PR #145, opened 2014-12-22
CDRIVER-489: Use the provided username, if available for X.509 auth

bjori:
Falls back to extracting the subject from the cert for backwards
compatibility


hanumantmk:
merged, thanks!

PR #146, opened 2014-12-23
Errors missing in some cases

alexeyvo:
Example:

```
mongoc_database_remove_user(db, 'not existing user', &err); // returns false but err is not written
```


hanumantmk:
Merged, thanks!

PR #147, opened 2014-12-23
CDRIVER-492: Report WriteConcern error on writes

bjori:
Renamed the field, and fixed the failure check.
There can only ever be one WriteConcernError, which is a bson object
containing the code (int), errInfo (object) and errmsg (cstring),
so there is no arrays to merge.


hanumantmk:
Looks good. Thanks for the help

PR #150, opened 2014-12-27
Clear some clang warnings

paulmelnikow:


src/mongoc/mongoc-counters.c
```diff
@@ -23,6 +23,7 @@
 #include <string.h>
 
 #ifdef BSON_OS_UNIX
+#undef MAX
```
hanumantmk: the ordering around this seems somewhat platform specific (sys/mman.h either doesn't pull MAX for me on linux, or has already been included).

I'm going to drop this change out and try to pull something together to get rid of all the non-prefixed MACROS, which as you've noted is the more robust approach.

hanumantmk:
I'll take care of MAX, MIN, RETURN, etc... in something more holistic.

Thanks for the contribution!

src/mongoc/mongoc-counters.c
paulmelnikow: Sounds good.

paulmelnikow:
Thanks!

PR #153, opened 2015-01-05
Change mongoc_gridfs_file_seek delta type from uint64_t to int64_t

alexeyvo:
If whence == SEEK_END it's required(according to implementation) to pass negative delta to achieve offset from end. In addition fseek's offset is declared as signed and if this function mimicing fseek it should be signed too.


hanumantmk:
@chergert, hate to bother, but do you have any idea if changing function parameters from signed to unsigned is an abi break?  I'm 99% sure there can't be a difference

* Sign extension would only come up if you were passing negative values in, which wasn't ok for the api before.
* Seeking more than 2^63 bits in a gridfs file isn't a real case.

I'll probably just take this as is, but I'm not 100% sure.

chergert:
I think it's just fine. LGTM

hanumantmk:
Thanks for all the gridfs attention!

PR #157, opened 2015-01-08
Fix for gridfs file seek if page == NULL

alexeyvo:
[Last gridfs fix](https://github.com/mongodb/mongo-c-driver/commit/e42287710db3ee3aa7412a804d32d78d119d6198)  don't cover case when page == NULL.

```
if (offset / file->chunk_size != file->pos / file->chunk_size) {
  ...
} else {
   _mongoc_gridfs_file_page_seek (file->page, offset % file->chunk_size);
}
```
If page == NULL, pos == 0 and offset < chunk_size it will try to seek in NULL page


samantharitter:
Hi @alexeyvo, I've created a [JIRA ticket for this issue](https://jira.mongodb.org/browse/CDRIVER-510), we'll address it from there for tracking purposes.  Thanks!

PR #159, opened 2015-01-09
Use void and size_t types for mongoc_iovec_t on windows

alexeyvo:
In [linux](http://linux.die.net/include/bits/uio.h) and [freebsd](http://fxr.watson.org/fxr/source/sys/_iovec.h) iovec struct defined as:
```
struct iovec
{
  void *iov_base;	/* Pointer to data.  */
  size_t iov_len;	/* Length of data.  */
};
```
in windows u_long defined as:
```
// WinSock2.h
typedef unsigned long   u_long;
```
so there is inconsistency between windows and other OS

### DIFF
--- a/doc/mongoc_iovec_t.page
+++ b/doc/mongoc_iovec_t.page
@@ -22,8 +22,8 @@
 #ifdef _WIN32
 typedef struct
 {
-   u_long  iov_len;
-   char   *iov_base;
+   void   *iov_base;
+   size_t  iov_len;
 } mongoc_iovec_t;
 #else
 typedef struct iovec mongoc_iovec_t;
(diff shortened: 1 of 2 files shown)


hanumantmk:
Unfortunately this isn't quite right.  We actually need mongoc_iovec_t to have a different layout on windows because we're passing the structure directly to WSASend, which expects a WSABUF.

The WSABUF is

```c
typedef struct __WSABUF {
  u_long   len;
  char FAR *buf;
} WSABUF, *LPWSABUF;
```

What was your motivation for the change, other than just making things match up?

alexeyvo:
Understood, just thought it can match. I'm closing this pull.

PR #167, opened 2015-01-30
Support for proper bulk updates

kaimast:
I am benchmarking mongodb and was confused that the c driver behaves differently than the python driver.
Integrating this patch can speed up bulk updates by an order of magnitude.

I am not sure if I broke any of the legacy stuff, make check runs fine though.


src/mongoc/mongoc-write-command.c
```diff
@@ -973,12 +1015,12 @@ _mongoc_write_result_merge_legacy (mongoc_write_result_t  *result, /* IN */
           * upsert if _id is not an ObjectId.
           */
          result->nUpserted += 1;
-         if (bson_iter_init_find (&iter, command->u.update.update, "_id") ||
```
kaimast: I am not really sure what this does and what I am supposed to do here....

src/mongoc/mongoc-write-command.c
```diff
@@ -31,8 +31,6 @@
  *    - Try to receive GLE on the stack (legacy only?)
  */
 
-
-#define MAX_INSERT_BATCH 1000
```
kaimast: This was hardcoded and did not respect the servers actual limits...

src/mongoc/mongoc-write-command.c
```diff
@@ -706,18 +781,65 @@ _mongoc_write_command_update (mongoc_write_command_t       *command,
...
-   bson_append_array_end (&cmd, &ar);
+   BSON_APPEND_BOOL (&cmd, "ordered", command->u.insert.ordered);
+
+   if ((command->u.update.updates->len < client->cluster.max_bson_size) &&
+       (command->u.update.updates->len < client->cluster.max_msg_size) &&
+       (command->u.update.n_updates <= max_update_batch)) {
+       // there is enough space to send the whole query at once...
+      BSON_APPEND_ARRAY (&cmd, "updates", command->u.update.updates);
+   } else {
+      bson_append_array_begin (&cmd, "updates", strlen("updates"), &ar);
```
kaimast: This is really ugly. We first built the updates and them split them up again. But this is the same behaviour as insert. 

It might make much more sense to split it up in mongoc_bulk_update_append. I tried this but for some reason the information about nodes might not yet exist in the client.

In general, this codebase has so many codepaths. Maybe try removing some of the deprecated stuff in the future?

chergert: FWIW, a lot of these crazy fallback paths were to deal with the multitude of changes that happened in every release from 1.8->2.6. Probably don't need to support back to 1.8, but in general good idea to test as far back as 2.0.

src/mongoc/mongoc-bulk-operation.c
```diff
@@ -219,8 +256,10 @@ mongoc_bulk_operation_update (mongoc_bulk_operation_t *bulk,
                               const bson_t            *document,
                               bool                     upsert)
 {
+   bool multi = true;
```
hanumantmk: Is there a code path where multi should be false?

kaimast: It was hardcoded before so I didn't change it. I could add multi as another function parameter (just like upsert)

src/mongoc/mongoc-write-command.c
```diff
@@ -109,6 +107,71 @@ _mongoc_write_command_insert_append (mongoc_write_command_t *command,
...
+   BSON_ASSERT (selector && update);
+
+   bson_init(&doc);
+   BSON_APPEND_DOCUMENT (&doc, "q", selector);
+   BSON_APPEND_DOCUMENT (&doc, "u", update);
+   BSON_APPEND_BOOL (&doc, "upsert", upsert);
+   BSON_APPEND_BOOL (&doc, "multi", multi);
+
+   pos = command->u.update.n_updates;
+   sprintf(id, "%i", pos);
```
hanumantmk: There's actually a bson_uint32_to_string() function which we've been preferring over sprintf for these kind of conversions

src/mongoc/mongoc-write-command.c
```diff
@@ -706,18 +914,66 @@ _mongoc_write_command_update (mongoc_write_command_t       *command,
...
-   BSON_APPEND_BOOL (&child, "multi", command->u.update.multi);
-   BSON_APPEND_BOOL (&child, "upsert", command->u.update.upsert);
-   bson_append_document_end (&ar, &child);
-   bson_append_array_end (&cmd, &ar);
+   BSON_APPEND_BOOL (&cmd, "ordered", command->u.insert.ordered);
+
+   if ((command->u.update.updates->len < client->cluster.max_bson_size) &&
+       (command->u.update.updates->len < client->cluster.max_msg_size) &&
+       (command->u.update.n_updates <= max_update_batch)) {
+       // there is enough space to send the whole query at once...
```
hanumantmk: Unfortunately, msvc doesn't like c++ style comments in c mode

src/mongoc/mongoc-write-command.c
hanumantmk: pre-2.6 mongod didn't return an _id for an upsert if _id wasn't an oid.  So we have to go dig the upserted id out of the update or the selector.

Basically, the conditional breaks down as:

* n == 1
  * we ran a single update
* updatedExisting is boolean false
  * we didn't update, so an upsert happened

Then we check for an _id field in:

* the update
* the selector

The || short circuits, so we append whichever one we find first.

The whole thing should work because we only get there if n == 1 and upserted wasn't an oid, which means the server couldn't have generated one and we should have provided one.

It's a bit convoluted, but that's the logic

hanumantmk:
The whole bulk api is fairly gnarly, so thanks for taking a stab at this.  I'm going to stare at the code a bit harder over a day or two, but it looks like a great start and I'd like to roll this in for the next release if I can.

Other than the fixme I responded to inline, everything looks pretty good.

I'll run this past some older versions of the server and I'll let you know what pops out of the woodwork.

Thanks for the PR!

bjori:
Cross referencing with https://jira.mongodb.org/browse/CDRIVER-523

PR #169, opened 2015-02-05
CDRIVER-521 run WT tests conditionally and with correct capitalization

samantharitter:


tests/test-mongoc-collection.c
```diff
@@ -2,6 +2,8 @@
 #include <mongoc.h>
 #include <mongoc-client-private.h>
 
+#include <stdio.h>
```
TylerBrock: I don't think we need this.

samantharitter: Whoops leftover from debugging good catch

tests/test-mongoc-collection.c
```diff
@@ -648,13 +697,22 @@ test_index_storage (void)
...
-   mongoc_index_opt_init (&opt);
-   mongoc_index_opt_wt_init (&wt_opt);
+   const char *engine;
 
    client = mongoc_client_new (gTestUri);
    ASSERT (client);
 
+   /* Skip unless we are on WT */
+   engine = storage_engine(client);
+   if (strcmp(engine, "wiredtiger") != 0 &&
```
TylerBrock: I believe it's more idiomatic to say `strcmp(engine, "wiredtiger") &&  strcmp(engine, ...)`

"wiredTiger" will be the only one in the future, not sure if all lower case is still a possibility (or ever was)? Can you confirm that it is?

samantharitter: This string is coming straight from the command line options passed to mongod.  I guess as long as we only test with 3.0.0-rc7 or later there's no reason to include "wiredtiger" because the db won't take it.  On earlier rc's, though, it's valid

tests/test-mongoc-collection.c
```diff
@@ -637,6 +639,53 @@ test_index_geo (void)
...
+   /* NOTE: this default will change eventually */
+   const char *unspecified = "mmapv1";
+
+   /* from getCmdLineOpts, if engine was specified, result looks like:
+      { "parsed" : { "storage" : { "engine" : "wiredtiger" }, ... }, ... } */
+   BSON_APPEND_INT32 (&cmd, "getCmdLineOpts", 1);
+   r = mongoc_client_command_simple(client, "admin", &cmd, NULL, &reply, &error);
+   ASSERT (r);
+
+   ASSERT (bson_iter_init_find (&iter, &reply, "parsed"));
```
TylerBrock: We can eliminate some code and variables in this method by using a dotted argument to bson_iter_init_find:

```c
if( bson_iter_init_find(&iter, &reply, "parsed.storage.engine") ) {
    /// found
}
```

tests/test-mongoc-collection.c
```diff
@@ -637,6 +639,53 @@ test_index_geo (void)
...
+   bson_iter_document(&iter, &len, &data);
+   ASSERT (bson_init_static(&parsed, data, len));
+   if (!bson_iter_init_find (&parsed_iter, &parsed, "storage")) {
+      return unspecified;
```
TylerBrock: If we return from this method early does do reply/cmd/etc get properly freed?

tests/test-mongoc-collection.c
```diff
@@ -637,6 +639,53 @@ test_index_geo (void)
...
+   }
+
+   engine = bson_iter_utf8(&storage_iter, NULL);
+
+   bson_destroy (&storage);
+   bson_destroy (&parsed);
+   bson_destroy (&reply);
+   bson_destroy (&cmd);
+
+   return engine;
```
TylerBrock: The returned pointer from this method will not point to valid memory.

tests/test-mongoc-collection.c
chergert: Id be more concerned that storage_engine() will absolutely always return you non-NULL (so you don't crash strcmp). Idiomatically, it really depends who you are talking to. I find the != show more clearly that you definitely didin't accidentally add a strcmp() and forgot that it was 0 based. Or my personal preference, !!strcmp(..).

Either way, I'd probably suggest a NULL check.

samantharitter:
@TylerBrock thanks!  This definitely is looking much better now, avoiding bson_iter craziness.

samantharitter:
Another patch, NULL check added, cleanup streamlined as per @chergert 's comment.

tests/test-mongoc-collection.c
```diff
@@ -637,24 +637,57 @@ test_index_geo (void)
...
+   mongoc_client_t *client = NULL;
    mongoc_index_opt_t opt;
    mongoc_index_opt_wt_t wt_opt;
    bson_error_t error;
    bool r;
    bson_t keys;
-
-   mongoc_index_opt_init (&opt);
-   mongoc_index_opt_wt_init (&wt_opt);
+   char *engine = NULL;
```
TylerBrock: This declaration has to go above the function calls.

tests/test-mongoc-collection.c
```diff
@@ -637,24 +637,57 @@ test_index_geo (void)
...
 
    client = mongoc_client_new (gTestUri);
    ASSERT (client);
 
+   /* Skip unless we are on WT */
+   engine = storage_engine(client);
+   if (!engine || strcmp("wiredTiger", engine) != 0) {
```
TylerBrock: I think engine is always non null given how storage_engine is written. We don't need the != 0.

samantharitter: Well, if ```bson_iter_utf8(&iter, NULL)``` within storage_engine returned NULL (with properly-formatted bson input I don't think it would) then we would get NULL from that method.  Better safe than sorry on that front?

chergert: I think the "safe" route is unnecessary. If you hit the bson_return_val_if_fail() macros, it indicates invalid API use (and therefore nothing we can do to help you, this is C after all).

tests/test-mongoc-collection.c
samantharitter: ...it is above the function calls (?)

TylerBrock: Ah sorry I misread the diff yesterday. Looks good.

samantharitter: No worries!  Thanks.

PR #171, opened 2015-02-08
Connection pool improvements

mschoenlaub:
In case we have more than "min_pool_size" connections in the pool, pushing one connection back to the pool will result in the LRU connection being destroyed.

Furthermore, unhealthy connections will be tried to repaired before returning them to the pool. 

A first simple test has been included.

### DIFF
--- a/src/mongoc/mongoc-client-pool.c
+++ b/src/mongoc/mongoc-client-pool.c
@@ -17,6 +17,7 @@
 
 #include "mongoc-counters-private.h"
 #include "mongoc-client-pool.h"
+#include "mongoc-client-private.h"
 #include "mongoc-queue-private.h"
 #include "mongoc-thread-private.h"
 #include "mongoc-trace.h"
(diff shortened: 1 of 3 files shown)


src/mongoc/mongoc-client-pool.c
```diff
@@ -193,23 +198,47 @@ mongoc_client_pool_push (mongoc_client_pool_t *pool,
...
-   _mongoc_queue_push_head(&pool->queue, client);
-   mongoc_cond_signal(&pool->cond);
-   mongoc_mutex_unlock(&pool->mutex);
+   mongoc_cond_signal (&pool->cond);
+   mongoc_mutex_unlock (&pool->mutex);
 
    EXIT;
 }
+
+uint
```
hanumantmk: I'd like to stick to explicitly sized types (out of stdint), or standard types.

So either returning size_t or uint32_t

src/mongoc/mongoc-client-pool.h
```diff
@@ -37,15 +37,31 @@ BSON_BEGIN_DECLS
...
+mongoc_client_pool_pop (mongoc_client_pool_t *pool);
+
+void
+mongoc_client_pool_push (mongoc_client_pool_t *pool,
+                         mongoc_client_t      *client);
+
+mongoc_client_t      *
+mongoc_client_pool_try_pop (mongoc_client_pool_t *pool);
+
+uint
```
hanumantmk: Same comment as above on uint

tests/test-mongoc-client-pool.c
```diff
@@ -38,9 +37,37 @@ test_mongoc_client_pool_try_pop (void)
...
+{
+   mongoc_client_pool_t *pool;
+   mongoc_client_t *client;
+   mongoc_uri_t *uri;
+   int c;
+
+      uri = mongoc_uri_new("mongodb://127.0.0.1?maxpoolsize=3&minpoolsize=2");
+      pool = mongoc_client_pool_new(uri);
+
+      mongoc_client_t *clients[3];
```
hanumantmk: msvc unfortunately demands variable declarations at the top of the block, so this needs to move up by the other declarations

hanumantmk:
Thanks for your interest in libmongoc and thanks for the patch.  It looks pretty reasonable to me.

I left some comments inline, but I'd also like to minimize some of the pure whitespace formatting.  If you can back those out I'm happy to take the patch the rest of the way.

Thanks!

PR #172, opened 2015-02-13
SDAM / SS review

samantharitter:
This branch represents our work to implement the new [Server Discovery and Monitoring](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst) and [Server Selection](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst) specs.

This branch still needs some fixing up before a merge, but it's going to be a big review so I figured it's best to start that ASAP.  @ajdavis and @amidvidy, would you please review?

Work yet to be done:

- normalize hostnames to lowercase
- ismasters must be run whenever a stream is created, and these should be used to avoid max_wire_version race condition
- document methods that take mutex, or that must be run while caller is holding mutex.

Thanks!


src/mongoc/mongoc-server-description.c
```diff
@@ -0,0 +1,483 @@
...
+         if (! BSON_ITER_HOLDS_ARRAY (&iter)) goto ERROR;
+         bson_iter_array (&iter, &len, &bytes);
+         bson_init_static (&sd->arbiters, bytes, len);
+      } else if (strcmp ("primary", bson_iter_key (&iter)) == 0) {
+         if (! BSON_ITER_HOLDS_UTF8 (&iter)) goto ERROR;
+         sd->current_primary = bson_iter_utf8 (&iter, NULL);
+      } else if (strcmp ("arbiterOnly", bson_iter_key (&iter)) == 0) {
+         if (! BSON_ITER_HOLDS_BOOL (&iter)) goto ERROR;
+         is_arbiter = bson_iter_bool (&iter);
+         /*
```
amidvidy: Jesse thinks that the below commented portions can just be removed.

src/mongoc/mongoc-server-description.c
```diff
@@ -0,0 +1,483 @@
...
+      } else if (is_secondary) {
+         sd->type = MONGOC_SERVER_RS_SECONDARY;
+      } else if (is_arbiter) {
+         sd->type = MONGOC_SERVER_RS_ARBITER;
+      } else {
+         sd->type = MONGOC_SERVER_RS_OTHER;
+      }
+   } else if (is_replicaset) {
+      sd->type = MONGOC_SERVER_RS_GHOST;
+   } else if (is_master) {
```
amidvidy: according to the SDAM spec, a server should have a type of STANDALONE if "No "msg: isdbgrid", no setName, and no "isreplicaset: true". - it isn't actually required to have `isMaster=true`.

src/mongoc/mongoc-server-description-private.h
```diff
@@ -0,0 +1,89 @@
...
+   bson_t                           last_is_master;
+
+   /* The following fields are filled from the last_is_master and are zeroed on
+    * parse.  So order matters here.  DON'T move set_name */
+   const char                      *set_name;
+   const char                      *connection_address;
+   char                            *error; // TODO: what type should this be?
+   mongoc_server_description_type_t type;
+   int32_t                          min_wire_version;
+   int32_t                          max_wire_version;
```
ajdavis: Let's talk about whether it's safe to use monitoring to discover wire version, max_write_batch_size, etc.: there were races in PyMongo that I solved by *not* using monitoring to find these values, and instead calling "ismaster" on each new socket during socket setup and using those per-socket values for formatting messages to send on that socket.

src/mongoc/mongoc-server-description.c
```diff
@@ -0,0 +1,483 @@
...
+   memset (&sd->set_name, 0, sizeof (*sd) - ((char*)&sd->set_name - (char*)sd));
+
+   while (bson_iter_next (&iter)) {
+      /* TODO: do we need to handle maxBsonObjSize */
+      if (strcmp ("ismaster", bson_iter_key (&iter)) == 0) {
+         if (! BSON_ITER_HOLDS_BOOL (&iter)) goto ERROR;
+         is_master = bson_iter_bool (&iter);
+      } else if (strcmp ("maxMessageSizeBytes", bson_iter_key (&iter)) == 0) {
+         if (! BSON_ITER_HOLDS_INT32 (&iter)) goto ERROR;
+         sd->max_write_batch_size = bson_iter_int32 (&iter);
```
ajdavis: The ismaster response's "maxWriteBatchSize" is the *number* of documents in a batch, not the number of bytes. It's distinct from maxMessageSizeBytes.

src/mongoc/mongoc-server-description.c
```diff
@@ -0,0 +1,483 @@
...
+      } else if (strcmp ("minWireVersion", bson_iter_key (&iter)) == 0) {
+         if (! BSON_ITER_HOLDS_INT32 (&iter)) goto ERROR;
+         sd->min_wire_version = bson_iter_int32 (&iter);
+      } else if (strcmp ("maxWireVersion", bson_iter_key (&iter)) == 0) {
+         if (! BSON_ITER_HOLDS_INT32 (&iter)) goto ERROR;
+         sd->max_wire_version = bson_iter_int32 (&iter);
+      } else if (strcmp ("msg", bson_iter_key (&iter)) == 0) {
+         if (! BSON_ITER_HOLDS_BOOL (&iter)) goto ERROR;
```
amidvidy: you should check that the value of the 'msg' field is 'isdbgrid'

src/mongoc/mongoc-server-description.c
```diff
@@ -0,0 +1,483 @@
...
+      sd->type = MONGOC_SERVER_STANDALONE;
+   } else {
+      goto ERROR;
+   }
+
+   if (sd->round_trip_time == -1) {
+      sd->round_trip_time = rtt_msec;
+   } else {
+      /* calculate round trip time as an exponentially weighted moving average
+       * with a weigt of ALPHA */
```
amidvidy: typo in comment

src/mongoc/mongoc-topology-description.c
```diff
@@ -0,0 +1,996 @@
...
+   bson_return_val_if_fail (topology, false);
+   bson_return_val_if_fail (sd, false);
+
+   mongoc_server_description_handle_ismaster (sd, ismaster_response, rtt_msec, error);
+
+   if (gSDAMTransitionTable[sd->type][topology->type]) {
+      gSDAMTransitionTable[sd->type][topology->type](topology, sd);
+   }
+
+   if (mongoc_topology_description_server_by_id (topology, sd->id)) {
```
amidvidy: can we simplify this if to `return mongoc_topology_description_server_by_id (topology, sd->id) != NULL`

or have no return type for this function since nobody cares if it returns false.

src/mongoc/mongoc-topology-description.c
```diff
@@ -0,0 +1,996 @@
...
+mongoc_topology_description_handle_ismaster (
+   mongoc_topology_description_t *topology,
+   mongoc_server_description_t   *sd,
+   const bson_t                  *ismaster_response,
+   int64_t                        rtt_msec,
+   bson_error_t                  *error)
+{
+   bson_return_val_if_fail (topology, false);
+   bson_return_val_if_fail (sd, false);
+
```
ajdavis: i think here you can return if `(!_mongoc_topology_description_topology_has_server(...)) { return false; }` here -- we should add a server to the topology description *before* we start monitoring it, so if it's not in the topology description that means the primary told us to stop monitoring it while we awaited this response, so we can ignore this response. this is a refactoring: you already check `_mongoc_topology_description_topology_has_server` where it matters elsewhere, but you can pull it up to this level. See:

https://github.com/mongodb/mongo-python-driver/blob/3.0-dev/pymongo/topology.py#L131

PyMongo doesn't even parse an ismaster response from a server if another thread has removed the server from the topology description while we awaited its response.

src/mongoc/mongoc-topology-description.c
```diff
@@ -0,0 +1,996 @@
...
+   else if (topology->set_name != server->set_name) {
+      _mongoc_topology_description_remove_server(topology, server);
+      return;
+   }
+
+   /* Begin monitoring any new servers that this server knows about */
+   // TODO TOPOLOGY_DESCRIPTION this loop.
+
+   /* If this server thinks there is a primary, label it POSSIBLE_PRIMARY */
+   _mongoc_topology_description_label_unknown_member(topology,
```
amidvidy: we should check that server->current_primary is non-null

src/mongoc/mongoc-topology-description.c
```diff
@@ -0,0 +1,996 @@
...
+                                                   const char *address,
+                                                   mongoc_server_description_type_t type)
+{
+   mongoc_server_description_t *server_iter;
+   mongoc_set_t *set = description->servers;
+   int i;
+
+   for (i = 0; i < set->items_len; i++) {
+      server_iter = set->items[i].item;
+      if (server_iter->connection_address == address &&
```
ajdavis: strcmp, not "==", and normalize case according to spec:

https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#hostnames-are-normalized-to-lower-case

amidvidy:
a general comment that having a test harness for the [SDAM yaml tests](https://github.com/mongodb/specifications/tree/master/source/server-discovery-and-monitoring/tests) would inspire a lot of confidence in the correctness of the topology/server description state machine.

src/mongoc/mongoc-server-description.c
```diff
@@ -0,0 +1,483 @@
...
+         bson_iter_array (&iter, &len, &bytes);
+         bson_init_static (&sd->hosts, bytes, len);
+      } else if (strcmp ("passives", bson_iter_key (&iter)) == 0) {
+         if (! BSON_ITER_HOLDS_ARRAY (&iter)) goto ERROR;
+         bson_iter_array (&iter, &len, &bytes);
+         bson_init_static (&sd->passives, bytes, len);
+      } else if (strcmp ("arbiters", bson_iter_key (&iter)) == 0) {
+         if (! BSON_ITER_HOLDS_ARRAY (&iter)) goto ERROR;
+         bson_iter_array (&iter, &len, &bytes);
+         bson_init_static (&sd->arbiters, bytes, len);
```
ajdavis: In several places we must ensure hostnames are all lowercased, perhaps it should be here, when we first encounter hostnames from the ismaster response. The other place we first encounter hostnames is when parsing the seed list.

src/mongoc/mongoc-topology-description.c
```diff
@@ -0,0 +1,996 @@
...
+ *--------------------------------------------------------------------------
+ */
+static void
+_mongoc_topology_description_update_unknown_with_standalone (mongoc_topology_description_t *topology,
+                                                             mongoc_server_description_t   *server)
+{
+   if (!_mongoc_topology_description_has_server(topology, server->connection_address)) return;
+
+   if (topology->servers->items_len > 1) {
+      /* This cluster contains other servers, it cannot be a standalone. */
```
amidvidy: nit: even though this logic is correct it would be nice if it directly mirrored the [SDAM pseudocode](http://emptysqua.re/server-discovery-and-monitoring.html#actions)

src/mongoc/mongoc-topology-description.c
```diff
@@ -0,0 +1,996 @@
...
+ *--------------------------------------------------------------------------
+ */
+static void
+_mongoc_topology_description_update_rs_with_primary_from_member (mongoc_topology_description_t *topology,
+                                                                 mongoc_server_description_t   *server)
+{
+   if (!_mongoc_topology_description_has_server(topology, server->connection_address)) return;
+
+   /* set_name should never be null here */
+   if (topology->set_name != server->set_name) {
```
ajdavis: strcmp, not "!="

src/mongoc/mongoc-topology-description.c
```diff
@@ -0,0 +1,996 @@
...
+   if (!_mongoc_topology_description_has_server(topology, server->connection_address)) return;
+
+   /*
+    * 'Server' can only be the primary if it has the right rs name.
+    */
+
+   if (!topology->set_name) {
+      int len = strlen(server->set_name) + 1;
+      topology->set_name = bson_malloc (len);
+      memcpy (topology->set_name, server->set_name, len);
```
ajdavis: Use bson_strdup?

src/mongoc/mongoc-topology-description.c
```diff
@@ -0,0 +1,996 @@
...
+   /*
+    *'Server' is the primary! Invalidate other primaries if found.
+    */
+
+   for (i = 0; i < set->items_len; i++) {
+      server_iter = set->items[i].item;
+      if (server_iter->id != server->id &&
+          server_iter->type == MONGOC_SERVER_RS_PRIMARY) {
+         mongoc_server_description_set_state(server, MONGOC_SERVER_UNKNOWN);
+         // TODO SDAM set other states to defaults?
```
ajdavis: No, the primary changed but the other servers' states are probably unchanged, just wait for the next scan to re-check them.

src/mongoc/mongoc-topology-description.c
```diff
@@ -0,0 +1,996 @@
...
+   /*
+    * Stop monitoring any servers primary doesn't know about.
+    */
+
+   for (i = 0; i < set->items_len; i++) {
+      server_iter = set->items[i].item;
+      if (!mongoc_server_description_has_rs_member(server, server_iter->connection_address)) {
+         _mongoc_topology_description_remove_server(topology, server_iter);
+      }
+   }
```
ajdavis: I have some concurrency concerns here that may or may not be valid. Can anything read the topology description in an inconsistent state while we update it here?

src/mongoc/mongoc-topology-description.c
```diff
@@ -0,0 +1,996 @@
...
+   /* set_name should never be null here */
+   if (topology->set_name != server->set_name) {
+      _mongoc_topology_description_remove_server(topology, server);
+   }
+
+   /* If there is no primary, label server's current_primary as the POSSIBLE_PRIMARY */
+   if (!_mongoc_topology_description_has_primary(topology)) {
+      _mongoc_topology_description_set_state(topology, MONGOC_TOPOLOGY_RS_NO_PRIMARY);
+      _mongoc_topology_description_label_unknown_member(topology,
+                                                        server->current_primary,
```
ajdavis: what if current_primary is NULL?

amidvidy:
@ajdavis and I are done with an initial review of the state machine transitions and isMaster parsing. We plan to cover overall architecture and concurrency next.

amidvidy:
I'm going to take a look at the async command logic. Please don't respond to comments I add on those files until I finish a full pass over.

src/mongoc/mongoc-async-cmd.c
```diff
@@ -0,0 +1,352 @@
...
+_mongoc_async_cmd_init_send (mongoc_async_cmd_t *acmd,
+                             const char         *dbname)
+{
+   bson_snprintf (acmd->ns, sizeof acmd->ns, "%s.$cmd", dbname);
+
+   acmd->rpc.query.msg_len = 0;
+   acmd->rpc.query.request_id = ++acmd->async->request_id;
+   acmd->rpc.query.response_to = 0;
+   acmd->rpc.query.opcode = MONGOC_OPCODE_QUERY;
+   acmd->rpc.query.flags = MONGOC_QUERY_SLAVE_OK;
```
amidvidy: if this is ever used for commands other than `isMaster`, this should probably be a parameter

src/mongoc/mongoc-topology-scanner.c
```diff
@@ -0,0 +1,396 @@
...
+                                          void                     *data,
+                                          bson_error_t             *error)
+{
+   mongoc_topology_scanner_node_t *node;
+
+   bson_return_if_fail (data);
+
+   node = (mongoc_topology_scanner_node_t *)data;
+   node->cmd = NULL;
+
```
amidvidy: we should explicitly handle a MONGOC_ASYNC_CMD_TIMEOUT here, even though we currently pass a NULL ismaster_response in that case.

amidvidy:
i've read the async code. It LGTM from a correctness perspective. Obviously storing pending commands in a sorted doubly-linked list wont scale incredibly high, but since we only have to deal with 50 replset nodes, it doesn't seem to be a problem. It might be worth using a more efficient datastructure in the future if we are e.g. monitoring 1000 mongoses or similar.

src/mongoc/mongoc-client-pool.c
```diff
@@ -80,6 +84,17 @@ mongoc_client_pool_new (const mongoc_uri_t *uri)
    pool->max_pool_size = 100;
    pool->size = 0;
 
+   topology = mongoc_topology_new(uri);
+   if (!topology) {
+      return NULL;
```
bjori: mongoc_topology_new always returns alloced topology, never NULL

src/mongoc/mongoc-client.c
```diff
@@ -672,9 +662,48 @@ _mongoc_client_recv_gle (mongoc_client_t  *client,
...
+   if (!(uri = mongoc_uri_new(uri_string))) {
+      return NULL;
+   }
+
+   topology = mongoc_topology_new(uri);
+
+   mongoc_uri_destroy (uri);
+
+   return _mongoc_client_new(uri_string, topology);
+}
```
bjori: I'm very confused by this function.
It seems to be doing exactly the same as _mongoc_client_new() - defaulting to mongodb://127.0.0.1 if no uri_string is given. And validate the uri_string.
I don't quite understand the need for this wrapper?

src/mongoc/mongoc-cluster.c
```diff
@@ -1804,6 +859,7 @@ _mongoc_cluster_auth_node_scram (mongoc_cluster_t      *cluster,
          goto failure;
       }
 
+      bson_subtype_t btype;
```
bjori: all variable declarations on top for MSCV pleasure.

src/mongoc/mongoc-cluster.c
```diff
@@ -1904,559 +962,436 @@ _mongoc_cluster_auth_node (mongoc_cluster_t      *cluster,
...
-   node->primary = false;
-   node->ping_avg_msec = -1;
-   memset(node->pings, 0xFF, sizeof node->pings);
-   node->pings_pos = 0;
-   node->stream = NULL;
-   node->stamp++;
-   bson_init(&node->tags);
-
-   stream = _mongoc_client_create_stream (cluster->client, hosts, error);
+   stream = _mongoc_client_create_stream(cluster->client, &description->host, error);
```
bjori: pedantic warning: cs. Space before (.
likely in other places too :cookie:

src/mongoc/mongoc-cluster.c
```diff
@@ -1904,559 +962,436 @@ _mongoc_cluster_auth_node (mongoc_cluster_t      *cluster,
...
-                                                   error))) {
-         MONGOC_INFO("%s: Lost connection during ping.",
-                     host.host_and_port);
-         _mongoc_cluster_node_destroy (&cluster->nodes[i]);
-         continue;
-      }
+   cluster->uri = mongoc_uri_copy(uri);
+   cluster->client = client;
+   cluster->max_msg_size = 1024 * 1024 * 48;
+   cluster->max_bson_size = 1024 * 1024 * 16;
```
bjori: max_bson_size should be set to 4194304, and max_message_size to max_bson_size * 2

src/mongoc/mongoc-cluster.c
```diff
@@ -1904,559 +962,436 @@ _mongoc_cluster_auth_node (mongoc_cluster_t      *cluster,
...
+                             mongoc_uri_get_auth_mechanism(uri));
 
-      _mongoc_cluster_node_track_ping(&cluster->nodes[i], ping);
+   /* TODO consider making these configurable */
+   cluster->sec_latency_ms = 15;
+   cluster->sockettimeoutms = DEFAULT_SOCKET_TIMEOUT_MSEC;
 
-      i++;
-   }
+   cluster->nodes = mongoc_set_new(8, _mongoc_cluster_stream_dtor, NULL);
```
bjori: Why 8?
Is that some sort of global average ? :)

src/mongoc/mongoc-read-prefs.c
```diff
@@ -37,6 +37,7 @@ mongoc_read_prefs_new (mongoc_read_mode_t mode)
 mongoc_read_mode_t
 mongoc_read_prefs_get_mode (const mongoc_read_prefs_t *read_prefs)
 {
+   assert(read_prefs);
```
bjori: ? that doesn't look like it belongs here

src/mongoc/mongoc-cluster.c
```diff
@@ -1904,559 +962,436 @@ _mongoc_cluster_auth_node (mongoc_cluster_t      *cluster,
...
-      ret = _mongoc_cluster_reconnect_sharded_cluster (cluster, error);
-      RETURN (ret);
-   default:
-      break;
+   /* we can run queries on secondaries if given the right read mode */
+   if (optype == MONGOC_SS_WRITE &&
+       opcode == MONGOC_OPCODE_QUERY) {
+      read_mode = mongoc_read_prefs_get_mode(read_prefs);
+      if ((read_mode & MONGOC_READ_SECONDARY) != 0) {
+         optype = MONGOC_SS_READ;
```
bjori: what about secondary_preferred?

bjori: Nevermind.
MONGOC_READ_SECONDARY_PREFERRED = (1 << 2) | MONGOC_READ_SECONDARY

src/mongoc/mongoc-cluster.c
```diff
@@ -2756,162 +1635,72 @@ _mongoc_cluster_sendv (mongoc_cluster_t             *cluster,
...
+   server_id = mongoc_cluster_select(cluster, rpcs, rpcs_len,
+                                     write_concern, read_prefs, error);
+   if (server_id < 1) {
+      return server_id;
    }
 
-   RETURN(node->index + 1);
+   if(mongoc_cluster_sendv_to_server(cluster, rpcs, rpcs_len, server_id,
+                                     write_concern, error)) {
+      return true;
```
bjori: the docblock said:
+ * Server id, or 0.

src/mongoc/mongoc-client.c
samantharitter: It's because of the way the topology object must be handled.  Client creation either happens by users calling ```mongoc_client_new()``` directly, or happens through a client pool, which calls the internal ```_mongoc_client_new()```.  When clients are created through a pool, they share a single topology object that the pool owns.  When clients are created directly, they each create and own separate topology objects.  This wrapper handles the creation of the topology object and then calls the internal method to complete client setup.

src/mongoc/mongoc-server-description.c
samantharitter: done

src/mongoc/mongoc-server-description.c
samantharitter: done

src/mongoc/mongoc-topology-description.c
samantharitter: done

src/mongoc/mongoc-topology-description.c
samantharitter: done

src/mongoc/mongoc-server-description.c
samantharitter: I'm going to hold off on this until SPEC-146 is resolved, I've put a note about it in the description above.

src/mongoc/mongoc-write-command.c
```diff
@@ -780,13 +809,15 @@ _mongoc_write_command_execute (mongoc_write_command_t       *command,       /* I
 
    command->hint = hint;
 
-   node = &client->cluster.nodes [hint - 1];
-   mode = SUPPORTS_WRITE_COMMANDS (node);
+   server = mongoc_topology_server_by_id(client->topology, hint);
+   mode = (server->min_wire_version <= 2 && server->max_wire_version >= 2);
```
bjori: Thats a very magical 2. I liked the macro better :/
Maybe define 2 as WRITE_COMMANDS or something?

src/mongoc/mongoc-topology.c
```diff
@@ -0,0 +1,664 @@
...
+ */
+void
+mongoc_topology_release (mongoc_topology_t *topology)
+{
+   bson_return_if_fail(topology);
+
+   mongoc_mutex_lock (&topology->mutex);
+   topology->users--;
+   if (topology->users < 1) {
+      mongoc_mutex_unlock (&topology->mutex);
```
bjori: Don't you want to unlock it just before mongoc_mutex_destroy() instead?

src/mongoc/mongoc-topology-description.c
samantharitter: done

src/mongoc/mongoc-async-cmd.c
samantharitter: This is true.  To simplify/expedite the SDAM work I think making async_cmd a general-purpose tool should be left for another time.  I've opened CDRIVER-548 to make a record of this.

src/mongoc/mongoc-topology-scanner.c
samantharitter: done

src/mongoc/mongoc-client-pool.c
samantharitter: done

src/mongoc/mongoc-cluster.c
samantharitter: done

src/mongoc/mongoc-read-prefs.c
samantharitter: done

src/mongoc/mongoc-write-command.c
samantharitter: done

src/mongoc/mongoc-topology.c
samantharitter: Do you mean after?  If the refcount has gone to 0, it shouldn't really matter, because we're the only writer/reader left.

hanumantmk: I think it's that there's a possible race if two threads both have the topology pointer and one releases (dropping the count to 0), as 1 grabs (taking the count to 1 as we enter destroy).

That said, I don't think there's anyway to work around something going wrong in that case (if you moved the unlock later, you'd just attempt to lock the free'd mutex).  In practice we gate new acquiring threads through the pool, so we really aren't getting anything out of the refcount.  thread_join() is the real synchronization primitive we're relying on here.

samantharitter: For the refcount to reach 0 the client pool has to get destroyed.  The client pool holds a reference to topology, so as long as all spawned clients are checked into the pool before the pool is destroyed, this shouldn't be a race.  I can imagine that horror scenarios are possible, when some client somehow outlives the pool object (unlikely because the pool destroys its clients before releasing the topology), but if that is happening then the race condition here is just a symptom of larger problems.  

If we aren't using a pool, then clients each own their own topology object, and run single-threaded so that's a non-issue.

bjori: "It shall be safe to destroy an initialized mutex that is unlocked. Attempting to destroy a locked mutex results in undefined behavior. " -- http://linux.die.net/man/3/pthread_mutex_destroy

But I meant calling mongoc_mutex_unlock() as last thing before mongoc_mutex_destroy().

I imagine if something happens between the unlock and the destroy, like @hanumantmk is suggesting
Or, someone in the future accidentally calls mongoc_topology_destroy with a locked mutex...

Its likely fine as-is, was just curious why the lock was released and then lot of work is done and then the mutex destroyed - which upon cursorary thought looked a little dangerous. Given it is about to get destroyed anyway, I figured there is no harm in keeping the lock until its destroyed.

src/mongoc/mongoc-cursor.c
```diff
@@ -208,7 +208,7 @@ _mongoc_cursor_new (mongoc_client_t           *client,
 
    /* we can't have exhaust queries with sharded clusters */
    if ((flags & MONGOC_QUERY_EXHAUST) &&
-       (client->cluster.mode == MONGOC_CLUSTER_SHARDED_CLUSTER)) {
+       (client->topology->description.type == MONGOC_TOPOLOGY_SHARDED)) {
```
amidvidy: do you need to hold the topology lock here?

src/mongoc/mongoc-cursor.c
```diff
@@ -937,10 +935,19 @@ _mongoc_cursor_get_host (mongoc_cursor_t    *cursor,
...
-   host->next = NULL;
-}
+   description = mongoc_topology_server_by_id(cursor->client->topology, cursor->hint);
+   if (!description) {
+      MONGOC_WARNING("%s(): Invalid cursor hint, no matching host.",
+                     __FUNCTION__);
+      return;
+   }
 
+   *host = description->host;
```
amidvidy: this is pretty minor, but you should null out host.next so you don't leave it dangling.

src/mongoc/mongoc-write-command.c
```diff
@@ -698,13 +718,22 @@ _mongoc_write_command_update (mongoc_write_command_t       *command,
...
+   if (!server) {
+      EXIT;
+   }
+
+   if ((server->min_wire_version == 0) &&
        !_mongoc_write_concern_needs_gle (write_concern)) {
       _mongoc_write_command_update_legacy (command, client, hint, database,
                                            collection, write_concern, result,
                                            error);
+      mongoc_server_description_destroy (server);
```
amidvidy: maybe move this to a cleanup label that would go on line 760 and then GOTO(cleanup) here? It would avoid the second call to mongoc_server_description_destroy, which is a bit hacky IMO.

src/mongoc/mongoc-stream.c
```diff
@@ -271,7 +270,53 @@ mongoc_stream_get_base_stream (mongoc_stream_t *stream) /* IN */
       return stream->get_base_stream (stream);
    }
 
-   return NULL;
+   return stream;
```
amidvidy: why are we not returning NULL now?

amidvidy:
We should be super clear to users of the driver that they *should be using the client pool if they are going to be making multiple connections to the same replica set*, to avoid redundant background threads monitoring the same set.

src/mongoc/mongoc-cluster.c
```diff
@@ -1904,560 +957,557 @@ _mongoc_cluster_auth_node (mongoc_cluster_t      *cluster,
...
-                     "Invalid host list supplied.");
-      RETURN(false);
-   }
+   BSON_ASSERT (cluster);
 
-   cluster->last_reconnect = bson_get_monotonic_time();
+   MONGOC_DEBUG ("Adding new server to cluster: %s", sd->connection_address);
 
-   node = &cluster->nodes[0];
+   /* if the topology is single-threaded, we share its streams */
```
amidvidy: do we need the topology lock here as well?

samantharitter: In theory ```topology->single_threaded``` shouldn't change but yes, we probably should get the lock to read this attribute.

src/mongoc/mongoc-cluster.c
```diff
@@ -1904,560 +957,557 @@ _mongoc_cluster_auth_node (mongoc_cluster_t      *cluster,
...
 
-      if (-1 == (ping = _mongoc_cluster_ping_node (cluster,
-                                                   &cluster->nodes[i],
-                                                   error))) {
-         MONGOC_INFO("%s: Lost connection during ping.",
-                     host.host_and_port);
-         _mongoc_cluster_node_destroy (&cluster->nodes[i]);
-         continue;
-      }
+   cluster->sec_latency_ms = 15;
```
amidvidy: is this used anywhere?

src/mongoc/mongoc-cluster.c
```diff
@@ -1904,560 +957,557 @@ _mongoc_cluster_auth_node (mongoc_cluster_t      *cluster,
...
+ * mongoc_cluster_destroy --
+ *
+ *       Clean up after @cluster and destroy all active connections.
+ *       All resources for @cluster are released.
+ *
+ * Returns:
+ *       None.
+ *
+ * Side effects:
+ *       Everything.
```
amidvidy: um

src/mongoc/mongoc-cluster.c
samantharitter: good catch, it should be used in _mongoc_cluster_select_by_optype() as the latency value, but it is not

samantharitter:
This work has been merged with master and pulled into the new 1.2.0-dev branch.

PR #174, opened 2015-02-18
CDRIVER-532 - Coalesce small vecrotized TLS writes

hanumantmk:
Rather than calling BIO_write multiple times for each iovec_t in a call
to mongoc_stream_tls_writev(), buffer them and send them all at once.
Attempt to "do the right thing"tm, when buffering provides no
performance benefit.

### DIFF
--- a/src/mongoc/mongoc-stream-tls.c
+++ b/src/mongoc/mongoc-stream-tls.c
@@ -42,6 +42,8 @@
 #undef MONGOC_LOG_DOMAIN
 #define MONGOC_LOG_DOMAIN "stream-tls"
 
+#define MONGOC_STREAM_TLS_BUFFER_SIZE 4096
+
 
 /**
  * mongoc_stream_tls_t:
@@ -448,6 +450,53 @@ _mongoc_stream_tls_flush (mongoc_stream_t *stream)
 }
 
 
+static ssize_t
+_mongoc_stream_tls_write (mongoc_stream_tls_t *tls,
+                          char                *buf,
+                          size_t               buf_len)
+{
+   ssize_t ret;
+
+   int64_t now;
+   int64_t expire = 0;
+
+   BSON_ASSERT (tls);
+   BSON_ASSERT (buf);
+   BSON_ASSERT (buf_len);
+
+   if (tls->timeout_msec >= 0) {
+      expire = bson_get_monotonic_time () + (tls->timeout_msec * 1000UL);
+   }
+
+   ret = BIO_write (tls->bio, buf, buf_len);
+
+   if (ret < 0) {
+      return ret;
+   }
+
+   if (expire) {
+      now = bson_get_monotonic_time ();
+
+      if ((expire - now) < 0) {
+         if (ret < buf_len) {
+            mongoc_counter_streams_timeout_inc();
+#ifdef _WIN32
+            errno = WSAETIMEDOUT;
+#else
+            errno = ETIMEDOUT;
+#endif
+         }
+
+         tls->timeout_msec = 0;
+      } else {
+         tls->timeout_msec = (expire - now) / 1000L;
+      }
+   }
+
+   return ret;
+}
+
+
 /*
  *--------------------------------------------------------------------------
  *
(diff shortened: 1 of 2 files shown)


src/mongoc/mongoc-stream-tls.c
```diff
@@ -495,40 +497,131 @@ _mongoc_stream_tls_writev (mongoc_stream_t *stream,
...
+              *   - buffered bytes already
+              *   - another iovec to send after this one and we don't have more
+              *     bytes to send than the size of the buffer.
+              *
+              * copy into the buffer */
 
-         if (write_ret < 0) {
-            return write_ret;
-         }
+             bytes = BSON_MIN (iov[i].iov_len - iov_pos, buf_end - buf_tail);
```
samantharitter: this only runs if ```(buf_end - buf_tail) > (iov[i].iov_len - iov_pos)``` so you don't need MIN here, you can just take ```(iov[i].iov_len - iov_pos)```

src/mongoc/mongoc-stream-tls.c
```diff
@@ -495,40 +497,131 @@ _mongoc_stream_tls_writev (mongoc_stream_t *stream,
       iov_pos = 0;
 
       while (iov_pos < iov[i].iov_len) {
-         write_ret = BIO_write (tls->bio, (char *)iov[i].iov_base + iov_pos,
-                              (int)(iov[i].iov_len - iov_pos));
+         if (buf && (buf_head != buf_tail ||
+             ((i + 1 < iovcnt) && ((buf_end - buf_tail) > (iov[i].iov_len - iov_pos))))) {
```
samantharitter: ...do you want >= here on the second conditional?

src/mongoc/mongoc-stream-tls.c
```diff
@@ -495,40 +497,131 @@ _mongoc_stream_tls_writev (mongoc_stream_t *stream,
...
+             /* If we have:
+              *   - buffered bytes already
+              *   - another iovec to send after this one and we don't have more
```
samantharitter: add an ```or``` here just for clarity

src/mongoc/mongoc-stream-tls.c
amidvidy: i know you two may have discussed IRL, but why is this not >= ?

src/mongoc/mongoc-stream-tls.c
```diff
@@ -495,40 +497,131 @@ _mongoc_stream_tls_writev (mongoc_stream_t *stream,
...
+                   /* we timed out, so send back what we could send */
+
+                   return ret;
+                }
+
+                /* reset all the pointers */
+                buf_tail = buf_head = buf;
+             }
+          } else {
+            /* Didn't buffer, so just write it through */
```
amidvidy: Jesse and I think that it would make the logic clearer if this was factored out into a separate function named `_mongoc_stream_tls_write` or similar, instead of calling itself with a NULL buf.

src/mongoc/mongoc-stream-tls.c
samantharitter: If you haven't satisfied the first part of the or clause (```buf_head != buf_tail```) then you have not buffered anything yet.  If the size of the iov is equal to the buffer length, then you'd just end up buffering the message and sending it immediately because it filled the buffer, so don't bother buffering it.

src/mongoc/mongoc-stream-tls.c
```diff
@@ -495,40 +497,131 @@ _mongoc_stream_tls_writev (mongoc_stream_t *stream,
...
             }
+
+            ret += write_ret;
+            iov_pos += write_ret;
          }
+      }
+   }
+
+   if (buf && buf_head != buf_tail) {
+      /* If we have any bytes buffered, send */
```
amidvidy: can you clarify how we would end up with buffered bytes that haven't been written after the loop terminates?

tests/ssl-test.c
```diff
@@ -9,6 +9,8 @@
 
 #define TIMEOUT 1000
 
+#define MANY_SIZE 2000
```
amidvidy: could we name something more descriptive?

amidvidy:
Jesse and I reviewed this together:

as a general comment it would be nice to have more test cases that try with different buffer sizes so that all the possible code paths are executed. This is a pretty tricky function as written and it takes a bit of time to understand how the logic flows.

src/mongoc/mongoc-stream-tls.c
amidvidy: Cool, that makes sense. Thanks @samantharitter. @hanumantmk I would add this explanation into the comments.

hanumantmk:
Refactored the actual write apart to make everything less ridiculous.

src/mongoc/mongoc-stream-tls.c
hanumantmk: You only flush the buffer if it's full in the main loop.  So if your vector adds up to less than 4k you'll never actually send in the main loop (you'll always accumulate because you had something in the buffer, including on the last iovec)

src/mongoc/mongoc-stream-tls.c
hanumantmk: done

src/mongoc/mongoc-stream-tls.c
hanumantmk: done

src/mongoc/mongoc-stream-tls.c
amidvidy: Ah, thanks.

ajdavis:
LGTM.

amidvidy:
LGTM as well.

PR #177, opened 2015-02-21
Client pool improvements

mschoenlaub:
In case we have more than "min_pool_size" connections in the pool, pushing one connection back to the pool will result in the LRU connection being destroyed.

Furthermore, unhealthy connections will be tried to repaired before returning them to the pool.

This is basically a cleaned up version of PR #171

### DIFF
--- a/src/mongoc/mongoc-client-pool.c
+++ b/src/mongoc/mongoc-client-pool.c
@@ -19,6 +19,8 @@
 #include "mongoc-client-pool.h"
 #include "mongoc-queue-private.h"
 #include "mongoc-thread-private.h"
+#include "mongoc-cluster-private.h"
+#include "mongoc-client-private.h"
 #include "mongoc-trace.h"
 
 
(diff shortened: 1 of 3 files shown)


hanumantmk:
Thanks for cleaning this up, just a couple of small questions:

* You had a test bundled with this that seems to have gotten lost, was that intentional?
* Are you using mongoc_client_pool_get_size() anywhere?  I generally ere on the side of not expanding the API surface where possible, but I'm willing if you have a motivating example.  Also, If I do take it there will also be a little dance I'll have to do for a 1.1.2 release (I'll take the fix, but not publicly export the poll_get_size() symbol and the symbol will make it into 1.2.0)

mschoenlaub:
Hey,

nope, this was not intentional, it just got lost. It's back in the PR now. 
Actually this test is the only location where I use mongoc_client_pool_get_size() right now.
Just thought that getting the actual size of a pool could prove useful one day.

hanumantmk:
Thanks for your pr!  This will go out with 1.1.2 within the next week

PR #178, opened 2015-02-25
client: Fix setting of the connectTimeoutMS value

ac000:
When specifying the connectTimeoutMS option in the URI e.g

    ?connectTimeoutMS=250

this would be added internally under the name as specified, ie in this
case connectTimeoutMS however when mongoc_client_connect_tcp() comes to
look for this setting, it does so case sensitively as all lowercase, i.e

    if ((options = mongoc_uri_get_options (uri)) &&
        bson_iter_init_find (&iter, options, "connecttimeoutms") &&

and won't find the setting and will use the default connectTimeoutMS
value instead.

It should use bson_iter_init_find_case() instead for a case insensitive
search.

Signed-off-by: Andrew Clayton <andrew@digital-domain.net>


hanumantmk:
Thanks for the PR!

PR #180, opened 2015-02-26
CDRIVER-546: Use the error message from the server rather then cryptic h...

bjori:
...ome grown one


hanumantmk:
Thanks for the fix!

PR #181, opened 2015-02-27
CDRIVER-536 hostnames are normalized to lowercase

samantharitter:
@hanumantmk and @ajdavis if you get a minute today, would you please review?  Note that this only canonicalizes hostnames within URIs.  Once it's merged in the SDAM branch will need a bit more work to complete the task.

### DIFF
--- a/src/mongoc/mongoc-uri.c
+++ b/src/mongoc/mongoc-uri.c
@@ -23,7 +23,7 @@
 #include "mongoc-host-list-private.h"
 #include "mongoc-log.h"
 #include "mongoc-socket.h"
-#include "mongoc-uri.h"
+#include "mongoc-uri-private.h"
 
 
 #if defined(_WIN32) && !defined(strcasecmp)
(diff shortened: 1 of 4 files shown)


src/mongoc/mongoc-uri.c
```diff
@@ -55,6 +55,21 @@ mongoc_uri_do_unescape (char **str)
...
+                           int len)
+{
+   bson_unichar_t c;
+   const char *iter;
+   char *buf_iter;
+
+   for (iter = src, buf_iter = buf;
+        iter && *iter && (c = bson_utf8_get_char(iter)) && buf_iter - buf < len;
+        iter = bson_utf8_next_char(iter), buf_iter++) {
+      *buf_iter = tolower(c);
```
hanumantmk: tolower returns an int.  You'll need another strategy to move through buf_iter that isn't bytewise if the input isn't ascii

Also, unsure if migrating to lower case changes the size of the output.

... honestly, I'm not sure I know how this is supposed to work.

samantharitter: hm ok let me do some more research then.

hanumantmk:
LGTM

PR #182, opened 2015-03-01
Updated readme to include info for VS2013

gideondsouza:
Visual Studio 2013 has some new shortcuts and names, so I updated the readme accordingly.
The update includes (1) The cmake directive needed (2) The fact the Visual Studio Command Prompt has new name. 
Both these changes confuse many developers on a general basis.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -122,6 +122,11 @@ To build on Windows Vista or newer with Visual Studio 2010, do the following::
   msbuild.exe ALL_BUILD.vcxproj
   msbuild.exe INSTALL.vcxproj
 
+Note: For Visual Studio 2013
+
+1. You'll need to say ``-G "Visual Studio 12 2013"``. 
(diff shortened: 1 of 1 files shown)


hanumantmk:
You can probably continue using "Visual Studio 10 Win64", it's just a matter of which version of the project files you're putting out.

I wonder if we'd be better off just referring to the cmake generators documentation here (http://www.cmake.org/cmake/help/v3.1/manual/cmake-generators.7.html#ide-build-tool-generators)

gideondsouza:
hey @hanumantmk,

>You can probably continue using "Visual Studio 10 Win64"
Nope, if you put "Visual Studio 10 Win64" and have VS2013 then cmake throws an error. Which is what I googled and then came across the solution that I've updated the docs with.

I don't remember the error but I can try and reproduce it and put it here.

ajdavis:
Fixed in https://github.com/mongodb/mongo-c-driver/commit/7bd6891aa14a6f0e8811b0caf2de3ca0cbf843ef.

PR #185, opened 2015-03-05
client pool: Add some missing locking in _push()

ac000:
Commit a8c1da4 ("Client pool tries to repair unhealthy connections...")
introduced more code to mongoc_client_pool_push() operating on the
connection pool. However locking was missing on the first if () section
which deals with various aspects of the connection pool and requires
locking.

Protect the if () with a mongoc_mutex_lock/unlock using pool->mutex.

Signed-off-by: Andrew Clayton <andrew@digital-domain.net>


ac000:
Going to close this pull request. I have another client pool fix that I'll put with this one in a more appropriately named branch.

PR #186, opened 2015-03-05
Fixed link/name to the current release

elfranne:


ajdavis:
Thanks @elfranne I fixed this problem for the latest release (1.1.4) and updated our release instructions so we stop making this mistake.

PR #187, opened 2015-03-05
client pool: Fix some missing locking and a deadlock

ac000:
client pool: Add some missing locking in _push()

This is adds a missing mongoc_mutex_lock/unlock pair in
mongoc_client_pool_push()

client pool: Prevent a deadlock in mongoc_client_pool_pop()

There is situation a where a client could deadlock in 
mongoc_client_pool_pop() which this fix's.


hanumantmk:
Thanks for contribution!

PR #189, opened 2015-03-11
CDRIVER-560: Include private header in mongoc-client-pool.c.

ajdavis:


src/mongoc/mongoc-client-pool.c
```diff
@@ -17,6 +17,7 @@
 
 #include "mongoc-counters-private.h"
 #include "mongoc-client-pool.h"
+#include "mongoc-client-pool-private.h"
```
samantharitter: mongoc-client-pool-private.h includes mongoc-client-pool.h, so you can just include the private header here.

ajdavis:
Updated to avoid duplicate includes.

PR #190, opened 2015-03-11
CDRIVER-529: Spurious "unstable version" warning from configure.

ajdavis:
We set an odd micro version (1.1.1, 1.1.3, etc.) between releases to mark the
code unstable, but the configure script was checking the minor version, not the
micro, to decide whether to warn.

### DIFF
--- a/build/autotools/PrintBuildConfiguration.m4
+++ b/build/autotools/PrintBuildConfiguration.m4
@@ -1,6 +1,6 @@
 AC_OUTPUT
 
-if test $(( ${MONGOC_MINOR_VERSION} % 2 )) -eq 1; then
+if test $(( ${MONGOC_MICRO_VERSION} % 2 )) -eq 1; then
 cat << EOF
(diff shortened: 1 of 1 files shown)


chergert:
We were following typical semantic versioning. (See cairo project for great example of how to do this for long term tracking).

Micro versions are only used for patch releases (ie: hotfix/bugfix release). New development always happens on a bumped minor version (1.3 is unstable dev for 1.4).

hanumantmk:
I don't really like MAJOR.(IF_ODD_UNSTABLE).PATCH release numbers.  To some extent, they seem wasteful, as a rc/dev tag would work just as well.  They also don't offer an obvious scheme for unstable patch releases.

Rolling it into the patch number, if a bit non-standard, at least offers a way to keep master in an unstable place between patch versions, which I like.  But it's just as much of a hack... 

What are your thoughts on just having a -dev?  That seems like the actual semver solution.

PR #191, opened 2015-03-11
CDRIVER-568 normalize hostnames to lowercase in SDAM code

samantharitter:


src/mongoc/mongoc-server-description.c
```diff
@@ -179,13 +189,33 @@ mongoc_server_description_has_rs_member(mongoc_server_description_t *server,
...
          while (bson_iter_next (&member_iter)) {
-            if (strcmp (address, bson_iter_utf8 (&member_iter, NULL)) == 0) {
+            /* TODO we convert to lowercase each time we do this, it might be
+               better to store lowercased member names instead */
+            name = bson_iter_utf8 (&member_iter, NULL);
+            name_len = strlen (name);
+
+            /* re-use previously malloced buffer if we can */
+            if (name_len > buf_len) {
+               if (buf) bson_free (buf);
```
hanumantmk: You can just malloc two buffers the size of address.  If the member you're looking at is larger than address (that second parameter to bson_iter_utf8()), you don't need to lower case it to need it's different.  

More generally, you only need to strcmp strings that are the same size.  Otherwise they're obviously different if they're different lengths

hanumantmk:
I'm sorry.  But let's pivot and just use strcasecmp.  You can grab a define that'll get that functionality on windows out of mongoc-uri.c

samantharitter:
Replaced by https://github.com/mongodb/mongo-c-driver/pull/200

PR #193, opened 2015-03-12
prerelease version

hanumantmk:
add prerelease version to get away from major/minor/patch encoding
instability


hanumantmk:
@chergert 

What do you think of this, as an alternative to even odd minor/patch versions?

chergert:
Not a fan. I far prefer semantic versioning with even/odd minor and stability branches (ie: branch 1.2 as part of tagging 1.2. patch releases are backported and tagged off 1.2 branch). then development continues as 1.3.

hanumantmk:
Any particular reason why you prefer it?

I don't love the fact that you can't always determine stability just by looking at the version numbers (or it means I can never have unstable branches for patch releases).  I.e. 1.2.0 in the version file could mean 1.2.0 stable, or it could mean the 1.2.0 branch, but only lightly tested and pre-bump to 1.2.1.

It also seems like a workflow more appropriate to projects with really long release cycles.  I can see the value if you plan to use patch numbers to indicate progress, or if you make external releases you want to organize around, but I don't see it for the short internal sprints we're working off of.

chergert:
How about an odd micro number also indicating development?

1.3.x => stream to 1.4
1.2.1 => post-release bump to 1.2.2

PR #194, opened 2015-03-13
CDRIVER-580: fsync or j in write concern can imply GLE

jmikola:
https://jira.mongodb.org/browse/CDRIVER-580

### DIFF
--- a/src/mongoc/mongoc-uri.c
+++ b/src/mongoc/mongoc-uri.c
@@ -26,6 +26,7 @@
 #include "mongoc-log.h"
 #include "mongoc-socket.h"
 #include "mongoc-uri-private.h"
+#include "mongoc-write-concern-private.h"
 
 
 #if defined(_WIN32) && !defined(strcasecmp)
@@ -737,13 +738,13 @@ _mongoc_uri_build_write_concern (mongoc_uri_t *uri) /* IN */
          value = bson_iter_int32 (&iter);
 
          switch (value) {
-         case -1:
-            mongoc_write_concern_set_w (write_concern,
-                                        MONGOC_WRITE_CONCERN_W_ERRORS_IGNORED);
-            break;
-         case 0:
-            mongoc_write_concern_set_w (write_concern,
-                                        MONGOC_WRITE_CONCERN_W_UNACKNOWLEDGED);
+         case MONGOC_WRITE_CONCERN_W_ERRORS_IGNORED:
+         case MONGOC_WRITE_CONCERN_W_UNACKNOWLEDGED:
+            /* Warn on conflict, since write concern will be validated later */
+            if (mongoc_write_concern_get_journal(write_concern)) {
+               MONGOC_WARNING("Journal conflicts with w value [w=%d].", value);
+            }
+            mongoc_write_concern_set_w(write_concern, value);
             break;
          default:
             if (value > 0) {
@@ -789,9 +790,14 @@ mongoc_uri_new (const char *uri_string)
       return NULL;
    }
 
-   uri->str = bson_strdup(uri_string);
-
    _mongoc_uri_build_write_concern (uri);
+
+   if (!_mongoc_write_concern_is_valid(uri->write_concern)) {
+      mongoc_uri_destroy(uri);
+      return NULL;
+   }
+
+   uri->str = bson_strdup(uri_string);
 
    return uri;
 }
--- a/src/mongoc/mongoc-write-concern.c
+++ b/src/mongoc/mongoc-write-concern.c
@@ -48,7 +48,7 @@ mongoc_write_concern_new (void)
    mongoc_write_concern_t *write_concern;
 
    write_concern = bson_malloc0(sizeof *write_concern);
-   write_concern->w = -2;
+   write_concern->w = MONGOC_WRITE_CONCERN_W_DEFAULT;
 
    return write_concern;
 }
@@ -214,7 +214,7 @@ bool
 mongoc_write_concern_get_wmajority (const mongoc_write_concern_t *write_concern)
 {
    bson_return_val_if_fail(write_concern, false);
-   return (write_concern->w == -3);
+   return (write_concern->w == MONGOC_WRITE_CONCERN_W_MAJORITY);
 }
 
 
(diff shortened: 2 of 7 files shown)


jmikola:
Regarding the magic number replacement, using `MONGOC_WRITE_CONCERN_W_DEFAULT()` in the `new()` function seemed consistent, and I just ended up hunting down the other cases.  I left the `(w >= -3)` range check in `mongoc_write_concern_set_w()` as-is since it wasn't referring to a specific value. If some/all of those replacements are unnecessary, just let me know and I'll amend or cut the commit entirely.

src/mongoc/mongoc-write-concern.c
```diff
@@ -380,7 +382,10 @@ bool
 _mongoc_write_concern_needs_gle (const mongoc_write_concern_t *write_concern)
 {
    if (write_concern) {
-      return ((write_concern->w != 0) && (write_concern->w != -1));
+      return (((write_concern->w != MONGOC_WRITE_CONCERN_W_UNACKNOWLEDGED) &&
+               (write_concern->w != MONGOC_WRITE_CONCERN_W_ERRORS_IGNORED)) ||
+              (write_concern->fsync_ != MONGOC_WRITE_CONCERN_FSYNC_DEFAULT && !!write_concern->fsync_) ||
+              (write_concern->journal != MONGOC_WRITE_CONCERN_JOURNAL_DEFAULT && !!write_concern->journal));
```
jmikola: These two lines duplicate logic within `mongoc_write_concern_get_fsync()` and `mongoc_write_concern_get_journal()`. Let me know if it makes sense to simply call the functions.

ajdavis: Let's refactor indeed.

jmikola:
@ajdavis: Updated and rebased.

src/mongoc/mongoc-write-concern.c
```diff
@@ -98,7 +100,7 @@ bool
 mongoc_write_concern_get_fsync (const mongoc_write_concern_t *write_concern)
 {
    bson_return_val_if_fail(write_concern, false);
-   return write_concern->fsync_;
+   return (write_concern->fsync_ != MONGOC_WRITE_CONCERN_FSYNC_DEFAULT && !!write_concern->fsync_);
```
hanumantmk: Are all of these checks basically:

(x != -1 && !!x)

?

If so, let's just move to == 1

src/mongoc/mongoc-write-concern.c
```diff
@@ -98,7 +100,7 @@ bool
 mongoc_write_concern_get_fsync (const mongoc_write_concern_t *write_concern)
 {
    bson_return_val_if_fail(write_concern, false);
-   return write_concern->fsync_;
+   return (write_concern->fsync_ == true);
```
jmikola: @hanumantmk: since the setter functions use bool arguments, are we guaranteed that only `true` or `false` can ultimately be assigned? I modified the tests to assign arbitrary integers and that does seem to be the case -- I just wasn't expecting the compiler to do that for us.

chergert: They should probably be like "uint8_t fsync_ : 1;" anyway (just make sure unsigned). They probably weren't originally just to make it easier to assign since you can't take and address of a bitfield.

hanumantmk: You can either do > 0 or !!fsync on the way in to make sure it isn't >1

src/mongoc/mongoc-uri.c
```diff
@@ -737,13 +738,13 @@ _mongoc_uri_build_write_concern (mongoc_uri_t *uri) /* IN */
...
-                                        MONGOC_WRITE_CONCERN_W_ERRORS_IGNORED);
-            break;
-         case 0:
-            mongoc_write_concern_set_w (write_concern,
-                                        MONGOC_WRITE_CONCERN_W_UNACKNOWLEDGED);
+         case MONGOC_WRITE_CONCERN_W_ERRORS_IGNORED:
+         case MONGOC_WRITE_CONCERN_W_UNACKNOWLEDGED:
+            /* Warn on conflict, since write concern will be validated later */
+            if (mongoc_write_concern_get_journal(write_concern)) {
+               MONGOC_WARNING("Journal conflicts with w value [w=%d].", value);
```
jmikola: When running tests, this emits:

```
WARNING:       mongoc: Journal conflicts with w value [w=0].
WARNING:       mongoc: Journal conflicts with w value [w=-1].
```

Since `mongoc_uri_new()` really doesn't give us an error reporting mechanism, I think warnings are helpful. Is there a convenient way to suppress the log output during the test, though? I saw some other tests using `mongoc_log_set_handler()`, but that was through a `main()` function that the URI test doesn't have, so I'm not sure of the best way to go about that.

bjori: fyi; I have not found a realistic solution on how to upgrade these warnings to be able to programmatically deal with them in phongo.

chergert: In glib we have g_test_expect_warning or something to th nature that asserts if we don't get that warning string at test end or next log message.

You could also add a new_with_error variant that gives a bson_error_t if null is returned.

hanumantmk: There's a little bit of a facility in test-libmongoc.c right now (suppress_one_message(), which suppresses N messages based on incrementing a global).

It's a little too broad right now because the suppressions include messages which aren't usually printed (INFO level).  And if we want to use it to verify that warnings are triggered we should probably do something fancier, but that's the facility you're basically looking for.

tests/test-mongoc-uri.c
```diff
@@ -438,9 +438,17 @@ test_mongoc_uri_write_concern (void)
...
+      { "mongodb://localhost/?journal=true", true, MONGOC_WRITE_CONCERN_W_DEFAULT },
+      { "mongodb://localhost/?w=0&journal=true", false, MONGOC_WRITE_CONCERN_W_UNACKNOWLEDGED },
+      { "mongodb://localhost/?w=-1&journal=true", false, MONGOC_WRITE_CONCERN_W_ERRORS_IGNORED },
+      { "mongodb://localhost/?w=1&journal=true", true, 1 },
       { NULL }
    };
 
+   /* Suppress warnings from two invalid URIs ("journal" and "w" conflict) */
+   suppress_one_message();
+   suppress_one_message();
```
jmikola: @hanumantmk: Two suppressions added. I think we're good to go.

ajdavis:
Thanks for the contribution!

PR #196, opened 2015-03-13
CDRIVER-567 deprecate get_max_* methods on client

samantharitter:


src/mongoc/mongoc-cluster.c
```diff
@@ -207,6 +207,32 @@ _mongoc_cluster_run_command (mongoc_cluster_t      *cluster,
 /*
  *--------------------------------------------------------------------------
  *
+ * _mongoc_cluster_is_single_threaded --
```
hanumantmk: Let's ditch this.  In single threaded you don't need the lock.  In multi-threaded it doesn't matter, because you set that member variable before thread starts, which synchronizes (and it never changes during the life of the object).

hanumantmk:
beyond the comment, seems like a good commit to me

samantharitter:
Patch updated, would you mind giving it another once-over?

src/mongoc/mongoc-cluster.c
```diff
@@ -1794,51 +1858,20 @@ _mongoc_cluster_min_of_max_msg_size_nodes (void *item,
...
-         mongoc_set_for_each (topology->description.servers,
-                              _mongoc_cluster_min_of_max_obj_size_sds,
-                              &max_bson_obj_size);
-      }
+   if (!cluster->client->topology->single_threaded) {
+      mongoc_set_for_each (cluster->nodes,
+                           _mongoc_cluster_min_of_max_obj_size_nodes,
+                           &max_bson_obj_size);
+   } else {
+      mongoc_set_for_each (cluster->client->topology->description.servers,
```
hanumantmk: I know we were doing it before, but I don't think this is thread safe.  Other uses of for_each are probably error prone as well

src/mongoc/mongoc-cluster.c
```diff
@@ -1866,51 +1896,18 @@ mongoc_cluster_get_max_bson_obj_size (mongoc_cluster_t *cluster,
...
-         mongoc_set_for_each (topology->description.servers,
-                              _mongoc_cluster_min_of_max_msg_size_sds,
-                              &max_msg_size);
-      }
+   if (!cluster->client->topology->single_threaded) {
+      mongoc_set_for_each (cluster->nodes,
+                           _mongoc_cluster_min_of_max_msg_size_nodes,
+                           &max_msg_size);
+   } else {
+      mongoc_set_for_each (cluster->client->topology->description.servers,
```
hanumantmk: same as above.  this probably isn't thread safe

samantharitter: actually, this only gets run in the single-threaded case, so I think it should be ok

samantharitter: sorry that was a confusingly-structured conditional

src/mongoc/mongoc-cluster.c
samantharitter: As below, I believe this is fine

PR #197, opened 2015-03-13
CDRIVER-565 make mongoc_cluster_fetch_stream thread-safe

samantharitter:


src/mongoc/mongoc-cluster.c
```diff
@@ -1214,19 +1214,31 @@ mongoc_cluster_fetch_stream (mongoc_cluster_t *cluster,
...
+   int64_t timestamp;
+   bool single_threaded;
 
    bson_return_val_if_fail(cluster, NULL);
 
-   scanner_node = mongoc_topology_scanner_get_node(cluster->client->topology->scanner, server_id);
+   topology = cluster->client->topology;
+   mongoc_mutex_lock (&topology->mutex);
+
+   scanner_node = mongoc_topology_scanner_get_node(topology->scanner, server_id);
```
hanumantmk: Let's migrate this logic into a get_timestamp method which handles the mutex logic for us.

hanumantmk: You'll want to strip out the other uses of mutex_lock for the topology mutex as well.

PR #204, opened 2015-03-17
working

hanumantmk:
Abstracts stream initiation and async command setup so you don't have to use streams from libmongoc.

### DIFF
--- a/src/mongoc/mongoc-topology-scanner.c
+++ b/src/mongoc/mongoc-topology-scanner.c
@@ -37,8 +37,9 @@ static void
 mongoc_topology_scanner_node_destroy (mongoc_topology_scanner_node_t *node);
 
 mongoc_topology_scanner_t *
-mongoc_topology_scanner_new (mongoc_topology_scanner_cb_t cb,
-                             void                        *cb_data)
+mongoc_topology_scanner_new (const mongoc_uri_t          *uri,
+                             mongoc_topology_scanner_cb_t cb,
+                             void                        *data)
 {
    mongoc_topology_scanner_t *ts = bson_malloc0 (sizeof (*ts));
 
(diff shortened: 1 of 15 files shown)


hanumantmk:
@jmikola @bjori 

This threads support for custom stream initiators through to the single threaded client's.

It also has an interface on the topology scanner to provide a setup function which you could use to manage tls setup if you so desire.

I think it's basically what you'll need for your integration

bjori:
Thanks!

This seems to be working fairly well, but https://jira.mongodb.org/browse/CDRIVER-582 is skewing all of my tests though so difficult to tell :)

src/mongoc/mongoc-topology-scanner.c
```diff
@@ -47,11 +48,31 @@ mongoc_topology_scanner_new (mongoc_topology_scanner_cb_t cb,
    BSON_APPEND_INT32 (&ts->ismaster_cmd, "isMaster", 1);
 
    ts->cb = cb;
-   ts->cb_data = cb_data;
+   ts->cb_data = data;
+   ts->uri = uri;
```
bjori: you need to mongoc_uri_copy() the uri, otherwise it'll be destroyed by mongoc_client_new() which calls mongoc_topology_new() then mongoc_topology_scanner_new()

bjori:
Is it intentional that I have to register the setup() callback to?
Its return value is a bit iffy. It also seems to be called multiple times per stream.

Also, I don't get any calls to the client->initiator() anymore, is that dead code now?

hanumantmk:
I pushed this into 1.2.0-dev

src/mongoc/mongoc-topology-scanner.c
hanumantmk: topology_new copies the uri, and cleans it up.  so the scanner can have a pointer and not worry about ownership

hanumantmk: ah, I see.  I'll patch 1.2.0-dev

PR #206, opened 2015-03-18
CDRIVER-582: Fix mongoc_server_description leaks

bjori:


bjori:
@ajdavis @hanumantmk could you have a look a this and merge it if all ok?

hanumantmk:
We don't actually need to copy then destroy, in single threaded mode we can just look into the description directly and skip the copy.  I'll fix that in a bit

hanumantmk:
Fixed in https://github.com/mongodb/mongo-c-driver/commit/94aa18d21ea03472609e0a0b22cd0f807e60b954

PR #207, opened 2015-03-20
Pass the proper server hint to the executor - and reassign it after exec...

bjori:
...uting

We need to re-assign it after executing if the server id wasn't
actually set


bjori:
Related to https://jira.mongodb.org/browse/CDRIVER-586

PR #208, opened 2015-03-23
CDRIVER-586 Direct bulk writes to hinted nodes

hanumantmk:
Bulk writes could have a hint set, which was used for decisions around
legacy op, size of write commands, etc.  After all those checks, we
dispatched to whichever node happened to be the primary though.

Added support for a command_simple_with_hint, which allows for a hint.
This lets us dispatch to a specific node, which will error on a non
master, rather than possibly silently succeeding if we route to a
primary.


bjori:
LGTM

hanumantmk:
Closed in https://github.com/mongodb/mongo-c-driver/commit/b887dab50bec002f67c9e9b0c877bfa36522baa0

PR #211, opened 2015-03-24
CDRIVER-590: Support readPreference connection string option

jmikola:
https://jira.mongodb.org/browse/CDRIVER-590

### DIFF
--- a/src/mongoc/mongoc-uri.c
+++ b/src/mongoc/mongoc-uri.c
@@ -42,7 +42,7 @@ struct _mongoc_uri_t
    char                   *database;
    bson_t                  options;
    bson_t                  credentials;
-   bson_t                  read_prefs;
+   mongoc_read_prefs_t    *read_prefs;
    mongoc_write_concern_t *write_concern;
 };
 
@@ -437,16 +437,13 @@ mongoc_uri_parse_auth_mechanism_properties (mongoc_uri_t *uri,
 
 static void
 mongoc_uri_parse_tags (mongoc_uri_t *uri, /* IN */
-                       const char   *str, /* IN */
-                       bson_t       *doc) /* IN */
+                       const char   *str) /* IN */
 {
    const char *end_keyval;
    const char *end_key;
    bson_t b;
    char *keyval;
    char *key;
-   char keystr[32];
-   int i;
 
    bson_init(&b);
 
@@ -466,9 +463,7 @@ again:
       }
    }
 
-   i = bson_count_keys(doc);
-   bson_snprintf(keystr, sizeof keystr, "%u", i);
-   bson_append_document(doc, keystr, -1, &b);
+   mongoc_read_prefs_add_tag(uri->read_prefs, &b);
    bson_destroy(&b);
 }
 
@@ -518,7 +513,7 @@ mongoc_uri_parse_option (mongoc_uri_t *uri,
                         (0 == strcasecmp (value, "t")) ||
                         (0 == strcmp (value, "1")));
    } else if (!strcasecmp(key, "readpreferencetags")) {
-      mongoc_uri_parse_tags(uri, value, &uri->read_prefs);
+      mongoc_uri_parse_tags(uri, value);
    } else if (!strcasecmp(key, "authmechanism") ||
               !strcasecmp(key, "authsource")) {
       bson_append_utf8(&uri->credentials, key, -1, value, -1);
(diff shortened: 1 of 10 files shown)


src/mongoc/mongoc-uri.c
```diff
@@ -704,6 +704,54 @@ mongoc_uri_get_mechanism_properties (const mongoc_uri_t *uri, bson_t *properties
...
+         MONGOC_WARNING("Unsupported readPreference value [readPreference=%s].", str);
+      }
+   }
+
+   if (!bson_empty(&uri->read_prefs_tags)) {
+      /* Warn on conflict, since read preference will be validated later */
+      if (mongoc_read_prefs_get_mode(read_prefs) == MONGOC_READ_PRIMARY) {
+         MONGOC_WARNING("Primary read preference mode conflicts with tags.");
+      }
+      mongoc_read_prefs_set_tags(read_prefs, &uri->read_prefs_tags);
```
jmikola: Since this function is always called between options parsing and the URI constructor returning, we could immediately destroy `&uri->read_prefs_tag` here and modify the should-be-deprecated `mongoc_uri_get_read_prefs()` function to fetch tags directly from `uri->read_prefs`.

src/mongoc/mongoc-uri.c
```diff
@@ -990,6 +1046,15 @@ mongoc_uri_unescape (const char *escaped_string)
 }
 
 
+const mongoc_read_prefs_t *
+mongoc_uri_get_read_prefs_t (const mongoc_uri_t *uri) /* IN */
```
jmikola: Since `mongoc_uri_get_read_prefs()` was unfortunately taken by the getter function for the tags `bson_t`, I added `_t` here on @hanumantmk's advice. Since this adds a public API function, should I include new documentation in this PR?

My second question: how should we go about deprecating (at least in documentation) the `mongoc_uri_get_read_prefs()` function? A cursory grep for "deprecated" in the project didn't turn up much.

tests/test-mongoc-uri.c
```diff
@@ -96,12 +98,15 @@ test_mongoc_uri_new (void)
    ASSERT_CMPSTR(bson_iter_utf8(&iter, NULL), "1");
    mongoc_uri_destroy(uri);
 
-   uri = mongoc_uri_new("mongodb://localhost:27017/?readPreferenceTags=dc:ny&readPreferenceTags=");
+   uri = mongoc_uri_new("mongodb://localhost:27017/?readPreference=secondaryPreferred&readPreferenceTags=dc:ny&readPreferenceTags=");
```
jmikola: Since we now validate the read preference, I needed to specify something other than `primary` (the default) here to ensure URI construction actually succeeds.

tests/test-mongoc-uri.c
```diff
@@ -411,6 +418,101 @@ test_mongoc_uri_unescape (void)
...
+      /* readPreference should take priority over slaveOk */
+      { "mongodb://localhost/?slaveOk=false&readPreference=secondary", true, MONGOC_READ_SECONDARY, NULL },
+      /* readPreferenceTags conflict with primary mode */
+      { "mongodb://localhost/?readPreferenceTags=", false },
+      { "mongodb://localhost/?readPreference=primary&readPreferenceTags=", false },
+      { "mongodb://localhost/?slaveOk=false&readPreferenceTags=", false },
+      { "mongodb://localhost/?readPreference=secondaryPreferred&readPreferenceTags=", true, MONGOC_READ_SECONDARY_PREFERRED, tags_empty },
+      { "mongodb://localhost/?readPreference=secondaryPreferred&readPreferenceTags=dc:ny", true, MONGOC_READ_SECONDARY_PREFERRED, tags_dcny },
+      { "mongodb://localhost/?readPreference=nearest&readPreferenceTags=dc:ny&readPreferenceTags=", true, MONGOC_READ_NEAREST, tags_dcny_empty },
+      { "mongodb://localhost/?readPreference=nearest&readPreferenceTags=dc:ny,use:ssd&readPreferenceTags=dc:sf&readPreferenceTags=", true, MONGOC_READ_NEAREST, tags_dcnyusessd_dcsf_empty },
```
jmikola: While testing this, I realized that `readPreferenceTags` aren't really validated, either. For instance, `readPreferenceTags=foo` doesn't raise any errors. I haven't looked into what BSON that generates, but that's probably worth fixing in a subsequent ticket (either against 1.1.x or 1.2.x).

src/mongoc/mongoc-uri.c
```diff
@@ -42,7 +42,8 @@ struct _mongoc_uri_t
    char                   *database;
    bson_t                  options;
    bson_t                  credentials;
-   bson_t                  read_prefs;
+   bson_t                  read_prefs_tags;
```
hanumantmk: Do we actually need that?  It seems like we can get the read_prefs_tags from the read_prefs via mongoc_read_prefs_get_tags()

src/mongoc/mongoc-uri.c
hanumantmk: Yeah, that makes sense to me.  I'd have that just pull from the read prefs we're making here.

src/mongoc/mongoc-uri.c
hanumantmk: Since it's an API addition, see: https://github.com/mongodb/mongo-c-driver/blob/master/CONTRIBUTING.md for where to add the symbol for export/documentation.

As for deprecation, see examples like:

```c
void mongoc_bulk_operation_delete_one  (mongoc_bulk_operation_t       *bulk,
                                        const bson_t                  *selector)
   BSON_GNUC_DEPRECATED_FOR (mongoc_bulk_operation_remove_one);
```

That will trigger a deprecation warning, mentioning mongoc_bulk_operation_remove_one, every time a consumer uses mongoc_bulk_operation_delete_one

tests/test-mongoc-uri.c
hanumantmk: Hmm, that seems like a bit of a regression.  Not sure how I feel about it

jmikola: Previously, such a URI would result in the client's read preference having tags with the primary mode, which is invalid according to the read preference spec and `mongoc_read_prefs_is_valid()`. That said, `mongoc_read_prefs_is_valid()` was never actually called anywhere, so I'm not sure how libmongoc handled such a structure.

src/mongoc/mongoc-uri.c
jmikola: We can certainly fetch the final tag set from the `mongoc_read_prefs_t` struct, but `mongoc_uri_parse_tags()` is invoked each time `readPreferenceTags` is encountered during query string parsing and needs a `bson_t` to build upon.

If we initialize the `mongoc_read_prefs_t` struct earlier (i.e. during URI initialization), instead of in `_mongoc_uri_build_read_prefs()`, we can have `mongoc_uri_parse_tags()` call `mongoc_read_prefs_add_tag()` each time.

The actual mode should still be set just once in `_mongoc_uri_build_read_prefs()`, after all query string parameters are parsed, so that we can enforce `readPreference` taking priority over `slaveOk`.

src/mongoc/mongoc-uri.c
jmikola: Noted. I'll add this deprecation and the update the documentation in this PR when I get a chance.

jmikola: New symbol docs and deprecation of the old function are now done.

tests/test-mongoc-uri.c
jmikola: @hanumantmk, @ajdavis: Any follow-up here?

tests/test-mongoc-uri.c
jmikola: Opened [CDRIVER-598](https://jira.mongodb.org/browse/CDRIVER-598) to track this outside of this ticket/PR.

tests/test-mongoc-uri.c
ajdavis: @hanumantmk and I approve this change, we'll note it prominently in the changelog.

PR #213, opened 2015-03-27
CDRIVER-594 fix all the SS/SDAM timers

hanumantmk:
There's a bunch in here.

@ajdavis, mind taking a look to see if the logic is correct?

### DIFF
--- a/src/mongoc/mongoc-thread-private.h
+++ b/src/mongoc/mongoc-thread-private.h
@@ -41,11 +41,14 @@ mongoc_cond_timedwait (pthread_cond_t  *cond,
 {
    struct timespec to;
    struct timeval tv;
+   int64_t msec;
 
    bson_gettimeofday (&tv);
 
-   to.tv_sec = tv.tv_sec + timeout_msec / 1000;
-   to.tv_nsec = (timeout_msec % 1000) * 1000000;
+   msec = (tv.tv_sec * 1000) + (tv.tv_usec / 1000) + timeout_msec;
+
+   to.tv_sec = msec / 1000;
+   to.tv_nsec = (msec % 1000) * 1000 * 1000;
 
    return pthread_cond_timedwait (cond, mutex, &to);
 }
--- a/src/mongoc/mongoc-topology-description.c
+++ b/src/mongoc/mongoc-topology-description.c
@@ -445,7 +445,13 @@ mongoc_topology_description_select (mongoc_topology_description_t *topology,
    }
 
    if (topology->type == MONGOC_TOPOLOGY_SINGLE) {
-      return topology->servers->items[0].item;
+      sd = topology->servers->items[0].item;
+
+      if (sd->has_is_master) {
+         RETURN(sd);
+      } else {
+         RETURN(NULL);
+      }
    }
 
    _mongoc_array_init(&suitable_servers, sizeof(mongoc_server_description_t *));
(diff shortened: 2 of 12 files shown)


src/mongoc/mongoc-topology.c
```diff
@@ -399,7 +433,7 @@ mongoc_topology_server_by_id (mongoc_topology_t *topology, uint32_t id)
  */
 
 void
-_mongoc_topology_request_scan (mongoc_topology_t *topology)
+mongoc_topology_request_scan (mongoc_topology_t *topology)
```
hanumantmk: in retrospect, we should be able to call this from topology_select, but it's just not right (it shouldn't start the scan and shouldn't possibly block).

src/mongoc/mongoc-topology.c
```diff
@@ -321,8 +349,10 @@ mongoc_topology_select (mongoc_topology_t         *topology,
                                                            error);
 
       if (! selected_server) {
-         /* TODO request an immediate topology check here */
-         r = mongoc_cond_timedwait (&topology->cond_client, &topology->mutex, expire_at - now);
+         topology->scan_requested = true;
```
hanumantmk: we shouldn't have to do this.   I just got hung up on the fact that we have the mutex here.  It's fine to either have a version of request_scan which assumes we have the lock, or to drop the lock and use request scan, then resume it

src/mongoc/mongoc-thread-private.h
```diff
@@ -41,11 +41,14 @@ mongoc_cond_timedwait (pthread_cond_t  *cond,
 {
    struct timespec to;
    struct timeval tv;
+   int64_t msec;
```
hanumantmk: I really messed this up the first time 'round -_-;

src/mongoc/mongoc-topology-description.c
```diff
@@ -445,7 +445,13 @@ mongoc_topology_description_select (mongoc_topology_description_t *topology,
    }
 
    if (topology->type == MONGOC_TOPOLOGY_SINGLE) {
-      return topology->servers->items[0].item;
+      sd = topology->servers->items[0].item;
```
hanumantmk: I think it's better for server selection to say we don't have a server if we've never managed to ismaster it at all

src/mongoc/mongoc-topology-private.h
```diff
@@ -59,7 +61,8 @@ typedef struct _mongoc_topology_t
 } mongoc_topology_t;
 
 mongoc_topology_t *
-mongoc_topology_new (const mongoc_uri_t *uri);
+mongoc_topology_new (const mongoc_uri_t *uri,
+                     bool                single_threaded);
```
hanumantmk: I think it makes more sense for the topology to be created in single/multi threaded mode and for it to own it's background thread

src/mongoc/mongoc-topology-scanner-private.h
```diff
@@ -44,6 +44,7 @@ typedef struct mongoc_topology_scanner_node
    mongoc_async_cmd_t             *cmd;
    mongoc_stream_t                *stream;
    int64_t                         timestamp;
+   int64_t                         last_used;
```
hanumantmk: This is messy, but I'm not quite sure where else to hang it.  Maybe on the streams themselves?  Bleh

src/mongoc/mongoc-topology.c
```diff
@@ -48,21 +54,17 @@ _mongoc_topology_scanner_cb (uint32_t      id,
 
    topology = data;
 
-   /* if multithreaded, driver may write through invalidate_server, so lock */
```
hanumantmk: This is probably wrong to remove actually

ajdavis:
LGTM.

PR #215, opened 2015-04-07
CDRIVER-609: Differenciate between a failed stream, and cleanup/destroy

bjori:
This adds a new `failed` handler to the mongoc_stream_t, along with
mongoc_stream_failed().

The failed handler shall be called when the stream truely needs to be
destroyed (due to errors, removal from topology, etc etc).
The destroy handler shall be called during normal _destroy() routines,
like mongoc_destroy().

By default, for BC, if no failed handler is provided we will fallback
and call the `destroy` handler.

### DIFF
--- a/build/autotools/versions.ldscript
+++ b/build/autotools/versions.ldscript
@@ -250,5 +250,6 @@ LIBMONGOC_1.2 {
         mongoc_server_description_host;
         mongoc_server_description_id;
         mongoc_server_description_new_copy;
+        mongoc_stream_failed;
         mongoc_uri_get_read_prefs_t;
 } LIBMONGOC_1.1;
--- a/src/mongoc/mongoc-stream.c
+++ b/src/mongoc/mongoc-stream.c
@@ -61,6 +61,32 @@ mongoc_stream_close (mongoc_stream_t *stream)
 
 
 /**
+ * mongoc_stream_failed:
+ * @stream: A mongoc_stream_t.
+ *
+ * Frees any resources referenced by @stream, including the memory allocation
+ * for @stream.
+ * This handler is called upon stream failure, such as network errors, invalid replies
+ * or replicaset reconfigures deleteing the stream
+ */
+void
+mongoc_stream_failed (mongoc_stream_t *stream)
+{
+   ENTRY;
+
+   bson_return_if_fail(stream);
+
+   if (stream->failed) {
+	   stream->failed(stream);
+   } else {
+	   stream->destroy(stream);
+   }
+
+   EXIT;
+}
+
+
+/**
  * mongoc_stream_destroy:
  * @stream: A mongoc_stream_t.
  *
--- a/src/mongoc/mongoc-stream.h
+++ b/src/mongoc/mongoc-stream.h
@@ -61,7 +61,7 @@ struct _mongoc_stream_t
    ssize_t          (*poll)            (mongoc_stream_poll_t *streams,
                                         size_t                nstreams,
                                         int32_t               timeout);
-   void             *padding [7];
+   void             *padding [6];
 };
 
 
--- a/src/mongoc/mongoc-stream.h
+++ b/src/mongoc/mongoc-stream.h
@@ -61,13 +61,15 @@ struct _mongoc_stream_t
    ssize_t          (*poll)            (mongoc_stream_poll_t *streams,
                                         size_t                nstreams,
                                         int32_t               timeout);
-   void             *padding [6];
+   void             (*failed)          (mongoc_stream_t *stream);
+   void             *padding [5];
 };
 
 
 mongoc_stream_t *mongoc_stream_get_base_stream (mongoc_stream_t       *stream);
 int              mongoc_stream_close           (mongoc_stream_t       *stream);
 void             mongoc_stream_destroy         (mongoc_stream_t       *stream);
+void             mongoc_stream_failed          (mongoc_stream_t       *stream);
 int              mongoc_stream_flush           (mongoc_stream_t       *stream);
 ssize_t          mongoc_stream_writev          (mongoc_stream_t       *stream,
                                                 mongoc_iovec_t        *iov,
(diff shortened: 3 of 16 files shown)


build/autotools/versions.ldscript
```diff
@@ -251,4 +251,5 @@ LIBMONGOC_1.2 {
         mongoc_server_description_id;
         mongoc_server_description_new_copy;
         mongoc_uri_get_read_prefs_t;
+        mongoc_stream_failed;
```
hanumantmk: sort this in with the rest

src/mongoc/mongoc-stream.h
```diff
@@ -61,6 +61,7 @@ struct _mongoc_stream_t
    ssize_t          (*poll)            (mongoc_stream_poll_t *streams,
                                         size_t                nstreams,
                                         int32_t               timeout);
+   void             (*failed)          (mongoc_stream_t *stream);
    void             *padding [7];
```
hanumantmk: remove one padding for every pointer you add to the struct

src/mongoc/mongoc-stream.h
```diff
@@ -61,6 +61,7 @@ struct _mongoc_stream_t
    ssize_t          (*poll)            (mongoc_stream_poll_t *streams,
                                         size_t                nstreams,
                                         int32_t               timeout);
+   void             (*failed)          (mongoc_stream_t *stream);
```
hanumantmk: remove one padding for each pointer you add to the struct

src/mongoc/mongoc-topology-scanner.c
```diff
@@ -167,7 +171,7 @@ mongoc_topology_scanner_rm (mongoc_topology_scanner_t *ts,
 
    ele = mongoc_topology_scanner_get_node (ts, id);
    if (ele) {
-      mongoc_topology_scanner_node_destroy (ele);
+      mongoc_topology_scanner_node_destroy (ele, true);
```
hanumantmk: It looks like scanner_rm never gets called anywhere.  I'd just get rid of the whole function

build/autotools/versions.ldscript
bjori: My bad, I thought the list was order of introduction :)

src/mongoc/mongoc-stream.h
bjori: The padding seems off by one already, it was not decremented with the addition of poll.
Should I decrement it for poll in a separate commit?

src/mongoc/mongoc-topology-scanner.c
bjori: Nice catch!

src/mongoc/mongoc-stream.c
```diff
@@ -61,6 +61,32 @@ mongoc_stream_close (mongoc_stream_t *stream)
...
+ * or replicaset reconfigures deleteing the stream
+ */
+void
+mongoc_stream_failed (mongoc_stream_t *stream)
+{
+   ENTRY;
+
+   bson_return_if_fail(stream);
+
+   if (stream->failed) {
```
hanumantmk: I'm not sure if this does what you want.  The top level object is probably a buffer.  That won't support fail, so we'll go to mongoc_stream_buffered_destroy().  That calls mongoc_stream_destroy(), which won't call to fail.

hanumantmk:
I think you want something simpler.  Maybe keep failed a (void *), but make it null by default.  Offer a mongoc_stream_fail() method which tags the stream as failed.  Then you can look for that in your destroy and do something intelligent with it there without re-working everything

src/mongoc/mongoc-stream.h
hanumantmk: crap.  Yeah.  We want to make sure that sizeof(mongoc_stream_t) is consistent between 1.1 and 1.2

src/mongoc/mongoc-stream.c
bjori: Why would it be a buffer?
The top level stream is my provided stream abstraction.

It may be a buffer for normal mongoc behavior, in which case its fine to destroy() it in stream_buffered_destroy() as mongoc doesn't care for the difference

hanumantmk: Ah, you're right.  For some reason I thought the buffer was above the stream_initiator

I guess I'm a little concerned about messing with the abstraction.  It feels a bit fragile to have failed dispatch to destroy, at least for me.

bjori: I'm fine with adding the skeleton for the failed() handler for the buffer and tls streams, that in the end just do the same as destroy() - it does indeed give better message.
This was just the quickest way to goal :)

bjori:
I'm not sure how "tagging the stream as failed" and then calling destroy would be any different -- it would just be extra integer on the stream plus a mongoc_stream_*() function that wasn't abstraction function ?

We'd still need this rework to separate the actual reason for destroying the stream.
The new `failed` handler doesn't change anything for those who don't need to know the difference, and only adds 5lines of code for those who do need to know :]

src/mongoc/mongoc-stream.c
bjori: Note that the if (stream->failed) still needs to be here for BC for other external implementations

build/autotools/versions.ldscript
bjori: done

src/mongoc/mongoc-stream.h
bjori: done

src/mongoc/mongoc-topology-scanner.c
bjori: done

src/mongoc/mongoc-stream.c
bjori: done

hanumantmk:
I was thinking that you'd just look into the stream_t in your regular destroy, but you've won me over, special failed method pointer it is

hanumantmk:
LGTM

PR #216, opened 2015-04-08
CDRIVER-611: Consistent treatment of integer URI options

bjori:
We now consitently allow =0 to mean "whatever the default value is"

### DIFF
--- a/src/mongoc/mongoc-cluster.c
+++ b/src/mongoc/mongoc-cluster.c
@@ -1338,7 +1338,6 @@ mongoc_cluster_init (mongoc_cluster_t   *cluster,
                      const mongoc_uri_t *uri,
                      void               *client)
 {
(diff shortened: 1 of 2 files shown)


bjori:
@ajdavis You OK with this? :)

ajdavis:
It's weird, but if there's a precedent then I suppose it should be consistent. Could you also look over the docs and make sure it's super-clear to people that the C Driver considers 0 equal to 300,000, at least as far as sockettimeoutms is concerned? Otherwise people might think that normal arithmetic works in our universe....

bjori:
Merged in https://github.com/mongodb/mongo-c-driver/commit/94c4091ab040d21af9c92e224b81cd87fbb70f54

Also documented the new *MS values since I was there.

ajdavis:
Beautiful, thank you!

PR #217, opened 2015-04-08
min_bytes is unsigned -- this will never be false

bjori:
-Wtype-limits


chergert:
I think this parameter used to be ssize_t with -1 to "opportunistically fill". LGTM.

PR #221, opened 2015-04-15
Move mongoc_cursor_cursorid_t typedef into private header

jmikola:
The PHP driver needs access to this structure, since it reimplements some of the logic within `_mongoc_cursor_cursorid_prime()`.


jmikola:
Thanks!

PR #222, opened 2015-04-17
add helper function to get gridfs files_id from mongoc_gridfs_file_t.

lloydzhou:
i have warp one library for luajit to using mongo-c-driver [luajit-mongo](https://github.com/lloydzhou/luajit-mongo).
i try to get the files_id, from struct "mongoc_gridfs_file_t". but, it always return "00...00".
so i want to add this helper function, to get the file's object_id when save file.


ajdavis:
Hi Lloyd, thanks for the contribution! I have three requests:

1. Could you rename it to mongoc_gridfs_file_get_id, with no "files"? It's shorter and equally descriptive.
2. The function should return a bson_value_t, not a bson_oid_t. Some drivers can store GridFS files with _ids of any type, not just ObjectId, so the C Driver must be able to read such files.
3. Please copy the value out instead of returning a pointer to it. It's much safer that way.

So I think the signature should be:
```
bool mongoc_gridfs_file_get_id (bson_value_t *file_id);
```
It will return false if it fails for any reason, otherwise it uses `bson_value_copy` to copy the id to the provided location.

ajdavis:
By the way I don't understand what you mean when you write,

> I try to get the files_id, from struct "mongoc_gridfs_file_t". but, it always return "00...00".

Can you show me an example of what you're doing?

lloydzhou:
@ajdavis 
1. thanks for you suggestion.
2. i have write the test file [here](https://github.com/lloydzhou/luajit-mongo/blob/test-gridfs-file-t/test.lua#L15):

    local file = fs.create('testname')
    print (tostring(file))
    file.write("sdfsdfgf")
    file.save()
    print (file.get_files_id())
    print (file.files_id.value.v_oid)

the results:

    ➜  luajit-mongo git:(test-gridfs-file-t) ✗ luajit test.lua
    2015/04/18 15:32:25.0508: [ 8856]:    DEBUG:      cluster: Client initialized in direct mode.
    "cdata<struct 513 *>: 0x009a3bd0"
    553208897221f12298327f83
    000000000000000000000000

when i using the helper function "mongoc_gridfs_file_get_files_id()" can get the right objectid, but when i try to get the id from the struct, get "000000000000000000000000"...

lloydzhou:
@ajdavis 
i have update the "mongo-c-driver", and the [test file](https://github.com/lloydzhou/luajit-mongo/blob/f60e190e7d1bf9ae86d70b93d075f4be5937364a/test.lua#L16) in "luajit-mongo", can work  now!

doc/mongoc_gridfs_file_get_files_id.page
```diff
@@ -0,0 +1,37 @@
...
+<page xmlns="http://projectmallard.org/1.0/"
+      type="topic"
+      style="function"
+      xmlns:api="http://projectmallard.org/experimental/api/"
+      xmlns:ui="http://projectmallard.org/experimental/ui/"
+      id="mongoc_gridfs_file_get_files_id">
+  <info>
+    <link type="guide" xref="mongoc_gridfs_file_t" group="function"/>
+  </info>
+  <title>mongoc_gridfs_file_get_files_id()</title>
```
ajdavis: Remember to change the title.

ajdavis: And the filename!

src/mongoc/mongoc-gridfs-file.c
```diff
@@ -769,6 +769,16 @@ mongoc_gridfs_file_error (mongoc_gridfs_file_t *file,
    RETURN(false);
 }
 
+bool
+mongoc_gridfs_file_get_id (mongoc_gridfs_file_t *file, bson_value_t *file_id)
+{
+   bson_return_val_if_fail (file, -1);
```
ajdavis: "bson_return_val_if_fail (file, false);"

ajdavis:
Looking good, very close.

lloydzhou:
Thank's for review my code!

ajdavis:
Thanks for this; I'll merge it into the 1.2.0-dev branch.

ajdavis:
Merged in https://github.com/mongodb/mongo-c-driver/commit/6789a7a678046a0e123a09c21151aee12030c508

PR #223, opened 2015-04-21
CDRIVER-619 fail fast after TLS socket hangup

ajdavis:


ajdavis:
I'll later refactor test-mongoc-stream-tls-hangup with ssl-test of course.

PR #225, opened 2015-04-28
NULL checked log function ptr

stoned7:


ajdavis:
Merged in https://github.com/mongodb/mongo-c-driver/commit/6ae03f5d78d99fd174d50faef75cf98510c05af1.

PR #227, opened 2015-05-04
Another shot at batched updates/deletes

hanumantmk:
More work on top of what Machyne and Kai Mast put together.

This needs a thorough review, but it passes tests for 2.4 and 2.6 without any fixmes...


ajdavis:
Thanks @hanumantmk, Kai, and @Machyne!

PR #228, opened 2015-05-05
Avoid use of SHA1 for scram for FIPS

hanumantmk:
instead use EVP_sha1 so our scram auth is compatible with openssl in FIPS
mode.


ajdavis:
Thanks!

PR #229, opened 2015-05-08
fix hang in TLS handshake

ajdavis:
CDRIVER-630 default_stream_initiator can block forever in TLS handshake


src/mongoc/mongoc-client.c
```diff
@@ -336,14 +337,18 @@ mongoc_client_default_stream_initiator (const mongoc_uri_t       *uri,
             return NULL;
          }
 
-         if (!mongoc_stream_tls_do_handshake (base_stream, -1) ||
+         if (bson_iter_init_find_case (&iter, options, "connecttimeoutms") &&
+             BSON_ITER_HOLDS_INT32 (&iter)) {
+            connecttimeoutms = bson_iter_int32 (&iter);
```
bjori: I think in other places optionname=0 means "use default".
-1 means "no limit"

ajdavis: Great catch; fixed.

PR #239, opened 2015-06-09
CDRIVER-678: Only the first member of seedlist is used

bjori:
Previously we would only add the node (topology description) to the
topology scanner when extracting it from a seedlist.
This resulted in no new nodes were ever added/connected to -- we would
only ever connect to the nodes in the seedlist -- no actual topology
discovery was being done.

Now we make sure we add the node to the scanner when we get a new topology
description, and start monitoring it.
When we have new ismaster results, we add new nodes, and remove old
nodes to/from the scanner.

Also removed weird mongoc_topology_cb_t -- which maybe was originally thought to handle this, but ultimately is not capable of doing so (and wasn't used at all).

### DIFF
--- a/src/mongoc/mongoc-cluster.c
+++ b/src/mongoc/mongoc-cluster.c
@@ -2071,7 +2071,7 @@ _mongoc_cluster_check_interval (mongoc_cluster_t *cluster,
          }
 
          mongoc_topology_description_handle_ismaster (
-            &topology->description, sd, &reply,
+            &topology->description, topology->scanner, sd, &reply,
             (now - before_ismaster) / 1000,    /* RTT_MS */
             error);
 
--- a/src/mongoc/mongoc-topology-description-private.h
+++ b/src/mongoc/mongoc-topology-description-private.h
@@ -33,11 +33,6 @@ typedef enum
       MONGOC_TOPOLOGY_DESCRIPTION_TYPES
    } mongoc_topology_description_type_t;
 
-typedef struct {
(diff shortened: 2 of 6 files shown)


bjori:
https://github.com/bjori/mongo-c-driver/commit/796506496ab15b9b8f4d6c9bde23b9d94f9b5871 is the money shot

ajdavis:
I'm sorry I didn't think of this earlier, but here's what I think you should do instead of this:

* Removing `mongoc_topology_cb_t` is great, keep that change.
* Don't pass the scanner around within mongoc-topology-description.c
* Instead, in `_mongoc_topology_scanner_cb`, right before `mongoc_cond_broadcast`, reconcile the list of servers in `topology->description` with the list of nodes in `topology->scanner`:
  1.  Write a function like `mongoc_host_list_t *mongoc_topology_description_get_host_list (mongoc_topology_description_t *)`
  2. Then write `mongoc_topology_scanner_reconcile (mongoc_host_list_t *)`, which adds and destroys scanner nodes so they match the list of server descriptions

The "reconcile" function is much like PyMongo's `Topology._update_servers`:

https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/topology.py#L322

... which is called after each ismaster response is processed, before giving up the lock:

https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/topology.py#L146

bjori:
Replaced with internal codereview

PR #243, opened 2015-06-10
CDRIVER-645: Setting "URI arguments" explicitly

bjori:


src/mongoc/mongoc-client.c
```diff
@@ -92,13 +90,8 @@ mongoc_client_connect_tcp (const mongoc_uri_t       *uri,
...
 
-   if ((options = mongoc_uri_get_options (uri)) &&
-       bson_iter_init_find_case (&iter, options, "connecttimeoutms") &&
-       BSON_ITER_HOLDS_INT32 (&iter)) {
-      if (!(connecttimeoutms = bson_iter_int32(&iter))) {
-         connecttimeoutms = MONGOC_DEFAULT_CONNECTTIMEOUTMS;
-      }
-   }
+   connecttimeoutms = mongoc_uri_get_option_as_int32(uri, "connecttimeoutms",
+         MONGOC_DEFAULT_CONNECTTIMEOUTMS);
```
ajdavis: Let's keep to a style more like this, where args are left-aligned and lines are less than 80 chars:
```
   connecttimeoutms = mongoc_uri_get_option_as_int32(
      uri, "connecttimeoutms",
      MONGOC_DEFAULT_CONNECTTIMEOUTMS);
```

src/mongoc/mongoc-client.c
```diff
@@ -337,12 +330,8 @@ mongoc_client_default_stream_initiator (const mongoc_uri_t       *uri,
...
          }
 
-         if (bson_iter_init_find_case (&iter, options, "connecttimeoutms") &&
-             BSON_ITER_HOLDS_INT32 (&iter)) {
-            if (!(connecttimeoutms = bson_iter_int32(&iter))) {
-               connecttimeoutms = MONGOC_DEFAULT_CONNECTTIMEOUTMS;
-            }
-         }
+         connecttimeoutms = mongoc_uri_get_option_as_int32(uri, "connecttimeoutms",
+               MONGOC_DEFAULT_CONNECTTIMEOUTMS);
```
ajdavis: style

src/mongoc/mongoc-cluster.c
```diff
@@ -1355,19 +1352,11 @@ mongoc_cluster_init (mongoc_cluster_t   *cluster,
...
    cluster->sec_latency_ms = 15;
 
-   cluster->sockettimeoutms = MONGOC_DEFAULT_SOCKETTIMEOUTMS;
-   if (bson_iter_init_find_case(&iter, options, "sockettimeoutms")) {
-      if (!(cluster->sockettimeoutms = bson_iter_int32 (&iter))) {
-         cluster->sockettimeoutms = MONGOC_DEFAULT_SOCKETTIMEOUTMS;
-      }
-   }
+   cluster->sockettimeoutms = mongoc_uri_get_option_as_int32(uri, "sockettimeoutms",
+         MONGOC_DEFAULT_SOCKETTIMEOUTMS);
```
ajdavis: same

src/mongoc/mongoc-cluster.c
```diff
@@ -1355,19 +1352,11 @@ mongoc_cluster_init (mongoc_cluster_t   *cluster,
...
 
-   cluster->socketcheckintervalms = MONGOC_TOPOLOGY_SOCKET_CHECK_INTERVAL_MS;
-   if (bson_iter_init_find_case(&iter, options, "socketcheckintervalms")) {
-      if (!(cluster->socketcheckintervalms = bson_iter_int32 (&iter))) {
-         cluster->socketcheckintervalms = MONGOC_TOPOLOGY_SOCKET_CHECK_INTERVAL_MS;
-	  }
-   }
+   cluster->socketcheckintervalms = mongoc_uri_get_option_as_int32(uri,
+         "socketcheckintervalms", MONGOC_TOPOLOGY_SOCKET_CHECK_INTERVAL_MS);
```
ajdavis: same

src/mongoc/mongoc-uri.c
```diff
@@ -468,6 +468,41 @@ mongoc_uri_parse_tags (mongoc_uri_t *uri, /* IN */
...
+
+bool
+mongoc_uri_option_is_utf8 (const char *key)
+{
+   return !strcasecmp(key, "authmechanism") ||
+              !strcasecmp(key, "authsource") ||
+              (!mongoc_uri_option_is_bool(key) && !mongoc_uri_option_is_int32(key)
+               && !!strcasecmp(key, "readpreferencetags") &&
+               !!strcasecmp(key, "authmechanismproperties"));
+}
```
ajdavis: Can you find a way to make this more readable? I suggest a couple of separate return statements, maybe temp variables. No need to squish all the logic into one expression.

src/mongoc/mongoc-uri.c
```diff
@@ -762,10 +782,7 @@ _mongoc_uri_build_write_concern (mongoc_uri_t *uri) /* IN */
                                   bson_iter_bool (&iter) ? 1 : MONGOC_WRITE_CONCERN_W_UNACKNOWLEDGED);
    }
 
-   if (bson_iter_init_find_case (&iter, &uri->options, "wtimeoutms") &&
-       BSON_ITER_HOLDS_INT32 (&iter)) {
-      wtimeoutms = bson_iter_int32 (&iter);
-   }
+   wtimeoutms = mongoc_uri_get_option_as_int32(uri, "wtimeoutms", 0);
 
    if (bson_iter_init_find_case (&iter, &uri->options, "journal") &&
```
ajdavis: write a mongoc_uri_get_option_as_bool?

src/mongoc/mongoc-uri.c
```diff
@@ -884,6 +910,15 @@ mongoc_uri_get_password (const mongoc_uri_t *uri)
    return uri->password;
 }
 
+bool
+mongoc_uri_set_password (mongoc_uri_t *uri, const char *password)
+{
+   bson_return_val_if_fail(password, false);
+   uri->password = bson_strdup(password);
```
ajdavis: Free the previous password.

src/mongoc/mongoc-uri.c
```diff
@@ -876,6 +893,15 @@ mongoc_uri_get_username (const mongoc_uri_t *uri)
    return uri->username;
 }
 
+bool
+mongoc_uri_set_username (mongoc_uri_t *uri, const char *username)
+{
+   bson_return_val_if_fail(username, false);
+   uri->username = bson_strdup(username);
```
ajdavis: Free the previous username.

src/mongoc/mongoc-uri.c
```diff
@@ -892,6 +927,15 @@ mongoc_uri_get_database (const mongoc_uri_t *uri)
    return uri->database;
 }
 
+bool
+mongoc_uri_set_database (mongoc_uri_t *uri, const char *database)
+{
+   bson_return_val_if_fail(database, false);
+   uri->database = bson_strdup(database);
```
ajdavis: Free the previous database.

src/mongoc/mongoc-uri.c
```diff
@@ -1073,3 +1117,102 @@ mongoc_uri_get_ssl (const mongoc_uri_t *uri) /* IN */
...
+      return false;
+   }
+
+   if (!mongoc_uri_option_is_utf8 (option)) {
+      return false;
+   }
+
+   if ((options = mongoc_uri_get_options (uri)) &&
+       bson_iter_init_find_case (&iter, options, option)) {
+      return false;
```
ajdavis: Let's make this act the same as mongoc_uri_set_option_as_bool, etc. -- replace the previous value -- or at least document the difference.

ajdavis:
This is a definite improvement.

It needs tests before I can merge it.

I think there's a bug in _mongoc_uri_build_write_concern, it throws away wtimeoutms unless w="majority". Check if there's a spec, otherwise just do the obvious thing: send wtimeoutms *unless* w=0 or w=-1. (w=-1 is buggy and deprecated, but we should avoid adding *more* bugs to it.)

I'd like to add these functions to the public API and document them. Would you like to do that in this PR or a later one?

src/mongoc/mongoc-uri.c
ajdavis: I've thought about this again, I think it *must* replace the previous value. That will be a little complex so make sure it's well-tested.

bjori: It'll have to duplicate the bson_t and overwrite it as we cannot do inline modifications of different size.
Not sure if that is such a good idea?

src/mongoc/mongoc-uri.c
bjori: Nice catch

src/mongoc/mongoc-client.c
bjori: Are you sure about that? That isn't the style that has been used so far.
It currently is "include as many arguments as fit in 80chars, then align all subsequent arguments at ( on next line, up 80, and repeat"

bjori:
If we put it into the public API we'd need a way to deal with complex options, such as SSL and authMechanismProperties -- or at the very least have specific functions for those (like we do for read_pref, write_concern, ....).

That something you want included as well?

src/mongoc/mongoc-uri.c
ajdavis: Yeah, let's do it. Test it well. It's not a performance issue since this is in configuration code. I suggest iterate the source bson key-by-key, copying values to the destination bson, until you reach "option", then write "value" instead. bson_iter_value and bson_append_value will allow you to ignore the types until you reach a key with the name "option". If you don't find that key, put option and value at the end.

My argument is that we shouldn't allow users to overwrite ints and bools, but prohibit them from overwriting strings, just because one is easy and the other is hard. That's an implementation detail.

src/mongoc/mongoc-client.c
ajdavis: It's the style I prefer. =)

ajdavis:
OK, let's leave this as private APIs for the C Driver and PHP driver for now.

bjori:
Replaced with "Rietveld Code Review"

PR #244, opened 2015-06-11
WIP: Bubbling up errors on SDAM/SS failure

bjori:


src/mongoc/mongoc-topology.c
```diff
@@ -351,29 +359,23 @@ mongoc_topology_select (mongoc_topology_t         *topology,
...
+         RETURN (mongoc_server_description_new_copy(selected_server));
+      }
 
-         if (selected_server) {
-            return mongoc_server_description_new_copy(selected_server);
-         }
+      bson_set_error(error,
+                     MONGOC_ERROR_SERVER_SELECTION,
+                     MONGOC_ERROR_SERVER_SELECTION_NO_MATCH,
+                     "Could not find any server matching");
```
ajdavis: How about "Could not find any server for %s", and substitute "writes" or a representation of read_pref?

src/mongoc/mongoc-topology.c
```diff
@@ -351,29 +359,23 @@ mongoc_topology_select (mongoc_topology_t         *topology,
          topology->stale = false;
       }
 
-      /* until we find a server or timeout */
-      for (;;) {
```
ajdavis: Nice fix. mongoc_topology_description_select already handles looping until the timeout, right?

src/mongoc/mongoc-topology-scanner.c
```diff
@@ -398,13 +399,18 @@ mongoc_topology_scanner_start (mongoc_topology_scanner_t *ts,
 
    DL_FOREACH_SAFE (ts->nodes, node, tmp)
    {
-      if (mongoc_topology_scanner_node_setup (node)) {
+      /* FIXME: Gotta catch all the pokemons, not overwrite them */
+      bson_error_t error;
```
ajdavis: Consider some logic something like this:

https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/topology.py#L369

The tests are here:

https://github.com/mongodb/mongo-python-driver/blob/master/test/test_topology.py#L593

You don't need to be quite as thorough and complex as PyMongo is, but some sort of intelligence about error messages would be great.

bjori: I agree we should try to resolve the fundamental error to human digestible error message -- I'm a littlebit worried though that we may be hiding away the gory details.

It seems like if there is any connectible node then pymongo doesn't tell you why there isn't a matching node. It just tells you that there wasn't a matching node for the 'selector'.

I guess there is no way to correlate server errors to the "selector", but we could attach the server errors as a "full details" option somehow?
Not sure how to handle that though, but I was really hoping to be able to provide transport error messages :/
We should at least make sure they are logged.

src/mongoc/mongoc-topology.c
bjori: Yeah, the optype + read_pref should definitely be part of the fallthrough error message

src/mongoc/mongoc-topology.c
bjori: Which looping?
There isn't a automatic retry-loop anymore.

I do believe we should set topology->stale before the RETURN (NULL) case though, so subsequent user-triggered explicit retries has a chance of succeeding by triggering a rescan.
Although - I guess that is replaced with the blacklisting now.

src/mongoc/mongoc-topology-scanner.c
ajdavis: I think we should fill the remainder of the bson_error_t's 504-byte error message with a list of _mongoc_server_description_t.error strings, or provide a way to retrieve those errors from a topology. If there are no errors but we still can't select a server, just print a list of server types: "Can't select a server for read preference "primary", with servers of type SECONDARY, SECONDARY, ARBITER." If there are *no* servers of known type, then "All servers are down."

src/mongoc/mongoc-topology.c
ajdavis: Oh, wait, nevermind. We have to keep the "for (;;)" loop in. Don't change this code until / unless the SS spec changes. We're still implementing this:

https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#single-threaded-server-selection-implementation

In the future, an expire_at value of -42 or something may have special meaning, but not yet.

bjori: Right -- that retry-logic is the root of SPEC-217.
The coolDown SDAM option was for subsequent user triggered actions, which tweaks _mongoc_topology_time_to_scan() decision to call for a scan or not.

PR #246, opened 2015-07-01
CDRIVER-731: Do not trace SASL mutex callbacks

jmikola:
https://jira.mongodb.org/browse/CDRIVER-731

### DIFF
--- a/src/mongoc/mongoc-init.c
+++ b/src/mongoc/mongoc-init.c
@@ -37,46 +37,36 @@ mongoc_sasl_mutex_alloc (void)
 {
    mongoc_mutex_t *mutex;
 
-   ENTRY;
-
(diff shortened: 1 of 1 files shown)


ajdavis:
LGTM if you add a comment about why these functions shouldn't be traced.

src/mongoc/mongoc-init.c
```diff
@@ -90,6 +80,8 @@ static MONGOC_ONCE_FUN( _mongoc_do_init)
 #endif
 
 #ifdef MONGOC_ENABLE_SASL
+   /* The following functions should not use tracing, as they may be invoked
+    * before mongoc_log_set_handler() can complete. */
```
jmikola: @ajdavis: this seemed preferable to copy/pasting it for each of the four functions.

ajdavis:
Thanks!

PR #247, opened 2015-07-02
Fixing debian changelog for package

maybeshewill:


ajdavis:
Thanks, I'll fix my changelog script so it generates dates in the right format.

I'm curious, how did you notice this? We don't actually release a .deb file so I wonder why this was a problem for you?

maybeshewill:
I'm porting my current code from legacy driver to curreent because of not allowed read-only conections to secondary nodes in legacy driver. 

Decided to make a side deb package in my project and met these. Also there are some problems with tests by default if no mongo is running or it is configured in some other way from default.

ajdavis:
Thanks for the info. Yes, the tests require a mongod running, and how the
tests run is not ideal. There are instructions in CONTRIBUTING for
configuring your tests to match your mongod's username, password, host, and
port.

maybeshewill:
I'll make some pull requests on tests if i'll have enough time.

Thx for reply!

ajdavis:
Let me know if you plan to contribute to the tests, we can discuss: jesse@mongodb.com I want to make sure we agree on the general strategy before you spend effort writing code.

PR #251, opened 2015-07-24
B-0 fixed bugs in GridFSFile related to seeking and flushing current page

jsbattig:
Fixed bug when seeking into a  page has been deleted, and the page is not available in cursor. Signs used to compare if page is available were reversed in the comparison.
Also fixed problem flushing page when positioned one byte after the end of the page

### DIFF
--- a/src/mongoc/mongoc-gridfs-file-private.h
+++ b/src/mongoc/mongoc-gridfs-file-private.h
@@ -59,6 +59,7 @@ struct _mongoc_gridfs_file_t
    const char                *bson_content_type;
    bson_t                     bson_aliases;
    bson_t                     bson_metadata;
+   bool                       pos_after_read_write;
 };
 
 
--- a/src/mongoc/mongoc-gridfs-file.c
+++ b/src/mongoc/mongoc-gridfs-file.c
@@ -598,7 +598,7 @@ _mongoc_gridfs_file_refresh_page (mongoc_gridfs_file_t *file)
       /* if we have a cursor, but the cursor doesn't have the chunk we're going
        * to need, destroy it (we'll grab a new one immediately there after) */
       if (file->cursor &&
-          !(file->cursor_range[0] >= n && file->cursor_range[1] <= n)) {
+          !(file->cursor_range[0] <= n && file->cursor_range[1] >= n)) {
          mongoc_cursor_destroy (file->cursor);
          file->cursor = NULL;
       }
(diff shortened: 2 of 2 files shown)


ajdavis:
Thanks for this, I'm sorry I haven't reviewed it yet. We may not merge until after the upcoming 1.2-beta release, but I certainly appreciate the contribution and I'll take a close look soon.

ajdavis:
Still haven't forgotten about this, just continuing to put it off while we handle higher-priority issues in the 1.2 beta period.

jsbattig:
No problem.
Jesse, one more thing.
I saw on some comment I think in Jira or somewhere that apparently the team
working on the driver was planning to make GridFS files be almost
"unidirectional" streams at some point.
We definitively use them as "streams" and not like unidirectional files.
Why is this idea? I remember while working in the legacy driver, I went
thorough lengths to clean the GridFS code and make it stable and reliable
work as a bi-directional stream with minimal buffering to improve
performance.
That's how we use GridFS file here at Convey, so if you decide to go with
unidirectional mode it will force us to implement our wrapper on top using
two GridFS objects and somehow trying to keep them sync when operations
doing input/output get interleaved.

Thanks

ajdavis:
@ksuarz will you please take a look at this along with your other GridFS work in the C Driver?

ksuarz:
Sorry for the delay -- I've been looking at this in our JIRA project.

https://jira.mongodb.org/browse/CDRIVER-765

jsbattig:
@ksuarz , thanks for checking at my Pull Request.
Also, have you made any decision guys on making GridFS files unidirectional? I think it would be a bad idea... in fact, I would err on the side of making it high efficiency and bi-directional, otherwise you are cutting off many potential use cases that need to deal with Streams (bi-directional)

ajdavis:
Hi @jsbattig I changed my mind about that. We have an existing API that allows you to read, write, and seek in a GridFS file (until you close it). So with your help we're fixing all the bugs and improving documentation in that existing API.

There's a new spec that all MongoDB drivers must comply with in the next year, to reduce our inconsistencies:

https://github.com/10gen/specifications/tree/master/source/gridfs

I'd originally thought the spec prohibited mixed reads and writes, so I was planning to remove them in the C Driver 2.0. But after discussion with the GridFS spec authors we've decided that the C driver's ability to seek, read, and write in unclosed GridFS files is an acceptable extension to the spec. Developers like you depend on those features, so we won't remove that ability.

I added a comment on [CDRIVER-511, "Files written to GridFS must be saved before being read back"](https://jira.mongodb.org/browse/CDRIVER-511):

> Editing myself: we can't break backwards compatibility, we promised that mixed reads & writes are allowed so we have to fix any logic errors related to that. The GridFS spec does not mention mixed reads and writes, so it doesn't prohibit us from continuing to support (and bugfix) that feature.

ajdavis:
Thanks @jsbattig! We've included your comparison fix in https://github.com/mongodb/mongo-c-driver/pull/273. We took a more general approach to fixing the hack that the driver had used for calculating the page number.

PR #254, opened 2015-07-29
Removed memory leak in function mongoc_database_has_collection

canella:
If an invalid namespace is supplied to mongoc_database_has_collection the cursor that is used in this function is not freed causing a memory leak every time the function is called. This leads to a drastic increase in memory consumption if the function is called frequently.

### DIFF
--- a/src/mongoc/mongoc-database.c
+++ b/src/mongoc/mongoc-database.c
@@ -613,13 +613,15 @@ mongoc_database_has_collection (mongoc_database_t *database,
    BSON_APPEND_UTF8 (&filter, "name", name);
 
    cursor = mongoc_database_find_collections (database, &filter, error);
-
-   if (!cursor ||
-       (error &&
-        ((error->domain != 0) ||
-         (error->code != 0)))) {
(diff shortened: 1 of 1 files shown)


canella:
Two pictures showing the memory consumption before and after applying the fix. In both cases the program ran over a minute while calling the function continuously.

mongoc_database_has_collection_fix
mongoc_database_has_collection_nofix

src/mongoc/mongoc-database.c
```diff
@@ -618,6 +618,7 @@ mongoc_database_has_collection (mongoc_database_t *database,
        (error &&
         ((error->domain != 0) ||
          (error->code != 0)))) {
+      mongoc_cursor_destroy( cursor );
```
bjori: If we enter this clause with !cursor then mongoc_cursor_destroy() is going to be very pissed off.

canella: You are of course right, I had only tested it with an error, not an invalid cursor. I will fix this tonight by separating the if clause for cursor and error.

ajdavis:
Thanks for the contribution!

PR #255, opened 2015-07-29
MFH

bjori:
We really should do this more frequently.


src/mongoc/mongoc-topology-scanner-private.h
```diff
@@ -94,8 +93,7 @@ mongoc_topology_scanner_node_destroy (mongoc_topology_scanner_node_t *node,
 
 void
 mongoc_topology_scanner_start (mongoc_topology_scanner_t *ts,
-                               int32_t timeout_msec,
-                               bool obey_cooldown);
+                               int32_t                    timeout_msec);
```
ajdavis: obey_cooldown will be *deleted* by this merge?

tests/test-mongoc-client.c
```diff
@@ -686,6 +927,15 @@ test_mongoc_client_ipv6 (void)
 void
 test_client_install (TestSuite *suite)
 {
+   bool local;
+   local = !getenv ("MONGOC_DISABLE_MOCK_SERVER");
```
ajdavis: This was never right, let's not reintroduce the "local" bool or the MONGOC_DISABLE_MOCK_SERVER env var. We should always run the mock server tests.

src/mongoc/mongoc-topology-scanner-private.h
bjori: Fixed nice catch

tests/test-mongoc-client.c
bjori: fixed

ajdavis:
Thank you!!

PR #259, opened 2015-08-17
B-0 fixed critical error with connection pools

jsbattig:
In its current state connection pool doesn't work at all.
Object on pool are continuously destroyed every time an object is popped from it generating constant reconnections.


jsbattig:
I think my bad on this one. apparently it works as designed, but min_pool_size will be better to default to a value > 0

ajdavis:
Thanks for the info, would you like to file a bug report at jira.mongodb.org
in the "CDRIVER" project so we can discuss it there?

jsbattig:
I will, it's probably not a "bug" but it's an issue...

ajdavis:
Thanks for bringing it to our attention.

jsbattig:
No problem.
I created an issue for you guys to think about it if you want to do
something.

Sebastian

PR #260, opened 2015-08-20
B-0 fixed seek method. Bad interpretation of seek_end mode. it should…

jsbattig:
… seek from one byte more than number of bytes written. seek_end with offset zero should put pointer exactly at the end ready to receive more data

### DIFF
--- a/src/mongoc/mongoc-gridfs-file.c
+++ b/src/mongoc/mongoc-gridfs-file.c
@@ -709,7 +709,7 @@ mongoc_gridfs_file_seek (mongoc_gridfs_file_t *file,
       offset = file->pos + delta;
       break;
    case SEEK_END:
-      offset = (file->length - 1) + delta;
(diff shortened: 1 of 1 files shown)


ajdavis:
Thanks for this, and sorry again for the delay. Over the next six weeks @ksuarz and I will address your concerns about the GridFS implementation.

For this pull request, we can accept it if you add a test to test-mongoc-gridfs-file-page.c that demonstrates the original bug and shows how your patch fixes it.

Thank you!

ksuarz:
Apologies for the delay. I'm taking a look at this, and I've opened a ticket in our JIRA project:

https://jira.mongodb.org/browse/CDRIVER-839

ksuarz:
Thanks @jsbattig , after examining the seeking behavior you're totally right. We'll merge this pull request and then add new tests to ensure the proper behavior.

PR #262, opened 2015-08-31
allow driver to use mongoc-log-private

remicollet:
the mongo-php-driver-prototype (pecl ext mongodb 1.0.0alpha2) needs this.


bjori:
Thanks!

remicollet:
Fast!
thanks

PR #264, opened 2015-09-03
edits .gitmodules, fixes libbson url

rubicks:
was getting socket error timeouts on `git submodule update --init` within `autogen.sh`, this change makes it work as intended.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "src/libbson"]
 	path = src/libbson
-	url = git://github.com/mongodb/libbson.git
+	url = https://github.com/mongodb/libbson.git


ajdavis:
Interesting. You may be behind a proxy or firewall that blocks the git protocol but allows HTTPS. Does `git clone git://github.com/mongodb/libbson.git` in a separate directory work for you, or does that time out also?

rubicks:
Here's what that looks like for me:

    $ git clone git://github.com/mongodb/libbson.git
    Cloning into 'libbson'...
    fatal: unable to connect to github.com:
    github.com[0: 192.30.252.128]: errno=Operation timed out

I think we're agreed: my employer's firewall is probably to blame. Would my edit break anything for other people?

ajdavis:
Thanks! @rozza pointed out to me that Github recommends HTTPS:

https://help.github.com/articles/which-remote-url-should-i-use/#cloning-with-https-recommended

rubicks:
Sweet! Thanks @ajdavis

PR #270, opened 2015-09-17
CDRIVER-851: Negative wtimeout values are invalid

jmikola:
https://jira.mongodb.org/browse/CDRIVER-851

### DIFF
--- a/src/mongoc/mongoc-write-concern.c
+++ b/src/mongoc/mongoc-write-concern.c
@@ -198,13 +198,20 @@ mongoc_write_concern_get_wtimeout (const mongoc_write_concern_t *write_concern)
  * @wtimeout_msec: Number of milliseconds before timeout.
  *
  * Sets the number of milliseconds to wait before considering a write
- * request as failed.
+ * request as failed. A value of 0 indicates no write timeout.
+ *
+ * The @wtimeout_msec parameter must be positive or zero. Negative values will
+ * be ignored.
  */
 void
 mongoc_write_concern_set_wtimeout (mongoc_write_concern_t *write_concern,
                                    int32_t            wtimeout_msec)
 {
    BSON_ASSERT (write_concern);
+
+   if (wtimeout_msec < 0) {
+      return;
+   }
 
    if (!_mongoc_write_concern_warn_frozen(write_concern)) {
       write_concern->wtimeout = wtimeout_msec;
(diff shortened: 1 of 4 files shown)


src/mongoc/mongoc-write-concern.c
```diff
@@ -205,6 +205,7 @@ mongoc_write_concern_set_wtimeout (mongoc_write_concern_t *write_concern,
                                    int32_t            wtimeout_msec)
 {
    bson_return_if_fail(write_concern);
+   bson_return_if_fail(wtimeout_msec >= 0);
```
jmikola: This is prone to printing the following on stderr when running the tests with a debug build:

```
mongoc_write_concern_set_wtimeout(): precondition failed: wtimeout_msec >= 0
```

I talked with @ajdavis about the downsides of using `bson_return_if_fail()` to check user input, noting that it is compiled out of non-debug builds; however, we agreed it best to deal with that in a separate ticket (likely an epic).

In this case, is there anything I can do to avoid spamming stderr when tests are executed? Do any other tests trigger these `bson_return_*()` conditions?

ajdavis: That's a great question. I think you should not use bson_return_if_fail, even though using it is consistent with the surrounding code. In this case, consistency with bad code is bad. As long as we are improving the driver by validating wtimeout_msec, and improving it further by adding good tests, let's be inconsistent: if wtimeout_msec < 0, MONGOC_WARN the user and refuse to set the value.

If you would like to add a test framework feature to catch and verify the warning instead of printing it to stderr, that'd be awesome. Otherwise I can come in after you and do that.

Eventually CDRIVER-697 will require a wholesale replacement of bson_return_if_fail to validate user input, along with tests of the improved validation. In the meantime, let's head in that direction incrementally instead of adding more bad uses of bson_return_if_fail.

bjori: The MONGOC_WARNING() and friends are problematic for drivers built on top of mongoc.
It has no semantical understanding of what raised the warning or how to convey that to the caller.

If a function cal fail it must report that in form of a return value or out param. Or invent errno.

This specific case can be fixed by documenting it to be a programming error to pass wtimeout < 0, and say such value will be ignored.

The bson_return_if_fail() are for checking complex types/pointers, not to validate their value.
Illegal values should be checked for by normal input validation.

jmikola:
Updated functions and documentation to ignore negative `wtimeout_msec` values. I think this is ready for a second look.

ajdavis:
LGTM.


---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mongodb/mongo-c-driver/270)

ajdavis:
<sup>**[doc/mongoc_write_concern_set_wmajority.page, line 24 \[r2\]](https://reviewable.io:443/reviews/mongodb/mongo-c-driver/270#-K02a7u3qP6bnriSkxtB)** ([raw file](https://github.com/mongodb/mongo-c-driver/blob/934dd338b3717e5235659b20d9349e7dfd3f6365/doc/mongoc_write_concern_set_wmajority.page#L24)):</sup>
Glad to see a grammatical fix slipped in with the technical improvement.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mongodb/mongo-c-driver/270)

PR #277, opened 2015-09-29
CDRIVER-886: fix wrong HTML docs configure flag

ksuarz:
Second attempt, now with correct names.

The correct flag is `--enable-html-docs`.


bjori:
kudos

PR #281, opened 2015-10-02
CDRIVER-822: clean generated documentation

ksuarz:
Running `make clean` now also cleans the generated HTML documentation and man pages. Yelp HTML files have been added to `CLEANFILES` as well.

There are also new rules for cleaning just the documentation: "html-clean" and "man-clean".

Thanks @bjori for the help and suggestions!


bjori:
Thanks!

PR #283, opened 2015-10-04
CDRIVER-903: Hinted cursors should still apply read prefs

jmikola:
https://jira.mongodb.org/browse/CDRIVER-903

Note: you may want to double-check that the test adjustments. I'm not too familiar with how the request mocking works with respect to topology types, but I assumed that it was mainly using standalone connections, where slaveOk would be injected.

### DIFF
--- a/src/mongoc/mongoc-client.c
+++ b/src/mongoc/mongoc-client.c
@@ -1159,7 +1159,7 @@ mongoc_client_command_simple (mongoc_client_t           *client,
                               bson_error_t              *error)
 {
    return _mongoc_client_command_simple_with_hint (client, db_name, command,
-                                                   read_prefs, reply, 0, error);
+                                                   read_prefs, false, reply, 0, error);
 }
 
 bool
@@ -1167,6 +1167,7 @@ _mongoc_client_command_simple_with_hint (mongoc_client_t           *client,
                                          const char                *db_name,
                                          const bson_t              *command,
                                          const mongoc_read_prefs_t *read_prefs,
+                                         bool                       is_write_command,
                                          bson_t                    *reply,
                                          uint32_t                   hint,
                                          bson_error_t              *error)
@@ -1183,6 +1184,7 @@ _mongoc_client_command_simple_with_hint (mongoc_client_t           *client,
                                    command, NULL, read_prefs);
 
    cursor->hint = hint;
+   cursor->is_write_command = true;
 
    ret = mongoc_cursor_next (cursor, &doc);
 
--- a/src/mongoc/mongoc-cursor-private.h
+++ b/src/mongoc/mongoc-cursor-private.h
@@ -55,14 +55,15 @@ struct _mongoc_cursor_t
    uint32_t                   hint;
    uint32_t                   stamp;
 
-   unsigned                   is_command   : 1;
-   unsigned                   sent         : 1;
-   unsigned                   done         : 1;
-   unsigned                   failed       : 1;
-   unsigned                   end_of_event : 1;
-   unsigned                   in_exhaust   : 1;
-   unsigned                   redir_primary: 1;
-   unsigned                   has_fields   : 1;
+   unsigned                   is_command      : 1;
+   unsigned                   sent            : 1;
+   unsigned                   done            : 1;
+   unsigned                   failed          : 1;
+   unsigned                   end_of_event    : 1;
+   unsigned                   in_exhaust      : 1;
+   unsigned                   redir_primary   : 1;
+   unsigned                   has_fields      : 1;
+   unsigned                   is_write_command: 1;
 
    bson_t                     query;
    bson_t                     fields;
--- a/tests/test-mongoc-collection.c
+++ b/tests/test-mongoc-collection.c
@@ -1451,7 +1451,7 @@ test_aggregate_legacy (void *data)
 
    /* no "cursor" argument */
    request = mock_server_receives_command (
-      server, "db", MONGOC_QUERY_NONE,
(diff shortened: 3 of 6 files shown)


tests/test-bulk.c
```diff
@@ -1557,7 +1557,7 @@ _test_write_concern (bool has_write_commands, bool ordered)
       request = mock_server_receives_command (
          mock_server,
          "test",
-         MONGOC_QUERY_NONE,
+         MONGOC_QUERY_SLAVE_OK,
```
ajdavis: It seems minorly buggy to me that the driver now sets SlaveOk on write commands. Server Selection Spec: ["clients MUST always set the slaveOK wire protocol flag on reads"](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#topology-type-single). It's reasonable to infer that it shouldn't set SlaveOk on writes, and other drivers (PyMongo) don't, that I know of.

Worse, I think the driver will now send $readPreference in write commands when connected to a single mongos?

The reason it didn't set SlaveOk / $readPreference on writes before, is because write commands set the cursor's hint:

https://github.com/mongodb/mongo-c-driver/blob/1941a0a6fd7271ac26d7a3a152df30f810a662a8/src/mongoc/mongoc-client.c#L1186-L1186

The hint comes from a call to _mongoc_client_preselect up in _mongoc_write_command_execute. Call stack, most recent call on top:
```
_mongoc_cursor_query mongoc-cursor.c:602
_mongoc_cursor_next mongoc-cursor.c:948
mongoc_cursor_next mongoc-cursor.c:876
_mongoc_client_command_simple_with_hint mongoc-client.c:1188
_mongoc_write_command mongoc-write-command.c:889
_mongoc_write_command_execute mongoc-write-command.c:961
```
It works by accident: since write commands always set a hint on the cursor, they aren't sent with SlaveOk.

Now, I think, we need a private bool member is_write_command, and a private function 
```
_mongoc_cursor_set_write_command (cursor, bool)
```
If the bool is set, don't apply read preferences. 

I guess _mongoc_client_command_simple_with_hint needs an is_write_command parameter now, too, so _mongoc_write_command can pass it as "true".

src/mongoc/mongoc-cursor-private.h
```diff
@@ -55,14 +55,15 @@ struct _mongoc_cursor_t
...
-   unsigned                   has_fields   : 1;
+   unsigned                   is_command      : 1;
+   unsigned                   sent            : 1;
+   unsigned                   done            : 1;
+   unsigned                   failed          : 1;
+   unsigned                   end_of_event    : 1;
+   unsigned                   in_exhaust      : 1;
+   unsigned                   redir_primary   : 1;
+   unsigned                   has_fields      : 1;
+   unsigned                   is_write_command: 1;
```
jmikola: ASCII art for @samantharitter :+1:

src/mongoc/mongoc-client.c
```diff
@@ -1183,6 +1184,7 @@ _mongoc_client_command_simple_with_hint (mongoc_client_t           *client,
                                    command, NULL, read_prefs);
 
    cursor->hint = hint;
+   cursor->is_write_command = true;
```
jmikola: I reasoned that we didn't need a private setter function, since the driver already has access to the cursor struct.

ajdavis:
Almost there!


---

Reviewed 6 of 6 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[src/mongoc/mongoc-cursor.c, line 651 \[r1\]](https://reviewable.io:443/reviews/mongodb/mongo-c-driver/283#-K07Z0kpFH4kIHAa7nS2)** ([raw file](https://github.com/mongodb/mongo-c-driver/blob/2282e159ede254b5f255435c40360ee5eef40521/src/mongoc/mongoc-cursor.c#L651)):</sup>
Let's refactor this repeated logic.


---

<sup>**[tests/test-mongoc-collection.c, line 1802 \[r1\]](https://reviewable.io:443/reviews/mongodb/mongo-c-driver/283#-K07YUf2x3Nb1EfHoto5)** ([raw file](https://github.com/mongodb/mongo-c-driver/blob/2282e159ede254b5f255435c40360ee5eef40521/tests/test-mongoc-collection.c#L1802)):</sup>
Do we need this change?


---

<sup>**[tests/test-mongoc-collection.c, line 1844 \[r1\]](https://reviewable.io:443/reviews/mongodb/mongo-c-driver/283#-K07YhKft7fdPKmSYGSL)** ([raw file](https://github.com/mongodb/mongo-c-driver/blob/2282e159ede254b5f255435c40360ee5eef40521/tests/test-mongoc-collection.c#L1844)):</sup>
Do we need this change for some reason?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/mongodb/mongo-c-driver/283)

tests/test-mongoc-collection.c
```diff
@@ -1451,7 +1451,7 @@ test_aggregate_legacy (void *data)
 
    /* no "cursor" argument */
    request = mock_server_receives_command (
-      server, "db", MONGOC_QUERY_NONE,
+      server, "db", MONGOC_QUERY_SLAVE_OK,
```
jmikola: > Do we need this change?

The test does fail without this change. The only other use of mock_server_receives_command() in this file (excluding the other aggregate test below) was for an insert command, which we've decided not to inject slaveOk into per earlier discussion.

jmikola:
Superseded by #292, which contains requested revisions from this PR re-opened against the master branch.

PR #285, opened 2015-10-06
CDRIVER-701 partial fix

bauman:
now allows the leaf compare to be more mongo like if right side is an array.  Still returns correctly for previous tests on exact document match.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -259,6 +259,11 @@ mongoc_add_test(test-sharded-cluster FALSE
    ${SOURCE_DIR}/tests/test-sharded-cluster.c
    ${SOURCE_DIR}/tests/mongoc-tests.c
    ${SOURCE_DIR}/tests/ha-test.c)
+mongoc_add_test(test-matcher-cdriver701 FALSE
+        ${SOURCE_DIR}/tests/test-mongoc-matcher_cdriver701.c
+        ${SOURCE_DIR}/tests/mongoc-tests.c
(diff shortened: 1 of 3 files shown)


ajdavis:
Thanks for this, taking a look now.

ajdavis:
Why do you say this is a "partial" fix? What's missing regarding array matching?

bauman:
It's got 100% coverage for -701.

It will only match correctly if the list is the last item.

{'a': {'b': [1, 2]}}
spec {"a.b":1} correctly works.

I've spent a few cylces trying to make it work with lists inside the namespace, but at the end of the day, I don't need it.  As you say, it's not supposed to be close to mongodb.  Decided to fight that battle another day and do the pull now since it does what I need.

{'a': [{'b': [1, 2]}, {'b': [3, 4]}]}
spec {"a.b":1} does not work.

user would need to know a = a list of 2 items and build 2 specs
{"a.0.b":1} && {"a.1.b":1} would work.

ajdavis:
I see. The reason the matcher doesn't work like mongod's is (for this scenario) that the C Driver's matcher uses bson_iter_find_descendant to search for "a.b" in the document. bson_iter_find_descendant search through arrays like mongod's matcher does.

ajdavis:
@bauman as we've discussed, we're not continuing development of the matcher, [it'll be removed in libmongoc 2.0](https://jira.mongodb.org/browse/CDRIVER-955).

PR #293, opened 2015-10-15
Fix UNIX domain socket connection

u2yg:
It concatenated the port number to the filename and thus failed on connection.


bjori:
Thanks!

This is a little embarrassing. We are looking into how we can automate testing for this before we merge.
See https://jira.mongodb.org/browse/CDRIVER-939

u2yg:
Thanks. That would be great. We're facing difficulty in automatically testing this ourselves, so I'm looking forward to what you will come up with.

bjori:
Merged as part of CDRIVER-939
https://github.com/mongodb/mongo-c-driver/commit/6614693e481b3614a8ae8af2dc10bac2c768a82a among others.

Thanks for discovery & fixing this!

PR #295, opened 2015-11-02
Define variables at the beginning of scope

alexeyvo:
To allow compilation using MSVC 2010

### DIFF
--- a/tests/test-mongoc-uri.c
+++ b/tests/test-mongoc-uri.c
@@ -567,27 +567,19 @@ test_mongoc_uri_read_prefs (void)
    const read_prefs_test *t;
    int i;
 
-   bson_t *tags_dcny;
-   bson_t *tags_dcny_empty;
-   bson_t *tags_dcnyusessd_dcsf_empty;
-   bson_t *tags_empty;
-
-   tags_dcny = BCON_NEW(
+   bson_t *tags_dcny = BCON_NEW(
       "0", "{", "dc", "ny", "}"
    );
-
-   tags_dcny_empty = BCON_NEW(
+   bson_t *tags_dcny_empty = BCON_NEW(
       "0", "{", "dc", "ny", "}",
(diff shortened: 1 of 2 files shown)


tests/test-mongoc-uri.c
```diff
@@ -567,27 +567,19 @@ test_mongoc_uri_read_prefs (void)
...
-
-   tags_dcnyusessd_dcsf_empty = BCON_NEW(
+   bson_t *tags_dcnyusessd_dcsf_empty = BCON_NEW(
       "0", "{", "dc", "ny", "use", "ssd", "}",
       "1", "{", "dc", "sf", "}",
       "2", "{", "}"
    );
-
-   tags_empty = BCON_NEW(
+   bson_t *tags_empty = BCON_NEW(
```
bjori: Does this actually change anything?

alexeyvo: This code doesn't compile using MSVS 2010 see https://jira.mongodb.org/browse/CDRIVER-981?jql=project%20%3D%20CDRIVER%20AND%20resolution%20%3D%20Unresolved%20ORDER%20BY%20priority%20DESC
MSVS doesn't support c99 standard(only partial support). ansi c standard requires variable definition at the beginning of the scope.

bjori: I understand, but I'm not seeing where exactly a variable is defined elsewhere in this change?

All the variables are defined at the top, and then assigned to BCON_NEW(...).

All you changed here was to do the assignment with the variable declaration

bjori: Ahh!! I see it now. The *next* line continues the variable declarations. sorry, got it! thanks :)

alexeyvo: Yes there are another variable definition below, so if we define and declare this variables at the same time all variables will be declared at the top of scope

PR #296, opened 2015-11-03
Cleanup opessl properly

alexeyvo:
Some memory leak decetion tools(vld, valgrind) reports memory leaks


bjori:
Hi @alexeyvo 

Thanks for the effort you've put into this, but I'm not sure this is the right thing to do.

Have a look at https://jira.mongodb.org/browse/CDRIVER-549?focusedCommentId=842108&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-842108

alexeyvo:
makes sense

PR #297, opened 2015-11-19
CDRIVER-942 add copy methods for db and collection

Machyne:
This adds methods for copying `mongoc_database_t`'s and `mongoc_collection_t`'s.

These would be nice to add into the C++11 driver :) Happy to make changes if you have feedback!


ajdavis:
Matt, features are like children: conceived in a moment of passion, they must be supported for decades. Let's talk about how necessary these new functions are.

PR #298, opened 2015-11-19
Fix date specfile date format

vleschuk:
rpmbuild (at least on CentOS 6) doesn't understand
the following date format in %changelog: Thu October 29 2015 ...
It requires short month names: Thu Oct 29 2015 ...

### DIFF
--- a/build/rpm/mongo-c-driver.spec
+++ b/build/rpm/mongo-c-driver.spec
@@ -114,25 +114,25 @@ find $RPM_BUILD_ROOT -name '*.la' -exec rm -f {} ';'
 
 
 %changelog -n %{DriverName}
-* Thu October 29 2015 A. Jesse Jiryu Davis <jesse@mongodb.com> - 1.2.1-1
+* Thu Oct 29 2015 A. Jesse Jiryu Davis <jesse@mongodb.com> - 1.2.1-1
 - Release 1.2.1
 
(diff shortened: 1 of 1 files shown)


bjori:
This file should actually probably be removed as we don't actually really maintain it - as you can tel if it doesn't even work :)
It even appears to be incorrectly located: https://jira.mongodb.org/browse/CDRIVER-883

ajdavis:
I agree, we should remove it. We've just got approval for a project to properly package libbson and libmongoc for Fedora this quarter.

@vleschuk I'd like to know more about your needs. Could you comment here, or email me jesse@mongodb.com, and tell me more about why you were trying to create an RPM for libmongoc?

vleschuk:
Hello, thanks for the answers. 

> tell me more about why you were trying to create an RPM for libmongoc

I am using mongoc-c-client in one of my projects (target platform is CentOS 6, a significant number of machines). I decided NOT to make "slackware" from the all the systems with ./configure && make && make install but to provide the customer with all required dependencies as RPMS (maybe with private repository). The spec file actually works (after fixes I showed in patch) and resulting rpm packages work fine. Also the final application will be distributed as rpm too, and I would like to set proper dependencies.

ajdavis:
Thanks for the answer. Does @remicollet's RPM package of the driver work for you instead?:

http://rpmfind.net/linux/rpm2html/search.php?query=mongo-c-driver

We'd rather not have two different RPMs for the driver to maintain.

PR #300, opened 2015-11-23
Fix C99 support

fortizc:
Warning in C99

### DIFF
--- a/src/mongoc/mongoc-socket.h
+++ b/src/mongoc/mongoc-socket.h
@@ -43,7 +43,7 @@
 
 BSON_BEGIN_DECLS
 
-
+struct addrinfo;
 typedef struct _mongoc_socket_t mongoc_socket_t;
 
 typedef struct


ajdavis:
Hi, can you tell me how you're compiling the driver, and what error message you see without this change? I'd like to know the OS and compiler you're using, and I want to see the sequence of commands you do that results in a build failure.

fortizc:
Hi, it seems to be partly my mistake, today I update my copy of mongo-c-driver compile and test, but only show a warning.
All of this is tested in Debian Jessie with GCC 4.9.2 the driver is compiled as follow:
```
$ ./autogen.sh --with-libbson=bundled
$ make
$ sudo make install
```
The warning is:
```
In file included from /home/fortiz/lib/include/libmongoc-1.0/mongoc-stream.h:25:0,
                 from /home/fortiz/lib/include/libmongoc-1.0/mongoc-gridfs.h:26,
                 from /home/fortiz/lib/include/libmongoc-1.0/mongoc-client.h:30,
                 from /home/fortiz/lib/include/libmongoc-1.0/mongoc.h:26,
                 from insert.c:2:
/home/fortiz/lib/include/libmongoc-1.0/mongoc-socket.h:98:44: warning: 'struct addrinfo' declared inside parameter list
                                            size_t                   buflen);
                                            ^
/home/fortiz/lib/include/libmongoc-1.0/mongoc-socket.h:98:44: warning: its scope is only this definition or declaration, which is probably not what you want
```
With the proposed change the warning disappears. I update the PR message

ajdavis:
And what's in "insert.c" please and how are you compiling it? I want to make sure I see the same warning as you, before I accept your change. Thanks!

fortizc:
insert.c is an example taken from [here](http://api.mongodb.org/c/1.2.0/tutorial.html#insert) and I use this Makefile:
```
CC=gcc
INCLUDES=-I/home/fortiz/lib/include/libmongoc-1.0 -I/home/fortiz/lib/include/libbson-1.0
COPTS=-Wall -g -std=c99
CFLAGS=$(COPTS) $(INCLUDES)
LDLIBS=-L/home/fortiz/lib/lib -lsasl2 -lssl -lcrypto -lrt -lmongoc-1.0 -lbson-1.0
TARGETS=insert

all:$(TARGETS)

clean:
	rm -f $(TARGETS)

default:
	$(CC) $(CFLAGS) $< -o $@ $(LDLIBS)
```
I don't use pkg-config because I install mongo-c-driver in a custom path for test.

ajdavis:
This has taken a bit of research on my part, but this isn't a C Driver bug. "struct addrinfo" and everything else related to sys/socket.h is not in the C99 standard. If you want to compile with C99 you can use "-std=gnu99" instead of "-std=c99". Or, you can use "-std=c99 -D_XOPEN_SOURCE=600".

fortizc:
Ok, thanks for check it. I believe that this must to be present in the documentation, for example [here](https://docs.mongodb.org/ecosystem/drivers/c/) say C99 not some POSIX variants

ajdavis:
Good point, thanks.

PR #304, opened 2015-12-14
CMakeLists.txt build is incomplete and makes the driver unusable for the C++ driver

duckie:
See JIRA issue: https://jira.mongodb.org/browse/CDRIVER-1052

The CMakeLists.txt is missing installation instructions for a header file and for pkg-config files.

Therefore, a installation with the CMakeLists.txt makes pkg-config unable to detect the library. This affects for instance any build of the C++ driver.

The missing header affects at least the build of the new C++ driver.


acmorrow:
FYI - I have a PR that hopefully removes the need for the private symbol: https://github.com/mongodb/mongo-cxx-driver/pull/400

CMakeLists.txt
```diff
@@ -379,3 +380,30 @@ install(
   FILES ${HEADERS}
   DESTINATION "include/libmongoc-${MONGOC_API_VERSION}"
 )
+
+# Define pkg-config files
+set(VERSION "${MONGOC_VERSION}")
+set(prefix "${CMAKE_INSTALL_PREFIX}")
+set(libdir "\${prefix}/lib")
```
bjori: Is this backspace intentional? It looks out of place

CMakeLists.txt
```diff
@@ -177,6 +177,7 @@ set (HEADERS
    ${SOURCE_DIR}/src/mongoc/mongoc-opcode.h
    ${SOURCE_DIR}/src/mongoc/mongoc-opcode-private.h
    ${SOURCE_DIR}/src/mongoc/mongoc-read-concern.h
+   ${SOURCE_DIR}/src/mongoc/mongoc-read-concern-private.h
```
acmorrow: Please note that the C++11 driver no longer requires this private header. I recommend against installing it.

CMakeLists.txt
acmorrow: I think it is there so that CMake doesn't expand it, and instead creates a variable containing a literal `${prefix}`, as this is what is expected in the interpolated .pc file.

duckie: @acmorrow is right, it is meant to avoid being expanded as a variable

CMakeLists.txt
duckie: I will remove it from the push request

bjori:
Merged as https://github.com/mongodb/mongo-c-driver/commit/3ef2137fb77040591a9e1ea7c294f3292ba0cbf1 and https://github.com/mongodb/mongo-c-driver/commit/ae550197df66e03d1199224e1edec4c635d3a872

Thanks!

PR #309, opened 2016-01-21
Tweaks for 1.3.1-1 Debian packaging

rcsanchez97:
Here are the last few tweaks I had to make prior to uploading 1.3.1-1.


ajdavis:
Thanks!

PR #310, opened 2016-01-26
CDRIVER-1054: Support negative cursor limit

jmikola:
https://jira.mongodb.org/browse/CDRIVER-1054

This adds internal support for specifying a negative cursor limit (i.e. single-batch mode) for both OP_QUERY and find command code paths. Note that this does not change the public API, which will be handled for 2.0 by CDRIVER-1053.


ajdavis:
Gorgeous.

PR #311, opened 2016-01-27
Workaround to create an issue

EmielBruijntjes:
When you follow the instructions from the README file, you do not end up with a working mongo c driver. This is either a mistake in the README file, or the latest "1.3.1" release was incomplete (there is for example no ./configure in 1.3.1).

You can probably ignore this pull request, but this seemed to be the only way to send in an issue as the "issues" button was missing on this github repository.

Would be awesome if you could fix this for 1.3.2 so that it becomes easier to install the C driver.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -92,7 +92,7 @@ http://api.mongodb.org/c/current/installing.html
 From a tarball
 --------------
 
-Download the latest release from `the release page <https://github.com/mongodb/mongo-c-driver/releases>`_, then::
+Download an older release (it is very important _not_ to get the latest release, because that one does not work as it does not have a ./configure script inside, and libbson is not inside it, so "make" doesn't work either) from `the release page <https://github.com/mongodb/mongo-c-driver/releases>`_, then::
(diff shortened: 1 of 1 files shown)


ajdavis:
Hi Emiel, the "configure" script and libbson are not in the mongo-c-driver repository. I generated them by running ./autogen.sh, which initializes the libbson submodule and generates ./configure. Then I package the result in the release tarball. The tarball has the configure script and libbson:

https://github.com/mongodb/mongo-c-driver/releases/download/1.3.1/mongo-c-driver-1.3.1.tar.gz

To report an issue with the driver, please open a ticket in Jira; instructions are here:

https://github.com/mongodb/mongo-c-driver#bugs--feature-requests

Thanks!

EmielBruijntjes:
Thanks for your answer. Solves the problem indeed. It turned out that I was using the "Source code (tar.gz)" link from github instead of the "mongo-c-driver-1.3.1.tar.gz" link.

PR #312, opened 2016-02-18
Add stricter dependency on libbson >= 1.3.1

oerdnj:
Hi, your package is missing required dependency on libbson (>= 1.3.1) and FTBFS if libbson 1.3.1 is not found on the system.


oerdnj:
Same as with libbson - added BC support for dbgsym packages.

ajdavis:
Thank you!

PR #313, opened 2016-02-19
Debian

oerdnj:
Same change as in libbson - just remove -dbg package and use new dbgsym automatic packages.


debian/changelog
```diff
@@ -1,3 +1,19 @@
+libmongoc (1.3.1-2~1.gbp00ef93) UNRELEASED; urgency=medium
```
oerdnj: And please make the next upload target unstable and not experimental.

ajdavis:
Thanks very much! We'll reupload later this week.

PR #314, opened 2016-02-29
allow to skip online tests

remicollet:
In Fedora infrastructure, builder don't have internet access.
So test suite fails, see https://kojipkgs.fedoraproject.org//work/tasks/7900/13177900/build.log

    FAIL
    Assert Failure: "No suitable servers found (`serverselectiontryonce` set): [Failed to resolve 'doesntexist'] [Failed to resolve 'example.com']" does not contain "[connection error calling ismaster on 'example.com:2']"

This small enhancement allow to define SKIP_INLINE env. var. before running to skip such test (for now there is only 1)

### DIFF
--- a/tests/test-libmongoc.c
+++ b/tests/test-libmongoc.c
@@ -1209,6 +1209,12 @@ int test_framework_skip_if_max_version_version_less_than_4 (void)
 }
 
 
+int test_framework_skip_if_offline (void)
+{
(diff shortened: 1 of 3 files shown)


ajdavis:
Cool, let's rename the environment variable MONGOC_TEST_OFFLINE and rename the function test_framework_skip_if_offline, since I think that better describes its purpose.

remicollet:
renamed.

ajdavis:
Thanks!

PR #315, opened 2016-03-29
CDRIVER-1096: Use const for mongoc_client_select_server() read prefs

jmikola:
Noticed while implementing [PHPC-607](https://jira.mongodb.org/browse/PHPC-607). This seemed like an oversight compared to the private `mongoc_topology_select()` function and other client functions that accept a read preference.

Related to: https://github.com/mongodb/mongo-c-driver/commit/929e8590a0ff49979ff83cc8175844d8aa992890


ajdavis:
Great. Remove installing.page (which is generated by ./autogen.sh, shouldn't be checked in) and then I'll merge.

jmikola:
Must have fat-fingered the `git add`. Fixed!

PR #317, opened 2016-03-31
BUILD-893 Upgrade OS X builders to OS X 10.10

bsamek:


bjori:
This branch is no longer maintained and should be removed from evg. I'll file a ticket to get it removed :)

PR #322, opened 2016-04-12
CDRIVER-1193, CDRIVER-1198: Expose mongoc_server_description_type_t and improve SDAM logging

jmikola:
* https://jira.mongodb.org/browse/CDRIVER-1193
 * https://jira.mongodb.org/browse/CDRIVER-1198

Cross-referencing with https://github.com/mongodb/mongo-php-driver/pull/291.

### DIFF
--- a/src/mongoc/mongoc-server-description-private.h
+++ b/src/mongoc/mongoc-server-description-private.h
@@ -28,19 +28,9 @@
 /* represent a server or topology with no replica set config version */
 #define MONGOC_NO_SET_VERSION -1
 
-typedef enum
-   {
-      MONGOC_SERVER_UNKNOWN,
-      MONGOC_SERVER_STANDALONE,
-      MONGOC_SERVER_MONGOS,
-      MONGOC_SERVER_POSSIBLE_PRIMARY,
-      MONGOC_SERVER_RS_PRIMARY,
-      MONGOC_SERVER_RS_SECONDARY,
-      MONGOC_SERVER_RS_ARBITER,
-      MONGOC_SERVER_RS_OTHER,
-      MONGOC_SERVER_RS_GHOST,
-      MONGOC_SERVER_DESCRIPTION_TYPES,
-   } mongoc_server_description_type_t;
+/* this is used for declaring gSDAMTransitionTable and should be defined as the
+ * number of unique states in mongoc_server_description_type_t */
+#define MONGOC_SERVER_DESCRIPTION_TYPES 9
 
 struct _mongoc_server_description_t
 {
@@ -104,6 +94,7 @@ mongoc_server_description_set_set_version (mongoc_server_description_t *descript
 void
 mongoc_server_description_set_election_id (mongoc_server_description_t *description,
                                            const bson_oid_t            *election_id);
+
 void
 mongoc_server_description_update_rtt (mongoc_server_description_t *server,
                                       int64_t                      new_time);
--- a/src/mongoc/mongoc-server-description-private.h
+++ b/src/mongoc/mongoc-server-description-private.h
@@ -95,6 +95,9 @@ void
 mongoc_server_description_set_election_id (mongoc_server_description_t *description,
                                            const bson_oid_t            *election_id);
 
+const char *
+mongoc_server_description_type_string (const mongoc_server_description_t *description);
+
 void
 mongoc_server_description_update_rtt (mongoc_server_description_t *server,
                                       int64_t                      new_time);
--- a/src/mongoc/mongoc-server-description.c
+++ b/src/mongoc/mongoc-server-description.c
@@ -299,38 +299,15 @@ mongoc_server_description_round_trip_time (mongoc_server_description_t *descript
  *      Discovery And Monitoring Spec.
  *
  * Returns:
- *      A string.
+ *      A mongoc_server_description_type_t, this server's type.
  *
  *--------------------------------------------------------------------------
  */
 
-const char *
-mongoc_server_description_type (mongoc_server_description_t *description)
+mongoc_server_description_type_t
+mongoc_server_description_type (const mongoc_server_description_t *description)
 {
-   switch (description->type) {
-   case MONGOC_SERVER_UNKNOWN:
-      return "Unknown";
-   case MONGOC_SERVER_STANDALONE:
-      return "Standalone";
-   case MONGOC_SERVER_MONGOS:
-      return "Mongos";
-   case MONGOC_SERVER_POSSIBLE_PRIMARY:
-      return "PossiblePrimary";
-   case MONGOC_SERVER_RS_PRIMARY:
-      return "RSPrimary";
-   case MONGOC_SERVER_RS_SECONDARY:
-      return "RSSecondary";
-   case MONGOC_SERVER_RS_ARBITER:
-      return "RSArbiter";
-   case MONGOC_SERVER_RS_OTHER:
-      return "RSOther";
-   case MONGOC_SERVER_RS_GHOST:
-      return "RSGhost";
-   case MONGOC_SERVER_DESCRIPTION_TYPES:
-   default:
-      MONGOC_ERROR ("Invalid mongoc_server_description_t type\n");
-      return "Invalid";
-   }
+   return description->type;
 }
 
 /*
--- a/src/mongoc/mongoc-server-description.c
+++ b/src/mongoc/mongoc-server-description.c
@@ -313,6 +313,48 @@ mongoc_server_description_type (const mongoc_server_description_t *description)
 /*
  *--------------------------------------------------------------------------
  *
+ * mongoc_server_description_type_string --
+ *
+ *      Get this server's type, one of the types defined in the Server
+ *      Discovery And Monitoring Spec, as a string.
+ *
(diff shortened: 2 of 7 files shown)


src/mongoc/mongoc-server-description.h
```diff
@@ -22,6 +22,19 @@
...
+{
+   MONGOC_SERVER_UNKNOWN,
+   MONGOC_SERVER_STANDALONE,
+   MONGOC_SERVER_MONGOS,
+   MONGOC_SERVER_POSSIBLE_PRIMARY,
+   MONGOC_SERVER_RS_PRIMARY,
+   MONGOC_SERVER_RS_SECONDARY,
+   MONGOC_SERVER_RS_ARBITER,
+   MONGOC_SERVER_RS_OTHER,
+   MONGOC_SERVER_RS_GHOST,
```
derickr: Because the MONGOC_SERVER_DESCRIPTION_TYPES definition is decoupled from the enum, you probably should add at least a comment here that that value needs to be updated if the enum is extended. I would probably rather have the MONGOC_SERVER_DESCRIPTION_TYPES here to prevent them from going out of sync.

src/mongoc/mongoc-server-description.c
```diff
@@ -279,14 +279,34 @@ mongoc_server_description_host (mongoc_server_description_t *description)
...
+ *      Discovery And Monitoring Spec, as a string.
+ *
+ * Returns:
+ *      A string denoting this server's type.
+ *
+ *--------------------------------------------------------------------------
+ */
+
+const char *
+mongoc_server_description_type_string (mongoc_server_description_t *description)
```
derickr: Right, we don't need this exposed.

src/mongoc/mongoc-topology-description.c
```diff
@@ -1370,6 +1370,46 @@ gSDAMTransitionTable[MONGOC_SERVER_DESCRIPTION_TYPES][MONGOC_TOPOLOGY_DESCRIPTIO
...
+ * Returns:
+ *       A string.
+ *
+ * Side effects:
+ *       None.
+ *
+ *--------------------------------------------------------------------------
+ */
+static const char *
+_mongoc_topology_description_type_string (mongoc_topology_description_t *topology)
```
derickr: I believe this was something that @ajdavis mentioned in the email thread, and that's why I added it. I also don't think we use it.

src/mongoc/mongoc-server-description.h
jmikola: I didn't want to define MONGOC_SERVER_DESCRIPTION_TYPES in the public API since it's only used internally; however, I've added a comment here mentioning that the constant should be updated if this enum is ever extended.

src/mongoc/mongoc-server-description.c
jmikola: I moved it to the private API.

src/mongoc/mongoc-server-description-private.h
```diff
@@ -28,19 +28,9 @@
...
-      MONGOC_SERVER_RS_PRIMARY,
-      MONGOC_SERVER_RS_SECONDARY,
-      MONGOC_SERVER_RS_ARBITER,
-      MONGOC_SERVER_RS_OTHER,
-      MONGOC_SERVER_RS_GHOST,
-      MONGOC_SERVER_DESCRIPTION_TYPES,
-   } mongoc_server_description_type_t;
+/* this is used for declaring gSDAMTransitionTable and should be defined as the
+ * number of unique states in mongoc_server_description_type_t */
+#define MONGOC_SERVER_DESCRIPTION_TYPES 9
```
bjori: It tells us how many elements we need to account for in the transition table and was a neat way of keeping track of it.

I think moving it to the .c file it is used in may be the best course of action

src/mongoc/mongoc-server-description-private.h
```diff
@@ -104,6 +94,10 @@ mongoc_server_description_set_set_version (mongoc_server_description_t *descript
 void
 mongoc_server_description_set_election_id (mongoc_server_description_t *description,
                                            const bson_oid_t            *election_id);
+
+const char *
+mongoc_server_description_type_string (const mongoc_server_description_t *description);
```
bjori: Errrrrrrr. Private functions should start with _, but I guess no other function here does that :/

src/mongoc/mongoc-server-description.c
```diff
@@ -299,13 +299,33 @@ mongoc_server_description_round_trip_time (mongoc_server_description_t *descript
...
+ *
+ * Returns:
+ *      A string denoting this server's type.
  *
  *--------------------------------------------------------------------------
  */
 
 const char *
-mongoc_server_description_type (mongoc_server_description_t *description)
+mongoc_server_description_type_string (const mongoc_server_description_t *description)
```
bjori: I'm not seeing this used anywhere?

src/mongoc/mongoc-topology-description.c
bjori: The only usecase for this function and mongoc_server_description_type_string () is the one TRACE() entry?
I don't think thats useful enough, and prefer keeping the trace lines as-is I guess

src/mongoc/mongoc-topology-description.c
```diff
@@ -1403,9 +1443,9 @@ mongoc_topology_description_handle_ismaster (
                                               error);
 
    if (gSDAMTransitionTable[sd->type][topology->type]) {
-      TRACE("Transitioning to %d for %d", topology->type, sd->type);
+      TRACE("Transitioning to %s for %s", _mongoc_topology_description_type_string(topology), mongoc_server_description_type_string(sd));
       gSDAMTransitionTable[sd->type][topology->type] (topology, sd);
    } else {
-      TRACE("No transition entry to %d for %d", topology->type, sd->type);
+      TRACE("No transition entry to %s for %s", _mongoc_topology_description_type_string(topology), mongoc_server_description_type_string(sd));
```
bjori: This line seems to be the only reason to introduce the _string () variants of these functions... I don't think thats good enough reason.
It'll cause useless extra work and headscratching in the future

chergert: Maybe also add some assertions before dereferencing that jump table.

jmikola: @bjori: This was the only use of the private `mongoc_server_description_type_string()` function. I created the separate ticket to track the enhancement of using string names here. Since these messages frequently appear in driver logs, it'd be very helpful if we used descriptive names instead of integers, which require counting through enum values to resolve.

> Maybe also add some assertions before dereferencing that jump table.

I assume that applies to the `if (gSDAMTransitionTable[sd->type][topology->type])` line?

src/mongoc/mongoc-topology-description.c
jmikola: This is only for [CDRIVER-1198](https://jira.mongodb.org/browse/CDRIVER-1198). As I mentioned [below](https://github.com/mongodb/mongo-c-driver/pull/322/files#r59925833), the current trace lines require hoop-jumping to decipher, and a single log often has a bunch of those state changes. Using strings would be a huge convenience for debugging.

src/mongoc/mongoc-server-description-private.h
jmikola: Existing private functions in this file didn't use an underscore prefix, so I didn't use it. I can change this in the second commit if you decide that [CDRIVER-1198](https://jira.mongodb.org/browse/CDRIVER-1198) is worthwhile.

src/mongoc/mongoc-server-description-private.h
jmikola: `mongoc-topology-description.c`? That seems a bit odd given that it has a `MONGOC_SERVER_DESCRIPTION` prefix. If moved there, would you want it at the top of the C file below the `#include` lines?

src/mongoc/mongoc-topology-description.c
chergert: @jmikola correct

src/mongoc/mongoc-server-description-private.h
bjori: It seems weird to define an enum in one place, the count of it in another place, and then its only used in the 3rd place.

Its unfortunate that this enum must be exposed, because it means we are knowingly breaking the ABI as soon as we add another entry to it, which will break your "enum to string" right off the bat.

I strongly recommend instead that we rather return the stringification of the types.
If you want to represent that as a number somewhere else, then thats fine. Any unknown strings can then be represented as a 0 in the app or something....

jmikola:
Closing this. [CDRIVER-1198](https://jira.mongodb.org/browse/CDRIVER-1198) was split off into https://github.com/mongodb/mongo-c-driver/pull/323 and we decided to maintain a server description type map manually in the PHP driver (see https://github.com/mongodb/mongo-php-driver/pull/291).

src/mongoc/mongoc-topology-description.c
jmikola: Opened [CDRIVER-1206](https://jira.mongodb.org/browse/CDRIVER-1206) to track that suggestion.

PR #323, opened 2016-04-18
CDRIVER-1198: Use strings for server and topology types in SDAM transition logs

jmikola:
https://jira.mongodb.org/browse/CDRIVER-1198

This PR is split off from https://github.com/mongodb/mongo-c-driver/pull/322 to remove publication of the server description's type enum.


bjori:
thanks!

PR #329, opened 2016-06-08
CDRIVER-1316: copy strings from mongoc_ssl_opt_t

puppyofkosh:


ajdavis:
Thanks!

src/mongoc/mongoc-ssl.h
```diff
@@ -32,14 +32,14 @@ typedef struct _mongoc_ssl_opt_t mongoc_ssl_opt_t;
...
 {
-   const char *pem_file;
-   const char *pem_pwd;
-   const char *ca_file;
-   const char *ca_dir;
-   const char *crl_file;
-   bool        weak_cert_validation;
-   bool        allow_invalid_hostname;
-   void       *padding [7];
+   char *pem_file;
```
acmorrow: @ajdavis @puppyofkosh @xdg - So... what happened to the const-ness here? It is causing warnings in the C++11 driver appveyor build:

https://ci.appveyor.com/project/markbenvenuto/mongo-cxx-driver/build/3.0.0.182#L2004

ajdavis: Since the driver now owns and frees these strings, it seemed right to un-const them?

acmorrow: The const here refers to the pointed to bytes. It means that if I pass in a pointer to a character constant, I don't have to worry about it. Right now, the following is a warning/error:
```
const char* ca_file = "/foo/bar/baz";
mongoc_ssl_opt_t options = {0};
options.ca_file = ca_file;
```
Because the ca_file field in the options structure isn't const.

puppyofkosh:
I put a fix for this into code review

PR #332, opened 2016-06-13
avoid abicheck failure

remicollet:
We encounter recent failure because of this script.

See: https://bugzilla.redhat.com/1345868
And : https://apps.fedoraproject.org/koschei/package/mongo-c-driver

Build is back to "green" thanks to this minor fix.

### DIFF
--- a/tests/abicheck.sh
+++ b/tests/abicheck.sh
@@ -1,6 +1,6 @@
 #! /bin/sh
 
-cpp -P ${cppargs} ${1:-./src/libmongoc.symbols} | sed -e '/^$/d' -e 's/ G_GNUC.*$//' -e 's/ PRIVATE//' -e 's/ DATA//' | sort > expected-abi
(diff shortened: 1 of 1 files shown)


remicollet:
My 1st idea was to use "diff -uwB", but using "cpp --freestanding" seems better (tested with old GCC 4.4 in RHEL-6 and newer GCC 6.1 in Fedora rawhide)

ajdavis:
How interesting, thanks. What does --freestanding do?

remicollet:
--freestanding is the same than -ffreestanding

Quoting Florian Weimer (glibc maintainer): 

> Use “cpp --freestanding -P”, so that you do not end up with any whitespace generated from implementation header files.

ajdavis:
I sort of understand, "A freestanding environment is one in which the standard library may not exist, and program startup may not necessarily be at main. The most obvious example is an OS kernel. This is equivalent to -fno-hosted."

https://gcc.gnu.org/onlinedocs/gcc/C-Dialect-Options.html

Thanks!

PR #338, opened 2016-06-17
added fiona and ian's ssl ping example

puppyofkosh:
This is what we did on our first day, and it looks like it'll actually get committed!


bjori:
This isn't actually generic enough I think to be included as an example.
It was a decent exercise in seeing how to use the driver though :)

PR #355, opened 2016-07-22
CDRIVER-1381: small optimization in reply_local

fionaRowan:
By using reply_ptr, removed need for bson_copy_to call in mongoc_collection_find_and_modify_with_opts


bjori:
thanks!

PR #358, opened 2016-07-25
CDRIVER-1322: mongoc_collection_aggregate_with_write_concern

fionaRowan:
Implemented and tested mongoc_collection_aggregate_with_write_concern, where mongoc_collection_aggregate is a thin wrapper that passes NULL as write concern.


ajdavis:
Good, but see https://jira.mongodb.org/browse/CDRIVER-1414.

Also, we need mongoc_collection_aggregate_with_write_concern.page.

PR #359, opened 2016-07-25
fixed handling of other cursor errors in mongoc_collection_find_indexes

cschwarz-inco:
previous behavior: function returned freed cursor if an error different from the two hardcoded error cases occurred. caller then ran into a "use-after-free" issue.

new behavior: functions returns a NULL pointer if an error different from the two hardcoded error cases occurred

### DIFF
--- a/src/mongoc/mongoc-collection.c
+++ b/src/mongoc/mongoc-collection.c
@@ -1209,6 +1209,9 @@ mongoc_collection_find_indexes (mongoc_collection_t *collection,
             error->code = 0;
             error->domain = 0;
             cursor = _mongoc_collection_find_indexes_legacy (collection, error);
+         } else {
+            /* other error, to be handled by caller */
+            cursor = NULL;
          }
       }
    }


ajdavis:
Good catch. Was there a specific case that reproduced the bug for you?

cschwarz-inco:
It was coincidence. Due to a programming error in our application, multiple threads used the same mongoc_client_t connection concurrently. This caused all kinds of funny errors.

ajdavis:
Did you mean to push another commit to this PR? It's not related to your first change.

cschwarz-inco:
No I didn't. The second commit was not supposed to go into the first pull request. (maybe a second pull request). I'll try to restore the original pull request...

cschwarz-inco:
Ok. I removed the second commit.

ajdavis:
Filed the bug:

https://jira.mongodb.org/browse/CDRIVER-1420

I've written a test, awaiting my colleagues' review.

ajdavis:
Merged in https://github.com/mongodb/mongo-c-driver/commit/542c42d535b17ec406c97a04a5ade4633ef81ceb.

PR #363, opened 2016-07-27
mongoc_bulk_operation_execute: ensure reply is initialized before returning

cschwarz-inco:
The documentation for "mongoc_bulk_operation_execute" says "reply is always initialized, even upon failure. Callers must call bson_destroy() to release this potential allocation."

However, implementation does not initialize reply in all cases, making it impossible for the caller to determine whether reply has been initialized or not. This is contrary to documentation, which says that reply is ALWAYS initialized.

This commit changes the implementation to be in accordance with the documentation. reply is initialized before any return.


ajdavis:
Thanks, we'll merge in the next few days.

ajdavis:
Tracking in https://jira.mongodb.org/browse/CDRIVER-1433

ajdavis:
Merged in https://github.com/mongodb/mongo-c-driver/commit/0e0ae00cffdccaa7e3436f40a39bad61eae1e3a4

PR #394, opened 2016-08-12
CDRIVER-1460: database_drop, parse wc err and return false

fionaRowan:
changed behavior for mongoc_database_drop command function on writeConcernError: parse writeConcernError and return false.


ajdavis:
Merged in https://github.com/mongodb/mongo-c-driver/commit/dde593cbe7a000999f5357dd7c3196b73e3794ae, thanks for everything!

PR #398, opened 2016-10-07
CDRIVER-1639 seeds rand for use in selecting servers

tomnomnom:
As per https://jira.mongodb.org/browse/CDRIVER-1639:

The call to rand() used when selecting a server is unseeded (https://github.com/mongodb/mongo-c-driver/blob/a4bcfffa9c7ffa2e541a3351ec26cfad417cbdcf/src/mongoc/mongoc-topology-description.c#L693)

This tends to manifest itself as uneven balancing between nodes in a cluster.

As a specific example: when libmongoc is used as part of, say, the php-driver:

1. A sudden increase in traffic creates new children (e.g. Apache workers)
2. The new workers create new connections to mongo using the php-driver, and therefore libmongoc
3. All workers initially choose the same server from the cluster as the first value from rand() is always the same
4. One mongo node takes the vast majority of the queries - possibly causing it to become overloaded while other nodes in the cluster receive little traffic

### DIFF
--- a/src/mongoc/mongoc-topology-description.c
+++ b/src/mongoc/mongoc-topology-description.c
@@ -56,6 +56,8 @@ mongoc_topology_description_init (mongoc_topology_description_t      *descriptio
                         type == MONGOC_TOPOLOGY_SINGLE ||
                         type == MONGOC_TOPOLOGY_RS_NO_PRIMARY);
 
+   srand(time(0));
+
    memset (description, 0, sizeof (*description));
 
    bson_oid_init (&description->topology_id, NULL);


tomnomnom:
Disclaimer: I'm not a C programmer. There might be a better place to put the `srand` call, or a better thing to seed it with. From a layman's perspective the topology init seemed like a logical place to put it, and the seed value is just what the legacy PHP driver was using.

acmorrow:
I'm not 100% sure, and I'm not one of the C driver devs, but I think it might be ill advised for a library to invoke srand, as that call has global side effects. An application might want or need explicit seeding, and it is not the place of any library to undermine that request.

stolenegg:
Fair point @acmorrow. Although in the context given (Apache process / PHP request), it's tricky to know where the seeding should happen as the 'application' code is not C. @tomnomnom How about moving the srand call up a layer to the php driver module init function?
https://github.com/mongodb/mongo-php-driver/blob/1fd77a1367552dab4b8985602e78d39a9fa8ba03/php_phongo.c#L2023

Possibly behind a config option to disable it if other modules conflict with this behaviour.

This would leave libmongoc agnostic (i.e. it always assumes that the calling code is responsible for seeding )

Disclaimer: I'm also not a C programmer! :-)

ajdavis:
@derickr and I think the right thing is for the C Driver to switch from rand to using rand_r (or rand_s on Windows), with a seed that's initially the system time. Then we have a properly seed PRNG that doesn't interfere with the global "rand" state.

tomnomnom:
@ajdavis that sounds like a nicer solution, but almost certainly outside of my comfort zone to implement myself. Is that something that's already in progress?

ajdavis:
Not yet in progress but we can complete it soon.

tomnomnom:
@ajdavis that's great to hear, thanks! I think I'll just build a patched version of the .so and use that until it's been fixed properly then.

tomnomnom:
This should no-longer be an issue as of 5f33f7d78.

Massive thanks to @ajdavis et al for that :)

ajdavis:
You're welcome! The C Driver 1.5.0 release that includes this fix is a month or two away, however. Do you need it sooner, or can you either work around the bug in your code, or live with the bug for now?

tomnomnom:
We could certainly do with it sooner than that. Our traffic can be pretty spiky, which is when the problem is most noticeable.

I think we'll probably look at temporarily forking the php driver and pointing libmongoc at master (with much testing before it hits prod of course :)). We're rolling our own RPMs anyway, so it shouldn't be too much hassle.

ajdavis:
Thanks, in that case we won't release this fix until 1.5.0, in order to minimize risk of changes for existing users of the driver.

PR #400, opened 2016-10-12
Upstream release 1.4.2 (upgrade from 1.4.1)

ajdavis:
@rcsanchez97 and @bjori please review.

I'm hoping that Debian release 1.4.2-1 has all the headers it needs, unlike Debian release 1.4.1-1:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=840528

Look at the `update-libmongoc-dev.install.sh` script I wrote and the changes the script made to `debian/libmongoc-dev.install`. I think that, since this is version 1.4.2 and it doesn't include Hannes's changes for [CDRIVER-989](https://jira.mongodb.org/browse/CDRIVER-989), it's correct for now to install all the private headers.

*OR*, perhaps this is dumb? Should I delete `update-libmongoc-dev.install.sh` and just use a file pattern in `libmongoc-dev.install` like this?:

```
usr/include/libmongoc-1.0/*.{h,def,defs}
```

Once we package 1.5.0 for Debian, we can make new decisions about which header files to include.


i-rinat:
There is a file exclusion (`-X-private.h`) in https://github.com/mongodb/mongo-c-driver/blob/debian/debian/rules#L42, so even `*-private.h` files are mentioned in `libmongoc-dev.install` files, they won't make it to the `libmongoc-dev` package.

ajdavis:
Good point @i-rinat, thanks very much. I think the right thing to do here is update `libmongoc-dev.install.sh` to include all files with this pattern:
```
usr/include/libmongoc-1.0/*.{h,def,defs}
```
... and delete the `-X-private.h` exclusion rule from `rules`.

Beginning with 1.5.0 and [CDRIVER-989](https://jira.mongodb.org/browse/CDRIVER-989), we can remove the private headers again by deleting the pattern from `libmongoc-dev.install.sh` and restoring the exclusion rule in `rules`.

ajdavis:
I've updated the Debian config now to simply include all headers by glob pattern. This builds correctly and the contents of libmongoc-dev built with this config are shown here:

https://gist.github.com/ajdavis/858f6b4f1789e226319bbda962c7c652

i-rinat:
I can't say anything about code changes, but changes to files in `debian/*` look great, thanks.

By the way, was exposing of `*-private.h` a right thing to do? From their name I can guess they are somehow private to the library itself. And the only public header including a private one was `mongoc-stream-tls.h`. The only thing that private header defines is `struct _mongoc_stream_tls_t`, which is then used to define type `mongoc_stream_tls_t`. But `typedef` from a `struct` is fine even without definition of structure itself. Moreover, `mongoc_stream_tls_t` is not used anywhere in public headers at all.

ajdavis:
We fixed the private headers issue, but not in 1.4.2:

https://jira.mongodb.org/browse/CDRIVER-989

i-rinat:
> https://jira.mongodb.org/browse/CDRIVER-989

Have had a look more carefully this time, and have found a patch with a change I was (kind of) talking about: https://github.com/mongodb/mongo-c-driver/commit/1e60b28923d9cca888d97c8e15a18a54bb49d609. That's great. Thank you for explaining!

PR #401, opened 2016-10-14
CDRIVER-1702 don't fail when returned platform is truncated

remicollet:
platform can be truncated in _append_platform_field (according to doc length, HANDSHAKE_MAX_SIZE...)


bjori:
Thanks!

PR #403, opened 2016-11-04
no braces in debian install script

ajdavis:
Resolves lintian warning brace-expansion-in-debhelper-config-file, "This
debhelper config file appears to use shell brace expansion. This happens
to work due to an accident of implementation but is not a supported
feature."


ajdavis:
@rcsanchez97 could you take a look?

rcsanchez97:
@ajdavis That fix looks exactly right.

rcsanchez97:
@ajdavis That fix looks exactly right.

PR #404, opened 2016-11-07
CDRIVER-1907 update libmongoc-dev dependencies

ajdavis:
@rcsanchez97 could you take a look? This is a proposed fix for https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=843495

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -18,6 +18,8 @@ Package: libmongoc-dev
 Section: libdevel
 Architecture: any
 Depends: libmongoc-1.0-0 (= ${binary:Version}),
+         libbson-dev (>= ${binary:Version}),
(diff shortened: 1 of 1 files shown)


rcsanchez97:
@ajdavis sorry I lost track of this.  It looks good, but you will need
to add a comma after libssl-dev or the package build will fail because
of the control file not being properly formatted. Also, the associated
debian/changelog message should include "(Closes: #843495)" after the
text of the entry so that the bug is closed automatically upon upload.

ajdavis:
Merged in https://github.com/mongodb/mongo-c-driver/commit/8cbd100cf02faa5d5536624b5d17cb3cb49db73d

PR #405, opened 2016-11-17
CDRIVER-1921 cursor from static reply

xdg:
Prior to this commit, mongoc_cursor_new_from_command_reply fails if the
reply is read-only.  This commit makes a copy of static bson_t instead of
failing.


bjori:
Merged as https://github.com/mongodb/mongo-c-driver/commit/111ad1a8fb39fa996c707fbf7f93b30f4b8c988e

PR #406, opened 2016-11-30
Update README links to mongoc host

christopherjwang:
I should probably bookmark the c docs, but I always come to the README to get to them. 

`http://api.mongodb.org/c/current/` redirects to `http://mongoc.org/` so I updated that as well.


ajdavis:
Thanks Christopher!

PR #408, opened 2016-11-30
fix UBSAN problem

jeroenooms:
If `items_len == 0` then `malloc(items_len) == NULL` which results in `memcpy(NULL, ...)` which is undefined behaviour.


bjori:
I think I've seen this before, and thought we fixed it.. need to give it a further look

PR #409, opened 2016-12-01
hp-ux port

malexzx:
Hello!

Following patch allows  the mongo-c driver to work on HP-UX 11.31 ia64 platform.
Note: to work, with this patch, we need a patch to libbson, that PR-ed just now (https://github.com/mongodb/libbson/pull/179). 

To compile on HPUX we need: 

- gcc-4.8.5
- pass -D_REENTRANT -D_INCLUDE_HPUX_SOURCE to CFLAGS
- pass -mlp64 to CFLAGS and LDFLAGS for 64bit version

configure opts: --enable-shm-counters=no -sinse it not ported yet

### DIFF
--- a/src/mongoc/mongoc-counters-private.h
+++ b/src/mongoc/mongoc-counters-private.h
@@ -35,6 +35,9 @@
 # include <sys/param.h>
 #endif
 
+#ifdef __hpux__
+#include <sys/pstat.h>
+#endif
 
 BSON_BEGIN_DECLS
 
@@ -48,6 +51,13 @@ _mongoc_get_cpu_count (void)
 {
 #if defined(__linux__)
    return get_nprocs ();
+#elif defined(__hpux__)
+  struct pst_dynamic psd;
+  if (pstat_getdynamic (&psd, sizeof (psd), (size_t) 1, 0) != -1) {
+     return psd.psd_max_proc_cnt;
+  }
+
+  return 1;
 #elif defined(__FreeBSD__) || \
       defined(__NetBSD__) || \
       defined(__DragonFly__) || \
--- a/src/mongoc/mongoc-cursor-private.h
+++ b/src/mongoc/mongoc-cursor-private.h
@@ -46,13 +46,7 @@ struct _mongoc_cursor_interface_t
    void             (*get_host) (mongoc_cursor_t        *cursor,
                                  mongoc_host_list_t     *host);
 };
-#ifdef MIN
-#undef MIN
-#endif
 
-#ifdef MAX
-#undef MAX
-#endif
 #define ALLOW_PARTIAL_RESULTS "allowPartialResults"
 #define ALLOW_PARTIAL_RESULTS_LEN 19
 #define AWAIT_DATA "awaitData"
@@ -73,16 +67,16 @@ struct _mongoc_cursor_interface_t
 #define HINT_LEN 4
 #define LIMIT "limit"
 #define LIMIT_LEN 5
-#define MAX "max"
-#define MAX_LEN 3
+#define MAX_STRING "max"
+#define MAX_STRING_LEN 3
 #define MAX_AWAIT_TIME_MS "maxAwaitTimeMS"
 #define MAX_AWAIT_TIME_MS_LEN 14
 #define MAX_SCAN "maxScan"
 #define MAX_SCAN_LEN 7
 #define MAX_TIME_MS "maxTimeMS"
 #define MAX_TIME_MS_LEN 9
-#define MIN "min"
-#define MIN_LEN 3
+#define MIN_STRING "min"
+#define MIN_STRING_LEN 3
 #define NO_CURSOR_TIMEOUT "noCursorTimeout"
 #define NO_CURSOR_TIMEOUT_LEN 15
 #define OPLOG_REPLAY "oplogReplay"
--- a/src/mongoc/mongoc-cursor-private.h
+++ b/src/mongoc/mongoc-cursor-private.h
@@ -46,7 +46,13 @@ struct _mongoc_cursor_interface_t
    void             (*get_host) (mongoc_cursor_t        *cursor,
                                  mongoc_host_list_t     *host);
 };
+#ifdef MIN
+#undef MIN
+#endif
 
+#ifdef MAX
+#undef MAX
+#endif
 #define ALLOW_PARTIAL_RESULTS "allowPartialResults"
 #define ALLOW_PARTIAL_RESULTS_LEN 19
 #define AWAIT_DATA "awaitData"
--- a/src/mongoc/mongoc-socket.c
+++ b/src/mongoc/mongoc-socket.c
@@ -333,7 +333,7 @@ _mongoc_socket_capture_errno (mongoc_socket_t *sock) /* IN */
 #else
    sock->errno_ = errno;
 #endif
-   TRACE("setting errno: %d", sock->errno_);
+   TRACE("setting errno: %d %s", sock->errno_, strerror(sock->errno_));
 }
 
 
@@ -412,7 +412,7 @@ mongoc_socket_accept_ex (mongoc_socket_t *sock,      /* IN */
 {
    mongoc_socket_t *client;
    struct sockaddr_in addr = { 0 };
-   socklen_t addrlen = sizeof addr;
+   mongoc_socklen_t addrlen = sizeof addr;
    bool try_again = false;
    bool failed = false;
 #ifdef _WIN32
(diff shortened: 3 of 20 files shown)


src/mongoc/mongoc-cursor-private.h
```diff
@@ -46,7 +46,13 @@ struct _mongoc_cursor_interface_t
    void             (*get_host) (mongoc_cursor_t        *cursor,
                                  mongoc_host_list_t     *host);
 };
+#ifdef MIN
+#undef MIN
+#endif
 
+#ifdef MAX
+#undef MAX
+#endif
```
bjori: I feel this is a bit unfortunate. we should probably rename the macros below instead

src/mongoc/mongoc-socket.h
```diff
@@ -38,6 +38,12 @@
 # include <sys/un.h>
 #endif
 
+#ifdef __hpux__
+typedef int       mongoc_socklen_t;
```
bjori: we should probably move this to autoconf check, I can image more unixes not having `socklen_t`

src/mongoc/mongoc-socket.c
```diff
@@ -487,7 +487,7 @@ mongoc_socket_accept_ex (mongoc_socket_t *sock,      /* IN */
 int
 mongoc_socket_bind (mongoc_socket_t       *sock,    /* IN */
                     const struct sockaddr *addr,    /* IN */
-                    socklen_t              addrlen) /* IN */
+                    mongoc_socklen_t              addrlen) /* IN */
```
bjori: here, and everywhere else, the argument should be aligned in a pretty ascii art with the argument above

src/mongoc/mongoc-counters-private.h
```diff
@@ -48,6 +51,12 @@ _mongoc_get_cpu_count (void)
 {
 #if defined(__linux__)
    return get_nprocs ();
+#elif defined(__hpux__)
+  struct pst_dynamic psd;
+  if (pstat_getdynamic (&psd, sizeof (psd), (size_t) 1, 0) != -1)
```
bjori: our coding standard requires braces, even for onelines

src/mongoc/mongoc-socket.c
```diff
@@ -333,7 +333,7 @@ _mongoc_socket_capture_errno (mongoc_socket_t *sock) /* IN */
 #else
    sock->errno_ = errno;
 #endif
-   TRACE("setting errno: %d", sock->errno_);
+   TRACE("setting errno: %d %s", sock->errno_, strerror(sock->errno_));
```
bjori: makes sense :D

src/mongoc/mongoc-cursor-private.h
malexzx: Hello @bjori 
I can update this PR.

I will try to rename, but it comes from HP system header, and probably no luck with it.

src/mongoc/mongoc-counters-private.h
malexzx: ok, undestand

src/mongoc/mongoc-cursor-private.h
malexzx: The code below this macro - define that macro as 
#define MAX "max" - so gcc warn about macro redefinition.

bjori: sorry, yeah. I mean, we should rename our MAX macro there to something else.

MAX and MIN are quite common macros which we shouldn't be imposing on.
we should probably rename our macros to MAX_STRING, and MAX_STRING_LEN maybe?

src/mongoc/mongoc-socket.h
bjori: will do ! leave it as is then, its fine :)

src/mongoc/mongoc-cursor-private.h
malexzx: yes, that should be reasonable

malexzx:
now amend this commit with update allignment of variables and add braces to conform coding standart.
if I have missed something, please tell me about that.

malexzx:
note, this PR requires libbson PR

src/mongoc/mongoc-cursor-private.h
malexzx: done

bjori:
Sorry to leave you in the dark on this.
We are working on getting consistent access to HPUX so we can ensure it will continue to be supported platform.

malexzx:
HPUX 11.31 (11i v3)
Integrity (aka ia64) is the latest hw platform supported by HPUX.
Sure, it supported https://h20392.www2.hpe.com/portal/swdepot/displayProductsList.do?category=OE

bjori:
I've created https://jira.mongodb.org/browse/CDRIVER-1973 so we can keep track of this patch when we prepare for our next release.
I find it unlikely we would include this in 1.5.2, but we plan on shipping 1.6.0 fairly soon

malexzx:
cool news! I see some conflicts in this branch. But now I'm at vacation now (to 12 January) , so not possible to help to resolve it. But as most of it - is a indentation and type replacement. That is not hard to change.
Anyway, very glad to that fact, so this patch is ready for including in release!

bjori:
Don't worry about the conflict, they are caused by our coding standard consistency commit in `master`. I'll fix it up when I merge this

bjori:
Merged as
- https://github.com/mongodb/mongo-c-driver/commit/98345054e1393385d406f0c0ce16275e4a83fd78
- https://github.com/mongodb/mongo-c-driver/commit/4ec9a155b4a9beb8e8fda02102c1c29df53825f0
- https://github.com/mongodb/mongo-c-driver/commit/fb80a4931167916c825d23f6a3b1b95b81ec86a9

Thanks for the patch!

malexzx:
Hello @bjori !
Thank you for support!

Everithing works, tests pass, but small fixes need:

1) The method for determine of using socklen_t does not work. socklen_t present on HPUX. It needs different socket implementation. I suggest to check use int by test macro _XOPEN_SOURCE_EXTENDED not set. For example, the ACE library check it with this code:
```
#    if defined (_XOPEN_SOURCE_EXTENDED)
typedef socklen_t ACE_SOCKET_LEN;
#    else
typedef int ACE_SOCKET_LEN;
#    endif /* _XOPEN_SOURCE_EXTENDED */
#  else
typedef socklen_t ACE_SOCKET_LEN;
#  endif /* __hpux */
#elif defined (ACE_HAS_SIZET_SOCKET_LEN)
typedef size_t ACE_SOCKET_LEN;
#else
typedef int ACE_SOCKET_LEN;
#endif /* ACE_HAS_SIZET_SOCKET_LEN */

``` 
(see http://nixdoc.net/man-pages/HP-UX/man2/connect.2.html for more information)
Lack of this macro will select native BSD socket implementation (instead of emulating GNU style sockets, which has a bugs). I always use BSD variant.
2) getting versions from VERSION_* files with HPUX utils fails, if those files not contain \n at end (I will investigate to avoid this).

bjori:
The above code will blindly ignore socklen_t and always pass `int` as the 3rd argument to `connect`, rather then `size_t` in the weird case where HPUX actually expects `size_t`.
If the type is however available, we use the native type, which will match the type expected by `connect`. I don't understand what you mean that how we typedef `mongoc_socklen_t` will affect which socket implementation will be used by HPUX?

I don't really think this matters much as the value of addrlen is never going to be so large for this to matter. Matching the strict prototype though would avoid a possible compiler warning.


As for the VERSION files -- I'm very curious, I couldn't see any such issues when I was building on HPUX the other day. Could you file a ticket for the issue at https://jira.mongodb.org/browse/CDRIVER ?

malexzx:
HPUX have several implementation of sockets. With case of connect() call - conversion things happens
, but passing pointers to receive (like in getsockname(... , int *)) will produce issue (on big endian platform). (I file the bug https://jira.mongodb.org/browse/CDRIVER-2015)
You can try simple test to view issue:
test.c

```
#include <stdio.h>
#include <sys/socket.h>

void test(int *i, socklen_t *s){
  printf("i=%d s=%d\n", *i, *s);
}

int main(int ac, char *av[]){
 socklen_t a = 1;
 unsigned int i = 10;


 test((int *)&i, (socklen_t*)&i);
 test((int *)&a, (socklen_t*)&a);

 return 0;
}

```
```
Output:
BIG_ENDIAN: only two combinations work
i=10 s=314336
i=0 s=1

LITTLE_ENDIAN: portable in any direction
i=10 s=10
i=1 s=1

```
More detail in https://curl.haxx.se/mail/lib-2009-04/0287.html

PR #410, opened 2016-12-02
Updated Debian packaging for 1.5.0 (without actual 1.5.0 upstream merge)

oerdnj:


debian/patches/0001-Compile-examples-with-lpthread.patch
```diff
@@ -0,0 +1,20 @@
...
+
+diff --git a/examples/Makefile.am b/examples/Makefile.am
+index da993e4..93e4c5b 100644
+--- a/examples/Makefile.am
++++ b/examples/Makefile.am
+@@ -1,6 +1,6 @@
+ EXAMPLE_LDADD = libmongoc-1.0.la
+ if EXPLICIT_LIBS
+-EXAMPLE_LDADD += $(BSON_LIBS)
++EXAMPLE_LDADD += $(BSON_LIBS) -lpthread
```
bjori: Is this a debian specific requirement?
I'd rather conditionally add it to the buildflags during configure then to maintain branches for something we should be doing anyway

oerdnj: I have no idea, but the build was failing. Anyway I saw Debian specific flags in d/rules before, but they were removed.

ajdavis:
Thanks, I've merged most of your changes!

We're not going to build with `--disable-automatic-init-and-cleanup`, that's a breaking change for users who *don't* call `mongoc_init` in their code but expect it to be called automatically. We'll make that change in libmongoc 2.0.

I also left the symbols-file out. We already have good tools that are independent of Debian and that ensure we maintain ABI stability, so we're discussing whether to maintain a symbols file for Debian or not.

PR #416, opened 2016-12-27
Support for ipv6-only hostnames

alexey-detr:
For cases when connection endpoint specified as hostname with ipv6
address only.

We have got this problem when tried to use PHP mongodb driver https://github.com/mongodb/mongo-php-driver with PHP 7. Our mongodb servers have ipv6 addresses only, so we can't rely on AF_INET.

### DIFF
--- a/src/mongoc/mongoc-uri.c
+++ b/src/mongoc/mongoc-uri.c
@@ -104,7 +104,7 @@ mongoc_uri_append_host (mongoc_uri_t *uri, const char *host, uint16_t port)
                      "%s:%hu",
                      host,
                      port);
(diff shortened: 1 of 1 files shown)


bjori:
Interesting. Nice catch.

I've created https://jira.mongodb.org/browse/CDRIVER-1972 to track this issue in hope to address it before our next release

bjori:
Thanks for the patch.

This has been merged as
- https://github.com/mongodb/mongo-c-driver/commit/333cbc2cd2f54f3650f51c39a2490c28c355cc0f (r1.5)
- https://github.com/mongodb/mongo-c-driver/commit/8729c1448782481f392e4b51e513c14bb9736a5b (master)

PR #417, opened 2017-01-02
CDRIVER-1522 CDRIVER-1970 update define constants for "find" opts to …

ooglek:
Improving on CDRIVER-1522 with a move to defining constants, I found that when compiled on FreeBSD, helper macros for MIN and MAX in sys/params.h, only included on FreeBSD, NetBSD, DragonFly and OpenBSD, clobbered the defines in mongoc-cursor-private.h. 

This improves the uniqueness in the define values to avoid future namespace collisions.

Compiled, tested and passed on a FreeBSD and OSX machine. Have not tested linux nor windows.


bjori:
Thanks!

PR #419, opened 2017-01-11
fix test for ipv6

remicollet:
Related to https://jira.mongodb.org/browse/CDRIVER-1988 (but partial)
Related to change in 8729c1448782481f392e4b51e513c14bb9736a5b


remicollet:
Notice: I haven kept AF_INET (which seems unneeded), but I supose you have run test sutie before release... so needed in some cases...

remicollet:
Closing as this is not anought to fix issue. Current IPv6 support breaks lot of things.

PR #420, opened 2017-02-06
Skip over documents that are too large to upload to MongoDB

aseering:
I discovered, when working with some code written on top of mongo-cxx, that, if I attempted to upload a document that was larger than 16mb, the driver would go into an infinite loop and hang.

The infinite loop occurs in the function that I've edited with this patch, in the underlying mongo-c driver.  The driver detects that the document is too long.  It looks to me like the intent of the driver is to flag an error for that document and continue to the next.  It does flag the error; it does not continue to the next.  So it just spins forever, repeatedly failing to send the too-large document.

I imagine this wants a unit test.  I have a local test, but it uses mongo-cxx and bson-cxx because I'm not very familiar with the C API, so it would be hard to include it in your test suite...  I figured I ought to at least post this code, though.

### DIFF
--- a/src/mongoc/mongoc-write-command.c
+++ b/src/mongoc/mongoc-write-command.c
@@ -1375,6 +1375,11 @@ again:
       too_large_error (error, i, len, max_bson_obj_size, NULL);
       result->failed = true;
       ret = false;
+
+      /* The current document is too large for MongoDB.  Continue to the next. */
+      if (!bson_iter_next (&iter)) {
+         GOTO (cleanup);
+      }
    } else {
       ret = mongoc_cluster_run_command_monitored (&client->cluster,
                                                   server_stream,
@@ -1405,6 +1410,7 @@ again:
       GOTO (again);
    }
 
+cleanup:
    bson_destroy (&cmd);
    EXIT;
 }


ajdavis:
Thanks very much! We'll plan to merge and release for 1.7.0:

https://jira.mongodb.org/browse/CDRIVER-589

Can you share your test **logic** with us to help write a unittest in C?

aseering:
Sure -- so, at a high level, my test does something like this:

* Construct a BSON document that looks something like `{"key":"zzzzzzzzzzzzzz...zzzzzzzzzzzzzzz"}`, where the value contains 20,000,000 `z`s
* Attempt to insert the document into the database.

I'm inserting using mongocxx's `collection::update_one()` with "upsert" enabled.  I don't have the stack trace in front of me, but I believe that mapped to `mongoc_bulk_operation_insert()` in the C API.

Just from reading the code, it looks like `command->flags.ordered` must be set to `false` for the mongoc write command, however that command is generated.  Otherwise I would expect the existing code to correctly exit on the first error.

Let me know if that's enough detail.  If not, I can go digging.

bjori:
This was merged as https://github.com/mongodb/mongo-c-driver/commit/d38df5714f898810d83a7f7a704530dfd9a90a9e

PR #421, opened 2017-02-10
CDRIVER-2042 skip test_create_index_fail when offline

remicollet:
From test_create_index_fail, requires internet connectivity (DNS):

       client = mongoc_client_new ("mongodb://example.com/?connectTimeoutMS=10");

trivial fix


ajdavis:
Thanks!

PR #425, opened 2017-03-03
CDRIVER-2072: Always initialize cursor filter and opts

jmikola:
https://jira.mongodb.org/browse/CDRIVER-2072

`mongoc_cursor_destroy()` always attempts to destroy these documents, which would previously crash if they were left uninitialized when an error was reported by `_mongoc_cursor_new_with_opts()`.

### DIFF
--- a/src/mongoc/mongoc-cursor.c
+++ b/src/mongoc/mongoc-cursor.c
@@ -229,6 +229,9 @@ _mongoc_cursor_new_with_opts (mongoc_client_t *client,
    cursor->client = client;
    cursor->is_command = is_command ? 1 : 0;
 
+   bson_init (&cursor->filter);
+   bson_init (&cursor->opts);
+
    if (filter) {
       if (!bson_validate (filter, BSON_VALIDATE_EMPTY_KEYS, NULL)) {
          MARK_FAILED (cursor);
@@ -239,9 +242,8 @@ _mongoc_cursor_new_with_opts (mongoc_client_t *client,
          GOTO (finish);
       }
 
+      bson_destroy (&cursor->filter);
       bson_copy_to (filter, &cursor->filter);
-   } else {
-      bson_init (&cursor->filter);
    }
 
    if (opts) {
@@ -263,7 +265,6 @@ _mongoc_cursor_new_with_opts (mongoc_client_t *client,
          GOTO (finish);
       }
 
-      bson_init (&cursor->opts);
       bson_copy_to_excluding_noinit (opts, &cursor->opts, "serverId", NULL);
 
       /* true if there's a valid serverId or no serverId, false on err */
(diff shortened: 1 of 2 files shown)


src/mongoc/mongoc-cursor.c
```diff
@@ -229,6 +229,9 @@ _mongoc_cursor_new_with_opts (mongoc_client_t *client,
    cursor->client = client;
    cursor->is_command = is_command ? 1 : 0;
 
+   bson_init (&cursor->filter);
+   bson_init (&cursor->opts);
```
jmikola: This covers both code paths for basic cursor creation, since `_mongoc_cursor_new()` already always initialized these documents. This might also avoid possible crash in `_mongoc_cursor_cursorid_init()`, as that always attempts to destroy `cursor->filter` before copying the command to it. The only code path I spotted for that private function is through `mongoc_cursor_new_from_command_reply()`, which itself calls `_mongoc_cursor_new_with_opts()`, so I believe we're safe.

_mongoc_cursor_cursorid_init

src/mongoc/mongoc-cursor.c
```diff
@@ -239,9 +242,8 @@ _mongoc_cursor_new_with_opts (mongoc_client_t *client,
          GOTO (finish);
       }
 
+      bson_destroy (&cursor->filter);
```
jmikola: This is the same approach I saw used elsewhere when SDAM regularly updates the last `isMaster` response. Unlike `opts`, which was previously initialized for all but the error case (a straightforward patch IMO), with `filter` we're potentially initializing an inline, static document and overwriting it if the user passed a filter. This is a bit redundant, but I felt the approach was much cleaner than a conditional `bson_init()` if `filter` failed validation.

tests/test-mongoc-cursor.c
```diff
@@ -733,6 +733,80 @@ test_cursor_new_invalid (void)
...
+{
+   mock_server_t *server;
+   mongoc_client_t *client;
+   mongoc_collection_t *collection;
+   mongoc_cursor_t *cursor;
+   bson_error_t error;
+
+   server = mock_server_new ();
+   mock_server_run (server);
+   client = mongoc_client_new_from_uri (mock_server_get_uri (server));
```
ajdavis: Don't need a mock server and all that ceremony - you can just do client_new and then find_with_opts.

(For future reference: mock server is useful to simulating server errors that are hard to trigger with a real server, or for reading messages sent *from* the driver and validating their contents. If you *do* need a mock server for such a scenario later on, don't forget to mock_server_destroy!)

jmikola: Fixed.

ajdavis:
Thanks!

PR #426, opened 2017-03-07
CDRIVER-2078 use bundled libson documentation instead of online copy

remicollet:


remicollet:
Fedora build with this patch applied
https://kojipkgs.fedoraproject.org//work/tasks/472/18250472/build.log

```
+ make -j48 man V=1
mkdir -p doc/man
/usr/libexec/python2-sphinx/sphinx-build -qEW -b man ./doc doc/man
+ exit 0

```

ajdavis:
Thanks Remi! I don't think this is the right solution, though - I'll explain in https://jira.mongodb.org/browse/CDRIVER-2078

PR #427, opened 2017-03-15
CDRIVER-2080 detect socket API parameters for legacy platforms

malexzx:
add detection code to CMakeLists.txt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -177,61 +177,64 @@ if (MINGW)
    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D_CRT_RAND_S")
 endif ()
 
+
+function(mongoc_get_accept_args ARG2 ARG3)
+   SET(VAR 0)
+   foreach (ARG2_VAL "struct sockaddr" "void")
+   foreach (ARG3_VAL "socklen_t" "size_t" "int")
+
+   MATH(EXPR VAR "${VAR}+1")
+
+   FILE(WRITE ${CMAKE_CURRENT_BINARY_DIR}/accept_test${VAR}.cxx
+   "#include <sys/types.h>
+    #include <sys/socket.h>
+
+    int main ()
+    {
+      int a = 0;
+      ${ARG2_VAL} *b = 0;
+      ${ARG3_VAL} *c = 0;
+      accept (a, b, c);
+      return 0;
+   }
+  ")
+
+  TRY_COMPILE(RES ${CMAKE_CURRENT_BINARY_DIR}
+  ${CMAKE_CURRENT_BINARY_DIR}/accept_test${VAR}.cxx CMAKE_FLAGS
+  "-Werror -DCMAKE_CXX_LINK_EXECUTABLE='echo not linking now...'" OUTPUT_VARIABLE LOG2)
+
+  if (RES)
+     message(STATUS "Detected accept args is ${ARG2_VAL} and ${ARG3_VAL}")
+     set(${ARG2} ${ARG2_VAL} PARENT_SCOPE)
+     set(${ARG3} ${ARG3_VAL} PARENT_SCOPE)
+     return ()
+  endif ()
+
+  endforeach()
+  endforeach()
+
+endfunction()
+
+
 # Reasonable default. CheckCompiler.m4 checks thoroughly for HP-UX's sake.
 set (MONGOC_SOCKET_ARG2 "struct sockaddr")
 set (MONGOC_SOCKET_ARG3 "socklen_t")
 
+
 if (NOT WIN32)
 
-set(MONGOC_ACCEPT_ARG2_LIST "struct sockaddr" "void")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -179,24 +179,70 @@ endif ()
...
+   set(MONGOC_SOCKET_ARG3 ${MONGOC_ACCEPT_ARG3_VAL})
+   break()
+   break()
+  endif ()
+endforeach()
+endforeach()
 
 include(CheckTypeSize)
-if (MSVC)
-   SET(CMAKE_EXTRA_INCLUDE_FILES "ws2tcpip.h")
```
ajdavis: Have you confirmed this change for the Windows build is correct?

CMakeLists.txt
```diff
@@ -179,24 +179,70 @@ endif ()
...
+   ${MONGOC_ACCEPT_ARG2_VAL} *b = 0;
+   ${MONGOC_ACCEPT_ARG3_VAL} *c = 0;
+   accept (a, b, c);
+   return 0;
+}
+")
+  TRY_COMPILE(RES ${CMAKE_CURRENT_BINARY_DIR}
+  ${CMAKE_CURRENT_SOURCE_DIR}/accept_test.cxx CMAKE_FLAGS
+  "-DCMAKE_CXX_LINK_EXECUTABLE='echo not linking now...'" OUTPUT_VARIABLE
+  LOG2)
```
ajdavis: For the Autotools version it was important to pass "-Werror" to the compiler when testing each combination of parameter types, in order to ensure that we had found the *exact* right types, rather than a combination of types that the compiler would accept with a warning:

https://github.com/mongodb/mongo-c-driver/blob/master/build/autotools/m4/ax_prototype.m4#L220

Do you need to pass "-Werror" to clang and gcc here, and perhaps "/WX" if (MSVC)?

CMakeLists.txt
```diff
@@ -179,24 +179,70 @@ endif ()
...
+set(MONGOC_ACCEPT_ARG2_LIST "struct sockaddr" "void")
+set(MONGOC_ACCEPT_ARG3_LIST "socklen_t" "size_t" "int")
+
+list(LENGTH MONGOC_ACCEPT_ARG2_LIST MONGOC_ACCEPT_ARG2_LIST_LEN)
+math(EXPR MONGOC_ACCEPT_ARG2_LIST_LEN_R "${MONGOC_ACCEPT_ARG2_LIST_LEN} - 1")
+list(LENGTH MONGOC_ACCEPT_ARG3_LIST MONGOC_ACCEPT_ARG3_LIST_LEN)
+math(EXPR MONGOC_ACCEPT_ARG3_LIST_LEN_R "${MONGOC_ACCEPT_ARG3_LIST_LEN} - 1")
+
+foreach(MONGOC_ACCEPT_ARG2_L RANGE ${MONGOC_ACCEPT_ARG2_LIST_LEN_R})
+foreach(MONGOC_ACCEPT_ARG3_L RANGE ${MONGOC_ACCEPT_ARG3_LIST_LEN_R})
```
ajdavis: why not just?:

```
foreach (MONGOC_ACCEPT_ARG2_VAL in "struct sockaddr" "void")
foreach (MONGOC_ACCEPT_ARG3_VAL in "socklen_t" "size_t" "int")
```
Why do we need an indexed loop instead of looping over the values directly?

ajdavis:
This looks very clever, thanks for the contribution.

Quick question, how are you installing CMake on HP-UX? I tried from a 3.7.2 source download and got:
```
[ 36%] Building C object Utilities/cmliblzma/CMakeFiles/cmliblzma.dir/liblzma/common/index.c.o
(Bundled) cc: warning 922: "-Aa" is unsupported in the bundled compiler, ignored.
(Bundled) cc: warning 922: "-Ae" is unsupported in the bundled compiler, ignored.
"/tmp/cmake-3.7.2/Utilities/cmliblzma/liblzma/common/index.c", line 102: error #2070:
          incomplete type is not allowed
        index_record records[];
                     ^

1 error detected in the compilation of "/tmp/cmake-3.7.2/Utilities/cmliblzma/liblzma/common/index.c".
```

CMakeLists.txt
malexzx: ok, that would be more right.

CMakeLists.txt
malexzx: that check not need for Windows.
Only socklen_t type. But I will try to check.

CMakeLists.txt
malexzx: yes, that correct.

malexzx:
about cmake. I ask that question to our UX team and do provide them.
But I'm think, so this occurs because camke build need at least aCC compiller (not bundled by default).
Or may be CXX and LD may be corrected.

malexzx:
I've improved checking of arguments now

ajdavis:
Merged, thank you:

https://github.com/mongodb/mongo-c-driver/commit/9feb9772409771d8116616ae9e65e79b46d8eed8

PR #428, opened 2017-03-20
Family specify branch

Lag1101:
Specifying aliases for ipv6 hosts aren't parsed correctly now, and it raises error in getaddrinfo func.
With option "family" in uri as ipv4,ipv6,unix,unspec user can define it directly.


ajdavis:
Thanks for the contribution @Lag1101, it's an interesting idea. Unfortunately we can't accept it, since it will sharply diverge from other MongoDB drivers' standard behavior. All drivers in all programming languages conform to the Connection String Spec:

https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst

Right now you can connect to an IPv4 or IPv6 address by specifying it explicitly:

```
mongodb://127.0.0.1/
mongodb://[::1]/
```

Once I implement https://jira.mongodb.org/browse/CDRIVER-1972 then you can pass the C Driver a hostname and it will try both IPv4 and IPv6 addresses for that host, somewhat like the standard "Happy Eyeballs" spec RFC 6555.

Lag1101:
Thanks for explanation. I haven't took into account about connection string specs. 
Nevertheless, I think we need some posibility like --ipv6 flag in mongo-client.
Besides we will try both IPv4 and IPv6, and it tooks some time for new connections, and it doesn't seem C-way. What about some function, maybe mongoc_uri_set_family, for specifying behavior by user?

ajdavis:
I'm concerned about connection startup time, too. So in CDRIVER-1972 I plan to try both IPv4 and IPv6 concurrently, so that I don't add any extra latency.

Lag1101:
Doesn't it limit user possibilities?
what if I have nginx server configured correctly for ipv4 and ipv6 hosts both, but they proxy for different endpoints, which would be selected? Seems like, it can caus for painful behaivior.

ajdavis:
You couldn't connect to MongoDB through an Nginx proxy anyway, since the MongoDB wire protocol is not HTTP. We don't generally support proxies at all besides our own proxy, mongos; but in a situation like the one you describe, an IPv4 or an IPv6 address, instead of a hostname, would allow you to specify which protocol to use.

Lag1101:
Thanks. I am looking forward to updates.

PR #434, opened 2017-04-17
Remove unnecessary string comparisons in mongoc_uri_option_is_utf8()

jmikola:
The default return was changed to false in 16d03dc58f598c1671d20d658681c0d871f96514, which means we need only check for known options.


bjori:
heh. Nice catch, thanks!

PR #435, opened 2017-04-25
CDRIVER-2141: Set error if authSource is invalid GSSAPI and X509

jmikola:
https://jira.mongodb.org/browse/CDRIVER-2142


bjori:
Looks good. Thanks!

PR #436, opened 2017-06-13
In *BSD platforms, using seedless arc4random* function,

devnexen:
here for the simple random function.


ajdavis:
Thanks! Can you tell me please, why is this an improvement over rand_r?

devnexen:
Basically better randomisation distribution and range.

ajdavis:
Thanks for explaining. I'm not going to accept this change. The wrapper function `_mongoc_rand_simple` is only used to randomly distribute queries among a small number of servers (typically two or three of them) so the most important thing is to keep the code simple and portable.

PR #440, opened 2017-09-15
Add the missing CMake LIBRESSL option

AlexMasterov:
Unable to configure against `LibreSSL` on Alpine.
```sh
cmake -DENABLE_SSL=LIBRESSL
```
```sh
CMake Error at CMakeLists.txt:9 (_message):   
    ENABLE_SSL option must be DARWIN, WINDOWS, OPENSSL, AUTO, or OFF
```
[1.8.0/CMakeLists.txt#L113-L115](https://github.com/mongodb/mongo-c-driver/blob/1.8.0/CMakeLists.txt#L113-L115):
```cmake
if (ENABLE_SSL STREQUAL LIBRESSL)
  set (LIBRESSL 1)
endif ()
```
This condition will never be executed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ endfunction()
 set (ENABLE_SSL AUTO CACHE STRING
      "Enable TLS connections and SCRAM-SHA-1 authentication.\ Options are
      \"DARWIN\" to use Apple's Secure Transport, \"WINDOWS\" to use Windows
-     Secure Channel, \"OPENSSL\", \"AUTO\",\ or \"OFF\". These options are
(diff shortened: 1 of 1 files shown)


bjori:
oh. whops.

Note that I strongly suggest you use the normal auto* routine instead, (e.g. `./configure --enable-ssl=libressl` since that is the configuration we have in our CI

I've filed a ticket for this in our JIRA: https://jira.mongodb.org/browse/CDRIVER-2289

ajdavis:
Thanks for this contribution! We'll take a look soon, sorry for the delay.

ajdavis:
Actually, now that I look it's clear this is an improvement. Thank you.

PR #444, opened 2017-09-22
CDRIVER-2297: Space out parents to avoid unterminated string error

jmikola:
https://jira.mongodb.org/browse/CDRIVER-2297


ajdavis:
"Space out parens", otherwise LGTM

PR #457, opened 2017-10-13
CXX-2324 'test-mongoc-topology.c' does not compile if ENABLE_COMPRESSION is not set

derickr:


ajdavis:
LGTM, but could you open this against master, please, not r1.8? All development happens on master. If we discover a bug introduced by 1.8.0, we fix it on master and cherry-pick the relevant commits from master to the r1.8 branch, then release 1.8.x.

derickr:
Done

derickr:
Turns out that this is wrong, and that it only should be done for r1.8, as in master it has already been removed in d35a67a404e3311257b7ad331e4f12d69ce5ec04:

```
commit d35a67a404e3311257b7ad331e4f12d69ce5ec04
Author: Hannes Magnusson <bjori@php.net>
Date:   Fri Aug 25 10:40:24 2017 -0700

    CDRIVER-2219 Remove compression check as we support noop compressor too
```

I will swap it back so that this PR is against r1.8 again.

derickr:
Done

ajdavis:
Oh good. Sorry for my mistake.

PR #463, opened 2017-11-21
Fix Cmake error

tzulea:
this fixes the following issues:
CDRIVER-2396.
CDRIVER-2395.
Add include for check_symbol_exists module.
Set default values in cmake for macros. Thus avoiding the build errors regarding the comparison operator : 
invalid token at start of a preprocessor expression
#if MONGOC_HAVE_DNSAPI != 1


ajdavis:
Merged in https://github.com/mongodb/mongo-c-driver/commit/f81e63dd554946c5acc025ba4e9fffe90fbcfde9 and https://github.com/mongodb/mongo-c-driver/commit/7b92f88575268151fd9d2ed1ca083d398db19c67 thank you!

ajdavis:
Actually I'm reverting, this doesn't work. One part of the the fix is necessary:
```
include(CheckSymbolExists)
```
But another part of the fix is incorrect:
```
set (MONGOC_HAVE_DNSAPI 0)
set (MONGOC_HAVE_RES_NSEARCH 0)
set (MONGOC_HAVE_RES_NDESTROY 0)
set (MONGOC_HAVE_RES_NCLOSE 0)
set (MONGOC_HAVE_RES_SEARCH 0)

if (ENABLE_SRV STREQUAL ON OR ENABLE_SRV STREQUAL AUTO)
   if (WIN32)
      set (RESOLV_LIBS Dnsapi)
      set (MONGOC_HAVE_DNSAPI 1)
   else ()
      # Thread-safe DNS query function for _mongoc_client_get_srv.
      # Could be a macro, not a function, so use check_symbol_exists.
      check_symbol_exists (res_nsearch resolv.h MONGOC_HAVE_RES_NSEARCH)
      if (MONGOC_HAVE_RES_NSEARCH)
```
This doesn't work, unfortunately. We can't set a variable like MONGOC_HAVE_RES_NSEARCH to a default value of 0, then run check_symbol_exists, because check_symbol_exists does *not* overwrite the default value. See the implementation here, it quits if the variable is already defined:

https://github.com/Kitware/CMake/blob/master/Modules/CheckSymbolExists.cmake#L57

Now that I've learned this I hope we both remember it!

PR #465, opened 2017-11-22
[CDRIVER-2402]  Add option to compile the submodule libbson during mo…

tzulea:
…ngoc compilation.

Check if the current source directory is not the source directoy (i.e mongoc is a subproject). In this case build mongoc and libbson dependency


ajdavis:
Thanks, merged in https://github.com/mongodb/mongo-c-driver/commit/6a4ba9b403f215f3b06aa04fabf24ffe5b598079

PR #466, opened 2017-11-27
Do not overwrite {SNAPPY,ZLIB}_LIBS for system versions

michaelkuhn:
*_LIBS often contains library paths (-L) if returned by pkg-config.
Stripping them out causes the snappy and zlib libraries to not be found if they are not installed in a system library path.

This happens with (at least) the libmongoc package in Spack (https://github.com/spack/spack).

### DIFF
--- a/build/autotools/CheckZlib.m4
+++ b/build/autotools/CheckZlib.m4
@@ -3,7 +3,7 @@
 found_zlib=no
 
 AS_IF([test "x${with_zlib}" = xauto -o "x${with_zlib}" = xsystem], [
-   PKG_CHECK_MODULES(zlib, [zlib], [
+   PKG_CHECK_MODULES(ZLIB, [zlib], [
       found_zlib=yes
    ], [
       # If we didn't find zlib with pkgconfig, search manually. If that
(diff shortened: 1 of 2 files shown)


build/autotools/CheckZlib.m4
```diff
@@ -12,6 +12,7 @@ AS_IF([test "x${with_zlib}" = xauto -o "x${with_zlib}" = xsystem], [
       AC_CHECK_LIB([zlib], [compress2], [
```
ajdavis: Should "zlib" be uppercased here, too?: "AC_CHECK_LIB([ZLIB] ..."

michaelkuhn: As far as I understand it, no. This argument is used to construct the library name, that is, `ZLIB` would result in `libZLIB`.

ajdavis: No, it seems that "zlib" is correct here after all. Zlib installs a file at "${prefix}/lib/pkgconfig/zlib.pc", so pkg-config must search for "zlib", not "z".

ajdavis:
Thank you!

PR #467, opened 2017-11-30
Fix :Error parsing URI: Invalid host string in URI

sriharshav:
Fixes issue : 
mongoc: A '/' is required between the host list and any options.
mongoc: Error parsing URI: 'Invalid host string in URI'

mongoc_uri_parse_hosts in mongoc/mongoc-uri.c  expects '/' between host list and options.


ajdavis:
Thanks!

PR #470, opened 2017-12-08
CDRIVER-2419 mongoc_bulk_operation_set_client_session

jmikola:
https://jira.mongodb.org/browse/CDRIVER-2419

### DIFF
--- a/doc/mongoc_bulk_operation_set_client.rst
+++ b/doc/mongoc_bulk_operation_set_client.rst
@@ -1,0 +1,29 @@
+:man_page: mongoc_bulk_operation_set_client
+
+mongoc_bulk_operation_set_client()
+==================================
+
+Synopsis
+--------
+
+.. code-block:: c
+
+  void
+  mongoc_bulk_operation_set_client (mongoc_bulk_operation_t *bulk, void *client);
+
+Parameters
+----------
+
+* ``bulk``: A :symbol:`mongoc_bulk_operation_t`.
+* ``client``: A :symbol:`mongoc_client_t`.
+
+Description
+-----------
+
+Specifies a client to use for the operation. This function has an effect only if called before :symbol:`mongoc_bulk_operation_execute`.
+
+If a :symbol:`mongoc_client_session_t` has already been assigned with :symbol:`mongoc_bulk_operation_set_client_session`, the ``client`` parameter must be the same as the :symbol:`mongoc_client_t` associated with that :symbol:`mongoc_client_session_t`.
+
+Use ``mongoc_bulk_operation_set_client`` only for building a language driver that wraps the C Driver. When writing applications in C, a client is implicitly set by :symbol:`mongoc_collection_create_bulk_operation_with_opts`.
+
+The ``client`` parameter's type is a void pointer by mistake, but it must be preserved for ABI. You must only pass a :symbol:`mongoc_client_t` pointer for ``client``.
--- a/src/mongoc/mongoc-bulk-operation.c
+++ b/src/mongoc/mongoc-bulk-operation.c
@@ -841,6 +841,10 @@ mongoc_bulk_operation_set_client (mongoc_bulk_operation_t *bulk, void *client)
    BSON_ASSERT (bulk);
    BSON_ASSERT (client);
 
+   if (bulk->session) {
+      BSON_ASSERT (bulk->session->client == client);
+   }
+
    bulk->client = (mongoc_client_t *) client;
 
    /* if you call set_client, bulk was likely made by mongoc_bulk_operation_new,
@@ -849,6 +853,22 @@ mongoc_bulk_operation_set_client (mongoc_bulk_operation_t *bulk, void *client)
    if (!bulk->operation_id) {
       bulk->operation_id = ++bulk->client->cluster.operation_id;
    }
+}
+
+
+void
+mongoc_bulk_operation_set_client_session (
+   mongoc_bulk_operation_t *bulk,
+   struct _mongoc_client_session_t *client_session)
+{
+   BSON_ASSERT (bulk);
+   BSON_ASSERT (client_session);
+
+   if (bulk->client) {
+      BSON_ASSERT (bulk->client == client_session->client);
+   }
+
+   bulk->session = client_session;
 }
 
 
--- a/src/mongoc/mongoc-bulk-operation.c
+++ b/src/mongoc/mongoc-bulk-operation.c
@@ -839,6 +839,7 @@ void
 mongoc_bulk_operation_set_client (mongoc_bulk_operation_t *bulk, void *client)
 {
    BSON_ASSERT (bulk);
+   BSON_ASSERT (client);
 
    bulk->client = (mongoc_client_t *) client;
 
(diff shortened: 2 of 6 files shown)


src/mongoc/mongoc-bulk-operation.c
```diff
@@ -851,6 +851,16 @@ mongoc_bulk_operation_set_client (mongoc_bulk_operation_t *bulk, void *client)
 }
 
 
+void
+mongoc_bulk_operation_set_client_session (mongoc_bulk_operation_t *bulk,
+                                          void *client_session)
```
ajdavis: mongoc_client_t *, not void *

src/mongoc/mongoc-bulk-operation.h
```diff
@@ -130,6 +130,9 @@ mongoc_bulk_operation_set_collection (mongoc_bulk_operation_t *bulk,
                                       const char *collection);
 MONGOC_EXPORT (void)
 mongoc_bulk_operation_set_client (mongoc_bulk_operation_t *bulk, void *client);
```
ajdavis: Weird, I hadn't noticed client here was void *. Looks like a mistake.

src/mongoc/mongoc-bulk-operation.c
jmikola: I ran into a host of compile errors attempting to use `mongoc_client_session_t *` here. I assume it has something to do with include order, as it's immediately caused by including `mongoc-client-session.h` up top.

src/mongoc/mongoc-bulk-operation.c
```diff
@@ -851,6 +851,16 @@ mongoc_bulk_operation_set_client (mongoc_bulk_operation_t *bulk, void *client)
...
+{
+   BSON_ASSERT (bulk);
+
+   bulk->session = (mongoc_client_session_t *) client_session;
```
jmikola: I realized this makes it possible to assign a session belonging to a different client. A few options to address this:

 * Change this function to accept a `bson_error_t *` and return a success boolean
 * `BSON_ASSERT()` that `client_session->client` is the same as `bulk->client`

Either way, I think we also need to modify `mongoc_bulk_operation_set_client()` to unset the `client_session`. Otherwise, it'd be possible the make a valid session assignment and then screw it up by changing the client again.

Thoughts?

src/mongoc/mongoc-bulk-operation.c
ajdavis: Don't include mongoc-client-session.h, and add a forward declaration:
```
/* forward decl */
struct _mongoc_client_session_t;

void
mongoc_bulk_operation_set_client_session (mongoc_bulk_operation_t *bulk,
                                          struct _mongoc_client_session_t *client_session)
```
See my solution to a similar problem here:

https://github.com/mongodb/mongo-c-driver/blob/0dc8fb1aca4194df67afd431c55d438f9ecef56c/src/mongoc/mongoc-apm-private.h#L30-L29

Preserves typechecking, solves circular include problem.

src/mongoc/mongoc-bulk-operation.c
ajdavis: Let's BSON_ASSERT that client_session->client is the same as bulk->client, in both mongoc_bulk_operation_set_client and mongoc_bulk_operation_set_session, so the condition is checked when you set them in either order.

ajdavis:
Can you also add a .rst document for the new function? (Some existing bulk functions aren't documented but we're trying to catch up, CDRIVER-1930.)

jmikola:
> Can you also add a .rst document for the new function? (Some existing bulk functions aren't documented but we're trying to catch up, CDRIVER-1930.)

Done. I also created docs for `set_client` since I wanted to refer to it from `set_client_session`.

Also found that "sessionId" wasn't documented for `create_bulk_operation_with_opts`, so I added a paragraph there as well.

src/mongoc/mongoc-bulk-operation.h
```diff
@@ -31,6 +31,8 @@
 
 BSON_BEGIN_DECLS
 
+/* forward decl */
+typedef struct _mongoc_client_session_t mongoc_client_session_t;
```
jmikola: @ajdavis: Your example in https://github.com/mongodb/mongo-c-driver/blob/0dc8fb1aca4194df67afd431c55d438f9ecef56c/src/mongoc/mongoc-apm-private.h#L30-L29 was in a private header and I don't think we want `_mongoc_client_session_t` in the public ABI, so I didn't adapt it verbatim.

Is this kosher?

doc/mongoc_bulk_operation_set_client.rst
```diff
@@ -0,0 +1,27 @@
...
+mongoc_bulk_operation_set_client()
+==================================
+
+Synopsis
+--------
+
+.. code-block:: c
+
+  void
+  mongoc_bulk_operation_set_client (mongoc_bulk_operation_t *bulk, void *client);
```
jmikola: `void *client` may be regrettable, but it's already part of the public ABI. Should the documentation be consistent, or would you rather see `mongoc_client_t *` here?

src/mongoc/mongoc-bulk-operation.h
jmikola: Changed to `struct _mongoc_client_session_t`.

doc/mongoc_bulk_operation_set_client.rst
jmikola: Keeping documentation consistent.

doc/mongoc_bulk_operation_set_client.rst
```diff
@@ -0,0 +1,27 @@
...
+* ``client``: A :symbol:`mongoc_client_t`.
+
+Description
+-----------
+
+Specifies a client to use for the operation. This function has an effect only if called before :symbol:`mongoc_bulk_operation_execute`.
+
+If a :symbol:`mongoc_client_session_t` has already been assigned with :symbol:`mongoc_bulk_operation_set_client_session`, the ``client`` parameter must be the same as the :symbol:`mongoc_client_t` associated with that :symbol:`mongoc_client_session_t`.
+
+Use ``mongoc_bulk_operation_set_client`` only for building a language driver that wraps the C Driver. When writing applications in C, a client is implicitly set by :symbol:`mongoc_collection_create_bulk_operation_with_opts`.
```
ajdavis: Let's add a regrettable paragraph at the end, "The ``client`` parameter's type is a void pointer by mistake, but it must be preserved for ABI. You must only pass a ``mongoc_client_t`` pointer for ``client``."

src/mongoc/mongoc-bulk-operation.h
ajdavis: I'm not seeing that change in this GitHub PR.

jmikola: Looks like I updated the function but forgot to change this forward decl. Will revise that Monday morning.

doc/mongoc_bulk_operation_set_client.rst
jmikola: Done.

jmikola:
@ajdavis: Updated with fixed forward decl and `void *` documentation note. Please re-review.

PR #473, opened 2017-12-12
CDRIVER-2296 pass bson_validate_flags_t to update/insert functions

xy24:
User can now pass bson_validate_flags_t to update/insert functions via "validate" field in the opt.


ajdavis:
Thanks!

PR #474, opened 2017-12-12
Windows: fixes for implicit declaration of functions

jeroen:
This fixes two compiler warnings on mingw-w64:

```c
mongoc/mongoc-socket.c: In function 'mongoc_socket_close':
mongoc/mongoc-socket.c:774:4: warning: implicit declaration of function 'getpid' [-Wimplicit-function-declaration]
    owned = (sock->pid == (int) getpid ());
    ^
```

This PR switched to `_getpid()` on Windows (which is already used elsewhere in the code).

```c
mongoc/mongoc-util.c -o mongoc/mongoc-util.o
mongoc/mongoc-util.c: In function '_mongoc_rand_simple':
mongoc/mongoc-util.c:33:4: warning: implicit declaration of function 'rand_s' [-Wimplicit-function-declaration]
    err = rand_s (&ret);
    ^
```

The [windows documentation](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/rand-s) says:

> The `rand_s` function requires that constant `_CRT_RAND_S` be defined prior to the inclusion statement for the function to be declared.

I confirmed that this fixes the compiler warning.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,12 +250,6 @@ endif ()
 include(CheckIncludeFiles)
 CHECK_INCLUDE_FILES(strings.h HAVE_STRINGS_H)
 
-if (MINGW)
-   # Define _CRT_RAND_S, so that MinGW's stdlib.h declares Microsoft's rand_s().
-   message (" MinGW: define _CRT_RAND_S")
-   set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D_CRT_RAND_S")
-endif ()
-
 include(CheckTypeSize)
 if (WIN32)
    SET(CMAKE_EXTRA_INCLUDE_FILES "ws2tcpip.h")
--- a/src/mongoc/mongoc-cluster-sspi.c
+++ b/src/mongoc/mongoc-cluster-sspi.c
@@ -140,11 +140,11 @@ _mongoc_cluster_auth_node_sspi (mongoc_cluster_t *cluster,
 {
    mongoc_cmd_parts_t parts;
    mongoc_sspi_client_state_t *state;
-   uint8_t buf[4096] = {0};
+   SEC_CHAR buf[4096] = {0};
    bson_iter_t iter;
    uint32_t buflen;
    bson_t reply;
-   char *tmpstr;
+   const char *tmpstr;
    int conv_id;
    bson_t cmd;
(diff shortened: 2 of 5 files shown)


ajdavis:
Thanks! The getpid fix looks clearly correct to me.

In CMakeLists.txt we set the _CRT_RAND_S flag with a command line switch:
```
if (MINGW)
   # Define _CRT_RAND_S, so that MinGW's stdlib.h declares Microsoft's rand_s().
   message (" MinGW: define _CRT_RAND_S")
   set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D_CRT_RAND_S")
endif ()
```
I forget, do you use Autotools to build libmongoc with mingw-w64, or do you use CMake?

jeroen:
Ah okay I have manually hardcoded the config files for windows. I can add the macro to `mongo-config.h` instead; I just thought it's better to have it in the file with the actual `rand_s()` code.

jeroen:
Should I remove `_CRT_RAND_S` from the PR and pass it as a compiler flag instead?

jeroen:
Update: added a commit that fixes type warnings in the sspi code, see [CDRIVER-2427](https://jira.mongodb.org/browse/CDRIVER-2427)

ajdavis:
Thanks. What's your opinion, should we remove the flag from CMakeLists and define it in the source file instead? If so, go ahead and make that change here.

Sorry for vacillating.

jeroen:
In this case I prefer to define the macro in the source rather than a compiler flag because it is not an *option*, you need this macro to enable the `rand_s()` call.

Also the macro is only needed for a single `c` file, so better not enable globally which could cause other side effects (though in this particular macro that seems unlikely).

ajdavis:
Yeah, that sounds right to me too.

jeroen:
Should I restore it in this PR?

ajdavis:
Yes please, and delete the flag from CMakeLists

jeroen:
OK done

ajdavis:
Thanks!

ajdavis:
Merged to master and cherry-picked into the r1.9 branch as well.

jeroen:
Thanks. I don't see it (yet) in r1.9, can you push please?

ajdavis:
Oh yes. Sorry about that.

PR #476, opened 2017-12-14
Setting error to NULL when BSON creation succeeds

road-t:
Functions bson_new_from_json(), bson_init_from_json() and bson_json_reader_new_from_file() don't reset passed bson_error_t structure, causing unpredictable behavior in some cases

### DIFF
--- a/src/libbson/src/bson/bson-json.c
+++ b/src/libbson/src/bson/bson-json.c
@@ -2229,6 +2229,10 @@ bson_new_from_json (const uint8_t *data, /* IN */
    r = bson_json_reader_read (reader, bson, error);
    bson_json_reader_destroy (reader);
 
+   if (error) {
+      memset (error, 0, sizeof (bson_error_t));
+   }
+
    if (r == 0) {
       bson_set_error (error,
                       BSON_ERROR_JSON,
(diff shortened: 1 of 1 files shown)


ajdavis:
Hi, the contract for most libbson and libmongoc functions is that they fill out the error struct if and only if they return false or NULL. If the function *succeeds* then there's no guarantee about the contents of the error struct at all.

Are there any functions you know of that don't fulfill this contract?

road-t:
Hi

I've checked some common functions of mongoc and those DO fill with zeroes error struct before any bson_set_error calls. So I decided that it's uncommon for the whole library not to do that.
Also IMO if there's a some error-handling variable, it must be set to 0 if operation succeeds.

Weird thing is that bug led to crash of our application on Ubuntu but made no harm on MacOS X.

I am not aware of any other functions that don't fulfill the contract yet.

ajdavis:
It looks like bson_validate_with_error and bson_reader_new_from_file also don't clear the error on success. And in libmongoc there are dozens of functions that take an error pointer and don't clear it on success. Since it's not currently a consistent or documented behavior for libbson or libmongoc to clear errors on success, I'm not going to commit to it. Thanks for your contribution, though; even though I disagree I do appreciate it.

road-t:
I got your point. Looks like we gotta use our own fork with those enhancements in all functions that use pointer to error structure.

ajdavis:
Thanks for understanding. I'd suggest that instead of using a fork, you use the official libbson. Check the return value of each function call instead of checking whether the error code is zero. That's the documented method for error checking that I'm committed to preserving in libbson and libmongoc.

PR #477, opened 2017-12-19
 CDRIVER-2431 check for empty command before comparing name

jmikola:
https://jira.mongodb.org/browse/CDRIVER-2431

`clang-format` picked up bad formatting in `mongoc-client.c`, which I split off into a separate commit.

### DIFF
--- a/src/mongoc/mongoc-client.c
+++ b/src/mongoc/mongoc-client.c
@@ -1563,9 +1563,8 @@ retry:
     * a new writable stream and retry. If server selection fails or the selected
     * server does not support retryable writes, fall through and allow the
     * original error to be reported. */
-   if (!ret && is_retryable &&
-       (error->domain == MONGOC_ERROR_STREAM ||
-        mongoc_cluster_is_not_master_error (error))) {
+   if (!ret && is_retryable && (error->domain == MONGOC_ERROR_STREAM ||
+                                mongoc_cluster_is_not_master_error (error))) {
       bson_error_t ignored_error;
 
       /* each write command may be retried at most once */
@@ -1578,8 +1577,9 @@ retry:
       retry_server_stream =
          mongoc_cluster_stream_for_writes (&client->cluster, &ignored_error);
 
-      if (retry_server_stream && retry_server_stream->sd->max_wire_version >=
-                                    WIRE_VERSION_RETRY_WRITES) {
+      if (retry_server_stream &&
+          retry_server_stream->sd->max_wire_version >=
+             WIRE_VERSION_RETRY_WRITES) {
          parts->assembled.server_stream = retry_server_stream;
          GOTO (retry);
       }
--- a/src/mongoc/mongoc-client.c
+++ b/src/mongoc/mongoc-client.c
@@ -1713,6 +1713,7 @@ _mongoc_client_command_with_opts (mongoc_client_t *client,
                                   bson_error_t *error)
 {
    mongoc_cmd_parts_t parts;
+   const char *command_name;
    mongoc_server_stream_t *server_stream = NULL;
    mongoc_cluster_t *cluster;
    bson_t reply_local;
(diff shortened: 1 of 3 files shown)


src/mongoc/mongoc-client.c
```diff
@@ -1783,8 +1783,20 @@ _mongoc_client_command_with_opts (mongoc_client_t *client,
...
+
+         if (!command_name) {
+            bson_set_error (error,
+                            MONGOC_ERROR_COMMAND,
+                            MONGOC_ERROR_COMMAND_INVALID_ARG,
+                            "Empty command document");
+            GOTO (err);
+         }
+
+         is_fam = !strcasecmp (command_name, "findandmodify");
```
ajdavis: We have bson_strcasecmp, let's standardize on it.

src/mongoc/mongoc-client.c
```diff
@@ -1783,8 +1783,20 @@ _mongoc_client_command_with_opts (mongoc_client_t *client,
       if ((mode & MONGOC_CMD_WRITE) &&
           !mongoc_write_concern_is_default (default_wc) &&
           (!opts || !bson_has_field (opts, "writeConcern"))) {
-         bool is_fam =
-            !strcasecmp (_mongoc_get_command_name (command), "findandmodify");
+         const char *command_name;
+         bool is_fam;
+
+         command_name = _mongoc_get_command_name (command);
```
ajdavis: I suggest putting this check first in the function to do the simple check early. Otherwise we risk the same bug again later if we need the command name in a different branch of this function.

ajdavis:
If you want to make the proposed changes and then push to master and cherry-pick to r1.9 please be my guest!

src/mongoc/mongoc-client.c
jmikola: Can do. That certainly improves maintainability. I didn't start with that since it entailed adding overhead to the entire function.

src/mongoc/mongoc-client.c
jmikola: Considering that `strcasecmp()` isn't even defined on Windows, I'm not sure how this code would have even compiled. There are many more uses of `strcasecmp()` throughout libmongoc as well.

I ultimately traced it back to this bit in `mongoc-util-private.h`, which is commonly included:

```
/* string comparison functions for Windows */
#ifdef _WIN32
#define strcasecmp _stricmp
#define strncasecmp _strnicmp
#endif
```

I'm not going to change any of this in this PR, but I'll create a ticket so that you can standardize down the line. IMO, the `mongoc-util-private.h` approach is preferable to adding a custom function that delegates. Even if that function is inlined by the compiler, there is cognitive overhead with having to use `bson_strcasecmp()` for case-insensitive but `strcmp()` for regular comparisons.

src/mongoc/mongoc-client.c
ajdavis: OK. Checking whether the command name is NULL (meaning the command is empty) is cheap constant time, let's go ahead and do it. Checking if the command name is "findandmodify" could conceivably take a couple microseconds, on the other hand.

src/mongoc/mongoc-client.c
ajdavis: Fair.

jmikola: Opened [CDRIVER-2437](https://jira.mongodb.org/browse/CDRIVER-2437).

src/mongoc/mongoc-client.c
```diff
@@ -1730,6 +1731,16 @@ _mongoc_client_command_with_opts (mongoc_client_t *client,
...
    parts.is_write_command = (mode & MONGOC_CMD_WRITE);
 
+   command_name = _mongoc_get_command_name (command);
+
+   if (!command_name) {
+      bson_set_error (error,
+                      MONGOC_ERROR_COMMAND,
+                      MONGOC_ERROR_COMMAND_INVALID_ARG,
+                      "Empty command document");
+      GOTO (err);
```
jmikola: Note: I deferred this check until after `mongoc_cmd_parts_init()`, as the end of this function always calls `mongoc_cmd_parts_cleanup()`. While the `parts` struct is always stack-allocated, I'm not sure how the `bson_destroy()` calls within the cleanup function would deal with empty pointers (assuming `parts` is zero-allocated).

jmikola:
Merged to `master` and back-ported to `r1.9` in https://github.com/mongodb/mongo-c-driver/commit/471bfb037e8e380dde9409ba318a66f491d003be (only the segfault fix, not code formatting commit).

PR #478, opened 2017-12-22
Use gmtime_s instead of gmtime

Mansuro:
gmtime_s has security enhancements over gmtime, that is why windows generates a warning when trying to compile using gmtime, for more details:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/gmtime-s-gmtime32-s-gmtime64-s


ajdavis:
Thanks for your contribution! We'll take a look in a week when we're back from vacation.

ajdavis:
Thanks, merged in https://github.com/mongodb/mongo-c-driver/commit/8df194e8ea252c22beea3747e48270f85cee3594

PR #481, opened 2018-01-04
CDRIVER-2449 omit lsid from auth and monitoring commands

jmikola:
https://jira.mongodb.org/browse/CDRIVER-2449

Per [exceptions to sending the session ID to the server on all commands](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#exceptions-to-sending-the-session-id-to-the-server-on-all-commands), we should omit `lsid` from authentication commands and may also do so for monitoring commands. The spec only mentions `isMaster` for monitoring, but libmongoc uses `ping` for that purpose as well.


jmikola:
Cherry-picked to r1.9 in https://github.com/mongodb/mongo-c-driver/commit/f0e1ea8401802fec55e3ff5923807c1f317856bd.

PR #484, opened 2018-01-16
CDRIVER-2369 remove bundled snappy from .gitignore

jmikola:
This likely should have been removed as part of [CDRIVER-2369](https://jira.mongodb.org/browse/CDRIVER-2369).


ajdavis:
Thanks!

PR #489, opened 2018-04-02
Disable topology event callbacks when NULL is passed in

kmahar:
While most callbacks were correctly disabled when passing in `NULL`, this was not the case for topology events. This adds a fix for that, and a test - just a slight modification of `_test_topology_events` - confirming that the last event is *not* topology closed. 

Also, I guess clang-format caught a lingering error in `mongoc-topology.c`... if you'd rather I remove that from this PR let me know.


ajdavis:
Code LGTM. Have you opened a ticket for this? Please include it in the ticket number and follow our commit message style:

CDRIVER-NNNN lowercase commit message

kmahar:
Ok, done

ajdavis:
Thanks!

PR #494, opened 2018-04-26
CDRIVER-2439 don't gossip cluster time to standalones

jmikola:
https://jira.mongodb.org/browse/CDRIVER-2439

This addresses an edge case where an explicit session's cluster time may have been manually advanced. Under normal usage, sessions used with a standalone will always have an empty cluster time.

### DIFF
--- a/tests/test-mongoc-cluster.c
+++ b/tests/test-mongoc-cluster.c
@@ -1471,6 +1471,66 @@ test_cluster_command_error_op_query ()
    _test_cluster_command_error (false);
 }
 
+static void
+test_advanced_cluster_time_not_sent_to_standalone (void)
+{
+   mock_server_t *server;
+   mongoc_client_t *client;
+   mongoc_client_session_t *cs;
+   bson_t opts = BSON_INITIALIZER;
+   mongoc_collection_t *collection;
+   mongoc_cursor_t *cursor;
+   future_t *future;
+   request_t *request;
+   const bson_t *doc;
+   bson_error_t error;
+
+   server = mock_server_new ();
+   mock_server_auto_ismaster (server,
+                              "{'ok': 1.0,"
+                              " 'ismaster': true,"
+                              " 'minWireVersion': 0,"
+                              " 'maxWireVersion': 6,"
+                              " 'logicalSessionTimeoutMinutes': 30}");
+   mock_server_run (server);
+   client = mongoc_client_new_from_uri (mock_server_get_uri (server));
+
+   cs = mongoc_client_start_session (client, NULL, &error);
+   ASSERT_OR_PRINT (cs, error);
+
+   mongoc_client_session_advance_cluster_time (
+      cs, tmp_bson ("{'clusterTime': {'$timestamp': {'t': 1, 'i': 1}}}"));
+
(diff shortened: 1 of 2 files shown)


jmikola:
The test failures were actually due to mocking, so I've fixed those tests. For reasons unknown to me `_test_dollar_query()` was simulating primary/secondary responses but their isMaster indicated that they were secondaries. I was able to fix that by improving their isMaster responses. I originally looked into porting these tests to use a mock replica set (leading to https://github.com/mongodb/mongo-c-driver/pull/495), although I think that went to far. For the "omit cluster time" variations, I ended up having the tests use OP_QUERY instead of OP_MSG, which I don't think was your original intention (since OP_MSG is the only code path that cares about cluster time at all). I reverted back to basic server mocking and kept the logic to add/remove `"$clusterTime"` from each isMaster response. That said, I'm not sure that makes sense in the real world, unless cluster time requires FCV 3.6 while OP_MSG only requires 3.6.

`test_framework_clustertime_supported()` actually requires a live server, and standalones never report `"$clusterTime"` in their isMaster responses so I don't think any changes are necessary there.

jmikola:
Evergreen build: https://evergreen.mongodb.com/patch/5ae374952fbabe28b02e5cfe

tests/test-mongoc-cluster.c
```diff
@@ -1471,6 +1483,65 @@ test_cluster_command_error_op_query ()
    _test_cluster_command_error (false);
 }
 
+static void
+test_advanced_cluster_time_not_sent_to_standalone ()
```
ajdavis: Set parameters to "void", otherwise this function can accept any parameters.

tests/test-mongoc-cluster.c
```diff
@@ -1559,6 +1630,10 @@ test_cluster_install (TestSuite *suite)
                                 "/Cluster/cluster_time/comparison/pooled",
                                 test_cluster_time_comparison_pooled,
                                 test_framework_skip_if_slow);
+   TestSuite_AddMockServerTest (
+      suite,
+      "/Cluster/cluster_time/advanced_not_sent_to_standalone",
+      test_advanced_cluster_time_not_sent_to_standalone);
```
ajdavis: Skip this test if crypto's not compiled in.

jmikola:
Evergreen build: https://evergreen.mongodb.com/version/5ae3d1742fbabe28b037ea45

tests/test-mongoc-cluster.c
```diff
@@ -1471,6 +1483,65 @@ test_cluster_command_error_op_query ()
...
+   mock_server_auto_ismaster (server,
+                              "{'ok': 1.0,"
+                              " 'ismaster': true,"
+                              " 'minWireVersion': 0,"
+                              " 'maxWireVersion': 6,"
+                              " 'logicalSessionTimeoutMinutes': 30}");
+   mock_server_run (server);
+   client = mongoc_client_new_from_uri (mock_server_get_uri (server));
+
+   cs = mongoc_client_start_session (client, NULL, &error);
```
kevinAlbs: mongoc_client_session_destroy this later

PR #509, opened 2018-07-03
Update bson-memory.h

sgalipeau:
Fix calling convention for program not using __cdecl calling convention.

### DIFF
--- a/src/libbson/src/bson/bson-memory.h
+++ b/src/libbson/src/bson/bson-memory.h
@@ -31,14 +31,14 @@
 BSON_BEGIN_DECLS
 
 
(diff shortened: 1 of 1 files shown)


kevinAlbs:
Hi @sgalipeau, thank you for bringing this up and for providing a fix. This is an improvement but unfortunately not one we can make right now. Marking these functions with __cdecl would break ABI; applications could be relying on them not being marked with __cdecl.

libmongoc is ABI stable, but if we eventually decide to release another major version we will happily re-open and merge this PR. I've created a ticket, [CDRIVER-2737](https://jira.mongodb.org/browse/CDRIVER-2737) to track this.

PR #510, opened 2018-07-04
Correct the sample URI of the aggregation example

Pingmin:
Just add the missing "/".


kevinAlbs:
Thank you for the contribution @Pingmin!

Pingmin:
@kevinAlbs  It's my pleasure and Thank you too!

PR #511, opened 2018-07-09
CDRIVER-2658 Save minidump files if mongodb crashes during test

markbenvenuto:
This was already reviewed by @kevinAlbs but I do not have push permission I needed to open a pull request. 

Description from earlier CR:
There is a rare crash on Windows SSL with MongoDB that I would like to get a dump for. As such, I have migrated the Server's evergreen support for collecting crash dumps to the C Driver.

MDMP
Mongo Orchestration is running mongodb from c:\mongodb. Example is from the log in SERVER-35125

2018-05-21T16:23:22.131+0000 I CONTROL [conn9] writing minidump diagnostic file c:\mongodb\bin\mongod.2018-05-21T16-23-22.mdmp

Linux
MongoDB Evergreen hosts generate crashdumps with .core

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -302,6 +302,43 @@ functions:
         permissions: public-read
         content_type: ${content_type|text/plain}
         display_name: "orchestration.log"
+    # Gather and archive mongo coredumps.
+    - command: shell.exec
+      params:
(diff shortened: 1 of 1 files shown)


kevinAlbs:
Thanks for doing this Mark!

PR #513, opened 2018-07-09
update spec tests

spencemc:


kevinAlbs:
I only skimmed this over, but LGTM. We can merge once travis passes.

PR #516, opened 2018-07-11
CDRIVER-2530 crud spec tests

spencemc:


kevinAlbs:
Nice! LGTM. Can you comment on CDRIVER-2736 that we've omitted the one find test that triggered that failure?

PR #524, opened 2018-07-24
Document that client_set_ssl_opts overrides URI SSL options

derickr:


ajdavis:
LGTM, please squash into one commit and commit to master.

PR #525, opened 2018-07-25
Fix warning: arithmetic on a pointer to void

jeroen:
Arithmetic on a pointer to void is a GNU extension [-Wpointer-arith]


ajdavis:
Thanks!

PR #526, opened 2018-08-09
Fix for LibreSSL 2.7

jeroen:
LibreSSL version 2.7.0 and up use the OpenSSL 1.1 API
See also: https://wiki.freebsd.org/LibreSSL

I have tested this with:

 - libressl 2.7.4
 - libressl 2.7.1
 - libressl 2.6.4

And it does the right thing.


ajdavis:
Thanks!

ajdavis:
Tracked here: https://jira.mongodb.org/browse/CDRIVER-2808

PR #527, opened 2018-08-09
Document reply documents for update_one, update_many, replace_one

kmahar:
@jmikola pointed out in [mongo-swift-driver#104](https://github.com/mongodb/mongo-swift-driver/pull/104) that `upsertedCount` is not mentioned in the docs. It seems that the reply document behavior is identical across all three of these functions, so I've updated the docs to match.

### DIFF
--- a/src/libmongoc/doc/mongoc_collection_replace_one.rst
+++ b/src/libmongoc/doc/mongoc_collection_replace_one.rst
@@ -34,7 +34,7 @@ Description
 
 This function shall replace documents in ``collection`` that match ``selector`` with ``replacement``.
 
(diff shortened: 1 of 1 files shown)


ajdavis:
Thanks! Could you make the text of these three functions as similar as possible to the paragraph in mongoc_collection_insert_one.rst?:

> If you pass a non-NULL ``reply``, it is filled out with an "insertedCount" field. If there is a server error then ``reply`` contains either a "writeErrors" array with one subdocument or a "writeConcernErrors" array. The reply must be freed with :symbol:`bson:bson_destroy`.

kmahar:
oh, it looks like @jmikola beat me to making this change to update_one and update_many docs. I'll modify this to only update replace_one.

ajdavis:
Merged in https://github.com/mongodb/mongo-c-driver/commit/06d88597ff6adb10cb1f931e4dcd94e62fe33dfa thanks!

PR #528, opened 2018-08-10
Fix two warnings when using mongoc.h with the Visual Studio compiler

henrikedin:
Fix macro redefinition warning when including <inttypes.h> after mongoc.h using Visual Studio 2013 or newer.
Fix iov_len in mongoc_iovec_t on Windows to be size_t like on Linux


ajdavis:
Thanks!

PR #530, opened 2018-08-14
R1.12 small change for working on Android

CMLDMR:
**1.** in "mongo-c-driver-1.12.0\src\libmongoc\src\mongocmongoc-util.c" file

if you change this line 
`return rand_r (seed);` 
to 
`srand(seed); return rand();`

because android not support or has not a librt file. linux support but android.

code compiled successfuly with NDK R17 on windows and working well different android device and version.

I reported here: [jira.mongodb.org](https://jira.mongodb.org/browse/CXX-1625)

**2**. gridfs download file or attemp any function example: get file size or get file name affect application that crashed. may be it can became from different compiler version but both of Mingw32 5.30. but I changed Qt Library from 5.10 to 5.11. if it changed the compiler.


kevinAlbs:
Hi @CMLDMR, I don't think this pull request has the commits you want. I'm closing this, but feel free to open another.

PR #532, opened 2018-08-23
CDRIVER-2800 Restructure install directory.

henrikedin:
All mongoc headers are under mongoc/ and all bson headers are under
bson/. The preferred way of including the headers are mongoc/mongoc.h
and bson/bson.h respectively. Forwarding headers in the root are
provided for backwards compatibility.

Preparation work to be able to create darwin style frameworks.


ajdavis:
Thanks for doing this work!

PR #536, opened 2018-09-07
Improve documentation for CRUD replies

jmikola:
`update_one` and `update_many` were inconsistent with other methods.

Since `insert_many` is technically a bulk write, its reply may include multiple `writeErrors` alongside `writeConcernErrors` (with one subdocument). Revise that language and reference the bulk write examples to avoid repetition.

### DIFF
--- a/src/libmongoc/doc/mongoc_collection_insert_many.rst
+++ b/src/libmongoc/doc/mongoc_collection_insert_many.rst
@@ -38,7 +38,7 @@ To insert a single document, see :symbol:`mongoc_collection_insert_one`.
 
 For any document that does not have an "_id" field, a :symbol:`bson:bson_oid_t` will be generated locally and added to the document. If you must know the inserted document's ``_id``, generate it in your code and include it in the ``document``. The ``_id`` you generate can be a :symbol:`bson:bson_oid_t` or any other non-array BSON type.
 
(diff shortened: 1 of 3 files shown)


src/libmongoc/doc/mongoc_collection_insert_many.rst
```diff
@@ -38,7 +38,7 @@ To insert a single document, see :symbol:`mongoc_collection_insert_one`.
 
 For any document that does not have an "_id" field, a :symbol:`bson:bson_oid_t` will be generated locally and added to the document. If you must know the inserted document's ``_id``, generate it in your code and include it in the ``document``. The ``_id`` you generate can be a :symbol:`bson:bson_oid_t` or any other non-array BSON type.
 
-If you pass a non-NULL ``reply``, it is filled out with an "insertedCount" field. If there is a server error then ``reply`` contains either a "writeErrors" array with one subdocument or a "writeConcernErrors" array. The reply must be freed with :symbol:`bson:bson_destroy`.
+If you pass a non-NULL ``reply``, it is filled out with an "insertedCount" field. If there is a server error then ``reply`` may contain a "writeErrors" array and/or a "writeConcernErrors" array (see :doc:`Bulk Write Operations <bulk>` for examples). The reply must be freed with :symbol:`bson:bson_destroy`.
```
jmikola: If it's not clear that we're referencing "Bulk Write Operations" just for examples of these error fields in replies, I can add additional language. I think it makes sense to point to existing material rather than repeat ourselves too much for `insert_many`, though.

ajdavis:
Thank you!

PR #537, opened 2018-09-14
Fix for CVE-2018-16790 -- Verify bounds before binary length read.

sgayou:
As reported here: https://jira.mongodb.org/browse/CDRIVER-2819,
a heap overread occurs due a failure to correctly verify data
bounds.

In the original check, len - o returns the data left including the
sizeof(l) we just read. Instead, the comparison should check
against the data left NOT including the binary int32, i.e. just
subtype (byte*) instead of int32 subtype (byte*).


ajdavis:
Terrific, thank you. Would you please add the corrupt BSON data as a file in src/libbson/tests/binary and add a test like this one?:

https://github.com/mongodb/mongo-c-driver/blob/b3bdc57c2009eb918857bd9657bb8bf953a6840b/src/libbson/tests/test-bson.c#L1247

sgayou:
Yep, I'll see if I can get that integrated next week.

sgayou:
Added corrupt BSON example. Let me know if that wasn't what you were looking for.

ajdavis:
Thanks very much!

PR #540, opened 2018-11-02
CDRIVER-2864 CDRIVER-2861

acmorrow:
Fixes for some minor things that the embedded build is papering over right now.

- The framework "version" should be the ABI
- The short version can't have anything but x.y.z, but the regular bundle version can have some prerelaese stuff tacked onto it.


src/libbson/CMakeLists.txt
```diff
@@ -248,10 +248,17 @@ set_target_properties (bson_shared PROPERTIES VERSION 0.0.0 SOVERSION 0)
 set_target_properties (bson_shared PROPERTIES OUTPUT_NAME "bson-${BSON_API_VERSION}" PREFIX "lib")
 
 if (ENABLE_APPLE_FRAMEWORK)
+
+   set (MONGOC_FRAMEWORK_BUNDLE_VERSION ${MONGOC_MAJOR_VERSION}.${MONGOC_MINOR_VERSION}.${MONGOC_MICRO_VERSION})
+   if ("${MONGOC_PRELEASE_VERSION}" STREQUAL "")
```
ajdavis: Should this check include a "NOT"?

src/libbson/CMakeLists.txt
```diff
@@ -248,10 +248,17 @@ set_target_properties (bson_shared PROPERTIES VERSION 0.0.0 SOVERSION 0)
...
+      set (MONGOC_FRAMEWORK_BUNDLE_VERSION ${MONGOC_FRAMEWORK_BUNDLE_VERSION}${MONGOC_PRERELEASE_VERSION})
+   endif()
+
    set_target_properties(bson_shared PROPERTIES 
-      FRAMEWORK TRUE 
-      MACOSX_FRAMEWORK_BUNDLE_VERSION ${MONGOC_VERSION} 
-      MACOSX_FRAMEWORK_SHORT_VERSION_STRING ${MONGOC_VERSION}
+      FRAMEWORK TRUE
+      FRAMEWORK_VERSION ${BSON_API_VERSION}
+      MACOSX_FRAMEWORK_BUNDLE_VERSION ${MONGOC_FRAMEWORK_BUNDLE_VERSION}
```
ajdavis: If MACOSX_FRAMEWORK_BUNDLE_VERSION can include the prerelease version, then this could just be MONGOC_VERSION, right? Maybe I don't understand what you wrote in the PR comment, but this logic contradicts what you wrote there.

src/libmongoc/CMakeLists.txt
```diff
@@ -634,10 +634,17 @@ if (MONGOC_ENABLE_STATIC)
...
+   endif()
+
    set_target_properties (mongoc_shared PROPERTIES 
-      FRAMEWORK TRUE 
-      MACOSX_FRAMEWORK_BUNDLE_VERSION ${MONGOC_VERSION} 
-      MACOSX_FRAMEWORK_SHORT_VERSION_STRING ${MONGOC_VERSION}
+      FRAMEWORK TRUE
+      FRAMEWORK_VERSION ${MONGOC_API_VERSION}
+      MACOSX_FRAMEWORK_BUNDLE_VERSION ${MONGOC_FRAMEWORK_BUNDLE_VERSION}
+      MACOSX_FRAMEWORK_SHORT_VERSION_STRING ${MONGOC_RELEASED_VERSION}
```
kevinAlbs: Same here, `MONGOC_VERSION`.

src/libbson/CMakeLists.txt
acmorrow: The problem with just using MONGOC_VERSION for MACOSX_FRAMEWORK_BUNDLE_VERSION is that the `-` is disallowed. My work above basically tacks the prerelease version on sans `-` when it is set.

src/libbson/CMakeLists.txt
```diff
@@ -248,10 +248,17 @@ set_target_properties (bson_shared PROPERTIES VERSION 0.0.0 SOVERSION 0)
...
+      MACOSX_FRAMEWORK_SHORT_VERSION_STRING ${MONGOC_RELEASED_VERSION}
```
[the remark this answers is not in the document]
acmorrow: The short version string isn't allowed to be anything other than x.y.z, numeric. My understanding of MONGOC_RELEASED_VERSION is that it will, 1) always be an x.y.z version, and 2) will only differ meaningfully from MONGOC_VERSION when MONGOC_VERSION is in some prerelease state. That is exactly what we want to express here, I think: the short version string is the version that most closely models the actual three digit version we are, and if you need more detailed information (like that we are actually a prerelease of a newer version), then consult the BUNDLE_VERSION.

That would break down if there is ever a case where MONGOC_RELEASED_VERSION and MONGOC_VERSION are both just x.y.z but different.

src/libmongoc/CMakeLists.txt
acmorrow: See above.

src/libbson/CMakeLists.txt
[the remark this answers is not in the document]
kevinAlbs: Ah okay, so when VERSION_RELEASED is "1.13.0" and VERSION_CURRENT is "1.14.0-prerelease" we want to set the bundle short string to 1.13.0.

src/libbson/CMakeLists.txt
kevinAlbs: Also, according to the [docs for CFBundleVersion](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364)

> While developing a new version of your app, you can include a suffix after the number that is being updated; for example 3.1.3a1. The character in the suffix represents the stage of development for the new version. For example, you can represent development, alpha, beta, and final candidate, by d, a, b, and fc. The final number in the suffix is the build version, which cannot be 0 and cannot exceed 255. When you release the new version of your app, remove the suffix.

Right now, MONGOC_PRERELEASE_VERSION is "pre" on the master branch. Would this need another number between 0-255 appended?

acmorrow: Yeah, it might. There documentation doesn't say much about what can go here. I was assuming that `pre` would be fine, or `rc0` would be fine. But no way to know without trying I guess. I sort of don't want this to become an unreasonable constraint on your version scheme. I suppose another option would be to just completely ignore the actual value of `MONGOC_PRERELEASE_VERSION` here, and just always tack on something like `pre0`. But that is sort of a shame if we were ever in a situation where we had meaningful things like `rc1` but didn't add them.

So, I don't know. I'm unclear on how much validation is performed for the suffix. It is implied that at least some is, since it says that the terminal number must be less than 255. It also says it can't be zero, so `rc0` wouldn't actually work I guess. I think I missed that on the first time reading through that.

What do you think?

src/libbson/CMakeLists.txt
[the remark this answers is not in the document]
acmorrow: That is my best understanding, yes.

src/libbson/CMakeLists.txt
kevinAlbs: Although not ideal, I vote for always appending "pre1" unless there's some standard scheme.

Also, `MONGOC_PRERELEASE_VERSION` is only "pre" if there's any prerelease info. Normally, our full prerelease info includes a date and git hash. E.g. "1.14.0-20181015+git7e2e9caeba". As @rcsanchez97 helpfully explained to me, the reason `MONGOC_PRERELEASE_VERSION` is only "pre" is because the full version is included in the driver ismaster handshake string, and the full prerelease string would be too long.

But perhaps the full prerelease string is appropriate for the framework though, and the docs don't seem to prohibit it. If so, I think change line 254 to:

```
set (MONGOC_FRAMEWORK_BUNDLE_VERSION "${MONGOC_DIST_VERSION}pre1")
```

`MONGOC_DIST_VERSION` is the full `<major>.<minor>.<patch>-<date>+git<hash>` string.

acmorrow:
@kevinAlbs @rcsanchez97 - I rebased this and made the proposed change to just tack 'pre1' on in the pre-release case. PTAL?

acmorrow:
@kevinAlbs @rcsanchez97 and @jmikola - I'm also happy to hand this one over to the C driver team (or probably specifically @rcsanchez97) if that makes more sense. As it stands, I think this needs hand testing to verify that we get correct packages in all release/pre-release cases.

rcsanchez97:
@acmorrow The changes look right to me.  I agree that manual testing with several different branch/tag configurations is a necessity to ensure that some odd corner case does not end up with a nonsense version.  That said, I don't have ready access to a Mac.  I'd need an account set up for me on one that I could use to test this.

samantharitter:
@acmorrow I tested these changed on OSX, and unfortunately `MONGOC_RELEASED_VERSION` is no longer a variable we generate (it was removed during [CDRIVER-2973](https://jira.mongodb.org/browse/CDRIVER-2973)).  So, right now, this code won't build, because that variable is empty.  It also means that we have no good way to get the previously-released version.  Are you ok with the short version string being calculated like:
```
set (MONGOC_SHORT_VERSION ${MONGOC_MAJOR_VERSION}.${MONGOC_MINOR_VERSION}.${MONGOC_MICRO_VERSION})
```
and then we can use that value in for `MACOSX_FRAMEWORK_SHORT_VERSION_STRING`?  That will give the next version, which isn't ideal, but giving the previous version seems equally bad, and this approach is much easier to implement than calculating the previous version.  I spoke to Roberto about this, and he thinks it might be the lesser evil to report the next version rather than the previous version.

Also, it appears that your calculation for `MACOSX_FRAMEWORK_BUNDLE_VERSION` can still contain a "-", because `MONGOC_DIST_VERSION` can.  For me right now, that value is being written to the plist file as "1.14.0-devpre1".  It looks like `MONGOC_PRERELEASE_VERSION` doesn't actually contain the hyphen itself, but it is appended with a hyphen in `build/cmake/LoadVersion.cmake`.  If you are ok with the `MONGOC_SHORT_VERSION` idea I added above, I believe the framework version can be calculated like this, which removes the if statement:
```
set (MONGOC_FRAMEWORK_BUNDLE_VERSION "${MONGOC_SHORT_VERSION}${MONGOC_PRERELEASE_VERSION}")
```

samantharitter:
Closing this PR, this will be future work.

PR #556, opened 2018-12-12
Print leading zeros in datetime ms

jeroen:
See issue https://github.com/jeroen/mongolite/issues/163

This ensures that leading zero's get included in the ms part of a datetime. Currently dates get printed like this: `1998-11-07T00:00:00. 23Z` (note the blank instead of 0).

### DIFF
--- a/src/libbson/src/bson/bson-iso8601.c
+++ b/src/libbson/src/bson/bson-iso8601.c
@@ -325,7 +325,7 @@ _bson_iso8601_date_format (int64_t msec_since_epoch, bson_string_t *str)
 #endif
 
    if (msecs_part) {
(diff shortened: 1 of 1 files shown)


kevinAlbs:
Thanks for the contribution Jeroen. The travis run on your commit failed due to an unrelated failure. Can you rebase this on the tip of master and update? We'll run this through Evergreen and create a C driver ticket.

kevinAlbs:
Created https://jira.mongodb.org/browse/CDRIVER-2910 to track this bug. And this will be tested in our BSON corpus tests once https://github.com/mongodb/specifications/pull/438 is merged.

Jeroen, no need to rebase. This is a small change, and after running this through our continuous integration tests I'm confident this is correct. I'm rebasing and merging this. This will be part of the 1.14.0 release.

Thanks for investigating and fixing this!

PR #560, opened 2019-03-06
CDRIVER-2993 add missing header include checks and licenses

jmikola:
https://jira.mongodb.org/browse/CDRIVER-2993

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-opts-helpers-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-opts-helpers-private.h
@@ -1,3 +1,5 @@
+#include "mongoc/mongoc-prelude.h"
+
 #include <bson/bson.h>
 #include "mongoc/mongoc-client-session-private.h"
 #include "mongoc/mongoc-collection-private.h"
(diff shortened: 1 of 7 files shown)


src/libmongoc/src/mongoc/mongoc-opts-helpers.c
```diff
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2018 MongoDB, Inc.
```
kevinAlbs: Same

src/libmongoc/src/mongoc/mongoc-opts-helpers-private.h
```diff
@@ -1,3 +1,21 @@
+/*
+ * Copyright 2018 MongoDB, Inc.
```
kevinAlbs: 2019-present

jmikola:
> FWIW headers ending with -private aren't installed when someone does make install. But I think it is worth adding the prelude to all headers for consistency, or in case someone tries including a private header from the source.

From my comment in the ticket:

> Note: CDRIVER-2839 supposedly added an Evergreen task to test for this, so I'm not sure how these missing instances were not caught.

Is the fact that these omissions were in private headers the reason that they weren't caught by the Evergreen task? For the life of me, I couldn't figure out what test you originally added to even handle public headers :laughing:

PR #565, opened 2019-05-30
CDRIVER-3158 updated documentation for bson_iter_value

haris-sheikh:


kevinAlbs:
The failing abi-compilance-check task is unrelated to this docs change, and will be investigated later.

PR #575, opened 2019-07-10
travis: move osx to its own matrix include

grooverdan:
Rather than putting the Linux only or exceptions everywhere else.

Add linux-ppc64le as a test. Because the travis ppc64le test
infrastructure is based on xenial the distro mongodb is 2.x which isn't
suitable for the test. Pulling in 3.x from ubuntu bionic resulted in
conflicts so we use the enterprise mongo for ppc64le xenial to have
a workable test server for the client C implementation.


kevinAlbs:
Hi @grooverdan! Thank you for the contribution. But our tests on Travis are really "smoke tests". Our comprehensive testing occurs on our continuous integration system, Evergreen. We do test linux-ppc64le on Evergreen with [Ubuntu 16.04](https://evergreen.mongodb.com/build/mongo_c_driver_power8_ubuntu1604_0997aad00c42f8c0501cdaba3a6201c67bef3423_19_12_16_16_43_43), [Ubuntu 18.04](https://evergreen.mongodb.com/build/mongo_c_driver_power8_ubuntu1804_0997aad00c42f8c0501cdaba3a6201c67bef3423_19_12_16_16_43_43), and [RHEL 7.1](https://evergreen.mongodb.com/build/mongo_c_driver_power8_rhel71_0997aad00c42f8c0501cdaba3a6201c67bef3423_19_12_16_16_43_43). Additionally, in our experience we have not had many linux-ppc64le specific issues, so it does not seem warranted to add it to Travis.

PR #587, opened 2019-10-13
adding Installation instructions

efron:


kevinAlbs:
Thanks for the contribution @efron, but our top-level README.rst links to these instructions here:
https://github.com/mongodb/mongo-c-driver#installation

PR #588, opened 2019-11-08
Allow libbson and libmongoc library names to be changed externally

eldruin:
Internally we rename the libraries we integrate as to avoid interference with the library versions already available on the target systems by giving them a prefix.
This PR allows users to specify, for example:
```
-DBSON_OUTPUT_BASENAME:STRING=MYORG_bson
-DMONGOC_OUTPUT_BASENAME:STRING=MYORG_mongoc
```
I used `BASENAME` as the flavor and/or version are appended later.

Alternatively, it would be also fine for us to add a prefix variable like `OUTPUT_NAME_PREFIX` or so but I was unable to find a good name that cannot be confused the target `PREFIX` property (e.g. `lib`)

### DIFF
--- a/src/libbson/CMakeLists.txt
+++ b/src/libbson/CMakeLists.txt
@@ -8,6 +8,8 @@ if (ENABLE_TESTS AND NOT ENABLE_STATIC MATCHES "ON|AUTO")
    message (FATAL_ERROR "-DENABLE_STATIC=OFF also requires -DENABLE_TESTS=OFF")
 endif ()
 
+set(BSON_OUTPUT_BASENAME "bson" CACHE STRING "Output bson library base name")
+
 include (CheckFunctionExists)
 include (CheckIncludeFile)
 include (CheckStructHasMember)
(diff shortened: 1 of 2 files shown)


src/libbson/CMakeLists.txt
```diff
@@ -8,6 +8,8 @@ if (ENABLE_TESTS AND NOT ENABLE_STATIC MATCHES "ON|AUTO")
    message (FATAL_ERROR "-DENABLE_STATIC=OFF also requires -DENABLE_TESTS=OFF")
 endif ()
 
+set(BSON_OUTPUT_BASENAME "bson" CACHE STRING "Output bson library base name")
```
rcsanchez97: Please add a space here -> `set (` for stylistic consistency

src/libmongoc/CMakeLists.txt
```diff
@@ -28,6 +28,8 @@ set (MONGOC_ENABLE_COMPRESSION_SNAPPY 0)
 set (MONGOC_ENABLE_COMPRESSION_ZLIB 0)
 set (MONGOC_ENABLE_COMPRESSION_ZSTD 0)
 
+set(MONGOC_OUTPUT_BASENAME "mongoc" CACHE STRING "Output mongoc library base name")
```
rcsanchez97: Please add a space here -> `set (` for stylistic consistency

eldruin: done

src/libbson/CMakeLists.txt
eldruin: done

rcsanchez97:
@kevinAlbs Do these options to you like they should be documented in installing.rst?  We don't document every build option it seems and there isn't a place where these naturally fall in that document, so it would need a new section.  Also, do you have any reservations regarding this change?

rcsanchez97:
Some build tasks are failing in Evergreen.  A few I immediately recognize as known issues that are definitely not related to this change.  I will need to confirm the others, but I honestly do not expect that any of the failures will be attributable to this change.

kevinAlbs:
Thanks for the contribution @eldruin!

PR #589, opened 2019-11-13
Remove lib prefix

eldruin:
On Windows libraries normally do not have the "lib" prefix.
With this change the output libraries are named:
```
bson-1.0.dll
bson-1.0.lib
mongoc-1.0.dll
mongoc-1.0.lib
```
The lib prefix is already omitted on windows on the mongo-cxx-driver libraries, for example.
The default prefix on Unix platforms is still "lib" so there is no change there.

### DIFF
--- a/src/libbson/CMakeLists.txt
+++ b/src/libbson/CMakeLists.txt
@@ -220,7 +220,7 @@ set (CMAKE_CXX_VISIBILITY_PRESET hidden)
 target_compile_definitions (bson_shared PRIVATE BSON_COMPILATION JSONSL_PARSE_NAN)
 target_include_directories (bson_shared INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src> $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/src>)
 set_target_properties (bson_shared PROPERTIES VERSION 0.0.0 SOVERSION 0)
-set_target_properties (bson_shared PROPERTIES OUTPUT_NAME "${BSON_OUTPUT_BASENAME}-${BSON_API_VERSION}" PREFIX "lib")
+set_target_properties (bson_shared PROPERTIES OUTPUT_NAME "${BSON_OUTPUT_BASENAME}-${BSON_API_VERSION}")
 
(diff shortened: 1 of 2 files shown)


rcsanchez97:
@kevinAlbs I have reviewed the changes, fired off an Evergreen patch build based on the changes, and reviewed the build results.

It looks like the unexpected Evergreen task failures result from some of the helper scripts in the `.evergereen` directory still referencing the libraries with the `lib` prefix on Windows, but the fix there is trivial.  Since the CDRIVER-3047 change was made, the CMake package scripts automatically picked up the new library names, so anybody consuming the BSON or C driver libraries via CMake would not encounter breakage.  The pkg-config scripts appear to be unchanged, but that may be OK.  I think that the linker on Windows will search for libraries with and without the `lib` prefix.  Assuming that to be the case, the pkg-config scripts still look good for the name change.

The concerns that I have are related to my work on CDRIVER-3413 and for users directly consuming the libraries without the aid of CMake or pkg-config.  As far as CDRIVER-3413, it feels like this change should be done in conjunction with that work.  If we decide to go ahead with @eldruin 's proposed change, then I can bring it in as part of CDRIVER-3413 (and give him credit, naturally).  That way I can be sure that both the old CMake package scripts I am restoring reference the correct artifacts.  The other concern that I have is that changing the name of the library artifact has the possibility of breaking existing workflows.  In particular, linking a static library requires specifying the full path to the library archive itself, which would now be differently named on Windows.

The possible alternatives there are to produce artifacts with both the old and new names, or to postpone the change until we are ready to make compatibility-breaking changes (2.0, perhaps).  I would also appreciate the perspective of someone who is more familiar with the Windows platform to ensure I haven't missed something.

rcsanchez97:
This change will need to be accomplished in two steps.

For C driver 1.16.0: https://jira.mongodb.org/browse/CDRIVER-3449

For C driver 1.17.0: https://jira.mongodb.org/browse/CDRIVER-3450

aprilschoffer:
Hi @eldruin, thank you for your contribution. We're closing this pull request in favor of pulling the work for this into CDRIVER-3449 and CDRIVER-3450. The work you've done will be used as part of that effort. Please reach out if you have any questions.

PR #591, opened 2019-11-18
CDRIVER-3433 support linking against windows static runtime

mbroadst:
This allows dependent projects to optionally enable overriding the
default MSVC runtime (`/MD` for the dynamic mulitthreaded) to use
the static runtime (`/MT`)

### DIFF
--- a/src/libbson/CMakeLists.txt
+++ b/src/libbson/CMakeLists.txt
@@ -267,6 +267,10 @@ if (ENABLE_STATIC MATCHES "ON|AUTO")
    target_include_directories (bson_static INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src> $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/src>)
    set_target_properties (bson_static PROPERTIES VERSION 0.0.0)
    set_target_properties (bson_static PROPERTIES OUTPUT_NAME "${BSON_OUTPUT_BASENAME}-static-${BSON_API_VERSION}")
+   if (ENABLE_WINDOWS_STATIC_RUNTIME)
+      target_compile_options (bson_static PUBLIC /MT)
+   endif ()
+
    # We use CMAKE_THREAD_LIBS_INIT rather than Threads::Threads here because the
    # latter fails when building on Mac OS X
    target_link_libraries (bson_static ${CMAKE_THREAD_LIBS_INIT})


mbroadst:
@rcsanchez97 The build for the node addon will implicitly test this, but no I had not otherwise considered adding something to test it explicitly.

rcsanchez97:
Since the node addon will be the primary (only?) user of this feature,
are you comfortable with that implicit testing?  Or would you prefer
that we have some explicit testing in the C driver Evergreen builds?

mbroadst:
@rcsanchez97 I'm totally fine with the addon being the lone tester for now

src/libbson/CMakeLists.txt
```diff
@@ -267,6 +267,10 @@ if (ENABLE_STATIC MATCHES "ON|AUTO")
    target_include_directories (bson_static INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src> $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/src>)
    set_target_properties (bson_static PROPERTIES VERSION 0.0.0)
    set_target_properties (bson_static PROPERTIES OUTPUT_NAME "${BSON_OUTPUT_BASENAME}-static-${BSON_API_VERSION}")
+   if (ENABLE_WINDOWS_STATIC_RUNTIME)
```
kevinAlbs: Can this be changed to a cached option? Similar to options here: https://github.com/mongodb/mongo-c-driver/blob/1.15.0/CMakeLists.txt/#L57

A cached option will show up after configuring with cmake, ending up in sort of a help menu. Users can get a list of cached options along with the help text by running `cmake -LH` or using  cmake GUI or ccmake.

Arguably, the same should be done for https://github.com/mongodb/libmongocrypt/pull/78, but I think it's much less likely anyone outside of our drivers team is using libmongocrypt directly.

jmikola: I'm not familiar with CMake, but should this option only exist on Windows platforms?

Coming from an Autotools perspective, I am imagining that this behaves similar to a configure option, where we'd simply prefer to have it omitted entirely from the list of supported options if we know it's not relevant to the platform. If that's not the case for CMake, please correct me.

kevinAlbs: Sorry, some discussion/investigation has happened outside of this PR and I did not follow-up. This PR can hopefully be closed without merging. I was able to pass `/MT` through the relwithdebinfo Windows compile using `-DCMAKE_C_FLAGS_RELWITHDEBINFO`. @mbroadst if you can verify the changes on this branch work (https://github.com/kevinAlbs/libmongocrypt/tree/proposed-fix), shall we close this PR?

Though I do agree it'd probably make sense to have this as a Windows only option.

mbroadst:
closing because @kevinAlbs found a way to do this without changing libbson or libmongocrypt :v:

PR #597, opened 2020-03-17
Fix for mingw-w64 Windows build

jeroen:
I am not sure why Windows is special cased here, but mingw-w64 / gcc needs the same logic as other systems, otherwise we get compiler warnings about non-standard c99.


kevinAlbs:
LGTM. As I understand it, _WIN32 defines the platform, and _MSC_VER the compiler. So this switches from __FUNCTION__ to __func__ which is C99 standard according to https://gcc.gnu.org/onlinedocs/cpp/Standard-Predefined-Macros.html. Thank you for the contribution @jeroen.

PR #598, opened 2020-04-01
Bugfix/integer overflow

richarddonkin:
Fix a couple of issues that only appear on a 32-bit or 64-bit architecture.

Both issues are based on integer overflows and could have potentially
serious consequences.

### DIFF
--- a/src/libbson/src/bson/bson.c
+++ b/src/libbson/src/bson/bson.c
@@ -105,7 +105,7 @@ static const uint8_t gZero;
  *       a malloc based buffer.
  *
  * Returns:
- *       true if successful; otherwise false indicating INT_MAX overflow.
+ *       true if successful; otherwise false indicating BSON_MAX_SIZE overflow.
  *
  * Side effects:
  *       None.
@@ -127,7 +127,7 @@ _bson_impl_inline_grow (bson_impl_inline_t *impl, /* IN */
 
    req = bson_next_power_of_two (impl->len + size);
 
-   if (req <= INT32_MAX) {
+   if (req <= BSON_MAX_SIZE) {
       data = bson_malloc (req);
 
       memcpy (data, impl->data, impl->len);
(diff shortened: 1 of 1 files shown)


richarddonkin:
It's worth mentioning that I don't believe either of these issues constitutes a security vulnerability. The functions `bson_new_from_buffer()` and `bson_compare()` will however behave unexpectedly.

src/libbson/src/bson/bson.c
```diff
@@ -127,7 +127,7 @@ _bson_impl_inline_grow (bson_impl_inline_t *impl, /* IN */
 
    req = bson_next_power_of_two (impl->len + size);
 
-   if (req <= INT32_MAX) {
```
kevinAlbs, line 130: I think this is a reasonable change, since BSON_MAX_SIZE better conveys what this should be checking. As a sanity check, `BSON_MAX_SIZE` is equal to `INT32_MAX` right? There should be no observable change here.

src/libbson/src/bson/bson.c
```diff
@@ -2105,7 +2105,7 @@ bson_new_from_data (const uint8_t *data, size_t length)
 
    BSON_ASSERT (data);
 
-   if ((length < 5) || (length > INT_MAX) || data[length - 1]) {
```
kevinAlbs, line 2108: But this could be a bug, since `INT_MAX` may differ from `INT32_MAX`/`BSON_MAX_SIZE`

src/libbson/src/bson/bson.c
richarddonkin: Yes, BSON_MAX_SIZE is equal to INT32_MAX. This changed line will not change the behaviour of the code.

src/libbson/src/bson/bson.c
richarddonkin: Yes, this change is an actual behaviour change. I do believe using INT_MAX is a bug. The largest BSON document size allowed is BSON_MAX_SIZE, but INT_MAX can be much greater than that on a 64-bit architecture.

PR #599, opened 2020-04-07
Improve error messaging when a NULL parameter is encountered.

jweinst1:
## Summary

This pull request makes a change to the way the mongoc driver reports encountering `NULL` in function parameters where a non `NULL` pointer is expected. Currently, a confusing error message is given:

```
mongoc_collection_count_documents(): precondition failed: coll
```

This arises from the  `BSON_ASSERT` macro, defined as:

```c
#define BSON_ASSERT(test)                                  \
   do {                                                    \
      if (!(BSON_LIKELY (test))) {                         \
         fprintf (stderr,                                  \
                  "%s:%d %s(): precondition failed: %s\n", \
                  __FILE__,                                \
                  __LINE__,                                \
                  BSON_FUNC,                               \
                  #test);                                  \
         abort ();                                         \
      }                                                    \
   } while (0)
```

This can work for typical conditions, like `foo != 0`, where the condition is properly transformed into a C-string literal by the preprocessor. However, in the case of NULL values, it just transforms the parameter name, which makes the message confusing. One could think there is something wrong with the parameter, or many different causes, rather than it being NULL.

The solution proposed here is to give a more straight forward message for NULL parameters:

```c
#define BSON_ASSERT_PARAM(param)                                                   \
   do {                                                                            \
      if ((BSON_LIKELY (param == NULL))) {                                         \
         fprintf (stderr,                                                          \
                  "The parameter: %s, in function %s, cannot be NULL\n", \
                  #param,                                                          \
                  BSON_FUNC);                                                      \
         abort ();                                                                 \
      }                                                                            \
   } while (0)   
```

This is much more direct, and accurate way of communicating the error. 

### Notes

As a side note, I don't believe the line number is relevant for these error messages, because that line number does not correspond to where the function is called in a program linking the mongoc driver, it corresponds to the source code of the C driver itself.

### DIFF
--- a/src/libbson/src/bson/bson-macros.h
+++ b/src/libbson/src/bson/bson-macros.h
@@ -196,6 +196,18 @@
          abort ();                                         \
       }                                                    \
    } while (0)
+       
+/* Used for asserting parameters to provide a more precise error message */
+#define BSON_ASSERT_PARAM(param)                                                   \
+   do {                                                                            \
+      if ((BSON_LIKELY (param == NULL))) {                                         \
+         fprintf (stderr,                                                          \
+                  "The parameter: %s, in function %s, cannot be NULL\n", \
+                  #param,                                                          \
+                  BSON_FUNC);                                                      \
+         abort ();                                                                 \
+      }                                                                            \
+   } while (0)      
 
 /* obsolete macros, preserved for compatibility */
 #define BSON_STATIC_ASSERT(s) BSON_STATIC_ASSERT_ (s, __LINE__)
--- a/src/libbson/src/bson/bson-macros.h
+++ b/src/libbson/src/bson/bson-macros.h
@@ -200,7 +200,7 @@
 /* Used for asserting parameters to provide a more precise error message */
 #define BSON_ASSERT_PARAM(param)                                                   \
    do {                                                                            \
-      if ((BSON_UNLIKELY (param == NULL))) {                                         \
+      if ((BSON_UNLIKELY (param == NULL))) {                                       \
          fprintf (stderr,                                                          \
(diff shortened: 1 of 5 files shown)


src/libbson/src/bson/bson-macros.h
```diff
@@ -196,6 +196,18 @@
          abort ();                                         \
       }                                                    \
    } while (0)
+       
+/* Used for asserting parameters to provide a more precise error message */
+#define BSON_ASSERT_PARAM(param)                                                   \
+   do {                                                                            \
+      if ((BSON_LIKELY (param == NULL))) {                                         \
```
kevinAlbs, line 203: I think this should be `BSON_UNLIKELY` here.

src/libbson/src/bson/bson-macros.h
```diff
@@ -196,6 +196,18 @@
...
+         fprintf (stderr,                                                          \
+                  "The parameter: %s, in function %s, cannot be NULL\n", \
```
kevinAlbs, line 205: Please align the trailing `\`

src/libbson/src/bson/bson-macros.h
```diff
@@ -196,6 +196,18 @@
          abort ();                                         \
       }                                                    \
    } while (0)
+       
+/* Used for asserting parameters to provide a more precise error message */
+#define BSON_ASSERT_PARAM(param)                                                   \
+   do {                                                                            \
+      if ((BSON_UNLIKELY (param == NULL))) {                                         \
```
kevinAlbs, line 203: Align trailing `\`

kevinAlbs:
Thanks again for the contribution!

PR #600, opened 2020-04-15
Fix MATCHES usage so ENABLE_STATIC=BUILD_ONLY works

remicollet:
Because BUILD_**ON**LY matches **ON**, it doesn't work as expected.

I also add some debug message, feel free to drop them...


rcsanchez97:
@remicollet Well, that was an embarrassing oversight on my part.  Thanks so much for catching the error and fixing it!

PR #601, opened 2020-05-04
CDRIVER-3615 reduce races in SDAM err handling

kevinAlbs:
**Summary of changes**
- Replace the topology scanner "timestamp" with a server description "generation". This allows invalidating all connections to a server. The generation is used in the new SDAM tests.
- Use the generation of the connection to determine whether or not to handle an application connection error.  Application errors are network errors, or a state change (aka "not master"/"node is recovering), defined in SDAM.
- Update error handling so state change errors do not apply if they do not  include a newer topologyVersion.


src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -1682,3 +1659,165 @@ _mongoc_topology_bypass_cooldown (mongoc_topology_t *topology)
...
+   if (!sd) {
+      /* Server removed, ignore and ignore error. */
+      return;
+   }
+   TRACE ("clearing pool for server: %s", sd->host.host_and_port);
+   sd->generation++;
+}
+
+
+/* Handle an error from an app connection.
```
kevinAlbs, line 1704: Arguably, this logic should be in mongoc-cluster.c, which is responsible for creating and managing connections for application use. However, the new SDAM tests mock application errors. I found it simpler to consolidate the logic of the [Application Errors|https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#application-errors] section here, so this can be used by the tests and by mongoc-cluster.c.

Consequently, the spec tests only test the logic of updating the server description and generation on application errors. Shane has integration tests in the works (https://github.com/ShaneHarvey/specifications/pull/1). And libmongoc has some existing test coverage of handling "not master" errors with tests using the mock server.

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -266,7 +325,10 @@ mongoc_cluster_run_command_opquery (mongoc_cluster_t *cluster,
                                     cluster->iov.len,
                                     cluster->sockettimeoutms,
                                     error)) {
-      mongoc_cluster_disconnect_node (cluster, server_id, true, error);
```
kevinAlbs, line 269: libmongoc previously distinguished read and write timeouts (and would not invalidate a server for a read timeout). SDAM currently says a post-handshake timeout should not invalidate (and does not distinguish). I've been mostly convinced this is the wrong behavior and changed this to conform to the spec. Though I created https://jira.mongodb.org/browse/DRIVERS-1246 for further discussion.

src/libmongoc/src/mongoc/mongoc-cluster-private.h
```diff
@@ -43,14 +43,14 @@ BSON_BEGIN_DECLS
 typedef struct _mongoc_cluster_node_t {
    mongoc_stream_t *stream;
    char *connection_address;
+   uint32_t generation;
 
+   /* TODO CDRIVER-3653, these fields are unused. */
```
kevinAlbs, line 48: This is a general issue in the driver. A `mongoc_server_stream_t` is a `mongoc_stream_t` with a server description  attached. We attach that server description when a stream is fetched. But in between server selection and fetching the stream, another thread could mark the server description as Unknown. That has the unfortunate consequence of making wire version checks think we have maxWireVersion 0.

I think we should fix this, but this seemed out of scope for this change.

kevinAlbs:
Something is up with the patch build reporting these errors on variants with older git reporting: `error: unknown option merged`

That happens during cmake configuration. I suspect this is due to how patch builds are being applied from GitHub. I'll take a look, but I do not think those failures are due to these changes.

src/libmongoc/src/mongoc/mongoc-server-description.c
```diff
@@ -80,14 +80,12 @@ mongoc_server_description_reset (mongoc_server_description_t *sd)
    bson_destroy (&sd->arbiters);
    bson_destroy (&sd->tags);
    bson_destroy (&sd->compressors);
-   bson_destroy (&sd->topology_version);
```
kevinAlbs, line 83: Worth mentioning, this was a bug identified by the spec tests. When a server description is marked as Unknown, it should still keep the previous topology version. That way, future state change errors can be ignored. We should only remove the topology version if we see handle another ismaster reply that does not contain it.

src/libmongoc/tests/test-mongoc-error.c
```diff
@@ -133,6 +134,41 @@ test_has_label (void)
    BSON_ASSERT (!mongoc_error_has_label (tmp_bson ("{}"), "foo"));
 }
 
+static void
+test_state_change (void)
```
jmikola, line 138: If my earlier comment about also checking error domains is relevant, this may need to be adjusted.

src/libmongoc/tests/test-mongoc-exhaust.c
```diff
@@ -28,28 +28,24 @@ skip_if_mongos (void)
...
-      mongoc_cluster_node_t *cluster_node;
-
-      cluster_node = (mongoc_cluster_node_t *) mongoc_set_get (
-         client->cluster.nodes, server_id);
+   bson_mutex_lock (&client->topology->mutex);
+   sd = mongoc_topology_description_server_by_id (
+      &client->topology->description, server_id, &error);
+   ASSERT_OR_PRINT (sd, error);
+   generation = sd->generation;
+   bson_mutex_unlock (&client->topology->mutex);
```
jmikola, line 46: Like my earlier questions, should `mongoc_topology_description_server_by_id` be moved directly after `mongoc_topology_description_server_by_id`? I suppose if the assertion fails we don't care about leaving the mutex locked either way.

src/libmongoc/tests/test-mongoc-exhaust.c
```diff
@@ -160,7 +156,7 @@ test_exhaust_cursor (bool pooled)
       }
       BSON_ASSERT (r);
       BSON_ASSERT (doc);
-      ASSERT_CMPINT64 (timestamp1, <, get_timestamp (client, cursor2));
+      ASSERT_CMPINT64 (generation1, ==, get_generation (client, cursor2));
```
jmikola, line 159: Why did the previous timestamp comparison use `<` but we do an equality check here?

src/libmongoc/src/mongoc/mongoc-cluster-private.h
```diff
@@ -43,14 +43,14 @@ BSON_BEGIN_DECLS
 typedef struct _mongoc_cluster_node_t {
    mongoc_stream_t *stream;
    char *connection_address;
+   uint32_t generation;
```
jmikola, line 46: Would it make sense to name this `pool_generation` or is `mongoc_cluster_node_t` synonymous with a pool?

Perhaps this makes more sense for the `generation` field on the SD object, as I was getting confused as to whether that corresponded to a pool or not. Ideally, should an SD always have the same `generation` as its cluster unless it's stale?

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -1682,3 +1659,165 @@ _mongoc_topology_bypass_cooldown (mongoc_topology_t *topology)
...
+   bson_error_t error;
+
+   sd = mongoc_topology_description_server_by_id (
+      &topology->description, server_id, &error);
+   if (!sd) {
+      /* Server removed, ignore and ignore error. */
+      return;
+   }
+   TRACE ("clearing pool for server: %s", sd->host.host_and_port);
+   sd->generation++;
```
jmikola, line 1700: I may be missing it, but where is the cluster's `generation` field ever incremented?

src/libmongoc/tests/json-test.c
```diff
@@ -176,47 +176,149 @@ server_description_by_hostname (mongoc_topology_description_t *topology,
...
+            &app_error_field_iter, &app_error, "maxWireVersion"));
+         BSON_ASSERT (BSON_ITER_HOLDS_INT32 (&app_error_field_iter));
+         max_wire_version = bson_iter_int32 (&app_error_field_iter);
+
+         BSON_ASSERT (
+            bson_iter_init_find (&app_error_field_iter, &app_error, "when"));
+         BSON_ASSERT (BSON_ITER_HOLDS_UTF8 (&app_error_field_iter));
+         when_str = bson_iter_utf8 (&app_error_field_iter, 0);
+         if (0 == strcmp (when_str, "beforeHandshakeCompletes")) {
+            when = MONGOC_SDAM_APP_ERROR_BEFORE_HANDSHAKE;
```
jmikola, line 280: Is this the only appearance of `MONGOC_SDAM_APP_ERROR_BEFORE_HANDSHAKE` in the entire PR? I only recall seeing `MONGOC_SDAM_APP_ERROR_AFTER_HANDSHAKE` elsewhere.

Is that an oversight if so? I would have assumed there'd be at least once instance of the "before" case in the driver.

src/libmongoc/src/mongoc/mongoc-server-description.c
kevinAlbs: I don't think so. The addition of topology version is only on master, not released (even in beta). 857d9de942b01ad48b1baa4d5ca2bb0cd05b4c96

src/libmongoc/tests/test-mongoc-exhaust.c
kevinAlbs: This is an important distinction. Closing the socket for an exhaust cursor does not mean all connections to that server should be invalidated. So we should not have cleared the pool.

I added another check to assert the next cursor is using a different stream.

kevinAlbs: Welp... that new check didn't work out. My hypothesis is that a newly allocated stream could reuse the address. I changed this to use a separate client to check the connection count from serverStatus.

This is also what the stepdown tests do:
https://github.com/mongodb/specifications/blob/master/source/connections-survive-step-down/tests/README.rst#how-to-verify-connection-count

src/libmongoc/src/mongoc/mongoc-server-description.c
```diff
@@ -667,12 +670,20 @@ mongoc_server_description_handle_ismaster (mongoc_server_description_t *sd,
          bson_destroy (&sd->compressors);
          BSON_ASSERT (bson_init_static (&sd->compressors, bytes, len));
       } else if (strcmp ("topologyVersion", bson_iter_key (&iter)) == 0) {
-         if (!BSON_ITER_HOLDS_DOCUMENT (&iter)) {
+         if (!BSON_ITER_HOLDS_DOCUMENT (&iter) &&
+             !BSON_ITER_HOLDS_NULL (&iter)) {
```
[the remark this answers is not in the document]
jmikola: Always feels good when a casual comment on a driver CR results in a spec change :grin:

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -100,6 +100,66 @@ static void
...
+static void
+_handle_not_master_error (mongoc_cluster_t *cluster,
+                          const mongoc_server_stream_t *server_stream,
+                          const bson_t *reply)
+{
+   uint32_t server_id;
+
+   server_id = server_stream->sd->id;
+   bson_mutex_lock (&cluster->client->topology->mutex);
+   if (_mongoc_topology_handle_app_error (cluster->client->topology,
```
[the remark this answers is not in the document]
jmikola: Cross-referenced with [CDRIVER-3645](https://jira.mongodb.org/browse/CDRIVER-3645). Good to know!

src/libmongoc/src/mongoc/mongoc-cluster-private.h
jmikola: Thanks for the explanation. In this case I think the generic `generation` name would be preferable, as I was definitely misunderstanding the relationship between a cluster and pool.

src/libmongoc/tests/json-test.c
jmikola: Ah, I wasn't thinking that the enum was being used for documentation/clarity purposes. That's great so no concern here.

src/libmongoc/src/mongoc/mongoc-cluster-private.h
kevinAlbs: Gotcha, renamed it back to `generation`

PR #602, opened 2020-05-04
CDRIVER-3619 reject bad setname for single topology

kevinAlbs:
For a topology type single, if the user specified a setName in the URI, the server should be marked Unknown if it reports a different setName.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-trace-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-trace-private.h
@@ -89,6 +89,23 @@ BSON_BEGIN_DECLS
                   (int) _l);                              \
       mongoc_log_trace_bytes (MONGOC_LOG_DOMAIN, _b, _l); \
    } while (0)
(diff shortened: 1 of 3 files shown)


src/libmongoc/src/mongoc/mongoc-trace-private.h
```diff
@@ -89,6 +89,23 @@ BSON_BEGIN_DECLS
...
+      }                                                             \
+      mongoc_log (MONGOC_LOG_LEVEL_TRACE,                           \
+                  MONGOC_LOG_DOMAIN,                                \
+                  "TRACE: %s():%d %s = %s",                         \
+                  BSON_FUNC,                                        \
+                  __LINE__,                                         \
+                  #_bson,                                           \
+                  _bson_str);                                       \
+      bson_free (_bson_str);                                        \
+   } while (0)
```
kevinAlbs, line 108: Trace logs have come in handy for debugging some monitoring bugs. It seemed worthwhile to add to them for future debugging.

PR #604, opened 2020-05-06
CDRIVER-3303 always apply command options for OP_MSG

rcsanchez97:
The original plan was to eliminate `_mongoc_cmd_parts_ensure_copied` and refactor so that any changes to options along the way are always captured and never lost.  However, that proved to be more risky because of `mongoc_cmd_parts_append_opts` not being idempotent.  As a result, the approach in this PR was taken instead: to add calls to `_mongoc_cmd_parts_ensure_copied` in places where they were missing.  A separate ticket will be written to consider a more comprehensive solution.

### DIFF
--- a/src/libmongoc/tests/test-libmongoc.c
+++ b/src/libmongoc/tests/test-libmongoc.c
@@ -101,6 +101,8 @@ test_client_hedged_reads_install (TestSuite *suite);
 extern void
 test_client_pool_install (TestSuite *suite);
 extern void
+test_client_cmd_install (TestSuite *suite);
+extern void
 test_cluster_install (TestSuite *suite);
 extern void
 test_collection_install (TestSuite *suite);
--- a/src/libmongoc/tests/test-libmongoc.c
+++ b/src/libmongoc/tests/test-libmongoc.c
@@ -2438,6 +2438,7 @@ main (int argc, char *argv[])
    test_client_max_staleness_install (&suite);
    test_client_hedged_reads_install (&suite);
    test_client_pool_install (&suite);
+   test_client_cmd_install (&suite);
    test_write_command_install (&suite);
    test_bulk_install (&suite);
    test_cluster_install (&suite);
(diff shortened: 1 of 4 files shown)


src/libmongoc/tests/test-libmongoc.c
```diff
@@ -2438,6 +2438,7 @@ main (int argc, char *argv[])
    test_client_max_staleness_install (&suite);
    test_client_hedged_reads_install (&suite);
    test_client_pool_install (&suite);
+   test_client_cmd_install (&suite);
```
kevinAlbs, line 2441: Add an extern declaration above on L238, that should fix travis.

src/libmongoc/tests/test-mongoc-cmd.c
```diff
@@ -0,0 +1,86 @@
...
+      NULL,
+      &error);
+
+   request = mock_server_receives_msg (
+      server,
+      MONGOC_QUERY_NONE,
+      tmp_bson ("{'readConcern': { '$exists': true }}"));
+
+   mock_server_replies_simple (request, "{'ok': 1, 'n': 1}");
+   ASSERT_OR_PRINT (1 == future_get_bool (future), error);
```
kevinAlbs, line 68: I'd remove the "1 ==" since this is already a bool return.

src/libmongoc/tests/test-mongoc-cmd.c
```diff
@@ -0,0 +1,86 @@
...
+}
+
+
+void
+test_client_cmd_install (TestSuite *suite)
+{
+   TestSuite_AddMockServerTest (suite,
+                                "/Client/cmd/options",
+                                test_client_cmd_options);
+}
```
kevinAlbs, line 86: I think pedantic requires a newline at the end of the file.

rcsanchez97: That's interesting.  In my editor it shows that there is a newline at the end of the file.

kevinAlbs:
I'm not sure why Travis failed, and there are no attached logs. Otherwise LGTM.

For future reference, I think the workflow once new commits are ready for review is to "Re-request review". That way it will show up again in reviewer's list of "Review Requests".

rcsanchez97:
Interesting.  I wasn't aware of the "Re-request review" thing and
clearly didn't notice it.  I'll keep it in mind for next time.

rcsanchez97:
Merged as commit a4f02b73833b2421098e55dc2ad159ccd97cef5d

PR #605, opened 2020-05-06
CDRIVER-3573 hard code version on patches

kevinAlbs:
Not all evergreen variants have a new enough version of git to run calc_release_version.py on a non-master branch. This fixes patch builds made through GitHub PRs


kevinAlbs:
Looks like this is working. The Ubuntu 14.04 compile task has gotten past cmake.

PR #607, opened 2020-05-07
CDRIVER-3625 monitor with a thread-per-server

kevinAlbs:
Prerequisite work for implementing streamable ismaster in multi-threaded.
Background and rationale is described in WRITING-5411.

For a client pool, servers are now scanned independently using a separate thread per server.
SRV polling is now done on a separate thread.

PR #610, opened 2020-05-12
CDRIVER-3535 add mongoc_interrupt_t

kevinAlbs:
This is prerequisite work for streamable ismaster.

A mongoc_interrupt_t  creates a stream which can be safely interrupted from another thread while polling.

This is needed to support the [isMaster Cancellation](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#ismaster-cancellation) section. When this is integrated with the work of https://github.com/mongodb/mongo-c-driver/pull/607 the interrupt stream will be included in calls to `mongoc_stream_poll`, and will be interrupted when shutting down a server monitor.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-interrupt.c
+++ b/src/libmongoc/src/mongoc/mongoc-interrupt.c
@@ -1,0 +1,307 @@
+/*
+ * Copyright 2020-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "mongoc/mongoc-errno-private.h"
+#include "mongoc/mongoc-interrupt-private.h"
+#include "mongoc/mongoc-log.h"
+#include "mongoc/mongoc-socket-private.h"
+#include "mongoc/mongoc-stream-socket.h"
+#include "mongoc/mongoc-trace-private.h"
+#include "common-thread-private.h"
+
+/* The interrupt stream is implemented in two ways.
+ * On POSIX, this uses the self-pipe trick.
+ * On Windows, this uses a pair of TCP sockets.
+ */
+struct _mongoc_interrupt_t {
+   bson_mutex_t mutex;
+   int pipe_fds[2];
+
+   mongoc_socket_t *read_socket;
+   mongoc_socket_t *write_socket;
+
+   mongoc_stream_t *stream;
+};
+
+mongoc_stream_t *
+_mongoc_interrupt_get_stream (mongoc_interrupt_t *interrupt)
+{
+   return interrupt->stream;
+}
+
+static void
+_log_errno (char *prefix, int _errno)
+{
+   char buf[128] = {0};
+
+   bson_strerror_r (_errno, buf, sizeof (buf));
+   MONGOC_ERROR ("%s: (%d) %s", prefix, _errno, buf);
+}
+
+#ifdef _WIN32
+/* TCP socket pair implementation. */
+mongoc_interrupt_t *
+_mongoc_interrupt_new (uint32_t timeout_ms)
+{
+   mongoc_interrupt_t *interrupt;
+   mongoc_socket_t *listen_socket = NULL;
+   mongoc_socket_t *interrupt_socket = NULL;
+   struct sockaddr_storage server_addr;
+   mongoc_socklen_t sock_len;
+   int ret;
+   bool success = false;
+   struct sockaddr_in server_addr_in = {0};
+
+   ENTRY;
+
+   interrupt = (mongoc_interrupt_t *) bson_malloc0 (sizeof *interrupt);
+   bson_mutex_init (&interrupt->mutex);
+
+   /* Inspired by cpython's implementation of socketpair. */
+   listen_socket = mongoc_socket_new (AF_INET, SOCK_STREAM, 0);
+   if (!listen_socket) {
+      MONGOC_ERROR ("socket creation failed");
+      GOTO (fail);
+   }
+
+   memset (&server_addr_in, 0, sizeof (server_addr_in));
+   server_addr_in.sin_family = AF_INET;
+   server_addr_in.sin_addr.s_addr = htonl (INADDR_LOOPBACK);
+   ret = mongoc_socket_bind (listen_socket,
+                             (struct sockaddr *) &server_addr_in,
+                             sizeof (server_addr_in));
+   if (ret == -1) {
+      _log_errno ("bind failed", mongoc_socket_errno (listen_socket));
+      GOTO (fail);
+   }
+
+   ret = mongoc_socket_listen (listen_socket, 1);
+   if (ret == -1) {
+      _log_errno ("listen failed", mongoc_socket_errno (listen_socket));
+      GOTO (fail);
+   }
+
+   sock_len = sizeof (server_addr);
+   ret = mongoc_socket_getsockname (
+      listen_socket, (struct sockaddr *) &server_addr, &sock_len);
+   if (-1 == ret) {
+      _log_errno ("getsockname failed", mongoc_socket_errno (listen_socket));
+      GOTO (fail);
+   }
+
+   interrupt->read_socket =
+      mongoc_socket_new (server_addr.ss_family, SOCK_STREAM, 0);
+   if (!interrupt->read_socket) {
+      MONGOC_ERROR ("socket creation failed");
+      GOTO (fail);
+   }
+
+   /* Begin non-blocking connect. */
+   ret = mongoc_socket_connect (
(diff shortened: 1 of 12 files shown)


src/libmongoc/src/mongoc/mongoc-interrupt.c
```diff
@@ -0,0 +1,307 @@
...
+#include "mongoc/mongoc-interrupt-private.h"
+#include "mongoc/mongoc-log.h"
+#include "mongoc/mongoc-socket-private.h"
+#include "mongoc/mongoc-stream-socket.h"
+#include "mongoc/mongoc-trace-private.h"
+#include "common-thread-private.h"
+
+/* The interrupt stream is implemented in two ways.
+ * On POSIX, this uses the self-pipe trick.
+ * On Windows, this uses a pair of TCP sockets.
```
kevinAlbs, line 27: The self-pipe trick is one such solution to interrupting calls to poll/select.

Windows does not enable `select` to call on both sockets and non-sockets, so on Windows this uses a pair of TCP sockets. (Which seemed less preferable, as system limits on created sockets could be hit sooner).

A cleaner solution is likely to utilize epoll/kqueue/iocp in our socket primitives functions. This was brought up in the design, and consensus was that that implementing those would require more research (e.g. to see if those are portable enough to support all of our platforms), and would be a riskier change overall.

src/libmongoc/tests/test-libmongoc.c
```diff
@@ -2407,16 +2409,17 @@ windows_exception_handler (EXCEPTION_POINTERS *pExceptionInfo)
...
    char exception_string[128];
    bson_snprintf (exception_string,
-                  sizeof(exception_string),
+                  sizeof (exception_string),
                   (exception_code == EXCEPTION_ACCESS_VIOLATION)
-                  ? "(access violation)"
-                  : "0x%08X", exception_code);
+                     ? "(access violation)"
+                     : "0x%08X",
+                  exception_code);
```
kevinAlbs, line 2416: clang-format

src/libmongoc/src/mongoc/mongoc-interrupt.c
```diff
@@ -0,0 +1,307 @@
...
+
+bool
+_mongoc_interrupt_flush (mongoc_interrupt_t *interrupt)
+{
+   char c;
+   while (true) {
+      if (read (interrupt->pipe_fds[0], &c, 1) == -1) {
+         if (MONGOC_ERRNO_IS_AGAIN (errno)) {
+            /* Nothing left to read. */
+            break;
```
sgolemon, line 266: The break here reads a bit indirectly.  Maybe `return true;` here and an `UNREACHABLE` macro at the end.

kevinAlbs: Done.

src/libmongoc/src/mongoc/mongoc-interrupt.c
```diff
@@ -0,0 +1,307 @@
...
+#include "mongoc/mongoc-trace-private.h"
+#include "common-thread-private.h"
+
+/* The interrupt stream is implemented in two ways.
+ * On POSIX, this uses the self-pipe trick.
+ * On Windows, this uses a pair of TCP sockets.
+ */
+struct _mongoc_interrupt_t {
+   bson_mutex_t mutex;
+   int pipe_fds[2];
```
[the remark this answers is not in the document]
kevinAlbs: Good idea, changed to a union.

kevinAlbs:
Updated. Thanks for the review @sgolemon!

src/libmongoc/src/mongoc/mongoc-interrupt.c
```diff
@@ -0,0 +1,315 @@
...
+bool
+_mongoc_interrupt_flush (mongoc_interrupt_t *interrupt)
+{
+   uint8_t buf[1];
+   while (true) {
+      if (-1 == mongoc_socket_recv (
+                   interrupt->impl.socket_pair.read, buf, sizeof (buf), 0, 0)) {
+         if (MONGOC_ERRNO_IS_AGAIN (errno)) {
+            /* Nothing left to read. */
+            break;
```
bazile-clyde, line 175: Nit: For consistency, why not just return `true` here like the non-windows version?

src/libmongoc/src/mongoc/mongoc-interrupt.c
```diff
@@ -0,0 +1,315 @@
...
+            return true;
+         } else {
+            /* Unexpected error. */
+            MONGOC_ERROR ("failed to read from pipe: %d", errno);
+            return false;
+         }
+      }
+   }
+   /* Should never be reached. */
+   BSON_ASSERT (false);
```
bazile-clyde, line 282: Why not just delete this if it's never reached? I don't _think_ it'd cause any issues based on a quick local test.

src/libmongoc/src/mongoc/mongoc-interrupt.c
```diff
@@ -0,0 +1,315 @@
...
+}
+
+bool
+_mongoc_interrupt_interrupt (mongoc_interrupt_t *interrupt)
+{
+   bson_mutex_lock (&interrupt->mutex);
+   if (write (interrupt->impl.pipe_fds[1], "!", 1) == -1 &&
```
bazile-clyde, line 289: Is the `!` arbitrary?

src/libmongoc/src/mongoc/mongoc-interrupt.c
```diff
@@ -0,0 +1,315 @@
...
+_mongoc_interrupt_destroy (mongoc_interrupt_t *interrupt)
+{
+   if (!interrupt) {
+      return;
+   }
+
+   mongoc_socket_destroy (interrupt->impl.socket_pair.read);
+   mongoc_socket_destroy (interrupt->impl.socket_pair.write);
+   mongoc_stream_destroy (interrupt->stream);
+   bson_mutex_destroy (&interrupt->mutex);
```
bazile-clyde, line 162: Should this be destroyed first like in the non-windows version? That way nothing else can lock this mutex and use the members that are actively being destroyed.

src/libmongoc/src/mongoc/mongoc-interrupt.c
kevinAlbs: Yeah, https://www.man7.org/tlpi/code/online/diff/altio/self_pipe.c.html uses `x`. I also considered 'c' for libmongoc. But I chose '!' because if this character ends up visible anywhere, something definitely went wrong!

src/libmongoc/src/mongoc/mongoc-interrupt.c
```diff
@@ -0,0 +1,315 @@
...
+#else
+/* Pipe implementation. */
+
+/* Set non-blocking and close on exec. */
+static bool
+_set_pipe_flags (int pipe_fd)
+{
+   int flags;
+
+   flags = fcntl (pipe_fd, F_GETFL, pipe_fd);
```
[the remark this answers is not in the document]
kevinAlbs: Good catch. I copied this over from `_mongoc_socket_setflags` and fixed it there too.

src/libmongoc/src/mongoc/mongoc-interrupt.c
```diff
@@ -0,0 +1,315 @@
...
+
+   if (-1 == fcntl (pipe_fd, F_SETFL, (flags | O_NONBLOCK))) {
```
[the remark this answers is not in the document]
kevinAlbs: I was not sure either. Looking at https://stackoverflow.com/a/34943884/774658 led me to https://www.gnu.org/software/libc/manual/html_node/File-Status-Flags.html#File-Status-Flags. It looks like F_SETFL can be used to set other file status flags that are not file access mode or file creation flags (any of the operating modes on that page) which I think we'd want to preserve.

src/libmongoc/src/mongoc/mongoc-interrupt.c
```diff
@@ -0,0 +1,315 @@
...
+/* The interrupt stream is implemented in two ways.
+ * On POSIX, this uses the self-pipe trick.
+ * On Windows, this uses a pair of TCP sockets.
+ */
+struct _mongoc_interrupt_t {
+   bson_mutex_t mutex;
+
+   union {
+      /* For POSIX */
+      int pipe_fds[2];
```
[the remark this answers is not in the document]
kevinAlbs: Done.

src/libmongoc/src/mongoc/mongoc-interrupt.c
kevinAlbs: Ah true. Sara asked for the same change in the non-windows version: https://github.com/mongodb/mongo-c-driver/pull/610#discussion_r425292372

Done.

src/libmongoc/src/mongoc/mongoc-interrupt.c
kevinAlbs: I added this per Sara's comment about adding an `UNREACHABLE` macro. My only slight concern of removing the assertion is that compilers may warn saying this function may not return a value.

src/libmongoc/src/mongoc/mongoc-interrupt.c
kevinAlbs: Done for consistency. Once this is called, it is assumed no thread will access the interrupt again. If a thread locked the mutex after it was destroyed that would be an invalid access.

kevinAlbs:
Pushed in https://github.com/mongodb/mongo-c-driver/commit/40f8e03f8b53150c38311900f7aabfbd21655f62

PR #612, opened 2020-05-15
CDRIVER-3665 End-to-end test for OCSP caching

bazile-clyde:


build/evergreen_config_lib/tasks.py
```diff
@@ -920,17 +920,20 @@ def to_dict(self):
...
+        if self.test in [ 'test_3', 'test_4', 'soft_fail_test', 'cache']:
             stapling = 'disableStapling'
         if self.test in [ 'malicious_server_test_1', 'malicious_server_test_2' ]:
             stapling = 'mustStaple-disableStapling'
 
         orchestration_file = '%s-basic-tls-ocsp-%s' % (self.cert, stapling)
         orchestration = bootstrap(TOPOLOGY='server', SSL='ssl', OCSP='on', ORCHESTRATION_FILE=orchestration_file)
 
-        commands.append(shell_mongoc('TEST_COLUMN=%s CERT_TYPE=%s USE_DELEGATE=%s sh .evergreen/run-ocsp-responder.sh' % (self.test.upper(), self.cert, 'on' if self.delegate == 'delegate' else 'off')))
+        commands.append(shell_mongoc('TEST_COLUMN=%s CERT_TYPE=%s USE_DELEGATE=%s sh .evergreen/run-ocsp-responder.sh' % ('TEST_4' if self.test == 'cache' else self.test.upper(), self.cert, 'on' if self.delegate == 'delegate' else 'off')))
```
kevinAlbs, line 931: The cache test is a bit of a special case. I'd suggest adding an `if` to split up this line, or commenting saying we're using `TEST_4`'s responder configuration for the cache test.

src/libmongoc/tests/test-mongoc-cache.c
```diff
@@ -0,0 +1,89 @@
...
+   ca_file = argv[1];
+
+   mongoc_init ();
+
+   ASSERT (ping () == EXIT_FAILURE);
+   raise (SIGSTOP);
+   ASSERT (ping () == EXIT_FAILURE);
+
+#endif
+   return EXIT_SUCCESS;
```
kevinAlbs, line 88: Call `mongoc_cleanup()` before return. (though it should not make a difference for what is being tested here)

build/evergreen_config_lib/tasks.py
bazile-clyde: Added comment

src/libmongoc/tests/test-mongoc-cache.c
bazile-clyde: Done

.evergreen/run-ocsp-cache-test.sh
```diff
@@ -0,0 +1,70 @@
...
+    exit 1;
+fi
+
+if [ ! `pgrep -nf ocsp_mock` ]; then
+    echo "Cannot find mock OCSP responder. See file comments for help."
+    exit 1;
+fi
+
+# This test will hang after the first ping.
+${CDRIVER_BUILD}/src/libmongoc/test-mongoc-cache ${CDRIVER_ROOT}/.evergreen/ocsp/${CERT_TYPE}/ca.pem &
```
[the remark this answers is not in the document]
bazile-clyde: I'm not sure I understand. I think the first example is more in line with the expected behavior since this line doesn't terminate, i.e., return an exit code, in the background. It just hangs while the rest of the scripts continues. The sleep is just there to make sure that it can contact the OCSP responder before we kill the OCSP responder. But the program is still hanging around. Finally, kill will continue the program and return its exit code. If the test program does exit in the background for whatever reason this program still fails. Although I'm not exactly sure how I'd capture the exit status of a background that may either halt indefinitely or return.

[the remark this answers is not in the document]
kevinAlbs: Ah I was misunderstanding. I was thinking that if it terminated quickly, the script may quietly pass. But in that case the `kill` would fail

PR #616, opened 2020-05-19
CDRIVER-3673: Don't check error message for retryable write errors

alcaeus:
https://jira.mongodb.org/browse/CDRIVER-3673

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-error.c
+++ b/src/libmongoc/src/mongoc/mongoc-error.c
@@ -60,6 +60,17 @@ mongoc_error_has_label (const bson_t *reply, const char *label)
 }
 
 static bool
+_mongoc_error_is_server (bson_error_t *error)
+{
+   if (!error) {
+      return false;
+   }
+
+   return error->domain == MONGOC_ERROR_SERVER ||
+          error->domain == MONGOC_ERROR_WRITE_CONCERN;
+}
+
+static bool
 _mongoc_write_error_is_retryable (bson_error_t *error)
 {
    switch (error->code) {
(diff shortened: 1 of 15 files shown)


src/libmongoc/src/mongoc/mongoc-error.c
```diff
@@ -77,10 +77,6 @@ _mongoc_write_error_is_retryable (bson_error_t *error)
    case MONGOC_SERVER_ERR_NOTMASTERORSECONDARY:
       return true;
    default:
-      if (strstr (error->message, "not master") ||
```
kevinAlbs, line 80: Related to a comment Jeremy had made in the review for CDRIVER-3615: https://github.com/mongodb/mongo-c-driver/pull/601/files#r421624638

It may be worth checking that `error->domain == MONGOC_ERROR_SERVER` or `error->domain == MONGOC_ERROR_WRITE_CONCERN` in this function to assure that this is an error that originated from the server. This function should only be called with an error parsed using MONGOC_ERROR_API_VERSION_2, so checking those domains should be all that is necessary.

src/libmongoc/src/mongoc/mongoc-error.c
```diff
@@ -59,9 +59,24 @@ mongoc_error_has_label (const bson_t *reply, const char *label)
    return false;
 }
 
+static bool
+_mongoc_error_is_server (bson_error_t *error)
```
kevinAlbs, line 63: Nice

alcaeus:
Merged in 20c0aa9a8.

PR #617, opened 2020-05-19
CDRIVER-3680 bump valgrind timeout

kevinAlbs:
From a patch build, it appears they take 3.5 hours to fully execute.

https://evergreen.mongodb.com/version/5ec3d728d1fe073c379213e9

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -1835,7 +1835,7 @@ tasks:
   tags:
   - latest
   - test-valgrind
-  exec_timeout_secs: 7200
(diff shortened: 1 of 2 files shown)


.evergreen/config.yml
```diff
@@ -1835,7 +1835,7 @@ tasks:
   tags:
   - latest
   - test-valgrind
-  exec_timeout_secs: 7200
+  exec_timeout_secs: 14400
```
kevinAlbs, line 1838: Four hours.

PR #619, opened 2020-05-20
CDRIVER-3516 implement waitQueueTimeoutMS for client pools

rcsanchez97:
Full Evergreen patch build: https://evergreen.mongodb.com/version/5ec5886bd6d80a5f707ffba3

### DIFF
--- a/src/libmongoc/doc/connection-pooling.rst
+++ b/src/libmongoc/doc/connection-pooling.rst
@@ -51,4 +51,23 @@ When the driver is in pooled mode, your program's operations are unblocked as so
 
 The pool opens one connection per server for monitoring, and each client opens its own connection to each server it uses for application operations. The background thread re-scans the server topology roughly every 10 seconds. This interval is configurable with ``heartbeatFrequencyMS`` in the connection string. (See :symbol:`mongoc_uri_t`.)
 
+The connection string can also specify ``waitQueueTimeoutMS`` to limit the time that :symbol:`mongoc_client_pool_pop` will wait for a client from the pool.  (See :symbol:`mongoc_uri_t`.)  If ``waitQueueTimeoutMS`` is specified, then it is necessary to confirm that a client was actually returned:
+
+.. code-block:: c
+
+  mongoc_uri_t *uri = mongoc_uri_new (
+     "mongodb://hostA,hostB/?replicaSet=my_rs&waitQueueTimeoutMS=1000");
+
+  mongoc_client_pool_t *pool = mongoc_client_pool_new (uri);
+
+  mongoc_client_t *client = mongoc_client_pool_pop (pool);
+
+  if (client) {
+     /* use the client for operations ... */
+
+     mongoc_client_pool_push (pool, client);
+  } else {
+     /* take appropriate action for a timeout */
+  }
+
 See :ref:`connection_pool_options` to configure pool size and behavior, and see :symbol:`mongoc_client_pool_t` for an extended example of a multi-threaded program that uses the driver in pooled mode.
--- a/src/libmongoc/src/mongoc/mongoc-client-pool.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-pool.c
@@ -249,8 +249,8 @@ mongoc_client_pool_pop (mongoc_client_pool_t *pool)
 {
    mongoc_client_t *client;
    int32_t wait_queue_timeout_ms;
-   int32_t expire_at_ms = -1;
-   int32_t now_ms;
+   int64_t expire_at_ms = -1;
+   int64_t now_ms;
    int r;
 
    ENTRY;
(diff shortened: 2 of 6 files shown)


src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -262,7 +266,16 @@ mongoc_client_pool_pop (mongoc_client_pool_t *pool)
          _initialize_new_client (pool, client);
          pool->size++;
       } else {
-         mongoc_cond_wait (&pool->cond, &pool->mutex);
+         if (wait_queue_timeout_ms > 0) {
+            r = mongoc_cond_timedwait (
```
kevinAlbs, line 270: I interpreted the `waitQueueTimeoutMS` from CMAP to be the maximum total time a thread should wait to obtain a connection before failing (in our case, we're obtaining a client).

I think we need to track the total time that has passed. It's possible for `mongoc_cond_timedwait` to be called multiple times in the process of trying to obtain a client. For example:
- Thread T1 is waiting for a client, and is sleeping from the call to `mongoc_cond_timedwait`
- Thread T2 pushes a client on, signaling the condition variable.
- Before T1 wakes up, thread T3 pops the client.
- Thread T1 again fails to obtain a client, and sleeps again, but for another period of `maxQueueTimeoutMS`.

One way to do this would be have an `expire_at_ms` at the top.

`expire_at_ms = (bson_get_monotonic_time () / 1000) + wait_queue_timeout_ms;`

Then, before sleeping, check if the current time is after the `expire_at_ms`. If not, only sleep for `expire_at_ms - (current time in MS)`.

src/libmongoc/tests/test-mongoc-client-pool.c
```diff
@@ -41,6 +41,32 @@ test_mongoc_client_pool_try_pop (void)
...
+   int64_t start;
+   int64_t duration_usec;
+
+   uri = mongoc_uri_new (
+      "mongodb://127.0.0.1/?maxpoolsize=1&waitqueuetimeoutms=2000");
+   pool = mongoc_client_pool_new (uri);
+   client = mongoc_client_pool_pop (pool);
+   BSON_ASSERT (client);
+   start = bson_get_monotonic_time ();
+   BSON_ASSERT (!mongoc_client_pool_pop (pool));
```
kevinAlbs, line 59: It looks like this is timing out on macOS in the patch build. I'm not certain why, but my best guess is that it is related to what I noted in mongoc-client-pool.c. If somehow the call to `mongoc_cond_timedwait` was returning (i.e. because condition variable was being signaled), and another iteration of the loop happened, it would sleep for another period of `waitQueueTimeoutMS`. Though, that being said I'm not sure why the condition variable would be getting signaled.

If this persists after adding an expiration time, I'd suggest putting in log statements in `mongoc_client_pool_pop` to identify where it is getting stuck.

src/libmongoc/doc/connection-pooling.rst
```diff
@@ -51,4 +51,23 @@ When the driver is in pooled mode, your program's operations are unblocked as so
 
 The pool opens one connection per server for monitoring, and each client opens its own connection to each server it uses for application operations. The background thread re-scans the server topology roughly every 10 seconds. This interval is configurable with ``heartbeatFrequencyMS`` in the connection string. (See :symbol:`mongoc_uri_t`.)
 
+The connection string can also specify ``waitQueueTimeoutMS`` to limit the time that :symbol:`mongoc_client_pool_pop` will wait for a client from the pool.  (See :symbol:`mongoc_uri_t`.)  If ``waitQueueTimeoutMS`` is specified, then it is necessary to confirm that a client was actually returned:
```
kevinAlbs, line 54: Nice, this is good to document.

src/libmongoc/src/mongoc/mongoc-client-pool.c
rcsanchez97: Excellent point.  I will update the PR to implement the tracking of total wait time.

rcsanchez97:
Full patch build to go with second commit: https://evergreen.mongodb.com/version/5ec83af70ae60638153cb506

src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -249,6 +249,8 @@ mongoc_client_pool_pop (mongoc_client_pool_t *pool)
 {
    mongoc_client_t *client;
    int32_t wait_queue_timeout_ms;
+   int32_t expire_at_ms = -1;
+   int32_t now_ms;
```
kevinAlbs, line 253: `bson_get_monotonic_time()` returns an int64_t (and it does not necessarily correspond to the microseconds since epoch) I don't think we can be assured that it fits in an int32_t, even after converting to milliseconds. I'd suggest declaring these as int64_t, and casting the difference as int32_t when passing to `mongoc_cond_timedwait`, which should fit in an int32_t.

As a sidenote, I recently had an overflow issue while working on CDRIVER-3535 caused by me mishandling the result of bson_get_monotonic_time in int32_t.

rcsanchez97: It turns out that `mongoc_cond_timedwait` takes and `int64_t`, so I am not going to cast.  I'll just change the types.

rcsanchez97:
Closed via commit f56a51535

PR #620, opened 2020-05-25
CDRIVER-748 fix mock server for TSAN

kevinAlbs:
This is not urgent. This, and some other test fixes from https://github.com/mongodb/mongo-c-driver/pull/607 are prerequisites for adding a thread-sanitizer task to evergreen to automatically check for thread related bugs.

mock_server_hangs_up and mock_server_resets were closing a request stream directly from the main thread. But a request stream should only be accessed by the worker thread. This now adds a special reply to the worker thread's queue to signal a hangup or reset.

### DIFF
--- a/src/libmongoc/tests/mock_server/mock-server.c
+++ b/src/libmongoc/tests/mock_server/mock-server.c
@@ -1839,6 +1839,10 @@ _mock_server_reply_with_stream (mock_server_t *server,
    uint8_t *ptr;
    size_t len;
    bool is_op_msg;
+   mongoc_reply_flags_t flags = reply->flags;
+   const bson_t *docs = reply->docs;
+   int n_docs = reply->n_docs;
+   int64_t cursor_id = reply->cursor_id;
 
    if (reply->type == HANGUP) {
       mongoc_stream_close (client);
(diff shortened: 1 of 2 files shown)


src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -1218,12 +1223,14 @@ mock_server_receives_kill_cursors (mock_server_t *server, int64_t cursor_id)
...
+   reply_t *reply;
    test_suite_mock_server_log ("%5.2f  %hu <- %hu \thang up!",
                                mock_server_get_uptime_sec (request->server),
                                request->client_port,
                                request_get_server_port (request));
-
-   mongoc_stream_close (request->client);
+   reply = bson_malloc0 (sizeof (reply_t));
+   reply->type = HANGUP;
+   q_put (request->replies, reply);
```
kevinAlbs, line 1233: The mock server spawns a worker thread for every connection. And each worker thread maintains a thread-safe queue for sending replies back.

But the behavior of `mock_server_hangs_up` was to close the server side stream from the main thread. TSan flags a data race because in the process of closing, it overwrites the stream's file descriptor with an invalid one, causing a write/read race.

PR #623, opened 2020-05-29
CDRIVER-3668 support OCSP back to OpenSSL 1.0.1

kevinAlbs:
Adds polyfills and makes a few tweaks to support OpenSSL back to 1.0.1.

Summary of changes:

- change SSL_CTX_set_tlsext_status_type to SSL_set_tlsext_status_type.
- polyfill SSL_get0_verified_chain, NID_tlsfeature, and hostname check.
- check for status_request from the tlsfeature extension when inspecting peer certificate.
- skip time check for older OpenSSL when updating cache entries.
- perform the OCSP check after the handshake, since sometimes the peer certificate is not available in the callback in OpenSSL <= 1.0.2.
- check tlsDisableOCSPEndpointCheck before reaching out to a responder.
- make tlsDisableOCSPEndpointCheck and tlsDisableCertificateRevocationCheck URI options implicitly enable TLS.
- enable OCSP tests on OpenSSL and macOS that were skipped.
- add OCSP tests for OpenSSL 1.0.1.
- update OCSP OpenSSL documentation.
- change OCSP verification logs from MONGOC_DEBUG to TRACE in successful cases.

The first commit has the changes, the second has the config.yml regeneration.

Full evergreen patch build: https://evergreen.mongodb.com/version/5ed03422a4cf471c38658ada

### DIFF
--- a/build/evergreen_config_lib/tasks.py
+++ b/build/evergreen_config_lib/tasks.py
@@ -31,14 +31,15 @@ from evergreen_config_lib import shell_mongoc
 class CompileTask(NamedTask):
     def __init__(self, task_name, tags=None, config='debug',
                  compression='default', continue_on_err=False,
-                 extra_commands=None, depends_on=None,
-                 extra_script=None, **kwargs):
+                 suffix_commands=None, depends_on=None,
+                 extra_script=None, prefix_commands=None, **kwargs):
         super(CompileTask, self).__init__(task_name=task_name,
                                           depends_on=depends_on,
                                           tags=tags,
                                           **kwargs)
 
-        self.extra_commands = extra_commands or []
+        self.suffix_commands = suffix_commands or []
+        self.prefix_commands = prefix_commands or []
         if extra_script:
             self.extra_script = "\n" + extra_script
         else:
@@ -65,6 +66,8 @@ class CompileTask(NamedTask):
     def to_dict(self):
         task = super(CompileTask, self).to_dict()
 
+        task['commands'].extend(self.prefix_commands)
+
         script = ''
         for opt, value in sorted(self.compile_sh_opt.items()):
             script += 'export %s="%s"\n' % (opt, value)
@@ -73,7 +76,7 @@ class CompileTask(NamedTask):
             self.extra_script
         task['commands'].append(shell_mongoc(script))
         task['commands'].append(func('upload build'))
-        task['commands'].extend(self.extra_commands)
+        task['commands'].extend(self.suffix_commands)
         return task
 
 
@@ -135,7 +138,7 @@ class CompileWithClientSideEncryptionAsan(CompileTask):
 
 
 class LinkTask(NamedTask):
-    def __init__(self, task_name, extra_commands, orchestration=True, **kwargs):
+    def __init__(self, task_name, suffix_commands, orchestration=True, **kwargs):
         if orchestration == 'ssl':
             bootstrap_commands = [bootstrap(SSL=1)]
         elif orchestration:
@@ -147,7 +150,7 @@ class LinkTask(NamedTask):
             task_name=task_name,
             depends_on=OD([('name', 'make-release-archive'),
                            ('variant', 'releng')]),
-            commands=bootstrap_commands + extra_commands,
+            commands=bootstrap_commands + suffix_commands,
             **kwargs)
 
 
@@ -201,7 +204,7 @@ all_tasks = [
     SpecialTask('debug-compile-coverage',
                 tags=['debug-compile', 'coverage'],
                 COVERAGE='ON',
-                extra_commands=[func('upload coverage')]),
+                suffix_commands=[func('upload coverage')]),
     CompileTask('debug-compile-no-counters',
                 tags=['debug-compile', 'no-counters'],
                 ENABLE_SHM_COUNTERS='OFF'),
@@ -244,7 +247,7 @@ all_tasks = [
                 continue_on_err=True,
                 ANALYZE='ON',
                 CC='clang',
-                extra_commands=[
+                suffix_commands=[
                     func('upload scan artifacts'),
                     shell_mongoc('''
                         if find scan -name \*.html | grep -q html; then
@@ -300,70 +303,70 @@ all_tasks = [
                 tags=['debug-compile'],
                 SRV='OFF'),
     LinkTask('link-with-cmake',
-             extra_commands=[
+             suffix_commands=[
                  func('link sample program', BUILD_SAMPLE_WITH_CMAKE=1)]),
     LinkTask('link-with-cmake-ssl',
-             extra_commands=[
+             suffix_commands=[
                  func('link sample program',
                       BUILD_SAMPLE_WITH_CMAKE=1,
                       ENABLE_SSL=1)]),
     LinkTask('link-with-cmake-snappy',
-             extra_commands=[
+             suffix_commands=[
                  func('link sample program',
                       BUILD_SAMPLE_WITH_CMAKE=1,
                       ENABLE_SNAPPY=1)]),
     LinkTask('link-with-cmake-mac',
-             extra_commands=[
+             suffix_commands=[
                  func('link sample program', BUILD_SAMPLE_WITH_CMAKE=1)]),
     LinkTask('link-with-cmake-deprecated',
-             extra_commands=[
+             suffix_commands=[
                  func('link sample program',
                       BUILD_SAMPLE_WITH_CMAKE=1,
                       BUILD_SAMPLE_WITH_CMAKE_DEPRECATED=1)]),
     LinkTask('link-with-cmake-ssl-deprecated',
-             extra_commands=[
+             suffix_commands=[
                  func('link sample program',
                       BUILD_SAMPLE_WITH_CMAKE=1,
                       BUILD_SAMPLE_WITH_CMAKE_DEPRECATED=1,
                       ENABLE_SSL=1)]),
     LinkTask('link-with-cmake-snappy-deprecated',
-             extra_commands=[
+             suffix_commands=[
                  func('link sample program',
                       BUILD_SAMPLE_WITH_CMAKE=1,
                       BUILD_SAMPLE_WITH_CMAKE_DEPRECATED=1,
                       ENABLE_SNAPPY=1)]),
     LinkTask('link-with-cmake-mac-deprecated',
-             extra_commands=[
+             suffix_commands=[
                  func('link sample program',
                       BUILD_SAMPLE_WITH_CMAKE=1,
                       BUILD_SAMPLE_WITH_CMAKE_DEPRECATED=1)]),
     LinkTask('link-with-cmake-windows',
-             extra_commands=[func('link sample program MSVC')]),
+             suffix_commands=[func('link sample program MSVC')]),
     LinkTask('link-with-cmake-windows-ssl',
-             extra_commands=[func('link sample program MSVC', ENABLE_SSL=1)],
+             suffix_commands=[func('link sample program MSVC', ENABLE_SSL=1)],
              orchestration='ssl'),
     LinkTask('link-with-cmake-windows-snappy',
-             extra_commands=[
+             suffix_commands=[
                  func('link sample program MSVC', ENABLE_SNAPPY=1)]),
     LinkTask('link-with-cmake-mingw',
-             extra_commands=[func('link sample program mingw')]),
+             suffix_commands=[func('link sample program mingw')]),
     LinkTask('link-with-pkg-config',
-             extra_commands=[func('link sample program')]),
+             suffix_commands=[func('link sample program')]),
     LinkTask('link-with-pkg-config-mac',
-             extra_commands=[func('link sample program')]),
+             suffix_commands=[func('link sample program')]),
     LinkTask('link-with-pkg-config-ssl',
-             extra_commands=[func('link sample program', ENABLE_SSL=1)]),
+             suffix_commands=[func('link sample program', ENABLE_SSL=1)]),
     LinkTask('link-with-bson',
-             extra_commands=[func('link sample program bson')],
+             suffix_commands=[func('link sample program bson')],
              orchestration=False),
     LinkTask('link-with-bson-mac',
-             extra_commands=[func('link sample program bson')],
+             suffix_commands=[func('link sample program bson')],
              orchestration=False),
     LinkTask('link-with-bson-windows',
-             extra_commands=[func('link sample program MSVC bson')],
+             suffix_commands=[func('link sample program MSVC bson')],
              orchestration=False),
     LinkTask('link-with-bson-mingw',
-             extra_commands=[func('link sample program mingw bson')],
+             suffix_commands=[func('link sample program mingw bson')],
              orchestration=False),
     NamedTask('debian-package-build',
               commands=[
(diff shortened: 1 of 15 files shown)


build/evergreen_config_lib/tasks.py
```diff
@@ -31,14 +31,15 @@
...
+                 suffix_commands=None, depends_on=None,
+                 extra_script=None, prefix_commands=None, **kwargs):
         super(CompileTask, self).__init__(task_name=task_name,
                                           depends_on=depends_on,
                                           tags=tags,
                                           **kwargs)
 
-        self.extra_commands = extra_commands or []
+        self.suffix_commands = suffix_commands or []
+        self.prefix_commands = prefix_commands or []
```
kevinAlbs, line 42: I added "prefix_commands" here, so I could create a debug-compile task that used OpenSSL 1.0.1 specifically. That needs to call the "install ssl" evergreen function first.

src/libmongoc/src/mongoc/mongoc-ocsp-cache.c
```diff
@@ -80,6 +80,26 @@ update_entry (cache_entry_list_t *entry,
...
+#else
+static int
+_cmp_time (ASN1_TIME *a, ASN1_TIME *b)
+{
+   /* For older OpenSSL, always report that "a" is before "b". I.e. do not
+    * replace the entry.
+    * If a driver would accept a stapled OCSP response and that response has a
+    * later nextUpdate than the response already in the cache, drivers SHOULD
+    * replace the older entry in the cache with the fresher response. */
+   return -1;
```
kevinAlbs, line 99: I'll admit, this is not ideal. After searching a bit, and checking with Shreyas to see how the server polyfilled the time comparison, it seems like we'd need to parse the string representation of the time to something we can easily compare:
https://github.com/mongodb/mongo/blob/f627ff829ba96487fccb6945d0c2807b1bb51d52/src/mongo/util/net/ssl_manager_openssl.cpp#L375-L412

It doesn't look like it's required by the spec to check `nextUpdate` when updating the cache. By always comparing "a" to be before "b", we won't overwrite cache entries. They'll still get removed when they're detected as being expired by `OCSP_check_validity`.

If this seems ok with you, I'll create a ticket to do this as post-4.4 follow-up.

src/libmongoc/src/mongoc/mongoc-openssl.c
```diff
@@ -84,6 +86,14 @@ _mongoc_openssl_init (void)
       MONGOC_ERROR ("Failed to initialize OpenSSL.");
    }
 
+#ifdef NID_tlsfeature
+   tlsfeature_nid = NID_tlsfeature;
+#else
+   /* TLS versions before 1.1.0 did not define the TLS Feature extension. */
+   tlsfeature_nid =
+      OBJ_create ("1.3.6.1.5.5.7.1.24", "tlsfeature", "TLS Feature");
```
kevinAlbs, line 94: The server does something similar to define the OID when it is not predefined in OpenSSL:
https://github.com/mongodb/mongo/blob/f627ff829ba96487fccb6945d0c2807b1bb51d52/src/mongo/util/net/ssl_manager_openssl.cpp#L357-L361

src/libmongoc/src/mongoc/mongoc-openssl.c
```diff
@@ -477,6 +513,142 @@ _get_issuer (X509 *cert, STACK_OF (X509) * chain)
...
+{
+   return X509_get0_extensions (cert);
+}
+
+#else
+/* Polyfill functionality for pre 1.1.0 OpenSSL */
+
+STACK_OF (X509) * _get_verified_chain (SSL *ssl)
+{
+   X509_STORE *store = NULL;
```
kevinAlbs, line 541: I based this off of what the server doese: https://github.com/mongodb/mongo/blob/f627ff829ba96487fccb6945d0c2807b1bb51d52/src/mongo/util/net/ssl_manager_openssl.cpp#L312-L327

src/libmongoc/src/mongoc/mongoc-openssl.c
```diff
@@ -477,6 +513,142 @@ _get_issuer (X509 *cert, STACK_OF (X509) * chain)
...
+const STACK_OF (X509_EXTENSION) * _get_extensions (const X509 *cert)
+{
+   return cert->cert_info->extensions;
+}
+#endif /* OPENSSL_VERSION_NUMBER */
+
+
+#define TLSFEATURE_STATUS_REQUEST 5
+
+/* Parse just enough of a DER encoded data to check if a SEQUENCE of INTEGER
```
kevinAlbs, line 591: The server has a full DER parser. We only need to parse this one thing, a SEQUENCE of INTEGER.
I used https://docs.microsoft.com/en-us/windows/win32/seccertenroll/about-sequence as a reference for the expected sequence, and this website: http://lapo.it/asn1js/ to interactively view a parsed cert (here is an example of the [mustStaple cert on that website](http://lapo.it/asn1js/#MIIESDCCAzCgAwIBAgIET-rEjTANBgkqhkiG9w0BAQsFADB0MQswCQYDVQQGEwJVUzERMA8GA1UECAwITmV3IFlvcmsxFjAUBgNVBAcMDU5ldyBZb3JrIENpdHkxEDAOBgNVBAoMB01vbmdvREIxDzANBgNVBAsMBktlcm5lbDEXMBUGA1UEAwwOS2VybmVsIFRlc3QgQ0EwHhcNMjAwMjA2MjAxMzExWhcNNDAwMjA4MjAxMzExWjBiMRAwDgYDVQQKDAdNb25nb0RCMQ8wDQYDVQQLDAZLZXJuZWwxEjAQBgNVBAMMCWxvY2FsaG9zdDELMAkGA1UEBhMCVVMxCzAJBgNVBAgMAk5ZMQ8wDQYDVQQHDAZPQ1NQLTEwggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQDYXCFeRfhcbEDMXNb5D5MIO4VzejQFyGwWuWpIlYGTh1cZp_RrBASBg0hdy0TkhqrWaLvTUq-SaCGrP8Yg0nYXUXIwMe34QhcaTMLndMpAgCXjjC-KDNLX4210QE2RjKbf-13d_duf8yQIRxYrbumA2q6-qZf8gsSteDYf_JeWQ6IWJ7ttWyT9XpcnWLucfviSkDwQuhaghYHDeaWafXgIViVXfGtl7ivv-qx1JUfRyAADl_vcd_-Z-ZojpOUN7Li38I2O8qKTQ8JuMV80dgBOwH9gf0Ku9R9HXFqbypDdvE22WTKk6cvksR_tK_QTdOCRZWpmKRcnhDkYXc7fr97vAgMBAAGjgfMwgfAwCQYDVR0TBAIwADALBgNVHQ8EBAMCBaAwHQYDVR0lBBYwFAYIKwYBBQUHAwEGCCsGAQUFBwMCMB0GA1UdDgQWBBQ7I-iGMVTuZpVlzy9dyvGNZ8akOzBpBggrBgEFBQcBAQRdMFswLgYIKwYBBQUHMAGGImh0dHA6Ly9sb2NhbGhvc3Q6OTAwMS9wb3dlci9sZXZlbC8wKQYIKwYBBQUHMAGGHWh0dHA6Ly9sb2NhbGhvc3Q6ODEwMC9zdGF0dXMvMBEGCCsGAQUFBwEYBAUwAwIBBTAaBgNVHREEEzARgglsb2NhbGhvc3SHBH8AAAEwDQYJKoZIhvcNAQELBQADggEBAJ6z1nkMhirDDFlDrkJPX23kAh1kqMTbFu2wnnrARzw6JEFz2Tvhx0-ISBjEdsPK7BhAQ_wpGIFriTIatXILqWhUAsZN7vwg2h-xH41rpbDVXIcGQ8jAx1GSZznxoeqH7arqbBWrV2AzhQofr_rSUtkqlZNTHmPIOvMjpVPqDGaC9nnw0gC5Ow8ZqgAqD_TV6Uym14c_vtxUGgOUCo4JVDsa-x4uuN1t2iPHIcfo37BZwa9J8Bw3mJGkxnJonm1OurReQjaNoT3Zgag9B7q0cKFd4YHXa2vmmFszs2LhJI2khB-OB4tp9g581M_VUl6mAyaahHlEzTlA5HiHFNADOug))

src/libmongoc/src/mongoc/mongoc-openssl.c
```diff
@@ -492,7 +664,7 @@ _contact_ocsp_responder (OCSP_CERTID *id, X509 *peer)
    url_stack = X509_get1_ocsp (peer);
    for (i = 0; i < sk_OPENSSL_STRING_num (url_stack) && !resp; i++) {
       url = sk_OPENSSL_STRING_value (url_stack, i);
-      MONGOC_DEBUG ("Contacting OCSP responder '%s'", url);
+      TRACE ("Contacting OCSP responder '%s'", url);
```
kevinAlbs, line 667: I noticed that running example-client would log messages even on success. Debug logs go to stdout by default if no log handler is set. Though that seems like an odd default, I don't see much precedent for using debug logs for successful log messages, just non-critical errors. So I changed these to TRACE logs, which require the driver to be configured with `-DENABLE_TRACE`.

src/libmongoc/src/mongoc/mongoc-openssl.c
```diff
@@ -613,32 +784,33 @@ _mongoc_ocsp_tlsext_status_cb (SSL *ssl, void *arg)
...
+      TRACE ("%s", "Server does not contain a stapled response");
+      must_staple = _get_must_staple (peer);
       if (must_staple) {
          MONGOC_ERROR ("Server must contain a stapled response");
          ret = OCSP_CB_REVOKED;
          GOTO (done);
       }
 
-      if (!(resp = _contact_ocsp_responder (id, peer))) {
+      if (opts->disable_endpoint_check ||
```
kevinAlbs, line 805: Reach out to a responder unless the user set tlsDisableOCSPEndpointCheck in the URI.

src/libmongoc/src/mongoc/mongoc-stream-tls-openssl.c
```diff
@@ -564,31 +564,25 @@ _mongoc_stream_tls_openssl_handshake (mongoc_stream_t *stream,
...
 
-      peer = SSL_get_peer_certificate (ssl);
-      if (peer && (X509_check_host (peer, host, 0, 0, NULL) == 1 ||
-          X509_check_ip_asc (peer, host, 0) == 1)) {
-         X509_free (peer);
-         RETURN (true);
+#ifdef MONGOC_ENABLE_OCSP_OPENSSL
+      /* Validate OCSP */
+      if (openssl->ocsp_opts &&
+          1 != _mongoc_ocsp_tlsext_status (ssl, openssl->ocsp_opts)) {
```
kevinAlbs, line 572: This is no longer done in the callback. I observed on OpenSSL 1.0.2/1.0.1 that against some servers, the callback would be unable to obtain the peer cert (i.e. `SSL_get_peer_certificate` returned NULL).

So instead, like the hostname check, this now does the OCSP check after the handshake.

FWIW one advantage of the callback is that when the callback returns a failure, that alerts the server the reason for the failure. Doing the OCSP check outside of the handshake means a failure would just close the connection client side. That does not seem like a big loss.

src/libmongoc/tests/test-libmongoc.c
```diff
@@ -244,7 +248,7 @@ extern void
 test_aws_install (TestSuite *suite);
 extern void
 test_streamable_ismaster_install (TestSuite *suite);
-#ifdef MONGOC_ENABLE_OCSP_OPENSSL
+#if defined(MONGOC_ENABLE_OCSP_OPENSSL) && OPENSSL_VERSION_NUMBER >= 0x10101000L
 extern void
 test_ocsp_cache_install (TestSuite *suite);
```
kevinAlbs, line 253: Because of the behavior change with the time, I left the cache tests limited to 1.1.1. I'd like to add these tests back as future work (along with the time polyfill).

src/libmongoc/src/mongoc/mongoc-openssl.c
```diff
@@ -613,32 +784,33 @@ _mongoc_ocsp_tlsext_status_cb (SSL *ssl, void *arg)
...
       /* obtain an OCSP_RESPONSE object from the OCSP response */
-      if (!d2i_OCSP_RESPONSE (&resp, &r, len)) {
+      if (!d2i_OCSP_RESPONSE (&resp, &resp_data, len)) {
          MONGOC_ERROR ("Failed to parse OCSP response");
          ret = OCSP_CB_ERROR;
          GOTO (done);
       }
    } else {
-      MONGOC_DEBUG ("Server does not contain a stapled response");
-      must_staple = X509_get_ext_d2i (peer, NID_tlsfeature, 0, 0) != NULL;
```
kevinAlbs, line 627: This previously checked for the existence of the tlsfeature extension. But after looking at the server code I realized this should be looking for the status_request extension within the tlsfeature extension:
https://github.com/mongodb/mongo/blob/4dcdc501520558d602897aed9aa61781265328d7/src/mongo/util/net/ssl_manager_openssl.cpp#L1605-L1616

pymongo does the same:
https://github.com/mongodb/mongo-python-driver/blob/29960237dcdd9d90402d6c76c4296aec3d01d3fc/pymongo/ocsp_support.py#L292-L298

src/libmongoc/src/mongoc/mongoc-openssl.c
```diff
@@ -477,6 +513,142 @@ _get_issuer (X509 *cert, STACK_OF (X509) * chain)
...
+   /* Expect a sequence type, with a sequence length representable in one byte.
+    */
+   if (length < 3 || data[0] != 0x30 || data[1] >= 127) {
+      MONGOC_ERROR ("malformed tlsfeature extension sequence");
+      return false;
+   }
+
+   for (i = 2; i < length; i += 3) {
+      /* Expect an integer, representable in one byte. */
+      if (length < i + 3 || data[i] != 0x02 || data[i + 1] >= 127) {
```
markbenvenuto, line 610: You need to validate the length of the INTEGER is 1 byte otherwise your i+3 logic will start reading the wrong data.

Imagine if the first INTEGER was 2 bytes instead of 1. Your iterator logic would be completely broken.

The JS viewer is pretty cool. Here is my ASN.1 reference - https://www.itu.int/rec/T-REC-X.690-201508-I

kevinAlbs: Ah, good catch! I was checking for a multi-byte length, but not a multi-byte integer. And thanks for the reference.

This was also ripe for unit testing, so I added tests.

src/libmongoc/src/mongoc/mongoc-openssl.c
```diff
@@ -477,6 +513,144 @@ _get_issuer (X509 *cert, STACK_OF (X509) * chain)
...
+ * representable in one byte, and that each integer is representable in one
+ * byte. */
+bool
+_mongoc_tlsfeature_has_status_request (const uint8_t *data, int length)
+{
+   int i;
+
+   /* Expect a sequence type, with a sequence length representable in one byte.
+    */
+   if (length < 3 || data[0] != 0x30 || data[1] >= 127) {
```
bazile-clyde, line 605: What do the numbers `0x30` and `127` mean?

src/libmongoc/src/mongoc/mongoc-openssl.c
```diff
@@ -477,6 +513,144 @@ _get_issuer (X509 *cert, STACK_OF (X509) * chain)
...
+      MONGOC_ERROR ("malformed tlsfeature extension sequence");
+      return false;
+   }
+
+   for (i = 2; i < length; i += 3) {
+      /* Expect an integer, representable in one byte. */
+      if (length < i + 3 || data[i] != 0x02 || data[i + 1] != 1) {
```
bazile-clyde, line 612: I'm not exactly sure what's going on here and in the for loop above.

src/libmongoc/src/mongoc/mongoc-ocsp-cache.c
kevinAlbs: Created https://jira.mongodb.org/browse/CDRIVER-3707

PR #625, opened 2020-05-31
CDRIVER-3535 streamable ismaster

kevinAlbs:
Implementation of streamable ismaster. Integration tests and unit tests are on another branch and passing. I left them out of this PR to keep it a little smaller.

The pseudocode in the Server Monitoring spec captures the core logic: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#monitor-thread but I've commented inline pointing to relevant sections.

### DIFF
--- a/.evergreen/compile-unix.sh
+++ b/.evergreen/compile-unix.sh
@@ -256,6 +256,8 @@ else
    }
 fi
 
+export MONGOC_TEST_SERVER_LOG=stdout
+
 # Write stderr to error.log and to console. Turn off tracing to avoid spurious
 # log messages that CHECK_LOG considers failures.
 mkfifo pipe || true
--- a/.evergreen/debug-core-evergreen.sh
+++ b/.evergreen/debug-core-evergreen.sh
@@ -1,5 +1,14 @@
 #!/usr/bin/env bash
 
+OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+
+case "$OS" in
+   cygwin*)
+      echo "debug-core-evergreen.sh is not supported on Windows"
+      exit 0
+   ;;
+esac
+
 echo "Debugging core files"
 
 shopt -s nullglob
@@ -7,3 +16,12 @@ for i in *.core; do
    echo $i
    echo "backtrace full" | gdb -q ./src/libmongoc/test-libmongoc $i
 done
+
+# If there is still a test-libmongoc process running (perhaps due to
+# deadlock, or very slow test) attach a debugger and print stacks.
+TEST_LIBMONGOC_PID="$(pgrep test-libmongoc)"
+if [ -n "$TEST_LIBMONGOC_PID" ]; then
+   echo "test-libmongoc processes still running with PID=$TEST_LIBMONGOC_PID"
+   echo "backtrace full" | gdb -q -p $TEST_LIBMONGOC_PID
+   kill $TEST_LIBMONGOC_PID
+fi
--- a/.evergreen/run-tests.sh
+++ b/.evergreen/run-tests.sh
@@ -87,6 +87,8 @@ check_mongocryptd() {
    fi
 }
 
+export MONGOC_TEST_MONITORING_VERBOSE=on
+
 case "$OS" in
    cygwin*)
       export PATH=$PATH:/cygdrive/c/mongodb/bin:/cygdrive/c/libmongocrypt/bin
--- a/src/libmongoc/src/mongoc/mongoc-client.c
+++ b/src/libmongoc/src/mongoc/mongoc-client.c
@@ -957,13 +957,8 @@ _mongoc_client_recv (mongoc_client_t *client,
    BSON_ASSERT (buffer);
    BSON_ASSERT (server_stream);
 
-   if (!mongoc_cluster_try_recv (
-          &client->cluster, rpc, buffer, server_stream, error)) {
-      mongoc_topology_invalidate_server (
-         client->topology, server_stream->sd->id, error);
-      return false;
-   }
-   return true;
+   return mongoc_cluster_try_recv (
+      &client->cluster, rpc, buffer, server_stream, error);
 }
 
 
--- a/src/libmongoc/src/mongoc/mongoc-cluster.c
+++ b/src/libmongoc/src/mongoc/mongoc-cluster.c
@@ -25,6 +25,7 @@
 #include "mongoc-counters-private.h"
 #include "mongoc-config.h"
 #include "mongoc-error.h"
+#include "mongoc-flags-private.h"
 #include "mongoc-host-list-private.h"
 #include "mongoc-log.h"
 #include "mongoc-cluster-sasl-private.h"
@@ -42,6 +43,7 @@
 #include "mongoc-stream-tls.h"
 #include "mongoc-thread-private.h"
 #include "mongoc-topology-private.h"
+#include "mongoc-topology-background-monitoring-private.h"
 #include "mongoc-trace-private.h"
 #include "mongoc-util-private.h"
 #include "mongoc-write-concern-private.h"
@@ -61,22 +63,6 @@
 #define CHECK_CLOSED_DURATION_MSEC 1000
 
 #define IS_NOT_COMMAND(_name) (!!strcasecmp (cmd->command_name, _name))
-
-/**
- * mongoc_op_msg_flags_t:
- * @MONGOC_MSG_CHECKSUM_PRESENT: The message ends with 4 bytes containing a
- * CRC-32C checksum.
- * @MONGOC_MSG_MORE_TO_COME: If set to 0, wait for a server response. If set to
- * 1, do not expect a server response.
- * @MONGOC_MSG_EXHAUST_ALLOWED: If set, allows multiple replies to this request
- * using the moreToCome bit.
- */
-typedef enum {
-   MONGOC_MSG_NONE = 0,
-   MONGOC_MSG_CHECKSUM_PRESENT = 1 << 0,
-   MONGOC_MSG_MORE_TO_COME = 1 << 1,
-   MONGOC_EXHAUST_ALLOWED = 1 << 16,
-} mongoc_op_msg_flags_t;
 
 static mongoc_server_stream_t *
 mongoc_cluster_fetch_stream_single (mongoc_cluster_t *cluster,
@@ -2238,13 +2224,18 @@ _mongoc_cluster_stream_for_server (mongoc_cluster_t *cluster,
        * 3. Auth error during handshake.
        * Only (1) should mark the server unknown and clear the pool.
        * Network errors should be checked at a lower layer than this, when an
-       * operation on a stream fails.
+       * operation on a stream fails, and should take the connection generation
+       * into account.
        */
 
       mongoc_topology_invalidate_server (topology, server_id, err_ptr);
       mongoc_cluster_disconnect_node (cluster, server_id);
       bson_mutex_lock (&topology->mutex);
       _mongoc_topology_clear_connection_pool (topology, server_id);
+      if (!topology->single_threaded) {
+         _mongoc_topology_background_monitoring_cancel_check (topology,
+                                                              server_id);
+      }
       bson_mutex_unlock (&topology->mutex);
       _mongoc_bson_init_with_transient_txn_error (cs, reply);
    }
--- a/src/libmongoc/src/mongoc/mongoc-server-description.c
+++ b/src/libmongoc/src/mongoc/mongoc-server-description.c
@@ -120,7 +120,7 @@ mongoc_server_description_init (mongoc_server_description_t *sd,
 
    sd->id = id;
    sd->type = MONGOC_SERVER_UNKNOWN;
-   sd->round_trip_time_msec = -1;
+   sd->round_trip_time_msec = MONGOC_RTT_UNSET;
 
    if (!_mongoc_host_list_from_string (&sd->host, address)) {
       MONGOC_WARNING ("Failed to parse uri for %s", address);
@@ -457,13 +457,18 @@ mongoc_server_description_set_election_id (
  * Side effects:
  *       None.
  *
+ * If rtt_msec is MONGOC_RTT_UNSET, the value is not updated.
+ *
  *-------------------------------------------------------------------------
  */
 void
 mongoc_server_description_update_rtt (mongoc_server_description_t *server,
                                       int64_t rtt_msec)
 {
-   if (server->round_trip_time_msec == -1) {
+   if (rtt_msec == MONGOC_RTT_UNSET) {
+      return;
+   }
+   if (server->round_trip_time_msec == MONGOC_RTT_UNSET) {
       server->round_trip_time_msec = rtt_msec;
    } else {
       server->round_trip_time_msec = (int64_t) (
@@ -487,7 +492,7 @@ _mongoc_server_description_set_error (mongoc_server_description_t *sd,
 
    /* Server Discovery and Monitoring Spec: if the server type changes from a
     * known type to Unknown its RTT is set to null. */
-   sd->round_trip_time_msec = -1;
+   sd->round_trip_time_msec = MONGOC_RTT_UNSET;
 }
 
 
@@ -549,6 +554,9 @@ mongoc_server_description_handle_ismaster (mongoc_server_description_t *sd,
              * domain will be overwritten. */
             (void) _mongoc_cmd_check_ok (
                ismaster_response, MONGOC_ERROR_API_VERSION_2, &sd->error);
+            /* TODO CDRIVER-3696: this is an existing bug. If this is handling
+             * an ismaster reply that is NOT from a handshake, this should not
+             * be considered an auth error. */
             /* ismaster response returned ok: 0. According to auth spec: "If the
              * isMaster of the MongoDB Handshake fails with an error, drivers
              * MUST treat this an authentication error." */
@@ -718,7 +726,7 @@ mongoc_server_description_handle_ismaster (mongoc_server_description_t *sd,
 
 failure:
    sd->type = MONGOC_SERVER_UNKNOWN;
-   sd->round_trip_time_msec = -1;
+   sd->round_trip_time_msec = MONGOC_RTT_UNSET;
 
    EXIT;
 }
@@ -747,7 +755,7 @@ mongoc_server_description_new_copy (
    copy->id = description->id;
    copy->opened = description->opened;
    memcpy (&copy->host, &description->host, sizeof (copy->host));
-   copy->round_trip_time_msec = -1;
+   copy->round_trip_time_msec = MONGOC_RTT_UNSET;
 
    copy->connection_address = copy->host.host_and_port;
    bson_init (&copy->last_is_master);
@@ -1120,6 +1128,14 @@ _mongoc_server_description_equal (mongoc_server_description_t *sd1,
    }
 
    if (sd1->session_timeout_minutes != sd2->session_timeout_minutes) {
+      return false;
+   }
+
+   if (0 != memcmp (&sd1->error, &sd2->error, sizeof (bson_error_t))) {
+      return false;
+   }
+
+   if (!bson_equal (&sd1->topology_version, &sd2->topology_version)) {
       return false;
    }
 
--- a/src/libmongoc/src/mongoc/mongoc-server-monitor-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-server-monitor-private.h
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2020-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "mongoc-prelude.h"
+
+#ifndef MONGOC_SERVER_MONITOR_PRIVATE_H
+#define MONGOC_SERVER_MONITOR_PRIVATE_H
+
+#include "mongoc.h"
+#include "mongoc-server-description-private.h"
+#include "mongoc-topology-private.h"
+
+/* For background monitoring of a single server. */
+
+typedef struct _mongoc_server_monitor_t mongoc_server_monitor_t;
+
+mongoc_server_monitor_t *
+mongoc_server_monitor_new (mongoc_topology_t *topology,
+                           mongoc_server_description_t *init_description);
+
+mongoc_server_description_t *
+mongoc_server_monitor_check_server (
+   mongoc_server_monitor_t *server_monitor,
+   const mongoc_server_description_t *previous_description,
+   bool *cancelled);
+
+void
+mongoc_server_monitor_request_cancel (mongoc_server_monitor_t *server_monitor);
+
+void
+mongoc_server_monitor_request_scan (mongoc_server_monitor_t *server_monitor);
+
+bool
+mongoc_server_monitor_request_shutdown (
+   mongoc_server_monitor_t *server_monitor);
+
+void
+mongoc_server_monitor_wait_for_shutdown (
+   mongoc_server_monitor_t *server_monitor);
+
+void
+mongoc_server_monitor_destroy (mongoc_server_monitor_t *server_monitor);
+
+void
+mongoc_server_monitor_run (mongoc_server_monitor_t *server_monitor);
+
+void
+mongoc_server_monitor_run_as_rtt (mongoc_server_monitor_t *server_monitor);
+
+#endif /* MONGOC_SERVER_MONITOR_PRIVATE_H */
--- a/src/libmongoc/src/mongoc/mongoc-server-monitor.c
+++ b/src/libmongoc/src/mongoc/mongoc-server-monitor.c
@@ -1,0 +1,1267 @@
+/*
+ * Copyright 2020-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "common-thread-private.h"
+#include "mongoc-server-monitor-private.h"
+
+#include "mongoc/mongoc-client-private.h"
+#include "mongoc/mongoc-error-private.h"
+#include "mongoc/mongoc-flags-private.h"
+#include "mongoc/mongoc-ssl-private.h"
+#include "mongoc/mongoc-stream-private.h"
+#include "mongoc/mongoc-topology-background-monitoring-private.h"
+#include "mongoc/mongoc-topology-private.h"
+#include "mongoc/mongoc-trace-private.h"
+
+#undef MONGOC_LOG_DOMAIN
+#define MONGOC_LOG_DOMAIN "monitor"
+
+typedef enum {
+   MONGOC_THREAD_OFF = 0,
+   MONGOC_THREAD_RUNNING,
+   MONGOC_THREAD_SHUTTING_DOWN,
+   MONGOC_THREAD_JOINABLE
+} thread_state_t;
+
+/* Use a signed and wide return type for timeouts as long as you can. Cast only
+ * when you know what you're doing with it. */
+static int64_t
+_now_us (void)
+{
+   return bson_get_monotonic_time ();
+}
+
+static int64_t
+_now_ms (void)
+{
+   return _now_us () / 1000;
+}
+
+struct _mongoc_server_monitor_t {
+   mongoc_topology_t *topology;
+   bson_thread_t thread;
+
+   /* State accessed from multiple threads. */
+   struct {
+      bson_mutex_t mutex;
+      mongoc_cond_t cond;
+      thread_state_t state;
+      bool scan_requested;
+      bool cancel_requested;
+   } shared;
+
+   /* Default time to sleep between ismaster checks (reduced when a scan is
+    * requested) */
+   uint64_t heartbeat_frequency_ms;
+   /* The minimum time to sleep between ismaster checks. */
+   uint64_t min_heartbeat_frequency_ms;
+   int64_t connect_timeout_ms;
+   bool use_tls;
+#ifdef MONGOC_ENABLE_SSL
+   mongoc_ssl_opt_t *ssl_opts;
+#endif
+   mongoc_uri_t *uri;
+   /* A custom initiator may be set if a user provides overrides to create a
+    * stream. */
+   mongoc_stream_initiator_t initiator;
+   void *initiator_context;
+   int64_t request_id;
+   mongoc_apm_callbacks_t apm_callbacks;
+   void *apm_context;
+
+   mongoc_stream_t *stream;
+   bool more_to_come;
+   mongoc_server_description_t *description;
+   uint32_t server_id;
+   bool is_rtt;
+};
+
+static BSON_GNUC_PRINTF (3, 4) void _server_monitor_log (
+   mongoc_server_monitor_t *server_monitor,
+   mongoc_log_level_t level,
+   const char *format,
+   ...)
+{
+   va_list ap;
+   char *msg;
+
+   va_start (ap, format);
+   msg = bson_strdupv_printf (format, ap);
+   va_end (ap);
+
+   mongoc_log (level,
+               MONGOC_LOG_DOMAIN,
+               "[%s%s] %s",
+               server_monitor->description->host.host_and_port,
+               server_monitor->is_rtt ? "-RTT" : "",
+               msg);
+   bson_free (msg);
+}
+
+#ifdef MONGOC_TRACE
+#define MONITOR_LOG(sm, ...) \
+   _server_monitor_log (sm, MONGOC_LOG_LEVEL_TRACE, __VA_ARGS__)
+#else
+#define MONITOR_LOG(sm, ...) (void) 0
+#endif
+
+/* TODO CDRIVER-3710 use MONGOC_LOG_LEVEL_ERROR */
+#define MONITOR_LOG_ERROR(sm, ...) \
+   _server_monitor_log (sm, MONGOC_LOG_LEVEL_DEBUG, __VA_ARGS__)
+/* TODO CDRIVER-3710 use MONGOC_LOG_LEVEL_WARNING */
+#define MONITOR_LOG_WARNING(sm, ...) \
+   _server_monitor_log (sm, MONGOC_LOG_LEVEL_DEBUG, __VA_ARGS__)
+
+/* Called only from server monitor thread.
+ * Caller must hold no locks (user's callback may lock topology mutex).
+ * Locks APM mutex.
+ */
+static void
+_server_monitor_heartbeat_started (mongoc_server_monitor_t *server_monitor,
+                                   bool awaited)
+{
+   mongoc_apm_server_heartbeat_started_t event;
+
+   if (!server_monitor->apm_callbacks.server_heartbeat_started) {
+      return;
+   }
+
+   event.host = &server_monitor->description->host;
+   event.context = server_monitor->apm_context;
+   MONITOR_LOG (server_monitor,
+                "%s heartbeat started",
+                awaited ? "awaitable" : "regular");
+   event.awaited = awaited;
+   bson_mutex_lock (&server_monitor->topology->apm_mutex);
+   server_monitor->apm_callbacks.server_heartbeat_started (&event);
+   bson_mutex_unlock (&server_monitor->topology->apm_mutex);
+}
+
+/* Called only from server monitor thread.
+ * Caller must hold no locks (user's callback may lock topology mutex).
+ * Locks APM mutex.
+ */
+static void
+_server_monitor_heartbeat_succeeded (mongoc_server_monitor_t *server_monitor,
+                                     const bson_t *reply,
+                                     int64_t duration_usec,
+                                     bool awaited)
+{
+   mongoc_apm_server_heartbeat_succeeded_t event;
+
+   if (!server_monitor->apm_callbacks.server_heartbeat_succeeded) {
+      return;
+   }
+
+   event.host = &server_monitor->description->host;
+   event.context = server_monitor->apm_context;
+   event.reply = reply;
+   event.duration_usec = duration_usec;
+   MONITOR_LOG (server_monitor,
+                "%s heartbeat succeeded",
+                awaited ? "awaitable" : "regular");
+   event.awaited = awaited;
+   bson_mutex_lock (&server_monitor->topology->apm_mutex);
+   server_monitor->apm_callbacks.server_heartbeat_succeeded (&event);
+   bson_mutex_unlock (&server_monitor->topology->apm_mutex);
+}
+
+/* Called only from server monitor thread.
+ * Caller must hold no locks (user's callback may lock topology mutex).
+ * Locks APM mutex.
+ */
+static void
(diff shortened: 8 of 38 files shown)


.evergreen/compile-unix.sh
```diff
@@ -256,6 +256,8 @@ else
    }
 fi
 
+export MONGOC_TEST_SERVER_LOG=stdout
```
kevinAlbs, line 259: This logs commands and replies for the mock server. I left it in since it could be helpful in diagnosing flaky test failures.

.evergreen/debug-core-evergreen.sh
```diff
@@ -7,3 +7,11 @@ for i in *.core; do
...
    echo "backtrace full" | gdb -q ./src/libmongoc/test-libmongoc $i
 done
+
+# If there is still a test-libmongoc process running (perhaps due to
+# deadlock, or very slow test) attach a debugger and print stacks.
+TEST_LIBMONGOC_PID="$(pgrep test-libmongoc)"
+if [ -n "$TEST_LIBMONGOC_PID" ]; then
+   echo "test-libmongoc processes still running with PID=$TEST_LIBMONGOC_PID"
+   echo "backtrace full" | gdb -q -p $TEST_LIBMONGOC_PID
+fi
```
kevinAlbs, line 17: This hasn't been used on evergreen, but I'm adding this as a "just-in-case" a deadlock appears. I did introduce a deadlock locally in the integration tests (by joining on a thread while holding a lock that the thread needed to proceed) and attaching lldb and printing every threads stack trace helped me diagnose.

.evergreen/run-tests.sh
```diff
@@ -87,6 +87,8 @@ check_mongocryptd() {
    fi
 }
 
+export MONGOC_TEST_MONITORING_VERBOSE=on
```
kevinAlbs, line 90: Prints more debug logs for the JSON spec test runner.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -957,13 +957,8 @@ _mongoc_client_recv (mongoc_client_t *client,
...
    BSON_ASSERT (server_stream);
 
-   if (!mongoc_cluster_try_recv (
-          &client->cluster, rpc, buffer, server_stream, error)) {
-      mongoc_topology_invalidate_server (
-         client->topology, server_stream->sd->id, error);
-      return false;
-   }
-   return true;
+   return mongoc_cluster_try_recv (
```
kevinAlbs, line 960: `mongoc_topology_invalidate_server` marks a server description as Unknown.
I don't think it should be called in isolation, since there is other behavior that needs to happen when marking a server Unknown due to a network error (generation check, clear connection pool, and cancelling the streamable ismaster). Invalidating a server on a network error without the other logic would be a bug.

Fortunately `mongoc_cluster_recv` already does error handling on a network error... except for failure to decompress. But that behavior is already inconsistent (https://jira.mongodb.org/browse/CDRIVER-3694) and I'm also not positive a decompression failure is a network error.

I added at least a basic test to ensure that a hangup on `_mongoc_client_recv` does the right error handline.

src/libmongoc/src/mongoc/mongoc-server-description.c
```diff
@@ -1123,6 +1131,14 @@ _mongoc_server_description_equal (mongoc_server_description_t *sd1,
...
    }
 
+   if (0 != memcmp (&sd1->error, &sd2->error, sizeof (bson_error_t))) {
+      return false;
+   }
+
+   if (!bson_equal (&sd1->topology_version, &sd2->topology_version)) {
+      return false;
+   }
+
```
kevinAlbs, line 1141: The equality comparison for a server description is specified in SDAM:

https://github.com/mongodb/specifications/blob/a25c50976807b4db16f85fd8d5c5b0337cd29cf9/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#serverdescription

src/libmongoc/src/mongoc/mongoc-server-monitor-private.h
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+#include "mongoc-prelude.h"
+
+#ifndef MONGOC_SERVER_MONITOR_PRIVATE_H
+#define MONGOC_SERVER_MONITOR_PRIVATE_H
+
+#include "mongoc.h"
+
+/* For background monitoring of a single server. */
```
kevinAlbs, line 24: I've split off the logic for monitoring a single server from mongoc-topology-background-monitoring.c into this file.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1318 @@
...
+
+   if (!server_monitor->apm_callbacks.server_heartbeat_started) {
+      return;
+   }
+
+   event.host = &server_monitor->description->host;
+   event.context = server_monitor->apm_context;
+   MONITOR_LOG (server_monitor, "%s heartbeat started", awaited ? "awaitable" : "regular");
+   /* TODO: once awaitable support on heartbeat events is merged. */
+   /* event.awaited = awaited; */
```
kevinAlbs, line 141: After https://github.com/mongodb/mongo-c-driver/pull/621 is merged, this will be rebased, uncommented, and tested.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1318 @@
...
+      server_monitor, &ismaster_reply, &error);
+
+exit:
+   duration_us = _now_us () - start_us;
+   MONITOR_LOG (
+      server_monitor, "server check duration (us): %" PRId64, duration_us);
+
+   if (ret) {
+      int64_t rtt_ms;
+      char *reply_str;
```
kevinAlbs, line 910: Ah, some debugging leftover. Will remove.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1318 @@
...
+
+      reply_str = bson_as_json (&ismaster_reply, NULL);
+      /* Got a reply to ismaster.  Note, rtt_ms is 0 for awaitable ismaster,
```
kevinAlbs, line 913: Bad comment, rtt_ms is -1 for awaitable ismaster (to indicate the lack of a usable RTT).

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1318 @@
...
+{
+   bson_mutex_lock (&server_monitor->shared.mutex);
+   if (server_monitor->shared.state == MONGOC_THREAD_OFF) {
+      server_monitor->is_rtt = true;
+      server_monitor->shared.state = MONGOC_THREAD_RUNNING;
+      COMMON_PREFIX (thread_create)
+      (&server_monitor->thread, _server_monitor_rtt_thread, server_monitor);
+   }
+   bson_mutex_unlock (&server_monitor->shared.mutex);
+}
```
kevinAlbs, line 1231: I reused a lot of the logic for the RTT thread, which differs from the pseudocode proposed in https://github.com/mongodb/specifications/pull/807

There's a discussion between Shane and I here explaining why this diverges: https://github.com/mongodb/specifications/pull/807#discussion_r427647840

The pseudocode assumes ServerDescriptions are immutable, and that they are constructed with the running average RTT. In libmongoc, server descriptions hold the running average, and it is updated in place.

src/libmongoc/src/mongoc/mongoc-topology-description.c
```diff
@@ -1132,7 +1132,7 @@ mongoc_topology_description_invalidate_server (
    BSON_ASSERT (error);
 
    /* send NULL ismaster reply */
-   mongoc_topology_description_handle_ismaster (topology, id, NULL, 0, error);
+   mongoc_topology_description_handle_ismaster (topology, id, NULL, -1, error);
```
kevinAlbs, line 1135: This was not a bug before. Because error was always set, so the RTT was never updated. But I think -1 should consistently be the "invalid RTT" indicator.

src/libmongoc/src/mongoc/mongoc-topology-description.c
```diff
@@ -1955,6 +1955,7 @@ mongoc_topology_description_handle_ismaster (
 
       if (mongoc_server_description_topology_version_cmp (
              &sd->topology_version, &incoming_topology_version) == 1) {
+         MONGOC_DEBUG ("topology version is strictly less. Skipping.");
```
kevinAlbs, line 1958: This should be a TRACE. Will fix.

src/libmongoc/tests/test-mongoc-background-monitoring.c
```diff
@@ -166,6 +168,26 @@ _server_changed (const mongoc_apm_server_changed_t *event)
    bson_mutex_unlock (&tf->mutex);
 }
 
+#define TV \
+   "{ 'processId': { '$oid': 'AABBAABBAABBAABBAABBAABB' }, 'counter': 1 }"
+
+bool
+auto_respond_polling_ismaster (request_t *request, void *ctx)
```
kevinAlbs, line 175: This auto responder is also used in responding to RTT ismasters for the new streaming tests.

src/libmongoc/tests/test-mongoc-cluster.c
```diff
@@ -145,7 +147,6 @@ static void
 _test_cluster_node_disconnect (bool pooled)
 {
    mock_server_t *server;
-   const int32_t socket_timeout_ms = 100;
    mongoc_uri_t *uri;
```
kevinAlbs, line 150: I observed one failure of the mock server receiving a query in START_QUERY (but only once). My best guess was that sending the message did not succeed in 100ms. But inspecting this test, I don't think anything here relies on socket_timeout_ms being short (nor does it speed up the test).

src/libmongoc/tests/test-mongoc-primary-stepdown.c
```diff
@@ -54,9 +56,8 @@ _setup_test_with_client (mongoc_client_t *client)
 }
 
 static int
-_connection_count (mongoc_database_t *db)
+_connection_count (mongoc_client_t *client, uint32_t server_id)
```
kevinAlbs, line 59: This was a fun bug. These tests check the connection count that a server reports to make assertions about whether connections are created or not. But, the checks must happen on the same server. It was relying on a primary read preference to select the same server after failover. But the primary read preference would only select the same server if monitoring had not yet discovered the newly changed primary. But with streamable ismaster, the failover is handled immediately and the new primary discovered, making the new connection count incorrect.

I strongly suspect this is also why this test flakes: https://jira.mongodb.org/browse/CDRIVER-3569

Fortunately Matt observed the same behavior in Node's tests and he brought this up in the drivers-streamable-ismaster channel.

src/libmongoc/tests/test-mongoc-sdam-monitoring.c
```diff
@@ -825,19 +828,16 @@ _test_heartbeat_fails_dns (bool pooled)
       client_set_heartbeat_event_callbacks (client, &context);
    }
 
-   start = bson_get_monotonic_time ();
-
    /* trigger "ismaster" handshake */
    r = mongoc_client_command_simple (
       client, "admin", tmp_bson ("{'foo': 1}"), NULL, NULL, &error);
 
+   /* This should result in either a DNS failure or connection failure depending on the network. */
```
kevinAlbs, line 835: This is not due to new behavior. doesntexist.foobar may result in an immediate DNS failure, but getaddrinfo may also hang until server selection times out. I observed this on my local network. I don't think the tests should be relying on getaddrinfo returning immediately on a bad address if it is network dependent.

src/libmongoc/tests/test-mongoc-topology.c
```diff
@@ -58,11 +58,11 @@ checks_cmp (checks_t *checks, const char *metric, char cmp, int expected)
    bson_mutex_unlock (&checks->mutex);
 
    if (cmp == '=') {
-      return expected == actual;
+      return actual == expected;
    } else if (cmp == '>') {
-      return expected > actual;
+      return actual > expected;
```
kevinAlbs, line 63: I had these flipped...

src/libmongoc/tests/test-mongoc-topology.c
```diff
@@ -76,6 +76,7 @@ check_started (const mongoc_apm_server_heartbeat_started_t *event)
 
    c = (checks_t *) mongoc_apm_server_heartbeat_started_get_context (event);
    bson_mutex_lock (&c->mutex);
+   MONGOC_DEBUG ("hearbeat started");
```
kevinAlbs, line 79: Will remove since this doesn't add much to the test.

src/libmongoc/tests/test-mongoc-topology.c
```diff
@@ -474,7 +475,7 @@ _test_topology_invalidate_server (bool pooled)
    checks_init (&checks);
    uri = test_framework_get_uri ();
    /* no auto heartbeat */
-   mongoc_uri_set_option_as_int32 (uri, "heartbeatFrequencyMS", INT32_MAX);
+   mongoc_uri_set_option_as_int32 (uri, "heartbeatFrequencyMS", 60000);
```
kevinAlbs, line 478: Ah, unnecessary. Will revert.

kevinAlbs, line 478: I need to verify this, but setting INT32_MAX results in an immediate network error when sending the awaitable ismaster.

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -2245,6 +2230,10 @@ _mongoc_cluster_stream_for_server (mongoc_cluster_t *cluster,
       mongoc_cluster_disconnect_node (cluster, server_id);
```
kevinAlbs, line 2230: TODO: note generation count should be checked

src/libmongoc/src/mongoc/mongoc-topology-background-monitoring.c
```diff
@@ -692,6 +148,40 @@ _mongoc_topology_background_monitoring_start (mongoc_topology_t *topology)
    }
 }
```
kevinAlbs, line 150: TODO: comment that topology->mutex should be locked

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1318 @@
...
+      mongoc_cond_t cond;
+      thread_state_t state;
+      bool scan_requested;
+      bool cancel_requested;
+      mongoc_interrupt_t *interrupt;
+   } shared;
+
+   /* Default time to sleep between ismaster checks (reduced when a scan is
+    * requested) */
+   uint64_t heartbeat_frequency_ms;
```
kevinAlbs, line 69: TODO: consider not copying everything from topology.

kevinAlbs: Some tests may modify these on the shared topology, so this might not be so straightforward.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1318 @@
...
+static bool
+_server_monitor_poll_with_interrupt (mongoc_server_monitor_t *server_monitor,
+                                     int64_t expire_at_ms,
+                                     bool *cancelled,
+                                     bson_error_t *error)
+{
+   MONITOR_LOG (server_monitor,
+                "_server_monitor_poll_with_interrupt expire_at_ms: %" PRIu64,
+                expire_at_ms);
+   while (_now_ms () < expire_at_ms) {
```
kevinAlbs, line 388: Consider `(last_now_ms = _now_ms()) < expire_at_ms`

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1318 @@
...
+                      MONGOC_ERROR_STREAM_SOCKET,
+                      "unexpected error polling streams");
+      return false;
+   }
+   bson_set_error (error,
+                   MONGOC_ERROR_STREAM,
+                   MONGOC_ERROR_STREAM_SOCKET,
+                   "timeout polling");
+   MONITOR_LOG (
+      server_monitor, "timeout polling: exceeded %" PRIu64, expire_at_ms);
```
kevinAlbs, line 487: Consider making this a MONGOC_ERROR log, and anything else that is error-worthy.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1318 @@
...
+ * Locks server monitor mutex.
+ * Returns true if awaitable was enabled (or is already enabled).
+ * Returns false if awaitable cannot be enabled.
+ */
+static bool
+_server_monitor_enable_awaitable_ismaster (
+   mongoc_server_monitor_t *server_monitor)
+{
+   if (server_monitor->initiator) {
+      /* Disable for custom initiator. */
```
kevinAlbs, line 651: TODO: log here or warning

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1318 @@
...
+   }
+   bson_mutex_lock (&server_monitor->shared.mutex);
+   if (server_monitor->shared.interrupt) {
+      bson_mutex_unlock (&server_monitor->shared.mutex);
+      return true;
+   }
+   server_monitor->shared.interrupt =
+      _mongoc_interrupt_new (server_monitor->connect_timeout_ms);
+   if (!server_monitor->shared.interrupt) {
+      MONITOR_LOG (server_monitor, "failed to create interrupt");
```
kevinAlbs, line 662: This should be an error log.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
kevinAlbs: Done.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
kevinAlbs: min_heartbeat_frequency_msec is only overriden in tests before scanning starts. So copying the value on the creation of a server monitor should be ok. My preference is not to access the shared topology object directly outside of the topology mutex.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1329 @@
...
+{
+   if (server_monitor->initiator) {
+      /* Disable for custom initiator. */
+      MONITOR_LOG_WARNING (server_monitor, "unable to enable server monitor "
+                                           "streaming due to custom stream "
+                                           "initiators");
+      return false;
+   }
+   bson_mutex_lock (&server_monitor->shared.mutex);
+   if (server_monitor->shared.interrupt) {
```
kevinAlbs, line 670: TODO: let's see if we can remove this interrupt.
Poll with a short timeout and check if cancelled instead.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1329 @@
...
+      bson_mutex_unlock (&server_monitor->shared.mutex);
+
+      mongoc_server_description_destroy (previous_description);
+      previous_description = mongoc_server_description_new_copy (description);
+      mongoc_server_description_destroy (description);
+      description = mongoc_server_monitor_check_server (
+         server_monitor, previous_description, &cancelled);
+
+      if (cancelled) {
+         continue;
```
kevinAlbs, line 1101: This is missing a wait.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1250 @@
...
+#else
+#define MONITOR_LOG(sm, ...) (void) 0
+#endif
+
+/* TODO use MONGOC_LOG_LEVEL_ERROR */
+#define MONITOR_LOG_ERROR(sm, ...) \
+   _server_monitor_log (sm, MONGOC_LOG_LEVEL_DEBUG, __VA_ARGS__)
+/* TODO use MONGOC_LOG_LEVEL_WARNING */
+#define MONITOR_LOG_WARNING(sm, ...) \
+   _server_monitor_log (sm, MONGOC_LOG_LEVEL_DEBUG, __VA_ARGS__)
```
kevinAlbs, line 131: There are assertions in evergreen that nothing to stderr is printed in tests without being captured within the test. Since there are quite a few tests that use the mock server to simulate network errors on monitoring, I'd prefer doing this in a follow-up PR if that is ok with ya'll.

kevinAlbs:
The interrupt is now removed from the server monitor, and there are additional assertions to tests to ensure a wait occurs after cancellation.

This should be ready for review again @jmikola + @bazile-clyde

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1241 @@
...
+                   MONGOC_ERROR_STREAM_SOCKET,
+                   "connection timeout while polling");
+   return false;
+}
+
+/* Receive an awaitable ismaster reply.
+ *
+ * May be used to receive additional replies when moreToCome is set.
+ * Called only from server monitor thread.
+ * Locks server monitor mutex.
```
jmikola, line 459: I don't see `bson_mutex_lock` called in this function. Is it called indirectly through some other function?

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1241 @@
...
+      bson_init (ismaster_reply);
+   }
+   _mongoc_buffer_destroy (&buffer);
+   return ret;
+}
+
+/* Send and receive an awaitable ismaster.
+ *
+ * Called only from server monitor thread.
+ * Locks server monitor mutex.
```
jmikola, line 563: Similarly, `bson_mutex_lock` isn't called here. Is it called indirectly (perhaps when this defers to `_server_monitor_awaitable_ismaster_recv`)?

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1241 @@
...
+      ismaster_reply = &description->last_is_master;
+   }
+
+   if (ismaster_reply) {
+      _mongoc_topology_update_cluster_time (topology, ismaster_reply);
+   }
+
+   bson_mutex_lock (&topology->mutex);
+   if (topology->scanner_state != MONGOC_TOPOLOGY_SCANNER_SHUTTING_DOWN) {
+      /* This is the another case of topology->mutex -> server monitor mutex. */
```
jmikola, line 627: What does "topology->mutex -> server monitor mutex" mean here? Is it referring to taking the server monitor lock while topology lock is held (from two lines up)?

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1241 @@
...
+   mongoc_topology_t *topology;
+   bson_t *ismaster_reply = NULL;
+
+   topology = server_monitor->topology;
+   if (description->has_is_master) {
+      ismaster_reply = &description->last_is_master;
+   }
+
+   if (ismaster_reply) {
+      _mongoc_topology_update_cluster_time (topology, ismaster_reply);
```
jmikola, line 622: `_mongoc_topology_update_cluster_time` also takes the lock internally, but I suppose there's no way around that unless you were to extract the update logic into a separate function so it could be called while the lock is already held -- allowing you to consolidate with the lock you take out below.

Probably nothing to do here, but I wanted to point it out in case there's a concern with taking out locks too frequently.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1241 @@
...
+      server_monitor->stream =
+         server_monitor->initiator (server_monitor->uri,
+                                    &server_monitor->description->host,
+                                    server_monitor->initiator_context,
+                                    error);
+   } else {
+      void *ssl_opts_void = NULL;
+
+#ifdef MONGOC_ENABLE_SSL
+      ssl_opts_void = server_monitor->ssl_opts;
```
jmikola, line 725: Does this need an explicit case, or is that redundant since `ssl_opts_void` is already a `void *`?

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1241 @@
...
+      ret = _server_monitor_awaitable_ismaster (
+         server_monitor,
+         &previous_description->topology_version,
+         &ismaster_reply,
+         cancelled,
+         &error);
+      GOTO (exit);
+   }
+
+   MONITOR_LOG (server_monitor, "polling ismaster");
```
jmikola, line 830: `_server_monitor_polling_ismaster` already logs "polling". Is this redundant?

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1241 @@
...
+      int64_t rtt_ms;
+      /* rtt remains -1 if awaited. */
+      if (awaited) {
+         rtt_ms = -1;
+      } else {
+         rtt_ms = duration_us / 1000;
+      }
+      mongoc_server_description_handle_ismaster (
+         description, &ismaster_reply, rtt_ms, NULL);
+      /* Check if it is a command error ("ok":0). */
```
jmikola, line 851: Should this actually check `ok`? As-is, you're just looking at `error.code`. Feel free to clarify is that's simply the way that `mongoc_server_description_handle_ismaster` relays that `ok` was zero -- in which case it probably makes sense to change this comment to something like `error.code will be non-zero on command error ("ok":0)`

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1241 @@
...
+   }
+
+   bson_destroy (&ismaster_reply);
+   return description;
+}
+
+/* Request scan of a single server.
+ *
+ * Caller does not need to have topology mutex locked.
+ * Locks server monitor mutex to deliver scan_requested.
```
jmikola, line 904: Should this also have a " Caller must not hold lock" line like the following function? Curious if that applies for all cases where functions take their own locks (in case earlier functions also omitted the " Caller must not hold lock." line).

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1241 @@
...
+      if (now_ms >= scan_due_ms) {
+         break;
+      }
+
+      if (server_monitor->shared.scan_requested) {
+         server_monitor->shared.scan_requested = false;
+         scan_due_ms = start_ms + server_monitor->min_heartbeat_frequency_ms;
+      }
+
+
```
jmikola, line 965: Extra whitespace.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1241 @@
...
+      sleep_duration_ms = scan_due_ms - now_ms;
+
+      if (sleep_duration_ms > 0) {
+         MONITOR_LOG (
+            server_monitor, "sleeping for %" PRId64, sleep_duration_ms);
+         mongoc_cond_timedwait (&server_monitor->shared.cond,
```
jmikola, line 971: Should we be concerned about a non-successful return? Is there a case where this returns and the mutex is _not_ re-locked? See [`pthread_cond_timedwait`](https://linux.die.net/man/3/pthread_cond_timedwait) and the alternative definition in `mongoc-thread-private.h` for Windows.

If this results in adding more logic here, I might suggest refactoring the `if/else` to just have an early guard statement that breaks and allows us to leave this code at the top level of the `while` iteration. For example:

```c
if (sleep_duration_ms <= 0) {
   break;
}

MONITOR_LOG...
```

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1241 @@
...
+      }
+
+      now_ms = _now_ms ();
+      if (now_ms >= scan_due_ms) {
+         break;
+      }
+
+      if (server_monitor->shared.scan_requested) {
+         server_monitor->shared.scan_requested = false;
+         scan_due_ms = start_ms + server_monitor->min_heartbeat_frequency_ms;
```
jmikola, line 962: If we do successfully sleep below, we'll jump to a new iteration of the `while` loop. Assuming we don't break because (a) server monitor thread is no longer running and (b) `now_ms` has not surpassed `scan_due_ms`, then are we sure that `scan_due_ms` is getting assigned correctly? `start_ms + server_monitor->min_heartbeat_frequency_ms` will not have changed since the last iteration.

Should this be calculated using `now_ms` instead, such that `start_ms` is only used for the initial `scan_due_ms` calculation (before the `while` loop)?

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1241 @@
...
+      if (server_monitor->more_to_come) {
+         MONITOR_LOG (server_monitor,
+                      "immediately proceeding due to moreToCome");
+         continue;
+      }
+
+      /* ... or the server has just transitioned to Unknown due to a network
+       * error. */
+      if (_mongoc_error_is_network (&description->error) &&
+          previous_description->type != MONGOC_SERVER_UNKNOWN) {
```
jmikola, line 1038: Should we check that the current description is `UNKNOWN`, or is that a given if `_mongoc_error_is_network` returns true?

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1241 @@
...
+static bool
+_server_monitor_ping_server (mongoc_server_monitor_t *server_monitor,
+                             int64_t *rtt_ms)
+{
+   bool ret = false;
+   int64_t start_us = _now_us ();
+   bson_t ismaster_reply;
+   bson_error_t error;
+
+   *rtt_ms = -1;
```
jmikola, line 1064: Is `-1` a legitimate return value if we cannot conduct a ping? Just want to make sure it's handled properly elsewhere. Assuming this relates to `_server_monitor_rtt_thread` below, I might suggest using a defined constant like `RTT_UNSET` instead of `-1` for self-documenting purposes.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -0,0 +1,1241 @@
...
+
+         bson_mutex_lock (&server_monitor->topology->mutex);
+         sd = mongoc_topology_description_server_by_id (
+            &server_monitor->topology->description,
+            server_monitor->description->id,
+            &error);
+         if (sd) {
+            /* If the server description has been removed, the RTT thread will
+             * be terminated by background monitoring soon. */
+            mongoc_server_description_update_rtt (sd, rtt_ms);
```
jmikola, line 1118: Should `sd` be destroyed before we release the topology lock, or is this returning an `sd` directly from the topology's set, which is why we hold the topology lock while working with it?

src/libmongoc/src/mongoc/mongoc-topology-description.c
jmikola: Cue earlier suggestion about using a constant.

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -327,6 +328,11 @@ mongoc_topology_new (const mongoc_uri_t *uri, bool single_threaded)
    if (service) {
       memset (&rr_data, 0, sizeof (mongoc_rr_data_t));
 
+      /* Upon failure, the SRV polling thread still needs a duration. */
```
jmikola, line 331: What does "upon failure" mean? Is this to ensure you set these in case a jump is made to `srv_fail` later on?

Do I assume correctly that this has no impact on the original assignments later on and those should be left as-is?

src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -1450,6 +1475,10 @@ mock_server_destroy (mock_server_t *server)
    int64_t deadline = bson_get_monotonic_time () + 10 * 1000 * 1000;
    request_t *request;
 
+   if (!server) {
+      return;
+   }
```
jmikola, line 1480: Any reason for adding this? Did some test introduce a call to `mock_server_destroy` with a null argument, or was it just something you noticed and wanted to change for consistency?

src/libmongoc/tests/test-mongoc-background-monitoring.c
```diff
@@ -24,6 +24,7 @@
 #include "mongoc/mongoc-client-pool-private.h"
 #include "test-libmongoc.h"
 #include "TestSuite.h"
+#include "test-conveniences.h"
```
jmikola, line 27: Probably makes no difference, but is this typically ordered before `test-libmongoc.h`?

src/libmongoc/tests/test-mongoc-background-monitoring.c
```diff
@@ -253,7 +296,11 @@ tf_destroy (test_fixture_t *tf)
       bson_mutex_unlock (&_tf->mutex);                     \
    } while (0);
 
-#define WAIT _mongoc_usleep (10 * 1000);
+/* Wait for two periods of the faster heartbeat.
```
jmikola, line 299: What does "two periods of the faster heartbeat" refer to? If this pertains to `min_heartbeat_frequency_msec` being changed to `10` earlier, should we instead use a constant for that and change this to `2 * CONSTANT * 1000`?

src/libmongoc/tests/test-mongoc-background-monitoring.c
```diff
@@ -600,16 +683,441 @@ test_repeated_requestscan (void)
...
-      OBSERVE_SOON (tf, tf->observations->n_heartbeat_succeeded == i);
+   for (i = 0; i < 10; i++) {
+      _request_scan (tf);
+   }
+   request = mock_server_receives_ismaster (tf->server);
+   OBSERVE (tf, request);
+   for (i = 0; i < 10; i++) {
       _request_scan (tf);
    }
+   WAIT;
```
jmikola, line 697: Perhaps `WAIT_TWO_MIN_HEARTBEAT_MS` might be a more descriptive name.

src/libmongoc/tests/test-mongoc-background-monitoring.c
```diff
@@ -600,16 +683,441 @@ test_repeated_requestscan (void)
...
+
+   /* After handling a scan request */
+   tf = tf_new (TF_FAST_MIN_HEARTBEAT);
+   _request_scan (tf);
+   request = mock_server_receives_ismaster (tf->server);
+   OBSERVE (tf, tf->observations->n_heartbeat_started == 1);
+   mock_server_replies_ok_and_destroys (request);
+   OBSERVE_SOON (tf, tf->observations->n_heartbeat_succeeded == 1);
+   /* Wait for two min heartbeat periods. */
+   WAIT;
```
jmikola, line 719: Per my earlier suggestion, a more descriptive macro would also make the comment unnecessary.

src/libmongoc/tests/test-mongoc-background-monitoring.c
```diff
@@ -154,18 +166,43 @@ static void
...
+           mongoc_server_description_type (old_sd),
+           mongoc_server_description_type (new_sd));
    tf->observations->sd_type = new_sd->type;
    tf->observations->n_server_changed++;
    mongoc_cond_broadcast (&tf->cond);
    bson_mutex_unlock (&tf->mutex);
 }
 
+#define TV \
+   "{ 'processId': { '$oid': 'AABBAABBAABBAABBAABBAABB' }, 'counter': 1 }"
```
jmikola, line 187: Having looked at the tests in this file, I didn't see anything that tests RTT or increasing topology versions. Should either of those be tested here, or is that something covered by spec tests?

With regard to RTT, I'm curious how that factors in given that a server monitor for streamable isMaster would be accompanied by a separate RTT thread -- wouldn't the RTT pings interfere with the mocked auto-responses?

src/libmongoc/tests/test-mongoc-primary-stepdown.c
```diff
@@ -65,12 +66,10 @@ _connection_count (mongoc_database_t *db)
...
    int conns;
 
-   ASSERT (db);
-
    BSON_APPEND_INT32 (&cmd, "serverStatus", 1);
 
-   read_prefs = mongoc_read_prefs_new (MONGOC_READ_PRIMARY);
-   res = mongoc_database_command_simple (db, &cmd, read_prefs, &reply, &error);
+   res = mongoc_client_command_simple_with_server_id (
+      client, "admin", &cmd, NULL, server_id, &reply, &error);
```
jmikola, line 72: If I understand your previous comment correctly, the fix here was to use the same `server_id` instead of re-selecting with a primary read preference, correct?

src/libmongoc/tests/test-mongoc-sdam-monitoring.c
```diff
@@ -827,6 +829,7 @@ _test_heartbeat_fails_dns (bool pooled)
    size_t i;
 
    context_init (&context);
+   start = bson_get_monotonic_time ();
```
jmikola, line 832: Earlier `_test_heartbeat_events` had a comment that the start time should be taken before `mongoc_client_pool_pop`. Should there be a similar comment here?

src/libmongoc/tests/test-mongoc-sdam-monitoring.c
```diff
@@ -838,19 +841,17 @@ _test_heartbeat_fails_dns (bool pooled)
...
+   /* This should result in either a DNS failure or connection failure depending
+    * on the network. */
    ASSERT (!r);
    ASSERT_ERROR_CONTAINS (error,
                           MONGOC_ERROR_SERVER_SELECTION,
                           MONGOC_ERROR_SERVER_SELECTION_FAILURE,
-                          "Failed to resolve");
-
-   duration = bson_get_monotonic_time () - start;
+                          "");
```
jmikola, line 854: Is this forgoing any assertion on the string message and just relying on a domain/code check? If so, I'd suggest amending the comment above to explain something like "We assert the domain/code but not the message string".

src/libmongoc/tests/test-mongoc-sdam-monitoring.c
```diff
@@ -859,6 +860,8 @@ _test_heartbeat_fails_dns (bool pooled)
       mongoc_client_destroy (client);
    }
 
+   duration = bson_get_monotonic_time () - start;
```
jmikola, line 863: In this case, you moved the `duration` calculation _after_ returning the client to the pool. Does this warrant a comment to balance out the earlier case of computing the `start` time _before_ pool popping?

src/libmongoc/src/mongoc/mongoc-server-monitor.c
jmikola: Handling timeout errors works for me. Thanks.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
jmikola: Clarification seems fine. Thanks.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
jmikola: Ah, I think I was reading this all backwards. If the `while` loop repeats, that means `mongoc_cond_timedwait` didn't sleep for as long as we need to because something woke up the thread. That's why we abort this loop and return from the function if either `sleep_duration_ms` is negative (no waiting required) or `mongo_cond_ret_is_timedout` is true (we waited the full duration we needed to).

This LGTM.

src/libmongoc/src/mongoc/mongoc-topology-description.c
jmikola: Did you miss replacing this with `MONGOC_RTT_UNSET`?

src/libmongoc/tests/test-mongoc-primary-stepdown.c
```diff
@@ -235,7 +250,7 @@ test_not_master_keep_pool (mongoc_client_t *client)
    ASSERT (res);
 
    /* Verify that the connection pool has not been cleared */
-   ASSERT (conn_count == _connection_count (db));
+   ASSERT (conn_count == _connection_count (client, primary_id));
```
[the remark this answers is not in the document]
kevinAlbs: Good idea, done.

src/libmongoc/src/mongoc/mongoc-topology-description.c
kevinAlbs: Ah yes, fixed.

kevinAlbs:
Rebased and updated. I'll merge assuming tests pass. Thanks for reviewing!

PR #627, opened 2020-06-02
CDRIVER-3571 consolidate and simplify installation doc

rcsanchez97:
Note: no Evergreen patch build since this documentation only.

### DIFF
--- a/src/libmongoc/doc/installing.rst
+++ b/src/libmongoc/doc/installing.rst
@@ -8,16 +8,16 @@ The following guide will step you through the process of downloading, building, 
 Supported Platforms
 -------------------
 
-The MongoDB C Driver is `continuously tested <https://evergreen.mongodb.com/waterfall/mongo-c-driver>`_ on variety of platforms including:
+The MongoDB C Driver is `continuously tested <https://evergreen.mongodb.com/waterfall/mongo-c-driver>`_ on a variety of platforms including:
 
 - Archlinux
-- Debian 9.2
-- macOS 10.12
-- Microsoft Windows Server 2008
-- RHEL 6.7, 7.0, 7.1, 7.2
+- Debian 9.2, 10.0
+- macOS 10.14
+- Microsoft Windows Server 2008, 2016
+- RHEL 6.2, 7.0, 7.1, 8.2
 - Ubuntu 16.04, 18.04
-- Clang 3.4, 3.5, 3.7, 3.8
-- GCC 4.8, 4.9, 5.4, 6.3
+- Clang 3.4, 3.5, 3.7, 3.8, 6.0
+- GCC 4.8, 4.9, 5.4, 6.3, 8.2, 8.3
 - MinGW-W64
 - Visual Studio 2013, 2015, 2017
 - x86, x86_64, ARM (aarch64), Power8 (ppc64le), zSeries (s390x)
@@ -25,7 +25,7 @@ The MongoDB C Driver is `continuously tested <https://evergreen.mongodb.com/wate
 Install libmongoc with a Package Manager
 ----------------------------------------
 
-Several Linux distributions provide packages for libmongoc and its dependencies. One advantage of installing libmongoc with a package manager is that its dependencies (including libbson) will be installed automatically.
+Several Linux distributions provide packages for libmongoc and its dependencies. One advantage of installing libmongoc with a package manager is that its dependencies (including libbson) will be installed automatically. If you choose to install libmongoc from distribution packages, use the package manager to confirm the version being installed is sufficient for your needs.
 
 The libmongoc package is available on recent versions of Debian and Ubuntu.
 
@@ -45,6 +45,12 @@ On recent Red Hat systems, such as CentOS and RHEL 7, a mongo-c-driver package i
 
   $ yum install mongo-c-driver
 
+On macOS systems with Homebrew, the mongo-c-driver package can be installed with:
+
+.. code-block:: none
+
+  $ brew install mongo-c-driver
+
 
 .. _installing_libbson_with_pkg_manager:
 
@@ -55,7 +61,7 @@ The libbson package is available on recent versions of Debian and Ubuntu. If you
 
 .. code-block:: none
 
-  $ apt-get install libbson-1.0
+  $ apt-get install libbson-1.0-0
 
 On Fedora, a libbson package is available in the default repositories and can be installed with:
 
@@ -72,11 +78,14 @@ The package can be installed with:
 
   $ yum install libbson
 
-Building on Unix
-----------------
+Build environment
+-----------------
+
+Build environment on Unix
+^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Prerequisites for libmongoc
-^^^^^^^^^^^^^^^^^^^^^^^^^^^
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 OpenSSL is required for authentication or for TLS connections to MongoDB. Kerberos or LDAP support requires Cyrus SASL.
 
@@ -99,24 +108,63 @@ On FreeBSD:
   $ su -c 'pkg install cmake openssl cyrus-sasl'
 
 Prerequisites for libbson
-^^^^^^^^^^^^^^^^^^^^^^^^^
+~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The only prerequisite for building libbson is ``cmake``. The command lines above can be adjusted to install only ``cmake``.
 
-Building from a release tarball
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Build environment on macOS
+^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Unless you intend to contribute to mongo-c-driver and/or libbson, you will want to build from a release tarball.
+Install the XCode Command Line Tools:
 
-The most recent release of libmongoc and libbson, both of which are included in mongo-c-driver, can be `downloaded here <https://github.com/mongodb/mongo-c-driver/releases/latest>`_. The instructions in this document utilize ``cmake``'s out-of-source build feature to keep build artifacts separate from source files.
+.. code-block:: none
+
+  $ xcode-select --install
+
+The ``cmake`` utility is also required. First `install Homebrew according to its instructions <https://brew.sh/>`_, then:
+
+.. code-block:: none
+
+  $ brew install cmake
+
+.. _build-on-windows:
+
+Build environment on Windows with Visual Studio
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Building on Windows requires Windows Vista or newer and Visual Studio 2010 or newer. Additionally, ``cmake`` is required to generate Visual Studio project files.  Installation of these components on Windows is beyond the scope of this document.
+
+Build environment on Windows with MinGW-W64 and MSYS2
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Install MSYS2 from `msys2.github.io <http://msys2.github.io>`_. Choose the x86_64 version, not i686.
+
+Open the MingGW shell with ``c:\msys64\ming64.exe`` (not the msys2_shell). Install dependencies:
+
+.. code-block:: none
+
+  $ pacman --noconfirm -Syu
+  $ pacman --noconfirm -S mingw-w64-x86_64-gcc mingw-w64-x86_64-cmake
+  $ pacman --noconfirm -S mingw-w64-x86_64-extra-cmake-modules make tar
+  $ pacman --noconfirm -S mingw64/mingw-w64-x86_64-cyrus-sasl
+
+Configuring the build
+---------------------
+
+Before building libmongoc and/or libbson, it is necessary to configure, or prepare, the build.  The steps to prepare the build depend on how you obtained the source code and the build platform.
+
+Preparing a build from a release tarball
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The most recent release of libmongoc and libbson, both of which are included in mongo-c-driver, can be `downloaded here <https://github.com/mongodb/mongo-c-driver/releases/latest>`_. The instructions in this document utilize ``cmake``'s out-of-source build feature to keep build artifacts separate from source files. While the ``$`` prompt is used throughout, the instructions below will work on Linux, macOS, and Windows (assuming that CMake is in the user's shell path in all cases).  See the subsequent sections for additional platform-specific instructions.
 
 The following snippet will download and extract the driver, and configure it:
 
 .. parsed-literal::
 
-  $ wget https://github.com/mongodb/mongo-c-driver/releases/download/x.y.z/mongo-c-driver-x.y.z.tar.gz
-  $ tar xzf mongo-c-driver-x.y.z.tar.gz
-  $ cd mongo-c-driver-x.y.z
+  $ wget https://github.com/mongodb/mongo-c-driver/releases/download/|version|/mongo-c-driver-|version|.tar.gz
+  $ tar xzf mongo-c-driver-|version|.tar.gz
+  $ cd mongo-c-driver-|version|
   $ mkdir cmake-build
   $ cd cmake-build
   $ cmake -DENABLE_AUTOMATIC_INIT_AND_CLEANUP=OFF ..
@@ -127,11 +175,11 @@ If ``cmake`` completed successfully, you will see a considerable amount of outpu
 
 .. parsed-literal::
 
-  -- Build files have been written to: /home/user/mongo-c-driver-x.y.z/cmake-build
+  -- Build files have been written to: /home/user/mongo-c-driver-|version|/cmake-build
 
-If ``cmake`` concludes with anything different, then there is likely an error or some other problem with the build. Review the output to identify and correct the problem.
+If ``cmake`` concludes with anything different, then it is likely an error occurred.
 
-mongo-c-driver contains a copy of libbson, in case your system does not already have libbson installed. The build will detect if libbson is not installed and use the bundled libbson.
+mongo-c-driver contains a copy of libbson, in case your system does not already have libbson installed. The configuration will detect if libbson is not installed and use the bundled libbson.
 
 Additionally, it is possible to build only libbson by setting the ``-DENABLE_MONGOC=OFF`` option:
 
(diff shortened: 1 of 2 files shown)


src/libmongoc/doc/installing.rst
```diff
@@ -251,17 +252,23 @@ Let's start by generating Visual Studio project files. The following assumes we
 
 (Run ``cmake -LH ..`` for a list of other options.)
 
-Now that we have project files generated, we can either open the project in Visual Studio or compile from the command line. Let's build using the command line program ``msbuild.exe``:
+To see a complete list the CMake generators available on your specific system, use a command like this:
+
+.. parsed-literal::
+
+ $ cmake -E capabilities | json_pp -json_opt pretty,canonical
```
kevinAlbs, line 259: Can we replace this with `cmake --help` instead? At least on macOS that lists available generators.

src/libmongoc/doc/installing.rst
```diff
@@ -108,7 +108,7 @@ Building from a release tarball
 
 Unless you intend to contribute to mongo-c-driver and/or libbson, you will want to build from a release tarball.
 
-The most recent release of libmongoc and libbson, both of which are included in mongo-c-driver, can be `downloaded here <https://github.com/mongodb/mongo-c-driver/releases/latest>`_. The instructions in this document utilize ``cmake``'s out-of-source build feature to keep build artifacts separate from source files.
```
kevinAlbs, line 111: Can we remove the "Unless you intend to..." line above. That seems a little out of place to me.

src/libmongoc/doc/installing.rst
```diff
@@ -108,7 +108,7 @@ Building from a release tarball
...
+The most recent release of libmongoc and libbson, both of which are included in mongo-c-driver, can be `downloaded here <https://github.com/mongodb/mongo-c-driver/releases/latest>`_. The instructions in this document utilize ``cmake``'s out-of-source build feature to keep build artifacts separate from source files. While the ``$`` prompt is used throughout, the instructions below will work on Linux, macOS, and Windows (assuming that CMake is in the user's shell path in all cases).  See the subsequent sections for additional platform-specific instructions.
```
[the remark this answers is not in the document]
rcsanchez97: I like this suggestion.  The `-B` option is not one I am accustomed to using.  I've updated the instances of `..` and `--build .` and tried out the steps.  Everything works the same and the commands are clearer.

src/libmongoc/doc/installing.rst
```diff
@@ -141,9 +141,26 @@ Additionally, it is possible to build only libbson by setting the ``-DENABLE_MON
 
 A build configuration description similar to the one above will be displayed, though with fewer entries. Once the configuration is complete, the selected items can be built and installed with these commands:
 
-.. code-block:: none
+.. parsed-literal::
+
```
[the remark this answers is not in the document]
rcsanchez97: The rationale is to limit the parts that run with root permissions to the minimum.  Breaking it up into two steps allows the bulk of the build to proceed under the user's permissions, with only the actual installation steps running as root.

src/libmongoc/doc/installing.rst
```diff
@@ -161,7 +178,7 @@ The second way to uninstall is from within the build directory, assuming that it
 
 .. code-block:: none
 
-  $ sudo make uninstall
+  $ sudo cmake --build . --target uninstall
```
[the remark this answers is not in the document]
rcsanchez97: That's correct.  I also noted at L147 that use of `sudo` may or may not apply, depending on the system configuration.

src/libmongoc/doc/installing.rst
```diff
@@ -179,7 +196,8 @@ Clone the repository and build the current master or a particular release tag:
   $ mkdir cmake-build
   $ cd cmake-build
   $ cmake -DENABLE_AUTOMATIC_INIT_AND_CLEANUP=OFF ..
-  $ sudo make install
+  $ cmake --build .
+  $ sudo cmake --build . --target install
```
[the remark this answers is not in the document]
rcsanchez97: Same.

src/libmongoc/doc/installing.rst
```diff
@@ -55,7 +55,7 @@ The libbson package is available on recent versions of Debian and Ubuntu. If you
 
 .. code-block:: none
 
-  $ apt-get install libbson-1.0
+  $ apt-get install libbson-1.0-0
```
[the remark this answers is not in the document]
rcsanchez97: Done.

src/libmongoc/doc/installing.rst
rcsanchez97: Done.

src/libmongoc/doc/installing.rst
rcsanchez97: Done

src/libmongoc/doc/installing.rst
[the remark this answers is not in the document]
kevinAlbs: Ah, that seems sensible.

src/libmongoc/doc/installing.rst
```diff
@@ -141,9 +141,26 @@ Additionally, it is possible to build only libbson by setting the ``-DENABLE_MON
...
+
+(Note that the ``sudo`` command may not be applicable or available depending on the configuration of your system.)
+
+In the above commands, the first relies on the default target which builds all configured components.  For fine grained control over what gets built, the following command can be used (for Ninja and Makefile-based build systems) to list all available targets:
+
+.. parsed-literal::
+
+  $ sudo cmake --build . help
+
+For Visual Studio 2019 (16.4 and newer), this command can be used to list all available targets:
```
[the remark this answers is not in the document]
kevinAlbs: Additionally, the macOS and MinGW instructions say to "Follow the build instructions above.", which seems off (at least for MinGW).

That being said, we have more Windows specific instructions below describing the "uninstall" script.

I'm having trouble coming up with a great restructuring to fix that. I'm also ok to leaving this for now. I think having platform specific instructions on each page (with duplication) might be the most straightforward for users, which we can do in https://jira.mongodb.org/browse/CDRIVER-3703.

src/libmongoc/doc/installing.rst
```diff
@@ -211,24 +233,7 @@ The ``cmake`` utility is also required. First `install Homebrew according to its
...
-
-.. code-block:: none
-
-  $ mkdir cmake-build
-  $ cd cmake-build
-  $ cmake -DENABLE_AUTOMATIC_INIT_AND_CLEANUP=OFF ..
-  $ make install
-
-All of the same variations described above (e.g., building only libbson, building documentation, etc.) are available when building on macOS.
+Follow the build instructions above.
```
kevinAlbs, line 236: Above is a bit vague. I suggest linking directly to the "Build from release tarball section", which I believe you can do with a sphinx label and reference

`.. build-on-windows` below is one example, and is linked to like:
```
:ref:`when you ran CMake <build-on-windows>`
```

src/libmongoc/doc/installing.rst
```diff
@@ -281,22 +290,7 @@ Open the MingGW shell with ``c:\msys64\ming64.exe`` (not the msys2_shell). Insta
...
-
-Create a build directory and build and install with cmake:
-
-.. code-block:: none
-
-  $ mkdir cmake-build
-  $ cd cmake-build
-  $ CC=/mingw64/bin/gcc.exe /mingw64/bin/cmake -G "MSYS Makefiles" -DCMAKE_INSTALL_PREFIX="C:/mongo-c-driver" -DCMAKE_C_FLAGS="-D__USE_MINGW_ANSI_STDIO=1" ..
-  $ make install
+Follow the build instructions above.
```
kevinAlbs, line 293: Here, it seems like we may not be able to redirect users. The line:
```
CC=/mingw64/bin/gcc.exe /mingw64/bin/cmake -G "MSYS Makefiles" -DCMAKE_INSTALL_PREFIX="C:/mongo-c-driver" -DCMAKE_C_FLAGS="-D__USE_MINGW_ANSI_STDIO=1" ..
```
The generator and `-D__USE_MINGW_ANSI_STDIO=1` is specific to MinGW.

In that case, I'm fine with having the duplication of instructions here, and just replacing `make install` with `cmake --build . --target install` for consistency.

src/libmongoc/doc/installing.rst
```diff
@@ -305,8 +299,6 @@ In the event that you are building the BSON library and/or the C driver to embed
 
 .. code-block:: none
 
-  $ mkdir cmake-build
-  $ cd cmake-build
```
kevinAlbs, line 309: We need this back I think, sinc we cannot use `-B`.

src/libmongoc/doc/installing.rst
```diff
@@ -195,9 +197,11 @@ Clone the repository and build the current master or a particular release tag:
   $ cd mongo-c-driver
   $ git checkout |version|  # To build a particular release
```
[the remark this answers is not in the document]
rcsanchez97: Good catch.  Done.

src/libmongoc/doc/installing.rst
rcsanchez97: Ack.

src/libmongoc/doc/installing.rst
rcsanchez97: I think my restructuring takes care of this.

rcsanchez97:
@kevinAlbs When I took another look at `installing.rst` to implement your suggestions it became clear that the structure could be reworked in a more clear manner.  Specifically, the structure is now along the lines of:

- Supported platforms
- Installation with a package manager
- Build environment (prerequisites)
  - on Unix
  - on macOS
  - on Windows w/ Visual Studio
  - on Windows w/ MinGW/MSYS2
- Configuring the build
  - from a release tarball
  - from a Git repo clone
  - on Windows w/ Visual Studio
- Executing the build
  - on Unix, macOS, and Windows (MinGW-W64 and MSYS2)
  - on Windows with Visual Studio
  - Documentation
  - Uninstall
- Integrator options

That structure seems to keep things organized, groups related content, and minimizes duplication.  If you find this structure to be OK, then CDRIVER-3703 may not be necessary.

Please let me know your thoughts.

src/libmongoc/doc/installing.rst
```diff
@@ -127,7 +172,7 @@ If ``cmake`` completed successfully, you will see a considerable amount of outpu
 
 .. parsed-literal::
 
-  -- Build files have been written to: /home/user/mongo-c-driver-x.y.z/cmake-build
+  -- Build files have been written to: /home/user/mongo-c-driver-|version|/cmake-build
 
 If ``cmake`` concludes with anything different, then there is likely an error or some other problem with the build. Review the output to identify and correct the problem.
```
kevinAlbs, line 177: Can we also change:
> then there is likely an error or some other problem with the build. Review the output to identify and correct the problem.

to:
> then it is likely an error occurred.

This is the configure step, not the build step. And the second sentence does not seem helpful.

src/libmongoc/doc/installing.rst
```diff
@@ -127,7 +172,7 @@ If ``cmake`` completed successfully, you will see a considerable amount of outpu
...
```
kevinAlbs, line 178: Below too, "The build will detect" => "The configuration will detect"?

src/libmongoc/doc/installing.rst
```diff
@@ -141,162 +186,133 @@ Additionally, it is possible to build only libbson by setting the ``-DENABLE_MON
...
   $ cmake -DENABLE_AUTOMATIC_INIT_AND_CLEANUP=OFF ..
-  $ sudo make install
 
-Generating the documentation
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Preparing a build on Windows with Visual Studio
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Install `Sphinx <http://www.sphinx-doc.org/>`_, then:
+On the Windows platform with Visual Studio, it may be necessary to specify the CMake generator to use.  This is especially important if multiple versions of Visual Studio are installed on the system or if alternate built tools (e.g., MinGW, MSYS2, Cygwin, etc.) are present on the system.  Specifying the generator will ensure that the build configuration is known with certainty, rather than relying on the toolchain that CMake happens to find.
```
kevinAlbs, line 207: "alternate built" => "alternate build"

src/libmongoc/doc/installing.rst
```diff
@@ -141,162 +186,133 @@ Additionally, it is possible to build only libbson by setting the ``-DENABLE_MON
...
 
-.. code-block:: none
+Let's start by generating Visual Studio project files. The following assumes we are compiling for 64-bit Windows using Visual Studio 2015 Express, which can be freely downloaded from Microsoft. We will be utilizing ``cmake``'s out-of-source build feature to keep build artifacts separate from source files.
```
kevinAlbs, line 209: Remove "Let's" for consistency with other instructions. Most of the language in this guide is imperative, and uses "you" instead of "we". I suggest changing "assumes we" to "assumes you" and "We will be utilizing" to "Utilize" (out-of-source builds seems like the general recommendation, so making it imperative seems ok).

src/libmongoc/doc/installing.rst
```diff
@@ -141,162 +186,133 @@ Additionally, it is possible to build only libbson by setting the ``-DENABLE_MON
...
-  $ cd mongo-c-driver-x.y.z
+  $ cmake --build . help
+
+Building on Windows with Visual Studio
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Once the project files are generated, the project can be opened directly in Visual Studio or compiled from the command line.
 
-Build and install the driver:
+Let's build using the CMake build tool mode:
```
kevinAlbs, line 251: Remove "Let's" for consistency.

src/libmongoc/doc/installing.rst
```diff
@@ -141,162 +186,133 @@ Additionally, it is possible to build only libbson by setting the ``-DENABLE_MON
...
 .. code-block:: none
 
-  $ mkdir cmake-build
-  $ cd cmake-build
-  $ cmake -DENABLE_AUTOMATIC_INIT_AND_CLEANUP=OFF ..
-  $ make install
+  $ cmake --build . --config RelWithDebInfo
 
-All of the same variations described above (e.g., building only libbson, building documentation, etc.) are available when building on macOS.
+Visual Studio's default build type is ``Debug``, but we recommend a release build with debug info for production use. Now that libmongoc and libbson are compiled, let's install them. Components will be installed to the path specified by ``CMAKE_INSTALL_PREFIX``.
```
kevinAlbs, line 257: This "we" seems ok, since this is our recommendation, not something we're telling the reader to do. Remove the "let's".

src/libmongoc/doc/installing.rst
```diff
@@ -141,162 +186,133 @@ Additionally, it is possible to build only libbson by setting the ``-DENABLE_MON
...
-.. code-block:: none
+Let's start by generating Visual Studio project files. The following assumes we are compiling for 64-bit Windows using Visual Studio 2015 Express, which can be freely downloaded from Microsoft. We will be utilizing ``cmake``'s out-of-source build feature to keep build artifacts separate from source files.
 
-  $ cmake -DENABLE_MAN_PAGES=ON -DENABLE_HTML_DOCS=ON ..
-  $ make mongoc-doc
+.. parsed-literal::
 
-To build only the libbson documentation:
+  $ cd mongo-c-driver-|version|
+  $ cmake -G "Visual Studio 14 2015 Win64" \\
```
kevinAlbs, line 214: This needs the extra instructions:
```
mkdir cmake-build
cd cmake-build
```

src/libmongoc/doc/installing.rst
```diff
@@ -141,162 +186,133 @@ Additionally, it is possible to build only libbson by setting the ``-DENABLE_MON
...
 .. parsed-literal::
 
-  $ cd mongo-c-driver-x.y.z
-  $ mkdir cmake-build
-  $ cd cmake-build
-  $ cmake -G "Visual Studio 14 2015 Win64" \\
-      "-DCMAKE_INSTALL_PREFIX=C:\\mongo-c-driver" \\
-      "-DCMAKE_PREFIX_PATH=C:\\mongo-c-driver" \\
-      ..
+  $ cmake --build . /targets
```
kevinAlbs, line 269: I tested this locally. It requires an extra `--`, but I don't think this lists cmake targets.

```
cmake --build . -- /targets
```

Here is a subset of the output
```
CppClean
BeforeCppClean
AfterCppClean
CoreCppClean
PGInstrumentedClean
InitGetClCommandLines
FixupReferences
GetClCommandLines
GetProjectDirectories
GetGeneratedFiles
GetProjectReferencesInfo
GetAssemblyReferences
GetWinMDReferences
GetComReferences
GetSDKReferences
GetProjectReferences
GetForeignReferences
GetResolvedReferences
_PrintPropertyPages
_PrintIntellisenseInfo
_ValidatePathInfo
_PerfIntellisenseInfo
_PrintResolvedProjectReferences
SetTelemetryEnvironmentVariables
```

I believe those are msbuild targets based off of https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-build-specific-targets-in-solutions-by-using-msbuild-exe?view=vs-2019

Perhaps we should just leave the instruction to look at  `*.vcxproj` files as the list of build targets.

kevinAlbs:
I like the restructuring!

Can we add a top-level heading for "Build environment"? The sidebar in the docs lists top level headings:
```
Supported Platforms
Install libmongoc with a Package Manager
Install libbson with a Package Manager
Build environment on Unix
Build environment on macOS
Build environment on Windows with Visual Studio
Build environment on Windows with MinGW-W64 and MSYS2
Configuring the build
Executing a build
Additional Options for Integrators
```

If we add a top-level heading for "Build environment" that could show the a simpler outline:

```
Supported Platforms
Install libmongoc with a Package Manager
Install libbson with a Package Manager
Build environment
Configuring the build
Executing a build
Additional Options for Integrators
```

src/libmongoc/doc/installing.rst
rcsanchez97: Done

src/libmongoc/doc/installing.rst
rcsanchez97: Done

src/libmongoc/doc/installing.rst
rcsanchez97: Done.

src/libmongoc/doc/installing.rst
rcsanchez97: Done.

src/libmongoc/doc/installing.rst
rcsanchez97: Done

src/libmongoc/doc/installing.rst
rcsanchez97: Done.

src/libmongoc/doc/installing.rst
rcsanchez97: While CMake has a notion of targets, it creates any specified targets in the chosen build system.  Most build systems have default or built in targets and that seems to be the case here.  In any event, I've left it for now (though I fixed the command to add in the `--` that I missed).  If you still feel strongly like this won't be too helpful, then I'll remove it.

src/libmongoc/doc/installing.rst
rcsanchez97: Done.

rcsanchez97:
@kevinAlbs I think that this might be the last of the changes, except perhaps removing the target list for Visual Studio.  Please have another look and let me know what you think.

src/libmongoc/doc/installing.rst
kevinAlbs: True, users could still find this helpful then. This is fine with me.

PR #628, opened 2020-06-03
CDRIVER-3373 Avoid using readConcern snapshot in sharded transaction tests

bazile-clyde:


src/libmongoc/tests/json-test.c
```diff
@@ -1504,37 +1504,6 @@ _in_blacklist (const bson_t *test, char **blacklist, uint32_t blacklist_len)
 
    return false;
 }
-static bool
-_should_skip_due_to_server_39704 (const bson_t *test)
```
kevinAlbs, line 1508: Yay for less skip code!

PR #629, opened 2020-06-03
CDRIVER-3701 fix TRACE macro warnings

kevinAlbs:
Fix gcc warnings for `TRACE` macros with a single string argument (and trailing NULL).

clang doesn't warn for this, but I saw the same warnings with gcc 5.4 in a VM.

Also fixes a couple other warnings I observed, and adds `-Werror` to the evergreen task that builds with tracing. That builds with gcc 5.4. It was already logging warnings:
https://evergreen.mongodb.com/task_log_raw/mongo_c_driver_releng_compile_tracing_e0be759b857bbafb7cdfa9b543c927e94ff27871_20_06_03_18_22_13/0?type=T#L505 but was not treating them as build failures.

### DIFF
--- a/build/evergreen_config_lib/tasks.py
+++ b/build/evergreen_config_lib/tasks.py
@@ -251,7 +251,7 @@ all_tasks = [
                           exit 123
                         fi''')]),
     CompileTask('compile-tracing',
-                TRACING='ON'),
+                TRACING='ON', CFLAGS='-Werror -Wno-cast-align'),
     CompileTask('release-compile',
                 config='release',
                 depends_on=OD([('name', 'make-release-archive'),
--- a/src/libmongoc/src/mongoc/mongoc-crypt.c
+++ b/src/libmongoc/src/mongoc/mongoc-crypt.c
@@ -55,6 +55,9 @@ _log_callback (mongocrypt_log_level_t mongocrypt_log_level,
    case MONGOCRYPT_LOG_LEVEL_TRACE:
       log_level = MONGOC_LOG_LEVEL_TRACE;
       break;
+   default:
+      log_level = MONGOC_LOG_LEVEL_CRITICAL;
+      break;
    }
 
    mongoc_log (log_level, MONGOC_LOG_DOMAIN, "%s", message);
--- a/src/libmongoc/src/mongoc/mongoc-stream-tls-secure-channel.c
+++ b/src/libmongoc/src/mongoc/mongoc-stream-tls-secure-channel.c
@@ -107,7 +107,7 @@ _mongoc_stream_tls_secure_channel_destroy (mongoc_stream_t *stream)
     * Shutting Down an Schannel Connection
     */
 
-   TRACE ("shutting down SSL/TLS connection", NULL);
+   TRACE ("%s", "shutting down SSL/TLS connection");
 
    if (secure_channel->cred && secure_channel->ctxt) {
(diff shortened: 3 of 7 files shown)


build/evergreen_config_lib/tasks.py
```diff
@@ -251,7 +251,7 @@ def __init__(self, task_name, extra_commands, orchestration=True, **kwargs):
                           exit 123
                         fi''')]),
     CompileTask('compile-tracing',
-                TRACING='ON'),
+                TRACING='ON', CFLAGS='-Werror -Wno-cast-align'),
```
kevinAlbs, line 254: Copying the `-Wno-cast-align` from the debug-compile-with-warnings task. I believe we expect those when compiling with the alignment attributes.

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -55,6 +55,9 @@ _log_callback (mongocrypt_log_level_t mongocrypt_log_level,
    case MONGOCRYPT_LOG_LEVEL_TRACE:
       log_level = MONGOC_LOG_LEVEL_TRACE;
       break;
+   default:
+      log_level = MONGOC_LOG_LEVEL_CRITICAL;
+      break;
```
kevinAlbs, line 60: Also observed a warning on gcc 5.4 locally:

```
/vagrant/mongo-c-driver/src/libmongoc/src/mongoc/mongoc-crypt.c:42:4: error: switch missing default case [-Werror=switch-default]
    switch (mongocrypt_log_level) {
    ^
```

It seemed reasonable to me to consider an unknown libmongocrypt log level as critical.

src/libmongoc/src/mongoc/mongoc-stream-tls-secure-channel.c
```diff
@@ -107,7 +107,7 @@ _mongoc_stream_tls_secure_channel_destroy (mongoc_stream_t *stream)
     * Shutting Down an Schannel Connection
     */
 
-   TRACE ("shutting down SSL/TLS connection", NULL);
+   TRACE ("%s", "shutting down SSL/TLS connection");
```
kevinAlbs, line 110: This is Windows specific code, so gcc did not warn about this. But I did a mass regex replace. And might as well make it consistent.

src/libmongoc/tests/test-mongoc-topology.c
```diff
@@ -40,7 +40,7 @@ checks_cleanup (checks_t *checks)
 static bool
 checks_cmp (checks_t *checks, const char *metric, char cmp, int expected)
 {
-   int actual;
+   int actual = 0;
```
kevinAlbs, line 43: Another gcc warning fix. It was complaining about L61 being a potential read of uninitialized "actual". Though that should not happen since L55 aborts.

src/libmongoc/src/mongoc/mongoc-stream-tls-secure-channel.c
jmikola: Ah, I totally missed that `msg` was merely concatenated with the format string in the [`TRACE`](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/mongoc-trace-private.h#L39) definition. I thought it was itself an argument.

This seems very sensible and an easy way to fix things.

PR #632, opened 2020-06-07
Cdriver 3704

rcsanchez97:
The two commits in this PR:
* address the bug identified in CDRIVER-3704
* fix the persistent failure in the `debian-package-build` task by installing build dependencies from within the chroot environment

Evergreen patch build: https://evergreen.mongodb.com/version/5edc5eb0306615560ebd3eff

### DIFF
--- a/.evergreen/debian_package_build.sh
+++ b/.evergreen/debian_package_build.sh
@@ -41,10 +41,10 @@ cd ..
 
 git clone https://salsa.debian.org/installer-team/debootstrap.git debootstrap.git
 export DEBOOTSTRAP_DIR=`pwd`/debootstrap.git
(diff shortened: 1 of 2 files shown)


.evergreen/debian_package_build.sh
```diff
@@ -41,10 +41,10 @@ cd ..
 
 git clone https://salsa.debian.org/installer-team/debootstrap.git debootstrap.git
 export DEBOOTSTRAP_DIR=`pwd`/debootstrap.git
-# perl-openssl-defaults is explicitly listed to work around https://bugs.debian.org/907015
```
kevinAlbs, line 44: Ah nice, since that bug was closed, I suppose we no longer need `perl-openssl-defaults`.

rcsanchez97: That's correct.

PR #637, opened 2020-06-11
CDRIVER-3708 skip two tests on latest auth replset

kevinAlbs:
Until SERVER-48715 is resolved, skip tests with auth against a >4.4 replica set that rely on arbiters being discovered.

Also skips the test failure described in CDRIVER-3709.

[CDRIVER-3708](https://jira.mongodb.org/browse/CDRIVER-3708) has more context.

### DIFF
--- a/src/libmongoc/tests/test-mongoc-client.c
+++ b/src/libmongoc/tests/test-mongoc-client.c
@@ -2411,20 +2411,18 @@ test_mongoc_client_get_description_pooled (void)
    _test_mongoc_client_get_description (true);
 }
 
-
 static void
-test_mongoc_client_descriptions (void)
+test_mongoc_client_descriptions_single (void)
 {
    mongoc_client_t *client;
-   mongoc_client_pool_t *pool;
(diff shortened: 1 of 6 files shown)


src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -3969,8 +3980,16 @@ test_client_install (TestSuite *suite)
    TestSuite_AddLive (suite,
                       "/Client/get_description/pooled",
                       test_mongoc_client_get_description_pooled);
-   TestSuite_AddLive (
-      suite, "/Client/descriptions", test_mongoc_client_descriptions);
+   TestSuite_AddLive (suite,
+                      "/Client/descriptions/single",
+                      test_mongoc_client_descriptions_single);
```
kevinAlbs, line 3985: This does not fail on single, ironically due to a bug in libmongoc: https://jira.mongodb.org/browse/CDRIVER-3716 which causes libmongoc not to send a $clusterTime in the first ismaster command to the arbiter.

PR #640, opened 2020-06-18
CDRIVER-3322 warning log with destroying unused session in pooled mode

awitten1:
When in pooling mode, destroying a session that had never been used produced a warning log.  Fixed by not pushing unused sessions to session_pool for being reused because server is unaware of the existence of the session.

### DIFF
--- a/src/libmongoc/tests/test-mongoc-client-pool.c
+++ b/src/libmongoc/tests/test-mongoc-client-pool.c
@@ -361,33 +361,48 @@ test_client_pool_destroy_without_pushing (void)
    bson_destroy (cmd);
 }
 
-/* tests that creating and destroying session is ok */
+static void
+command_started_cb (const mongoc_apm_command_started_t *event)
+{
+   int *count;
+
+   if (strcmp (mongoc_apm_command_started_get_command_name (event),
+               "endSessions") != 0) {
+      return;
+   }
+
+   count = (int *) mongoc_apm_command_started_get_context (event);
+   count++;
+}
+
+/* tests that creating and destroying an unused session
+ * in pooled mode does not result in an error log */
 static void
 test_client_pool_create_unused_session (void)
 {
    mongoc_client_t *client;
    mongoc_client_pool_t *pool;
-   mongoc_uri_t *uri;
    mongoc_client_session_t *session;
-
+   mongoc_apm_callbacks_t *callbacks;
    bson_error_t error;
+   int count = 0;
 
    capture_logs (true);
 
-   uri = mongoc_uri_new_with_error ("mongodb://127.0.0.1/", &error);
-   ASSERT_NO_CAPTURED_LOGS (error.message);
-
-   pool = mongoc_client_pool_new (uri);
+   callbacks = mongoc_apm_callbacks_new ();
+   pool = test_framework_client_pool_new ();
    client = mongoc_client_pool_pop (pool);
    session = mongoc_client_start_session (client, NULL, &error);
-   ASSERT_NO_CAPTURED_LOGS (error.message);
+
+   mongoc_apm_set_command_started_cb (callbacks, command_started_cb);
+   mongoc_client_pool_set_apm_callbacks (pool, callbacks, &count);
 
    mongoc_client_session_destroy (session);
-   ASSERT_NO_CAPTURED_LOGS (error.message);
    mongoc_client_pool_push (pool, client);
-   ASSERT_NO_CAPTURED_LOGS (error.message);
    mongoc_client_pool_destroy (pool);
-   ASSERT_NO_CAPTURED_LOGS (error.message);
+   mongoc_apm_callbacks_destroy (callbacks);
+   ASSERT_CMPINT (count, ==, 0);
+   ASSERT_NO_CAPTURED_LOGS ("mongoc_client_pool_destroy");
 }
 
 void
--- a/src/libmongoc/tests/test-mongoc-client-pool.c
+++ b/src/libmongoc/tests/test-mongoc-client-pool.c
@@ -361,15 +361,41 @@ test_client_pool_destroy_without_pushing (void)
    bson_destroy (cmd);
 }
 
+/* tests that creating and destroying session is ok */
+static void
+test_mongoc_client_pool_create_unused_session (void)
+{
+   mongoc_client_t *client;
+   mongoc_client_pool_t *pool;
+   mongoc_uri_t *uri;
+   mongoc_client_session_t *session;
+
+   bson_error_t error;
+
+   capture_logs (true);
+
+   uri = mongoc_uri_new_with_error ("mongodb://127.0.0.1/", &error);
+   ASSERT_NO_CAPTURED_LOGS (error.message);
+
+   pool = mongoc_client_pool_new (uri);
+   client = mongoc_client_pool_pop (pool);
+   session = mongoc_client_start_session (client, NULL, &error);
+   ASSERT_NO_CAPTURED_LOGS (error.message);
+
+   mongoc_client_session_destroy (session);
+   mongoc_client_pool_push (pool, client);
+   mongoc_client_pool_destroy (pool);
+   ASSERT_NO_CAPTURED_LOGS (error.message);
+}
+
 void
 test_client_pool_install (TestSuite *suite)
 {
    TestSuite_Add (suite, "/ClientPool/basic", test_mongoc_client_pool_basic);
    TestSuite_Add (
       suite, "/ClientPool/try_pop", test_mongoc_client_pool_try_pop);
-   TestSuite_Add (suite,
-                  "/ClientPool/pop_timeout",
-                  test_mongoc_client_pool_pop_timeout);
+   TestSuite_Add (
+      suite, "/ClientPool/pop_timeout", test_mongoc_client_pool_pop_timeout);
    TestSuite_Add (suite,
                   "/ClientPool/min_size_zero",
                   test_mongoc_client_pool_min_size_zero);
@@ -384,9 +410,14 @@ test_client_pool_install (TestSuite *suite)
    TestSuite_Add (
       suite, "/ClientPool/handshake", test_mongoc_client_pool_handshake);
 
+   TestSuite_Add (suite,
+                  "/ClientPool/create_unused_session",
+                  test_mongoc_client_pool_create_unused_session);
 #ifndef MONGOC_ENABLE_SSL
    TestSuite_Add (
       suite, "/ClientPool/ssl_disabled", test_mongoc_client_pool_ssl_disabled);
 #endif
-   TestSuite_AddLive (suite, "/ClientPool/destroy_without_push", test_client_pool_destroy_without_pushing);
+   TestSuite_AddLive (suite,
+                      "/ClientPool/destroy_without_push",
+                      test_client_pool_destroy_without_pushing);
 }
--- a/src/libmongoc/tests/test-mongoc-client-pool.c
+++ b/src/libmongoc/tests/test-mongoc-client-pool.c
@@ -363,7 +363,7 @@ test_client_pool_destroy_without_pushing (void)
 
 /* tests that creating and destroying session is ok */
 static void
-test_mongoc_client_pool_create_unused_session (void)
+test_client_pool_create_unused_session (void)
 {
    mongoc_client_t *client;
    mongoc_client_pool_t *pool;
@@ -383,7 +383,9 @@ test_mongoc_client_pool_create_unused_session (void)
    ASSERT_NO_CAPTURED_LOGS (error.message);
 
    mongoc_client_session_destroy (session);
+   ASSERT_NO_CAPTURED_LOGS (error.message);
    mongoc_client_pool_push (pool, client);
+   ASSERT_NO_CAPTURED_LOGS (error.message);
    mongoc_client_pool_destroy (pool);
    ASSERT_NO_CAPTURED_LOGS (error.message);
 }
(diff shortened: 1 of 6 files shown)


src/libmongoc/tests/test-mongoc-client-pool.c
```diff
@@ -361,15 +361,43 @@ test_client_pool_destroy_without_pushing (void)
...
+   mongoc_client_t *client;
+   mongoc_client_pool_t *pool;
+   mongoc_uri_t *uri;
+   mongoc_client_session_t *session;
+
+   bson_error_t error;
+
+   capture_logs (true);
+
+   uri = mongoc_uri_new_with_error ("mongodb://127.0.0.1/", &error);
```
bazile-clyde, line 377: I think this is the cause of the memory leak on evergreen [here](https://evergreen.mongodb.com/task_log_raw/mongo_c_driver_asan_ubuntu_test_asan_latest_server_auth_nosasl_openssl_cse_patch_43c5aaf21185156cc9e56ac25ac105c51f0a44be_5eebc80ec9ec44013775ff97_20_06_18_20_01_19/0?type=T#L7251). Because `mongoc_uri_new_with_error` calls `bson_malloc0`, which allocates and zeros out memory, and sets this uri to that address we have to free it. You should just be able to add `mongoc_uri_destroy (uri);`  below to fix it.

src/libmongoc/tests/test-mongoc-client-pool.c
```diff
@@ -361,15 +361,43 @@ test_client_pool_destroy_without_pushing (void)
...
+   ASSERT_NO_CAPTURED_LOGS (error.message);
+
+   pool = mongoc_client_pool_new (uri);
```
kevinAlbs, line 380: In addition to checking that nothing was logged, I suggest also asserting that "endSessions" was never sent.

Other tests in `test-mongoc-client-session.c` make similar assertions by setting APM callbacks in `endsessions_test_init`.

There is some background on APM events in the documentation here: http://mongoc.org/libmongoc/current/application-performance-monitoring.html

For this test, callbacks can be set on the pool, like:

```
mongoc_apm_callbacks_t *callbacks;
int count = 0;

callbacks = mongoc_apm_callbacks_new ();
mongoc_apm_set_command_started_cb (callbacks, command_started_cb);
mongoc_client_pool_set_apm_callbacks (pool, callbacks, &count);
```

Then, define `command_started_cb` like so:
```
static void
command_started_cb (const mongoc_apm_command_started_t *event)
{
   int *count;

   if (strcmp (mongoc_apm_command_started_get_command_name (event),
               "endSessions") != 0) {
      return;
   }

   count = (int *) mongoc_apm_command_started_get_context (event);
   count++;
}
```

`count` is passed through the `void *` context. After the test calls `mongoc_client_pool_destroy`, assert that `count` is still 0 with `ASSERT_CMPINT(count, ==, 0)`.

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -793,12 +793,10 @@ _test_end_sessions_many (bool pooled)
    /*
     * sessions were ended on the server, ten thousand at a time
     */
-   ASSERT_CMPINT (test.started_calls, ==, 2);
-   ASSERT_CMPINT (test.succeeded_calls, ==, 2);
+   ASSERT_CMPINT (test.started_calls, ==, 1);
+   ASSERT_CMPINT (test.succeeded_calls, ==, 1);
```
kevinAlbs, line 797: Thinking more about it, I'd prefer if we kept this as a test of the behavior of `endSessions` batching 10,000 sessions at a time.

Your new test checks that we do not send `endSessions` when there are no used sessions. But we should still be testing that if the client has _used_ more than 10,000 sessions, the `endSessions` commands batch up in groups of 10,000.

I suggest adding to the loop above that creates the sessions to use the sessions in an operation. Then these assertions can remain as they were.

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -1213,7 +1211,7 @@ check_session_returned (session_test_t *test, const bson_t *lsid)
       }
    }
 
-   if (!found) {
+   if (!found && ss && ss->last_used_usec != SESSION_NEVER_USED) {
```
kevinAlbs, line 1214: This makes sense, we explicitly are deciding not to return unused sessions to the pool. Can you add a comment here to that effect?

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -2224,8 +2222,8 @@ test_cursor_implicit_session (void *ctx)
 
    /* push a new server session into the pool */
    mongoc_client_session_destroy (cs);
```
kevinAlbs, line 2224: I think we should use the `cs` client session so it does get added to the pool. You can use it by appending it to a command's options. With that, I think we can keep these assertions the same.

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -781,6 +780,13 @@ _test_end_sessions_many (bool pooled)
     */
    for (i = 0; i < sizeof sessions / sizeof (mongoc_client_session_t *); i++) {
       sessions[i] = mongoc_client_start_session (client, NULL, &error);
+      mongoc_client_command_with_opts (client,
+                                       "admin",
+                                       tmp_bson ("{'ping': 1}"),
```
kevinAlbs, line 785: This is more of an internal detail. But instead of using `tmp_bson` in the loop, I suggest creating a `bson_t` another way.

`tmp_bson` is a test convenience. Each time it is called, a `bson_t` is allocated on the heap and added to a global list. Currently, this does not get cleared until process exit. I'm not sure if there'd be any observable performance hit for the test runner, but creating 10,000 seems unnecessary.

If my suggestion below works out (sending ping instead of modifying `last_used_usec`), I suggest adding a helper to send `ping` and using that in both places. This could construct the ping command without `tmp_bson`:

```
static void send_ping (mongoc_client_t *client, mongoc_client_session_t *client_session) {
   bson_t ping_cmd = BSON_INITIALIZER;
   
   BCON_APPEND (&cmd, "ping", BCON_INT32 (1));
   /* ... */
}
```

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -2220,10 +2231,15 @@ test_cursor_implicit_session (void *ctx)
    ASSERT_POOL_SIZE (topology, 0);
    ASSERT_SESSIONS_MATCH (&test->sent_lsid, &find_lsid);
 
-   /* push a new server session into the pool */
+   /* push a new server session into the pool.  server session is only pushed
+    * if it is used.  therefore mark session as used prior to
+    * destroying session.  it isn't possible to use the server session
+    * to send a command because all messages must use the same lsid (for this
+    * test). */
```
kevinAlbs, line 2238: Looking into this test a bit more, notice how line 2245 resets the `lsid` that is captured in the APM events:
```
bson_reinit (&test->sent_lsid);
```
I believe the intention there is to bypass the assertion in the `started` callback, so the test can make assertions. This test is checking that a cursor ties to one session for all of the commands it sends. It checks that the same `lsid` is sent for the initial `find` command and subsequent `getMore` commands. It pushes a session into the pool in between the commands to ensure that the test doesn't pass just because the `getMore` obtains the same session from the pool.

I'd like to try once more for sending `ping` again to mark the session as used instead of overriding `last_used_usec`. I believe calling `bson_reinit (&test->sent_lsid);` before sending the `ping` should bypass the assertion.

If that doesn't work out, I'm ok with overriding `last_used_usec`. It is just a little less preferable to override since it is modifying a private field.

src/libmongoc/tests/test-mongoc-client-pool.c
```diff
@@ -384,9 +427,14 @@ test_client_pool_install (TestSuite *suite)
    TestSuite_Add (
       suite, "/ClientPool/handshake", test_mongoc_client_pool_handshake);
 
+   TestSuite_Add (suite,
+                  "/ClientPool/create_client_pool_unused_session",
+                  test_client_pool_create_unused_session);
```
kevinAlbs, line 432: Ah I missed this initially. `TestSuite_Add` adds this test to always run. This test requires running against a live server that supports sessions.

That is the cause of task failures on Evergreen like [this one](https://evergreen.mongodb.com/task/mongo_c_driver_asan_ubuntu_debug_compile_asan_clang_patch_43c5aaf21185156cc9e56ac25ac105c51f0a44be_5ef2622b2fbabe7b1c89a22e_20_06_23_20_12_29):
```
[2020/06/23 20:14:23.716] Begin /ClientPool/create_client_pool_unused_session, seed 2599939822
[2020/06/23 20:14:33.220] error calling ismaster: 'No suitable servers found: `serverselectiontimeoutms` timed out: [connection refused calling ismaster on 'localhost:27017']'
[2020/06/23 20:14:33.220] URI = mongodb://localhost:27017
```
The test is attempting to run, but the test runner in the compile task is meant to only run tests that do not require a live server (mock server tests and unit tests).

I believe this should do the trick:
```
TestSuite_AddFull (suite,
                    "/ClientPool/create_client_pool_unused_session",
                    test_client_pool_create_unused_session,
                    NULL /* dtor */,
                    NULL /* ctx */,
                    test_framework_skip_if_no_sessions);
```

You can verify this locally by running the test without a server running. The test should skip instead of failing.

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -3013,13 +3036,14 @@ test_session_install (TestSuite *suite)
...
+   TestSuite_AddFull (
+      suite,
+      "/Session/dirty",
+      test_session_dirty,
+      NULL /* dtor */,
+      NULL /* ctx */,
+      test_framework_skip_if_no_sessions,
+      test_framework_skip_if_no_failpoint,
+      /* Tests with retryable writes, requires non-standalone. */
+      test_framework_skip_if_single);
```
kevinAlbs, line 3048: As a note, we do not (yet) have automated lint tasks. So sometimes clang-format of a file will inadvertently change code you did not modify. This looks like the right output for clang-format 4.0.1.

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -781,6 +780,13 @@ _test_end_sessions_many (bool pooled)
...
    for (i = 0; i < sizeof sessions / sizeof (mongoc_client_session_t *); i++) {
       sessions[i] = mongoc_client_start_session (client, NULL, &error);
+      mongoc_client_command_with_opts (client,
+                                       "admin",
+                                       tmp_bson ("{'ping': 1}"),
+                                       NULL,
+                                       &sessions[i]->server_session->lsid,
+                                       NULL,
+                                       &error);
       ASSERT_OR_PRINT (sessions[i], error);
```
kevinAlbs, line 790: This assertion should be above the `ping`.

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -781,6 +780,13 @@ _test_end_sessions_many (bool pooled)
     */
    for (i = 0; i < sizeof sessions / sizeof (mongoc_client_session_t *); i++) {
       sessions[i] = mongoc_client_start_session (client, NULL, &error);
+      mongoc_client_command_with_opts (client,
```
kevinAlbs, line 783: Check the bool return value to ensure this command succeeds:

```
ret = mongoc_client_command_with_opts (...);
ASSERT_OR_PRINT (ret, error);
```

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -799,7 +820,7 @@ _test_end_sessions_many (bool pooled)
    endsessions_test_get_ended_lsids (&test, 0, &ended_lsids);
    ASSERT_CMPINT (bson_count_keys (&ended_lsids), ==, 10000);
    endsessions_test_get_ended_lsids (&test, 1, &ended_lsids);
-   ASSERT_CMPINT (bson_count_keys (&ended_lsids), ==, 1);
+   ASSERT_CMPINT (bson_count_keys (&ended_lsids), ==, 2);
```
kevinAlbs, line 823: I do not think this assertion should change. This is asserting that the second "endSessions" command ends one session. AFAICT there should still be 10001 server sessions on the pool by the end of this test, not 10002. Modifying `send_ping` above may resolve that.

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -757,6 +756,27 @@ test_end_sessions_pooled (void *ctx)
...
+   bson_error_t error;
+   bool ret;
+
+   BCON_APPEND (&ping_cmd, "ping", BCON_INT32 (1));
+
+   ret = mongoc_client_command_with_opts (client,
+                                          "admin",
+                                          &ping_cmd,
+                                          NULL,
+                                          &client_session->server_session->lsid,
```
kevinAlbs, line 774: Instead of using the `lsid` document directly (which is private), append the session into the options like so:

```
mongoc_client_session_append (client_session, &opts);
ret = mongoc_client_command_with_opts (client, ... &opts, NULL /* reply */, error);
```

Note, `mongoc_client_session_append` appends the id associated with the `mongoc_client_session_t`, which libmongoc uses to look up the `mongoc_client_session_t`. It does not append the `lsid` of the server session directly, so the behavior will differ.

kevinAlbs: See also: https://github.com/mongodb/mongo-c-driver/blob/1.16.0/src/libmongoc/tests/test-mongoc-client-session.c#L633-L640 for an example.

kevinAlbs:
Since Clyde's comment was addressed, you can proceed to merge.

PR #645, opened 2020-06-25
CDRIVER-3476 lower case options

awitten1:
Lower case options set with mongoc_uri_set_options_as_* functions.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-uri.c
+++ b/src/libmongoc/src/mongoc/mongoc-uri.c
@@ -123,12 +123,11 @@ count_dots (const char *s)
    return n;
 }
 
-/* return a lowercased string.  must be freed */
 static char *
-lowercase_const_str(const char *key)
+lowercase_str_new (const char *key)
 {
-   char *ret = bson_strdup(key);
-   mongoc_lowercase(key, ret);
+   char *ret = bson_strdup (key);
+   mongoc_lowercase (key, ret);
    return ret;
 }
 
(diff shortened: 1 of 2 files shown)


src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -123,6 +123,14 @@ count_dots (const char *s)
    return n;
 }
 
+/* return a lowercased string.  must be freed */
+static char *
+lowercase_const_str(const char *key)
+{
+   char *ret = bson_strdup(key);
+   mongoc_lowercase(key, ret);
```
kevinAlbs, line 131: Run clang-format.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -2795,6 +2809,7 @@ _mongoc_uri_set_option_as_int64_with_error (mongoc_uri_t *uri,
    const char *option;
    const bson_t *options;
    bson_iter_t iter;
+   char *lowercase_option = NULL;
```
kevinAlbs, line 2812: Minor, but rename to `option_lowercase` for consistency.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -2989,6 +3008,7 @@ mongoc_uri_set_option_as_utf8 (mongoc_uri_t *uri,
 {
    const char *option;
    size_t len;
+   char *lowercase_option = NULL;
```
kevinAlbs, line 3011: And here, `option_lowercase`.

src/libmongoc/tests/test-mongoc-uri.c
```diff
@@ -2720,6 +2720,22 @@ test_one_tls_option_enables_tls ()
    }
 }
 
+static void
+test_casing_options ()
+{
+   mongoc_uri_t* uri;
+   bson_error_t error;
+   capture_logs(true);
```
kevinAlbs, line 2728: It should not be necessary to capture logs.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -123,6 +123,14 @@ count_dots (const char *s)
    return n;
 }
 
+/* return a lowercased string.  must be freed */
+static char *
+lowercase_const_str(const char *key)
```
bazile-clyde, line 128: Throughout our codebase we the suffix `_new` in function names to signal to other developers that whatever the function returns is _newly allocated_ and must be freed. That being said, if this was renamed to something like `lowercase_str_new` we wouldn't need the comment above. Since the functions' name would signal: 1) it's returning a lowercase str 2) it needs to be freed. But I'm fine with the way it is currently too.

bazile-clyde:
LGTM!

awitten1:
Ok sounds good!  Will do next time.

PR #647, opened 2020-06-26
CDRIVER-3612 do not accept explicit session for collection count

rcsanchez97:
Full patch build: https://spruce.mongodb.com/version/5ef65493d1fe072a882ec1cf/tasks

### DIFF
--- a/src/libmongoc/tests/test-mongoc-collection.c
+++ b/src/libmongoc/tests/test-mongoc-collection.c
@@ -2621,6 +2621,18 @@ test_estimated_document_count (void)
    ASSERT_OR_PRINT (123 == future_get_int64_t (future), error);
    ASSERT_MATCH (&reply, server_reply);
 
+   future_destroy (future);
+
+   /* CDRIVER-3612: ensure that an explicit session triggers a client error */
+   future = future_collection_estimated_document_count (
+      collection, tmp_bson ("{'sessionId': 123}"), NULL, &reply, &error);
+
+   ASSERT (-1 == future_get_int64_t (future));
+   ASSERT_ERROR_CONTAINS (error,
+                          MONGOC_ERROR_COMMAND,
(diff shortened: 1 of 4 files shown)


src/libmongoc/tests/test-mongoc-collection.c
```diff
@@ -2621,6 +2621,18 @@ test_estimated_document_count (void)
    ASSERT_OR_PRINT (123 == future_get_int64_t (future), error);
    ASSERT_MATCH (&reply, server_reply);
 
+   future_destroy (future);
+
+   /* CDRIVER-3612: ensure that an explicit session triggers a client error */
+   future = future_collection_estimated_document_count (
```
kevinAlbs, line 2627: Since the mock server is not contacted, you can call `mongoc_collection_estimated_document_count` directly.

rcsanchez97:
On Mon, Jun 29, 2020 at 12:36:18PM -0700, Kevin Albertson wrote:
>    @kevinAlbs approved this pull request.
> 
>    Optional suggestion but otherwise LGTM!
> 
>    FYI the evergreen patch build attached to the PR can be reconfigured to
>    run a full patch build if you do not want to run them separately.
> 
I know.  However, I normally run a patch build (usually multiple) to
confirm that I am ready to actually create a PR.  By the time the PR is
created, the patch build is in the past, so I just go ahead and paste a
link to the build so that it doesn't go to waste.

kevinAlbs:
Ah that makes sense. You can also create the PR and wait for the patch
build before requesting reviewers, but either way is fine.

src/libmongoc/tests/test-mongoc-collection.c
rcsanchez97: My preference is leave it this way to keep the test pattern uniform.

PR #648, opened 2020-06-27
CDRIVER-3535 integration tests + RTT test

kevinAlbs:
Adds support for all SDAM integration tests and the prose RTT test.

This does not include the connectTimeoutMS=0 test introduced in [CDRIVER-3721](https://jira.mongodb.org/browse/CDRIVER-3721). libmongoc does not interpret `connectTimeoutMS=0` as an unlimited timeout, and instead interprets `0` as the option being unset (so it defaults to `10,000`). I'll address that as part of CDRIVER-3721 instead of folding it in here.

The JSON files are on the second commit.

### DIFF
--- a/src/libmongoc/tests/json-test-operations.c
+++ b/src/libmongoc/tests/json-test-operations.c
@@ -50,6 +50,118 @@ session_from_name (json_test_ctx_t *ctx, const char *session_name)
    }
 }
 
+static json_test_worker_thread_t *
+thread_from_name (json_test_ctx_t *ctx, const char *thread_name)
+{
+   json_test_worker_thread_t *wt = NULL;
+   BSON_ASSERT (thread_name);
+
+   if (!strcmp (thread_name, "thread1")) {
+      wt = ctx->worker_threads[0];
+   } else if (!strcmp (thread_name, "thread2")) {
+      wt = ctx->worker_threads[1];
+   } else {
+      test_error ("Unrecognized thread name: %s", thread_name);
+   }
+   if (!wt) {
+      test_error ("Test runner did not create worker threads");
+   }
+   return wt;
+}
+
+json_test_worker_thread_t *
+worker_thread_new (json_test_ctx_t *ctx)
+{
+   json_test_worker_thread_t *wt;
+
+   wt = bson_malloc0 (sizeof (json_test_worker_thread_t));
+   wt->ctx = ctx;
+   /* Note: thread safety of the queue is a bit redundant since the worker
+    * thread locks around it. */
+   wt->queue = q_new ();
+   bson_mutex_init (&wt->mutex);
+   mongoc_cond_init (&wt->cond);
+   return wt;
+}
+
+void
+worker_thread_destroy (json_test_worker_thread_t *wt)
+{
+   q_destroy (wt->queue);
+   bson_mutex_destroy (&wt->mutex);
+   mongoc_cond_destroy (&wt->cond);
+   bson_free (wt);
+}
+
+#define WORKER_THREAD_WAIT_TIMEOUT_MS 10 * 1000
+
+static BSON_THREAD_FUN (json_test_worker_thread_run, wt_void)
+{
+   json_test_worker_thread_t *wt;
+   bson_t *op = NULL;
+   bool shutdown = false;
+
+   wt = wt_void;
+
+   while (true) {
+      /* Wait for a new item in the queue or shutdown. */
+      bson_mutex_lock (&wt->mutex);
+      op = (bson_t *) q_get_nowait (wt->queue);
+      shutdown = wt->shutdown_requested;
+      while (!op && !shutdown) {
+         mongoc_cond_timedwait (
+            &wt->cond, &wt->mutex, WORKER_THREAD_WAIT_TIMEOUT_MS);
+         shutdown = wt->shutdown_requested;
+         op = q_get_nowait (wt->queue);
+      }
+      bson_mutex_unlock (&wt->mutex);
+
+      if (op) {
+         bson_t reply;
+
+         json_test_operation (wt->ctx,
+                              wt->ctx->config->scenario,
+                              op,
+                              wt->ctx->collection,
+                              NULL /* session */,
+                              &reply);
+         bson_destroy (op);
+         op = NULL;
+         bson_destroy (&reply);
+      } else if (shutdown) {
+         /* If there are no queued operations and shutdown was requested, exit
+          * thread. */
+         break;
+      }
+   }
+   BSON_THREAD_RETURN;
+}
+
+static void
+start_thread (json_test_worker_thread_t *wt)
+{
+   wt->shutdown_requested = false;
+   COMMON_PREFIX (thread_create) (&wt->thread, json_test_worker_thread_run, wt);
+}
+
+static void
+run_on_thread (json_test_worker_thread_t *wt, bson_t *op)
+{
+   bson_mutex_lock (&wt->mutex);
+   q_put (wt->queue, bson_copy (op));
+   mongoc_cond_broadcast (&wt->cond);
+   bson_mutex_unlock (&wt->mutex);
+}
+
+static void
+wait_for_thread (json_test_worker_thread_t *wt)
+{
+   bson_mutex_lock (&wt->mutex);
+   wt->shutdown_requested = true;
+   mongoc_cond_broadcast (&wt->cond);
+   bson_mutex_unlock (&wt->mutex);
+   COMMON_PREFIX (thread_join) (wt->thread);
+}
 
 void
 json_test_ctx_init (json_test_ctx_t *ctx,
@@ -106,6 +218,7 @@ json_test_ctx_init (json_test_ctx_t *ctx,
          bson_free (session_opts_path);
       }
    }
+   bson_mutex_init (&ctx->mutex);
 }
 
 
@@ -136,6 +249,7 @@ json_test_ctx_cleanup (json_test_ctx_t *ctx)
    mongoc_uri_destroy (ctx->test_framework_uri);
    bson_destroy (ctx->sent_lsids[0]);
    bson_destroy (ctx->sent_lsids[1]);
+   bson_mutex_destroy (&ctx->mutex);
 }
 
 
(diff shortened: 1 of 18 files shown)


src/libmongoc/tests/json-test-operations.c
```diff
@@ -1990,6 +2098,154 @@ index_exists (mongoc_client_t *client, const bson_t *operation)
...
+   int64_t measured = 0;
+   int64_t total = 0;
+
+   event_name = bson_lookup_utf8 (operation, "arguments.event");
+   count = bson_lookup_int32 (operation, "arguments.count");
+   expires_us = bson_get_monotonic_time () + 10 * 1000 * 1000;
+   while (!satisfied && bson_get_monotonic_time () < expires_us) {
+      int64_t diff;
+
+
```
kevinAlbs, line 2137: Will remove spacing.

src/libmongoc/tests/test-mongoc-sdam.c
```diff
@@ -215,6 +216,250 @@ test_sdam_cb (bson_t *test)
...
+      if (!ret) {
+         MONGOC_DEBUG ("error disabling failpoint: %s", error.message);
+      }
+   }
+
+   bson_destroy (&cmd);
+}
+
+static void
+run_one_integration_test (json_test_config_t *config, bson_t *test)
```
kevinAlbs, line 320: With the introduction of threaded operations, it seemed easier to implement the test runner outside of the general `run_json_general_test`. When there is a unified test runner spec, maybe we can consolidate this.

src/libmongoc/tests/json-test.c
```diff
@@ -1360,6 +1379,15 @@ set_uri_opts_from_bson (mongoc_uri_t *uri, const bson_t *opts)
...
          mongoc_write_concern_destroy (wc);
+      } else if (!strcmp (bson_iter_key (&iter), "appname")) {
+         mongoc_uri_set_option_as_utf8 (
+            uri, MONGOC_URI_APPNAME, bson_iter_utf8 (&iter, NULL));
+      } else if (!strcmp (bson_iter_key (&iter), "connectTimeoutMS")) {
+         mongoc_uri_set_option_as_int32 (
+            uri, MONGOC_URI_CONNECTTIMEOUTMS, bson_iter_int32 (&iter));
+      } else if (!strcmp (bson_iter_key (&iter), "serverSelectionTimeoutMS")) {
+         mongoc_uri_set_option_as_int32 (
+            uri, MONGOC_URI_SERVERSELECTIONTIMEOUTMS, bson_iter_int32 (&iter));
```
jmikola, line 1390: Would it make sense to generalize all of this logic, as we do in PHPC? See: [`php_phongo_apply_options_to_uri`](https://github.com/mongodb/mongo-php-driver/blob/41a96ba0e864450465f1ef41501ea3891cb72c83/php_phongo.c#L1478). Only RC, RP, and WC are handled separately, since those have unique setters. For everything else, we use methods like `mongoc_uri_option_is_int32` and `mongoc_uri_set_option_as_int32`.

That said, perhaps this isn't a concern since you do error on any unsupported option. That removes the risk of a test adding a new option and it being ignored by the libmongoc test runner.

Not suggesting this be addressed in this PR, but perhaps it warrants a todo item (possible `new-eng` work).

src/libmongoc/tests/test-mongoc-sdam.c
```diff
@@ -215,6 +216,250 @@ test_sdam_cb (bson_t *test)
...
+   bson_init (&ctx->events);
+   ctx->acknowledged = true;
+   ctx->verbose = test_framework_getenv_bool ("MONGOC_TEST_MONITORING_VERBOSE");
+   bson_init (&ctx->lsids[0]);
+   bson_init (&ctx->lsids[1]);
+   bson_mutex_init (&ctx->mutex);
+}
+
+static void
+sdam_json_test_ctx_init_client (json_test_ctx_t *ctx, mongoc_client_t *client)
```
jmikola, line 241: Is there any reason to split this up from `sdam_json_test_ctx_init_config` vs. having a single `sdam_json_test_ctx_init` function? They always seem to be called together.

Perhaps this is necessary because of the comment you have below in `run_one_integration_test`: "Initializing the test context in tandem with monitoring is a little tricky."

src/libmongoc/tests/test-mongoc-sdam.c
```diff
@@ -215,6 +216,250 @@ test_sdam_cb (bson_t *test)
...
+
+   bson_destroy (&reply);
+
+   return res;
+}
+
+/* Try to get a completely clean slate by disabling failpoints on all servers.
+ */
+static void
+deactivate_failpoints_on_all_servers (mongoc_client_t *client)
```
jmikola, line 287: Is this compensating for the `configureFailPoint` operation handler not tracking exactly which fail points are set? May be related to my earlier comment in the other file.

As-is, I might worry that this would stop working if any test were to set a fail point other than `failCommand` (I don't believe any currently do that).

src/libmongoc/tests/test-mongoc-sdam.c
```diff
@@ -215,6 +216,250 @@ test_sdam_cb (bson_t *test)
...
+   MONGOC_DEBUG ("running test: %s", bson_lookup_utf8 (test, "description"));
+
+   uri = test_framework_get_uri ();
+   if (bson_has_field (test, "clientOptions")) {
+      bson_t client_opts;
+
+      bson_lookup_doc (test, "clientOptions", &client_opts);
+      set_uri_opts_from_bson (uri, &client_opts);
+   }
+   mongoc_uri_set_option_as_bool (uri, MONGOC_URI_DIRECTCONNECTION, false);
```
jmikola, line 342: Is this necessary? Doesn't the connection string returned by `test_framework_get_uri` already ensure that replica sets also specify a `replicaSet` option that would ensure discovery behavior?

Or does this perhaps have implications for connection strings with only a single host? In that case, I expect the topology would still end up being sharded due to `isdbgrid` being found in the `isMaster` response.

src/libmongoc/tests/test-mongoc-sdam.c
```diff
@@ -215,6 +216,250 @@ test_sdam_cb (bson_t *test)
...
+                                 tmp_bson ("{'killAllSessions': []}"),
+                                 NULL,
+                                 NULL,
+                                 &error);
+
+   insert_data (db_name, coll_name, config->scenario);
+
+   if (bson_has_field (test, "failPoint")) {
+      activate_fail_point (setup_client, 0, test, "failPoint");
+   }
```
jmikola, line 367: AFAIK, this is for the top-level field. Do any of the SDAM tests actually use this, or is it just covering the bases since the SDAM spec readme alludes to the Transactions test format?

src/libmongoc/tests/test-mongoc-sdam.c
```diff
@@ -215,6 +216,250 @@ test_sdam_cb (bson_t *test)
...
+   set_apm_callbacks_pooled (&ctx, pool);
+
+   /* Pop a client, which starts topology scanning. */
+   test_client = mongoc_client_pool_pop (pool);
+
+   sdam_json_test_ctx_init_client (&ctx, test_client);
+
+   /* Set up test contexts for worker threads, which may be used by tests that
+    * have "startThread" operations. */
+   /* These get the same APM event listeners. But fortunately those are
```
jmikola, line 393: Should this block comment be joined with the previous?

src/libmongoc/tests/test-mongoc-sdam.c
```diff
@@ -250,6 +495,15 @@ test_all_spec_tests (TestSuite *suite)
...
                      resolved));
    install_json_test_suite (suite, resolved, &test_sdam_cb);
+
+   /* Integration tests. */
+   ASSERT (realpath (JSON_DIR "/server_discovery_and_monitoring/integration",
+                     resolved));
+   install_json_test_suite_with_check (suite,
+                                       resolved,
+                                       &test_sdam_integration_cb,
+                                       test_framework_skip_if_no_crypto,
```
jmikola, line 505: Do these tests actually care about whether crypto is available, or was this copied from transactions (which require sessions)?

src/libmongoc/tests/test-mongoc-sdam.c
```diff
@@ -443,7 +695,138 @@ test_existing_behavior (void *ctx)
...
+   /* Run a find command for the server to be discovered. */
+   coll = get_test_collection (client, "streamingRttTest");
+   cursor = mongoc_collection_find_with_opts (
+      coll, tmp_bson ("{}"), NULL /* opts */, NULL /* read prefs */);
+   mongoc_cursor_next (cursor, &doc);
+
+   /* Sleep for two seconds to allow multiple heartbeats to succeed. */
+   _mongoc_usleep (2 * 1000 * 1000);
+
+   /* Set a failpoint to make isMaster commands take longer. */
```
jmikola, line 765: [RTT Tests](https://github.com/mongodb/specifications/tree/master/source/server-discovery-and-monitoring/tests#rtt-tests) has a step to "Assert that each ServerDescriptionChangedEvent includes a non-zero RTT." between the 2-second sleep and `configureFailPoint` call. Is that missing?

src/libmongoc/tests/test-mongoc-sdam.c
```diff
@@ -443,7 +695,138 @@ test_existing_behavior (void *ctx)
...
+                BCON_INT32 (500),
+                "appName",
+                "streamingRttTest",
+                "}");
+   ret = mongoc_client_command_simple (
+      client, "admin", &cmd, NULL /* read prefs. */, NULL /* reply */, &error);
+   ASSERT_OR_PRINT (ret, error);
+
+   /* Wait for the server's RTT to exceed 250ms. If this does not happen for 20
+    * seconds, consider it a failure. */
```
jmikola, line 788: Is the 20-second limit necessary? I don't see this mentioned in the test readme.

src/libmongoc/tests/json-test-operations.c
```diff
@@ -1990,6 +2098,154 @@ index_exists (mongoc_client_t *client, const bson_t *operation)
...
+         bson_mutex_unlock (&ctx->mutex);
+      } else {
+         test_error ("Unknown event: %s", event_name);
+      }
+      if (!satisfied) {
+         _mongoc_usleep (10 * 1000);
+      }
+   }
+   if (!satisfied) {
+      test_error ("did not see enough %s events after 10s. %d total occurred. "
```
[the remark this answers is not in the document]
ShaneHarvey: SGTM. Yeah it was an oversight not to add a `timeoutMS` to waitForEvent too.

src/libmongoc/tests/test-mongoc-sdam.c
kevinAlbs: I forgot to update this comment. Doing that in the next commit (along with replacing other times with macros in this function).

src/libmongoc/tests/json-test-operations.c
```diff
@@ -50,6 +50,112 @@ session_from_name (json_test_ctx_t *ctx, const char *session_name)
...
+
+      if (op) {
+         bson_t reply;
+
+         json_test_operation (wt->ctx,
+                              wt->ctx->config->scenario,
+                              op,
+                              wt->ctx->collection,
+                              NULL /* session */,
+                              &reply);
```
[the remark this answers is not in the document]
jmikola: Ah, `mongoc_client_session_with_transaction` makes sense. The unified test runner shouldn't differ much from transactions (at least not functionally), so I think this would stick around either way.

src/libmongoc/tests/json-test-operations.c
```diff
@@ -50,6 +50,112 @@ session_from_name (json_test_ctx_t *ctx, const char *session_name)
...
+      op = (bson_t *) q_get_nowait (wt->queue);
+      shutdown = wt->shutdown_requested;
+      while (!op && !shutdown) {
+         mongoc_cond_timedwait (&wt->cond, &wt->mutex, 10 * 1000);
+         shutdown = wt->shutdown_requested;
+         op = q_get_nowait (wt->queue);
+      }
+      bson_mutex_unlock (&wt->mutex);
+
+      if (op) {
```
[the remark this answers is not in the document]
jmikola: > Your comment made me realize there is a latent bug here

Can you confirm that this was fixed? It looks like you did change the loop to exhaust any queued up ops, but I want to be sure before I resolve :)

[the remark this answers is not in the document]
jmikola: Ah, just got to your comment in https://github.com/mongodb/mongo-c-driver/pull/648#discussion_r448686303

This is good to resolve, then.

src/libmongoc/tests/json-test.c
```diff
@@ -1121,7 +1122,10 @@ check_outcome_collection (mongoc_collection_t *collection, bson_t *test)
       mongoc_collection_set_read_concern (collection, rc);
    }
 
-   cursor = mongoc_collection_find_with_opts (collection, &query, NULL, prefs);
+   bson_init (&opts);
+   BCON_APPEND (&opts, "sort", "{", "_id", BCON_INT32 (1), "}");
+   cursor = mongoc_collection_find_with_opts (collection, &query, &opts, prefs);
+   bson_destroy (&opts);
```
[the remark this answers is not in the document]
jmikola: To confirm, this will be split out into its own commit before merging? I noticed CDRIVER-3666 still wasn't listed in https://github.com/mongodb/mongo-c-driver/pull/648/commits (and the JIRA issue is just "open" with no link to this PR) so wanted to double check. OK if so, in which case you can probably move the JIRA ticket right to closed once its commit is merged.

src/libmongoc/tests/json-test-operations.c
```diff
@@ -2227,8 +2483,66 @@ json_test_operation (json_test_ctx_t *ctx,
...
+      } else if (!strcmp (op_name, "waitForEvent")) {
+         wait_for_event (ctx, operation);
+      } else if (!strcmp (op_name, "assertEventCount")) {
+         assert_event_count (ctx, operation);
+      } else if (!strcmp (op_name, "configureFailPoint")) {
+         mongoc_client_t *client;
+
+         client = mongoc_client_new_from_uri (ctx->client->uri);
+         test_framework_set_ssl_opts (client);
+         activate_fail_point (
```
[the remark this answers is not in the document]
jmikola: OK to punt on this for now, as it clearly doesn't affect any existing tests. Please create a separate CDRIVER ticket to revisit this down the line, though. I expect I will address this in the README for the unified spec test runner, too.

src/libmongoc/tests/test-mongoc-sdam.c
jmikola: Yup. Separate follow-up ticket works for me. I'll resolve this and you can drop a link to the CDRIVER ticket in the other thread (https://github.com/mongodb/mongo-c-driver/pull/648#discussion_r451223867) when you get a chance, and then resolve that accordingly.

src/libmongoc/tests/test-mongoc-sdam.c
jmikola: Noted.

IIRC, libmongoc makes decisions about feature support by server type (in addition to wire version, of course) and never the topology type. That's why things like retryable writes and transactions would still work when directly connected to a replica set primary or mongos node.

src/libmongoc/tests/test-mongoc-sdam.c
jmikola: Found it. Thanks!

src/libmongoc/tests/json-test-operations.c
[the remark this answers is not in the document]
kevinAlbs: Created https://jira.mongodb.org/browse/CDRIVER-3738

src/libmongoc/tests/json-test.c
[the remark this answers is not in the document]
kevinAlbs: Yes, I will revise the commits before merging. I was worried that force pushing to this PR could mess up the current conversations. Marked CDRIVER-3666 as in code review and commented with a link to this PR.

PR #649, opened 2020-06-29
CDRIVER-3702 adds test only asserts for mutex lock

awitten1:
Adds test only assertions that mutex is acquired upon entering certain functions that require that the caller have locked the mutex.  Assertions are enabled with the cmake configuration option -DENABLE_DEBUG_ASSERTIONS=ON.


src/common/common-thread.c
```diff
@@ -27,6 +27,18 @@ int COMMON_PREFIX (thread_join) (bson_thread_t thread)
 {
    return pthread_join (thread, NULL);
 }
+
+bool COMMON_PREFIX (mutex_is_locked) (bson_mutex_t *mutex)
+{
+#ifdef MONGOC_ENABLE_TESTING
```
kevinAlbs, line 33: This is not necessary. If `MONGOC_ENABLE_TESTING` is not defined, then `MONGOC_TEST_ASSERT` expands to `((void) 0)`.

Arguably, `mutex_is_locked` may never be called outside of test assertions. But having it return  `true` when the testing macro is not defined seems like potentially surprising behavior.

src/libmongoc/src/mongoc/mongoc-topology-background-monitoring.c
```diff
@@ -802,6 +802,9 @@ _mongoc_topology_background_monitoring_stop (mongoc_topology_t *topology)
    int i;
    bool is_srv_polling;
 
+   /* Test only assert. Is a noop unless -DMONGOC_ENABLE_TESTING is passed */
+   MONGOC_TEST_ASSERT (COMMON_PREFIX (mutex_is_locked) (&topology->mutex));
```
kevinAlbs, line 806: Additionally, mongoc-server-monitor.c has many functions that document "Caller must hold no locks" or "must hold topology lock" that we can apply this assertion to. You may need to rebase on an updated master branch, since that file is relatively recent (from CDRIVER-3535).

CMakeLists.txt
```diff
@@ -226,6 +227,9 @@ set (CMAKE_MACOSX_RPATH ON)
 # silence a CMake 3.11 warning that the old behavior is deprecated.
 cmake_policy (SET CMP0042 NEW)
 
+include_directories ("${PROJECT_BINARY_DIR}/src")
+include_directories ("${PROJECT_BINARY_DIR}/src/common")
```
kevinAlbs, line 231: I believe the cause of `*-cse` tasks failing is that they are only building libbson, not libmongoc, and do not have the include path to the generated common-config.h.

I would move this into `src/libmongoc/CMakeLists.txt` next to the `include_directories` directive for the source common headers: https://github.com/mongodb/mongo-c-driver/blob/1.16.2/src/libmongoc/CMakeLists.txt/#L465

And similar for src/libbson/CMakeLists.txt.

kevinAlbs: You can test this locally by configuring with `-DENABLE_MONGOC=OFF`:
```
cmake -DENABLE_MONGOC=OFF ..
cmake --build . --target bson_shared
```

kevinAlbs: Ah, I looked into this locally a bit. I think you stumbled onto an existing issue.

Add `add_subdirectory (src/common)` just below `add_subdirectory (src/libbson)` on L236. If only libbson is being configured without libmongoc, cmake is currently ignoring the common directory (and hence, not generating the common header).

You can verify that works locally by configuring and building just `bson_shared`.

src/common/CMakeLists.txt
```diff
@@ -1,8 +1,22 @@
+
+set (MONGOC_ENABLE_TESTING 0)
+
+if (ENABLE_TESTING)
+    set (MONGOC_ENABLE_TESTING 1)
+    unset (ENABLE_TESTING CACHE)
```
kevinAlbs, line 6: Though I am not a fan of cmake caching, since `ENABLE_TRACING` is cached, I'd argue for removing this for consistency. It seems like it'd be less of a surprise if this behaved the same way as the other options.

src/common/common-thread-private.h
```diff
@@ -29,32 +30,78 @@ BSON_BEGIN_DECLS
...
 #define bson_once pthread_once
 #define bson_once_t pthread_once_t
 #define bson_thread_t pthread_t
 #define BSON_THREAD_FUN(_function_name, _arg_name) \
    void *(_function_name) (void *(_arg_name))
 #define BSON_THREAD_FUN_TYPE(_function_name) void *(*(_function_name)) (void *)
 #define BSON_THREAD_RETURN return NULL
+
+/* this macro can be defined as a as a build configuration option
+ * with -DENABLE_TESTING=ON.  it's purpose is to allow for functions
```
kevinAlbs, line 46:
```suggestion
 * with -DENABLE_TESTING=ON.  its purpose is to allow for functions
```

src/libmongoc/src/mongoc/mongoc-thread-private.h
```diff
@@ -29,11 +29,20 @@
 #define mongoc_cond_t pthread_cond_t
 #define mongoc_cond_broadcast pthread_cond_broadcast
 #define mongoc_cond_init(_n) pthread_cond_init ((_n), NULL)
+
+#ifndef MONGOC_ENABLE_TESTING
 #define mongoc_cond_wait pthread_cond_wait
+#else
```
kevinAlbs, line 35: This should only be defined if on Unix.

src/common/common-thread-private.h
```diff
@@ -29,32 +30,78 @@ BSON_BEGIN_DECLS
...
+
+#ifndef MONGOC_ENABLE_TESTING
+
+#define bson_mutex_destroy pthread_mutex_destroy
+#define bson_mutex_init(_n) pthread_mutex_init ((_n), NULL)
+#define bson_mutex_lock pthread_mutex_lock
+#define bson_mutex_t pthread_mutex_t
+#define bson_mutex_unlock pthread_mutex_unlock
+
+#else
```
kevinAlbs, line 59: Since this is only supported on Unix, I suggest changing the condition to be `#if defined(MOGNOC_ENABLE_TESTING) && defined (BSON_OS_UNIX)`.

src/libmongoc/src/mongoc/mongoc-topology-private.h
```diff
@@ -37,6 +37,15 @@
 #define MONGOC_TOPOLOGY_HEARTBEAT_FREQUENCY_MS_SINGLE_THREADED 60000
 #define MONGOC_TOPOLOGY_MIN_RESCAN_SRV_INTERVAL_MS 60000
 
+/*
```
kevinAlbs, line 40: Though we're only using this in libmongoc now, I could see this being useful in libbson too. libmongoc can't reach inside libbson's private headers. But I like what you've done by creating a common-config.h.in. I'd argue for creating another common header, `common-macros.h`, and placing this inside.

kevinAlbs:
`debug-compile-aws` will be fixed after rebasing on https://github.com/mongodb/mongo-c-driver/commit/650ad0a3a5ec5d460b568a8e7e322f72f945c7fd

I have a suggested fix inline for the other tasks.

Can you note in debugging.rst how to configure with test assertions?

src/common/common-thread-private.h
```diff
@@ -29,21 +31,65 @@ BSON_BEGIN_DECLS
...
+   do {                                             \
+      pthread_mutex_lock (&(mutex)->wrapped_mutex); \
+      (mutex)->lock_owner = pthread_self ();        \
+      (mutex)->valid_tid = true;                    \
+   } while (0);
+
+#define bson_mutex_unlock(mutex)                      \
+   do {                                               \
+      pthread_mutex_unlock (&(mutex)->wrapped_mutex); \
+      (mutex)->valid_tid = false;                     \
```
kevinAlbs, line 88: I think `valid_tid` must be set false while the wrapped mutex is still locked. Setting it outside of the lock may interfere with the assertion. After changing that locally and running `/Cursor/kill/pooled` 100 times all succeeded.

CMakeLists.txt
```diff
@@ -246,6 +247,8 @@ add_definitions (-D_GNU_SOURCE)
 add_definitions (-D_BSD_SOURCE)
 add_definitions (-D_DEFAULT_SOURCE)
 
+add_subdirectory (src)
+
```
kevinAlbs, line 251: It seems my suggestion for moving this was not quite right, and is the cause of the `make-release-archive` task failing.

`src/CMakeLists.txt` makes the assumption that `SOURCE_DIR` is defined. But it is defined below on line 260. So it ends up being empty (because cmake, like bash, treats unset variables like empty strings) causing that error when producing the dist tarball. Also, `src/CMakeLists.txt` has comments indicating it assumes other directories have been included.

My new theory:
- Move this `add_subdirectory (src)` back.
- Add `add_subdirectory (src/common)` here, so the common-config.h.in is always configured. This handles the previous include errors when building only libbson with `-DENABLE_MONGOC=OFF`
- Remove `add_subdirectory (common)` from src/CMakeLists.txt

The problematic tasks were tasks that were building with `-DENABLE_MONGOC=OFF` and tasks that were building the release tarball. You can test both of these locally:

To build only libbson:
```
mkdir cmake-build-bsononly
cd cmake-build-bsononly
cmake -DENABLE_MONGOC=OFF ..
cmake --build . --target bson_shared
```

To build the distribution tarball:
```
mkdir cmake-build-dist
cd cmake-build-dist
cmake -DENABLE_BSON=ON -DENABLE_MAN_PAGES=ON -DENABLE_HTML_DOCS=ON ..
cmake --build . --target dist
```

Aside: cmake docs say [SOURCE_DIR is a built-in read-only variable](https://cmake.org/cmake/help/latest/prop_tgt/SOURCE_DIR.html). So that seems fishy... but a problem for another time. Updating our cmake scripts can be quite frustrating sometimes.

src/libmongoc/doc/debugging.rst
```diff
@@ -104,3 +104,9 @@ If you see "bson command installed by lldb_bson" at the beginning of your LLDB s
 
 .. _PyMongo: https://pypi.python.org/pypi/pymongo
 .. _install pip: https://pip.pypa.io/en/stable/installing/#installing-with-get-pip-py)
+
+Test assertions
+---------------
+
+To enable assertions that a mutex is locked upon entering functions that require the caller to lock the mutex, configure
```
kevinAlbs, line 111: I would leave this vague, like `To enable runtime test assertions, configure with ...`. I foresee us adding more types assertions behind this flag.

src/libmongoc/src/mongoc/mongoc-thread-private.h
kevinAlbs: ping

awitten1: Sorry I should have commented on this earlier.  I didn't change this because it is within the `#if defined(BSON_OS_UNIX)` block.  Should add a check for that in the inner if also?

CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ option (ENABLE_PIC
    "Enables building of position independent code for static library components."
    ON
 )
+option (ENABLE_TEST_ASSERTIONS "Turn on mutex lock asserts" OFF)
```
kevinAlbs, line 64: Thanks for renaming this. At the risk of continued bikeshedding, can you rename this to `ENABLE_DEBUG_ASSERTIONS`? I think that conveys better that these are assertions intended for debugging only. And can you reword the help description to be more generic, like "Turn on runtime test assertions for debugging"?

src/common/CMakeLists.txt
```diff
@@ -1,8 +1,22 @@
+
+set (MONGOC_ENABLE_TESTING 0)
```
kevinAlbs, line 2: Can you do a global find/replace to change `MONGOC_ENABLE_TESTING` with `MONGOC_ENABLE_DEBUG_ASSERTIONS`?

src/common/common-thread-private.h
kevinAlbs: Ah, nevermind. This is already nested within an `#if defined(BSON_OS_UNIX)`.

src/libmongoc/doc/debugging.rst
```diff
@@ -104,3 +104,9 @@ If you see "bson command installed by lldb_bson" at the beginning of your LLDB s
 
 .. _PyMongo: https://pypi.python.org/pypi/pymongo
 .. _install pip: https://pip.pypa.io/en/stable/installing/#installing-with-get-pip-py)
+
+Test assertions
+---------------
+
+To enable runtime test assertions, configure with ``-DENABLE_TEST_ASSERTIONS=ON``.
+``MONGOC_TEST_ASSERT`` is a noop if ``-DENABLE_TEST_ASSERTIONS=OFF``.
```
kevinAlbs, line 112: Remove "MONGOC_TEST_ASSERT is a noop if -DENABLE_TEST_ASSERTIONS=OFF". `MONGOC_TEST_ASSERT` is a private macro, so users need not be made aware of it.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -134,6 +134,8 @@ _server_monitor_heartbeat_started (mongoc_server_monitor_t *server_monitor,
                                    bool awaited)
 {
    mongoc_apm_server_heartbeat_started_t event;
+   MONGOC_TEST_ASSERT(!COMMON_PREFIX (mutex_is_locked) (&server_monitor->topology->apm_mutex));
```
kevinAlbs, line 137: Run clang format before merging.

src/libmongoc/src/mongoc/mongoc-thread-private.h
kevinAlbs: Ah nevermind. I had missed that, and in my other comment in `common-thread-private.h`. Resolving.

src/libmongoc/src/mongoc/mongoc-topology-background-monitoring.c
```diff
@@ -272,6 +275,10 @@ _mongoc_topology_background_monitoring_stop (mongoc_topology_t *topology)
    int i;
    bool is_srv_polling;
 
+   /* Test only assert. Is a noop unless -DENABLE_TEST_ASSERTIONS=ON is set
```
kevinAlbs, line 278: I'd argue for moving this comment, (and the ones in mongoc-topology.c), to the definition of `MONGOC_TEST_ASSERT` in common-macros-private.h.

src/common/common-macros-private.h
```diff
@@ -0,0 +1,16 @@
...
+#include "common-prelude.h"
+
+#ifndef MONGO_C_DRIVER_COMMON_MACROS_H
+#define MONGO_C_DRIVER_COMMON_MACROS_H
+
+/*
+ * configure with option -DENABLE_TEST_ASSERTIONS=ON
+ */
+#if defined (MONGOC_ENABLE_TESTING) && defined (BSON_OS_UNIX)
+#define MONGOC_TEST_ASSERT(statement) BSON_ASSERT (statement)
```
kevinAlbs, line 11: I suggest renaming `MONGOC_TEST_ASSERT` to `MONGOC_DEBUG_ASSERT`.

.evergreen/compile-unix.sh
```diff
@@ -134,6 +134,8 @@ if [ "$RELEASE" = "ON" ]; then
    mkdir build-dir
    $TAR xf ../mongoc.tar.gz -C build-dir --strip-components=1
    cd build-dir
+else
+  CONFIGURE_FLAGS="${CONFIGURE_FLAGS} -DENABLE_TEST_ASSERTIONS=ON"
```
kevinAlbs, line 138: Here too, `ENABLE_DEBUG_ASSERTIONS`

.evergreen/compile-windows.sh
```diff
@@ -103,7 +103,7 @@ if [ "$RELEASE" ]; then
    TEST_PATH="./src/libmongoc/RelWithDebInfo/test-libmongoc.exe"
    export PATH=$PATH:`pwd`/src/libbson/RelWithDebInfo:`pwd`/src/libmongoc/RelWithDebInfo:`pwd`/install-dir/bin
 else
-   CONFIGURE_FLAGS="$CONFIGURE_FLAGS"
+   CONFIGURE_FLAGS="${CONFIGURE_FLAGS} -DENABLE_TEST_ASSERTIONS=ON"
```
kevinAlbs, line 106: `ENABLE_DEBUG_ASSERTIONS`

src/common/CMakeLists.txt
```diff
@@ -1,8 +1,22 @@
+
+set (MONGOC_ENABLE_DEBUG_ASSERTIONS 0)
+
+if (ENABLE_TEST_ASSERTIONS)
```
kevinAlbs, line 4: `ENABLE_DEBUG_ASSERTIONS`

src/common/common-macros-private.h
```diff
@@ -0,0 +1,15 @@
+
+#include "common-prelude.h"
+
+#ifndef MONGO_C_DRIVER_COMMON_MACROS_H
+#define MONGO_C_DRIVER_COMMON_MACROS_H
+
+/* Test only assert. Is a noop unless -DENABLE_TEST_ASSERTIONS=ON is set
```
kevinAlbs, line 7: `ENABLE_DEBUG_ASSERTIONS`

src/common/common-thread-private.h
```diff
@@ -29,21 +31,65 @@ BSON_BEGIN_DECLS
...
 #define bson_once pthread_once
 #define bson_once_t pthread_once_t
 #define bson_thread_t pthread_t
 #define BSON_THREAD_FUN(_function_name, _arg_name) \
    void *(_function_name) (void *(_arg_name))
 #define BSON_THREAD_FUN_TYPE(_function_name) void *(*(_function_name)) (void *)
 #define BSON_THREAD_RETURN return NULL
+
+/* this macro can be defined as a as a build configuration option
+ * with -DENABLE_TEST_ASSERTIONS=ON.  its purpose is to allow for functions
```
kevinAlbs, line 47: `ENABLE_DEBUG_ASSERTIONS`

src/libmongoc/doc/debugging.rst
```diff
@@ -104,3 +104,8 @@ If you see "bson command installed by lldb_bson" at the beginning of your LLDB s
 
 .. _PyMongo: https://pypi.python.org/pypi/pymongo
 .. _install pip: https://pip.pypa.io/en/stable/installing/#installing-with-get-pip-py)
+
+Test assertions
+---------------
+
+To enable runtime test assertions, configure with ``-DENABLE_TEST_ASSERTIONS=ON``.
```
kevinAlbs, line 111: `ENABLE_DEBUG_ASSERTIONS`

src/libmongoc/doc/debugging.rst
```diff
@@ -104,3 +104,8 @@ If you see "bson command installed by lldb_bson" at the beginning of your LLDB s
 
 .. _PyMongo: https://pypi.python.org/pypi/pymongo
 .. _install pip: https://pip.pypa.io/en/stable/installing/#installing-with-get-pip-py)
+
+Test assertions
```
kevinAlbs, line 108: Debug assertions

awitten1:
Ugh, sorry I thought I got them all.  I don't know how I missed those.

src/common/CMakeLists.txt
awitten1: That one's correct I think? config.h.in will set the macro `MONGOC_ENABLE_DEBUG_ASSERTIONS` to whatever the variable cmake variable `MONGOC_ENABLE_DEBUG_ASSERTIONS` equals.

awitten1: Oh sorry nevermind.

awitten1:
I think it should be good now.

PR #650, opened 2020-06-29
CDRIVER-3513 saved errors in out parameter

awitten1:
Saved error in out parameters.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-host-list.c
+++ b/src/libmongoc/src/mongoc/mongoc-host-list.c
@@ -236,15 +236,31 @@ _mongoc_host_list_from_string_with_err (mongoc_host_list_t *link_,
       /* if present, the port should immediately follow after ] */
       sport = strchr (close_bracket, ':');
       if (sport > close_bracket + 1) {
+         bson_set_error (error,
+                         MONGOC_ERROR_COMMAND,
+                         MONGOC_ERROR_COMMAND_INVALID_ARG,
+                         "If present, port should immediately follow the \"]\""
+                         "in an IPv6 address"
+                         );
          return false;
       }
 
       /* otherwise ] should be the last char. */
       if (!sport && *(close_bracket + 1) != '\0') {
+         bson_set_error (error,
+                         MONGOC_ERROR_COMMAND,
+                         MONGOC_ERROR_COMMAND_INVALID_ARG,
+                         "If port is not supplied, \"[\" should be the last"
+                         "character");
          return false;
       }
 
       if (*address != '[') {
+         bson_set_error (error,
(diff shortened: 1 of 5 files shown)


src/libmongoc/src/mongoc/mongoc-host-list.c
```diff
@@ -236,15 +236,31 @@ _mongoc_host_list_from_string_with_err (mongoc_host_list_t *link_,
       /* if present, the port should immediately follow after ] */
       sport = strchr (close_bracket, ':');
       if (sport > close_bracket + 1) {
+         bson_set_error (error,
+                         MONGOC_ERROR_COMMAND,
+                         MONGOC_ERROR_COMMAND_INVALID_ARG,
+                         "If present, port should immediately follow the \"]\""
+                         "in an IPv6 address"
+                         );
```
kevinAlbs, line 244: Nice, I agree with using `MONGOC_ERROR_COMMAND` and `MONGOC_ERROR_COMMAND_INVALID_ARG` as the domain and code. This would error if the user supplied a wrong IPv6 address in the URI, or if a server were to report an invalid IPv6 address.

I looked at codecov.io and it looks like this case isn't encountered in tests.
https://codecov.io/gh/mongodb/mongo-c-driver/src/master/src/libmongoc/src/mongoc/mongoc-host-list.c
`test-mongoc-uri.c` has `test_mongoc_host_list_from_string` which may be an easy way to add a test for this case. Adding the test is out of the scope of this ticket, but if it's a quick win I suggest that we add it.

As a sidenote, codecov.io is not very accurate. It does not capture tests that run on platforms other than Linux, and it fails to upload very often. But it's still a useful measure sometimes.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -1127,6 +1127,11 @@ mongoc_uri_apply_options (mongoc_uri_t *uri,
             }
 
             if (!mongoc_uri_set_option_as_bool (uri, canon, bval)) {
+               bson_set_error (error,
+                               MONGOC_ERROR_COMMAND,
+                               MONGOC_ERROR_COMMAND_INVALID_ARG,
+                               "Failed to set %s to %d",
+                               canon, bval);
```
kevinAlbs, line 1134: As it stands, I don't think we can test this. `mongoc_uri_set_option_as_bool` only fails if attempting to set an option that is not a boolean option or if it's overwriting an option that is stored as something other than a boolean. This is already inside an if-block checking that `mongoc_uri_option_is_bool`. I don't think either of those scenarios are possible, but still worthwhile to set an error if that changes in the future.

PR #651, opened 2020-06-30
CDRIVER-3580 / CDRIVER-3719 soft-fail with Secure Channel + regenerate OCSP tests

kevinAlbs:
Change the default certificate validation with Secure Channel to ignore errors due to an OCSP responder / CRL distribution point being offline and due to a peer certificate not listing any revocation information. [CDRIVER-3580](https://jira.mongodb.org/browse/CDRIVER-3580) has more background.

Patch build with just the OCSP tests:
https://evergreen.mongodb.com/version/5efaada3d1fe072a883a5c2f

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-stream-tls-secure-channel.c
+++ b/src/libmongoc/src/mongoc/mongoc-stream-tls-secure-channel.c
@@ -974,15 +974,15 @@ mongoc_stream_tls_secure_channel_new (mongoc_stream_t *base_stream,
    schannel_cred.dwFlags = SCH_USE_STRONG_CRYPTO;
 #endif
 
+   /* By default, enable soft failing.
+    * A certificate with no revocation check is a soft failure. */
+   schannel_cred.dwFlags |= SCH_CRED_IGNORE_NO_REVOCATION_CHECK;
+   /* An offline OCSP responder / CRL distribution list is a soft failure. */
+   schannel_cred.dwFlags |= SCH_CRED_IGNORE_REVOCATION_OFFLINE;
+
    if (opt->weak_cert_validation) {
-      schannel_cred.dwFlags |= SCH_CRED_MANUAL_CRED_VALIDATION |
-                               SCH_CRED_IGNORE_NO_REVOCATION_CHECK |
-                               SCH_CRED_IGNORE_REVOCATION_OFFLINE;
+      schannel_cred.dwFlags |= SCH_CRED_MANUAL_CRED_VALIDATION;
       TRACE ("%s", "disabled server certificate checks");
-   } else if (_mongoc_ssl_opts_disable_certificate_revocation_check (opt)) {
-      schannel_cred.dwFlags |= SCH_CRED_IGNORE_NO_REVOCATION_CHECK |
-                               SCH_CRED_IGNORE_REVOCATION_OFFLINE;
-      TRACE ("%s", "disabled server certificate revocation checks");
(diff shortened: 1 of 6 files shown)


src/libmongoc/src/mongoc/mongoc-stream-tls-secure-channel.c
```diff
@@ -974,24 +974,23 @@ mongoc_stream_tls_secure_channel_new (mongoc_stream_t *base_stream,
    schannel_cred.dwFlags = SCH_USE_STRONG_CRYPTO;
 #endif
 
+   /* By default, enable soft failing.
```
samantharitter, line 977: Is there a way to opt into a "hard failure" here?  Would that be useful?

kevinAlbs: We could, but I’d prefer to wait unless a need arises though. The other TLS implementations already soft-fail and do not have a way to opt into hard fail, so this is at least consistent. We’d have to add an option that is specific to secure channel, and specific to the C driver.

kevinAlbs:
+ @shreyaskalyan

src/libmongoc/src/mongoc/mongoc-stream-tls-secure-channel.c
```diff
@@ -974,24 +974,23 @@ mongoc_stream_tls_secure_channel_new (mongoc_stream_t *base_stream,
...
+    * A certificate with no revocation check is a soft failure. */
+   schannel_cred.dwFlags |= SCH_CRED_IGNORE_NO_REVOCATION_CHECK;
+   /* An offline OCSP responder / CRL distribution list is a soft failure. */
+   schannel_cred.dwFlags |= SCH_CRED_IGNORE_REVOCATION_OFFLINE;
```
shreyaskalyan, line 981: This looks fine to me. My main concern is does this work? Did you test against both a down responder and a responder sending "STATUS_UNKNOWN"?

kevinAlbs: I tested against a down responder before, but good thing you reminded me about STATUS_UNKNOWN. The spec tests currently do not test that scenario.

I tested the scenarios on Windows by following these steps:

Starting a mongod on Windows with no stapling using the RSA cert  from drivers-evergreen-tools.

Starting the mock responder in drivers-evergreen-tools:

```
python.exe ..\ocsp_mock.py  --ca_file ca.pem --ocsp_responder_cert ca.crt --ocsp_responder_key ca.key -p 8100 -v --fault unknown
```

Flushing the OCSP response cache with:
```
certutil -urlcache * delete
certutil -flushcache 0 1
```

And verifying connection succeeded. I repeated the same with `--fault unknown` to verify connection failed. And did the same with no responder running to ensure offline is a soft-fail.

shreyaskalyan: Cool looks like soft-fail works! Also in the last paragraph I assume you mean `--fault revoked`?

shreyaskalyan:
lgtm

src/libmongoc/src/mongoc/mongoc-stream-tls-secure-channel.c
kevinAlbs: Ah yes, that was a typo, I meant `--fault revoked`.

PR #655, opened 2020-07-06
CDRIVER-3734 add Host header to OCSP requests

kevinAlbs:
Log messages in Evergreen were indicating soft-failures against Amazon KMS servers. In addition to checking responses locally, I inspected the logs of the `-cse` tasks to validate this is no longer failing.

The HTTP header Host is required by KMS servers (and is required by HTTP in general). Related issue in OpenSSL: https://github.com/openssl/openssl/issues/1986

Fixes a leak of `next_time` / `this_time` fields.

Permits verified chain be used to validate OCSP response signature certificate.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-ocsp-cache.c
+++ b/src/libmongoc/src/mongoc/mongoc-ocsp-cache.c
@@ -174,10 +174,8 @@ _mongoc_ocsp_cache_get_status (OCSP_CERTID *id,
 
    *cert_status = entry->cert_status;
    *reason = entry->reason;
-   *this_update =
-      ASN1_item_dup (ASN1_ITEM_rptr (ASN1_TIME), entry->this_update);
-   *next_update =
-      ASN1_item_dup (ASN1_ITEM_rptr (ASN1_TIME), entry->next_update);
+   *this_update = entry->this_update;
+   *next_update = entry->next_update;
 
    ret = true;
 done:
--- a/src/libmongoc/src/mongoc/mongoc-openssl.c
+++ b/src/libmongoc/src/mongoc/mongoc-openssl.c
@@ -658,7 +658,8 @@ _contact_ocsp_responder (OCSP_CERTID *id, X509 *peer)
 {
    STACK_OF (OPENSSL_STRING) *url_stack = NULL;
    OPENSSL_STRING url = NULL, host = NULL, path = NULL, port = NULL;
-   OCSP_REQUEST *req = NULL;
+   OCSP_REQUEST *req;
+   OCSP_REQ_CTX *sendreq_ctx = NULL;
    OCSP_RESPONSE *resp = NULL;
    BIO *bio = NULL;
    int i, ssl;
@@ -702,12 +703,39 @@ _contact_ocsp_responder (OCSP_CERTID *id, X509 *peer)
          GOTO (retry);
       }
 
-      if (!(resp = OCSP_sendreq_bio (bio, path, req))) {
-         MONGOC_DEBUG (
-            "Could not perform an OCSP request for url '%s'. Error: %s",
-            url,
-            ERR_STR);
+      /* Leave OCSP request NULL, set it onto the request context after setting
+       * the host header. */
+      sendreq_ctx =
+         OCSP_sendreq_new (bio, path, NULL /* OCSP request */, 0 /* maxline */);
+      if (host) {
+         if (0 == OCSP_REQ_CTX_add1_header (sendreq_ctx, "Host", host)) {
+            MONGOC_DEBUG ("Could not set OCSP request header for host: %s",
+                          host);
+            GOTO (retry);
+         }
       }
+
+      if (0 == OCSP_REQ_CTX_set1_req (sendreq_ctx, req)) {
+         MONGOC_DEBUG ("Could not set OCSP request");
+         GOTO (retry);
+      }
+
+      do {
+         int ret = OCSP_sendreq_nbio (&resp, sendreq_ctx);
+         if (ret == 1) {
+            /* Success. */
+            break;
+         } else if (ret == -1 && BIO_should_retry (bio)) {
+            /* Non-blocking write not finished, repeat. */
+            continue;
+         } else {
+            MONGOC_DEBUG ("Could not send OCSP request for url '%s'. Error: %s",
+                          url,
+                          ERR_STR);
+            GOTO (retry);
+         }
+      } while (true);
+
    retry:
       if (bio)
          BIO_free_all (bio);
@@ -719,6 +747,8 @@ _contact_ocsp_responder (OCSP_CERTID *id, X509 *peer)
          OPENSSL_free (path);
       if (req)
          OCSP_REQUEST_free (req);
+      if (sendreq_ctx)
+         OCSP_REQ_CTX_free (sendreq_ctx);
    }
 
    if (url_stack)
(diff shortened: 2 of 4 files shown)


src/libmongoc/src/mongoc/mongoc-ocsp-cache.c
```diff
@@ -174,10 +174,8 @@ _mongoc_ocsp_cache_get_status (OCSP_CERTID *id,
 
    *cert_status = entry->cert_status;
    *reason = entry->reason;
-   *this_update =
-      ASN1_item_dup (ASN1_ITEM_rptr (ASN1_TIME), entry->this_update);
-   *next_update =
-      ASN1_item_dup (ASN1_ITEM_rptr (ASN1_TIME), entry->next_update);
```
kevinAlbs, line 180: This is to fix leaks on ASAN that popped up after client side encryption tests started caching responses.

`_mongoc_ocsp_tlsext_status` expects to retrieve the `this_update` / `next_update` pair from the cache or from `OCSP_resp_find_status` and treats both the same. Since the `this_update` / `next_update` from `OCSP_resp_find_status` must not be freed, I changed this to behave the same way.

src/libmongoc/src/mongoc/mongoc-openssl.c
```diff
@@ -702,12 +703,39 @@ _contact_ocsp_responder (OCSP_CERTID *id, X509 *peer)
...
 
-      if (!(resp = OCSP_sendreq_bio (bio, path, req))) {
-         MONGOC_DEBUG (
-            "Could not perform an OCSP request for url '%s'. Error: %s",
-            url,
-            ERR_STR);
+      /* Leave OCSP request NULL, set it onto the request context after setting
+       * the host header. */
+      sendreq_ctx =
+         OCSP_sendreq_new (bio, path, NULL /* OCSP request */, 0 /* maxline */);
```
kevinAlbs, line 709: This now creates an `OCSP_REQ_CTX` to wrap around the `OCSP_REQUEST` and add the `Host` HTTP header.

src/libmongoc/src/mongoc/mongoc-openssl.c
```diff
@@ -702,12 +703,39 @@ _contact_ocsp_responder (OCSP_CERTID *id, X509 *peer)
...
+         }
+      }
+
+      if (0 == OCSP_REQ_CTX_set1_req (sendreq_ctx, req)) {
+         MONGOC_DEBUG ("Could not set OCSP request");
+         GOTO (retry);
       }
+
+      do {
+         int ret = OCSP_sendreq_nbio (&resp, sendreq_ctx);
```
kevinAlbs, line 724: `OCSP_sendreq_bio` works with an `OCSP_REQUEST`, not an `OCSP_REQ_CTX`. The only API I saw for sending `OCSP_REQ_CTX` was the non-blocking send, so this now accounts for retries.

src/libmongoc/src/mongoc/mongoc-openssl.c
```diff
@@ -844,8 +874,12 @@ _mongoc_ocsp_tlsext_status (SSL *ssl, mongoc_openssl_ocsp_opt_t *opts)
     * 2. Next, the function verifies the signature of the basic response.
     * 3. Finally, the function validates the signer cert, constructing the
     * validation path via the untrusted cert chain.
+    *
+    * cert_chain has already been verified. Use OCSP_TRUSTOTHER so the signer
+    * certificate can be considered verified if it is in cert_chain.
     */
-   if (OCSP_basic_verify (basic, cert_chain, store, 0) != OCSP_VERIFY_SUCCESS) {
+   if (OCSP_basic_verify (basic, cert_chain, store, OCSP_TRUSTOTHER) !=
+       OCSP_VERIFY_SUCCESS) {
```
kevinAlbs, line 882: After the `Host` header was included, I had noticed that evergreen hosts were still failing to validate the response. It was due to the difference of installed trusted certificates.

The kms.us-east-1.amazonaws.com returns a chain of four certs: 1 -> 2 -> 3 -> 4.
1 is the peer certificate, and 2 is the issuer.
The OCSP response is signed by 2.
Locally, I have 3 (Amazon Root CA) installed.
On Evergreen, 3 is not installed, but 4 is.

`OCSP_basic_verify` attempts to build the chain from the OCSP response's signing certificate (2) to verify the signature. Since I locally have 3 installed, it can fetch that directly from `store`. But on Evergreen, `store` does not have 3. But 3 is included in `cert_chain`, since that comes from the chain that the server presents.

`OCSP_basic_verify` considers `cert_chain` an untrusted list of certificates by default, and will not use it to verify the OCSP response. However, we've already verified `cert_chain`. We can use OCSP_TRUSTOTHER to indicate `cert_chain` contains verified certificates, and can be used to verify the OCSP signature certificate.

See also: https://www.mail-archive.com/openssl-users@openssl.org/msg22791.html for more documentation of the flags.

bazile-clyde: >  However, we've already verified `cert_chain`.

Where did we verify `cert_chain`?

src/libmongoc/src/mongoc/mongoc-openssl.c
bazile-clyde: Why not just move this line of code under the `host` check (to line 718) and _then_ add the request? That way we could eliminate lines 718-721.

src/libmongoc/tests/test-mongoc-ocsp-cache.c
```diff
@@ -60,129 +60,121 @@ create_cert_id (long serial)
...
+   ASN1_GENERALIZEDTIME_free (this_update_in);
+   ASN1_GENERALIZEDTIME_free (next_update_in);
 }
 
 static void
 test_mongoc_cache_update (void)
 {
    OCSP_CERTID *id;
-   ASN1_GENERALIZEDTIME *next_update, *nextupd, *thisupd;
+   ASN1_GENERALIZEDTIME *next_update_in, *next_update_out, *this_update_out;
```
bazile-clyde, line 110: Ah, this is a lot clearer.

src/libmongoc/src/mongoc/mongoc-openssl.c
kevinAlbs: This chain is obtained from `_get_verified_chain`. Since the OCSP check happens after `BIO_do_handshake`, at that the server certificate chain has already been verified.

src/libmongoc/src/mongoc/mongoc-openssl.c
kevinAlbs: `man OCSP_REQ_CTX_set1_req` says:

```
OCSP_REQ_CTX_set1_req() sets the OCSP request in rctx to req. This function should be called after any calls to OCSP_REQ_CTX_add1_header().
```

I interpreted that to mean headers must be set before setting the request, even if the request is set by `OCSP_sendreq_new`. Though it isn't totally clear from the man pages, this seemed safer.

src/libmongoc/tests/test-mongoc-ocsp-cache.c
kevinAlbs: Thanks!

src/libmongoc/src/mongoc/mongoc-openssl.c
bazile-clyde: Ah, I see.

src/libmongoc/src/mongoc/mongoc-openssl.c
bazile-clyde: Got it.

PR #658, opened 2020-07-10
CDRIVER-3739 Add 5 second timeout to OCSP

kevinAlbs:
- Reuse libmongoc's HTTP function instead of OCSP_sendreq_new
- Support TLS in OCSP endpoint
- Do not log a soft failure if a certificate has no OCSP endpoints

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-cluster-aws.c
+++ b/src/libmongoc/src/mongoc/mongoc-cluster-aws.c
@@ -126,7 +126,6 @@ fail:
 }
 
 
-#define MONGOC_AWS_SOCKET_TIMEOUT 10000
 /*
  * Send an HTTP request and get a response.
  * On success, returns true.
--- a/src/libmongoc/src/mongoc/mongoc-cluster-aws.c
+++ b/src/libmongoc/src/mongoc/mongoc-cluster-aws.c
@@ -25,6 +25,7 @@
 #include "mongoc-trace-private.h"
 #include "mongoc-uri-private.h"
 #include "mongoc-util-private.h"
+#include "mongoc-http-private.h"
 
 #undef MONGOC_LOG_DOMAIN
 #define MONGOC_LOG_DOMAIN "aws_auth"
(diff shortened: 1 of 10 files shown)


src/libmongoc/src/mongoc/mongoc-cluster-aws.c
```diff
@@ -125,6 +126,7 @@ _sasl_reply_parse_payload_as_bson (const bson_t *reply,
 }
 
 
+#define MONGOC_AWS_SOCKET_TIMEOUT 10000
```
bazile-clyde, line 129: Is this used somewhere? Was this meant to be used for `const int socket_timeout_ms = 10000;`?

src/libmongoc/src/mongoc/mongoc-http.c
```diff
@@ -0,0 +1,209 @@
...
+         error,
+         MONGOC_ERROR_STREAM,
+         MONGOC_ERROR_STREAM_SOCKET,
+         "Error occurred reading response: end of headers not found");
+      goto fail;
+   }
+
+   res->headers_len = ptr - http_response_str;
+   res->headers = bson_strndup (http_response_str, res->headers_len);
+   res->body_len = http_response_buf.len - res->headers_len - 4;
```
bazile-clyde, line 195: The `4` here and below on line 198 is for `"\r\n\r\n"` right?

src/libmongoc/tests/test-mongoc-http.c
```diff
@@ -0,0 +1,60 @@
...
+
+   _mongoc_http_request_init (&req);
+   _mongoc_http_response_init (&res);
+
+   /* Basic GET request */
+   req.method = "GET";
+   req.host = "example.com";
+   req.port = 80;
+   r = _mongoc_http_send (&req, 10000, false, NULL, &res, &error);
+   ASSERT_OR_PRINT (r, error);
```
bazile-clyde, line 38: Why do we expect the `GET` and `POST` request to succeed if they're using invalid addresses?

PR #659, opened 2020-07-12
CDRIVER-3745 add flag to disable rev check on Win

kevinAlbs:


src/libmongoc/doc/configuring_tls.rst
```diff
@@ -119,11 +119,13 @@ When compiled against the Windows native libraries, the ``ca_dir`` option of a :
 
 Encrypted PEM files (e.g., setting ``tlsCertificateKeyPassword``) are also not supported, and will result in error when attempting to load them.
 
-When ``tlsCAFile`` is set, the driver will only allow server certificates issued by the authority (or authorities) provided. When no ``tlsCAFile`` is set, the driver will look up the Certificate Authority using the ``System Local Machine Root`` certificate store to confirm the provided certificate or the ``Current user certificate store`` if the ``System Local Machine Root`` certificate store is unavailable.
```
kevinAlbs, line 122: Note, I unnecessarily added support for the CURRENT_USER certificate store in CDRIVER-3408. I realized there would be more to it (we'd need to use it for CRL store as well). I then removed the unnecessary addition of the user store in CDRIVER-3508, but did not update the documentation to reflect that.

PR #660, opened 2020-07-12
CDRIVER-3595 adds tests, removes extraneous check

awitten1:
The bugs mentioned in CDRIVER-3595 don't seem to be observable, as demonstrated in the test cases that this PR adds to test-mongoc-write-concern.c.  The only real change that this PR makes is in mongoc-collection.c.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-collection.c
+++ b/src/libmongoc/src/mongoc/mongoc-collection.c
@@ -3278,8 +3278,9 @@ mongoc_collection_find_and_modify_with_opts (
       }
 
       write_concern = appended_opts.writeConcern;
-   } else if (!write_concern) {
-      if (server_stream->sd->max_wire_version >=
+   }
+   /* inherit write concern from collection if not in transaction */
+   else if (server_stream->sd->max_wire_version >=
              WIRE_VERSION_FAM_WRITE_CONCERN &&
           !_mongoc_client_session_in_txn (parts.assembled.session)) {
          if (!mongoc_write_concern_is_valid (collection->write_concern)) {
@@ -3291,7 +3292,7 @@ mongoc_collection_find_and_modify_with_opts (
          }
 
          write_concern = collection->write_concern;
-      }
+
    }
 
    if (appended_opts.hint.value_type) {
--- a/src/libmongoc/src/mongoc/mongoc-collection.c
+++ b/src/libmongoc/src/mongoc/mongoc-collection.c
@@ -3278,13 +3278,10 @@ mongoc_collection_find_and_modify_with_opts (
       }
 
       write_concern = appended_opts.writeConcern;
-   }
-
-   if (!write_concern) {
+   } else if (!write_concern) {
       if (server_stream->sd->max_wire_version >=
              WIRE_VERSION_FAM_WRITE_CONCERN &&
-          (mongoc_write_concern_is_acknowledged (collection->write_concern) ||
-           !_mongoc_client_session_in_txn (parts.assembled.session))) {
+          !_mongoc_client_session_in_txn (parts.assembled.session)) {
          if (!mongoc_write_concern_is_valid (collection->write_concern)) {
             bson_set_error (error,
                             MONGOC_ERROR_COMMAND,
--- a/src/libmongoc/tests/test-mongoc-write-concern.c
+++ b/src/libmongoc/tests/test-mongoc-write-concern.c
@@ -621,7 +621,7 @@ write_concern_count (const mongoc_apm_command_started_t *event)
 /* tests that unacknowledged write concerns are inherited from the
  * collection in fam with options.  Addresses concerns from CDRIVER-3595 */
 void
-test_write_concern_inheritance_fam (void)
+test_write_concern_inheritance_fam (void *context)
 {
    mongoc_find_and_modify_opts_t *opts;
    bson_t *update;
@@ -771,13 +771,17 @@ test_write_concern_install (TestSuite *suite)
       suite, "/WriteConcern/prohibited", test_write_concern_prohibited);
    TestSuite_AddLive (
       suite, "/WriteConcern/unacknowledged", test_write_concern_unacknowledged);
-   TestSuite_AddLive (
-      suite, "/WriteConcern/inherited_fam", test_write_concern_inheritance_fam);
+   TestSuite_AddFull (suite,
+                      "/WriteConcern/inherited_fam",
+                      test_write_concern_inheritance_fam,
+                      NULL,
+                      NULL,
+                      test_framework_skip_if_max_wire_version_less_than_4);
    TestSuite_AddFull (suite,
                       "/WriteConcern/inherited_fam_txn",
                       test_write_concern_inheritance_fam_txn,
-                      NULL /* dtor */,
-                      NULL /* ctx */,
+                      NULL,
+                      NULL,
                       test_framework_skip_if_no_sessions,
                       test_framework_skip_if_no_txns);
 }
--- a/src/libmongoc/tests/test-mongoc-write-concern.c
+++ b/src/libmongoc/tests/test-mongoc-write-concern.c
@@ -669,7 +669,7 @@ test_write_concern_inheritance_fam (void)
  * write concerns are not inherited from collections.  Adresses concerns
  * in CDRIVER-3595 */
 void
-test_write_concern_inheritance_fam_txn (void)
+test_write_concern_inheritance_fam_txn (void *context)
 {
    mongoc_find_and_modify_opts_t *opts;
    bson_t *update;
@@ -701,6 +701,7 @@ test_write_concern_inheritance_fam_txn (void)
    BSON_APPEND_UTF8 (&query, "firstname", "Zlatan");
    update = bson_new ();
 
+   ASSERT_OR_PRINT (session, error);
    session_id = bson_new ();
    mongoc_client_session_append (session, session_id, &error);
    mongoc_client_session_start_transaction (session, NULL, &error);
@@ -713,6 +714,7 @@ test_write_concern_inheritance_fam_txn (void)
       collection, &query, opts, &reply, &error);
    /* check that the sent write concern is 1 (not inherited) */
    BSON_ASSERT (sent_w == 1);
+   bson_destroy (&reply);
 
    mongoc_write_concern_set_w (wc, MONGOC_WRITE_CONCERN_W_UNACKNOWLEDGED);
    mongoc_collection_set_write_concern (collection, wc);
@@ -735,6 +737,7 @@ test_write_concern_inheritance_fam_txn (void)
    bson_destroy (update);
    bson_destroy (&query);
    bson_destroy (session_id);
+   bson_destroy (&reply);
    mongoc_find_and_modify_opts_destroy (opts);
 }
 
@@ -770,7 +773,11 @@ test_write_concern_install (TestSuite *suite)
       suite, "/WriteConcern/unacknowledged", test_write_concern_unacknowledged);
    TestSuite_AddLive (
       suite, "/WriteConcern/inherited_fam", test_write_concern_inheritance_fam);
-   TestSuite_AddLive (suite,
+   TestSuite_AddFull (suite,
                       "/WriteConcern/inherited_fam_txn",
-                      test_write_concern_inheritance_fam_txn);
+                      test_write_concern_inheritance_fam_txn,
+                      NULL /* dtor */,
+                      NULL /* ctx */,
+                      test_framework_skip_if_no_sessions,
+                      test_framework_skip_if_no_txns);
 }
--- a/src/libmongoc/tests/test-mongoc-write-concern.c
+++ b/src/libmongoc/tests/test-mongoc-write-concern.c
@@ -613,65 +613,22 @@ write_concern_count (const mongoc_apm_command_started_t *event)
       if (bson_iter_find_descendant (&iter, "writeConcern.w", &iter_rec)) {
          *sent_collection_w = bson_iter_int32 (&iter_rec);
       } else {
-         *sent_collection_w = 1; /* no write concern sent. default used */
+         *sent_collection_w = MONGOC_WRITE_CONCERN_W_DEFAULT; /* no write
+                                                                 concern sent.
+                                                                 default used */
       }
    }
 }
 
-/* tests that unacknowledged write concerns are inherited from the
- * collection in fam with options.  Addresses concerns from CDRIVER-3595 */
-void
-test_write_concern_inheritance_fam (void *context)
-{
-   mongoc_find_and_modify_opts_t *opts;
-   bson_t *update;
-   bson_error_t error;
-   bson_t query = BSON_INITIALIZER;
-   mongoc_write_concern_t *wc;
-   bool success;
-   int sent_w = -1;
-   mongoc_apm_callbacks_t *callbacks;
-   mongoc_client_t *client = test_framework_client_new ();
-   mongoc_collection_t *collection =
-      mongoc_client_get_collection (client, "db", "collection");
-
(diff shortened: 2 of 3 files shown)


src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -3278,13 +3278,10 @@ mongoc_collection_find_and_modify_with_opts (
...
       write_concern = appended_opts.writeConcern;
-   }
-
-   if (!write_concern) {
+   } else if (!write_concern) {
       if (server_stream->sd->max_wire_version >=
              WIRE_VERSION_FAM_WRITE_CONCERN &&
-          (mongoc_write_concern_is_acknowledged (collection->write_concern) ||
-           !_mongoc_client_session_in_txn (parts.assembled.session))) {
+          !_mongoc_client_session_in_txn (parts.assembled.session)) {
```
awitten1, line 3284: It is safe to remove the `mongoc_write_concern_is_acknowledged (collection->write_concern)`.  The only case where 
```
server_stream->sd->max_wire_version >=
             WIRE_VERSION_FAM_WRITE_CONCERN &&
          (mongoc_write_concern_is_acknowledged (collection->write_concern) ||
           !_mongoc_client_session_in_txn (parts.assembled.session))
```
 is true and 
```
server_stream->sd->max_wire_version >=
             WIRE_VERSION_FAM_WRITE_CONCERN &&
           !_mongoc_client_session_in_txn (parts.assembled.session)
```
is false is when 
```
server_stream->sd->max_wire_version >=
             WIRE_VERSION_FAM_WRITE_CONCERN && 
mongoc_write_concern_is_acknowledged (collection->write_concern) && 
_mongoc_client_session_in_txn (parts.assembled.session)
``` 
is true.  In that case, a transaction is in progress and the call to `mongoc_cmd_parts_assemble` will always execute the check 

```
if (parts->assembled.is_txn_finish ||
          !_mongoc_client_session_in_txn (cs)) {
         _mongoc_cmd_parts_add_write_concern (parts);
 }
```
which will not add the write concern (as it should not).

Unlike we previously thought, it is not safe to remove the `!_mongoc_client_session_in_txn (parts.assembled.session)` check anywhere.  That is because the call to `mongoc_cmd_parts_assemble` will interpret that as the user attempting to send a session id with an unacknowledged write (which is forbidden) and will return an error, when actually the user is not doing that (because write concern should not propagate from the collection).

kevinAlbs: Nicely done analysis. I agree that we cannot remove the `!_mongoc_client_session_in_txn` here or in any other places where the collection's write concern is being inherited.

> That is because the call to mongoc_cmd_parts_assemble will interpret that as the user attempting to send a session id with an unacknowledged write (which is forbidden) and will return an error, when actually the user is not doing that (because write concern should not propagate from the collection) and the function should not return an error.

Ah right, that comes from https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#when-using-unacknowledged-writes

However, I do believe you've uncovered a bug. Though the transactions specification is [explicit about not inheriting the write concern from a client/database/collection](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#writeconcern), the session specification does not say to bypass write concern inheritance. If a user supplies a session (that is not in a transaction) and the collection has an unacknowledged write concern, I believe an error is expected.

Further evidence: an insert_one with a session and an unacknowledged write concern does produce an error:

```
   doc = bson_new ();
   wc = mongoc_write_concern_new ();
   mongoc_write_concern_set_w (wc, MONGOC_WRITE_CONCERN_W_UNACKNOWLEDGED);
   mongoc_collection_set_write_concern (collection, wc);
   mongoc_client_session_append (client_session, opts, &error);
   r = mongoc_collection_insert_one (collection, doc, opts, NULL /* reply */, &error);
   if (!r) {
      fprintf (stderr, "Insert Error=%s\n", error.message);
   }
```

This prints `Insert Error=Cannot use client session with unacknowledged writes`. And grepping mongoc-collection.c for `collection->write_concern` indicates that would be the same behavior in all other write operations except for `mongoc_find_and_modify_with_opts()`.

Can you add a test for this? Call `mongoc_find_and_modify_with_opts()` with a session on a collection with an unacknowledged write concern, and expect an error. I believe that would fail before, but succeed with your change.

Aside: I think a lot of the hairiness of this logic points to a deeper issue within libmongoc's inheritance and validation. I think it would be simpler to have validation of options based on session and transaction state centralized in `mongoc_cmd_parts_assemble`. But, inheriting the write concern/read concern/read preference complicates that. By the time `mongoc_cmd_parts_assemble` sees those options, it does not know whether they were inherited or explicitly passed for that operation. So in cases where inheritance should not occur (e.g. a write concern should not be inherited in a transaction.

src/libmongoc/tests/test-mongoc-write-concern.c
```diff
@@ -595,10 +596,150 @@ test_write_concern_wtimeout_preserved (void)
...
+   const bson_t *command = mongoc_apm_command_started_get_command (event);
+   bson_iter_t iter, iter_rec;
+
+   if (bson_iter_init (&iter, command)) {
+      int *sent_collection_w =
+         (int *) mongoc_apm_command_started_get_context (event);
+      if (bson_iter_find_descendant (&iter, "writeConcern.w", &iter_rec)) {
+         *sent_collection_w = bson_iter_int32 (&iter_rec);
+      } else {
+         *sent_collection_w = 1; /* no write concern sent. default used */
```
kevinAlbs, line 616: I suggest using `MONGOC_WRITE_CONCERN_W_DEFAULT`. That constant is used in write concern objects to indicate not to append anything to the command (so it shouldn't appear in the command). This way, this won't accidentally match a sent acknowledged write concern.

src/libmongoc/tests/test-mongoc-write-concern.c
```diff
@@ -595,10 +596,150 @@ test_write_concern_wtimeout_preserved (void)
...
+void
+test_write_concern_inheritance_fam (void *context)
+{
+   mongoc_find_and_modify_opts_t *opts;
+   bson_t *update;
+   bson_error_t error;
+   bson_t query = BSON_INITIALIZER;
+   mongoc_write_concern_t *wc;
+   bool success;
+   int sent_w = -1;
```
kevinAlbs, line 632: Use `MONGOC_WRITE_CONCERN_W_DEFAULT`.

src/libmongoc/tests/test-mongoc-write-concern.c
```diff
@@ -595,10 +596,150 @@ test_write_concern_wtimeout_preserved (void)
...
+
+   wc = mongoc_write_concern_new ();
+
+   mongoc_write_concern_set_w (wc, MONGOC_WRITE_CONCERN_W_UNACKNOWLEDGED);
+   mongoc_collection_set_write_concern (collection, wc);
+
+   success = mongoc_collection_find_and_modify_with_opts (
+      collection, &query, opts, NULL, &error);
+
+   BSON_ASSERT (success);
```
kevinAlbs, line 656: Oh, actually, you do not need a new test like I suggested in my other comment. This test can just expect an error.

src/libmongoc/tests/test-mongoc-write-concern.c
```diff
@@ -595,10 +596,150 @@ test_write_concern_wtimeout_preserved (void)
...
+   mongoc_apm_callbacks_destroy (callbacks);
+   mongoc_collection_destroy (collection);
+   mongoc_client_destroy (client);
+   bson_destroy (update);
+   bson_destroy (&query);
+   mongoc_find_and_modify_opts_destroy (opts);
+}
+
+/* tests that in fam operations, with a transaction in progress,
+ * write concerns are not inherited from collections.  Adresses concerns
```
kevinAlbs, line 669: "Addresses"

src/libmongoc/tests/test-mongoc-write-concern.c
```diff
@@ -595,10 +596,150 @@ test_write_concern_wtimeout_preserved (void)
...
+ * in CDRIVER-3595 */
+void
+test_write_concern_inheritance_fam_txn (void *context)
```
kevinAlbs, line 672: If this test is structurally similar to the other, I suggest consolidating them and parameterizing on a `bool in_txn`.

src/libmongoc/tests/test-mongoc-write-concern.c
```diff
@@ -595,10 +596,150 @@ test_write_concern_wtimeout_preserved (void)
...
+   bson_t *update;
+   bson_t *session_id;
+   bson_error_t error;
+   bson_t query = BSON_INITIALIZER;
+   mongoc_write_concern_t *wc;
+   mongoc_client_t *client;
+   mongoc_collection_t *collection;
+   mongoc_client_session_t *session;
+   mongoc_apm_callbacks_t *callbacks;
+   int sent_w = -1;
```
kevinAlbs, line 684: Use `MONGOC_WRITE_CONCERN_W_DEFAULT` (or consolidate the tests).

src/libmongoc/tests/test-mongoc-write-concern.c
```diff
@@ -595,10 +596,137 @@ test_write_concern_wtimeout_preserved (void)
...
+   mongoc_write_concern_set_w (wc, 2);
+   mongoc_collection_set_write_concern (collection, wc);
+
+   BSON_APPEND_UTF8 (&query, "firstname", "Zlatan");
+   update = bson_new ();
+   session_id = bson_new ();
+
+   opts = mongoc_find_and_modify_opts_new ();
+   session = mongoc_client_start_session (client, NULL, &error);
+
```
awitten1, line 662: In the case where `in_txn` is false `session` is not immediately used, but it gets used later.  I did it this way because I wanted to test that unacknowledged and acknowledged write concerns are both inherited with no transaction in progress (and also no lsid in the command). It gets used later because in the `!in_txn` case I also test that sending an lsid with an unacknowledged write concern fails.

kevinAlbs: Ah, in that case, I suggest further simplifying by parameterizing on both `in_txn` and `in_session`.  A struct of two bools could be used as the context.

```
typedef struct {
   bool in_session;
   bool in_txn;
} fam_inheritance_test_t;
```

The only minor annoyance is that the `test_write_concern_install` needs to create separate structs for each test.

```
fam_inheritance_test_t test_no_session = { false, false };
fam_inheritance_test_t test_with_session = { true, false };
fam_inheritance_test_t test_with_txn = { true, true };
/* ... */
```

But I think this further consolidates the test.

src/libmongoc/tests/test-mongoc-write-concern.c
```diff
@@ -595,10 +596,135 @@ test_write_concern_wtimeout_preserved (void)
...
+   client = test_framework_client_new ();
+   collection = mongoc_client_get_collection (client, "db", "collection");
+
+   callbacks = mongoc_apm_callbacks_new ();
+   mongoc_apm_set_command_started_cb (callbacks, write_concern_count);
+   mongoc_client_set_apm_callbacks (client, callbacks, (void *) &sent_w);
+
+   wc = mongoc_write_concern_new ();
+   mongoc_write_concern_set_w (wc, 2);
```
kevinAlbs, line 653: Sorry for not thinking about this before, but using `w` of `2` seems problematic for sharded cluster tests on Evergreen. 

E.g. https://evergreen.mongodb.com/task/mongo_c_driver_asan_ubuntu_test_asan_3.6_sharded_auth_nosasl_openssl_patch_368439f8410c39234f9d82d8466202176bf9d57a_5f0e8a82306615237fbcc50f_20_07_15_04_48_04 logs:

```
[2020/07/15 05:02:52.290] FAIL:/data/mci/fffe1073a527e393f08fee6bf5483918/mongoc/src/libmongoc/tests/test-mongoc-write-concern.c:681  test_write_concern_inheritance_fam_txn()
[2020/07/15 05:02:52.290]   success
[2020/07/15 05:02:52.290]   Write Concern error: Not enough data-bearing nodes at sh01
```

As well as standalones (which support sessions).

Since the default behavior is to not send any write concern, I think this can be set to 1 without adding any ambiguity.

src/libmongoc/src/mongoc/mongoc-collection.c
kevinAlbs: Gah right, I misread the original condition in the if statement. I agree there was no bug originally.

src/libmongoc/tests/test-mongoc-write-concern.c
```diff
@@ -595,10 +596,148 @@ test_write_concern_wtimeout_preserved (void)
...
+   mongoc_client_session_destroy (session);
+   mongoc_apm_callbacks_destroy (callbacks);
+   mongoc_collection_destroy (collection);
+   mongoc_client_destroy (client);
+   bson_destroy (update);
+   bson_destroy (&query);
+   bson_destroy (session_id);
+   mongoc_find_and_modify_opts_destroy (opts);
+}
+
```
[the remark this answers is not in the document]
kevinAlbs: Nice! IMO separate functions is clearer than passing through the context. The test context is often unused.

src/libmongoc/tests/test-mongoc-write-concern.c
```diff
@@ -650,21 +651,24 @@ test_write_concern_inheritance_fam_txn (void *context)
...
+   if (in_session) {
+      session = mongoc_client_start_session (client, NULL, &error);
+      ASSERT_OR_PRINT (session, error);
       mongoc_client_session_append (session, session_id, &error);
-      mongoc_client_session_start_transaction (session, NULL, &error);
       mongoc_find_and_modify_opts_append (opts, session_id);
    }
+   if (in_txn) {
+      BSON_ASSERT (in_session);
+      mongoc_client_session_start_transaction (session, NULL, &error);
```
kevinAlbs, line 670: Wrap this in an `ASSERT_OR_PRINT` (or store the return in a bool) in case starting the transaction fails, so it will be easier to debug if this ever fails.

src/libmongoc/tests/test-mongoc-write-concern.c
```diff
@@ -726,6 +721,24 @@ test_write_concern_inheritance_fam_txn (void *context)
    mongoc_find_and_modify_opts_destroy (opts);
 }
 
+static void
+test_fam_no_session_no_txn ()
```
kevinAlbs, line 725: Use `(void)` to denote no arguments. https://stackoverflow.com/a/51080/774658

awitten1:
Of course! I had no problem bearing with your requests.  The refactoring was definitely worth doing. I would have done it initially but it didn't occur to me.

src/libmongoc/tests/test-mongoc-write-concern.c
awitten1: Ah, that's interesting.  I didn't realize that.

src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -3278,13 +3278,10 @@ mongoc_collection_find_and_modify_with_opts (
       }
 
       write_concern = appended_opts.writeConcern;
-   }
-
-   if (!write_concern) {
+   } else if (!write_concern) {
```
bazile-clyde, line 3281: Won't `write_concern` always be false here? The only line that could set this to true is `3281` but since this is an `else if` either `appended_opts.writeConcern` is `true` and line `3280` is reached or that condition is false in this block will be entered. That makes this essential an `else`.

bazile-clyde:
Just one comment but otherwise everything looks good!

src/libmongoc/src/mongoc/mongoc-collection.c
awitten1: Yes! Good point.

PR #662, opened 2020-07-13
CDRIVER-3748 add warning about commitQuorum

kevinAlbs:


src/libmongoc/src/mongoc/mongoc-cmd.c
```diff
@@ -881,7 +910,8 @@ mongoc_cmd_parts_assemble (mongoc_cmd_parts_t *parts,
       }
 
       if (cs && _mongoc_client_session_in_txn (cs)) {
-         if (!IS_PREF_PRIMARY (cs->txn.opts.read_prefs) && !parts->is_write_command) {
```
kevinAlbs, line 884: clang format

src/libmongoc/src/mongoc/mongoc-cmd.c
```diff
@@ -183,6 +183,21 @@ mongoc_cmd_parts_append_opts (mongoc_cmd_parts_t *parts,
       } else if (BSON_ITER_IS_KEY (iter, "serverId") ||
                  BSON_ITER_IS_KEY (iter, "maxAwaitTimeMS")) {
          continue;
+      } else if (BSON_ITER_IS_KEY (iter, "commitQuorum")) {
```
jmikola, line 186: I noted the justification for this in https://github.com/mongodb/mongo-cxx-driver/pull/691#issuecomment-657796646, but this is certainly not ideal. Do we have a tracking ticket for implementing index management in libmongoc? I suppose even if that is implemented, we can't really remove this in a minor release until 2.0 because CXX 3.6.0 may end up linking to a new version of libmongoc.

At the very least, I'd propose adding a comment here to explain why we're handling a command-specific option here so a future contributor doesn't get the idea that other command-specific options should be lumped into this function.

It also sucks that we can't rely on the server to report an error when this is used (per the [Index Management Q&A](https://github.com/mongodb/specifications/blob/master/source/index-management.rst#q-a)).

kevinAlbs: I added a comment, but I am open to discussing this further. I agree that committing this means we're stuck with this check.

https://jira.mongodb.org/browse/CDRIVER-2096 is the ticket that originally deprecated the index helpers. The rationale there is lack of flexible options for individual index specs. But, I don't think there's strong rationale for not providing an alternative index creation helper. I'm can to create a ticket if that seems like the right way.

Not that we should emulate it, but there is some precedent for conditionally checking option support on the command at this level `mongoc_cmd_parts_set_write_concern` checks for `findAndModify` to determine whether a write concern can be applied. That would occur when running findAndModify through the generic command helpers. That check was added in https://jira.mongodb.org/browse/CDRIVER-2213 despite the fact that the check could have been done in `mongoc_collection_find_and_modify_with_opts`. 

If we do not have this check, and rely on wrapping drivers, we'd still have it documented that users should use the generic helpers to create indexes in http://mongoc.org/libmongoc/current/create-indexes.html

Without adding index helpers, the alternative is to do the suggestion I first proposed in https://github.com/mongodb/mongo-cxx-driver/pull/691#issuecomment-657796646. Do server selection in the C++ driver and check the wire version before applying the commitQuorum option. And in libmongoc, we could make a documentation change in http://mongoc.org/libmongoc/current/create-indexes.html to make it clear that commitQuorum may be quietly ignored on pre-4.4 servers.

jmikola: > make it clear that commitQuorum may be quietly ignored on pre-4.4 servers.

I don't think that's the case. The spec instructs drivers to raise errors because `commitQuorum` may have side effects on pre-4.4 servers (per [Q&A](https://github.com/mongodb/specifications/blob/master/source/index-management.rst#q-a)).

I'm not too worried about enforcing this for libmongoc users relying on the generic command helper, but it'd be reasonable to note the issue on that page. I just wouldn't phrase it as "the option will be ignored on pre-4.4 servers".

If performing eager server selection in C++ is feasible, I think that's the best option; however, I definitely don't want to hold up the 4.4 release for this. But if it's something @bazile-clyde can knock out in an hour or two, I'd say go for it.

As for [CDRIVER-2096](https://jira.mongodb.org/browse/CDRIVER-2096), it's a shame that `create_index_with_opts` is the existing, deprecated function name. The ideal API is to probably allow all index options to be expressed as a flexible `bson_t`, but either way this would require a new API on the collection class. It's probably worth jotting this down in a ticket just so we have a record of the idea, though.

kevinAlbs: Gah right, I noted the potential adverse side effect in pre 4.4 servers in a warning in the index creation documentation.

I created https://jira.mongodb.org/browse/CDRIVER-3755 to track adding an index creation helper.

Aside: my gut says generic command helpers should not be inspecting command / options, but I didn't recall the exact rationale at first. Here's what I've found glancing through specs.

1. [Retryable writes don't inspect commands](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst#behavioral-changes-for-write-commands)
2. [Collation is explicitly not checked in generic commands](https://github.com/mongodb/specifications/blob/master/source/collation/collation.rst/#generic-command-method)
3. [Read/write concern should not be inherited and write concern should not be checked for server support](https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#id6)

However, at a glance libmongoc unfortunately seems to ignore some of that. In mongoc-cmd.c,  `mongoc_cmd_parts_append_opts` which applies to the generic write command helpers, and `mongoc_cmd_parts_append_read_write` which applies options to the non read/write generic command helper, both check the wire version for collation support, read/write concern support. I suppose changing that behavior would be backwards breaking.

The [transactions spec notes](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#calling-committransaction-with-the-generic-runcommand-helper-is-undefined-behavior)

> The purpose of the generic runCommand method is to execute a command directly with minimum additional client-side logic.

But I don't see a clear reason why the client side checks are harmful. I suppose if a driver is being used to connect to something that is not a mongod server, the additional validation may not be applicable. And having the generic helpers not do any validation always provides some way of sending a raw command.

kevinAlbs:
This is a documentation-only change now.

src/libmongoc/src/mongoc/mongoc-cmd.c
jmikola: > Aside: my gut says generic command helpers should not be inspecting command / options, but I didn't recall the exact rationale at first

For context, there is really only one "generic command helper", and that's the `runCommand` equivalent. I've been a stickler in my spec reviews about excusing read- and write-aware helpers (see: [Retryable Reads](https://github.com/mongodb/specifications/blob/master/source/retryable-reads/retryable-reads.rst#unsupported-read-operations)). I _think_ that may apply to most of the citations you have above.

With respect to just Retryable Writes, I don't think we should be adding transaction IDs in `mongoc_write_command_with_opts` (and I don't believe libmongoc does that anyway).

> which applies options to the non read/write generic command helper

This isn't surprising, as I don't recall seeing the command helper type (e.g. `RAW`) checked in those functions, which are further down the stack.

> I suppose if a driver is being used to connect to something that is not a mongod server, the additional validation may not be applicable. And having the generic helpers not do any validation always provides some way of sending a raw command.

I think this is definitely one rationale. It is important that a driver provide some raw mechanism as an escape hatch. Perhaps doubly so given libmongoc's role as a building block for other drivers.

I'd propose transcribing the second half of your last comment into a JIRA ticket for posterity. We can debate the merits of correcting this in a future release (and the significance of a BC break). Worst case, we can mark it for 2.0 along with other wishlist changes.

bazile-clyde:
LGTM

src/libmongoc/src/mongoc/mongoc-cmd.c
kevinAlbs: Created https://jira.mongodb.org/browse/CDRIVER-3758 to track removing unnecessary validation. Perhaps even if it is a backwards break of behavior, it's not a very harmful one.

PR #664, opened 2020-07-16
Minor fixes to issues reported by Coverity

kevinAlbs:
These are not urgent, and will be released in 1.18.0.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-client.c
+++ b/src/libmongoc/src/mongoc/mongoc-client.c
@@ -429,6 +429,7 @@ _mongoc_get_rr_search (const char *service,
    bool dns_success;
    bool callback_success = true;
    int num_matching_records;
+   uint32_t ttl;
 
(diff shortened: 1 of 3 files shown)


src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -517,13 +518,9 @@ _mongoc_get_rr_search (const char *service,
 
       num_matching_records++;
 
-      if (rr_data) {
```
alcaeus, line 520: I assume that this check was removed because `rr_data` is already accessed in L489. However, it is not asserted against, so it could be `NULL` after all. Would it make sense to add an assertion at the top to guard against this?

kevinAlbs: Yeah that seems like a good idea. Adding a `BSON_ASSERT (rr_data)`

PR #665, opened 2020-07-16
CDRIVER-3751 document NULL for set_compressors

kevinAlbs:
There is already a test that asserts this behavior:

https://github.com/mongodb/mongo-c-driver/blob/1.16.2/src/libmongoc/tests/test-mongoc-uri.c/#L1039-L1043


kevinAlbs:
No problem! Closing and merging.

PR #666, opened 2020-07-17
CDRIVER-3757 inits pool cond, adds test

awitten1:
This PR initializes the `mongoc_client_pool_t` condition variable.  It also adds a test, asserting that the function `mongoc_client_pool_pop` returns when the max pool number of clients have already been allocated.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-client-pool.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-pool.c
@@ -117,6 +117,7 @@ mongoc_client_pool_new (const mongoc_uri_t *uri)
 
    pool = (mongoc_client_pool_t *) bson_malloc0 (sizeof *pool);
    bson_mutex_init (&pool->mutex);
+   mongoc_cond_init (&pool->cond);
    _mongoc_queue_init (&pool->queue);
    pool->uri = mongoc_uri_copy (uri);
    pool->min_pool_size = 0;
--- a/src/libmongoc/tests/test-mongoc-client-pool.c
+++ b/src/libmongoc/tests/test-mongoc-client-pool.c
@@ -409,7 +409,6 @@ test_client_pool_create_unused_session (void *context)
 /* Tests case where thread is blocked waiting for a client to be pushed back
  * into the client pool.  Specifically this tests that the program terminates.
  * Addresses CDRIVER-3757 */
-#ifdef BSON_OS_UNIX
 typedef struct pool_timeout {
    mongoc_client_pool_t *pool;
    bson_mutex_t mutex;
@@ -417,13 +416,12 @@ typedef struct pool_timeout {
    int nleft;
 } pool_timeout_args_t;
 
-static void *
-worker (void *arg)
+static BSON_THREAD_FUN (worker, arg)
 {
    pool_timeout_args_t *args = arg;
    mongoc_client_t *client = mongoc_client_pool_pop (args->pool);
    BSON_ASSERT (client);
-   sleep (1);
+   _mongoc_usleep (10);
    mongoc_client_pool_push (args->pool, client);
    bson_mutex_lock (&args->mutex);
    /* notify main thread that current thread has terminated */
@@ -438,9 +436,9 @@ test_client_pool_max_pool_size_exceeded (void)
 {
    mongoc_client_pool_t *pool;
    mongoc_uri_t *uri;
-   pthread_t thread1, thread2;
+   bson_thread_t thread1, thread2;
    pool_timeout_args_t *args = bson_malloc0 (sizeof (pool_timeout_args_t));
-   int wait_time = 4*1000; /* 4000 msec = 4 sec */
+   int wait_time = 2000; /* 2000 msec = 2 sec */
    int ret;
 
    uri = mongoc_uri_new ("mongodb://127.0.0.1/?maxpoolsize=1");
@@ -450,8 +448,8 @@ test_client_pool_max_pool_size_exceeded (void)
    bson_mutex_init (&args->mutex);
    mongoc_cond_init (&args->cond);
 
-   pthread_create (&thread1, NULL, worker, args);
-   pthread_create (&thread2, NULL, worker, args);
+   COMMON_PREFIX (thread_create) (&thread1, worker, args);
+   COMMON_PREFIX (thread_create) (&thread2, worker, args);
 
    bson_mutex_lock (&args->mutex);
    while (args->nleft > 0) {
@@ -461,14 +459,14 @@ test_client_pool_max_pool_size_exceeded (void)
    }
    bson_mutex_unlock (&args->mutex);
 
-   pthread_join (thread1, NULL);
-   pthread_join (thread2, NULL);
+   COMMON_PREFIX (thread_join) (thread1);
+   COMMON_PREFIX (thread_join) (thread2);
 
    mongoc_uri_destroy (uri);
    mongoc_client_pool_destroy (pool);
-   free (args);
+   bson_free (args);
 }
-#endif
+
 void
 test_client_pool_install (TestSuite *suite)
 {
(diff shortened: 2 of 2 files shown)


src/libmongoc/tests/test-mongoc-client-pool.c
```diff
@@ -405,6 +405,70 @@ test_client_pool_create_unused_session (void *context)
...
+   uri = mongoc_uri_new ("mongodb://127.0.0.1/?maxpoolsize=1");
+   pool = mongoc_client_pool_new (uri);
+   args->pool = pool;
+   args->nleft = 2;
+   bson_mutex_init (&args->mutex);
+   mongoc_cond_init (&args->cond);
+
+   pthread_create (&thread1, NULL, worker, args);
+   pthread_create (&thread2, NULL, worker, args);
+
```
awitten1, line 455: I think this is a worthwhile test, but maybe it should do more.  This seemed like a good way to do enforce the timeout, but there definitely might be better ways.  Another way that would work is to set an alarm to go off in 4 seconds (with the `alarm` system call), and if the function still hasn't terminated, then fail the test. It would require setting a signal handler for `SIGALRM`.  That would be a reasonable way to do it also.

src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -117,6 +117,7 @@ mongoc_client_pool_new (const mongoc_uri_t *uri)
 
    pool = (mongoc_client_pool_t *) bson_malloc0 (sizeof *pool);
    bson_mutex_init (&pool->mutex);
+   mongoc_cond_init (&pool->cond);
```
bazile-clyde, line 120: If you removed (or commented out) this line would your test still pass locally? For me it does so I'm wondering if there's something I'm doing wrong on my part. I also wouldn't be opposed to not adding a test since we're just initializing a member variable. @kevinAlbs let me know if you disagree.

awitten1: It fails for me when I comment out that line.  I believe it is undefined behavior and might fail or might not (and maybe always works on some systems).  There is also some non-determinism involved.  I would be ok with not adding the test.  I added it because I don't think we have coverage for this kind of case, where one thread is waiting for another thread to push a client to the pool.

kevinAlbs: It fails on my mac. It does not surprise me that this does not fail on Linux. Popping more clients than available in `maxPoolSize` seems like a common thing to do. If this does not cause any issues on Linux, that could explain why we have not had any reports about this.

I am in favor of the test. Even though the fix is simple, this should help prevent future bugs if we change the client pool behavior.

src/libmongoc/tests/test-mongoc-client-pool.c
```diff
@@ -405,6 +405,70 @@ test_client_pool_create_unused_session (void *context)
...
+   uri = mongoc_uri_new ("mongodb://127.0.0.1/?maxpoolsize=1");
+   pool = mongoc_client_pool_new (uri);
+   args->pool = pool;
+   args->nleft = 2;
+   bson_mutex_init (&args->mutex);
+   mongoc_cond_init (&args->cond);
+
+   pthread_create (&thread1, NULL, worker, args);
+   pthread_create (&thread2, NULL, worker, args);
```
kevinAlbs, line 454: If the only non-Windows specific calls are `pthread_create` and `pthread_join`, I'd argue for enabling this on Windows too. Use `COMMON_PREFIX (thread_create)`, `COMMON_PREFIX (thread_join)`, and define the thread function above as `static BSON_THREAD_FUN (worker, arg)`.

src/libmongoc/tests/test-mongoc-client-pool.c
```diff
@@ -405,6 +405,70 @@ test_client_pool_create_unused_session (void *context)
...
+      BSON_ASSERT (!ret);
+   }
+   bson_mutex_unlock (&args->mutex);
+
+   pthread_join (thread1, NULL);
+   pthread_join (thread2, NULL);
+
+   mongoc_uri_destroy (uri);
+   mongoc_client_pool_destroy (pool);
+   free (args);
```
kevinAlbs, line 469: Use `bson_free`.

src/libmongoc/tests/test-mongoc-client-pool.c
kevinAlbs: Nice! This seems like a good approach to me.

src/libmongoc/tests/test-mongoc-client-pool.c
```diff
@@ -405,6 +405,70 @@ test_client_pool_create_unused_session (void *context)
...
+   int nleft;
+} pool_timeout_args_t;
+
+static void *
+worker (void *arg)
+{
+   pool_timeout_args_t *args = arg;
+   mongoc_client_t *client = mongoc_client_pool_pop (args->pool);
+   BSON_ASSERT (client);
+   sleep (1);
```
kevinAlbs, line 426: IIUC this sleep is here to make the test fail more consistently, to make it likely that both threads attempt to pop a client at the same time.

I'd argue for removing one of the worker threads and removing this sleep to avoid a slow test:
1. main thread creates client pool with maxPoolSize=1
2. main thread pops a client from pool
3. main thread creates worker thread, which attempts to pop a client but must wait
4. main thread pushes client onto pool

awitten1: Yes, that's the reason!  It is to make it likely that thread2 attempts to the pop a client before the thread1 pushes its client back into the pool, causing thread2 to block forever in its call to `mongoc_client_pool_pop`.  I totally agree that this is a slow test.

I attempted to do what you described and encountered two issues:

1. The main thread seems to push its client before the worker thread attempts to pop a client (which is not the case that I would like to test), unless the main thread sleeps before pushing its client.
2. In the event that the worker thread attempts to pop a client before the main thread pushes its client, the main thread will block forever in its call to `mongoc_client_pool_push`, which is another manifestation of the bug.  The issue with that is that it then requires a different timeout mechanism.

I can do those things but I'm not sure they would necessarily fix the issue of the slow test, because it would still require a call to sleep, and it would also complicate the timeout mechanism (only a little bit).

To make the test faster I have replaced `sleep` with `_umongoc_sleep`, which allows a higher level of control on the sleep time.  Now instead of sleeping for ~2 seconds the test will sleep for ~20 micro seconds in the case where the test passes.  

Really the issue here is that I'm not using the clients for anything, which allows the threads to pop and push before the other threads can pop.  In general I'm not a fan of using `sleep` to try and control the order that things happen in, but in this case I don't mind it that much, because I view it as a simulation of doing something with the client.

In any case, using `_umongoc_sleep` definitely speeds it up.  I still think your idea may be better.  What do you think?

kevinAlbs: > 1. The main thread seems to push its client before the worker thread attempts to pop a client (which is not the case that I would like to test), unless the main thread sleeps before pushing its client.

Good point. You could synchronize the worker thread to signal to the main thread with a condition variable after popping a client.

> 2. In the event that the worker thread attempts to pop a client before the main thread pushes its client, the main thread will block forever in its call to mongoc_client_pool_push, which is another manifestation of the bug. The issue with that is that it then requires a different timeout mechanism.

That's true. If the failure case of this test is an infinite block, that seems much worse than a slightly slow / non-deterministic test.

src/libmongoc/tests/test-mongoc-client-pool.c
```diff
@@ -405,6 +405,68 @@ test_client_pool_create_unused_session (void *context)
...
+   mongoc_client_t *client = mongoc_client_pool_pop (args->pool);
+   BSON_ASSERT (client);
+   _mongoc_usleep (10);
+   mongoc_client_pool_push (args->pool, client);
+   bson_mutex_lock (&args->mutex);
+   /* notify main thread that current thread has terminated */
+   args->nleft--;
+   mongoc_cond_signal (&args->cond);
+   bson_mutex_unlock (&args->mutex);
+   return NULL;
```
bazile-clyde, line 431: Should use `BSON_THREAD_RETURN` here for portability.

PR #667, opened 2020-07-21
CDRIVER-3579 CDRIVER-3522 CDRIVER-3762

rcsanchez97:
Full patch build: https://spruce.mongodb.com/version/5f160475d1fe072b63cdf169/tasks?limit=100&page=0


src/libmongoc/src/mongoc/mongoc-stream-tls-openssl.c
```diff
@@ -712,12 +712,13 @@ mongoc_stream_tls_openssl_new (mongoc_stream_t *base_stream,
...
    if (!opt->allow_invalid_hostname) {
       struct in_addr addr;
+      struct in6_addr addr6;
       X509_VERIFY_PARAM *param = X509_VERIFY_PARAM_new ();
 
       X509_VERIFY_PARAM_set_hostflags (param,
                                        X509_CHECK_FLAG_NO_PARTIAL_WILDCARDS);
       if (inet_pton (AF_INET, host, &addr) ||
-          inet_pton (AF_INET6, host, &addr)) {
+          inet_pton (AF_INET6, host, &addr6)) {
```
jmikola, line 721: Happy to see this was such a straightforward fix. That's not to make any comment on how tricky it might have been to come to this solution :smile:

rcsanchez97: The interesting thing is that this particular change was driven by an ASAN failure that started after enabling certificate verification in the unit tests.  Kevin was actually the one who made the connection between the change I made and the specific JIRA ticket that we had not yet managed to figure out.

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -2101,21 +2101,44 @@ _test_mongoc_client_ipv6 (bool pooled)
...
    bson_error_t error;
 
    uri_str = test_framework_add_user_password_from_env ("mongodb://[::1]/");
    uri = mongoc_uri_new (uri_str);
    BSON_ASSERT (uri);
 
+#if (defined(__APPLE__) || defined(_WIN32)) && defined(MONGOC_ENABLE_SSL)
+   /* This is necessary because macOS & Windows seem to treat [::1] as not
+    * matching the hostname "0:0:0:0:0:0:0:1", which results in a certificate
+    * trust failure. */
```
kevinAlbs, line 2116: Can you create tickets to fix the hostname matching on IPv6 literals. Since the fixes for macOS and Windows would be separate, I think separate tickets makes sense. And note the ticket numbers in this comment so we can remove this once those are fixed.

src/libmongoc/tests/test-mongoc-dns.c
```diff
@@ -213,16 +225,28 @@ _test_dns_maybe_pooled (bson_t *test, bool pooled)
       BSON_ASSERT (
          mongoc_uri_get_tls (_mongoc_client_pool_get_topology (pool)->uri) ==
          expect_ssl);
+#ifdef MONGOC_ENABLE_SSL
+      mongoc_client_pool_set_ssl_opts (pool, &ssl_opts);
+#else
       test_framework_set_pool_ssl_opts (pool);
```
kevinAlbs, line 231: The DNS tests require libmongoc to have SSL enabled. `test_framework_set_pool_ssl_opts`  aborts if libmongoc was not built with SSL support. I think you can remove the `#ifdef MONGOC_ENABLE_SSL` in this file.

rcsanchez97: As we discussed in Slack, some of the SSL-related functions are conditionally compiled, while others are always compiled and throw a runtime error if SSL is not enabled.  We will need to retain these conditionals for now.

src/libmongoc/tests/test-mongoc-client.c
rcsanchez97: I wrote CDRIVER-3765 and CDRIVER-3766 and also updated the comment.  I'll push that change along with any others that are needed to finalize this PR.

PR #669, opened 2020-07-23
CDRIVER-3755 defined create index api

awitten1:
https://jira.mongodb.org/browse/CDRIVER-3755

Defines index management API.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-collection-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-collection-private.h
@@ -39,34 +39,6 @@ struct _mongoc_collection_t {
    bson_t *gle;
 };
 
-struct _mongoc_index_options_t {
-   bool background;
-   char *name;
-   bool unique;
-   bson_t *partial_filter_expression;
-   bool sparse;
-   int expire_after_seconds;
-   bool hidden;
-   bson_t *storage_engine;
-   bson_t *weights;
-   char *default_language;
-   char *anguage_override;
-   int text_index_version;
-   int bits;
-   int two_d_sphere_index_version;
-   int index_min;
-   int index_max;
-   int bucket_size;
-   bson_t *collation;
-   bson_t *wildcard_projection;
-};
-
-
-struct _mongoc_index_model_t {
-   bson_t *key;
-   char *name;
-   mongoc_index_options_t *opts;
-};
 
 mongoc_collection_t *
 _mongoc_collection_new (mongoc_client_t *client,
--- a/src/libmongoc/src/mongoc/mongoc-collection-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-collection-private.h
@@ -39,9 +39,33 @@ struct _mongoc_collection_t {
    bson_t *gle;
 };
 
+struct _mongoc_index_options_t {
+   bool background;
+   char *name;
+   bool unique;
+   bson_t *partial_filter_expression;
+   bool sparse;
+   int expire_after_seconds;
+   bool hidden;
+   bson_t *storage_engine;
+   bson_t *weights;
+   char *default_language;
+   char *anguage_override;
+   int text_index_version;
+   int bits;
+   int two_d_sphere_index_version;
+   int index_min;
+   int index_max;
+   int bucket_size;
+   bson_t *collation;
+   bson_t *wildcard_projection;
+};
+
+
 struct _mongoc_index_model_t {
    bson_t *key;
-   mongoc_index_opt_t opts;
+   char *name;
+   mongoc_index_options_t *opts;
 };
 
 mongoc_collection_t *
--- a/src/libmongoc/src/mongoc/mongoc-collection-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-collection-private.h
@@ -39,6 +39,10 @@ struct _mongoc_collection_t {
    bson_t *gle;
 };
 
+struct _mongoc_index_model_t {
+   bson_t *key;
+   mongoc_index_opt_t opts;
+};
 
 mongoc_collection_t *
 _mongoc_collection_new (mongoc_client_t *client,
--- a/src/libmongoc/src/mongoc/mongoc-collection.h
+++ b/src/libmongoc/src/mongoc/mongoc-collection.h
@@ -351,6 +351,57 @@ mongoc_collection_estimated_document_count (
    bson_t *reply,
    bson_error_t *error);
 
+
+typedef enum {
+   MONGOC_INDEX_BACKGROUND,
+   MONGOC_INDEX_NAME,
+   MONGOC_INDEX_UNIQUE,
+   MONGOC_INDEX_PARTIAL_FILTER_EXPRESSION,
+   MONGOC_INDEX_SPARSE,
+   MONGOC_INDEX_EXPIRE_AFTER_SECONDS,
+   MONGOC_INDEX_HIDDEN,
+   MONGOC_INDEX_STORAGE_ENGINE,
+   MONGOC_INDEX_WEIGHTS,
+   MONGOC_INDEX_DEFAULT_LANGUAGE,
+   MONGOC_INDEX_LANGUAGE_OVERRIDE,
+   MONGOC_INDEX_TEXT_INDEX_VERSION,
+   MONGOC_INDEX_BITS,
+   MONGOC_INDEX_TWO_D_SPHERE_INDEX_VERSION,
+   MONGOC_INDEX_MIN,
+   MONGOC_INDEX_MAX,
+   MONGOC_INDEX_BUCKET_SIZE,
+   MONGOC_INDEX_COLLATION,
+   MONGOC_INDEX_WILDCARD_PROJECTION,
+} mongoc_set_index_opt_t;
+
+
+typedef struct _mongoc_index_model_t mongoc_index_model_t;
+
+MONGOC_EXPORT (bool)
+mongoc_collection_create_index_with_commit_quorum (
+   mongoc_collection_t *collection,
+   mongoc_index_model_t **indexes,
+   size_t num_indexes,
+   bson_t *opts,
+   bson_error_t *error);
+
+MONGOC_EXPORT (mongoc_index_model_t *)
+mongoc_index_model_new (void);
+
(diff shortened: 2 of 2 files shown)


src/libmongoc/src/mongoc/mongoc-collection-private.h
```diff
@@ -39,6 +39,10 @@ struct _mongoc_collection_t {
    bson_t *gle;
 };
 
+struct _mongoc_index_model_t {
+   bson_t *key;
+   mongoc_index_opt_t opts;
```
awitten1, line 44: I was thinking it would be a good idea to use this type again.  This will be an opaque type anyway, so it should be safe, and I believe it will allow the flexibility to add new options to that struct without breaking anything.

src/libmongoc/src/mongoc/mongoc-collection.h
```diff
@@ -351,6 +351,57 @@ mongoc_collection_estimated_document_count (
...
+   MONGOC_INDEX_LANGUAGE_OVERRIDE,
+   MONGOC_INDEX_TEXT_INDEX_VERSION,
+   MONGOC_INDEX_BITS,
+   MONGOC_INDEX_TWO_D_SPHERE_INDEX_VERSION,
+   MONGOC_INDEX_MIN,
+   MONGOC_INDEX_MAX,
+   MONGOC_INDEX_BUCKET_SIZE,
+   MONGOC_INDEX_COLLATION,
+   MONGOC_INDEX_WILDCARD_PROJECTION,
+} mongoc_set_index_opt_t;
```
awitten1, line 375: Users can specify the option they are trying to set with this enum.

src/libmongoc/src/mongoc/mongoc-collection.h
```diff
@@ -351,6 +351,57 @@ mongoc_collection_estimated_document_count (
...
+   bson_error_t *error);
+
+MONGOC_EXPORT (mongoc_index_model_t *)
+mongoc_index_model_new (void);
+
+MONGOC_EXPORT (void)
+mongoc_index_model_destroy (mongoc_index_model_t *);
+
+MONGOC_EXPORT (void)
+mongoc_index_model_destroy (mongoc_index_model_t *);
```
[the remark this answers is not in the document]
awitten1: Oops, thanks.

src/libmongoc/src/mongoc/mongoc-collection.h
kevinAlbs: Though this aligns with APIs like cURL's [curl_easy_setopt](https://curl.haxx.se/libcurl/c/curl_easy_setopt.html) and there are trade-offs, I'd vote for continuing libmongoc's established pattern of representing command options with a `bson_t`, and not having this enum.

An advantage of accepting `bson_t` options is forward compatibility. The implementation will inspect the options for necessary client-side checks (i.e. check that `commitQuorum` is not being used on a pre-4.4 server), but otherwise append them to the command. As options are introduced in the future, we do not need to backport changes to support them (users can just append extra options).

src/libmongoc/src/mongoc/mongoc-collection.h
```diff
@@ -351,6 +351,54 @@ mongoc_collection_estimated_document_count (
...
+   MONGOC_INDEX_WILDCARD_PROJECTION,
+} mongoc_set_index_opt_t;
+
+
+typedef struct _mongoc_index_model_t mongoc_index_model_t;
+
+MONGOC_EXPORT (bool)
+mongoc_collection_create_index_with_commit_quorum (
+   mongoc_collection_t *collection,
+   mongoc_index_model_t **indexes,
```
kevinAlbs, line 383: I believe this can be `const mongoc_index_model_t **indexes`. We won't need to modify the `mongoc_index_model_t`

src/libmongoc/src/mongoc/mongoc-collection.h
```diff
@@ -351,6 +351,54 @@ mongoc_collection_estimated_document_count (
...
+   bson_error_t *error);
+
+MONGOC_EXPORT (mongoc_index_model_t *)
+mongoc_index_model_new (void);
+
+MONGOC_EXPORT (void)
+mongoc_index_model_destroy (mongoc_index_model_t *);
+
+MONGOC_EXPORT (void)
+mongoc_index_model_set_opt (mongoc_index_model_t *,
```
kevinAlbs, line 395: If my suggestion for representing the individual index options as `bson_t` sounds good, this could be `mongoc_index_model_set_opts (mongoc_index_model_t *, bson_t *opts)`.

src/libmongoc/src/mongoc/mongoc-collection.h
```diff
@@ -351,6 +351,30 @@ mongoc_collection_estimated_document_count (
...
+typedef struct _mongoc_index_options_t mongoc_index_options_t;
+
+MONGOC_EXPORT (bool)
+mongoc_collection_create_indexes (
+   mongoc_collection_t *collection,
+   const mongoc_index_model_t **indexes,
+   size_t num_indexes,
+   const bson_t *create_indexes_opts,
+   char **index_names, /* out parameter, populated with index names */
+   int *num_indexes_created, /* out parameter contained length of index_names */
```
awitten1, line 367: Added out parameters for index_names created and length of index_names because https://github.com/mongodb/specifications/blob/master/source/index-management.rst indicates that `create_indexes` should return names of indexes created.

src/libmongoc/src/mongoc/mongoc-collection.h
```diff
@@ -351,6 +351,30 @@ mongoc_collection_estimated_document_count (
...
+   bson_t *reply,
+   bson_error_t *error);
+
+MONGOC_EXPORT (mongoc_index_model_t *)
+mongoc_index_model_new (const bson_t *key, const bson_t *index_opts);
```
awitten1, line 372: I considered having `set_key` and `set_index_opts` functions but I like it a little more to take these at index_model creation time but I don't feel that strongly.

src/libmongoc/src/mongoc/mongoc-collection.h
```diff
@@ -351,6 +351,30 @@ mongoc_collection_estimated_document_count (
    bson_t *reply,
    bson_error_t *error);
 
+
+
+
+typedef struct _mongoc_index_model_t mongoc_index_model_t;
+typedef struct _mongoc_index_options_t mongoc_index_options_t;
```
awitten1, line 358: It's a bit annoying that this name is so similar to the deprecated `mongoc_index_opts_t`.  I do like having a type like this though, as long as it is opaque.

src/libmongoc/src/mongoc/mongoc-collection-private.h
```diff
@@ -39,6 +39,34 @@ struct _mongoc_collection_t {
...
+   int bucket_size;
+   bson_t *collation;
+   bson_t *wildcard_projection;
+};
+
+
+struct _mongoc_index_model_t {
+   bson_t *key;
+   char *name;
+   mongoc_index_options_t *opts;
```
awitten1, line 68: Maybe this should be a `bson_t *` instead.

src/libmongoc/src/mongoc/mongoc-collection-private.h
```diff
@@ -39,6 +39,34 @@ struct _mongoc_collection_t {
...
+   char *name;
+   bool unique;
+   bson_t *partial_filter_expression;
+   bool sparse;
+   int expire_after_seconds;
+   bool hidden;
+   bson_t *storage_engine;
+   bson_t *weights;
+   char *default_language;
+   char *anguage_override;
```
jmikola, line 53: This may be a typo.

src/libmongoc/src/mongoc/mongoc-collection-private.h
jmikola: I strongly suggest using a `bson_t*`, as that makes the driver more flexible going forward. Considering we don't even model most command options (e.g. [`find_with_opts`](http://mongoc.org/libmongoc/current/mongoc_collection_find_with_opts.html)) I think `bson_t*` is also more consistent with other libmongoc APIs.

src/libmongoc/src/mongoc/mongoc-collection.h
jmikola: I would rename this to `num_index_names` for consistency.

If you look at the [`createIndexes` output](https://docs.mongodb.com/manual/reference/command/createIndexes/#output), you'll see that it doesn't actually report which indexes were created. It merely says how many existed before and after the command was run. The return value here is derived entirely from the input argument and has no relation to the server response, apart from whether or not the command succeeded or failed.

IIRC, the index management spec doesn't address this detail, although I agree it'd be helpful to clarify that if you want to take a stab at creating a DRIVERS ticket to propose a change (Kevin, Clyde, or Alex can advise on workflow).

src/libmongoc/src/mongoc/mongoc-collection.h
jmikola: To your comment, I definitely don't think we need extra helpers to set options. If anything, I'd argue for an even simpler API.

The index management spec doesn't seem to give a clear definition of IndexModel, so I imagine it's just inferred from the proposed method APIs and [language examples](https://github.com/mongodb/specifications/blob/master/source/index-management.rst#examples).

At the command level, the index model is really just a document where `key` and `name` are always defined (driver populates `name` if not provided by the user) and other options may follow in the same document. Does it make sense for the index model to be anything more than a `bson_t*` internally, even if you intend to abstract it for the user (i.e. define the struct as an opaque type but the `private.h` header defines it as a `bson_t*`)?

In the case where index model is just a `bson_t*` under the hood, you then wouldn't have to do anything special when crafting the `createIndexes` command for the singular or multi-index helper methods. And the pseudo-code for constructing a new index model would roughly be:

 * Allocate a new `bson_t*` for the index model
 * Set the `key` field from the explicit argument here
 * If `name` isn't already in the options, generate it from the key and assign it in the document
 * Copy in any other options (`copy_to_excluding` BSON helpers may work here) into the index model

kevinAlbs:
It will be some time until we can pick this up again. When we do, we'll use this as a starting point.

PR #675, opened 2020-07-31
CRIVER-3728 set prohibit_lsid in cluster_auth_node_cyrus and cluster_…

pmgzl94:
…auth_node_sspi


kevinAlbs:
Thank you for the contribution @pmgzl94!

I validated this manually. Without this PR applied, I added a call to `_mongoc_rpc_printf` in `mongoc_cluster_run_opmsg` authenticating with Kerberos to print out the full auth commands. As expected, the `lsid` was present in `saslStart` and `saslContinue` commands. After applying your change, `lsid` was omitted.

The test failures against the latest server on the attached tests appear due to CDRIVER-3771, which appears due to an unrelated server change.

PR #679, opened 2020-08-26
CDRIVER-3779 resync change stream tests

kevinAlbs:
Note, the `cse` tasks on evergreen are failing due to a compiler warning from libmongocrypt. That is independent of this change.

### DIFF
--- a/src/libmongoc/tests/json-test.c
+++ b/src/libmongoc/tests/json-test.c
@@ -1194,11 +1194,17 @@ execute_test (const json_test_config_t *config,
 
    if (bson_has_field (test, "expectations")) {
       bson_t expectations;
+      bson_iter_t iter;
 
-      bson_lookup_doc (test, "expectations", &expectations);
-      check_json_apm_events (&ctx, &expectations);
-      if (config->events_check_cb) {
-         config->events_check_cb (&ctx.events);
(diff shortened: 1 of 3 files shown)


src/libmongoc/tests/json-test.c
```diff
@@ -1194,11 +1194,17 @@ execute_test (const json_test_config_t *config,
...
+      bson_iter_t iter;
 
-      bson_lookup_doc (test, "expectations", &expectations);
-      check_json_apm_events (&ctx, &expectations);
-      if (config->events_check_cb) {
-         config->events_check_cb (&ctx.events);
+      if (bson_iter_init_find (&iter, test, "expectations")) {
+         /* If 'expectations' is explicitly null, skip the check. */
+         if (!BSON_ITER_HOLDS_NULL (&iter)) {
+            bson_iter_bson (&iter, &expectations);
```
bazile-clyde, line 1202: What's the point of this line?

kevinAlbs: That replaces the previous `bson_lookup_doc`. It takes a `bson_iter_t` instead of the string key to look up the nested "expectations" array. I figured while we have a `bson_iter_t`, I'd might as well use that to look up instead of searching by the string key again.

bazile-clyde: Got it

PR #680, opened 2020-08-31
CDRIVER-3783 apply compile definition and include directory propertie…

rcsanchez97:
…s at the target level

Full patch build: https://spruce.mongodb.com/version/5f4d156f2fbabe76be4394e0/tasks

### DIFF
--- a/src/libbson/CMakeLists.txt
+++ b/src/libbson/CMakeLists.txt
@@ -150,18 +150,6 @@ if (ENABLE_APPLE_FRAMEWORK)
    )
 endif ()
 
-include_directories ("${PROJECT_BINARY_DIR}/src")
-# For the generated files, like bson-config.h, add src/bson to the include path
-# so it can be included relatively when building, like #include "bson-config.h"
-# But when installed, all libbson headers will be in the same directory.
-include_directories ("${PROJECT_BINARY_DIR}/src/bson")
-include_directories ("${PROJECT_SOURCE_DIR}/src")
-add_definitions (-DCOMMON_PREFIX_=_bson_common)
-include_directories ("${PROJECT_SOURCE_DIR}/../../src/common")
-
-include_directories ("${PROJECT_BINARY_DIR}/../../src")
-include_directories ("${PROJECT_BINARY_DIR}/../../src/common")
-
 
 if (APPLE)
    cmake_policy (SET CMP0042 OLD)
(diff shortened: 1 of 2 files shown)


src/libbson/CMakeLists.txt
```diff
@@ -281,13 +281,25 @@ if (MONGOC_ENABLE_STATIC_BUILD)
    add_library (bson_static STATIC ${SOURCES} ${HEADERS} ${HEADERS_FORWARDING})
    target_compile_definitions (bson_static
       PUBLIC BSON_STATIC
-      PRIVATE BSON_COMPILATION JSONSL_PARSE_NAN
+      PRIVATE BSON_COMPILATION JSONSL_PARSE_NAN COMMON_PREFIX_=_bson_common
```
RedBeard0531, line 284: Could you split these to multiple lines

src/libmongoc/CMakeLists.txt
```diff
@@ -799,7 +817,7 @@ function (mongoc_add_test test use_shared)
          target_link_libraries (${test} DbgHelp.dll)
       endif ()
       target_include_directories (${test} PRIVATE ${PROJECT_SOURCE_DIR}/tests)
-      target_compile_definitions (${test} PUBLIC "MONGOC_COMPILATION" PUBLIC "BSON_COMPILATION")
+      target_compile_definitions (${test} PUBLIC "MONGOC_COMPILATION" PUBLIC "BSON_COMPILATION" "COMMON_PREFIX_=_mongoc_common")
```
RedBeard0531, line 820: ditto

RedBeard0531 (approved):
You may want to scrub out all remaining usages of `CMAKE_BINARY_DIR` and `CMAKE_SOURCE_DIR`  to ensure the lib is `add_subdirectory()` clean.

src/libbson/CMakeLists.txt
rcsanchez97: done

src/libmongoc/CMakeLists.txt
rcsanchez97: done

rcsanchez97:
> You may want to scrub out all remaining usages of `CMAKE_BINARY_DIR` and `CMAKE_SOURCE_DIR` to ensure the lib is `add_subdirectory()` clean.

Good point.  It is only used a few other places.  However, I will write a separate ticket for that rather than including that change here.  In particular, the principal use of `CMAKE_BINARY_DIR` in the build is for the generation of the dist manifest and references to it.  Since that impacts the release process, I'd rather make that a separate change and test it accordingly.

src/libbson/CMakeLists.txt
```diff
@@ -236,8 +224,25 @@ set (HEADERS_FORWARDING
...
+   PRIVATE
+      BSON_COMPILATION
+      JSONSL_PARSE_NAN
+      COMMON_PREFIX_=_bson_common
+)
+# Several directories in the source and build trees contain headers we would like
+# include via relative reference, but they all end up in the same install path
+target_include_directories (
+   bson_shared
+   PUBLIC
```
kevinAlbs, line 237: Previously, some directories had `INTERFACE`, and the ones included through `include_directories` are implicitly 
 private. But AFAICT, the `PUBLIC` scope does not have any observable difference. You had mentioned in slack that cmake won't add extraneous directories in the installed package scripts. And I compared the install directories before and after these changes to confirm.

rcsanchez97:
> I will write a separate ticket for that rather than including that change here. In particular, the principal use of `CMAKE_BINARY_DIR` in the build is for the generation of the dist manifest and references to it. Since that impacts the release process, I'd rather make that a separate change and test it accordingly.

The new ticket is CDRIVER-3784.

PR #685, opened 2020-09-22
CDRIVER-3802 remove bits/signum.h

kevinAlbs:


src/libmongoc/tests/test-mongoc-cache.c
```diff
@@ -21,9 +21,11 @@
 #include <mongoc/mongoc.h>
 #include <stdio.h>
 #include <signal.h>
-#include <bits/signum.h>
 #include "TestSuite.h"
 
+#ifndef SIGSTOP
+#define SIGSTOP 19
+
```
bauman, line 28:
```suggestion
#endif
```
that's what I did on alpine 3 12 and it works.

kevinAlbs: Doh, thanks!

kevinAlbs:
Thanks for the report and help @bauman!

PR #686, opened 2020-10-01
CDRIVER-3805 fix build with system zlib in non-standard location

rcsanchez97:
Full Evergreen patch build: https://spruce.mongodb.com/version/5f75fa79e3c3312424d9b554/tasks

### DIFF
--- a/src/libmongoc/CMakeLists.txt
+++ b/src/libmongoc/CMakeLists.txt
@@ -55,6 +55,7 @@ configure_file (
    "${CMAKE_BINARY_DIR}/src/zlib-1.2.11/zconf.h"
    COPYONLY
 )
+set (ZLIB_INCLUDE_DIRS "")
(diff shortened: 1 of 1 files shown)


src/libmongoc/CMakeLists.txt
```diff
@@ -55,6 +55,7 @@ configure_file (
    "${CMAKE_BINARY_DIR}/src/zlib-1.2.11/zconf.h"
    COPYONLY
 )
+set (ZLIB_INCLUDE_DIRS "")
 if (ENABLE_ZLIB MATCHES "SYSTEM|AUTO")
    message (STATUS "Searching for zlib CMake packages")
    include (FindZLIB)
```
kevinAlbs, line 61: Ah I see, we were ignoring the reported value of `ZLIB_INCLUDE_DIRS`.

kevinAlbs (approved):
LGTM! Thanks again for diagnosing this!

PR #688, opened 2020-10-10
docs: fix simple typo, manimpulate -> manipulate

timgates42:
There is a small typo in src/libbson/src/bson/bcon.c

Should read manipulate rather than manimpulate.

(libbson PR https://github.com/mongodb/libbson/pull/223 requested application here)


kevinAlbs:
Thanks for the contribution!

PR #690, opened 2020-10-14
Implement BSON to JSON encoder that limits encoded string length

alcaeus:
This commit was extracted from #684, which is the POC for CDRIVER-3775. Since this logic is not entirely related to logging, I figured it was easier to review this separately.

This introduces a new `bson_as_json_with_opts` method, which is used to pass extra options to the encoding process. For now, the opts struct holds a mode (corresponding to the individual `bson_as_json`, `bson_as_canonical_extended_json`, and `bson_as_relaxed_extended_json` functions) and a maximum length, which is required by the new structured logging component.

The goal behind this function is that not all BSON is read and encoded, but that the process ends when the character limit is reached. For the time being, this works by checking the length limit in a `visit_after` visitor. This means that while we won't continue encoding BSON that is essentially dropped later, each field is always encoded in full and the JSON string is truncated later to not exceed the maximum length. This also stops the encoding process.

Example usage:
```c
// Encode up to 1000 characters in canonical extended JSON
bson_json_opts_t opts = { BSON_JSON_MODE_CANONICAL, 1000 };
bson_t b;

char *str = bson_as_json_with_opts (b, NULL, opts);
```

```c
// Encode the entire BSON document as relaxed extended json
bson_json_opts_t opts = { BSON_JSON_MODE_CANONICAL, BSON_MAX_LEN_UNLIMITED };
bson_t b;

char *str = bson_as_json_with_opts (b, NULL, opts);
```

The example above is equivalent to this:
```c
bson_t b;

char *str = bson_as_relaxed_extended_json (b, NULL);
```

While working on this, I noticed that we don't really cover the current logic. I've added tests that encode into both relaxed and canonical extended formats with and without length limits. I've skipped the legacy encoding mode for the time being, but can add test coverage for that as well if desired. I've also included documentation for the new types (including `bson_json_mode_t` which previously was private). Please let me know if I missed something there.

### DIFF
--- a/src/libbson/doc/bson_as_json_with_opts.rst
+++ b/src/libbson/doc/bson_as_json_with_opts.rst
@@ -1,0 +1,54 @@
+:man_page: bson_as_json_with_opts
+
+bson_as_json_with_opts()
+========================
+
+Synopsis
+--------
+
+.. code-block:: c
+
+  char *
+  bson_as_json_with_opts (const bson_t *bson, size_t *length, const bson_json_opts_t *opts);
+
+Parameters
+----------
+
+* ``bson``: A :symbol:`bson_t`.
+* ``length``: An optional location for the length of the resulting string.
+* ``opts``: A :symbol:`bson_json_opts_t`.
+
+Description
+-----------
+
+The :symbol:`bson_as_json_with_opts()` encodes ``bson`` as a UTF-8 string in the `MongoDB Extended JSON format`_.
+
+The caller is responsible for freeing the resulting UTF-8 encoded string by calling :symbol:`bson_free()` with the result.
+
+If non-NULL, ``length`` will be set to the length of the result in bytes.
+
+The ``opts`` structure is used to pass options for the encoding process. Please refer to the documentation of :symbol:`bson_json_opts_t` for more details.
+
+Returns
+-------
+
+If successful, a newly allocated UTF-8 encoded string and ``length`` is set.
+
+Upon failure, NULL is returned.
+
+Example
+-------
+
+.. code-block:: c
+
+  bson_json_opts_t opts = { BSON_JSON_MODE_CANONICAL, BSON_MAX_LEN_UNLIMITED };
+  char *str = bson_as_json_with_opts (doc, NULL, &opts);
+  printf ("%s\n", str);
+  bson_free (str);
+
+
+.. only:: html
+
+  .. include:: includes/seealso/bson-as-json.txt
+
+.. _MongoDB Extended JSON format: https://github.com/mongodb/specifications/blob/master/source/extended-json.rst
--- a/src/libbson/doc/bson_as_json_with_opts.rst
+++ b/src/libbson/doc/bson_as_json_with_opts.rst
@@ -41,10 +41,11 @@ Example
 
 .. code-block:: c
 
-  bson_json_opts_t opts = { BSON_JSON_MODE_CANONICAL, BSON_MAX_LEN_UNLIMITED };
-  char *str = bson_as_json_with_opts (doc, NULL, &opts);
+  bson_json_opts_t *opts = bson_json_opts_new (BSON_JSON_MODE_CANONICAL, BSON_MAX_LEN_UNLIMITED);
+  char *str = bson_as_json_with_opts (doc, NULL, opts);
   printf ("%s\n", str);
   bson_free (str);
+  bson_json_opts_destroy (opts);
 
 
 .. only:: html
--- a/src/libbson/src/bson/bson-types.h
+++ b/src/libbson/src/bson/bson-types.h
@@ -519,6 +519,40 @@ typedef struct _bson_error_t {
 } bson_error_t BSON_ALIGNED_END (8);
 
 
+/**
+ * BSON_MAX_LEN_UNLIMITED
+ *
+ * Denotes unlimited length limit when converting BSON to JSON.
+ */
+#define BSON_MAX_LEN_UNLIMITED -1
+
+/**
+ * bson_json_mode_t:
+ *
+ * This enumeration contains the different modes to serialize BSON into extended
+ * JSON.
+ */
+typedef enum {
+   BSON_JSON_MODE_LEGACY,
+   BSON_JSON_MODE_CANONICAL,
+   BSON_JSON_MODE_RELAXED,
+} bson_json_mode_t;
+
+/**
+ * bson_json_opts_t:
+ *
+ * This structure is used to pass options for serializing BSON into extended
+ * JSON to the respective serialization methods.
+ *
+ * max_len can be either a non-negative integer, or BSON_MAX_LEN_UNLIMITED to
+ * set no limit for serialization length.
+ */
+typedef struct {
+   bson_json_mode_t mode;
+   int32_t max_len;
+} bson_json_opts_t;
+
+
 BSON_STATIC_ASSERT2 (error_t, sizeof (bson_error_t) == 512);
 
 
--- a/src/libbson/src/bson/bson-types.h
+++ b/src/libbson/src/bson/bson-types.h
@@ -97,6 +97,17 @@ typedef enum {
  */
 typedef struct _bson_context_t bson_context_t;
 
+/**
+ * bson_json_opts_t:
+ *
+ * This structure is used to pass options for serializing BSON into extended
+ * JSON to the respective serialization methods.
+ *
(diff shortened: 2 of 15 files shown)


src/libbson/src/bson/bson.c
```diff
@@ -3174,7 +3238,9 @@ _bson_as_json_visit_all (const bson_t *bson,
       return NULL;
    }
 
-   bson_string_append (state.str, " }");
+   if (!state.max_len_reached) {
```
kevinAlbs, line 3241: Should this check that there is room for the last two characters? E.g. `state.str->len + 2 <= state.max_len`?

src/libbson/src/bson/bson.c
```diff
@@ -3249,7 +3333,9 @@ bson_array_as_json (const bson_t *bson, size_t *length)
       return NULL;
    }
 
-   bson_string_append (state.str, " ]");
+   if (!state.max_len_reached) {
```
kevinAlbs, line 3336: Similarly, should this check that there is room for the last two characters? E.g. `state.str->len + 2 <= state.max_len`?

src/libbson/tests/test-json.c
```diff
@@ -2622,18 +2622,18 @@ test_bson_array_as_json (void)
    char *str;
 
    str = bson_array_as_json (&d, &len);
-   BSON_ASSERT (0 == strcmp (str, "[ ]"));
-   BSON_ASSERT (len == 3);
+   ASSERT_CMPSTR (str, "[ ]");
+   ASSERT_CMPINT (len, ==, 3);
```
kevinAlbs, line 2626: Nice, thanks for updating these assertions with the more helpful `ASSERT_CMP*`

src/libbson/tests/test-json.c
```diff
@@ -2841,6 +2841,701 @@ test_bson_as_json_multi_object (void)
    TEST_JSON_PRODUCES_MULTIPLE ("[],[{'a': 1}]", 1, NULL);
 }
 
+#define TEST_BSON_AS_JSON_WITH_OPTS(_bson, _mode, _max_len, _expected) \
+   {                                                                   \
+      bson_json_opts_t _opts = {_mode, _max_len};                      \
+      size_t _json_len;                                                \
+      char *_str = bson_as_json_with_opts (_bson, &_json_len, &_opts); \
+      ASSERT_CMPSTR (_str, _expected);                                 \
```
kevinAlbs, line 2849: Suggest adding another check to ensure that the produced string does not exceed maximum specified length when truncation occurs.
```
if (_max_len != BSON_MAX_LEN_UNLIMITED) {
    ASSERT_CMPINT(_json_len, <=, _max_len);
}
```

src/libbson/tests/test-json.c
```diff
@@ -2841,6 +2841,701 @@ test_bson_as_json_multi_object (void)
...
+
+   TEST_BSON_AS_JSON_WITH_OPTS (
+      b,
+      BSON_JSON_MODE_CANONICAL,
+      53,
+      "{ \"v\" : { \"$numberDecimal\" : \"51922968585348276285304");
+
+   bson_destroy (b);
+}
+
```
kevinAlbs, line 3538: I suggest one more test. Create a BSON document with every type. Serialize it with no max length limit. Then iterate `i = 0..full string length` and serialize it with a max length of `i`. Assert that the produced string is exactly `i` characters.

Here is an example document:
https://gist.github.com/kevinAlbs/0807bde6f9e5b166abafb502a37f0da5

kevinAlbs (commented):
This looks great, I have a couple small suggestions an an alternate API proposal.

src/libbson/src/bson/bson-types.h
```diff
@@ -519,6 +519,40 @@ typedef struct _bson_error_t {
 } bson_error_t BSON_ALIGNED_END (8);
 
 
+/**
+ * BSON_MAX_LEN_UNLIMITED
+ *
+ * Denotes unlimited length limit when converting BSON to JSON.
+ */
+#define BSON_MAX_LEN_UNLIMITED -1
```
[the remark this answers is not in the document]
samantharitter: I'm inclined to say that `-1` is more in-line with other driver functions that deal with string length, like the `bson_append_*` functions that take `-1` to mean "no specified key length."  Here, it would be kind of parallel, "no specified length limit," or unlimited.  Though I'm not sure why you would use it, I would also expect setting `0` for the max length to mean empty logging strings, and I think it would be a little counterintuitive if `0` meant "unlimited."

src/libbson/src/bson/bson.c
alcaeus: `bson_array_as_json` is existing API, so we can't change the behaviour here without a BC break, hence always using `BSON_MAX_LEN_UNLIMITED`. This method is not used at all by libmongoc itself, so I don't think there's a use-case for adding the limit there. If there is, there should be a `bson_array_as_json_with_opts` that takes a `bson_json_opts_t` to pass the length, as was done with the other `bson_as_*` methods.

samantharitter: @alcaeus that makes a lot of sense!  Why not leave this code the way it was before, then?  It feels a bit unnecessary to me to check if we've reached max length if we don't ever have a length limit.  Do you feel there's a benefit to leaving this logic in?

alcaeus: Sorry for the late reply, I lost track of this comment thread. We can remove the logic from this method since it will never deal with max length and can add this back in if we do decide that we want to support this for `bson_array_as_json` eventually.

src/libbson/doc/bson_as_json_with_opts.rst
```diff
@@ -0,0 +1,54 @@
...
+If successful, a newly allocated UTF-8 encoded string and ``length`` is set.
+
+Upon failure, NULL is returned.
+
+Example
+-------
+
+.. code-block:: c
+
+  bson_json_opts_t opts = { BSON_JSON_MODE_CANONICAL, BSON_MAX_LEN_UNLIMITED };
```
kevinAlbs, line 44: This should use the new `bson_json_opts_new`.

src/libbson/src/bson/bson.c
```diff
@@ -3236,9 +3325,12 @@ bson_array_as_json (const bson_t *bson, size_t *length)
    state.depth = 0;
    state.err_offset = &err_offset;
    state.mode = BSON_JSON_MODE_LEGACY;
+   state.max_len = BSON_MAX_LEN_UNLIMITED; // TODO: how does the limit get here?
```
kevinAlbs, line 3328: I think we would need a `bson_array_as_json_with_opts` for callers to pass a limit, right?

src/libbson/tests/test-json.c
kevinAlbs: Oh that looks awesome! That gives me a lot of confidence for these changes.

src/libbson/tests/test-json.c
```diff
@@ -2841,6 +2841,538 @@ test_bson_as_json_multi_object (void)
...
+			   "\"5192296858534827628530496329220095\" } }");
+
+   bson_destroy (b);
+}
+
+static void
+test_bson_as_json_with_opts_all_types (void)
+{
+   const char *full_canonical;
+   const char *full_relaxed;
```
kevinAlbs, line 3326: `full_canonical` and `full_relaxed` should be non-const `char *`. Currently `debug-compile-with-warnings` task fails due to a warning - L3369 is freeing a `const char*`.

kevinAlbs (commented):
Looking great! Just have a few small comments.

src/libbson/doc/bson_as_json_with_opts.rst
samantharitter: Done.

src/libbson/src/bson/bson.c
samantharitter: Yes, that's correct.  Do we want to add this?  Or leave it as-is for now, with no limits on arrays?

src/libbson/tests/test-json.c
samantharitter: Done.

src/libbson/tests/test-json.c
samantharitter: 👍

src/libbson/src/bson/bson.c
samantharitter: Done.

kevinAlbs (approved):
LGTM!

PR #692, opened 2020-10-23
CDRIVER-3808 fix srv polling thread from spinning

kevinAlbs:
If an SRV URI is used to connect to a deployment other than a sharded cluster
the SRV polling thread spins since it bypasses the poll due to the topology
type being ineligible. This change terminates the thread when the topology
type is discovered to be ineligible.

I reran SRV polling tests described in [the spec](https://github.com/mongodb/specifications/tree/master/source/polling-srv-records-for-mongos-discovery/tests) using the steps outlined in https://github.com/kevinAlbs/SRVPollingTesting.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-topology-background-monitoring.c
+++ b/src/libmongoc/src/mongoc/mongoc-topology-background-monitoring.c
@@ -151,6 +151,7 @@ _mongoc_topology_background_monitoring_start (mongoc_topology_t *topology)
    _mongoc_topology_background_monitoring_reconcile (topology);
    /* Start SRV polling thread. */
    if (mongoc_topology_should_rescan_srv (topology)) {
+      topology->is_srv_polling = true;
       COMMON_PREFIX (thread_create)
       (&topology->srv_polling_thread, srv_polling_run, topology);
    }
@@ -279,7 +280,6 @@ _mongoc_topology_background_monitoring_stop (mongoc_topology_t *topology)
 {
    mongoc_server_monitor_t *server_monitor;
    int i;
-   bool is_srv_polling;
 
    MONGOC_DEBUG_ASSERT (COMMON_PREFIX (mutex_is_locked) (&topology->mutex));
 
@@ -292,9 +292,8 @@ _mongoc_topology_background_monitoring_stop (mongoc_topology_t *topology)
    topology->scanner_state = MONGOC_TOPOLOGY_SCANNER_SHUTTING_DOWN;
    TRACE ("%s", "background monitoring stopping");
 
-   is_srv_polling = NULL != mongoc_uri_get_service (topology->uri);
    /* Signal SRV polling to shut down (if it is started). */
-   if (is_srv_polling) {
+   if (topology->is_srv_polling) {
       mongoc_cond_signal (&topology->srv_polling_cond);
    }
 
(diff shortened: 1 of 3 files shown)


src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -634,19 +674,9 @@ mongoc_topology_rescan_srv (mongoc_topology_t *topology)
    bool ret;
 
    MONGOC_DEBUG_ASSERT (COMMON_PREFIX (mutex_is_locked) (&topology->mutex));
-
-   if ((topology->description.type != MONGOC_TOPOLOGY_SHARDED) &&
-       (topology->description.type != MONGOC_TOPOLOGY_UNKNOWN)) {
-      /* Only perform rescan for sharded topology. */
-      return;
```
kevinAlbs, line 641: This was the cause of the bug, if connecting to a replica set with an SRV string, the polling thread `srv_polling_run` would repeatedly attempt to scan.

src/libmongoc/src/mongoc/mongoc-topology-background-monitoring.c
```diff
@@ -48,6 +48,12 @@ static BSON_THREAD_FUN (srv_polling_run, topology_void)
       }
 
       /* This will check if a scan is due. */
+      if (!mongoc_topology_should_rescan_srv (topology)) {
```
kevinAlbs, line 51: Once we have determined that the topology does not require SRV polling, we can terminate the thread. The topology will remain Unknown until the first SDAM scan, which occurs independently of SRV polling.

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -613,12 +613,52 @@ mongoc_topology_apply_scanned_srv_hosts (mongoc_uri_t *uri,
...
+   if (!service) {
+      /* Only rescan if we have a mongodb+srv:// URI. */
+      return false;
+   }
+
+   if ((topology->description.type != MONGOC_TOPOLOGY_SHARDED) &&
+       (topology->description.type != MONGOC_TOPOLOGY_UNKNOWN)) {
+      /* Only perform rescan for sharded topology. */
+      return false;
+   }
```
bazile-clyde, line 648: Why create a thread at all if the topology description isn't sharded or unknown? It seems like this check could be moved to `_mongoc_topology_background_monitoring_start` or `_start_scanner_if_needed`.

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -613,12 +613,52 @@ mongoc_topology_apply_scanned_srv_hosts (mongoc_uri_t *uri,
...
+mongoc_topology_should_rescan_srv (mongoc_topology_t *topology) {
+   const char *service;
+
+   MONGOC_DEBUG_ASSERT (COMMON_PREFIX (mutex_is_locked) (&topology->mutex));
+
+   service = mongoc_uri_get_service (topology->uri);
+   if (!service) {
+      /* Only rescan if we have a mongodb+srv:// URI. */
+      return false;
+   }
```
bazile-clyde, line 642: It looks like `srv_polling_run` is only called if this condition is true ([here](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/mongoc-topology-background-monitoring.c#L147-L150)). So I don't think this check is needed.

bazile-clyde (approved):
LGTM with a couple of comments/questions.

src/libmongoc/src/mongoc/mongoc-topology.c
bazile-clyde: If we copied the deleted checks (or a call to `mongoc_topology_should_rescan_srv`) to `_mongoc_topology_background_monitoring_start` or `_start_scanner_if_needed` we could leave these checks in here too. That way there's no chance that the user (or one of us) would forget to assert the preconditions are met.

PR #694, opened 2020-11-02
fix /Cursor/common/opts/cmd_deprecated

kevinAlbs:
SERVER-51380 changed ping such that it no longer accepts extra parameters.

### DIFF
--- a/src/libmongoc/tests/test-mongoc-cursor.c
+++ b/src/libmongoc/tests/test-mongoc-cursor.c
@@ -232,7 +232,7 @@ _test_common_server_hint_command_started (
    const char *cmd = mongoc_apm_command_started_get_command_name (event);
    test_common_server_hint_ctx_t *ctx;
    /* only check command associated with cursor priming. */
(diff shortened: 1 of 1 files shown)


src/libmongoc/tests/test-mongoc-cursor.c
```diff
@@ -435,7 +435,7 @@ _make_cmd_deprecated_cursor (mongoc_collection_t *coll)
                                      0,
                                      0,
                                      0,
-                                     tmp_bson ("{'ping': 1}"),
+                                     tmp_bson ("{'isMaster': 1}"),
```
kevinAlbs, line 438: The C driver has a deprecated command cursor, which returns the command reply as the first (and only) document returned in the cursor. 

The cursor tests here (under "/Cursor/common/*") test common behavior across all cursor types. In the process, they send arguments alongside the command like `batchSize`. Since `ping` no longer ignores extra arguments, this changes to use `isMaster`.

bazile-clyde (approved):
LGTM

PR #695, opened 2020-11-09
CDRIVER-3799 support Azure/GCP in CSFLE

kevinAlbs:
Adds integration for libmongocrypt 1.1.0 (and now requires 1.1.0, since it uses new API) to support Azure and GCP KMS providers.


src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -681,71 +682,11 @@ _mongoc_crypt_new (const bson_t *kms_providers,
    mongocrypt_setopt_log_handler (
       crypt->handle, _log_callback, NULL /* context */);
 
-   /* Take options from the kms_providers map. */
```
kevinAlbs, line 684: libmongocrypt's new API for configuring KMS providers: `mongocrypt_setopt_kms_providers` no longer requires the calling driver to parse out options. We can pass the KMS providers BSON document directly.

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -1063,52 +1007,20 @@ _mongoc_crypt_create_datakey (_mongoc_crypt_t *crypt,
       goto fail;
    }
 
-   if (0 == strcmp ("aws", kms_provider) && masterkey) {
```
kevinAlbs, line 1066: Similarly, `mongocrypt_ctx_setopt_key_encryption_key` consolidates the configuration for the master key. So we can pass a BSON document of the options with the "provider" field appended.

src/libmongoc/src/mongoc/mongoc-util.c
```diff
@@ -576,7 +576,7 @@ char *
 _mongoc_getenv (const char *name)
 {
 #ifdef _MSC_VER
-   char buf[1024];
+   char buf[2048];
```
kevinAlbs, line 579: This needed to be bumped because the environment variable for the GCP private key data exceeds 1024 characters (`MONGOC_TEST_GCP_PRIVATEKEY`)

src/libmongoc/tests/json-test.h
```diff
@@ -119,4 +119,7 @@ check_scenario_version (const bson_t *scenario);
 void
 check_outcome_collection (mongoc_collection_t *collection, bson_t *test);
 
+void
+make_kms_provider (bson_t *out);
```
bazile-clyde, line 123: Is this defined somewhere?

kevinAlbs: Good catch, it wasn't. Removed.

bazile-clyde (approved):
LGTM

PR #697, opened 2020-11-14
CDRIVER-3813 correct linkage when using static OpenSSL

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/5fb0002761837d306c6e22b2/tasks?page=0&taskName=openssl-static

Note that the patch build has a pre-set filter to show only the new tasks introduced in this PR.  The new tasks that are introduced by this PR are restricted to Debian 9 (and newer) and Ubuntu 18.04 (and newer).

Older Debian and Ubuntu releases have static OpenSSL libraries that are not built in a way that allows them to be linked into a shared library or dynamic executable (because of relocation).  However, this issue is not present in newer releases, hence the selectivity of where the new tasks run.

Additionally, all Arch, RHEL, and derivatives lack static libraries for OpenSSL, making it impossible to test static linkage on those platforms.

I also tested the minimal example provided by the submitter of the Jira ticket.  The reported problem in the C++ driver build is resolved by the CMake changes to the C driver implemented in this PR.


build/evergreen_config_lib/variants.py
```diff
@@ -245,20 +253,26 @@ def days(n):
...
+             '.debug-compile !.sspi .openssl-static',
              '.debug-compile !.sspi .nossl',
              '.latest .openssl !.nosasl .server',
+             '.latest .openssl-static !.nosasl .server',
              '.latest .nossl',
              '.4.4 .openssl !.nosasl .server',
-             '.4.2 .openssl !.nosasl .server'],
+             '.4.4 .openssl-static !.nosasl .server',
+             '.4.2 .openssl !.nosasl .server',
+             '.4.2 .openssl-static !.nosasl .server'],
```
kevinAlbs, line 264: IMO we do not need to test more than one server version with `openssl-static`. How we link against OpenSSL should have no bearing on the behavior. It seems worthwhile to have one live test as a safety check, so I propose we only test with the latest server version.

kevinAlbs (approved):
I have a suggested reduction in testing matrix, but otherwise LGTM!

rcsanchez97:
Updated Evergreen patch build: https://evergreen.mongodb.com/version/5fb325e81e2d17303e63c8c2?redirect_spruce_users=true

PR #698, opened 2020-11-16
CDRIVER-3422 add structure and error helpers

kevinAlbs:
This is some of the groundwork for the unified test runner and some additional test conveniences.

### DIFF
--- a/src/libmongoc/tests/test-conveniences.c
+++ b/src/libmongoc/tests/test-conveniences.c
@@ -75,7 +75,7 @@ test_conveniences_cleanup ()
          char *str;
 
          str = _mongoc_array_index (&gTmpStringArray, char *, i);
-         bson_free (doc);
+         bson_free (str);
       }
 
       _mongoc_array_destroy (&gTmpBsonArray);
--- a/src/libmongoc/tests/test-conveniences.c
+++ b/src/libmongoc/tests/test-conveniences.c
@@ -38,6 +38,7 @@
 
 static bool gConveniencesInitialized = false;
 static mongoc_array_t gTmpBsonArray;
+static mongoc_array_t gTmpStringArray;
 
 static char *gHugeString;
 static size_t gHugeStringLength;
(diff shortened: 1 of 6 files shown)


src/libmongoc/tests/test-conveniences.c
```diff
@@ -127,7 +137,33 @@ tmp_bson (const char *json, ...)
    return doc;
 }
 
+const char *
+tmp_json (const bson_t *bson)
```
kevinAlbs, line 141: This is for ease of logging BSON as a JSON string. The caller does not need to free, so this enables statements like:

```
MONGOC_DEBUG ("document: %s", tmp_json(&bson));
```

... instead of the current way:
```
char *json = bson_as_json (&bson, NULL);
MONGOC_DEBUG ("document: %s", json);
bson_free (json);
```

src/libmongoc/tests/unified/runner.c
```diff
@@ -0,0 +1,120 @@
...
+static void
+test_ctx_cleanup ()
+{
+   bson_destroy (&test_ctx.command_started_events);
+   bson_destroy (&test_ctx.command_succeeded_events);
+   bson_destroy (&test_ctx.command_failed_events);
+}
+
+static void
+handle_abort (int signo)
```
kevinAlbs, line 53: I want to solve the problem of getting an assertion error in a spec test, and not knowing which operation failed (like an error "assert failed: 3 != 2"). Instead of changing every assertion, this captures the abort signal and prints out state from the global test context.

bazile-clyde: This is definitely a lot more helpful.

bazile-clyde (approved):
LGTM!

src/libmongoc/tests/test-conveniences.c
kevinAlbs: Thanks!

kevinAlbs:
> I'm not sure if the test failures are related.

The test failures on `*-latest` appear due to recent server changes (removal of geoHaystack, CDRIVER-3815) and the issue Roberto pointed out with `/BulkOperation/large_inserts_ordered`. I'll look into that after.

PR #701, opened 2020-11-30
CDRIVER-3794 document default values for connection options

rcsanchez97:


src/libmongoc/doc/mongoc_uri_t.rst
```diff
@@ -87,20 +87,20 @@ IPv4 and IPv6
...
+Constant                                   Key                               Default                           Description
+========================================== ================================= ================================= ============================================================================================================================================================================================================================================
+MONGOC_URI_RETRYREADS                      retryreads                        true                              If "true" and the server is a MongoDB 3.6+ standalone, replica set, or sharded cluster, the driver safely retries a read that failed due to a network error or replica set failover.
+MONGOC_URI_RETRYWRITES                     retrywrites                       true if driver built w/ SSL       If "true" and the server is a MongoDB 3.6+ replica set or sharded cluster, the driver safely retries a write that failed due to a network error or replica set failover. Only inserts, updates of single documents, or deletes of single
```
kevinAlbs, line 94: Use "TLS" instead of "SSL" since the SSL URI options (and server options) are deprecated in favor of the TLS ones.

kevinAlbs (approved):
LGTM

src/libmongoc/doc/mongoc_uri_t.rst
rcsanchez97: Done.

PR #705, opened 2020-12-07
CDRIVER-3819 collation is not an option for collection rename

rcsanchez97:
Note that while this accomplishes the objective of not listing `collation` in an instance where it is not a valid option, it results in duplicating part of the generated `opts` documentation.  If the `opts` change in a way that should be reflected here, then it is not clear how to ensure that change is not overlooked.  That said, this change is probably better than documenting something as valid that is not actually valid.


kevinAlbs (approved):
LGTM! It does not seem worthwhile to do anything particularly fancy to avoid duplicating the other three options.

PR #708, opened 2020-12-13
Logging using native NetBSD thread id call.

devnexen:


kevinAlbs:
Hi @devnexen, thank you for the contribution! This looks like an improvement to me. We do not regularly test against NetBSD, but once we verify this works we will merge this.

kevinAlbs:
Thank you for the contribution @devnexen!

And thank you @rcsanchez97 for testing these changes!

kevinAlbs (approved):

PR #713, opened 2020-12-29
CDRIVER-3825 remove BSON_ASSERT for server_id

samantharitter:
I simply removed this assert.  The logic right below should compare for server_id 0 and fail with a nice error.


kevinAlbs (approved):
LGTM!

PR #717, opened 2021-01-08
Client-side operations timeout struct

bazile-clyde:
This commit adds the timeoutMS struct to be used under the hood for client-side operations. This will not be a public-facing API. If the user needs the value of timeoutMS set on, for example, a client, then they could call a method like,
```
MONGOC_EXPORT (int64_t)
mongoc_client_get_timeout (const mongoc_client_t *client);
```
The only reason this couldn't just be an `int64_t` is that `0` and `unset` mean different things. This was done to prevent backward breaking changes as much as possible.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-timeout.c
+++ b/src/libmongoc/src/mongoc/mongoc-timeout.c
@@ -26,6 +26,7 @@ int64_t
 mongoc_timeout_get_timeout_ms (const mongoc_timeout_t *timeout)
 {
    BSON_ASSERT (timeout);
+   BSON_ASSERT (timeout->is_set);
 
    return timeout->timeout_ms;
 }
--- a/src/libmongoc/src/mongoc/mongoc-timeout.c
+++ b/src/libmongoc/src/mongoc/mongoc-timeout.c
@@ -80,7 +80,9 @@ mongoc_timeout_copy (const mongoc_timeout_t *timeout)
 {
    BSON_ASSERT (timeout);
 
-   mongoc_timeout_t *copy = mongoc_timeout_new ();
+   mongoc_timeout_t *copy = NULL;
+
+   copy = mongoc_timeout_new ();
    copy->timeout_ms = timeout->timeout_ms;
    copy->is_set = timeout->is_set;
 
--- a/src/libmongoc/src/mongoc/mongoc-timeout.c
+++ b/src/libmongoc/src/mongoc/mongoc-timeout.c
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2020 MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "mongoc-timeout-private.h"
+#include "mongoc.h"
+
+typedef struct _mongoc_timeout_t {
+   bool is_set;
+   int64_t timeout_ms;
+} mongoc_timeout_t;
+
+int64_t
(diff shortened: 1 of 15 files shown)


src/libmongoc/src/mongoc/mongoc-timeout.c
```diff
@@ -0,0 +1,102 @@
...
+   return _mongoc_timeout_set_timeout_ms (timeout, timeout_ms);
+}
+
+mongoc_timeout_t *
+mongoc_timeout_new ()
+{
+   mongoc_timeout_t *timeout;
+
+   timeout = (mongoc_timeout_t *) bson_malloc0 (sizeof *timeout);
+   timeout->timeout_ms = 0;
```
kevinAlbs, line 60: I think you can leave out lines 60 and 61 since `bson_malloc0` zeroes out each field.

src/libmongoc/src/mongoc/mongoc-timeout.c
```diff
@@ -0,0 +1,102 @@
...
+   copy->timeout_ms = timeout->timeout_ms;
+   copy->is_set = timeout->is_set;
+
+   return copy;
+}
+
+void
+mongoc_timeout_destroy (mongoc_timeout_t *timeout)
+{
+   bson_free (timeout);
```
kevinAlbs, line 95: I suggest making this a no-op if `timeout` is NULL. That enables the "done" pattern:
```
mongoc_timeout_t *timeout = NULL;

if (some condition) {
   goto done;
}

mongoc_timeout_destroy (timeout); /* works whether NULL or set. */
````

File: src/libmongoc/tests/test-mongoc-timeout.c [hunk 3]
```diff
@@ -0,0 +1,159 @@
...
+   mongoc_timeout_destroy (timeout);
+}
+
+void
+test_timeout_install (TestSuite *suite)
+{
+   TestSuite_Add (suite, "/Timeout/new", test_mongoc_timeout_new);
+   TestSuite_Add (suite, "/Timeout/set", test_mongoc_timeout_set);
+   TestSuite_Add (suite, "/Timeout/get", test_mongoc_timeout_get);
+   TestSuite_Add (suite, "/Timeout/copy", test_mongoc_timeout_copy);
```
kevinAlbs, line 158: Nice, these are some thorough tests!

kevinAlbs (commented):
Looking good!

src/libmongoc/src/mongoc/mongoc-timeout.c
```diff
@@ -0,0 +1,102 @@
...
+
+typedef struct _mongoc_timeout_t {
+   bool is_set;
+   int64_t timeout_ms;
+} mongoc_timeout_t;
+
+int64_t
+mongoc_timeout_get_timeout_ms (const mongoc_timeout_t *timeout)
+{
+   BSON_ASSERT (timeout);
```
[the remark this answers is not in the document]
bazile-clyde: Done. Updated the tests as well. They no longer check the timeout equals 0 after verifying the timeout is unset.

src/libmongoc/src/mongoc/mongoc-timeout.c
bazile-clyde: Done! This is now a one-liner.

src/libmongoc/src/mongoc/mongoc-timeout.c
bazile-clyde: I thought `bson_free` was a no-op if its argument is `NULL`. I added a `NULL` test for `mongoc_timeout_destroy` just in case it fails on some architectures. I also found an example of the "done" pattern with `bson_free` in `client-side-encryption-schema-map:create_schema_file:90`. I'm not 100% sure, though. Waiting for evg tests.

src/libmongoc/tests/test-mongoc-timeout.c
bazile-clyde: Thanks!

src/libmongoc/src/mongoc/mongoc-timeout.c
kevinAlbs: Ah that is correct, `bson_free` follows this pattern. My comment was a knee-jerk reaction since most `destroy` methods have that explicit check. If `mongoc_timeout_t` had other fields to free it would need it. Since it does not, there's nothing to do here.

kevinAlbs (approved):
LGTM!

PR #719, opened 2021-01-13
CDRIVER-3850 allow uri to accept timeout MS

samantharitter:


src/libmongoc/tests/test-mongoc-uri.c
```diff
@@ -255,6 +255,15 @@ test_mongoc_uri_new (void)
    ASSERT (!mongoc_uri_new (
       "mongodb://localhost/?" MONGOC_URI_HEARTBEATFREQUENCYMS "=499"));
 
+   /* timeoutMS */
```
bazile-clyde, line 258: Nit, is this comment needed?

src/libmongoc/tests/test-mongoc-uri.c
```diff
@@ -255,6 +255,15 @@ test_mongoc_uri_new (void)
...
+   uri = mongoc_uri_new ("mongodb://localhost/?" MONGOC_URI_TIMEOUTMS "=3000");
+   ASSERT (uri);
+   ASSERT_CMPINT32 (3000, ==, mongoc_uri_get_option_as_int32 (uri, MONGOC_URI_TIMEOUTMS, 0));
+   mongoc_uri_destroy (uri);
+
+   /* negative timeoutMS */
+   ASSERT (!mongoc_uri_new ("mongodb://localhost/?" MONGOC_URI_TIMEOUTMS "=-1000"));
```
bazile-clyde, line 265: Could we assert the correct warning was thrown here as well?

src/libmongoc/tests/test-mongoc-uri.c
```diff
@@ -605,6 +618,15 @@ test_mongoc_uri_functions (void)
       MONGOC_LOG_LEVEL_WARNING,
       "Invalid \"heartbeatfrequencyms\" of 499: must be at least 500");
 
+   /* timeoutMS must be non-negative */
```
bazile-clyde, line 621: Nit, is this comment needed? The captured log below states the same thing.

src/libmongoc/tests/test-mongoc-uri.c
bazile-clyde, line 258: We should also test non-numeric values throw a warning. An example from `source/uri-options/tests/connection-options.yml` from [this](https://github.com/mongodb/specifications/pull/873/files) PR is
```
        description: "Non-numeric timeoutMS causes a warning"
        uri: "mongodb://example.com/?timeoutMS=invalid"
        valid: true
        warning: true
        hosts: ~
        auth: ~
        options: {}
```

bazile-clyde (changes requested):

src/libmongoc/tests/test-mongoc-uri.c
samantharitter: Removed

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -2782,6 +2784,7 @@ mongoc_uri_set_option_as_int64 (mongoc_uri_t *uri,
 
    r = _mongoc_uri_set_option_as_int64_with_error (uri, option, value, &error);
    if (!r) {
+      printf ("got a warning for int64\n");
```
bazile-clyde, line 2787: Did you mean to leave this print statement here?

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -2817,6 +2820,16 @@ _mongoc_uri_set_option_as_int64_with_error (mongoc_uri_t *uri,
 
    option = mongoc_uri_canonicalize_option (option_orig);
 
+   /* timeoutMS may not be a negative number. */
+   if (!bson_strcasecmp (option, MONGOC_URI_TIMEOUTMS) && value < 0) {
+      printf("checking value for timeoutMS in set_int64\n");
```
bazile-clyde, line 2825: Did you mean to leave this print statement here?

src/libmongoc/tests/test-mongoc-uri.c
bazile-clyde: Why not just create a new function? This one is already really long as is ([code smell: long method](https://refactoring.guru/smells/long-method)). From the link

> As a rule of thumb, if you feel the need to comment on something inside a method, you should take this code and put it in a new method.

I think that's applicable here. In general, this function is pretty hard to read given that you can't see the entire function at once (unless your monitor is ridiculously tall). I wouldn't be opposed to breaking this function up either.

bazile-clyde (changes requested):

src/libmongoc/doc/mongoc_uri_t.rst
```diff
@@ -211,7 +212,7 @@ MONGOC_URI_W                               w                                 Det
...
                                                                              * tags = For replica sets, you can specify a tag set to require that all members of the set that have these tags configured return confirmation of the write operation.
-MONGOC_URI_WTIMEOUTMS                      wtimeoutms                        The time in milliseconds to wait for replication to succeed, as specified in the w option, before timing out. When wtimeoutMS is 0, write operations will never time out.
+MONGOC_URI_WTIMEOUTMS                      wtimeoutms                        Deprecated in favor of MONGOC_URI_TIMEOUTMS. The time in milliseconds to wait for replication to succeed, as specified in the w option, before timing out. When wtimeoutMS is 0, write operations will never time out.
```
kevinAlbs, line 215: I think we need to deprecate `mongoc_write_concern_set_wtimeout` and `mongoc_write_concern_get_wtimeout` as well.

Edit: I think this will be done in CDRIVER-3842. No change needed here!

src/libmongoc/tests/test-mongoc-uri.c
```diff
@@ -255,6 +255,25 @@ test_mongoc_uri_new (void)
    ASSERT (!mongoc_uri_new (
       "mongodb://localhost/?" MONGOC_URI_HEARTBEATFREQUENCYMS "=499"));
 
+   /* timeoutMS */
+   uri = mongoc_uri_new ("mongodb://localhost/?" MONGOC_URI_TIMEOUTMS "=3000");
+   ASSERT (uri);
+   ASSERT_CMPINT64 (3000, ==, mongoc_uri_get_option_as_int64 (uri, MONGOC_URI_TIMEOUTMS, 0));
```
kevinAlbs, line 261: Can you add a check for int32 as well? `mongoc_uri_set_option_as_int32` and `mongoc_uri_get_option_as_int32` should both work. There is some special behavior for `get_option_as_int32` to check for values within int32 range.

src/libmongoc/doc/mongoc_uri_t.rst
samantharitter: Acknowledged.

src/libmongoc/src/mongoc/mongoc-uri.c
samantharitter: Nope, removed

samantharitter (commented):
I've refactored a bunch of the tests in test-mongoc-uri.c.  The way the existing tests were laid out in that file made it difficult to add additional tests for this new option without making more of a mess.

@bazile-clyde did post a link to an article that stated the following:
“As a rule of thumb, if you feel the need to comment on something inside a method, you should take this code and put it in a new method.”

I’ve refactored that code away, so it’s a bit of a moot point.  But I wanted to respond to this comment because it touches on some style things that I think are important and where we might disagree.

In general, I do think it’s good to avoid creating endless methods (like the ones that already existed in this file).  But, another best practice in my book is to match the style in existing code, so as to avoid creating a big mashup of coding styles that makes the code less readable, and, in this case, would make it difficult to see whether there is testing parity between an existing option and something newly added.  

I chose to refactor the tests so that all int options would be tested in the same way by a new test function.  But my second choice would have been to leave my addition as-was, with a comment, to match the way existing and similar options like minHeartbeatFrequencyMS were already being tested in this file.  It would have made that (already long) function a little longer, but it also would have kept the same coding style existent in the file.

bazile-clyde:
> I've refactored a bunch of the tests in test-mongoc-uri.c. The way the existing tests were laid out in that file made it difficult to add additional tests for this new option without making more of a mess.
> 
> @bazile-clyde did post a link to an article that stated the following:
> “As a rule of thumb, if you feel the need to comment on something inside a method, you should take this code and put it in a new method.”
> 
> I’ve refactored that code away, so it’s a bit of a moot point. But I wanted to respond to this comment because it touches on some style things that I think are important and where we might disagree.
> 
> In general, I do think it’s good to avoid creating endless methods (like the ones that already existed in this file). But, another best practice in my book is to match the style in existing code, so as to avoid creating a big mashup of coding styles that makes the code less readable, and, in this case, would make it difficult to see whether there is testing parity between an existing option and something newly added.
> 
> I chose to refactor the tests so that all int options would be tested in the same way by a new test function. But my second choice would have been to leave my addition as-was, with a comment, to match the way existing and similar options like minHeartbeatFrequencyMS were already being tested in this file. It would have made that (already long) function a little longer, but it also would have kept the same coding style existent in the file.

I agree that we should strive to match the style of existing code wherever possible. The only obvious exception should be if the existing code is riddled with code smells, anti-patterns, or other faults. If you agree that this is a code smell then we shouldn't make matters worse by adding to the mess. That'll lead to more code smells. 

If we run into the case where a single file has a ton of functions then we could consider breaking that file into multiple files. If we have too many files we can further refactor them into separate modules/folders, etc. In this case, I was highlighting an anti-pattern, not only a style choice.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -2817,6 +2819,16 @@ _mongoc_uri_set_option_as_int64_with_error (mongoc_uri_t *uri,
 
    option = mongoc_uri_canonicalize_option (option_orig);
 
+   /* timeoutMS may not be a negative number. */
+   if (!bson_strcasecmp (option, MONGOC_URI_TIMEOUTMS) && value < 0 && value != -1) {
```
kevinAlbs, line 2823: Why is `-1` considered valid? If it is a sentinel value (perhaps meaning `unset`) I suggest defining it as a constant.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -727,21 +727,23 @@ mongoc_uri_option_is_int32 (const char *key)
...
           !strcasecmp (key, MONGOC_URI_SOCKETCHECKINTERVALMS) ||
-          !strcasecmp (key, MONGOC_URI_SOCKETTIMEOUTMS) ||
           !strcasecmp (key, MONGOC_URI_LOCALTHRESHOLDMS) ||
           !strcasecmp (key, MONGOC_URI_MAXPOOLSIZE) ||
           !strcasecmp (key, MONGOC_URI_MAXSTALENESSSECONDS) ||
           !strcasecmp (key, MONGOC_URI_MINPOOLSIZE) ||
           !strcasecmp (key, MONGOC_URI_MAXIDLETIMEMS) ||
           !strcasecmp (key, MONGOC_URI_WAITQUEUEMULTIPLE) ||
+          !strcasecmp (key, MONGOC_URI_ZLIBCOMPRESSIONLEVEL) ||
+          /* deprecated options */
```
kevinAlbs, line 737: If `MONGOC_URI_TIMEOUTMS` is mixed with a deprecated timeout option it will override the behavior of the deprecated option.

I suggest warning with `MONGOC_WARNING` if a URI contains `MONGOC_URI_WAITQUEUETIMEOUTMS` or `MONGOC_URI_SOCKETTIMEOUTMS`.

`mongoc_uri_apply_options` may be a good spot to add that check. That is a common to both ways options are set: options specified in the URI string or options retrieved from [TXT records](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst#default-connection-string-options). Though `timeoutMS` would not be applied through TXT records currently.

src/libmongoc/tests/test-mongoc-uri.c
```diff
@@ -705,52 +650,29 @@ test_mongoc_uri_functions (void)
...
-      "WayTooLongAppnameToBeValidSoThisShouldResultInAnErrorWayToLongAppnameToB"
-      "eValidSoThisShouldResultInAnErrorWayToLongAppnameToBeValidSoThisShouldRe"
-      "sultInAnError",
-      &error));
-   ASSERT_ERROR_CONTAINS (error,
-                          MONGOC_ERROR_COMMAND,
-                          MONGOC_ERROR_COMMAND_INVALID_ARG,
-                          "Unsupported value for \"appname\""); /* ... */
-
+   // TODO Sam move this test, it is not testing new_with_error.
```
kevinAlbs, line 675: It looks like this was moved in L691. I think you can remove this and the immediate test below.

src/libmongoc/tests/test-mongoc-uri.c
```diff
@@ -2416,10 +2268,10 @@ test_mongoc_uri_duplicates (void)
 
    RECREATE_URI (MONGOC_URI_READPREFERENCE
                  "=secondary&" MONGOC_URI_MAXSTALENESSSECONDS
-                 "=1&" MONGOC_URI_MAXSTALENESSSECONDS "=2");
+                 "=100&" MONGOC_URI_MAXSTALENESSSECONDS "=200");
```
kevinAlbs, line 2271: Was this intentional?

src/libmongoc/tests/test-mongoc-uri.c
```diff
@@ -2563,128 +2420,251 @@ test_mongoc_uri_duplicates (void)
    mongoc_uri_destroy (uri);
 }
 
+static void
+run_common_int_tests (const char *option,
+		      int32_t val)
```
kevinAlbs, line 2425: run clang-format

src/libmongoc/tests/test-mongoc-uri.c
```diff
@@ -2772,6 +2752,7 @@ test_uri_install (TestSuite *suite)
    TestSuite_Add (suite, "/Uri/utf8", test_mongoc_uri_utf8);
    TestSuite_Add (suite, "/Uri/duplicates", test_mongoc_uri_duplicates);
    TestSuite_Add (suite, "/Uri/int_options", test_mongoc_uri_int_options);
+   TestSuite_Add (suite, "/Uri/bool_options", test_mongoc_uri_bool_options);
```
kevinAlbs, line 2755: After warning, can you add a test to check that a combination of timeoutMS and socketTimeoutMS / waitQueueTimeoutMS produces a warning?

kevinAlbs (commented):
Looking good! I just have a request for warning when timeoutMS is mixed with a deprecated option it overrides.

src/libmongoc/src/mongoc/mongoc-uri.c
samantharitter: Done, it should not accept `-1`.

src/libmongoc/tests/test-mongoc-uri.c
samantharitter: Adding in additional tests revealed that we handle maxStalenessSeconds correctly:

From the MDB docs:
"You must specify a maxStalenessSeconds value of 90 seconds or longer: specifying a smaller maxStalenessSeconds value will raise an error."

From the C driver URI docs:
"The smallest allowed value for maxStalenessSeconds is 90 seconds."

Even though we aren't actually enforcing this lower limit at the time when we set this parameter, I thought it was silly to be using numbers <90 to test our "success" case.  We should probably fix this, but I didn't want to add that fix to this code review.

src/libmongoc/tests/test-mongoc-uri.c
samantharitter: Done.

src/libmongoc/src/mongoc/mongoc-uri.c
samantharitter: Done.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -2503,6 +2506,30 @@ mongoc_uri_get_option_as_int32 (const mongoc_uri_t *uri,
...
+static void
+_mongoc_uri_warn_for_bad_int_option_combos (const mongoc_uri_t *uri,
+					    const char *option)
+{
+   /* Warn for deprecated option combinations */
+   if (!strcasecmp (option, MONGOC_URI_TIMEOUTMS)) {
+      if (mongoc_uri_get_option_as_int64 (uri, MONGOC_URI_WAITQUEUETIMEOUTMS, -1) > 0 ||
+	  mongoc_uri_get_option_as_int64 (uri, MONGOC_URI_SOCKETTIMEOUTMS, -1) > 0 ||
+	  mongoc_uri_get_option_as_int64 (uri, MONGOC_URI_WTIMEOUTMS, -1) > 0) {
+	 MONGOC_WARNING ("Setting a deprecated timeout option in combination with timeoutMS");
```
kevinAlbs, line 2519: To give users more information on which option is deprecated:
```suggestion
	 MONGOC_WARNING ("Setting a deprecated timeout option %s in combination with timeoutMS", option);
```

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -2503,6 +2506,30 @@ mongoc_uri_get_option_as_int32 (const mongoc_uri_t *uri,
...
+      }
+   }
+
+   if (!strcasecmp (option, MONGOC_URI_WAITQUEUETIMEOUTMS) ||
+       !strcasecmp (option, MONGOC_URI_SOCKETTIMEOUTMS) ||
+       !strcasecmp (option, MONGOC_URI_WTIMEOUTMS)) {
+      if (mongoc_uri_get_option_as_int64 (uri, MONGOC_URI_TIMEOUTMS, -1) > 0) {
+	 MONGOC_WARNING ("Setting a deprecated timeout option in combination with timeoutMS");
```
kevinAlbs, line 2527:
```suggestion
	 MONGOC_WARNING ("Setting a deprecated timeout option %s in combination with timeoutMS", option);
```

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -2817,6 +2847,17 @@ _mongoc_uri_set_option_as_int64_with_error (mongoc_uri_t *uri,
 
    option = mongoc_uri_canonicalize_option (option_orig);
 
+   /* timeoutMS may not be a negative number. */
+   if (!bson_strcasecmp (option, MONGOC_URI_TIMEOUTMS) && value < 0) {
+      printf ("checking value for timeoutMS in set_int64\n");
```
kevinAlbs, line 2852: remove

src/libmongoc/tests/test-mongoc-uri.c
kevinAlbs: Thanks!

kevinAlbs (approved):
LGTM! I rebased the `csot` branch so this PR only shows the commits made for the URI changes.

bazile-clyde (approved):

src/libmongoc/src/mongoc/mongoc-uri.c
samantharitter: Done.

samantharitter:
Manually merged.

PR #720, opened 2021-01-16
log report, using native thread id support for Mac

devnexen:


kevinAlbs:
Thank you for the contribution @devnexen!

kevinAlbs (approved):

PR #722, opened 2021-01-21
CDRIVER-3422 remaining support of unified test format

kevinAlbs:
At long last, the last part of the unified test runner.

This adds:
- The GridFS bucket entity
- All remaining operations
- Result/expectation checking

It supports the 1.1 format excluding the `serverApi` option (which will be done as part of version API).

JSON tests are on the first commit (select commit range to filter during review).


src/libmongoc/tests/unified/operation.c
```diff
@@ -0,0 +1,2259 @@
...
+   entity_t *entity = NULL;
+   bson_parser_t *parser = NULL;
+   mongoc_change_stream_t *changestream = NULL;
+   bson_t *pipeline = NULL;
+   const bson_t *op_reply = NULL;
+   bson_error_t op_error = {0};
+
+   parser = bson_parser_new ();
+   /* Capture options as all extra fields, and pass them directly as change
+    * stream options. */
```
kevinAlbs, line 83: The rationale for passing extra options along is to avoid ignoring any future options added to tests.

Operations in libmongoc pass extra options along to commands. That will result in a server error, so it should result in a test failure.

src/libmongoc/tests/unified/entity-map.c
```diff
@@ -718,3 +858,382 @@ entity_map_get (entity_map_t *entity_map, const char *id, bson_error_t *error)
...
+}
+
+
+bool
+entity_map_end_session (entity_map_t *em, char *session_id, bson_error_t *error)
+{
+   bool ret = false;
+   entity_t *entity = NULL;
+
+   entity = entity_map_get (em, session_id, error);
```
bazile-clyde, line 1178: So because we use `entity_map_get` and not `entity_map_get_session` this function is safe to call on sessions that have already ended?

src/libmongoc/tests/unified/test-diagnostics.c
```diff
@@ -0,0 +1,158 @@
...
+   struct _msg_t *next;
+} msg_t;
+
+typedef struct {
+   msg_t *test_info;
+   msg_t *error_info;
+   bson_mutex_t mutex;
+} test_diagnostics_t;
+
+test_diagnostics_t diagnostics;
```
bazile-clyde, line 35: Should this be `static`? It looks like it's only used in this file.

src/libmongoc/tests/unified/test-diagnostics.c
```diff
@@ -0,0 +1,158 @@
...
+
+typedef struct _msg_t {
+   char *string;
+   struct _msg_t *next;
+} msg_t;
+
+typedef struct {
+   msg_t *test_info;
+   msg_t *error_info;
+   bson_mutex_t mutex;
```
bazile-clyde, line 32: It looks like this mutex is only used when adding to this struct (in `_test_diagnostics_add`). Should we also use it before reading from this struct? We do in other places in our codebase. For example:

```
mongoc_topology_description_type_t
_mongoc_topology_get_type (mongoc_topology_t *topology)
{
   mongoc_topology_description_type_t td_type;

   bson_mutex_lock (&topology->mutex);

   td_type = topology->description.type;

   bson_mutex_unlock (&topology->mutex);

   return td_type;
}
```
It seems like we could run into the case where information is changing as we're reading it. Though, I'm not entirely sure.

bazile-clyde (changes requested):

src/libmongoc/tests/unified/entity-map.c
```diff
@@ -265,27 +276,10 @@ entity_client_new (bson_t *bson, bson_error_t *error)
...
-            test_set_error (error,
-                            "useMultiMongoses is false, so expected single "
-                            "host listed, but got: %s",
-                            mongoc_uri_get_string (uri));
-            goto done;
-         }
+   if (use_multiple_mongoses) {
+      if (!test_framework_uri_apply_multi_mongos (
+             uri, *use_multiple_mongoses, error)) {
+         goto done;
```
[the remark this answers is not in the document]
bazile-clyde: Ah, I see.

src/libmongoc/tests/unified/operation.c
```diff
@@ -0,0 +1,2261 @@
...
+
+   if (!op->session) {
+      test_set_error (error, "%s", "session unset");
+      goto done;
+   }
+
+   if (check_dirty) {
+      if (check_dirty != mongoc_client_session_get_dirty (op->session)) {
+         test_set_error (error,
+                         "expected session to%s be dirty but was not",
```
bazile-clyde, line 1492: nit `to%s`. It looks like a space should be added here or before "not" below or the output might be `tonot`.

src/libmongoc/tests/unified/operation.c
```diff
@@ -0,0 +1,2261 @@
...
+                         check_dirty ? "" : "not");
```
bazile-clyde, line 1493: If `check_dirty == false` then this line would not be reached because of the check on line 1489. Maybe delete the check on 1489?

src/libmongoc/tests/unified/operation.c
```diff
@@ -0,0 +1,2261 @@
...
+         tmp_json (b->command));
+      goto done;
+   }
+
+   bson_iter_bson (&iter, &b_lsid);
+
+   if (check_same != bson_equal (&a_lsid, &b_lsid)) {
+      test_set_error (error,
+                      "expected $lsid's to be%s equal, but got: %s and %s",
+                      check_same ? "" : "not",
```
bazile-clyde, line 1608: nit "not" -> " not". Add space.

src/libmongoc/tests/unified/operation.c
```diff
@@ -0,0 +1,2261 @@
...
+   case MONGOC_TRANSACTION_IN_PROGRESS:
+      return "in_progress";
+   case MONGOC_TRANSACTION_COMMITTED:
+      return "committed";
+   case MONGOC_TRANSACTION_ABORTED:
+      return "aborted";
+   default:
+      return "invalid";
+   }
+   return "invalid";
```
bazile-clyde, line 1722: This line will never be reached.

src/libmongoc/tests/unified/result.c
```diff
@@ -0,0 +1,564 @@
...
+         test_set_error (error,
+                         "expected error, but no error: %s",
+                         bson_val_to_json (result->value));
+         goto done;
+      }
+
+      if (is_client_error && *is_client_error) {
+         /* from errors.rst: "In Version 2, error codes originating on the
+          * server always have error domain ``MONGOC_ERROR_SERVER`` or
+          * ``MONGOC_ERROR_WRITE_CONCERN``" */
```
bazile-clyde, line 390: Ack.

src/libmongoc/tests/unified/result.c
```diff
@@ -0,0 +1,564 @@
...
+            test_set_error (error, "utf8 codeName not found in error reply: %s",
+      tmp_json (result->reply)); goto done;
+         }
+
+         if (0 !=
+             bson_strcasecmp (bson_iter_utf8 (&iter, NULL), error_code_name)) {
+            test_set_error (error, "expected codeName: %s, got: %s",
+      error_code_name, bson_iter_utf8 (&iter, NULL)); goto done;
+         }
+      }
```
bazile-clyde, line 443: Ack

bazile-clyde (changes requested):
Just these last few comments. Otherwise, LGTM.

bazile-clyde (approved):

PR #723, opened 2021-01-21
get_rand mac supports arc4random api too

devnexen:


kevinAlbs:
Thank you for the contribution!

PR #724, opened 2021-01-21
CSOT public setters and getters

samantharitter:
I broke this work out from the change_streams CSOT work because these helpers seemed like potential blockers for a lot of the tickets in the Client-Side Operations Timeout ticket.


src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -506,3 +513,40 @@ mongoc_client_pool_enable_auto_encryption (mongoc_client_pool_t *pool,
    return _mongoc_cse_client_pool_enable_auto_encryption (
       pool->topology, opts, error);
 }
+
+bool
+mongoc_client_pool_set_timeout_ms (mongoc_client_pool_t *pool,
+				   int64_t timeout,
+				   bson_error_t *error) {
+   if (!pool) {
```
samantharitter, line 521: Checking that the class parameter in class methods is not NULL isn't generally something we do at all in the driver right now.  So if a user uses most of those methods with a NULL client/pool/db/coll, the program could seg fault or return garbage of some kind (see `mongoc_client_pool_enable_auto_encryption` just above this method).  

This seemed like a much more graceful way to handle that situation.  If you like it, then perhaps it could become our convention for methods like this.

src/libmongoc/src/mongoc/mongoc-collection-private.h
```diff
@@ -37,6 +37,9 @@ struct _mongoc_collection_t {
    mongoc_read_concern_t *read_concern;
    mongoc_write_concern_t *write_concern;
    bson_t *gle;
+   int64_t timeout_ms;
+   bool timeout_set;
+   mongoc_database_t *parent_db;
```
samantharitter, line 42: Up until this point, collections have not tracked their parent database objects, if they had one.  Keeping track of that relationship is necessary if we want to have timeoutMS inherit up the hierarchy according to spec.  However, it does introduce a lifetime dependency of collections on their parent databases.

jmikola: Note that collections obtained via `mongoc_client_get_collection` wouldn't even have a parent database object.

Is there any reason not to inherit the timeout value at the time the collection is instantiated (from either the database or client) and not worry about the parent object's state being modified after the fact. AFAIK, that's how RP/RC/WC inheritance works.

src/libmongoc/doc/mongoc_client_pool_get_timeout_ms.rst
```diff
@@ -0,0 +1,24 @@
+:man_page: mongoc_client_pool_get_timeout_ms
+
+mongoc_client_pool_get_timeout_ms()
```
jmikola, line 3: Does `mongoc_client_pool_t` require an API to get/set `timeoutMS`?

This may be naive, but I was expecting `timeoutMS` to be managed like RP/RC/WC. Those are maintained on the pool's topology (which contains a `mongoc_uri_t`) and fed to popped clients via `_mongoc_client_new_from_uri`.

src/libmongoc/doc/mongoc_client_set_timeout_ms.rst
```diff
@@ -0,0 +1,32 @@
...
+
+Synopsis
+--------
+
+.. code-block:: c
+
+  bool
+  mongoc_client_set_timeout_ms (mongoc_client_t *client, int64_t *timeout_ms, bson_error_t *error)
+
+Sets the client-side operation timeout value for ``client`` to the given ``timeout_ms``. If there is an error setting the timeout, this method will return false and ``error`` will be set. If ``client`` belongs to a :symbol:`mongoc_client_pool_t`, its timeout will be reset to match the pool's timeout once it is pushed back into the pool.
```
jmikola, line 14: This last sentence makes me wonder about other modifications made to clients checked out from the pool.

Suppose a pool is created with `w:majority` via its URI. And suppose a client is popped, modified with `w:1` (or worse), and later pushed back into the pool. When some other thread happens to pop the same client from the pool, would it be in for a surprise if it expected the client to have the original WC from the URI (`w:majority`)? If so, this sounds like a general bug worthy of a ticket (e.g. "Revert state `mongoc_client_t` when pushing back to pool").

src/libmongoc/src/mongoc/mongoc-client-pool.c
jmikola: Isn't `BSON_ASSERT` commonly used in other class methods? Despite the name, it doesn't use the `assert()` (which could be compiled out). On failure, it will `fprintf` to `stderr` and `abort()`.

I see some other pool methods use `BSON_ASSERT`, so maybe this was just missed in `mongoc_client_pool_enable_auto_encryption`.

src/libmongoc/src/mongoc/mongoc-client-private.h
```diff
@@ -122,6 +122,9 @@ struct _mongoc_client_t {
    mongoc_set_t *client_sessions;
    unsigned int csid_rand_seed;
 
+   int64_t timeout_ms;
+   bool timeout_set;
```
jmikola, line 126: Nit suggestion: `is_timeout_set`. Elsewhere in the driver, `is_` is commonly used as a field name prefix for bools.

src/libmongoc/doc/mongoc_client_pool_get_timeout_ms.rst
bazile-clyde: +1

src/libmongoc/doc/mongoc_client_set_timeout_ms.rst
bazile-clyde, line 14: > If ``client`` belongs to a :symbol:`mongoc_client_pool_t`, its timeout will be reset to match the pool's timeout once it is pushed back into the pool.

Is this required by the spec? Also, I don't think we'd need to set timeout MS on a pool.

src/libmongoc/doc/mongoc_client_set_timeout_ms.rst
```diff
@@ -0,0 +1,32 @@
...
+mongoc_client_set_timeout_ms()
+==============================
+
+Synopsis
+--------
+
+.. code-block:: c
+
+  bool
+  mongoc_client_set_timeout_ms (mongoc_client_t *client, int64_t *timeout_ms, bson_error_t *error)
```
bazile-clyde, line 12: Why is ` int64_t *timeout_ms` a pointer? Here and for the other setters as well.

src/libmongoc/doc/mongoc_client_set_timeout_ms.rst
```diff
@@ -0,0 +1,32 @@
...
+This function will set ``error`` and return false in the following cases:
+
+* ``client`` is NULL
+* ``timeout_ms`` is a negative number
+
+Parameters
+----------
+
+* ``client``: A :symbol:`mongoc_client_t`.
+* ``timeout_ms``: A value, in milliseconds, for the timeout.
```
bazile-clyde, line 26: Nit, Maybe we should also specify a "non-negative value" here too.

src/libmongoc/doc/mongoc_database_create_collection.rst
```diff
@@ -25,7 +25,7 @@ Parameters
 Description
 -----------
 
-This function creates a :symbol:`mongoc_collection_t` from the given :symbol:`mongoc_database_t`.
+This function creates a :symbol:`mongoc_collection_t` from the given :symbol:`mongoc_database_t`. ``database`` must not be destroyed while the new :symbol:`mongoc_collection_t` is in use.
```
bazile-clyde, line 28: Why do we need the `mongoc_collection_t` struct to depend on the `mongoc_database_t` struct? Here and in `src/libmongoc/doc/mongoc_database_get_collection.rst`

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1406,7 +1406,8 @@ mongoc_client_get_collection (mongoc_client_t *client,
                                   collection,
                                   client->read_prefs,
                                   client->read_concern,
-                                  client->write_concern);
+                                  client->write_concern,
+				  NULL);
```
bazile-clyde, line 1410: Why not simply have `_mongoc_collection_new` take the `timeout_ms` here instead of the database? That's what we do for the other options, e.g., `write_concern`.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -3080,3 +3081,40 @@ mongoc_client_enable_auto_encryption (mongoc_client_t *client,
...
+mongoc_client_set_timeout_ms (mongoc_client_t *client,
+			      int64_t timeout,
+			      bson_error_t *error) {
+   if (!client) {
+      bson_set_error (error,
+		      MONGOC_ERROR_CLIENT,
+		      MONGOC_ERROR_CLIENT_NULL,
+		      "NULL client parameter");
+	 return false;		      
+   }
```
bazile-clyde, line 3095: For consistency, we should `BSON_ASSERT` the client exists like the other functions in this file.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -3080,3 +3081,40 @@ mongoc_client_enable_auto_encryption (mongoc_client_t *client,
...
+		      MONGOC_ERROR_TIMEOUT,
+		      MONGOC_ERROR_TIMEOUT_INVALID,
+		      "timeoutMS must be a non-negative integer");
+	 return false;
+   }
+
+   client->timeout_ms = timeout;
+   client->timeout_set = true;
+
+   return true;
```
bazile-clyde, line 3108: There's a timeout struct in `mongoc-timeout.h` that ensures a negative timeout is never set and keeps `is_set` and `timeout_ms` in sync among other things. We should use that. Here and in the other setters/getters.

src/libmongoc/src/mongoc/mongoc-collection-private.h
bazile-clyde: +1

bazile-clyde (changes requested):

src/libmongoc/src/mongoc/mongoc-collection-private.h
samantharitter: 🤔  that would mean that if you later changed the parent database's timeoutMS, the child collection would be stuck with an old value.  There's no rule in the spec about not being able to set the timeoutMUS multiple times, just that you can't reset it to "unset" once it is set.

So, I suppose, do we care if the collection doesn't track changes in the parent db's timeoutMS over time?

src/libmongoc/doc/mongoc_client_pool_get_timeout_ms.rst
samantharitter: As far as I can tell the spec is not explicit about how timeoutMS gets set.  It only says "This value MUST be configurable at the level of a MongoClient, MongoDatabase, MongoCollection, or of a single operation."  It does not mention pools, but with the approach that I took here of adding setters and getters, it seemed odd to not also be able to set the timeout on the pool.

It sounds like you would prefer that this option can only be set once initially through the URI, which simplifies the child-tracking-parent-inheritance issue as well.

jmikola: libmongoc is unique among drivers in providing a pool of MongoClients, rather than a connection pool _within_ a MongoClient -- so I doubt we''ll ever see a spec address the API for a pool (apart from something like CMAP, perhaps).

My question about "does `mongo_client_pool_t` require an API" was more about libmongoc's own needs, as I wasn't sure if this was being provided for the potential benefit of users, wrapping drivers (N/A for PHP, but I dunno about C++), etc.

If users haven't complained about RC/RP/WC not being configurable on a pool then I'd err on the side of not introducing other configuration methods like `timeoutMS`.

src/libmongoc/doc/mongoc_client_set_timeout_ms.rst
kevinAlbs: Yes! I filed https://jira.mongodb.org/browse/CDRIVER-3537 some time ago when I noticed this for read concern.

kmahar: this sent me into a panic thinking we might be relying on the "reset" assumption in Swift but fortunately, we are not 🙂

samantharitter: You're right, that does sound like potential surprising and buggy behavior!  We currently don't make any alterations to the client when we return it to the pool, so it could be "dirty" in unexpected ways.  I've opened https://jira.mongodb.org/browse/CDRIVER-3878 to investigate that.

src/libmongoc/doc/mongoc_client_pool_get_timeout_ms.rst
samantharitter: Ok, sounds good!  In my next patch I'll remove the setters for pool and client, but leave them in for database and collection.

src/libmongoc/src/mongoc/mongoc-client-pool.c
samantharitter: Found `BSON_ASSERT_PARAM`, which seems to be made for this purpose!

src/libmongoc/doc/mongoc_client_set_timeout_ms.rst
samantharitter: Ah whoops closed my ticket as a duplicate :D

samantharitter: The spec doesn't say anything about pooling, but this behavior can be addressed in https://jira.mongodb.org/browse/CDRIVER-3537 at a later point in time.

src/libmongoc/doc/mongoc_client_set_timeout_ms.rst
samantharitter: Done.

src/libmongoc/doc/mongoc_database_create_collection.rst
samantharitter: It was so that collection could track changes in the parent db's timeout value to inherit them, but I've just removed that behavior so this dependency is gone, too.

src/libmongoc/src/mongoc/mongoc-client-private.h
samantharitter: Done.

src/libmongoc/src/mongoc/mongoc-client.c
samantharitter: I've changed this to just hold the timeout value, but we only take it if it's set on the database, otherwise it remains unset in the collection as well.

src/libmongoc/src/mongoc/mongoc-client.c
samantharitter: Changed.

src/libmongoc/src/mongoc/mongoc-client.c
samantharitter: Oh nice, I didn't realize we had that type.  Switched to use it.

src/libmongoc/src/mongoc/mongoc-collection-private.h
samantharitter: Changed so that we take the parent db's initial timeout value, but do not track it for changes over time.

src/libmongoc/doc/mongoc_client_set_timeout_ms.rst
samantharitter: Typo! Fixed.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -3080,3 +3088,33 @@ mongoc_client_enable_auto_encryption (mongoc_client_t *client,
...
+}
+
+int64_t
+mongoc_client_get_timeout_ms (const mongoc_client_t *client)
+{
+   BSON_ASSERT_PARAM (client);
+
+   if (!mongoc_timeout_is_set (client->timeout_ms)) {
+      return 0;
+   }
```
bazile-clyde, line 3117: We should also allow the user to distinguish between unset and `0`. The only way I can think of doing that is by exposing `timeout->is_set`.

bazile-clyde (changes requested):

src/libmongoc/src/mongoc/mongoc-client.c
samantharitter: Yes, I suppose they may want to know that as well.  I'll add an additional getter for whether the timeout is set.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -3080,3 +3088,33 @@ mongoc_client_enable_auto_encryption (mongoc_client_t *client,
...
+{
+   BSON_ASSERT_PARAM (client);
+
+   if (timeout < 0) {
+      bson_set_error (error,
+		      MONGOC_ERROR_TIMEOUT,
+		      MONGOC_ERROR_TIMEOUT_INVALID,
+		      "timeoutMS must be a non-negative integer");
+	 return false;
+   }
```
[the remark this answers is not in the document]
samantharitter: Done.

src/libmongoc/doc/mongoc_client_pool_get_timeout_ms.rst
kevinAlbs: Is there value in providing client level setters/getters, when `timeoutMS` can be configured from the URI?

I can see an argument about consistency with `mongoc_client_set_read_concern`, `mongoc_client_set_write_concern`, `mongoc_client_set_read_prefs`, but that already has some weirdness with pooled clients hanging onto that configuration.

It might be worth removing the client (not just client pool) as well. I think it is still spec compliant – you can still configure it at a client level with a URI option. But it avoids the shenanigans of pooled clients.

src/libmongoc/tests/test-mongoc-collection.c
```diff
@@ -6350,6 +6350,63 @@ test_remove_multi (void)
    mongoc_client_destroy (client);
 }
 
+static void
+test_timeout_ms (void)
+{
+   mongoc_client_t *client = mongoc_client_new ("mongodb://host/db_name");
+   mongoc_database_t *db;
+   mongoc_collection_t *coll = mongoc_client_get_collection (client, "db", "test");
+   mongoc_collection_t *coll2;
```
bazile-clyde, line 6359: To be safe, we should set `db` and `coll2` pointers to `NULL`. Here and for other new declarations.

src/libmongoc/src/mongoc/mongoc-client.c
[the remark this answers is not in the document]
bazile-clyde: I agree. Feel free to change whatever. Yeah, this isn't public-facing, so no reason we can't change that type as needed. The only real pain will be updating the expected error messages in `test-mongoc-timeout.c`.

bazile-clyde (changes requested):

src/libmongoc/src/mongoc/mongoc-collection-private.h
jmikola: @samantharitter: I concur. Just like RC/RP/WC, I think it makes sense to inherit at the time of instantiation and not track. Feel free to resolve this thread.

src/libmongoc/doc/mongoc_client_set_timeout_ms.rst
jmikola: Resolving, since [CDRIVER-3537](https://jira.mongodb.org/browse/CDRIVER-3537) exists.

src/libmongoc/doc/mongoc_client_pool_get_timeout_ms.rst
jmikola: If there are already known issues with pooled clients retaining state ([CDRIVER-3537](https://jira.mongodb.org/browse/CDRIVER-3537)), I think it makes sense to hold back on adding more client-level setters and instead create a ticket (dependent on CDRIVER-3537) to add that API down the line.

src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -166,7 +167,8 @@ _mongoc_collection_new (mongoc_client_t *client,
                         const char *collection,
                         const mongoc_read_prefs_t *read_prefs,
                         const mongoc_read_concern_t *read_concern,
-                        const mongoc_write_concern_t *write_concern)
+                        const mongoc_write_concern_t *write_concern,
+			const mongoc_timeout_t *timeout_ms)
```
jmikola, line 171: Don't forget to run `clang-format` before merging.

src/libmongoc/src/mongoc/mongoc-client.c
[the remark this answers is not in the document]
jmikola: > It doesn't really look like mongoc_timeout_t is used elsewhere in the codebase yet, and it's not a public-facing type,

This answers my previous question about the getters returning `int64_t` instead of `mongoc_timeout_t` objects.

If the type is entirely private, would it make sense to use a static struct initializer (e.g. `{0}`) instead of memory allocation in `mongoc_timeout_new()` and just embed the timeout struct in other types? This isn't performance sensitive like BSON and I don't feel strongly about it but wanted to ask in case it was never considered.

jmikola (approved):
All of my previous concerns were addressed, so I'll defer to @bazile-clyde for the rest.

src/libmongoc/doc/mongoc_client_pool_get_timeout_ms.rst
samantharitter: All right, I'll remove the client setter as well.  Since the URI functionality is being added in a separate review, I'll just comment out some of the new tests until we can actually set timeout on a client via URI.

src/libmongoc/tests/test-mongoc-collection.c
samantharitter: It should be fine to leave these newly-declared pointers undefined at the start of a test method.  It's our typical style to do so throughout the codebase.  I agree that it's safer to explicitly set to `NULL`, but given the current conventions in the codebase, I do feel it's a bit unnecessary to do so here (unless there is a push to change this convention that I'm not aware of).

src/libmongoc/src/mongoc/mongoc-collection.c
samantharitter: 👍

src/libmongoc/src/mongoc/mongoc-client.c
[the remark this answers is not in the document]
bazile-clyde: I didn't consider it. I'd be open to it though.

src/libmongoc/tests/test-mongoc-collection.c
bazile-clyde: If you believe it's safer to do so, then we probably should. No need to leave the door open for easily preventable bugs because we've done so in the past. It'd be difficult to improve code quality - in this case, make our codebase safer - if we appeal to tradition.

That being said, I've certainly forgotten to do this at times, so feel free to point it out in my PRs as well.

src/libmongoc/src/mongoc/mongoc-client.c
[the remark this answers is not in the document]
samantharitter: In the interest of getting this code review to the finish line, I'd like to leave `mongoc_timeout_t` as-is for now, but I'd be open to changing this in the future

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -3853,6 +3853,43 @@ test_mongoc_client_recv_network_error (void)
...
+   res = mongoc_client_set_timeout_ms (client, -300, &error);
+   BSON_ASSERT (!res);
+   ASSERT_ERROR_CONTAINS (error,
+			  MONGOC_ERROR_TIMEOUT,
+			  MONGOC_ERROR_TIMEOUT_INVALID,
+			  "timeoutMS must be a non-negative integer");
+
+   /* successful timeouts */
+   ASSERT_OR_PRINT (mongoc_client_set_timeout_ms (client, 300, &error), error);
+   ASSERT_CMPINT (mongoc_client_get_timeout_ms (client), ==, 300);
```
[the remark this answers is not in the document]
samantharitter: Done.

src/libmongoc/tests/test-mongoc-collection.c
samantharitter: The reason I'm pushing back a bit here is that, to me, this kind of feedback falls into a category where, though it's not _purely_ a matter of personal preference, it's close, it isn't super substantive either, and it delays the review for little to no functional difference or benefit in the code.  The code as I have written it here is not incorrect, nor is it risky, since it is in a test.  I don't think that this sort of change is one we need to be pushing for.

I think there has to be a balance in our reviews between the ideal of following a golden standard of style, and following existing conventions in the codebase in a way that allows us to get code into production efficiently.

Perhaps @kevinAlbs has additional thoughts.

kevinAlbs: To weigh in, I do agree. Raising codebase quality is an admirable goal. But unless we have an agreed team standard, style or non-correctness changes may not be consistent and do add to review turnaround time. I think there is a good middle ground though. Code reviews do provide a good forum for sharing ideas and learning as well. I am happy to discuss offline.

src/libmongoc/doc/mongoc_client_get_timeout_ms.rst
```diff
@@ -0,0 +1,24 @@
...
+
+Synopsis
+--------
+
+.. code-block:: c
+
+  int64_t
+  mongoc_client_get_timeout_ms (const mongoc_client_t *client)
+
+Returns the client-side operation timeout value for this object. If ``client`` does not have a timeout currently set, this method will return the default timeout, which is 0, or infinite timeout.
```
kevinAlbs, line 14: Based on the `Backwards Breaking Considerations` section of the spec, I think unset will mean "keep the existing behavior". Otherwise, when we introduce `timeoutMS` support, it could silently override existing default timeout.

E.g. the URI "mongodb://localhost:27017/?sockettimeoutms=1000" may suddenly change to mean no socket timeout (with a deprecation warning).

This might just be a doc change:
```suggestion
Returns the client-side operation timeout value for this object. If ``client`` does not have a timeout currently set, this method will return 0.
```

`0` will have a double meaning here. I am ok with that since callers can determine with `mongoc_client_timeout_is_set`. If we want to avoid that, perhaps this could return an `const int64_t*`, which is NULL on unset. Though I'm fine with this as-is.

src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -3503,3 +3510,39 @@ mongoc_collection_watch (const mongoc_collection_t *coll,
 {
    return _mongoc_change_stream_new_from_collection (coll, pipeline, opts);
 }
+
+bool
+mongoc_collection_set_timeout_ms (mongoc_collection_t *coll,
+				  int64_t timeout,
+				  bson_error_t *error)
+{
+   BSON_ASSERT (coll);
```
kevinAlbs, line 3519:
```suggestion
   BSON_ASSERT_PARAM (coll);
```

src/libmongoc/doc/mongoc_database_set_timeout_ms.rst
```diff
@@ -0,0 +1,31 @@
...
+
+  bool
+  mongoc_database_set_timeout_ms (mongoc_database_t *database, int64_t timeout_ms, bson_error_t *error)
+
+Sets the client-side operation timeout value for ``database`` to the given ``timeout_ms``. If there is an error setting the timeout, this method will return false and ``error`` will be set.
+
+
+This function will set ``error`` and return false in the following cases:
+
+* ``database`` is NULL
```
kevinAlbs, line 19: Remove, the setter asserts `database` is not NULL.

src/libmongoc/doc/mongoc_collection_set_timeout_ms.rst
```diff
@@ -0,0 +1,31 @@
...
+
+  bool
+  mongoc_collection_set_timeout_ms (mongoc_collection_t *collection, int64_t timeout_ms, bson_error_t *error)
+
+Sets the client-side operation timeout value for ``collection`` to the given ``timeout_ms``. If there is an error setting the timeout, this method will return false and ``error`` will be set.
+
+
+This function will set ``error`` and return false in the following cases:
+
+* ``collection`` is NULL
```
kevinAlbs, line 19: Remove, the setter asserts instead.

src/libmongoc/src/mongoc/mongoc-error.h
```diff
@@ -122,7 +123,9 @@ typedef enum {
 
    /* An error related to initializing client side encryption. */
    MONGOC_ERROR_CLIENT_INVALID_ENCRYPTION_STATE,
-   MONGOC_ERROR_CLIENT_INVALID_ENCRYPTION_ARG
+   MONGOC_ERROR_CLIENT_INVALID_ENCRYPTION_ARG,
+
+   MONGOC_ERROR_TIMEOUT_INVALID
```
kevinAlbs, line 128: Can you add to the errors.rst tables with the new domain/code?

src/libmongoc/doc/mongoc_client_get_timeout_ms.rst
jmikola: Would the returned `int64_t*` point to the `int64_t` _within_ the internal timeout struct?

I suppose this is consistent with the RC/RP/WC getters, which return the pointer _within_ the client struct. In that case, the pointers are valid for either the lifetime of the client struct or until a new RC/RP/WC is assigned, whichever comes first.

It seems a bit odd to do for an integer type, though. AFAIK, the CSOT spec doesn't prohibit you from returning a negative value (i.e. -1) to communicate that it's unset -- I don't imagine that'd be different than other languages returning a nullable integer. You could also define a public constant (e.g. `TIMEOUTMS_UNSET` or `TIMEOUTMS_DEFAULT`), as is done for `MONGOC_WRITE_CONCERN_W_DEFAULT`. I think the only difference from WC would be that you wouldn't be able to specify this constant via the setter methods -- but that seems fine.

Just my $.02.

kevinAlbs: Yes, I was thinking it would return a pointer to within the struct.

A negative `TIMEOUTMS_UNSET` sentinel seems like a good idea too. I think it is unlikely a negative `timeoutMS` will ever mean something.

FWIW I am also ok with the current proposed API with two functions.

jmikola: I suppose if you opt for defining `TIMEOUTMS_UNSET` as a negative, you could get by with removing both the `is_set` function _and_ the struct. The public APIs that set `timeoutMS` (e.g. URI, setter methods, method options) could just require a non-negative value (I assume they may already) and fall back to -1 or inherit as the default behavior accordingly.

Seems out of scope for this PR. Do you want to take it offline for future consideration?

src/libmongoc/doc/mongoc_collection_set_timeout_ms.rst
samantharitter: Done.

src/libmongoc/src/mongoc/mongoc-error.h
samantharitter: Done,

samantharitter: Done.

src/libmongoc/doc/mongoc_database_set_timeout_ms.rst
samantharitter: Done.

src/libmongoc/doc/mongoc_client_get_timeout_ms.rst
samantharitter: I've changed the documentation as @kevinAlbs suggested.

I agree with @jmikola that it would be simpler to use -1 as a special value to mean `TIMEOUTMS_UNSET`.  In the interest of not flip-flopping too much on this code review, I'd like consensus before making that change (because it will change a lot of the code in this review).  So, thoughts, @kevinAlbs and @bazile-clyde ?  If we use -1 as a special value, I will remove the `is_set` getter methods.

kevinAlbs: Defining a `TIMEOUTMS_UNSET` as `-1` sounds good to me!

bazile-clyde: Sounds good to me too.

src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -3503,3 +3508,35 @@ mongoc_collection_watch (const mongoc_collection_t *coll,
...
 }
+
+bool
+mongoc_collection_set_timeout_ms (mongoc_collection_t *coll,
+                                  int64_t timeout_ms,
+                                  bson_error_t *error)
+{
+   BSON_ASSERT_PARAM (coll);
+
+   if (timeout_ms < -1) {
```
kevinAlbs, line 3519: I think this can error if `timeout_ms < 0`.

According to the spec, users should not be able to set `timeoutMS` to unset after it has been set:

> Drivers MUST error if a negative value is specified. This value MUST be configurable at the level of a MongoClient, MongoDatabase, MongoCollection, or of a single operation. However, if the option is specified at any level, it cannot be later changed to unset.

src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -3503,3 +3508,35 @@ mongoc_collection_watch (const mongoc_collection_t *coll,
...
+   return true;
+}
+
+int64_t
+mongoc_collection_get_timeout_ms (const mongoc_collection_t *coll)
+{
+   BSON_ASSERT_PARAM (coll);
+
+   if (coll->timeout_ms == MONGOC_TIMEOUTMS_UNSET) {
+      return mongoc_client_get_timeout_ms (coll->client);
```
kevinAlbs, line 3538: I think this should return from the database.

src/libmongoc/src/mongoc/mongoc-database.c
```diff
@@ -1029,3 +1032,34 @@ mongoc_database_watch (const mongoc_database_t *db,
...
 }
+
+bool
+mongoc_database_set_timeout_ms (mongoc_database_t *db,
+                                int64_t timeout_ms,
+                                bson_error_t *error)
+{
+   BSON_ASSERT_PARAM (db);
+
+   if (timeout_ms < -1) {
```
kevinAlbs, line 1043: Same comment, I think this can be a strict negative check and disallow `-1` too.

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -3853,6 +3853,23 @@ test_mongoc_client_recv_network_error (void)
...
 }
 
+void
+test_mongoc_client_timeout_ms (void)
+{
+   mongoc_client_t *client;
+
+   client = mongoc_client_new ("mongodb://localhost");
+
+   /* no timeout returns 0 */
```
kevinAlbs, line 3863:
```suggestion
   /* no timeout returns -1 */
```

kevinAlbs (commented):
Using the unset sentinel does look cleaner!

samantharitter:
This is ready for re-review!  The automated checks are not running for some reason, so here's a manual evergreen patch: https://evergreen.mongodb.com/version/6023f6773066155e29379918

src/libmongoc/src/mongoc/mongoc-collection.c
samantharitter: So, the collection doesn't keep track of its parent database.  If the parent db did have a timeoutMS set when it created this collection, it would have passed that value on to the collection, and coll->timeout_ms would not be UNSET.  Returning the client's timeout seemed like the best possible choice here, and still makes some attempt to follow the hierarchy.  

An earlier patch of this code review added in a parent db pointer onto the collection, so it could track changes in the parent db's timeout.  The feedback I got was that this wasn't ideal or necessary, since other things we inherit in a hierarchical fasion (like write_concern) function fine without it.

src/libmongoc/src/mongoc/mongoc-collection.c
samantharitter: Oh good point!  Will change.

build/evergreen_config_lib/functions.py
```diff
@@ -82,6 +82,11 @@
         sh .evergreen/check-release-archive.sh
         ''', xtrace=False),
     )),
+    ('set api version', Function(
+        shell_mongoc(r'''
+        export MONGODB_API_VERSION=1
+        '''),
+    )),
```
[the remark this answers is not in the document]
samantharitter: ahhh whoops!  Will remove.

src/libmongoc/src/mongoc/mongoc-database.c
samantharitter: Will change.

src/libmongoc/src/mongoc/mongoc-collection.c
kevinAlbs: Discussed offline. It seems preferable to be consistent with read concern / write concern / read preferences. Those are inherited when a collection struct is constructed, and will not change if the parent database / client's options change.

samantharitter:
Another manual run on Evergreen for the most recent changes: https://evergreen.mongodb.com/version/60242bd557e85a0a0bfe96ae

src/libmongoc/src/mongoc/mongoc-client.h
```diff
@@ -268,6 +268,9 @@ mongoc_client_enable_auto_encryption (mongoc_client_t *client,
                                       mongoc_auto_encryption_opts_t *opts,
                                       bson_error_t *error);
 
+MONGOC_EXPORT (int64_t)
+mongoc_client_get_timeout_ms (const mongoc_client_t *client);
+
```
bazile-clyde, line 273: Why not add `mongoc_client_set_timeout_ms`? There's plenty of setters for key-values that can also be set in the URI. For example, `appname`, `readPrefs`, etc.

src/libmongoc/src/mongoc/mongoc-database-private.h
```diff
@@ -43,7 +44,8 @@ _mongoc_database_new (mongoc_client_t *client,
                       const char *name,
                       const mongoc_read_prefs_t *read_prefs,
                       const mongoc_read_concern_t *read_concern,
-                      const mongoc_write_concern_t *write_concern);
+                      const mongoc_write_concern_t *write_concern,
+		      int64_t timeout_ms);
```
bazile-clyde, line 48: clang-format?

src/libmongoc/src/mongoc/mongoc-timeout-private.h
```diff
@@ -22,13 +22,15 @@
 #include <stdint.h>
 #include <stdbool.h>
 
+#include <bson/bson.h>
+
```
bazile-clyde, line 26: Does this PR need any of these changes to the timeout struct? Since we aren't using it anymore I think the changes can be reverted.

src/libmongoc/tests/test-mongoc-collection.c
```diff
@@ -6350,6 +6350,76 @@ test_remove_multi (void)
...
+                  ==,
+                  mongoc_database_get_timeout_ms (db));
+
+   /* We do not track parent db timeout changes over time, though */
+   ASSERT_OR_PRINT (mongoc_database_set_timeout_ms (db, 200, &error), error);
+   ASSERT_CMPINT (mongoc_collection_get_timeout_ms (coll2), ==, 50);
+
+   ASSERT_OR_PRINT (mongoc_collection_set_timeout_ms (coll2, 300, &error),
+                    error);
+   ASSERT_CMPINT (mongoc_collection_get_timeout_ms (coll2), ==, 300);
```
bazile-clyde, line 6416: Duplicate tests? This same thing is done on lines 6381-6383

bazile-clyde (changes requested):

src/libmongoc/src/mongoc/mongoc-collection.c
samantharitter: Done.

src/libmongoc/doc/mongoc_client_get_timeout_ms.rst
```diff
@@ -0,0 +1,24 @@
...
+
+Parameters
+----------
+
+* ``client``: A :symbol:`mongoc_client_t`.
+
+Returns
+-------
+
+The timeout set on this client or 0.
```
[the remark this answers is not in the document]
samantharitter: Good catch, missed that change!

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1327,7 +1330,8 @@ mongoc_client_get_database (mongoc_client_t *client, const char *name)
                                 name,
                                 client->read_prefs,
                                 client->read_concern,
-                                client->write_concern);
+                                client->write_concern,
+				client->timeout_ms);
```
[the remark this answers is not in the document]
samantharitter: Yes, I'll run that again... something about my editor makes these lines appear lined up with each other when they actually aren't, maybe it's a tabs vs. spaces thing.

src/libmongoc/src/mongoc/mongoc-client.h
samantharitter: I originally did have a `mongoc_client_set_timeout_ms`, but I removed it.  The feedback I got from @jmikola and @kevinAlbs was that we should only allow timeoutMS to be set on clients through the URI, at least for now, since that helps avoid some of the weirdness of keeping properties of pooled clients consistent with their parent pools.  The spec does not mandate that there be a setter for clients, just that the timeout "MUST be configurable at the level of a MongoClient."

src/libmongoc/src/mongoc/mongoc-database-private.h
samantharitter: Yes, same weirdness with my editor, I'll do a final clang-format before committing this PR.

src/libmongoc/src/mongoc/mongoc-timeout-private.h
samantharitter: I will remove the changes to the timeout struct.

src/libmongoc/tests/test-mongoc-collection.c
samantharitter: These lines are to test that the collection doesn't keep using the timeout set on its parent `db` on L6411.  It's just a coincidence that it uses the same value `300`, it does not matter what value it uses as long as it's different from the parent `db`'s timeoutMS.

src/libmongoc/src/mongoc/mongoc-client.h
bazile-clyde: Oh, [CDRIVER-3537](https://jira.mongodb.org/browse/CDRIVER-3537) right? Gotcha.

kevinAlbs (approved):
LGTM with clang-format and the timeout struct changes reverted.

samantharitter:
New manual patch test: https://evergreen.mongodb.com/version/602695fa2a60ed253f65555a

I squashed my commits, rebased, resolved the conflicts, and force-pushed to this branch to get it ready to merge, but it looks like all the rebased commits from master are now in this PR.  Sorry about that!

bazile-clyde (approved):
LGTM!

samantharitter:
Closing this PR, manually pushed.

PR #726, opened 2021-01-24
CDRIVER-3780 add client_session_get_dirty

kevinAlbs:
Unified test format (and driver sessions tests) make assertions on whether a session has been marked dirty. This adds a getter to `mongoc_client_session_t` so wrapping drivers can easily implement those assertions.

### DIFF
--- a/src/libmongoc/doc/mongoc_client_session_get_dirty.rst
+++ b/src/libmongoc/doc/mongoc_client_session_get_dirty.rst
@@ -1,0 +1,33 @@
+:man_page: mongoc_client_session_get_dirty
+
+mongoc_client_session_get_dirty()
+=================================
+
+Synopsis
+--------
+
+.. code-block:: c
+
+  bool
+  mongoc_client_session_get_dirty (const mongoc_client_session_t *session);
+
+Indicates whether ``session`` has been marked "dirty" as defined in the `driver sessions specification <https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst>`_.
+
+Parameters
+----------
+
+* ``session``: A const :symbol:`mongoc_client_session_t`.
+
+Description
+-----------
+
+This function is intended for use by drivers that wrap libmongoc. It is not useful in client applications.
+
+Returns
+-------
+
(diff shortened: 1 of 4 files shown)


src/libmongoc/doc/mongoc_client_session_get_dirty.rst
```diff
@@ -0,0 +1,33 @@
+:man_page: mongoc_client_session_get_dirty
+
+mongoc_client_session_get_dirty()
```
kevinAlbs, line 3: The name is admittedly odd. But the rationale is:
1. `_get` is consistent with other client session methods like `mongoc_client_session_get_server_id`.
2. `dirty` matches terminology from the specification.

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -1613,3 +1613,11 @@ _mongoc_client_session_pin (mongoc_client_session_t *session,
 
    session->server_id = server_id;
 }
+
+bool
+mongoc_client_session_get_dirty (mongoc_client_session_t *session)
+{
+   BSON_ASSERT_PARAM (session);
+
+   return session->server_session->dirty;
```
kevinAlbs, line 1622: Note, no need to assert `server_session` as it is added on construction of a client session. `mongoc_client_session_get_lsid` makes the same assumption.

bazile-clyde: I see. It does look like we should assert the `server_session` arg in `_mongoc_client_session_new` is never `NULL`, though—that way, our assumption that a `mongoc_client_session_t` always has a non-NULL `server_session` is enforced. 

Since that function is only used in by `mongoc_client_start_session`, which checks the server session is non-NULL, it doesn't look like it'd be a problem for now. So I'm happy leaving this as is.

bazile-clyde (approved):
LGTM. One optional comment.

src/libmongoc/src/mongoc/mongoc-client-session.c
kevinAlbs: I agree, added.

PR #727, opened 2021-01-24
CDRIVER-3873 clear fam error on retry

kevinAlbs:
If `mongoc_collection_find_and_modify_with_opts` or `mongoc_collection_find_and_modify` succeeds after retrying due to retryable writes being enabled, it correctly returns `true` but leaves the output `bson_error_t` set to the original error.

This clears the error on a successful retry.

Note, this came up during implementation of the unified test runner. I implemented error checking by checking the out `bson_error_t` on operations (not the boolean return value).

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-collection.c
+++ b/src/libmongoc/src/mongoc/mongoc-collection.c
@@ -3281,18 +3281,17 @@ mongoc_collection_find_and_modify_with_opts (
    }
    /* inherit write concern from collection if not in transaction */
    else if (server_stream->sd->max_wire_version >=
-             WIRE_VERSION_FAM_WRITE_CONCERN &&
-          !_mongoc_client_session_in_txn (parts.assembled.session)) {
-         if (!mongoc_write_concern_is_valid (collection->write_concern)) {
-            bson_set_error (error,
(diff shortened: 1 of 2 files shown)


src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -3281,18 +3281,17 @@ mongoc_collection_find_and_modify_with_opts (
...
+            !_mongoc_client_session_in_txn (parts.assembled.session)) {
+      if (!mongoc_write_concern_is_valid (collection->write_concern)) {
+         bson_set_error (error,
+                         MONGOC_ERROR_COMMAND,
+                         MONGOC_ERROR_COMMAND_INVALID_ARG,
+                         "The write concern is invalid.");
+         GOTO (done);
+      }
 
+      write_concern = collection->write_concern;
```
bazile-clyde, line 3294: Is this just clang-format?

bazile-clyde (approved):

src/libmongoc/src/mongoc/mongoc-collection.c
kevinAlbs: Yep! Just clang-format.

PR #728, opened 2021-01-25
CDRIVER-3147 add cmd errors to write results

kevinAlbs:
Add code, codeName, and errmsg to write results from a reply
when a command error (i.e. not a writeError / writeConcernError)
occurs during a write.

### DIFF
--- a/src/libmongoc/tests/test-mongoc-write-commands.c
+++ b/src/libmongoc/tests/test-mongoc-write-commands.c
@@ -634,6 +634,7 @@ _test_command_error (void *unused)
    bool ret;
    bson_t reply;
    bson_error_t error;
+   mongoc_bulk_operation_t *bulk;
 
    client = test_framework_client_new ();
    mongoc_client_set_error_api (client, MONGOC_ERROR_API_VERSION_2);
@@ -651,8 +652,27 @@ _test_command_error (void *unused)
    BSON_ASSERT (bson_has_field (&reply, "errmsg"));
    ASSERT_CMPINT (123, ==, bson_lookup_int32 (&reply, "code"));
    _configure_failpoint (client, "'off'", "{}");
-
    bson_destroy (&reply);
+
+   /* Test a bulk write insert. */
+   bulk = mongoc_collection_create_bulk_operation_with_opts (coll, NULL);
+   ASSERT_OR_PRINT (mongoc_bulk_operation_insert_with_opts (
+                       bulk, tmp_bson ("{'x': 1}"), NULL, &error),
+                    error);
+   _configure_failpoint (client,
+                         "{'times': 2}",
+                         "{'failCommands': ['insert'], 'errorCode' : 123 }");
+   ret = mongoc_bulk_operation_execute (bulk, &reply, &error);
+   BSON_ASSERT (!ret);
+   ASSERT_ERROR_CONTAINS (error, MONGOC_ERROR_SERVER, 123, "");
+   BSON_ASSERT (bson_has_field (&reply, "code"));
+   BSON_ASSERT (bson_has_field (&reply, "codeName"));
+   BSON_ASSERT (bson_has_field (&reply, "errmsg"));
+   ASSERT_CMPINT (123, ==, bson_lookup_int32 (&reply, "code"));
+   _configure_failpoint (client, "'off'", "{}");
+   bson_destroy (&reply);
+   mongoc_bulk_operation_destroy (bulk);
+
    mongoc_collection_destroy (coll);
    mongoc_client_destroy (client);
 }
--- a/src/libmongoc/tests/test-mongoc-write-commands.c
+++ b/src/libmongoc/tests/test-mongoc-write-commands.c
@@ -626,6 +626,37 @@ _test_invalid_wc_server_error (void *unused)
    mongoc_client_destroy (client);
 }
 
+static void
+_test_command_error (void *unused)
+{
+   mongoc_client_t *client;
+   mongoc_collection_t *coll;
+   bool ret;
+   bson_t reply;
+   bson_error_t error;
+
+   client = test_framework_client_new ();
+   mongoc_client_set_error_api (client, MONGOC_ERROR_API_VERSION_2);
+   coll = get_test_collection (client, "server_wc_error");
+
+   _configure_failpoint (client,
+                         "{'times': 2}",
+                         "{'failCommands': ['insert'], 'errorCode' : 123 }");
+   ret = mongoc_collection_insert_one (
+      coll, tmp_bson ("{'x':1}"), NULL /* opts */, &reply, &error);
+   BSON_ASSERT (!ret);
+   ASSERT_ERROR_CONTAINS (error, MONGOC_ERROR_SERVER, 123, "");
+   BSON_ASSERT (bson_has_field (&reply, "code"));
+   BSON_ASSERT (bson_has_field (&reply, "codeName"));
+   BSON_ASSERT (bson_has_field (&reply, "errmsg"));
+   ASSERT_CMPINT (123, ==, bson_lookup_int32 (&reply, "code"));
+   _configure_failpoint (client, "'off'", "{}");
+
+   bson_destroy (&reply);
+   mongoc_collection_destroy (coll);
+   mongoc_client_destroy (client);
+}
+
 void
 test_write_command_install (TestSuite *suite)
 {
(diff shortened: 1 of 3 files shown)


src/libmongoc/tests/test-mongoc-write-commands.c
```diff
@@ -626,6 +626,37 @@ _test_invalid_wc_server_error (void *unused)
...
+   _configure_failpoint (client,
+                         "{'times': 2}",
+                         "{'failCommands': ['insert'], 'errorCode' : 123 }");
+   ret = mongoc_collection_insert_one (
+      coll, tmp_bson ("{'x':1}"), NULL /* opts */, &reply, &error);
+   BSON_ASSERT (!ret);
+   ASSERT_ERROR_CONTAINS (error, MONGOC_ERROR_SERVER, 123, "");
+   BSON_ASSERT (bson_has_field (&reply, "code"));
+   BSON_ASSERT (bson_has_field (&reply, "codeName"));
+   BSON_ASSERT (bson_has_field (&reply, "errmsg"));
```
jmikola, line 651: Would it be too much trouble to ask that you also verify these fields will be available via the bulk API? See [PHPC-1386](https://jira.mongodb.org/browse/PHPC-1386) for how we originally came across this issue.

kevinAlbs: Not at all! AFAIK all write results go through the same code path for constructing and merging write results, but it seems like a good idea to be sure.

I added a test for bulk.

bazile-clyde (approved):

src/libmongoc/tests/test-mongoc-write-commands.c
```diff
@@ -626,6 +626,57 @@ _test_invalid_wc_server_error (void *unused)
...
+   ASSERT_OR_PRINT (mongoc_bulk_operation_insert_with_opts (
+                       bulk, tmp_bson ("{'x': 1}"), NULL, &error),
+                    error);
+   _configure_failpoint (client,
+                         "{'times': 2}",
+                         "{'failCommands': ['insert'], 'errorCode' : 123 }");
+   ret = mongoc_bulk_operation_execute (bulk, &reply, &error);
+   BSON_ASSERT (!ret);
+   ASSERT_ERROR_CONTAINS (error, MONGOC_ERROR_SERVER, 123, "");
+   BSON_ASSERT (bson_has_field (&reply, "code"));
```
jmikola, line 668: Consider a case where a bulk write with `unordered=true` encountered multiple command failures. Would the last command clobber the `code`, `codeName`, and `errmsg` fields set by an earlier command failure?

Per the documentation examples in [Bulk Insert](http://mongoc.org/libmongoc/1.17.1/bulk.html#bulk-insert), I would have expected the error to be reported within a `writeErrors` array (or potentially `writeConcernErrors`). Looking at the [`mongoc_collection_insert_one`](http://mongoc.org/libmongoc/1.17.0/mongoc_collection_insert_one.html#description), that method uses the same format for its reply document (granted, the arrays would only ever contain at most one document).

Perhaps this problem is unique to the CRUD methods, since they return formatted result documents instead of raw server replies. In that case, would it be preferable to amend the CRUD spec to support an optional `codeName` field on the WriteError and WriteConcernError objects described in [Error Handling](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#error-handling)?

kevinAlbs: Good point. Right now, if `unordered=true` encounters multiple command failures this would only keep the last.

I'll investigate a bit, but it feels wrong to report server errors through `writeErrors`. `writeErrors` maps to the input documents.

Example:
```
> db.runCommand({insert: "coll", documents:[{x:1},{_id:1},{_id:1},{_id:2},{_id:2}], ordered:false})
{
	"ok" : 1,
	"writeErrors" : [
        { "index" : 1, "code" : 11000, ... },
        { "index" : 4, "code" : 11000, ... }
    ],
    ...
}
```
Can report writeErrors with indexes 1 and 4 for the two duplicate ID errors.

A command error doesn't directly map to documents:
```
db.runCommand({insert: "coll", documents:[{x:1},{y:1}], bad:1})
{
	"ok" : 0,
	"errmsg" : "BSON field 'insert.bad' is an unknown field.",
	"code" : 40415,
	"codeName" : "Location40415"
}
```

I'm not sure how to best represent that in writeErrors.

I'll investigate further.
1. I'm not sure what other drivers do in the case of an unordered bulk write that fails with multiple command errors. 
2. I wonder if we should be stopping an unordered bulk write on the first command (`ok:0`) error. Glancing over the [On Errors](https://github.com/mongodb/specifications/blob/master/source/driver-bulk-update.rst#on-errors) section of the bulk write spec gives me that impression.

kevinAlbs: I tested in [Go](https://gist.github.com/kevinAlbs/11e6d6864804e8cb296d15878dc7f19e) and [pymongo](https://gist.github.com/kevinAlbs/1f8e0d6961499be70eeddd19041e410d).

Go continues to send commands after a command error, and clobbers the last command error in the case of multiple command errors. That logic occurs here: https://github.com/mongodb/mongo-go-driver/blob/release/1.4/mongo/bulk_write.go#L76:L88. 

pymongo raises an OperationError on the first command error.

The [General Rule of Error Handling](https://github.com/mongodb/specifications/blob/master/source/driver-bulk-update.rst#general-rule-of-handling-errors) seems to agree with pymongo:
> A top level error means the whole command failed and should cause a command failure error.

But I think future spec changes will address how multiple bulk write command errors should be reported. (SPEC-976 / new bulk write spec).

`writeErrors` seems like the wrong place to report command errors.

Introducing an array of `commandErrors` might be risky until we have know it is spec compliant (it also begs the question of how wrapping drivers should interpret that array of errors).

Perhaps I should leave code/codeName/errmsg out of unordered bulk write results until it is specified?

jmikola: The CRUD spec's [Error Handling](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#error-handling) doesn't seem to refer to command failures (i.e. `{ok:0}`) at all. I believe the intention is that those would result in a separate exception type (e.g. CommandFailureException), which isn't documented in the spec. And of course there's no discussion about how command failures pertain to bulk writes, which is where [SPEC-976](https://jira.mongodb.org/browse/SPEC-976) comes in.

AFAIK, libmongoc doesn't claim to fully conform to the CRUD spec, but its reply documents for these CRUD methods are, for the most part, modeled after the classes in the CRUD spec. Perhaps the relevant API change to libmongoc would be to claim that in the event of a command failure, `reply` shifts from being modeled after a CRUD result/error and instead just contains the original command response in its entirety. Would that be more consistent with other methods in the driver? One question would be: do you feel comfortable addressing this now or would it be wiser to defer until we can address this in the specs? So far, if this is only relevant to wrapping drivers attempting to implement spec tests, they can work around that by managing a map of code names [as I did in PHPLIB](https://github.com/mongodb/mongo-php-library/blob/1.8.0/tests/SpecTests/ErrorExpectation.php#L201).

This still leaves bulk writes as something to address. And perhaps we still need a separate spec ticket to clarify how command failures should be reported for CRUD methods (not merely with respect to bulk writes per [SPEC-976]([SPEC-976](https://jira.mongodb.org/browse/SPEC-976)).

kevinAlbs: > Perhaps the relevant API change to libmongoc would be to claim that in the event of a command failure, reply shifts from being modeled after a CRUD result/error and instead just contains the original command response in its entirety. Would that be more consistent with other methods in the driver?

That could work. I could also see an option of changing the `reply` to have a field containing the original server reply on error:

```
{
   insertedCount: 0,
   errorLabels: [...] // error labels from reply with labels added from driver
   errorReply: { ok: 0, errmsg: "foo", errorLabels: [...] ... }
}
```

That may help future proof. In the case of additional server error fields (`errinfo` comes to mind), they will continue to be exposed. Depending on the outcome of SPEC-976, it could be an array of server replies.

> One question would be: do you feel comfortable addressing this now or would it be wiser to defer until we can address this in the specs? So far, if this is only relevant to wrapping drivers attempting to implement spec tests, they can work around that by managing a map of code names as I did in PHPLIB.

Since there is a straightforward workaround, I would rather wait until after updating the specs. I filed https://jira.mongodb.org/browse/DRIVERS-1547 to clarify how command errors should be reported.

jmikola: I concur with waiting until we can sort this out in a spec. This isn't urgent for PHP and I don't imagine Swift is much different.

PR #733, opened 2021-02-04
Allow custom bind_ip for outgoing client connections

samantharitter:
This PR is still in a pretty rough state, but I wanted to get some feedback on it before continuing.


src/libmongoc/doc/mongoc_client_pool_set_bind_ip.rst
```diff
@@ -0,0 +1,32 @@
...
+   bool
+   mongoc_client_pool_set_bind_ip (mongoc_client_pool_t *pool, const char *bind_ip, bson_error_t *error);
+
+Sets a custom bind IP for ``pool``.  All clients from ``pool`` will bind() all outgoing TCP socket connections to the address ``bind_ip`` before calling connect(). Call this method immediately after creating ``pool`` and before getting a :symbol:`mongoc_client_t` from ``pool`` to ensure that no connections are opened before the custom IP address is set.
+
+``bind_ip`` is copied, and does not have to remain valid after the call to ``mongoc_client_pool_set_bind_ip()``.
+
+This function will set ``error`` and return false in the following cases:
+
+* ``bind_ip`` is not a valid IPv4 address
```
samantharitter, line 20: I would like to also support IPv6.

src/libmongoc/doc/mongoc_client_pool_set_bind_ip.rst
```diff
@@ -0,0 +1,32 @@
+:man_page: mongoc_client_pool_set_bind_ip
+
+mongoc_client_pool_set_bind_ip()
```
samantharitter, line 3: If we only offered a client-level setter, then there would be a gap between when the first outgoing connections are made in the driver and when the user could set a custom bind_ip.  So, monitoring might not use the custom bind_ip right away.  Offering a pool-level setter avoids this gap.

src/libmongoc/src/mongoc/mongoc-http.c
```diff
@@ -72,7 +72,7 @@ _mongoc_http_send (mongoc_http_request_t *req,
       goto fail;
    }
 
-   stream = mongoc_client_connect_tcp (timeout_ms, &host_list, error);
+   stream = mongoc_client_connect_tcp (NULL, timeout_ms, &host_list, error);
```
samantharitter, line 75: I don't believe it's necessary to support custom bind_ips for http servers

src/libmongoc/src/mongoc/mongoc-socket.c
```diff
@@ -782,6 +782,108 @@ mongoc_socket_bind (mongoc_socket_t *sock,       /* IN */
...
+      }
+   } else {
+      MONGOC_WARNING ("Invalid bind address: %s", bind_ip);
+      return NULL;
+   }
+
+   fprintf (stderr, "passed-in family/domain is %d\n", domain);
+   fprintf (stderr, "determined family is %d\n", bind_addr.sin_family);
+
+   // TODO SAM: this loop shouldn't be infinite. Use expire_at ? Num retries? One retry?
```
samantharitter, line 826: Thoughts on how to structure this loop?

This article talks about `SO_REUSEADDR` and explains why this retry loop is here:
https://idea.popcount.org/2014-04-03-bind-before-connect/

It is probably not strictly necessary to use `SO_REUSEADDR`, but I think it's a little more resilient to use it.  In a situation where the user's machine was opening a lot (like, a _lot_) of connections, including many made through the C driver, `SO_REUSEADDR` would help push up the hard limit on the number of connections that could be made.  But if we don't think it's worth using, then the logic in this method would be much simpler.

src/libmongoc/src/mongoc/mongoc-topology-scanner-private.h
```diff
@@ -109,6 +110,9 @@ typedef struct mongoc_topology_scanner {
    bool negotiate_sasl_supported_mechs;
    bool bypass_cooldown;
    bool speculative_authentication;
+
+   bson_mutex_t ip_mutex;
+   char *bind_ip;
```
samantharitter, line 115: The topology scanner is the owner of the custom `bind_ip.`  Pools and clients and the SDAM framework all have access to the topology scanner and so can share the same custom `bind_ip` if it is stored here.

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -2195,6 +2195,49 @@ test_mongoc_client_unix_domain_socket (void *context)
...
+   ASSERT_OR_PRINT (
+      mongoc_client_read_command_with_opts (
+         client, "admin", tmp_bson ("{'ping': 1}"), NULL, NULL, NULL, &error),
+      error);
+
+   mongoc_client_destroy (client);
+   client = mongoc_client_new ("mongodb://localhost");
+   BSON_ASSERT (client);
+
+   // Test an IPv6 address
```
samantharitter, line 2228: IPv6 does not work right now.

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -2195,6 +2195,49 @@ test_mongoc_client_unix_domain_socket (void *context)
...
+      error);
+   mongoc_client_destroy (client);
+   fprintf (stderr, "ran sanity check ping\n");
+
+   client = mongoc_client_new ("mongodb://localhost");
+   BSON_ASSERT (client);
+
+   // Test an IPv4 address
+   ASSERT_OR_PRINT (mongoc_client_set_bind_ip (client, "127.0.0.1", &error), error);
+   //ASSERT_OR_PRINT (mongoc_client_set_bind_ip (client, "192.168.1.14", &error), error);
```
samantharitter, line 2218: This is my machine's current IP address, which I used for some manual testing.  TBD on a way to programmatically use an IP besides "127.0.0.1" on evergreen hosts.

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -3792,6 +3835,31 @@ test_ssl_opts_padding_not_null (void)
 }
 #endif
 
+/* static void */
+/* test_mongoc_bind_ip (void) */
```
samantharitter, line 3839: I will remove this before committing, it's a test that works with a simple test server I wrote to inspect incoming connections.

src/libmongoc/tests/test-mongoc-interrupt.c
```diff
@@ -106,7 +106,7 @@ test_interrupt (void)
    mock_server_run (server);
    uri = mock_server_get_uri (server);
    stream =
-      mongoc_client_connect_tcp (10000, mongoc_uri_get_hosts (uri), &error);
+      mongoc_client_connect_tcp (NULL, 10000, mongoc_uri_get_hosts (uri), &error);
```
samantharitter, line 109: I don't think we need to support a custom `bind_ip` for this class.

src/libmongoc/src/mongoc/mongoc-socket.c
kevinAlbs: Oh interesting, thanks for linking to that article.

> It is probably not strictly necessary to use SO_REUSEADDR, but I think it's a little more resilient to use it. 

I agree. From what I understand, closed sockets can be in the `TIME_WAIT` state for quite some time before they are eligible for reuse if `SO_REUSEADDR` is not set. It seems like 15 seconds on macOS `sysctl net.inet.tcp.msl`

If `SO_REUSEADDR` is set, it might be rare a user hits that limit. Though you make a good point that they may be creating sockets outside of the C driver. Even if binding to an IP, and using `SO_REUSEADDR` gives you 64k unique sockets to a destination IP, another application may have exhausted many of those sockets outside of the use of the C driver.

Waiting until `expire_at` seems reasonable. I am not sure how we could avoid spinning, if all sockets are exhausted, for, say 5 seconds, I'm not sure if it would loop at full CPU usage.

What if we don't retry? We don't know how this will be used. We can investigate more complex retry logic if we discover there is a need for it.

src/libmongoc/tests/test-mongoc-client.c
kevinAlbs: If it makes sense, we could determine the IP of the interface to bind to with bash, and pass it in as an environment variable.

src/libmongoc/tests/test-mongoc-interrupt.c
kevinAlbs: Agreed, mongoc_interrupt can be removed entirely.

src/libmongoc/src/mongoc/mongoc-socket.c
samantharitter: I think that if we are going to set `SO_REUSEADDR`, we should retry at least once.  Because we are forcing the OS to pick an ephemeral socket for us before it knows which remote host we are connecting to, `SO_REUSEADDR` makes it possible that the OS could give us a port number we are already using to talk to that host.  This would make our { bind_ip, bind_port, connect_ip, connect_port } tuple a repeat, and `connect()` will fail.  As far as I understand it, ending up in this scenario isn't the same situation as running out of sockets, I think the OS could return a repeat ephemeral port number at any time if we are using `SO_REUSEADDR` (but this could also vary by OS, not entirely sure).

If we drop the retry logic, I think we should just not use `SO_REUSEADDR`.  Since it seems rare that a user would get into a situation where `SO_REUSEADDR` was really necessary, I would be fine with removing that logic if you agree!

kevinAlbs: Yeah I am fine with removing it. Users may not need it, and it is not a requirement as far as we know. We can start with the more straightforward implementation without retrying, and add complexity later if we discover a need.

PR #736, opened 2021-02-11
CDRIVER-3904 move log handler above init in logging example

bazile-clyde:
This PR moves `mongoc_log_set_handler` above `mongoc_init` in our logging docs example. This clarifies the log handler can be set at any time, even before a call to `mongoc_init`.

### DIFF
--- a/src/libmongoc/doc/logging.rst
+++ b/src/libmongoc/doc/logging.rst
@@ -91,8 +91,8 @@ For example, you could register a custom handler to suppress messages at INFO le
   int
   main (int argc, char *argv[])
   {
-     mongoc_init ();
      mongoc_log_set_handler (my_logger, NULL);
+     mongoc_init ();
 
      /* ... your code ...  */
 


src/libmongoc/doc/logging.rst
```diff
@@ -91,8 +91,8 @@ For example, you could register a custom handler to suppress messages at INFO le
   int
   main (int argc, char *argv[])
   {
-     mongoc_init ();
      mongoc_log_set_handler (my_logger, NULL);
+     mongoc_init ();
```
kevinAlbs, line 95: Looking through `mongoc_log_set_handler` I agree with your assessment. It looks like `mongoc_log_set_handler` is safe to call before `mongoc_init` since it has its own `bson_once` initialization.

kevinAlbs (approved):
Thanks for investigating this!

src/libmongoc/doc/logging.rst
fgalan: In addition to the change in logging.rst (which is great!) maybe http://mongoc.org/libmongoc/current/mongoc_init.html should also modified, as the following statement:

> Initialize the MongoDB C Driver by calling mongoc_init() exactly once at the beginning of your program

Is a bit mislead and could be improved by adding something like this:

> Initialize the MongoDB C Driver by calling mongoc_init() exactly once at the beginning of your program. Exception to this is mongoc_log_set_handler, which should be called *before* mongoc_init() or some log traces would not use your log handling function. See [link to logging document] for details.

I can do a pull request with the modification if you want.

bazile-clyde: That change looks good. If you wouldn't mind submitting a pull request that'd be great!

fgalan: Of course! This is the PR: https://github.com/mongodb/mongo-c-driver/pull/739/

PR #737, opened 2021-02-12
CDRIVER-3821 implement versioned server api

samantharitter:


src/libmongoc/doc/mongoc_server_api_version_from_string.rst
```diff
@@ -0,0 +1,26 @@
...
+  bool
+  mongoc_server_api_version_from_string (const char *version,
+                                         mongoc_server_api_version_t *out);
+
+Given a string ``version``, populates ``out`` with the equivalent :symbol:`mongoc_server_api_version_t` if ``version`` represents a valid API version.
+
+Paremeters
+----------
+
+* ``version``: A string representing the version number.
```
alcaeus, line 20: I would avoid using the term "version number" here and in other documentation. The version is not required to be a number, it just so happens that the first one is "1". There may be other versions in the future that are not identifier by a number.

src/libmongoc/src/mongoc/mongoc-server-api.c
```diff
@@ -0,0 +1,100 @@
...
+#include "mongoc-server-api-private.h"
+
+const char *
+mongoc_server_api_version_to_string (mongoc_server_api_version_t version)
+{
+   const char *out;
+
+   /* Until version numbers top 1000, this should be fine. After that point,
+      this method will need to be altered to hold memory for larger strings. */
+   bson_uint32_to_string (version, &out, NULL, 0);
```
alcaeus, line 28: This assumes that API versions will always be numeric. I think it would make more sense to use a `switch` here to ensure we're not implicitly restricting future API versions.

src/libmongoc/doc/mongoc_server_api_version_from_string.rst
samantharitter: Oh that's really interesting!  So maybe we'll start naming MDB versions after desserts??  Ha.  I will change this to be "version identifier" instead of "version number".

src/libmongoc/src/mongoc/mongoc-server-api.c
samantharitter: Done.

src/libmongoc/doc/mongoc_server_api_version_from_string.rst
alcaeus: Desserts are already taken by Android, but we could always do something else, like cocktails for example 🍹

samantharitter: `MONGOC_VERSION_DARK_AND_STORMY`

src/libmongoc/doc/mongoc_server_api_copy.rst
```diff
@@ -0,0 +1,24 @@
...
+mongoc_server_api_copy()
+========================
+
+Synopsis
+--------
+
+.. code-block:: c
+
+  mongoc_server_api_t *
+  mongoc_server_api_copy (mongoc_server_api_t *api);
```
kevinAlbs, line 12:
```suggestion
  mongoc_server_api_copy (const mongoc_server_api_t *api);
```

src/libmongoc/src/mongoc/mongoc-server-api.c
```diff
@@ -0,0 +1,98 @@
...
+{
+   if (version == MONGOC_SERVER_API_V1) {
+      return "1";
+   } else {
+      return NULL;
+   }
+}
+
+bool
+mongoc_server_api_version_from_string (const char *version,
```
kevinAlbs, line 32: Should this be public? I think the spec only requires a way to pass a string as the API version if an enum is not possible.

src/libmongoc/doc/mongoc_server_api_version_t.rst
```diff
@@ -0,0 +1,34 @@
...
+Supported API Versions
+----------------------
+
+The driver currently supports the following MongoDB API versions:
+
+====================  ======================
+Enum value            MongoDB version number
+====================  ======================
+MONGOC_SERVER_API_V1  1
+====================  ======================
```
alcaeus, line 22: Referring to version string instead of version number and quoting the value would help convey that versions are not limited to numbers but can be any arbitrary string (e.g. "luscious-lollipop"):

```suggestion
====================  ======================
Enum value            MongoDB version string
====================  ======================
MONGOC_SERVER_API_V1  "1"
====================  ======================
```

src/libmongoc/tests/bsonutil/bson-val.c
```diff
@@ -45,9 +47,16 @@ bson_val_from_value (const bson_value_t *value)
    bson_init (&tmp);
    bson_append_value (&tmp, "v", 1, value);
    as_string = bson_as_canonical_extended_json (&tmp, NULL);
-   /* This produces: { "v" : {...} }. Strip off the wrapping "v" and braces. */
-   val->as_string = bson_strndup (as_string + 7, strlen (as_string) - 9);
-   bson_free (as_string);
+
+   if (!as_string) {
```
kevinAlbs, line 51: Interesting, did this occur in tests? I think `bson_as_canonical_extended_json` only returns `NULL` if the document is malformed.

src/libmongoc/src/mongoc/mongoc-cmd.c
```diff
@@ -785,6 +785,29 @@ _is_retryable_read (const mongoc_cmd_parts_t *parts,
...
+   if (_mongoc_client_session_txn_in_progress (cs)) {
+      return true;
+   }
+
+   /* commitTransaction counts, too. */
+   if (parts->assembled.is_txn_finish) {
+      return true;
+   }
+
+   return false;
```
alcaeus, line 807: I feel a ternary return helps understand the logic better after the early return above:

```suggestion
   cs = parts->prohibit_lsid ? NULL : parts->assembled.session;
   if (!cs) {
      return false;
   }

   return _mongoc_client_session_txn_in_progress (cs)
      /* commitTransaction counts as in progress, too. */
      || parts->assembled.is_txn_finish);
```

src/libmongoc/src/mongoc/mongoc-cmd.c
```diff
@@ -811,12 +834,14 @@ mongoc_cmd_parts_assemble (mongoc_cmd_parts_t *parts,
                            bson_error_t *error)
 {
    mongoc_server_description_type_t server_type;
+   mongoc_server_api_t *api;
```
alcaeus, line 837: This is totally unrelated, but is there a reason to not declare `*api` scoped in the condition below? Is this a matter of preference or is this not supported in some C standards?

src/libmongoc/src/mongoc/mongoc-server-api.c
```diff
@@ -0,0 +1,98 @@
...
+#include "mongoc-server-api-private.h"
+
+const char *
+mongoc_server_api_version_to_string (mongoc_server_api_version_t version)
+{
+   if (version == MONGOC_SERVER_API_V1) {
+      return "1";
+   } else {
+      return NULL;
+   }
```
alcaeus, line 28: Using a `switch` here would ensure that all `enum` values are covered:

```suggestion
   switch (version) {
      case MONGOC_SERVER_API_V1:
         return "1";
   }
   
   return NULL;
```

This also fixes the nit of not liking `return` nested in `else` as it may just be folded out of the conditional.

src/libmongoc/src/mongoc/mongoc-server-api.c
```diff
@@ -0,0 +1,98 @@
...
+bool
+mongoc_server_api_version_from_string (const char *version,
+                                       mongoc_server_api_version_t *out)
+{
+   if (strcmp (version, "1") == 0) {
+      *out = MONGOC_SERVER_API_V1;
+      return true;
+   } else {
+      return false;
+   }
```
alcaeus, line 40: Nit: consider not nesting `return` within `else`:

```suggestion

   return false;
```

src/libmongoc/src/mongoc/mongoc-server-api.h
```diff
@@ -0,0 +1,57 @@
...
+#ifndef MONGOC_SERVER_API_H
+#define MONGOC_SERVER_API_H
+
+#include <bson/bson.h>
+
+#include "mongoc-macros.h"
+
+BSON_BEGIN_DECLS
+
+typedef enum { MONGOC_SERVER_API_V1 = 1 } mongoc_server_api_version_t;
```
alcaeus, line 28: Do we manually need to assign values to the enum? IMO, this could sow confusion whether versions are strings or numbers.

src/libmongoc/src/mongoc/mongoc-cmd.c
```diff
@@ -785,6 +785,29 @@ _is_retryable_read (const mongoc_cmd_parts_t *parts,
...
+   if (!cs) {
+      return false;
+   }
+
+   if (_mongoc_client_session_txn_in_progress (cs)) {
+      return true;
+   }
+
+   /* commitTransaction counts, too. */
+   if (parts->assembled.is_txn_finish) {
```
kevinAlbs, line 803: `is_txn_finish` is also true for `abortTransaction`. I think that's right. It might not be exercised in the transaction-handing.json test.

cc @alcaeus Can you confirm `abortTransaction` should skip adding server API options?

src/libmongoc/tests/test-mongoc-versioned-api.c
```diff
@@ -0,0 +1,150 @@
...
+                    error);
+
+   /* Cannot change server API once it is set */
+   ASSERT (!mongoc_client_pool_set_server_api (pool, api, &error));
+   ASSERT_ERROR_CONTAINS (error,
+                          MONGOC_ERROR_POOL,
+                          MONGOC_ERROR_POOL_API_ALREADY_SET,
+                          "Cannot set server api more than once");
+
+   /* Clients popped from pool have matching API */
```
kevinAlbs, line 121: Nice tests! Checking for clients getting the server API from the parent pool is a good idea.

src/libmongoc/tests/unified/runner.c
```diff
@@ -359,6 +358,7 @@ test_file_new (test_runner_t *test_runner, bson_t *bson)
    bson_parser_array_optional (
       parser, "createEntities", &test_file->create_entities);
    bson_parser_array_optional (parser, "initialData", &test_file->initial_data);
+   bson_parser_doc_optional (parser, "_yamlAnchors", &test_file->yaml_anchors);
```
alcaeus, line 361: Parsing this is not necessary as per the unified test format spec:
> This is only used to define anchors within the YAML files and MUST NOT be used by test runners.

alcaeus (changes requested):
I've left some suggestions in the comments below, feel free to apply them through the UI for convenience. Changes look good, but there are a few points below. Of these, only the first should be considered for this PR, the others can be relegated to follow-up tickets.

##### `runOnRequirements.serverParameters`

This PR omits the new `serverParameters` field in `runOnRequirements`. From the spec:
> `serverParameters`: Optional object of server parameters to check against. To check server parameters, drivers send a `{ getParameter: 1, <parameter>: 1 }` command to the server using the internal MongoClient. Drivers MAY also choose to send a `{ getParameter: '*' }` command and fetch all parameters at once. The result SHOULD be cached to avoid repeated calls to fetch the same parameter. Test runners MUST apply the rules specified in Flexible Numeric Comparisons when comparing values. If a server does not support a parameter, test runners MUST treat the comparison as not equal and skip the test. This includes errors that occur when fetching a single parameter using getParameter.
Not implementing this portion will cause overhead once we start testing with `requiredApiVersion`. See my next point for details.

##### Testing with required API version

The spec calls for adding evergreen configuration to test against servers with the `requiredApiVersion` parameter set: https://github.com/mongodb/specifications/blob/master/source/versioned-api/tests/README.rst#testing-with-required-api-version
Since I was going to do this for PHP in the coming days, I'm ok with deferring on those tests to get the feature merged and create a PR here while I do PHP. mongo-orchestration provides the necessary tools, and integrating this into the runner should be straightforward. The tests require the `serverParameters` requirement logic explained above, so having this would make this process a lot easier and more straightforward for me.

##### Exposing `mongoc_server_api_version_t` internals

During the implementation of the POC for PHP, I noticed that there are no getters for the values in `mongoc_server_api_version_t`. In PHP we expose a `MongoDB\Driver\ServerApi` class for people to use and need to be able to represent it:
```
object(MongoDB\Driver\ServerApi)#%d (%d) {
   ["version"]=>
   string(1) "1"
   ["strict"]=>
   NULL
   ["deprecationErrors"]=>
   NULL
 }
```
Getters are convenient as it makes the object in PHP more lightweight. As it is, we need to store the values given when creating the `ServerApi` instance even though they are only needed to construct the `mongoc_server_api_t`. Was there a reason to omit the getters? No need to add them in this PR if it wasn't, I can add them when wrapping up work on PHP.

kevinAlbs (commented):
Looking good!

src/libmongoc/tests/bsonutil/bson-val.c
samantharitter: Yes, it showed up as a seg fault in testing.  I guess we were running a malformed document through!  Returning `val` without a string still caused a test failure, but one that was much easier to debug.  I can't remember which test was causing this now, though.

src/libmongoc/tests/unified/runner.c
samantharitter: With the way that the `bson_parser` works, it is necessary to add this field to the `parser` here to avoid it complaining.  We still aren't using `_yamlAnchors` anywhere, though.

src/libmongoc/src/mongoc/mongoc-server-api.h
samantharitter: No, I suppose we don't need to do this, now that I've changed the to string logic to not rely on the numeric value of the enum.  Will change.

src/libmongoc/tests/test-mongoc-versioned-api.c
samantharitter: Thanks!

src/libmongoc/src/mongoc/mongoc-server-api.c
samantharitter: Actually, I don't think the spec requires either of these helpers to be exposed.  It does seem like something a user might be interested in, but I'm also happy to make these private.  I do think it would be odd to provide an `enum -> string` method without the parallel `string -> enum` method, so we should either expose both or hide both IMO.

src/libmongoc/src/mongoc/mongoc-cmd.c
samantharitter: It's a matter of preference for me, and is the style followed in this file. I like for my code additions to be consistent with existing styles in the surrounding code whenever it is reasonable to do so.  As far as I know C99 would allow the variable declaration at the beginning of any scope, though.

src/libmongoc/src/mongoc/mongoc-cmd.c
samantharitter: Eh, I usually like to keep multi-step logic in distinct steps for readability, but I don't feel strongly about it in this situation and I am happy to use your suggestion instead.

samantharitter:
Hi @alcaeus, thank you for your feedback!  Let me address your questions one by one.

> This PR omits the new `serverParameters` field in `runOnRequirements`.

We are actually already parsing and validating the `serverParameters` field, I believe @kevinAlbs added this when he implemented the new unified test runner: https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/tests/unified/runner.c#L606-L625

> The spec calls for adding evergreen configuration to test against servers with the `requiredApiVersion` parameter set: https://github.com/mongodb/specifications/blob/master/source/versioned-api/tests/README.rst#testing-with-required-api-version

Yes, let's do this in a follow-up ticket.  I've opened [CDRIVER-3917](https://jira.mongodb.org/browse/CDRIVER-3917) for this work.

> Was there a reason to omit the getters?

Since `deprecationErrors` and `strict` are optional bool parameters, exposing them is non-trivial, so I decided not to add the getters unless they were deemed useful.  It would be nice if we offered a `mongoc_optional_t` type that could encapsulate such values, rather than adding two separate `get_strict()` and `get_strict_set()` methods.  I'm happy to talk about a design we like for this and do it in a follow-up ticket, or let you add getters later.

alcaeus:
mongoc_optional_t sounds good to me. Happy to defer that to a followup ticket

src/libmongoc/src/mongoc/mongoc-cmd.c
alcaeus: Yes, `abortTransaction` should not include versioned API options, as it is still bound to the API version set when the transaction was started.

src/libmongoc/src/mongoc/mongoc-server-api.c
alcaeus: PHP makes use of this: https://github.com/mongodb/mongo-php-driver/pull/1204/files#diff-4ddc908579a884b62260840125c4a168a175022f1a5f764db111cd22a0529899R35. Not having this helper exposed would require us to duplicate that logic in PHP to be able to create a `mongoc_server_api_version_t`.

src/libmongoc/src/mongoc/mongoc-cmd.c
alcaeus: I see, thanks for explaining!

src/libmongoc/src/mongoc/mongoc-cmd.c
samantharitter: Great!  I'll just change the comment, then.

src/libmongoc/src/mongoc/mongoc-server-api.c
samantharitter: Ok, then I will leave these as-is.

alcaeus (approved):

kevinAlbs (approved):
LGTM!

samantharitter:
I've manually resolved the conflicts here and pushed this work into master.  Closing this PR.

PR #740, opened 2021-02-23
Improve docs of cursor returning functions

fgalan:
Comparing the "Returns" section in src/libmongoc/doc/mongoc_collection_aggreage.rst with the one in src/libmongoc/doc/mongoc_collection_find.rst and src/libmongoc/doc/mongoc_collection_find_with_options.rst it seems the later ones are missing important information about how errors are reported (related with https://stackoverflow.com/questions/66027858/how-to-get-errors-when-calling-mongoc-collection-find-function).

I'd like to have feedback on this pull request, please. If it is right, please merge to improve the documentation. In addition, if you know any other document (others operations returning mongo_cursor_t*, probably) that may need a similar improvement, I can include it in this PR before merging.

Thanks!

### DIFF
--- a/src/libmongoc/doc/includes/returns-cursor.txt
+++ b/src/libmongoc/doc/includes/returns-cursor.txt
@@ -1,0 +1,8 @@
+This function returns a newly allocated :symbol:`mongoc_cursor_t` that should be freed with :symbol:`mongoc_cursor_destroy()` when no longer in use. The returned :symbol:`mongoc_cursor_t` is never ``NULL``; if the parameters are invalid, the :symbol:`bson:bson_error_t` in the :symbol:`mongoc_cursor_t` is filled out, and the :symbol:`mongoc_cursor_t` is returned before the server is selected. The user must call :symbol:`mongoc_cursor_next()` on the returned :symbol:`mongoc_cursor_t` to execute the aggregation pipeline.
+
+Cursor errors can be checked with :symbol:`mongoc_cursor_error_document`. It always fills out the :symbol:`bson_error_t` if an error occurred, and optionally includes a server reply document if the error occurred server-side.
+
+.. warning::
+
+  Failure to handle the result of this function is a programming error.
+
--- a/src/libmongoc/doc/includes/returns-cursor.txt
+++ b/src/libmongoc/doc/includes/returns-cursor.txt
@@ -1,8 +1,7 @@
-This function returns a newly allocated :symbol:`mongoc_cursor_t` that should be freed with :symbol:`mongoc_cursor_destroy()` when no longer in use. The returned :symbol:`mongoc_cursor_t` is never ``NULL``; if the parameters are invalid, the :symbol:`bson:bson_error_t` in the :symbol:`mongoc_cursor_t` is filled out, and the :symbol:`mongoc_cursor_t` is returned before the server is selected. The user must call :symbol:`mongoc_cursor_next()` on the returned :symbol:`mongoc_cursor_t` to execute the aggregation pipeline.
+This function returns a newly allocated :symbol:`mongoc_cursor_t` that should be freed with :symbol:`mongoc_cursor_destroy()` when no longer in use. The returned :symbol:`mongoc_cursor_t` is never ``NULL``, even on error. The user must call :symbol:`mongoc_cursor_next()` on the returned :symbol:`mongoc_cursor_t` to execute the initial command.
 
 Cursor errors can be checked with :symbol:`mongoc_cursor_error_document`. It always fills out the :symbol:`bson_error_t` if an error occurred, and optionally includes a server reply document if the error occurred server-side.
 
 .. warning::
 
   Failure to handle the result of this function is a programming error.
-
--- a/src/libmongoc/doc/mongoc_collection_find.rst
+++ b/src/libmongoc/doc/mongoc_collection_find.rst
@@ -51,7 +51,12 @@ If no options are necessary, ``query`` can simply contain a query such as ``{a:1
 Returns
 -------
 
-A newly allocated :symbol:`mongoc_cursor_t` that should be freed with :symbol:`mongoc_cursor_destroy()` when no longer in use.
+This function returns a newly allocated :symbol:`mongoc_cursor_t` that should be freed with :symbol:`mongoc_cursor_destroy()` when no longer in use. The returned :symbol:`mongoc_cursor_t` is never ``NULL``; if the parameters are invalid, the :symbol:`bson:bson_error_t` in the :symbol:`mongoc_cursor_t` is filled out, and the :symbol:`mongoc_cursor_t` is returned before the server is selected. The user must call :symbol:`mongoc_cursor_next()` on the returned :symbol:`mongoc_cursor_t` to execute the aggregation pipeline.
+
+.. warning::
+
+  Failure to handle the result of this function is a programming error.
+
 
 Example
 -------
--- a/src/libmongoc/doc/mongoc_collection_find.rst
+++ b/src/libmongoc/doc/mongoc_collection_find.rst
@@ -51,7 +51,7 @@ If no options are necessary, ``query`` can simply contain a query such as ``{a:1
 Returns
 -------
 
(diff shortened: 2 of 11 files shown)


kevinAlbs (commented):
Thank you for the contribution @fgalan!

The error handling for a `mongoc_cursor_t` should apply to any function that returns a `mongoc_cursor_t`. There is documentation describing error handling of cursors on: http://mongoc.org/libmongoc/current/cursors.html But a blurb, like the one you've written, on each function that returns a `mongoc_cursor_t` seems generally helpful to users.

If you are feeling ambitious, I'd suggest putting this blurb into a reusable include within: `src/libmongoc/doc/includes` and including that. There are many functions that return a `mongoc_cursor_t`:

- mongoc_database_aggregate
- mongoc_database_command
- mongoc_database_find_collections
- mongoc_database_find_collections_with_opts
- mongoc_collection_find_with_opts
- mongoc_collection_aggregate
- mongoc_collection_command
- mongoc_collection_find_indexes
- mongoc_collection_find_indexes_with_opts
- mongoc_collection_find

We'd be happy to pick it up from here as well.

fgalan:
@kevinAlbs thanks for the feedback!

I'd have a look and try to implement your proposals.

src/libmongoc/doc/mongoc_collection_find_indexes.rst
```diff
@@ -37,4 +37,9 @@ Errors are propagated via the ``error`` parameter.
 Returns
 -------
 
-A cursor where each result corresponds to the server's representation of an index on this collection. If the collection does not exist on the server, the cursor will be empty.
+.. include:: includes/returns-cursor.txt
+
+In the returned cursor each result corresponds to the server's representation of an index on this collection. If the collection does not exist on the server, the cursor will be empty.
+
+The cursor functions :symbol:`mongoc_cursor_set_limit`, :symbol:`mongoc_cursor_set_batch_size`, and :symbol:`mongoc_cursor_set_max_await_time_ms` have no use on the returned cursor.
```
fgalan, line 44: I see a similar paragraph in mongoc_collection_find_indexes_with_opts.rst and I have included it under the understanding that both functions behave the same.

However, if I'm wrong, just tell me and I'll remove the paragraph.

src/libmongoc/doc/mongoc_database_find_collections.rst
```diff
@@ -43,5 +43,9 @@ Errors are propagated via the ``error`` parameter.
 Returns
 -------
 
-A cursor where each result corresponds to the server's representation of a collection in this database.
+.. include:: includes/returns-cursor.txt
+
+In the returned cursor each result corresponds to the server's representation of a collection in this database.
+
+The cursor functions :symbol:`mongoc_cursor_set_limit`, :symbol:`mongoc_cursor_set_batch_size`, and :symbol:`mongoc_cursor_set_max_await_time_ms` have no use on the returned cursor.
```
fgalan, line 50: I see a similar paragraph in src/libmongoc/doc/mongoc_database_find_collections_with_opts.rst and I have included it under the understanding that both functions behave the same way.

However, if I'm wrong, just tell me and I'll remove the paragraph.

src/libmongoc/doc/mongoc_collection_find.rst
```diff
@@ -51,7 +51,12 @@ If no options are necessary, ``query`` can simply contain a query such as ``{a:1
 Returns
 -------
 
-A newly allocated :symbol:`mongoc_cursor_t` that should be freed with :symbol:`mongoc_cursor_destroy()` when no longer in use.
+This function returns a newly allocated :symbol:`mongoc_cursor_t` that should be freed with :symbol:`mongoc_cursor_destroy()` when no longer in use. The returned :symbol:`mongoc_cursor_t` is never ``NULL``; if the parameters are invalid, the :symbol:`bson:bson_error_t` in the :symbol:`mongoc_cursor_t` is filled out, and the :symbol:`mongoc_cursor_t` is returned before the server is selected. The user must call :symbol:`mongoc_cursor_next()` on the returned :symbol:`mongoc_cursor_t` to execute the query.
```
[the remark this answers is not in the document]
fgalan: Done in commit b14f277

I have used part of your comment in the note. If you prefer a different working, a comment including a [github suggestion](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/incorporating-feedback-in-your-pull-request) will be great.

fgalan:
@kevinAlbs I have updated the PR based on your feedback, creating an include file and using it in the 10 functions you mention. It would be ready for a next review.

Thanks!

src/libmongoc/doc/includes/returns-cursor.txt
```diff
@@ -0,0 +1,8 @@
+This function returns a newly allocated :symbol:`mongoc_cursor_t` that should be freed with :symbol:`mongoc_cursor_destroy()` when no longer in use. The returned :symbol:`mongoc_cursor_t` is never ``NULL``; if the parameters are invalid, the :symbol:`bson:bson_error_t` in the :symbol:`mongoc_cursor_t` is filled out, and the :symbol:`mongoc_cursor_t` is returned before the server is selected. The user must call :symbol:`mongoc_cursor_next()` on the returned :symbol:`mongoc_cursor_t` to execute the aggregation pipeline.
```
kevinAlbs, line 1:
```suggestion
This function returns a newly allocated :symbol:`mongoc_cursor_t` that should be freed with :symbol:`mongoc_cursor_destroy()` when no longer in use. The returned :symbol:`mongoc_cursor_t` is never ``NULL``, even on error. The user must call :symbol:`mongoc_cursor_next()` on the returned :symbol:`mongoc_cursor_t` to execute the initial command.
```
I realize this came from existing documentation :) but I think it is best to avoid mentioning the `bson_error_t` inside the `mongoc_cursor_t`. `mongoc_cursor_t` is an opaque type, so it seems preferable to just note that an error could be set on the returned cursor (before the first call to `mongoc_cursor_next`.

Cursor returning functions like `mongoc_collection_find_with_opts` would run a find command initially instead of an aggregate command, so I think it is preferable to be generic.

src/libmongoc/doc/mongoc_collection_find_indexes.rst
kevinAlbs: Good catch! This is an improvement. `mongoc_collection_find_indexes` directly calls `mongoc_collection_find_indexes_with_opts`, so the same behavior applies to the returned cursor.

src/libmongoc/doc/mongoc_database_find_collections.rst
kevinAlbs: Adding this paragraph is a good idea. This internally calls `mongoc_database_find_collections_with_opts`. The only difference is that this returns `NULL` if an error occurred.

src/libmongoc/doc/mongoc_database_find_collections.rst
```diff
@@ -43,5 +43,9 @@ Errors are propagated via the ``error`` parameter.
 Returns
 -------
 
-A cursor where each result corresponds to the server's representation of a collection in this database.
+.. include:: includes/returns-cursor.txt
```
kevinAlbs, line 46: This may be a case to not use the include directly.

Part of the deprecated behavior of this function is that it returns `NULL` if an error occurred:
https://github.com/mongodb/mongo-c-driver/blob/6bacf922669ab780fa5fb37e9c561b8709e27e29/src/libmongoc/src/mongoc/mongoc-database.c#L758:L758

Instead of using the `returns-cursor.txt` directly, this might be a special case, and to paste everything except for the `The returned mongoc_cursor_t is never NULL, even on error.` sentence:


```suggestion
This function returns a newly allocated :symbol:`mongoc_cursor_t` that should be freed with :symbol:`mongoc_cursor_destroy()` when no longer in use. The user must call :symbol:`mongoc_cursor_next()` on the returned :symbol:`mongoc_cursor_t` to execute the initial command.
```

kevinAlbs (changes requested):
Thank you for updating the PR! I added a few suggestions, and have run our Evergreen CI attached to this PR, which builds the documentation as part of the tests.

src/libmongoc/doc/mongoc_database_find_collections.rst
fgalan: Maybe documentation should make explicit that in this case `NULL` can be returned in case of error? I mean something like this:

```suggestion
This function returns a newly allocated :symbol:`mongoc_cursor_t` that should be freed with :symbol:`mongoc_cursor_destroy()` when no longer in use, or `NULL` in case of error. The user must call :symbol:`mongoc_cursor_next()` on the returned :symbol:`mongoc_cursor_t` to execute the initial command.
```

What do you think?

fgalan:
Is there anything pending on my side in this PR to be merged, please? Thanks!

kevinAlbs (approved):
LGTM!

kevinAlbs:
Apologies for the delay @fgalan. Thank you again for the contribution!

PR #741, opened 2021-03-05
CDRIVER-3856 skip CSE listCollections cmds

bazile-clyde:
The first commit syncs the spec tests. The second is the actual changes.

### DIFF
--- a/src/libmongoc/tests/json-test-monitoring.c
+++ b/src/libmongoc/tests/json-test-monitoring.c
@@ -406,8 +406,8 @@ apm_match_visitor (match_ctx_t *ctx,
       }
    } else if (strstr (ctx->path, "updates.")) {
       /* tests expect "multi: false" and "upsert: false" explicitly;
-      * we don't send them. fix when path is like "updates.0", "updates.1", ...
-      */
+       * we don't send them. fix when path is like "updates.0", "updates.1", ...
+       */
 
       if (!strcmp (key, "multi") && !bson_iter_bool (pattern_iter)) {
          return MATCH_ACTION_SKIP;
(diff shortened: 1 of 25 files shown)


src/libmongoc/tests/json-test-monitoring.c
```diff
@@ -442,6 +442,25 @@ _apm_match_error_context (const bson_t *actual, const bson_t *expectations)
    bson_free (actual_str);
    bson_free (expectations_str);
 }
+
+bool skip_cse_list_collections(const bson_t *doc) {
+   // see CDRIVER-3856: Sharing a MongoClient for metadata lookup can lead to
+   // deadlock in drivers using automatic encryption
```
kevinAlbs, line 448: Use `/* */` comments for c89.

Can you note in this comment that the C driver explicitly does not use a separate `mongoc_client_t` for listCollections / find on the key vault, which is why we skip checks for them?

src/libmongoc/tests/json-test-monitoring.c
```diff
@@ -442,6 +442,25 @@ _apm_match_error_context (const bson_t *actual, const bson_t *expectations)
...
+   const char* val;
+
+   if (!bson_has_field(doc, "command_started_event.command.listCollections"))
+      return false;
```
kevinAlbs, line 452: If this returns true if we should skip the check, I would think this should be:

```
if (bson_has_field(doc, "command_started_event.command.listCollections"))
    return true;

if (bson_has_field(doc, "command_started_event.command.$db") && 0 == strcmp (bson_lookup_utf8(doc, "command_started_event.command.$db"), "keyvault"))
    return true;

return false;

Comment (bazile-clyde) [reply src/libmongoc/tests/json-test-monitoring.c:452]: I thought the logic here was to skip _only_ the `listCollections` commands on the "keyvault" database. As you wrote it, this function would skip all `listCollections` commands _and_ any commands on the "keyvault" database. I updated this function to do that but just wanted to double-check that was the intent.

File: src/libmongoc/tests/json-test-monitoring.c (hunk 1 above)
Comment (bazile-clyde) [reply src/libmongoc/tests/json-test-monitoring.c]: Done

File: src/libmongoc/tests/json-test-monitoring.c [hunk 3]
```diff
@@ -520,16 +543,20 @@ check_json_apm_events (json_test_ctx_t *ctx, const bson_t *expectations)
...
-            test_error ("could not match APM event\n"
-                        "\texpected: %s\n\n"
-                        "\tactual  : %s\n\n"
-                        "\terror   : %s\n\n",
-                        bson_as_canonical_extended_json (&expectation, NULL),
-                        bson_as_canonical_extended_json (&actual, NULL),
-                        match_ctx.errmsg);
          }
+
+         if (skip_cse_list_collections (&actual)) {
```
kevinAlbs, line 548: This skip logic is specific to CSFLE tests, but `check_json_apm_events` applies to all spec tests that exercise this code, (CRUD, transactions).

There's a very small risk a non CSFLE test could skip an event if there is a legitimate `listCollections` against a `keyvault` database in a spec test.

I'm not too concerned though. It seems very unlikely. And it would result in a test failure (not a loss of test coverage). If it becomes an issue, we can make this configurable later.

kevinAlbs (approved):
LGTM with undoing the requested change to `skip_cse_list_collections` (sorry for the initial incorrect comment).

bazile-clyde:
> LGTM with undoing the requested change to `skip_cse_list_collections` (sorry for the initial incorrect comment).

Got it. Reverted the commit. I'm waiting until the test pass and pushing.

PR #743, opened 2021-03-08
CDRIVER-3272 Correctly document bson gettimeofday

benjirewis:
[CDRIVER-3272](https://jira.mongodb.org/browse/CDRIVER-3272)

Document `bson_gettimeofday()` correctly. The docs previously listed an incorrect, additional `timezone` param that does not actually exist in [bson-clock.h](https://github.com/mongodb/mongo-c-driver/blob/master/src/libbson/src/bson/bson-clock.h#L35).


bazile-clyde (approved):
LGTM! Feel free to merge.

PR #745, opened 2021-03-09
Don't inspect error messages for "not master" or "node is recovering"

alcaeus:
Covers CDRIVER-3885 and CDRIVER-3884.


src/libmongoc/src/mongoc/mongoc-error.c
```diff
@@ -268,9 +270,13 @@ _mongoc_error_is_not_master (bson_error_t *error)
    switch (error->code) {
    case MONGOC_SERVER_ERR_NOTMASTER:
    case MONGOC_SERVER_ERR_NOTMASTERNOSLAVEOK:
+   case MONGOC_SERVER_ERR_LEGACYNOTPRIMARY:
       return true;
-   default:
+      /* All errors where no code was found are marked as MONGOC_ERROR_QUERY_FAILURE */
+   case MONGOC_ERROR_QUERY_FAILURE:
```
alcaeus, line 276: I dislike this part. `_mongoc_cmd_check_ok` and `_mongoc_cmd_check_ok_no_wce` rely on `_parse_error_reply` to extract message and code information from a `bson_t` reply. If no code is found in the reply document, a code of `0` is returned.

The two check methods then take this information and build a `bson_error_t` of it, but only after changing the error code to `MONGOC_ERROR_QUERY_FAILURE` (17) if it was 0, essentially defaulting to that error code. I've covered for this by checking for that error code here, but I feel like this is a dirty workaround. `_mongoc_cmd_check_ok_no_wce` and `_mongoc_cmd_check_ok` should not change the error code, but instead this logic should be moved to wherever the `MONGOC_ERROR_QUERY_FAILURE` is needed. I'm not very familiar with this part of libmongoc, so if you have any pointers to where this might require changes I'm happy to investigate.

src/libmongoc/src/mongoc/mongoc-change-stream.c
```diff
@@ -73,6 +73,7 @@ _is_resumable_error (mongoc_change_stream_t *stream, const bson_t *reply)
    case MONGOC_SERVER_ERR_STALECONFIG:
    case MONGOC_SERVER_ERR_RETRYCHANGESTREAM:
    case MONGOC_SERVER_ERR_FAILEDTOSATISFYREADPREFERENCE:
+   case MONGOC_SERVER_ERR_LEGACYNOTPRIMARY:
```
alcaeus, line 76: Not sure if this is mandated by the spec, but I added this error code to all places where we check for the other two errors that indicate that the server is not a primary.

alcaeus (commented):
@bazile-clyde @samantharitter: The changes are complete, but I'm not sure about two things that I've highlighted below. I'd appreciate your opinion on how to best proceed.

src/libmongoc/src/mongoc/mongoc-change-stream.c
alcaeus: On second thought, I've removed this as it does not seem intentional per the spec text (and probably does not apply since the error code was used pre-3.2)

src/libmongoc/src/mongoc/mongoc-error.c
alcaeus: Tracked in CDRIVER-3928.

bazile-clyde (changes requested):
looks good. Only a few comments.

bazile-clyde (approved):
LGTM!

PR #751, opened 2021-03-09
CDRIVER-3773 Support regular expressions with no options

benjirewis:
[CDRIVER-3773](https://jira.mongodb.org/browse/CDRIVER-3773)

python-bsonjs, which relies on libbson, requested the ability to load regexes without the "options" field through JSON into BSON. I see no language in the extended JSON [spec](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#regularexpression) forbidding this behavior, and it's unlikely an existing user would actively expect a "missing options" error, so it seems ok to add.

Removes the "missing options" errors in `_bson_json_read_append_regex` and adds two new tests to verify correct parsing and absence of error for legacy and canonical regexes without options.

### DIFF
--- a/src/libbson/src/bson/bson-json.c
+++ b/src/libbson/src/bson/bson-json.c
@@ -1389,18 +1389,9 @@ _bson_json_read_append_regex (bson_json_reader_t *reader,    /* IN */
                                     "Missing \"$regex\" after \"$options\"");
          return;
       }
-      if (!data->regex.has_options) {
-         _bson_json_read_set_error (reader,
-                                    "Missing \"$options\" after \"$regex\"");
-         return;
-      }
    } else if (!data->regex.has_pattern) {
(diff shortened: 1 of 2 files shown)


bazile-clyde (approved):
LGTM!

alcaeus:
@benjirewis this change breaks the bson-corpus test suite in PHP. There is a test that expects a parse error when presented a regular expression without options: https://github.com/mongodb/specifications/blob/master/source/bson-corpus/tests/top.json#L74..L77.

I'll skip the test in PHP, but wanted to give you a heads up that this is covered by the corpus.

benjirewis:
Apologies for overlooking that test. Looking into how important the change is for python-bsonjs. I'll remove this change no matter what since it's spec-test breaking. If we still want to make the change, I'll do it through a drivers ticket with an actual spec change instead.

PR #753, opened 2021-03-11
CDRIVER-3929 Include apiVersion in handshake commands

alcaeus:
CDRIVER-3929

This fixes an oversight in the versioned API implementation where the handshake did not send API Version information to the server. This required exposing the API Version to the topology scanner as that's where the actual commands are handled. I've decided to allow changing the API Version multiple times as this part of libmongoc is very much internal.

During testing I discovered a couple of edge cases around setting API Version on a pooled client:
* When no API Version has been set, a user can set it after clients have been popped. This causes an inconsistent state in the pool. Solution is to prevent changing the API Version when the pool size is greater than 0. I understand this can cause an inconsistency when the last client is destroyed where one could once again set the API Version. Let me know if we should also prevent that case.
* When fetching a client from a pool that had no server API set, the api was not locked against modifications, which could once again lead to an inconsistent state in the pool. For clients fetched from a pool, the server API is mutable. This may apply to other properties of the client as well, so we may want to create a follow-up ticket for this.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-client-pool.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-pool.c
@@ -51,6 +51,7 @@ struct _mongoc_client_pool_t {
    int32_t error_api_version;
    bool error_api_set;
    mongoc_server_api_t *api;
+   bool client_initialized;
 };
 
 
@@ -240,12 +241,13 @@ _initialize_new_client (mongoc_client_pool_t *pool, mongoc_client_t *client)
       pool->topology->scanner->initiator,
       pool->topology->scanner->initiator_context);
 
+   pool->client_initialized = true;
+   client->is_pooled = true;
    client->error_api_version = pool->error_api_version;
    _mongoc_client_set_apm_callbacks_private (
       client, &pool->apm_callbacks, pool->apm_context);
 
    client->api = mongoc_server_api_copy (pool->api);
-   client->api_set = true;
 
 #ifdef MONGOC_ENABLE_SSL
    if (pool->ssl_opts_set) {
@@ -529,7 +531,7 @@ mongoc_client_pool_set_server_api (mongoc_client_pool_t *pool,
       return false;
    }
 
-   if (mongoc_client_pool_get_size (pool)) {
+   if (pool->client_initialized) {
       bson_set_error (error,
                       MONGOC_ERROR_POOL,
                       MONGOC_ERROR_POOL_API_TOO_LATE,
--- a/src/libmongoc/src/mongoc/mongoc-client-pool.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-pool.c
@@ -245,6 +245,7 @@ _initialize_new_client (mongoc_client_pool_t *pool, mongoc_client_t *client)
       client, &pool->apm_callbacks, pool->apm_context);
 
    client->api = mongoc_server_api_copy (pool->api);
+   client->api_set = true;
 
 #ifdef MONGOC_ENABLE_SSL
    if (pool->ssl_opts_set) {
@@ -525,6 +526,14 @@ mongoc_client_pool_set_server_api (mongoc_client_pool_t *pool,
                       MONGOC_ERROR_POOL,
                       MONGOC_ERROR_POOL_API_ALREADY_SET,
                       "Cannot set server api more than once per pool");
+      return false;
+   }
+
+   if (mongoc_client_pool_get_size (pool)) {
+      bson_set_error (error,
+                      MONGOC_ERROR_POOL,
+                      MONGOC_ERROR_POOL_API_TOO_LATE,
+                      "Cannot set server api after a client has been created");
       return false;
    }
 
--- a/src/libmongoc/src/mongoc/mongoc-client-pool.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-pool.c
@@ -529,5 +529,6 @@ mongoc_client_pool_set_server_api (mongoc_client_pool_t *pool,
    }
 
    pool->api = mongoc_server_api_copy (api);
+   _mongoc_topology_scanner_set_server_api (pool->topology->scanner, api);
    return true;
 }
--- a/src/libmongoc/src/mongoc/mongoc-client-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-client-private.h
@@ -97,6 +97,7 @@ struct _mongoc_client_t {
    mongoc_uri_t *uri;
    mongoc_cluster_t cluster;
    bool in_exhaust;
+   bool is_pooled;
 
    mongoc_stream_initiator_t initiator;
    void *initiator_data;
@@ -119,7 +120,6 @@ struct _mongoc_client_t {
    bool error_api_set;
 
    mongoc_server_api_t *api;
-   bool api_set;
 
    /* mongoc_client_session_t's in use, to look up lsids and clusterTimes */
    mongoc_set_t *client_sessions;
--- a/src/libmongoc/src/mongoc/mongoc-client-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-client-private.h
@@ -119,6 +119,7 @@ struct _mongoc_client_t {
    bool error_api_set;
 
    mongoc_server_api_t *api;
+   bool api_set;
 
    /* mongoc_client_session_t's in use, to look up lsids and clusterTimes */
    mongoc_set_t *client_sessions;
--- a/src/libmongoc/src/mongoc/mongoc-cmd.c
+++ b/src/libmongoc/src/mongoc/mongoc-cmd.c
@@ -796,8 +796,8 @@ _txn_in_progress (mongoc_cmd_parts_t *parts)
    }
 
    return (_mongoc_client_session_txn_in_progress (cs)
-	   /* commitTransaction and abortTransaction count as in progress, too. */
-	   || parts->assembled.is_txn_finish);
+      /* commitTransaction and abortTransaction count as in progress, too. */
+      || parts->assembled.is_txn_finish);
 }
 
 
--- a/src/libmongoc/src/mongoc/mongoc-cmd.c
+++ b/src/libmongoc/src/mongoc/mongoc-cmd.c
@@ -1150,6 +1150,7 @@ _mongoc_cmd_append_payload_as_array (const mongoc_cmd_t *cmd, bson_t *out)
  *
  * Pre-conditions:
  *    - @api is initialized.
+ *    - @command_body is initialised
  *
  *--------------------------------------------------------------------------
  */
@@ -1159,6 +1160,7 @@ _mongoc_cmd_append_server_api (bson_t *command_body,
 {
    const char *string_version;
 
+   BSON_ASSERT (command_body);
    BSON_ASSERT (api);
 
    string_version = mongoc_server_api_version_to_string (api->version);
--- a/src/libmongoc/src/mongoc/mongoc-cmd.c
+++ b/src/libmongoc/src/mongoc/mongoc-cmd.c
@@ -827,14 +827,12 @@ mongoc_cmd_parts_assemble (mongoc_cmd_parts_t *parts,
                            bson_error_t *error)
 {
    mongoc_server_description_type_t server_type;
-   mongoc_server_api_t *api;
    mongoc_client_session_t *cs;
    const bson_t *cluster_time = NULL;
    mongoc_read_prefs_t *prefs = NULL;
    const char *cmd_name;
    bool is_get_more;
    const mongoc_read_prefs_t *prefs_ptr;
-   const char *string_version;
    bool ret = false;
 
    ENTRY;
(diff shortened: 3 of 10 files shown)


src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -528,6 +529,15 @@ mongoc_client_pool_set_server_api (mongoc_client_pool_t *pool,
...
+   if (mongoc_client_pool_get_size (pool)) {
+      bson_set_error (error,
+                      MONGOC_ERROR_POOL,
+                      MONGOC_ERROR_POOL_API_TOO_LATE,
+                      "Cannot set server api after a client has been created");
+      return false;
+   }
+
    pool->api = mongoc_server_api_copy (api);
+   _mongoc_topology_scanner_set_server_api (pool->topology->scanner, api);
```
alcaeus, line 541: Not sure if this is considered kosher or not. Please advise if this should be injected into the scanner through other means.

src/libmongoc/src/mongoc/mongoc-cmd.c
```diff
@@ -796,8 +796,8 @@ _txn_in_progress (mongoc_cmd_parts_t *parts)
    }
 
    return (_mongoc_client_session_txn_in_progress (cs)
-	   /* commitTransaction and abortTransaction count as in progress, too. */
-	   || parts->assembled.is_txn_finish);
+      /* commitTransaction and abortTransaction count as in progress, too. */
```
alcaeus, line 799: Unrelated whitespace change, but the tabs really stood out.

samantharitter: Ugh that's from my text editor, I'm not sure why clang-format didn't fix it for me.  Sorry!

src/libmongoc/tests/test-mongoc-topology.c
```diff
@@ -2202,6 +2203,94 @@ test_slow_server_pooled (void)
...
+   } else {
+      client = mongoc_client_new_from_uri (uri);
+      ASSERT_OR_PRINT (mongoc_client_set_server_api (client, api, &error),
+                       error);
+   }
+
+   ismaster = bson_strdup_printf ("{'ok': 1,"
+                                  " 'ismaster': true,"
+                                  " 'setName': 'rs',"
+                                  "  'minWireVersion': 2,"
```
samantharitter, line 2243: super nit, weird spacing

src/libmongoc/tests/test-mongoc-versioned-api.c
```diff
@@ -135,6 +135,43 @@ _test_mongoc_server_api_client_pool (void)
...
+      error,
+      MONGOC_ERROR_POOL,
+      MONGOC_ERROR_POOL_API_TOO_LATE,
+      "Cannot set server api after a client has been created");
+
+   ASSERT (!mongoc_client_set_server_api (client, api, &error));
+   ASSERT_ERROR_CONTAINS (error,
+                          MONGOC_ERROR_CLIENT,
+                          MONGOC_ERROR_CLIENT_API_ALREADY_SET,
+                          "Cannot set server api more than once");
```
samantharitter, line 167: This is interesting.  So, when we get the client from the pool, it does *not* have an api set (`client->api` is NULL) but we've set the `client->api_set` flag to prevent users from setting the `api` directly on the client?  It's a bit of a misleading message, I would rather that the message said something like "can't set server api directly on a pooled client, please set it on the parent pool"

samantharitter (changes requested):
I like the refactoring you did in cmd, and these changes seem like they are on the right track, just a couple of questions from me.

src/libmongoc/src/mongoc/mongoc-client-private.h
```diff
@@ -119,6 +119,7 @@ struct _mongoc_client_t {
    bool error_api_set;
 
    mongoc_server_api_t *api;
+   bool api_set;
```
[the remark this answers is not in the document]
bazile-clyde: +1

src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -528,6 +529,15 @@ mongoc_client_pool_set_server_api (mongoc_client_pool_t *pool,
       return false;
    }
 
+   if (mongoc_client_pool_get_size (pool)) {
+      bson_set_error (error,
+                      MONGOC_ERROR_POOL,
+                      MONGOC_ERROR_POOL_API_TOO_LATE,
+                      "Cannot set server api after a client has been created");
+      return false;
+   }
```
bazile-clyde, line 538: To avoid the edge you mentioned of having the pool size fall back down to zero, alternatively, I think adding `bool api_set` to the pool, rather than the client (per Sam's comment), and setting it to `true` once a client has been created (popped) would solve the problem. Here we'd need to check that boolean (which would be immutable once changed), and not the pool size. Though, the name `api_set` might be confusing since the actual pointer to the API struct could still be `NULL`.

src/libmongoc/src/mongoc/mongoc-client-private.h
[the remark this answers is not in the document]
bazile-clyde: I think this solves the issue of a client being created from a pool without an API set. The client's API pointer would be left `NULL`, but we wouldn't want to allow users to set the API pointer either.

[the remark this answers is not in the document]
alcaeus: That's exactly the reason. On second thought and after our conversation yesterday where you brought up CDRIVER-3537, I'll rename this to `is_pooled` and use it to track whether the client is part of a client pool. This allows me to print a more appropriate error message and also allow us to re-use the flag later when implementing other checks.

src/libmongoc/src/mongoc/mongoc-topology-scanner.c
```diff
@@ -1324,3 +1339,20 @@ _jumpstart_other_acmds (mongoc_topology_scanner_node_t *node,
       }
    }
 }
+
+void
+_mongoc_topology_scanner_set_server_api (mongoc_topology_scanner_t *ts,
+                                         const mongoc_server_api_t *api)
+{
+   mongoc_server_api_t *prev_api;
```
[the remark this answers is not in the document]
alcaeus: I believe this was left over from a previous version of the code. I've updated this and am now calling `mongoc_server_api_destroy` unconditionally.

src/libmongoc/src/mongoc/mongoc-topology-scanner.c
```diff
@@ -104,9 +104,22 @@ _jumpstart_other_acmds (mongoc_topology_scanner_node_t *node,
...
+
+   if (api) {
+      _mongoc_cmd_append_server_api (cmd, api);
+   }
+}
+
+static void
+_reset_ismaster (mongoc_topology_scanner_t *ts)
+{
+   bson_init (&ts->ismaster_cmd);
```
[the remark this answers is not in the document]
alcaeus: Yes, this may have leaked memory. `bson_reinit` only works on a `bson_t` that was previously initialised, so I've split up the method. I've also realised that we shouldn't clear the `cluster_time`, so I've removed that from `_reset_ismaster` as well.

src/libmongoc/tests/test-mongoc-topology.c
alcaeus: Copypasta strikes again. I've aligned the indentation across all occurrences in the file.

src/libmongoc/tests/test-mongoc-versioned-api.c
alcaeus: As per above, I'll change the flag name to `is_pooled` and will check that for pooled client. This makes the flag reusable in CDRIVER-3537 if we decide to block setting other properties.

src/libmongoc/src/mongoc/mongoc-client-pool.c
alcaeus: Good point, I was considering that. I'll add a `client_created` flag that gets set automatically when the first client is created.

src/libmongoc/src/mongoc/mongoc-client-private.h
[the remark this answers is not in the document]
samantharitter: Sounds good!

src/libmongoc/tests/test-mongoc-topology.c
```diff
@@ -2202,6 +2203,94 @@ test_slow_server_pooled (void)
...
+   request_t *request;
+   bson_error_t error;
+   mongoc_server_api_version_t version;
+   mongoc_server_api_t *api;
+
+   server = mock_server_new ();
+   mock_server_run (server);
+   uri = mongoc_uri_copy (mock_server_get_uri (server));
+
+   mongoc_server_api_version_from_string ("1", &version);
```
bazile-clyde, line 2225: Nit, because this returns `false` on an error you could `BSON_ASSERT` this works.

bazile-clyde (commented):
Looks good. Just a few comments.

src/libmongoc/src/mongoc/mongoc-topology-scanner.c
```diff
@@ -1324,3 +1346,15 @@ _jumpstart_other_acmds (mongoc_topology_scanner_node_t *node,
...
+void
+_mongoc_topology_scanner_set_server_api (mongoc_topology_scanner_t *ts,
+                                         const mongoc_server_api_t *api)
+{
+   BSON_ASSERT (ts);
+   BSON_ASSERT (api);
+
+   mongoc_server_api_destroy (ts->api);
+   ts->api = mongoc_server_api_copy (api);
+   _reset_ismaster (ts);
```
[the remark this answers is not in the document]
alcaeus: When creating a new client using `mongoc_client_new`, this implicitly also creates the `topology_scanner_t` through `mongoc_topology_new` and `mongoc_topology_scanner_new`. This last method prepares the basic `isMaster`/`hello` command. At that point in time, the user hasn't had a chance to call `mongoc_client_set_server_api`, hence the need to reset any previously created handshake commands to recreate them with the versioned API options.

Since `_mongoc_topology_scanner_set_server_api` is part of the private API, I haven't bothered to add checks against changing the previously declared API version, especially since the logic required due to the initialisation flow (i.e. changing from no declared API to a specific API) completely covers the scenario of changing from one declared API version to another.

src/libmongoc/src/mongoc/mongoc-cmd.c
```diff
@@ -1155,3 +1138,39 @@ _mongoc_cmd_append_payload_as_array (const mongoc_cmd_t *cmd, bson_t *out)
...
+ *
+ * _mongoc_cmd_append_server_api --
+ *    Append versioned API fields to a mongoc_cmd_t
+ *
+ * Arguments:
+ *    cmd The mongoc_cmd_t, which will have versioned API fields added
+ *    api A mongoc_server_api_t holding server API information
+ *
+ * Pre-conditions:
+ *    - @api is initialized.
```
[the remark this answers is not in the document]
alcaeus: Done

src/libmongoc/tests/test-mongoc-topology.c
alcaeus: Not a nit ;)

Done.

src/libmongoc/src/mongoc/mongoc-topology-scanner.c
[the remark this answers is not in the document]
bazile-clyde: Got it. Thanks for the explanation.

bazile-clyde (approved):

src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -51,6 +51,7 @@ struct _mongoc_client_pool_t {
    int32_t error_api_version;
    bool error_api_set;
    mongoc_server_api_t *api;
+   bool client_initialized;
```
samantharitter, line 54: to clarify, this is meant to be set when the first client created by this pool is initialized?

src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -528,6 +531,15 @@ mongoc_client_pool_set_server_api (mongoc_client_pool_t *pool,
       return false;
    }
 
+   if (pool->client_initialized) {
```
samantharitter, line 534: Can we use `pool->size > 0` here instead and lose the new flag?  Unless you specified a max pool size of 0 (which I don't think would allow the pool to create any clients at all) then `pool->size` should always be > 0 if a client has been created.

samantharitter (commented):
I have one question about the `pool->client_initialized flag`, but the rest of the changes look good!

src/libmongoc/src/mongoc/mongoc-client-pool.c
alcaeus: Yes

src/libmongoc/src/mongoc/mongoc-client-pool.c
alcaeus: On second thought, that would've worked. I had this in previously but refactored this since the logic would allow changes when `pool->size == 0`. Reading the logic in `mongoc_client_pool_push`, this case can't happen, as old clients are only dropped from the pool if `pool->min_pool_size` is truish, in which case you also can't have the last client dropped. I can change it back to the previous check with `pool->size`, but I feel like the flag is more explicit, even though it requires additional memory.

PR #754, opened 2021-03-11
CDRIVER-3773 Remove support for canonical regular expressions with no options

benjirewis:
Reverts mongodb/mongo-c-driver#751.

We cannot support regular expressions with no options without a spec change because of an existing spec [test](https://github.com/mongodb/specifications/blob/master/source/bson-corpus/tests/top.json#L74..L77) that checks for a parse error on a regex with no options.

[CDRIVER-3773](https://jira.mongodb.org/browse/CDRIVER-3773) would require a larger spec change, which I will look into.


bazile-clyde (approved):
LGTM! I'd wait for a few of the tests to pass and then merge.

bazile-clyde (approved):

PR #756, opened 2021-03-15
CDRIVER-3918 expose server api options as mongoc_optional_t

samantharitter:
This PR adds a new `mongoc_optional_t` type (which represents an optional bool) and returns that type for boolean options for versioned API.  I thought through a couple of questions when implementing this:

- What about other optional types?  Is it odd that `mongoc_optional_t` is only for bools?
The only other type I could see us wanting to represent this way is an integer.  Everything else can be pointer-based to indicate NULL-ness.  I think we could implement a `mongoc_optional_int_t` later on if we wanted to, and we actually may need specifically named types for int32, int64, etc.  Given that, I decided using `mongoc_optional_t` here would be nicer than `mongoc_optional_bool_t`.

- Is this over-engineered?
Maybe!  We do represent optional booleans in a fair number of places, and I think this little type is a nice way to wrap that up.

- Expose the type, or not?
It's our convention in most places to hide struct definitions and allow users to access them only through getters and setters.  However, we lose a lot of the convenience that a `mongoc_optional_t` offers if we require `opt.value` and `opt.is_set` to be accessed through getters and setters.  I chose to expose the struct and offer getters and setters, but happy to discuss this.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-optional.h
+++ b/src/libmongoc/src/mongoc/mongoc-optional.h
@@ -26,15 +26,15 @@
 BSON_BEGIN_DECLS
 
 typedef struct {
-    bool value;
-    bool is_set;
+   bool value;
+   bool is_set;
 } mongoc_optional_t;
 
 MONGOC_EXPORT (void)
 mongoc_optional_init (mongoc_optional_t *opt);
 
 MONGOC_EXPORT (bool)
-mongoc_optional_set (mongoc_optional_t *opt);
+mongoc_optional_is_set (mongoc_optional_t *opt);
 
 MONGOC_EXPORT (bool)
 mongoc_optional_value (mongoc_optional_t *opt);
--- a/src/libmongoc/src/mongoc/mongoc-optional.h
+++ b/src/libmongoc/src/mongoc/mongoc-optional.h
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2021 MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


src/libmongoc/src/mongoc/mongoc-optional.h
```diff
@@ -0,0 +1,50 @@
...
+typedef struct {
+    bool value;
+    bool is_set;
+} mongoc_optional_t;
+
+MONGOC_EXPORT (void)
+mongoc_optional_init (mongoc_optional_t *opt);
+
+MONGOC_EXPORT (bool)
+mongoc_optional_set (mongoc_optional_t *opt);
```
alcaeus, line 37: Nit: `mongoc_optional_set` doesn't sound like a getter. Maybe consider renaming this to `mongoc_optional_is_set`?

alcaeus (changes requested):
Code LGTM in general. There are still a couple of usages of the old `strict_set` and `deprecation_errors_set` members in mongoc-cmd.c and test-mongoc-versioned-api.c.

src/libmongoc/src/mongoc/mongoc-optional.h
samantharitter: oh ha I didn't consider that!  Yes, is_set is a much clearer choice.

samantharitter:
Hey @alcaeus this is ready for re-review!  Apologies about the lingering references to `strict_set` and `deprecation_errors_set`, there was some code I neglected to push to this branch before opening the PR.  Hopefully the patch will be more green now.

alcaeus (approved):
LGTM! We'll see if evergreen agrees with me...

PR #758, opened 2021-03-16
CDRIVER-3905 Change estimatedDocumentCount to use collStats on 4.9+

benjirewis:
[CDRIVER-3905](https://jira.mongodb.org/browse/CDRIVER-3905)

Test changes in first three commits, code changes in remaining three.

Updates `mongoc_collection_estimated_document_count` to use an aggregate with the $collStats pipeline stage when running against server version 4.9 or higher. Adds a new wire version macro for wire version 12.

Adds new spec tests for estimatedDocumentCount under `crud/unified` and moves old CRUD tests to `crud/legacy`. Updates unified test runner accordingly. Updates retryable reads and versioned API spec tests involving estimatedDocumentCount.


src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -825,24 +882,45 @@ mongoc_collection_estimated_document_count (
...
+      ret = mongoc_collection_read_command_with_opts (
+         coll, &cmd, read_prefs, opts, reply_ptr, error);
+
+      if (error && error->code == MONGOC_ERROR_COLLECTION_DOES_NOT_EXIST) {
+         /* Collection does not exist. From spec: return 0 but no err:
+          * https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#estimateddocumentcount
+          */
+         error = NULL;
+         count = 0;
+         GOTO (done);
```
benjirewis, line 918: Although this swallows the error and returns 0, it does not touch the passed-in `reply` which still contains the NamespaceNotFound error. This is problematic, as users will not expect the reply to contain an error when the returned `error` is NULL. I could instead destroy `reply` and document that `reply` may be NULL when running against a non-existent collection. I could also mock a server response in `reply`, but mocking server behavior client-side sounds like trouble. 

@alcaeus I'm curious what would make more sense in PHP.

src/libmongoc/tests/unified/result.c
```diff
@@ -349,11 +349,6 @@ result_check (result_t *result,
 
    /* check result. */
    if (expect_result) {
-      if (!result->ok) {
-         test_set_error (
-            error, "expected result, but got error: %s", result->error.message);
-         goto done;
-      }
```
benjirewis, line 356: This was an existing bug in the unified test runner. We should not necessarily be expecting the result to be ok when `expectEvents` are specified according to the [spec](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#expectedevent). I'll handle this change through a separate PR and ticket and rebase this PR eventually.

src/libmongoc/tests/bsonutil/bson-parser.c
```diff
@@ -638,30 +638,33 @@ bson_parser_parse (bson_parser_t *parser, bson_t *in, bson_error_t *error)
...
    bson_parser_entry_t *entry = NULL;
 
-   BSON_FOREACH (in, iter)
-   {
-      const char *key = bson_iter_key (&iter);
-      bson_parser_entry_t *matched = NULL;
-      /* Check for a corresponding entry. */
-      LL_FOREACH (parser->entries, entry)
+   /* Verify that key is not NULL. */
+   if (in) {
```
benjirewis, line 642: This was also an existing bug in the unified test runner. `bson_parser_parse` is sometimes called when the passed in document `in` is NULL, and it will cause an error in `BSON_FOREACH` if unchecked. I'll handle this change through a separate PR and ticket and rebase this PR eventually.

src/libmongoc/src/mongoc/mongoc-collection.c
alcaeus: PHP does not use the CRUD API from libmongoc but implements its own in the PHP library. As such, we're not affected by changes to this method.

alcaeus: Upon taking a closer look, we'll have to consider the server reply in general: even when the collection exists, the reply will be different from what a user might expect to be returned. I'm not sure how much of this is covered by a BC promise, but users won't see the usual `count` reply document when running on 4.9+.

src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -825,24 +882,45 @@ mongoc_collection_estimated_document_count (
...
+      /* On >= 4.9, use aggregate with collStats for estimatedDocumentCount */
+      _make_aggregate_for_edc (coll, &cmd);
+      ret = mongoc_collection_read_command_with_opts (
+         coll, &cmd, read_prefs, opts, reply_ptr, error);
+
+      if (error && error->code == MONGOC_ERROR_COLLECTION_DOES_NOT_EXIST) {
+         /* Collection does not exist. From spec: return 0 but no err:
+          * https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#estimateddocumentcount
+          */
+         error = NULL;
```
alcaeus, line 916: I may be mistaken, but I believe that this will cause memory to leak as we're not resetting error information, but rather setting the pointer to `NULL`. This should either use `bson_set_error` or `memset` to make sure there's no error information.

src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -825,24 +882,45 @@ mongoc_collection_estimated_document_count (
...
+         count = 0;
+         GOTO (done);
+      }
+      if (ret && bson_iter_init(&iter, reply_ptr)) {
+         if (bson_iter_find_descendant (&iter, "cursor.firstBatch.0.n", &iter)) {
+            count = bson_iter_as_int64 (&iter);
+         }
```
alcaeus, line 923: If the command was successful but we don't find the data we're looking for, we're returning a count of `-1` without exposing any additional information. It may be sensible to set `error` to something that indicates that we didn't receive the information we were looking for.

src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -825,24 +882,45 @@ mongoc_collection_estimated_document_count (
...
+                                             coll->read_concern,
+                                             coll->write_concern,
+                                             reply_ptr,
+                                             error);
+      if (ret) {
+         if (bson_iter_init_find (&iter, reply_ptr, "n")) {
+            count = bson_iter_as_int64 (&iter);
+         }
+      }
+   } else {
```
alcaeus, line 906: I mentioned this in the other comment and want to leave the final decision to @kevinAlbs and @bazile-clyde, but we may want to consider crafting a `reply` that matches the `count` command reply document when using agg pipeline to determine document count. While technically the details of how we get to a document count are internal and shouldn't be relied on by users, the fact that the reply document changes significantly may present a problem for some users.

I don't think we need to do anything for the error case, but in case we're returning any count we may want to craft a document like `{ "n" : <count>, "ok" : 1 }` to return when using agg pipeline to retrieve document count.

src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -796,6 +796,59 @@ mongoc_collection_count_with_opts (
...
+
+   bson_append_document_begin (&pipeline, keys[key], 1, &group_stage);
+   bson_append_document_begin (&group_stage, "$group", 6, &group_stage_doc);
+   bson_append_int32 (&group_stage_doc, "_id", 3, 1);
+   bson_append_document_begin (&group_stage_doc, "n", 1, &sum);
+   bson_append_utf8 (&sum, "$sum", 4, "$count", 6);
+   bson_append_document_end (&group_stage_doc, &sum);
+   bson_append_document_end (&group_stage, &group_stage_doc);
+   bson_append_document_end (&pipeline, &group_stage);
+   bson_append_array_end (out, &pipeline);
```
alcaeus, line 850: Using the `BSON_APPEND_` macros would make this a little easier to read as it removes the need to pass string lengths for everything. Compare `bson_append_utf8` for example:
```
bson_append_utf8 (out, "aggregate", 9, coll->collection, coll->collectionlen);

BSON_APPEND_UTF8 (out, "aggregate", coll->collection);
```

Depending on whether clang-format allows you to do so,  leveraging indentation could make this a little more readable as well:
```
BSON_APPEND_UTF8 (out, "aggregate", coll->collection);

[...]

BSON_APPEND_DOCUMENT_BEGIN (&pipeline, keys[key++], &coll_stats_stage);
   BSON_APPEND_DOCUMENT_BEGIN (&coll_stats_stage, "$collStats", &coll_stats_stage_doc);
      BSON_APPEND_DOCUMENT_BEGIN (&coll_stats_stage_doc, "count", &empty);
      bson_append_document_end (&coll_stats_stage_doc, &empty);
   bson_append_document_end (&coll_stats_stage, &coll_stats_stage_doc);
bson_append_document_end (&pipeline, &coll_stats_stage);
```

This could help make sense of the nesting a little better, but as I said, not sure if clang-format will allow you to do this.

alcaeus (commented):
Have a couple of notes, but the changes LGTM in general. I'd like to let @bazile-clyde or @kevinAlbs chime in on the reply and error pointer comments once they're available.

src/libmongoc/src/mongoc/mongoc-collection.c
benjirewis: Awesome! I wasn't aware of these macros. I've added them in but unfortunately cannot do the indentation you described. As you predicted, clang-format will not allow indentation of regular statements like this.

src/libmongoc/src/mongoc/mongoc-collection.c
bazile-clyde: I'm not sure if this a memory leak but I'm sure it's a no-op. Because C passes everything by value, this line changes a copy of the parameter's address and not the actual parameter itself or its contents. If this was passed by reference somehow then this would be a memory leak. Here's a code snippet of what's happening:
```
void faux_mod(int *ptr) {
	ptr = NULL;
	printf("faux_mod: address=%p\n", ptr ? ptr : 0);
}

void real_mod(int *ptr) {
	*ptr = 2;
	printf("real_mod: address=%p,val=%d\n", ptr ? ptr : 0, ptr ? *ptr : 0);
}

int main() {
   int* ptr;
   ptr = (int*) malloc(sizeof(int));
   *ptr = 1;

   printf("address=%p,val=%d\n", ptr ? ptr : 0, ptr ? *ptr : 0);

   faux_mod(ptr);
   printf("address=%p,val=%d\n", ptr ? ptr : 0, ptr ? *ptr : 0);

   real_mod(ptr);
   printf("address=%p,val=%d\n", ptr ? ptr : 0, ptr ? *ptr : 0);

   printf("set to NULL\n");
   ptr = NULL;
   printf("address=%p,val=%d\n", ptr ? ptr : 0, ptr ? *ptr : 0);

   free(ptr);
}

...
# output
address=0x558bcab3e2a0,val=1
faux_mod: address=(nil)
address=0x558bcab3e2a0,val=1
real_mod: address=0x558bcab3e2a0,val=2
address=0x558bcab3e2a0,val=2
set to NULL
address=(nil),val=0

```

src/libmongoc/src/mongoc/mongoc-collection.c
benjirewis: It's a great point, and it seems like we should put some thought into the consistency of the returned `reply`. I've created [CDRIVER-3933](https://jira.mongodb.org/browse/CDRIVER-3933) to do this work separately from this PR. The spec only requires us to return an int64 `count` and makes no mention of returning the server response, so we may get rid of `reply` entirely.

src/libmongoc/src/mongoc/mongoc-collection.c
benjirewis: Yep, you're totally right! Very helpful explanation, thank you @bazile-clyde . I only use `bson_set_error` now.

src/libmongoc/src/mongoc/mongoc-collection.c
benjirewis: As mentioned in the other thread, [CDRIVER-3933](https://jira.mongodb.org/browse/CDRIVER-3933) will be used for considering the future of the estimated document count `reply`.

src/libmongoc/src/mongoc/mongoc-collection.c
benjirewis: An interesting point! The estimated document count function did not return an error on a successful but malformed server response before this added functionality, so I'm not sure what a meaningful error would be. Any thoughts @bazile-clyde ?

src/libmongoc/tests/bsonutil/bson-parser.c
benjirewis: Done.

src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -796,6 +796,60 @@ mongoc_collection_count_with_opts (
...
+   int key = 0;
+
+   bson_init (out);
+   BSON_APPEND_UTF8 (out, "aggregate", coll->collection);
+   BSON_APPEND_DOCUMENT_BEGIN (out, "cursor", &cursor_empty);
+   bson_append_document_end (out, &cursor_empty);
+   BSON_APPEND_ARRAY_BEGIN (out, "pipeline", &pipeline);
+
+   BSON_APPEND_DOCUMENT_BEGIN (&pipeline, keys[key], &coll_stats_stage);
+   key++;
```
alcaeus, line 835: Nit: I feel like the `keys` variable doesn't bring any additional value:
```suggestion
   BSON_APPEND_DOCUMENT_BEGIN (&pipeline, "0", &coll_stats_stage);
```

src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -796,6 +796,60 @@ mongoc_collection_count_with_opts (
...
+   BSON_APPEND_DOCUMENT_BEGIN (
+      &coll_stats_stage, "$collStats", &coll_stats_stage_doc);
+   BSON_APPEND_DOCUMENT_BEGIN (&coll_stats_stage_doc, "count", &empty);
+   bson_append_document_end (&coll_stats_stage_doc, &empty);
+   bson_append_document_end (&coll_stats_stage, &coll_stats_stage_doc);
+   bson_append_document_end (&pipeline, &coll_stats_stage);
+
+   BSON_APPEND_DOCUMENT_BEGIN (&pipeline, keys[key], &group_stage);
```
alcaeus, line 843: Same nit as above:
```suggestion
   BSON_APPEND_DOCUMENT_BEGIN (&pipeline, "1", &group_stage);
```

src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -796,6 +796,60 @@ mongoc_collection_count_with_opts (
...
+   bson_t pipeline;
+   bson_t coll_stats_stage;
+   bson_t coll_stats_stage_doc;
+   bson_t group_stage;
+   bson_t group_stage_doc;
+   bson_t sum;
+   bson_t cursor_empty;
+   bson_t empty;
+   const char *keys[] = {"0", "1"};
+   int key = 0;
```
alcaeus, line 826: Can be dropped if you decide to accept the suggestions below:
```suggestion
```

alcaeus (approved):
LGTM besides nits I couldn't help myself with.

src/libmongoc/src/mongoc/mongoc-collection.c
alcaeus: Sounds good to me, thanks for creating the follow-up ticket!

src/libmongoc/src/mongoc/mongoc-collection.c
alcaeus: Since this is an extremely narrow use-case (only thing I can imagine is the server not returning an initial batch in the cursor), I'm fine leaving this as is and revisiting as part of CDRIVER-3933 or another ticket.

src/libmongoc/src/mongoc/mongoc-collection.c
alcaeus: Thanks for that explanation @bazile-clyde, I completely forgot about pass-by-value 🤦‍♂️

src/libmongoc/src/mongoc/mongoc-collection.c
benjirewis: Sounds good! Done.

src/libmongoc/src/mongoc/mongoc-collection.c
benjirewis: Done.

src/libmongoc/src/mongoc/mongoc-collection.c
benjirewis: Done.

alcaeus:
You're a victim of your own success - want to include changes from https://github.com/mongodb/specifications/pull/933 in this PR?

bazile-clyde (approved):
LGTM!

src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -825,24 +879,47 @@ mongoc_collection_estimated_document_count (
...
+      ret = mongoc_collection_read_command_with_opts (
+         coll, &cmd, read_prefs, opts, reply_ptr, error);
+
+      if (error && error->code == MONGOC_ERROR_COLLECTION_DOES_NOT_EXIST) {
+         /* Collection does not exist. From spec: return 0 but no err:
+          * https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#estimateddocumentcount
+          */
+         error->code = 0;
+         error->domain = 0;
+         memset (error->message, 0, strlen (error->message));
```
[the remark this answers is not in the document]
benjirewis: Ah that seems better! Thanks.

benjirewis:
I can't remove the skipReason just yet, as the "latest" server version hasn't updated on Evergreen to include [this commit](https://evergreen.mongodb.com/version/mongodb_mongo_master_075e51fefecc54e219c56e75c3ba08c993b5c187). So, I'll do CDRIVER-3932 in a separate PR.

PR #759, opened 2021-03-17
CDRIVER-3792 Support parsing $uuid as extended JSON

alcaeus:
CDRIVER-3792

This adds support for parsing UUIDs from extended JSON into BSON. To avoid using a UUID library, this uses sscanf to first validate the UUID and extract the data skipping hyphens, then decodes the hex values into the format understood by the BSON binary type. Any malformed UUID (e.g. wrong length, wrong position of hyphens, etc.) is rejected.

### DIFF
--- a/src/libbson/src/bson/bson-json.c
+++ b/src/libbson/src/bson/bson-json.c
@@ -940,7 +940,6 @@ _bson_json_parse_binary_elem (bson_json_reader_t *reader,
       uuid[32] = '\0';
 
       if (nread != 36 || val_w_null[nread] != '\0') {
-         binary_len = 0;
          _bson_json_read_set_error (reader,
                                     "Invalid input string \"%s\", looking for "
                                     "a dash-separated UUID string",
@@ -953,7 +952,7 @@ _bson_json_parse_binary_elem (bson_json_reader_t *reader,
       _bson_json_buf_ensure (&bson->bson_type_buf[0], (size_t) binary_len + 1);
 
       if (!_unhexlify_uuid (
-             &uuid[0], bson->bson_type_buf[0].buf, (size_t) binary_len + 1)) {
+             &uuid[0], bson->bson_type_buf[0].buf, (size_t) binary_len)) {
          _bson_json_read_set_error (reader,
                                     "Invalid input string \"%s\", looking for "
                                     "a dash-separated UUID string",
--- a/src/libbson/src/bson/bson-json.c
+++ b/src/libbson/src/bson/bson-json.c
@@ -103,7 +103,8 @@ static const char *read_state_names[] = {FOREACH_READ_STATE (GENERATE_STRING)};
    BS (DECIMAL128)                                                   \
    BS (DBPOINTER)                                                    \
    BS (SYMBOL)                                                       \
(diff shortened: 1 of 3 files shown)


src/libbson/src/bson/bson-json.c
```diff
@@ -833,8 +835,29 @@ _bson_json_read_int64_or_set_error (bson_json_reader_t *reader, /* IN */
    return true;
 }
 
+static bool
+_unhexlify_uuid (const char *uuid, uint8_t *out, size_t max)
```
alcaeus, line 839: Shamelessly copied this from `corpus_test_unhexlify`. Let me know if this functionality should be extracted. I've adapted the code to work directly on the BSON buffer but can refactor this to allocate a buffer of the correct size, as is done in the original method.

If there's value in having this available in general, I can also extract this (e.g. to `bson_uuid_pton` to be consistent with the `bson_b64_*` methods.

src/libbson/src/bson/bson-json.c
```diff
@@ -894,6 +917,50 @@ _bson_json_parse_binary_elem (bson_json_reader_t *reader,
...
+
+      data->binary.has_binary = true;
+      data->binary.has_subtype = true;
+      data->binary.type = BSON_SUBTYPE_UUID;
+
+      /* Validate the UUID and extract relevant portions */
+      /* We can't use %x here as it allows +, -, and 0x prefixes */
+      SSCANF (val_w_null,
+              "%8[0-9a-fA-F]-%4[0-9a-fA-F]-%4[0-9a-fA-F]-%4[0-9a-fA-F]-%12[0-"
+              "9a-fA-F]%n",
```
alcaeus, line 932: Using `sscanf` seemed like the most plausible way to check this. We don't check the return value here since we test that we read 36 bytes and encountered a null byte at the end below. This ensures that we don't accept strings that start out like a valid UUID but contain additional data in the end.

src/libbson/src/bson/bson-json.c
```diff
@@ -967,6 +1034,8 @@ _bson_json_read_string (bson_json_reader_t *reader, /* IN */
       case BSON_JSON_LF_BINARY:
       case BSON_JSON_LF_TYPE:
          bson->bson_type_data.binary.is_legacy = true;
+         /* Break omitted intentionally */
```
alcaeus, line 1037: I assumed that we don't need to set `is_legacy` here, but I'm not very familiar with what it's supposed to do. Please advise if I should change this.

src/libbson/src/bson/bson-json.c
alcaeus: Looking at the failures, I would suspect that not every system supports the `[0-9a-fA-F]` notation. I've changed it to use `%8c` instead but am getting access violation on Windows. The linux versions seem to work fine, so I'm at a bit of a loss as to what's causing this.

alcaeus: Found it: the `sscanf_s` function used in Windows environments requires an additional argument specifying the buffer size when using `%c`.

bazile-clyde (approved):
LGTM. Just one comment

PR #760, opened 2021-03-17
CDRIVER-3931 Check for null document in bson parser parse

benjirewis:
[CDRIVER-3931](https://jira.mongodb.org/browse/CDRIVER-3931)

The bson test util `bson_parser_parse` can run BSON_FOREACH on a NULL document causing an error in BSON_ASSERT. Adds a check to verify the document isn't NULL.


bazile-clyde (approved):
LGTM! Thanks for creating the ticket for this.

PR #762, opened 2021-03-18
CDRIVER-3917 route all test client and client_pool creation through test framework

samantharitter:
This PR renames some of the test framework methods around client and client_pool creation so they match the signatures of the `mongoc_client_new` and `mongoc_client_pool_new` methods.  It also creates new methods in the test framework for client and pool creation, so that all clients and pools in our tests are created through the framework.

The framework stub methods right now have simple pass through to `mongoc_client_new` and `mongoc_client_pool_new`, so these changes should have no behavior changes in our tests.  The next piece of work for CDRIVER-3917 will be to change these stub methods to add in versioned API information to client and pool creation.

### DIFF
--- a/src/libmongoc/tests/test-libmongoc.h
+++ b/src/libmongoc/tests/test-libmongoc.h
@@ -109,7 +109,7 @@ test_framework_client_new_from_uri (const mongoc_uri_t *uri);
 mongoc_client_pool_t *
 test_framework_new_default_client_pool (void);
 mongoc_client_pool_t *
-test_framework_client_pool_new (const mongoc_uri_t *uri);
+test_framework_client_pool_new_from_uri (const mongoc_uri_t *uri);
 
 bool
 test_framework_is_mongos (void);
--- a/src/libmongoc/tests/test-libmongoc.h
+++ b/src/libmongoc/tests/test-libmongoc.h
@@ -98,10 +98,18 @@ void
 test_framework_set_ssl_opts (mongoc_client_t *client);
 void
 test_framework_set_pool_ssl_opts (mongoc_client_pool_t *pool);
+
 mongoc_client_t *
-test_framework_client_new (void);
+test_framework_new_default_client (void);
+mongoc_client_t *
+test_framework_client_new (const char *uri_str);
+mongoc_client_t *
+test_framework_client_new_from_uri (const mongoc_uri_t *uri);
+
 mongoc_client_pool_t *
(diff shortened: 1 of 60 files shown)


src/libmongoc/tests/test-libmongoc.h
```diff
@@ -98,10 +98,18 @@ void
...
+test_framework_client_new (const char *uri_str);
 mongoc_client_t *
-test_framework_client_new (void);
+test_framework_client_new_from_uri (const mongoc_uri_t *uri);
+
+mongoc_client_pool_t *
+test_framework_new_default_client_pool (void);
 mongoc_client_pool_t *
-test_framework_client_pool_new (void);
+test_framework_client_pool_new (const mongoc_uri_t *uri);
```
alcaeus, line 112: I dislike that this does not match the signature of `test_framework_client_new` but rather `test_framework_client_new_from_uri`.

My suggestion would be:
```suggestion
test_framework_client_pool_new (const char *uri_str);
mongoc_client_t *
test_framework_client_pool_new_from_uri (const mongoc_uri_t *uri);
```

alcaeus (commented):
I didn't bother to look at the result of the refactoring, so I'll let evergreen decide on that.

src/libmongoc/tests/test-libmongoc.h
samantharitter: The reason I changed this to take a uri was to match the signature of `mongoc_client_pool_new`, which takes a uri:

```
MONGOC_EXPORT (mongoc_client_pool_t *)
mongoc_client_pool_new (const mongoc_uri_t *uri);
```

I find it confusing that the test framework doesn't match the real class' methods.  It's also a bit weird that `mongoc_client_pool_t` and `mongoc_client_t` have methods whose signatures don't agree, but that's a discussion for another day.

Do you think it's more important that `test_framework_client_pool_*` agrees with `mongoc_client_pool_*`, or that it agrees with `test_framework_client_*` ? Unfortunately we can't have both :(

alcaeus: I think it's more natural to have the `test_framework_client_*` methods have consistent signatures and deal with any API differences internally. This at least fixes the wrong signature issue in tests, leaving us to figure out the actual API on a different day.

alcaeus (approved):

kevinAlbs:
Being done in #787.

PR #763, opened 2021-03-23
CDRIVER-3215 Assert transaction state does not change on client-side error

benjirewis:
[CDRIVER-3215](https://jira.mongodb.org/browse/CDRIVER-3215)

Adds new spec test to assert that transaction state does not move from both "starting" or "in progress" when a client-side error is encountered.

Updates test runner to support "assertSessionTransactionState" described [here](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#assertsessiontransactionstate).

### DIFF
--- a/src/libmongoc/tests/json-test-operations.c
+++ b/src/libmongoc/tests/json-test-operations.c
@@ -2255,6 +2255,49 @@ assert_event_count (json_test_ctx_t *ctx, const bson_t *operation)
    }
 }
 
+static void
+assert_session_transaction_state (json_test_ctx_t *ctx, const bson_t *operation)
+{
+   const char *expected_state;
+   const mongoc_client_session_t *session;
+
+   expected_state = bson_lookup_utf8 (operation, "arguments.state");
+   session = session_from_name (
+      ctx, bson_lookup_utf8 (operation, "arguments.session"));
+
+   if (!strcmp (expected_state, "none")) {
+      if (session->txn.state != MONGOC_INTERNAL_TRANSACTION_NONE) {
+         test_error ("expected session transaction state none, but have %i",
+                     session->txn.state);
+      }
+   } else if (!strcmp (expected_state, "starting")) {
+      if (session->txn.state != MONGOC_INTERNAL_TRANSACTION_STARTING) {
+         test_error ("expected session transaction state starting, but have %i",
+                     session->txn.state);
+      }
+   } else if (!strcmp (expected_state, "in_progress")) {
+      if (session->txn.state != MONGOC_INTERNAL_TRANSACTION_IN_PROGRESS) {
+         test_error (
+            "expected session transaction state in progress, but have %i",
+            session->txn.state);
+      }
+   } else if (!strcmp (expected_state, "committed")) {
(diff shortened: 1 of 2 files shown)


bazile-clyde (approved):
LGTM! Just two small comments. 
1) since the transaction state is an enum the error message we'd return would contain a number and not a name of a transaction state. For example:
```
expected session transaction state in progress, but have 2
```
2) Alternatively, you could've switched on the transaction state to make sure we do something for each transaction state as opposed to ensuring we do something for each `assertSessionTransactionState.state`. Though, I think this is fine since will still error if the two are mismatched.

src/libmongoc/tests/json-test-operations.c
```diff
@@ -2255,6 +2255,49 @@ assert_event_count (json_test_ctx_t *ctx, const bson_t *operation)
...
+{
+   const char *expected_state;
+   const mongoc_client_session_t *session;
+
+   expected_state = bson_lookup_utf8 (operation, "arguments.state");
+   session = session_from_name (
+      ctx, bson_lookup_utf8 (operation, "arguments.session"));
+
+   if (!strcmp (expected_state, "none")) {
+      if (session->txn.state != MONGOC_INTERNAL_TRANSACTION_NONE) {
```
bazile-clyde, line 2269: Here and below. I'd used `mongoc_client_session_get_transaction_state` instead of reaching into the session to handle the case where `state == MONGOC_INTERNAL_TRANSACTION_COMMITTED_EMPTY`

benjirewis: Sounds good! Used `mongoc_client_session_get_transaction_state` instead. Added `state_strings` to print actual names of states, too.

I could have used the switch on the transaction state, but since `assertSessionTransactionState.state` represents a subset of possible transaction states, I thought the else if ladder was more readable.

src/libmongoc/tests/json-test-operations.c
```diff
@@ -2260,37 +2260,46 @@ assert_session_transaction_state (json_test_ctx_t *ctx, const bson_t *operation)
...
    expected_state = bson_lookup_utf8 (operation, "arguments.state");
    session = session_from_name (
       ctx, bson_lookup_utf8 (operation, "arguments.session"));
+   state = mongoc_client_session_get_transaction_state (session);
 
    if (!strcmp (expected_state, "none")) {
-      if (session->txn.state != MONGOC_INTERNAL_TRANSACTION_NONE) {
-         test_error ("expected session transaction state none, but have %i",
-                     session->txn.state);
+      if (state != MONGOC_INTERNAL_TRANSACTION_NONE) {
```
bazile-clyde, line 2278: lastly `MONGOC_INTERNAL_TRANSACTION_NONE` should be `MONGOC_TRANSACTION_NONE` to correspond to the states in `mongoc_transaction_state_t` as opposed to `mongoc_internal_transaction_state_t`

benjirewis: Done!

bazile-clyde (approved):
LGTM

PR #766, opened 2021-03-24
CDRIVER-3933 Reinit reply in edc for NamespaceNotFound

benjirewis:
[CDRIVER-3933](https://jira.mongodb.org/browse/CDRIVER-3933)

Destroys and re-initializes `reply` in `mongoc_collection_estimated_document_count` in the case of a NamespaceNotFound error caused by an aggregate on a non-existent collection. Also removes the workaround for 3933 in `operation.c`.

Per the spec, drivers are required to swallow a NamespaceNotFound error and return a count of 0. While the `bson_error_t` was being unset, and a count of 0 was being returned, `reply` still housed the error, which is arguably unexpected behavior for the user.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-collection.c
+++ b/src/libmongoc/src/mongoc/mongoc-collection.c
@@ -909,7 +909,9 @@ mongoc_collection_estimated_document_count (
          /* Collection does not exist. From spec: return 0 but no err:
           * https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#estimateddocumentcount
           */
-         bson_reinit (reply);
+         if (reply) {
+            bson_reinit (reply);
+         }
          memset (error, 0, sizeof *error);
          count = 0;
          GOTO (done);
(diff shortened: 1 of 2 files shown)


src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -909,6 +909,7 @@ mongoc_collection_estimated_document_count (
          /* Collection does not exist. From spec: return 0 but no err:
           * https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#estimateddocumentcount
           */
+         bson_reinit (reply);
```
bazile-clyde, line 912: the reply may be null but `bson_reinit` asserts that it isn't before attempting to reinitialize it. The simplest solution would be to make sure you have a reply before call `bson_reinit`.

bazile-clyde (changes requested):

src/libmongoc/src/mongoc/mongoc-collection.c
benjirewis: Great point! Added in a null check.

bazile-clyde (approved):
LGTM

kevinAlbs (approved):
LGTM!

PR #767, opened 2021-03-25
CDRIVER-3924 Ignore RCs when comparing test version strings

benjirewis:
[CDRIVER-3924](https://jira.mongodb.org/browse/CDRIVER-3924)

Ignores the fourth field of version strings in `_parse_server_version` and `test_framework_str_to_version`. Updates `test_version_cmp` to test this behavior.

According to the unified test format [spec](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#version-string), we should only consider `major`, `minor` and `patch` when comparing server version strings for testing. Release candidates like `"3.4.0.-49"` should not be distinct from non-rcs like `"3.4.0"`.

I _think_ the changes in this PR will also affect tests outside of the unified test format, so let me know if that seems like a problem.

### DIFF
--- a/src/libmongoc/tests/test-libmongoc.c
+++ b/src/libmongoc/tests/test-libmongoc.c
@@ -2096,14 +2096,10 @@ _parse_server_version (const bson_t *buildinfo)
    /* Server returns a 4-part version like [3, 2, 0, 0], or like [3, 2, 0, -49]
     * for an RC. Ignore the 4th part since RCs are equivalent to non-RCs for
     * testing purposes. */
-   i = 0;
-   while (bson_iter_next (&array_iter)) {
-      if (i == N_SERVER_VERSION_PARTS) {
-         break;
-      }
+   for (i = 0; i < N_SERVER_VERSION_PARTS && bson_iter_next (&array_iter);
+        i++) {
       ret *= 1000;
       ret += 100 + bson_iter_as_int64 (&array_iter);
-      i++;
    }
 
    ASSERT_CMPINT (i, ==, N_SERVER_VERSION_PARTS);
@@ -2142,20 +2138,15 @@ test_framework_str_to_version (const char *version_str)
    server_version_t ret = 0;
 
    str_copy = bson_strdup (version_str);
-   i = 0;
    part = strtok (str_copy, ".");
 
    /* Versions can have 4 parts like "3.2.0.0", or like "3.2.0.-49" for an RC.
     * Ignore the 4th part since RCs are equivalent to non-RCs for testing
     * purposes. */
-   while (part) {
-      if (i == N_SERVER_VERSION_PARTS) {
-         break;
-      }
+   for (i = 0; i < N_SERVER_VERSION_PARTS && part; i++) {
       ret *= 1000;
       ret += 100 + bson_ascii_strtoll (part, &end, 10);
       part = strtok (NULL, ".");
-      i++;
    }
 
    /* pad out a short version like "3.0" to three parts */
--- a/src/libmongoc/tests/test-libmongoc.c
+++ b/src/libmongoc/tests/test-libmongoc.c
@@ -2080,7 +2080,7 @@ test_framework_max_write_batch_size (void)
    return size;
 }
 
-#define N_SERVER_VERSION_PARTS 4
+#define N_SERVER_VERSION_PARTS 3
 
 static server_version_t
 _parse_server_version (const bson_t *buildinfo)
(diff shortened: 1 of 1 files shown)


alcaeus (approved):
In theory, changing other tests shouldn't affect this, as the intent is always the same. Our spec tests don't care about pre-release versions but want to test things against `4.9.0` for example. Any pre-release version is fine, since our test suite only tests against the latest version for 4.9 anyways. With the original logic, 4.9.0-rc0 wouldn't pass for 4.9.0 in most drivers (including PHP where I first noticed this), which is why this was clarified in the spec.

src/libmongoc/tests/test-libmongoc.c
```diff
@@ -2094,17 +2094,18 @@ _parse_server_version (const bson_t *buildinfo)
    ASSERT (bson_iter_recurse (&iter, &array_iter));
 
    /* Server returns a 4-part version like [3, 2, 0, 0], or like [3, 2, 0, -49]
-    * for an RC. Bail if number of parts is ever not 4. */
+    * for an RC. Ignore the 4th part since RCs are equivalent to non-RCs for
+    * testing purposes. */
    i = 0;
    while (bson_iter_next (&array_iter)) {
+      if (++i == N_SERVER_VERSION_PARTS) {
```
bazile-clyde, line 2101: Why not change `N_SERVER_VERSION_PARTS` to 3. I think that would ensure we're only looking at the first 3 parts though I haven't tested this locally so it may be something I'm missing.

benjirewis: Done. It's about the same amount of code, as you still need to break early when iterating through server versions, but it seems more accurate to change `N_SERVER_VERSION_PARTS` to 3.

src/libmongoc/tests/test-libmongoc.c
```diff
@@ -2094,17 +2094,19 @@ _parse_server_version (const bson_t *buildinfo)
    ASSERT (bson_iter_recurse (&iter, &array_iter));
 
    /* Server returns a 4-part version like [3, 2, 0, 0], or like [3, 2, 0, -49]
-    * for an RC. Bail if number of parts is ever not 4. */
+    * for an RC. Ignore the 4th part since RCs are equivalent to non-RCs for
+    * testing purposes. */
    i = 0;
    while (bson_iter_next (&array_iter)) {
+      if (i == N_SERVER_VERSION_PARTS) {
```
bazile-clyde, line 2101: I think if the while loop above was a for loop this could be added to (or in place of, assuming the server always returns a 4-part version) the current conditional statement. This appears to be a more verbose for loop already so it shouldn't cause any changes to the logic really but should be a bit easier to reason about.

Here and below on line 2152.

bazile-clyde (changes requested):

src/libmongoc/tests/test-libmongoc.c
benjirewis: You're totally right; still have vacation brain 🤦 🌴  .

Switched both to `for` loops. The conditional statements do have two parts, one for `N_SERVER_VERSION_PARTS` and one for iterating over the BSON/string.

bazile-clyde (approved):
LGTM!

kevinAlbs (commented):
I think this is a good cleanup change, but to check, I don't think this changes functionality, right?

> I think the changes in this PR will also affect tests outside of the unified test format, so let me know if that seems like a problem.


I don't think that should be a problem. There other two places I can think of where we do server version checks.

1. The unified test runner's `check_run_on_requirement`. It also appears to ignore RC. I ran a few examples and modified poc-crud.json to check against a 5.0.0-alpha server like so:

```
# modify runOnRequirements in poc-crud.json an run to check:
MONGOC_JSON_SUBTEST="Aggregate with $listLocalSessions" ./cmake-build/src/libmongoc/test-libmongoc --no-fork -d -l "/unified/poc-crud"
```

2. The old spec test runner's `check_version_info`, which also goes through `_parse_server_version`.

src/libmongoc/tests/test-libmongoc.c
```diff
@@ -2096,11 +2096,8 @@ _parse_server_version (const bson_t *buildinfo)
    /* Server returns a 4-part version like [3, 2, 0, 0], or like [3, 2, 0, -49]
     * for an RC. Ignore the 4th part since RCs are equivalent to non-RCs for
```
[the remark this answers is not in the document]
benjirewis: Slightly confusing: when implementing versioned API testing in Evergreen ([CDRIVER-3917](https://github.com/mongodb/mongo-c-driver/pull/773)), I needed to ignore RCs to get `4.9.0-alpha...` to be equivalent to `4.9.0`... Was planning on rebasing off of this PR, but Evergreen testing was approved sooner. So, this PR is now just a cleanup but does not change the functionality in master.

And great! Thanks for finding those two other server version checks.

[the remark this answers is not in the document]
kevinAlbs: Ah that makes sense, thank you for the explanation!

kevinAlbs (approved):
LGTM!

PR #770, opened 2021-03-30
CDRIVER-3940 Fix missing versioned API options in server monitor

alcaeus:
CDRIVER-3940

The server monitor assembles its own `isMaster` command without using the topology scanner. This logic was missing the Versioned API fields, causing errors in background monitoring when connected to servers with `requireApiVersion` enabled.

This PR changes the logic to always fetch the hello command from the topology scanner. To make the distinction easier, I renamed `_mongoc_topology_scanner_get_ismaster` to `_mongoc_topology_scanner_get_handshake_cmd` and added a new `_mongoc_topology_scanner_get_hello_cmd` to return the base command without additional handshake data. This contains Versioned API fields, and we append the necessary fields for awaitable hello.

Note that this logic will likely have to change again in the near future, as the server monitor needs to know whether a server supports `hello` or if it needs to revert to `isMaster`. Versioned API will also bring more changes to the handshake process.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-server-monitor.c
+++ b/src/libmongoc/src/mongoc/mongoc-server-monitor.c
@@ -326,10 +326,12 @@ _server_monitor_polling_ismaster (mongoc_server_monitor_t *server_monitor,
                                   bson_error_t *error)
 {
    bson_t cmd;
+   const bson_t *hello;
    bool ret;
 
-   bson_init (&cmd);
-   bson_append_int32 (&cmd, "isMaster", 8, 1);
+   hello = _mongoc_topology_scanner_get_hello_cmd (server_monitor->topology->scanner);
+   bson_copy_to (hello, &cmd);
+
    _server_monitor_append_cluster_time (server_monitor, &cmd);
    ret = _server_monitor_send_and_recv_opquery (
       server_monitor, &cmd, ismaster_reply, error);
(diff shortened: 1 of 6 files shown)


src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -326,10 +326,12 @@ _server_monitor_polling_ismaster (mongoc_server_monitor_t *server_monitor,
                                   bson_error_t *error)
 {
    bson_t cmd;
+   const bson_t *hello;
    bool ret;
 
-   bson_init (&cmd);
-   bson_append_int32 (&cmd, "isMaster", 8, 1);
+   hello = _mongoc_topology_scanner_get_hello_cmd (server_monitor->topology->scanner);
```
kevinAlbs, line 332: Renaming this was a good idea. And reducing the duplicate generation of the hello command will make it easier to update when we need to deal with helloOk.

There's no need to change it now. But in CDRIVER-3682 I intend to remove the dependency of the server monitor on the topology scanner to keep then separate. The topology scanner is exclusive to single-threaded scanning, and the server monitor is exclusive to multi-threaded. I can make a note to move the handshake generation outside of the topology scanner as part of CDRIVER-3682 if that sounds ok to you.

kevinAlbs (approved):
LGTM! And thanks for looking into this! I added a comment on CDRIVER-3796. Since there is more divergence in monitoring between single and pooled clients, we can avoid this in the future if we test on both.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
alcaeus: That sounds good. Thanks!

PR #771, opened 2021-03-31
CDRIVER-3861 upload deb/rpm packages to deterministic location

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/6063a6510305b93f1b789de0/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


benjirewis (approved):

prashantmital:
I'm a bit confused by the patch build - which variant should I be looking at?

rcsanchez97:
If you are looking at the `Files` tab of one of the tasks in the build,
the second file (`deb.tar.gz` or `rpm.tar.gz`, depending on which task
you are viewing) is the one that has been added.  The first file is
going to a 'latest' location and so may not be the same file when
retrieved multiple times.  The second file is placed at a location that
ensures it will not be overwritten by a new artifact later on, following
the pattern used in libmongocrypt.

prashantmital:
Ok that sounds like it should work but I still don't know what task I need to be looking at.

kevinAlbs:
The `debian-package-build` task: https://spruce.mongodb.com/task/mongo_c_driver_packaging_debian_package_build_patch_6bacf922669ab780fa5fb37e9c561b8709e27e29_6063a6510305b93f1b789de0_21_03_30_22_29_38/files?execution=0

kevinAlbs (approved):
LGTM! I added a fix version of 1.17.5 to CDRIVER-3861.

PR #772, opened 2021-03-31
Lock topology mutex before setting server API

alcaeus:
Calling `_mongoc_topology_scanner_set_server_api` resets `ismaster_cmd_with_handshake`, which requires obtaining a lock on the topology.


kevinAlbs (approved):
LGTM!

PR #774, opened 2021-03-31
CDRIVER-3935 Lower Evergreen batchtimes

benjirewis:
[CDRIVER-3935](https://jira.mongodb.org/browse/CDRIVER-3935)

Lowers the `batchtime` of the OCSP tests in Evergreen to 10080, or one week. 

The OCSP tests in Evergreen currently have a `batchtime` of 20180, meaning they run only every two weeks. This is not ideal, as many commits can happen in the span of two weeks that could break the OCSP tests. 

We _could_ lower the `batchtime` to something like 4 days, too. Then, we'd probably want to lower the batchtimes of the TSAN and Valgrind tests, as well.


kevinAlbs (approved):
LGTM! Thanks again for researching how `batchtime` works.

bazile-clyde:
> . Then, we'd probably want to lower the batchtimes of the TSAN and Valgrind tests, as well.

I wouldn't be opposed to lowering the batch times of TSAN and Valgrind tests to daily, though I'm not sure why they were weekly in the first place. @kevinAlbs thoughts? It looks like this was done by you in [CDRIVER-748](https://jira.mongodb.org/browse/CDRIVER-748) ([PR](https://github.com/mongodb/mongo-c-driver/pull/653))

.evergreen/config.yml
```diff
@@ -23127,7 +23127,7 @@ buildvariants:
   - name: .ocsp-openssl-1.0.1
     distros:
     - ubuntu1804-test
-  batchtime: 20160
```
bazile-clyde, line 23130: See the top of this file. We don't modify this file directly but instead rely on the python scripts in `mongo-c-driver/build/evergreen_config_lib` to change it for us. In `variants.py` (see line 638), the batch times are in days as well so you shouldn't have to do any calculations from days to seconds directly.

benjirewis: Ah sounds good, I've modified `variants.py` instead of modifying `config.yml` directly.

After offline discussion, I also lowered the TSAN and Valgrind tests' `batchtime`s to 1 day instead of 1 week.

build/evergreen_config_generator/__init__.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2018-present MongoDB, Inc.
```
benjirewis, line 1: Note that I had to add `evergreen_config_generator/*` locally to get the package to work.

bazile-clyde (approved):

build/evergreen_config_generator/__init__.py
kevinAlbs: Sweet, thanks! This will save time for future developers.

kevinAlbs (approved):

PR #776, opened 2021-04-02
Fix packaging

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/6067864dc9ec445641c32eed/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/.evergreen/build_snapshot_rpm.sh
+++ b/.evergreen/build_snapshot_rpm.sh
@@ -89,14 +89,12 @@ sudo mock -r ${config} --use-bootstrap-image --isolation=simple --init
 mock_root=$(sudo mock -r ${config} --use-bootstrap-image --isolation=simple --print-root-path)
 sudo mock -r ${config} --use-bootstrap-image --isolation=simple --install rpmdevtools git rpm-build cmake python python3-sphinx gcc openssl-devel
 sudo mock -r ${config} --use-bootstrap-image --isolation=simple --copyin "$(pwd)" "$(pwd)/${spec_file}" /tmp
-if [ ! -f VERSION_CURRENT ]; then
-  sudo mock -r ${config} --use-bootstrap-image --isolation=simple --cwd "/tmp/${build_dir}" --chroot -- /bin/sh -c "(
-    set -o xtrace ;
-    python build/calc_release_version.py | sed -E 's/([^-]+).*/\1/' > VERSION_CURRENT ;
-    python build/calc_release_version.py -p > VERSION_RELEASED
-    )"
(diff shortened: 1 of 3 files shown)


.evergreen/config.yml
```diff
@@ -1758,7 +1758,7 @@ tasks:
     params:
       aws_key: ${aws_key}
       aws_secret: ${aws_secret}
-      remote_file: ${project}/${branch_name}/${revision}/${version_id}/${build_id}/${execution}/mongo-c-driver-debian-packages.tar.gz
+      remote_file: ${project}/${branch_name}/${revision}/${version_id}/${build_id}/${execution}/mongo-c-driver-rpm-packages.tar.gz
```
kevinAlbs, line 1761: Oh, good catch.

.evergreen/build_snapshot_rpm.sh
```diff
@@ -89,14 +89,12 @@ sudo mock -r ${config} --use-bootstrap-image --isolation=simple --init
 mock_root=$(sudo mock -r ${config} --use-bootstrap-image --isolation=simple --print-root-path)
 sudo mock -r ${config} --use-bootstrap-image --isolation=simple --install rpmdevtools git rpm-build cmake python python3-sphinx gcc openssl-devel
 sudo mock -r ${config} --use-bootstrap-image --isolation=simple --copyin "$(pwd)" "$(pwd)/${spec_file}" /tmp
-if [ ! -f VERSION_CURRENT ]; then
```
kevinAlbs, line 92: Nice, it's better to use calc_release_version.py rather than rely on the hard coded version anyway. Curious, was this necessary to work around an issue? I think it is hard coded to `1.18.0` based on https://github.com/mongodb/mongo-c-driver/blob/master/.evergreen/config.yml#L29:L29, and `1.17.0-pre` on the `r1.17` branch.

kevinAlbs (approved):
LGTM! I just have a couple questions to check my understanding.

.evergreen/build_snapshot_rpm.sh
rcsanchez97: Right.  The issue that hard-coding of VERSION_CURRENT is meant to work around is the age of Git (and possibly Python) on some of the older platforms, making the determination of the version from the Git history problematic. However, when we build the RPM snapshot packages, we always do so in a very recent chroot environment, so we can actually just calculate the correct version.  Additionally, we can't just use {{x.y.z-pre}} in the spec file, because the "-" is not a valid character in that part of the version string.  Hence, it is better to calculate it in the way the snapshot script does it for building the RPMs.

PR #778, opened 2021-04-17
Improve documentation in mongoc_stream_file_new.rst regarding the return value

esquerbatua:
Fix return struct of fn mongoc_stream_file_new

### DIFF
--- a/src/libmongoc/doc/mongoc_stream_file_new.rst
+++ b/src/libmongoc/doc/mongoc_stream_file_new.rst
@@ -8,7 +8,7 @@ Synopsis
 
 .. code-block:: c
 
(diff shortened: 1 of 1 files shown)


kevinAlbs:
Thank you for the contribution @esquerbatua! I believe the documentation for `mongoc_stream_file_new` is correct as-is. The public header returns a `mongoc_stream_t*`: https://github.com/mongodb/mongo-c-driver/blob/c64caa4093d52f36c8f20ee34b1126e4b7be7bed/src/libmongoc/src/mongoc/mongoc-stream-file.h#L32 (though the underlying implementation is a `mongoc_stream_file_t*`).

PR #783, opened 2021-04-21
CDRIVER-3949, CDRIVER-3963, CDRIVER-3966 aborting txn unpins session

jmikola:
https://jira.mongodb.org/browse/CDRIVER-3949
https://jira.mongodb.org/browse/CDRIVER-3963
https://jira.mongodb.org/browse/CDRIVER-3966

There appear to be numerous changes to legacy transaction tests, which leads me to believe libmongoc hadn't been synced in some time. For reference [e728dfb/source/transactions/tests](https://github.com/mongodb/specifications/tree/e728dfbc77e5e3f00a9b8b4301ec59fe14d87eda/source/transactions/tests) is the state of legacy transaction tests before they were moved into a `legacy/` sub-directory. Some of the changes there go back more than a year.


src/libmongoc/tests/unified/operation.c
```diff
@@ -851,14 +852,22 @@ operation_find (test_t *test,
       goto done;
    }
 
+   opts = bson_copy (bson_parser_get_extra (parser));
+   if (op->session) {
+      if (!mongoc_client_session_append (op->session, opts, error)) {
```
jmikola, line 857: I was surprised to discover that the "session" argument was not being handled for all operations. See: [CDRIVER-3964](https://jira.mongodb.org/browse/CDRIVER-3964).

jmikola:
I believe the four ASan task failures are expected.

On a separate note, there will likely be _another_ related libmongoc ticket created for [DRIVERS-1655](https://jira.mongodb.org/browse/DRIVERS-1655) once that is merged. It's up to you if you'd like to sit on this for the time being so we can lump it into this PR.

src/libmongoc/tests/unified/operation.c
kevinAlbs: Good catch, and thank you for filing a ticket to track this.

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -627,7 +627,7 @@ _mongoc_client_session_handle_reply (mongoc_client_session_t *session,
        * fails with a TransientTransactionError". If the server reply included
        * a TransientTransactionError, we unpin here. If a network error caused
        * us to add a label client-side, we unpin in network_error_reply. */
-      session->server_id = 0;
+      _mongoc_client_session_unpin (session);
```
kevinAlbs, line 630: Good eye!

kevinAlbs (approved):
LGTM

> I believe the four ASan task failures are expected.

I agree, checking the logs of all four looks like they are from old server builds not recognizing `$collStats` in the versioned API. CDRIVER-3946 should resolve this.


> On a separate note, there will likely be _another_ related libmongoc ticket created for [DRIVERS-1655](https://jira.mongodb.org/browse/DRIVERS-1655) once that is merged. It's up to you if you'd like to sit on this for the time being so we can lump it into this PR.

At a glance, the behavior in libmongoc may already agree with the proposed changes. `txn_abort` unpins the session even if the abortTransaction command fails. It may just be a test update. I am OK with merging this now and addressing it once DRIVERS-1655 is merged.

PR #784, opened 2021-04-22
CDRIVER-3947 Use "hello" command when an API version is declared

alcaeus:
CDRIVER-3947

This changes the topology scanner to use the `hello` command when an API version is declared on the client. To make this work properly, I had to extend the mock server to not only handle legacy hello commands but also the new `hello` command.

While making those changes, I changed some of the terminology but didn't make a complete effort, as that's covered by the "Remove oppressive language" project.
I also started making changes to the hello autoresponder for mock servers to avoid tests having to specify their own responder for the purpose of checking the hello command body. However, I did not wrap up this refactoring as the diff is already getting large. I'll follow up in a separate PR.

### DIFF
--- a/src/libmongoc/doc/mongoc_server_description_hello_response.rst
+++ b/src/libmongoc/doc/mongoc_server_description_hello_response.rst
@@ -1,6 +1,6 @@
-:man_page: mongoc_server_description_last_hello_response
+:man_page: mongoc_server_description_hello_response
 
-mongoc_server_description_last_hello_response()
+mongoc_server_description_hello_response()
 ===============================================
 
 Synopsis
@@ -9,7 +9,7 @@ Synopsis
 .. code-block:: c
 
   const bson_t *
-  mongoc_server_description_last_hello_response (
+  mongoc_server_description_hello_response (
      const mongoc_server_description_t *description);
 
 Parameters
@@ -24,7 +24,7 @@ The client or client pool periodically runs a
 `"hello" <https://docs.mongodb.org/manual/reference/command/isMaster/>`_
 command on each server, to update its view of the MongoDB deployment. Use
 :symbol:`mongoc_client_get_server_descriptions()` and
-``mongoc_server_description_last_hello_response()`` to get the most recent "hello"
+``mongoc_server_description_hello_response()`` to get the most recent "hello"
 response.
 
 Returns
--- a/src/libmongoc/doc/mongoc_server_description_hello_response.rst
+++ b/src/libmongoc/doc/mongoc_server_description_hello_response.rst
@@ -1,0 +1,34 @@
+:man_page: mongoc_server_description_hello_response
+
+mongoc_server_description_hello_response()
+==========================================
+
+Synopsis
+--------
+
+.. code-block:: c
+
+  const bson_t *
+  mongoc_server_description_hello_response (
+     const mongoc_server_description_t *description);
+
+Parameters
+----------
+
+* ``description``: A :symbol:`mongoc_server_description_t`.
+
+Description
+-----------
+
+The client or client pool periodically runs a
+`"hello" <https://docs.mongodb.org/manual/reference/command/isMaster/>`_
+command on each server, to update its view of the MongoDB deployment. Use
+:symbol:`mongoc_client_get_server_descriptions()` and
+``mongoc_server_description_hello_response()`` to get the most recent "hello"
+response.
+
+Returns
+-------
+
+A reference to a BSON document, owned by the server description. The document is empty if the driver is not connected to the server.
+
--- a/src/libmongoc/doc/mongoc_server_description_ismaster.rst
+++ b/src/libmongoc/doc/mongoc_server_description_ismaster.rst
@@ -12,6 +12,15 @@ Synopsis
   mongoc_server_description_ismaster (
      const mongoc_server_description_t *description);
 
+Deprecated
+----------
+
+.. warning::
+
+  This function is deprecated and should not be used in new code.
+
+Please use :doc:`mongoc_server_description_hello_response() <mongoc_server_description_hello_response>` instead.
+
 Parameters
 ----------
 
--- a/src/libmongoc/doc/mongoc_server_description_ismaster.rst
+++ b/src/libmongoc/doc/mongoc_server_description_ismaster.rst
@@ -19,7 +19,7 @@ Deprecated
 
   This function is deprecated and should not be used in new code.
 
-Please use :doc:`mongoc_server_description_hello_response() <mongoc_server_description_hello_response>` instead.
+Please use :doc:`mongoc_server_description_last_hello_response() <mongoc_server_description_last_hello_response>` instead.
 
 Parameters
 ----------
--- a/src/libmongoc/doc/mongoc_server_description_ismaster.rst
+++ b/src/libmongoc/doc/mongoc_server_description_ismaster.rst
@@ -19,7 +19,7 @@ Deprecated
 
   This function is deprecated and should not be used in new code.
 
-Please use :doc:`mongoc_server_description_last_hello_response() <mongoc_server_description_last_hello_response>` instead.
+Please use :doc:`mongoc_server_description_hello_response() <mongoc_server_description_hello_response>` instead.
 
 Parameters
 ----------
--- a/src/libmongoc/src/mongoc/mongoc-topology-scanner.c
+++ b/src/libmongoc/src/mongoc/mongoc-topology-scanner.c
@@ -296,7 +296,7 @@ _mongoc_topology_scanner_get_handshake_cmd (mongoc_topology_scanner_t *ts)
    if (bson_empty (&ts->ismaster_cmd_with_handshake)) {
       ts->handshake_ok_to_send = _build_ismaster_with_handshake (ts);
       if (!ts->handshake_ok_to_send) {
-         MONGOC_WARNING ("Handshake doc too big, not including in isMaster");
+         MONGOC_WARNING ("Handshake doc too big, not including in hello");
       }
    }
 
@@ -697,7 +697,7 @@ _async_error_or_timeout (mongoc_async_cmd_t *acmd,
       bson_set_error (&node->last_error,
                       MONGOC_ERROR_CLIENT,
                       MONGOC_ERROR_STREAM_CONNECT,
-                      "%s calling ismaster on \'%s\'",
+                      "%s calling hello on \'%s\'",
                       message,
                       node->host.host_and_port);
 
@@ -1014,7 +1014,7 @@ mongoc_topology_scanner_node_setup (mongoc_topology_scanner_node_t *node,
  * mongoc_topology_scanner_node_in_cooldown --
  *
  *      Return true if @node has experienced a network error attempting
- *      to call "ismaster" less than 5 seconds before @when, a timestamp in
+ *      to call "hello" less than 5 seconds before @when, a timestamp in
  *      microseconds.
  *
  *      Server Discovery and Monitoring Spec: "After a single-threaded client
--- a/src/libmongoc/src/mongoc/mongoc-topology-scanner.c
+++ b/src/libmongoc/src/mongoc/mongoc-topology-scanner.c
@@ -113,7 +113,7 @@ _add_ismaster (mongoc_topology_scanner_t *ts)
       BSON_APPEND_INT32 (cmd, "hello", 1);
       _mongoc_cmd_append_server_api (cmd, api);
    } else {
-      BSON_APPEND_INT32 (cmd, "isMaster", 1);
+      BSON_APPEND_INT32 (cmd, HANDSHAKE_CMD_LEGACY_HELLO, 1);
    }
 }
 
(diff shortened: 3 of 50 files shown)


src/libmongoc/doc/mongoc_server_description_hello_response.rst
```diff
@@ -0,0 +1,34 @@
+:man_page: mongoc_server_description_hello_response
+
+mongoc_server_description_hello_response()
```
alcaeus, line 3: Alternatively, we can name this `mongoc_server_description_last_hello_response` to better indicate what this returns.

src/libmongoc/tests/json/server_discovery_and_monitoring/integration/isMaster-command-error.json
```diff
@@ -17,7 +17,8 @@
         },
         "data": {
           "failCommands": [
-            "isMaster"
+            "isMaster",
```
alcaeus, line 20: These test changes are from the corresponding [spec PR](https://github.com/mongodb/specifications/pull/969). I'll amend the commit once the spec change is approved.

src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -488,43 +492,119 @@ mock_server_remove_autoresponder (mock_server_t *server, int id)
...
+   }
+
+   // Convert responses for legacy hello
+   if (bson_iter_init_find (&iter, &response, "isWritablePrimary")) {
+      BSON_APPEND_BOOL (
+         &response, HANDSHAKE_RESPONSE_LEGACY_HELLO, bson_iter_bool (&iter));
+   } else if (bson_iter_init_find (
+                 &iter, &response, HANDSHAKE_RESPONSE_LEGACY_HELLO)) {
+      BSON_APPEND_BOOL (&response, "isWritablePrimary", bson_iter_bool (&iter));
+   }
```
alcaeus, line 550: This piece of code allows us to always use `nonWritablePrimary` in mocked responses, while retaining functionality for tests that might expect the legacy hello response.

src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -935,13 +1002,71 @@ _mock_server_receives_msg (mock_server_t *server, uint32_t flags, ...)
  */
 
 request_t *
-mock_server_receives_ismaster (mock_server_t *server)
+mock_server_receives_legacy_hello (mock_server_t *server,
```
alcaeus, line 1005: I changed the semantics of this function a bit: it now expects either `hello` or legacy hello commands, while the new `mock_server_receives_hello` function always expects a `hello` command.

src/libmongoc/tests/test-libmongoc.c
```diff
@@ -1132,10 +1132,20 @@ call_ismaster_with_host_and_port (char *host_and_port, bson_t *reply)
...
+             client,
+             "admin",
+             tmp_bson ("{'" HANDSHAKE_CMD_LEGACY_HELLO "': 1}"),
+             NULL,
+             reply,
+             &error)) {
+         fprintf (stderr, "error calling hello: '%s'\n", error.message);
+         fprintf (stderr, "URI = %s\n", uri_str);
+         abort ();
+      }
```
alcaeus, line 1148: I opted for manually calling `hello` first and then falling back to legacy hello if the server does not support it. Other tests still use legacy hello to avoid failing tests on older server versions that don't support `hello`. We could wrap this in a `test_mongoc_client_command_hello` function and use that exclusively. Please let me know if you'd like me to add that.

src/libmongoc/tests/test-mongoc-bulk.c
```diff
@@ -2984,26 +2984,26 @@ test_unordered_bulk_writes_with_error (void)
    collection = mongoc_client_get_collection (client, "db", "test");
    /* use an unordered bulk write; we expect to continue on error */
    bulk = mongoc_collection_create_bulk_operation_with_opts (
-      collection, tmp_bson("{'ordered': false}"));
+      collection, tmp_bson ("{'ordered': false}"));
```
alcaeus, line 2987: clang-format fixed some unrelated code, which I didn't bother to exclude from the diff.

src/libmongoc/tests/mock_server/mock-server.h
```diff
@@ -64,9 +69,13 @@ void
 mock_server_remove_autoresponder (mock_server_t *server, int id);
 
 int
-mock_server_auto_ismaster (mock_server_t *server,
-                           const char *response_json,
-                           ...);
+mock_server_auto_hello_callback (mock_server_t *server,
```
kevinAlbs, line 72: IIUC this won't be used outside of mock-server.c. I think this can be removed from the header and marked as static.

src/libmongoc/tests/test-libmongoc.c
```diff
@@ -1132,10 +1132,20 @@ call_ismaster_with_host_and_port (char *host_and_port, bson_t *reply)
...
+      bson_destroy (reply);
+
+      if (!mongoc_client_command_simple (
+             client,
+             "admin",
+             tmp_bson ("{'" HANDSHAKE_CMD_LEGACY_HELLO "': 1}"),
+             NULL,
+             reply,
+             &error)) {
+         fprintf (stderr, "error calling hello: '%s'\n", error.message);
```
kevinAlbs, line 1145:
```suggestion
         fprintf (stderr, "error calling legacy hello: '%s'\n", error.message);
```

src/libmongoc/tests/test-libmongoc.c
kevinAlbs: Nice, I don't feel strongly either way about switching other tests. But perhaps it will be necessary anyway with versioned API testing to use `hello` everywhere?

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -3135,8 +3134,8 @@ test_mongoc_handshake_pool (void)
    mongoc_client_t *client1;
    mongoc_client_t *client2;
    mongoc_client_pool_t *pool;
-   const char *const server_reply =
-      "{'ok': 1, 'ismaster': true, 'minWireVersion': 2, 'maxWireVersion': 5}";
+   const char *const server_reply = "{'ok': 1, 'isWritablePrimary': true, "
```
kevinAlbs, line 3137: Minor: Since `maxWireVersion` is mocked as 5, I suggest using the legacy response field to make it more accurate. 
```suggestion
   const char *const server_reply = "{'ok': 1, '" HANDSHAKE_RESPONSE_LEGACY_HELLO "': true, "
```

Edit: there are a lot of cases like this, where the field name does not match with the mocked wire version. I don't feel too strongly since I think it is a low risk that SDAM incorrectly skips `isWritablePrimary` and we don't catch it somewhere else.

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -3675,6 +3672,7 @@ test_client_reset_connections (void)
 
    request = mock_server_receives_command (
       server, "admin", MONGOC_QUERY_SLAVE_OK, "{'ping': 1}");
+   BSON_ASSERT (request);
```
kevinAlbs, line 3675: Good eye!

src/libmongoc/tests/test-mongoc-cluster.c
```diff
@@ -1231,10 +1229,14 @@ auto_ismaster (request_t *request, void *data)
    dollar_query_test_t *test;
```
kevinAlbs, line 1229: Rename this function to `auto_hello`?

src/libmongoc/tests/test-mongoc-handshake.c
```diff
@@ -405,17 +404,16 @@ test_mongoc_handshake_too_big (void)
    /* Send a ping, mock server deals with it */
    future = future_client_command_simple (
       client, "admin", tmp_bson ("{'ping': 1}"), NULL, NULL, NULL);
-   request = mock_server_receives_ismaster (server);
+   request = mock_server_receives_legacy_hello (server, NULL);
 
    /* Make sure the isMaster request has a handshake field, and it's not huge */
```
kevinAlbs, line 409:
```suggestion
   /* Make sure the hello request has a handshake field, and it's not huge */
```

src/libmongoc/tests/test-mongoc-handshake.c
```diff
@@ -405,17 +404,16 @@ test_mongoc_handshake_too_big (void)
...
    ASSERT (request);
-   ismaster_doc = request_get_doc (request, 0);
-   ASSERT (ismaster_doc);
-   ASSERT (bson_has_field (ismaster_doc, "isMaster"));
-   ASSERT (bson_has_field (ismaster_doc, HANDSHAKE_FIELD));
+   hello_doc = request_get_doc (request, 0);
+   ASSERT (hello_doc);
+   ASSERT (bson_has_field (hello_doc, HANDSHAKE_FIELD));
 
    /* isMaster with handshake isn't too big */
```
kevinAlbs, line 415:
```suggestion
   /* hello with handshake isn't too big */
```

src/libmongoc/tests/test-libmongoc.c
alcaeus: Yes, there's going to be follow-up work once DRIVERS-1293 is ready for implementation. I'll finish this change then.

src/libmongoc/src/mongoc/mongoc-topology-scanner.c
```diff
@@ -249,7 +253,8 @@ _build_ismaster_with_handshake (mongoc_topology_scanner_t *ts)
    int count = 0;
    char buf[16];
 
-   _add_ismaster (doc, ts->api);
+   bson_destroy (doc);
```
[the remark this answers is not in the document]
alcaeus: The old `_add_ismaster` added the command to the `bson_t *` that was passed to the function. `_add_hello` takes the `mongoc_topology_scanner_t *` to always work on the Versioned API declared on it, then adds it to the `hello_cmd` field. I figured that it made more sense to copy that and append additional fields to have `hello_cmd` as the source of truth.

As a side note, DRIVERS-1293 will bring in more changes to this logic, as the server monitor will use `hello` even without an API version declared if the initial handshake response includes the `helloOk` field.

src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -537,31 +617,19 @@ auto_ismaster (request_t *request, void *data)
...
-   }
-   if (!bson_iter_init_find (&iter, tmp, "maxWireVersion")) {
-      BSON_APPEND_INT32 (tmp, "maxWireVersion", WIRE_VERSION_OP_MSG - 1);
-   }
-   bson_free (formatted_response_json);
-   formatted_response_json = bson_as_json (tmp, 0);
 
-   return mock_server_autoresponds (
-      server, auto_ismaster, (void *) formatted_response_json, bson_free);
+   return mock_server_auto_hello_callback (server,
```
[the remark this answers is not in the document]
alcaeus: test-mongoc-cluster.c now uses this callback to build its own response. This can be used in other tests that need to do this or make assertions against the content of the `hello` command.

src/libmongoc/doc/mongoc_server_description_hello_response.rst
alcaeus: Renamed to `mongoc_server_description_last_hello_response`. Changing this to be immutable sounds sensible.

src/libmongoc/tests/test-mongoc-cluster.c
alcaeus: Has been changed to `auto_hello_callback` during refactoring.

src/libmongoc/tests/test-mongoc-client.c
alcaeus: I changed this one and didn't see there were other ones affected as well. I was trying to reproduce the server logic, instead opting for consistent usage of `isWritablePrimary` and only checking for legacy fields in a few specific tests.

alcaeus:
Rebased to account for CDRIVER-3954; the entire Versioned API test suite still passes.

alcaeus:
I'm observing failures in tests I added in this PR related to CDRIVER-3934, so I'll investigate those and then put the PR up for another review.

alcaeus:
Correction: those were tests I added inadvertently. I've created CDRIVER-3993 to sync SDAM spec tests and fix any resulting failures.

kevinAlbs (approved):
LGTM

src/libmongoc/doc/mongoc_server_description_ismaster.rst
```diff
@@ -19,7 +19,7 @@ Deprecated
 
   This function is deprecated and should not be used in new code.
 
-Please use :doc:`mongoc_server_description_hello_response() <mongoc_server_description_hello_response>` instead.
+Please use :doc:`mongoc_server_description_last_hello_response() <mongoc_server_description_last_hello_response>` instead.
```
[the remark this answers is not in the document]
alcaeus: No worries. I've renamed the method as you suggested.

alcaeus:
Failures in the versioned_api task are due to changes required for CDRIVER-3974, which I'll fix in a PR as soon as this is merged.

PR #787, opened 2021-04-28
CDRIVER-3954 Run all integration tests in Versioned API evergreen tests

alcaeus:
This PR includes changes from #762 and adds logic to always set a server API when the corresponding env variable is set. The Versioned API tests are expanded to run all tests to confirm that all commands (including handshake and auth which can't be tested using command monitoring) declare an Api Version.

I'd like to merge this PR before #784 as these tests would provide useful to make sure the driver behaves correctly when using "hello" as well.

@kevinAlbs I couldn't figure out how to add the Versioned API tests to the list of tasks run for every PR. Could you please point me in the right direction?

### DIFF
--- a/.evergreen/integration-tests.sh
+++ b/.evergreen/integration-tests.sh
@@ -166,14 +166,11 @@ fi
 
 if [ ! -z "$REQUIRE_API_VERSION" ]; then
   MONGO_SHELL_CONNECTION_FLAGS="${MONGO_SHELL_CONNECTION_FLAGS} --apiVersion=1"
+  # Set the requireApiVersion parameter.
+  `pwd`/mongodb/bin/mongo $MONGO_SHELL_CONNECTION_FLAGS $DIR/require-api-version.js
 fi
 
 echo $MONGO_SHELL_CONNECTION_FLAGS
 
 `pwd`/mongodb/bin/mongo $MONGO_SHELL_CONNECTION_FLAGS --eval 'printjson(db.serverBuildInfo())' admin
 `pwd`/mongodb/bin/mongo $MONGO_SHELL_CONNECTION_FLAGS --eval 'printjson(db.isMaster())' admin
-
-# Set the requireApiVersion parameter.
-if [ ! -z "$REQUIRE_API_VERSION" ]; then
-  `pwd`/mongodb/bin/mongo $MONGO_SHELL_CONNECTION_FLAGS $DIR/require-api-version.js
-fi
--- a/.evergreen/integration-tests.sh
+++ b/.evergreen/integration-tests.sh
@@ -164,6 +164,10 @@ elif [ "$SSL" != "nossl" ]; then
    MONGO_SHELL_CONNECTION_FLAGS="${MONGO_SHELL_CONNECTION_FLAGS} --host localhost --ssl --sslCAFile=$MONGO_ORCHESTRATION_HOME/lib/ca.pem --sslPEMKeyFile=$MONGO_ORCHESTRATION_HOME/lib/client.pem"
 fi
 
+if [ ! -z "$REQUIRE_API_VERSION" ]; then
+  MONGO_SHELL_CONNECTION_FLAGS="${MONGO_SHELL_CONNECTION_FLAGS} --apiVersion=1"
+fi
+
 echo $MONGO_SHELL_CONNECTION_FLAGS
 
 `pwd`/mongodb/bin/mongo $MONGO_SHELL_CONNECTION_FLAGS --eval 'printjson(db.serverBuildInfo())' admin
--- a/src/libmongoc/tests/TestSuite.c
+++ b/src/libmongoc/tests/TestSuite.c
@@ -808,7 +808,8 @@ TestSuite_PrintJsonHeader (TestSuite *suite, /* IN */
             "  },\n"
             "  \"options\": {\n"
             "    \"fork\": %s,\n"
-            "    \"tracing\": %s\n"
+            "    \"tracing\": %s,\n"
+            "    \"apiVersion\": %s\n"
             "  },\n"
             "  \"results\": [\n",
             egetenv ("MONGOC_TEST_USER"),
@@ -839,7 +840,8 @@ TestSuite_PrintJsonHeader (TestSuite *suite, /* IN */
                                                                  : "false",
             test_framework_getenv_bool ("MONGOC_CHECK_IPV6") ? "true" : "false",
             (suite->flags & TEST_NOFORK) ? "false" : "true",
-            (suite->flags & TEST_TRACE) ? "true" : "false");
+            (suite->flags & TEST_TRACE) ? "true" : "false",
+            egetenv ("MONGODB_API_VERSION"));
 
    bson_free (hostname);
    bson_free (udspath);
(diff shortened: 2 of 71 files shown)


alcaeus:
The build failure in the versioned-api tests is due to `acceptApiVersion2` not being accepted by the server version being tested (see explanation in https://github.com/mongodb/mongo-c-driver/pull/780). The `requireApiVersion` tests pass.

.evergreen/integration-tests.sh
```diff
@@ -165,12 +165,13 @@ elif [ "$SSL" != "nossl" ]; then
    MONGO_SHELL_CONNECTION_FLAGS="${MONGO_SHELL_CONNECTION_FLAGS} --host localhost --ssl --sslCAFile=$MONGO_ORCHESTRATION_HOME/lib/ca.pem --sslPEMKeyFile=$MONGO_ORCHESTRATION_HOME/lib/client.pem"
 fi
 
+if [ ! -z "$REQUIRE_API_VERSION" ]; then
+  MONGO_SHELL_CONNECTION_FLAGS="${MONGO_SHELL_CONNECTION_FLAGS} --apiVersion=1"
```
benjirewis, line 169: Nice, forgot to do this when I put it way down at the bottom.

src/libmongoc/tests/test-libmongoc.c
```diff
@@ -1649,6 +1649,78 @@ test_framework_client_new ()
    return client;
 }
 
+mongoc_server_api_t *
+test_framework_get_default_server_api (void)
+{
+   char *require_api_version = test_framework_getenv ("MONGODB_API_VERSION");
```
benjirewis, line 1655: [nit] Maybe just `api_version`? `require_api_version` implies to me that this would be a `bool`.

src/libmongoc/tests/test-mongoc-bulk.c
```diff
@@ -3773,6 +3777,14 @@ test_bulk_max_msg_size (void)
                        strlen ("$clusterTime") + 2;
    }
 
+   /* Versioned API may be appended */
+   if (client->api) {
+      /* TODO: Figure out why 7 works. */
```
benjirewis, line 3782: Hmm, we would expect to just subtract `2`, right? For the null terminators of the two strings? Was it just guess and check that made 7 work?

src/libmongoc/tests/test-mongoc-bulk.c
```diff
@@ -3773,6 +3777,14 @@ test_bulk_max_msg_size (void)
                        strlen ("$clusterTime") + 2;
    }
 
+   /* Versioned API may be appended */
```
benjirewis, line 3780:
```suggestion
   /* API version may be appended */
```

src/libmongoc/tests/test-mongoc-cache.c
```diff
@@ -41,6 +41,7 @@ ping ()
 
    uri = bson_strdup_printf ("mongodb://localhost/?tls=true&tlsCAFile=%s",
                              ca_file);
+   /* TODO 3954: Replace with test_framework_client_new */
```
benjirewis, line 44: Isn't that this ticket?

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -3464,11 +3464,11 @@ with_transaction_example (bson_error_t *error)
     * members in the URI string; e.g.
     * uri_repl = "mongodb://mongodb0.example.com:27017,mongodb1.example.com:" \
     *    "27017/?replicaSet=myRepl";
-    * client = mongoc_client_new (uri_repl);
+    * client = test_framework_client_new (uri_repl);
```
benjirewis, line 3467: Ah good catch in the examples.

benjirewis (commented):
That's a lot of function renames! Just some nits/questions 😄 

That build failure for `test-versioned-api-accept-version-two` is interesting. From what I can tell, those tests are running against the newest server version and on Ubuntu 18.04, so I think there might be a separate build issue. Looking into it...

src/libmongoc/tests/test-mongoc-cache.c
alcaeus: That todo was obsolete: the code wouldn't compile with the new method, which is something @samantharitter also experienced. I've left this as-is for the time being.

src/libmongoc/tests/test-libmongoc.c
alcaeus: Done

src/libmongoc/tests/test-mongoc-bulk.c
alcaeus: Yes, careful trial-and-error revealed 7 as the magic number. All others append 2, so I'm not entirely sure what's going on. I'll step-debug this and then we'll see.

alcaeus: Through debugging I was able to confirm that the `bson_append_utf8` call in [`_mongoc_cmd_append_server_api`](https://github.com/mongodb/mongo-c-driver/blob/15cbc7865a513c09859fd4c88ba296b9b01139e2/src/libmongoc/src/mongoc/mongoc-cmd.c#L1168) indeed adds 18 bytes to the BSON. 10 for the "apiVersion" string, 1 for the actual string, plus 7 characters that do something. I've removed the comment, and maybe someone with knowledge of this could see if we can find a better way to do this.

kevinAlbs (approved):
LGTM!

> @kevinAlbs I couldn't figure out how to add the Versioned API tests to the list of tasks run for every PR. Could you please point me in the right direction?

This is done as part of the evergreen project config by modifying the task/variant selectors for "GitHub Patch Definitions". I added a definition for all tasks on the versioned-api variant.

src/libmongoc/tests/TestSuite.c
```diff
@@ -839,7 +840,8 @@ TestSuite_PrintJsonHeader (TestSuite *suite, /* IN */
                                                                  : "false",
             test_framework_getenv_bool ("MONGOC_CHECK_IPV6") ? "true" : "false",
             (suite->flags & TEST_NOFORK) ? "false" : "true",
-            (suite->flags & TEST_TRACE) ? "true" : "false");
+            (suite->flags & TEST_TRACE) ? "true" : "false",
+            egetenv ("MONGODB_API_VERSION"));
```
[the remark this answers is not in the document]
alcaeus: Added a section where we document all other environment variables.

src/libmongoc/tests/test-libmongoc.c
```diff
@@ -1649,6 +1649,78 @@ test_framework_client_new ()
...
+
+   ASSERT (mongoc_server_api_version_from_string (api_version, &version));
+
+   bson_free (api_version);
+
+   return mongoc_server_api_new (version);
+}
+
+mongoc_client_t *
+test_framework_client_new (const char *uri_str, const mongoc_server_api_t *api)
```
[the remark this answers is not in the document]
alcaeus: Added a comment here and for all other methods that take an api argument.

benjirewis (commented):
Here's a patch to fix the versioned API build failure that would be good to add in this PR: 
[vapi_patch.txt](https://github.com/mongodb/mongo-c-driver/files/6442305/vapi_patch.txt).

It runs the `test-versioned-api-accept-version-two` on `noauth` and `nossl`. This is what helped the task pass in the Go driver, and `auth` + `ssl` are unnecessary to run the single [test-commands-deprecation-errors](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/tests/json/versioned_api/test-commands-deprecation-errors.json). It also removes redundant `fetch build` steps in the versioned API tasks.

PR #788, opened 2021-04-29
CDRIVER-3937 Remove usages of xtrace in build scripts

alcaeus:
CDRIVER-3937

The first commit is necessary as the `remote_file` is prepended with the project dir automatically. The already generated config.yml was correct, but regenerating broke it without this commit.

### DIFF
--- a/build/evergreen_config_lib/tasks.py
+++ b/build/evergreen_config_lib/tasks.py
@@ -390,7 +390,7 @@ all_tasks = [
                          remote_file='${branch_name}/mongo-c-driver-debian-packages-${CURRENT_VERSION}.tar.gz',
                          content_type='${content_type|application/x-gzip}'),
                   s3_put(local_file='deb.tar.gz',
-                         remote_file='${project}/${branch_name}/${revision}/${version_id}/${build_id}/${execution}/mongo-c-driver-debian-packages.tar.gz',
+                         remote_file='${branch_name}/${revision}/${version_id}/${build_id}/${execution}/mongo-c-driver-debian-packages.tar.gz',
                          content_type='${content_type|application/x-gzip}')]),
     NamedTask('rpm-package-build',
               commands=[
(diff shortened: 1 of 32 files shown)


build/evergreen_config_lib/variants.py
```diff
@@ -635,7 +635,7 @@ def days(n):
         'ubuntu1604-small',
         ['.tsan'],
         {'CC': '/opt/mongodbtoolchain/v3/bin/clang'},
-        batchtime=days(1))
+        batchtime=days(1)),
```
benjirewis, line 638: Ah thank you!

benjirewis (approved):
LGTM!

build/evergreen_config_lib/tasks.py
```diff
@@ -390,7 +390,7 @@ def __init__(self, task_name, suffix_commands, orchestration=True, **kwargs):
                          remote_file='${branch_name}/mongo-c-driver-debian-packages-${CURRENT_VERSION}.tar.gz',
                          content_type='${content_type|application/x-gzip}'),
                   s3_put(local_file='deb.tar.gz',
-                         remote_file='${project}/${branch_name}/${revision}/${version_id}/${build_id}/${execution}/mongo-c-driver-debian-packages.tar.gz',
+                         remote_file='${branch_name}/${revision}/${version_id}/${build_id}/${execution}/mongo-c-driver-debian-packages.tar.gz',
                          content_type='${content_type|application/x-gzip}')]),
```
rcsanchez97, line 394: This change seems unrelated to the xtrace issue.  Also, this change would make the C driver publish distro package artifacts under a path that is not consistent with the pattern used in libmongocrypt and the C++ driver.  Can `${project}/` be restored?

build/evergreen_config_lib/tasks.py
```diff
@@ -399,7 +399,7 @@ def __init__(self, task_name, suffix_commands, orchestration=True, **kwargs):
                          remote_file='${branch_name}/mongo-c-driver-rpm-packages-${CURRENT_VERSION}.tar.gz',
                          content_type='${content_type|application/x-gzip}'),
                   s3_put(local_file='rpm.tar.gz',
-                         remote_file='${project}/${branch_name}/${revision}/${version_id}/${build_id}/${execution}/mongo-c-driver-rpm-packages.tar.gz',
+                         remote_file='${branch_name}/${revision}/${version_id}/${build_id}/${execution}/mongo-c-driver-rpm-packages.tar.gz',
```
rcsanchez97, line 402: Ibid.

build/evergreen_config_lib/tasks.py
alcaeus: This change was introduced in https://github.com/mongodb/mongo-c-driver/commit/c883dab9ff915b2cff60f2ad0ac6695094506eed, but regenerating the config file introduces a double `${project}/` path component in config.yml. This is confirmed by the surrounding lines, where the `remote_file` option does not start with `${project}`, and by the fact that while the template in tasks.py changes, the generated config file does not.

rcsanchez97: OK.  That was totally not expected.  It is clearly a copy/paste on my part in the commit where I introduced the change.  Thanks for catching and fixing it and thanks for the explanation!

rcsanchez97 (approved):
LGTM

PR #789, opened 2021-04-29
document fam reply is empty for w0

kevinAlbs:
If an unacknowledged write concern is used as an option for a
mongoc_find_and_modify_with_opts, the output reply document is
always empty.


benjirewis (approved):
LGTM but PR name seems off!

kevinAlbs:
> LGTM but PR name seems off!

Oops fixed!

PR #790, opened 2021-05-03
CDRIVER-3942 Bump mongocryptd server selection t/o to 10 seconds

benjirewis:
CDRIVER-3942

Bumps the mongocryptd server selection timeout in client side encryption from 5 seconds to 10 seconds.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
@@ -1284,10 +1284,10 @@ _mongoc_cse_client_enable_auto_encryption (mongoc_client_t *client,
                          "Unable to create client to mongocryptd");
          GOTO (fail);
       }
-      /* Similarly, single threaded clients will by default wait for 10 second
+      /* Similarly, single threaded clients will by default wait for 5 second
        * cooldown period after failing to connect to a server before making
        * another attempt. Meaning if the first attempt to mongocryptd fails
-       * to connect, then the user observes a 10 second delay. This is not
+       * to connect, then the user observes a 5 second delay. This is not
        * configurable in the URI, so override. */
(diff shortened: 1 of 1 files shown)


src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
```diff
@@ -1283,19 +1284,19 @@ _mongoc_cse_client_enable_auto_encryption (mongoc_client_t *client,
                          "Unable to create client to mongocryptd");
          GOTO (fail);
       }
-      /* Similarly, single threaded clients will by default wait for 5 second
+      /* Similarly, single threaded clients will by default wait for 10 second
```
kevinAlbs, line 1287: The cooldown timeout is... yet another timeout :)

It is defined in the server monitoring specification: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#cooldownms

This comment and the one below should remain with at 5 seconds.

This is meant to explain why we're bypassing the fixed 5 second cooldown. By default, when client side encryption is enabled, the driver will [spawn the mongocryptd process](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#managing-mongocryptd). If the first operation requiring encryption occurs shortly after, it may occur before mongocryptd is listening for connections. With the cooldown in place, the network error would force a delay of 5 seconds before attempting to reconnect. Bypassing the cooldown permits server monitoring to attempt to reconnect immediately.

kevinAlbs (approved):
LGTM with a revert to two comments!

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
benjirewis: Gotcha, thanks for the explanation! Reverted those comments...

PR #792, opened 2021-05-18
CDRIVER-3974 Append Versioned API options to all commands

alcaeus:
CDRIVER-3974

This fixes test failures in the versioned_api builds. Other builds against latest will continue to fail due `batchSize` calculation for `getMore` resulting in additional `getMore` calls.


alcaeus:
I ran a build with a temporary patch for the cursor/getMore issue: https://spruce.mongodb.com/version/60a3b3f80305b944948f6b9b/tasks. No idea why debug-compile-sspi-winssl failed, but the rest of the tests is stable.

src/libmongoc/src/mongoc/mongoc-cmd.c
```diff
@@ -988,13 +972,10 @@ mongoc_cmd_parts_assemble (mongoc_cmd_parts_t *parts,
             &parts->assembled_body, "$clusterTime", 12, cluster_time);
       }
 
-      /* Add versioned server api, if it is set. Do not add api if we are
-         sending a getmore, or if we are in a transaction. */
+      /* Add versioned server api, if it is set. */
       if (parts->client->api) {
-         if (!is_get_more && !_txn_in_progress (parts)) {
```
kevinAlbs, line 994: Nice! Always appending a versioned API does help simplify this logic.

kevinAlbs (approved):
LGTM!

> No idea why debug-compile-sspi-winssl failed, but the rest of the tests is stable.

Unfortunately, this is a known flaky test failure, tracked in: CDRIVER-3760. I think it is unrelated to these changes.

PR #793, opened 2021-05-19
CDRIVER-4001 Temporarily fix batchSize behaviour for getMore

alcaeus:
CDRIVER-4001

This PR proposes a fix for a number of tests that are currently failing due to slot-based execution (SBE) in the latest 5.0 builds. With SBE enabled we see the server keep the cursor alive when the end of a cursor is reached at the end of a batch. This usually doesn't occur too often, but the getMore command logic replicates the OP_GET_MORE behaviour of reducing the size of the last batch to fetch exactly `limit` documents. This triggers the unintended server behaviour for every cursor. While neither mandated nor prohibited, this behaviour is somewhat enforced by the command monitoring spec tests.

This PR updates the logic for the getMore command (OP_GET_MORE is unaffected) to always send the intended batch size. These changes can be safely reverted once the server behaviour is fixed, or when a better solution comes around.


kevinAlbs (approved):

alcaeus:
Some builds are now failing due to the newer server version no longer rejecting documents with dots and dollars in keys. Will have a separate PR up to fix that.

PR #795, opened 2021-05-20
Skip tests with dots and dollars in field names on 5.0+

alcaeus:
Follows #793. Versioned API tests are fixed by #792. With all patches applied, tests pass again: https://spruce.mongodb.com/version/60a62eabc9ec443716160e3d/tasks

Note that these tests will be replaced by spec tests after the completion of DRIVERS-1237 (see https://github.com/mongodb/specifications/pull/980)

### DIFF
--- a/src/libmongoc/tests/test-mongoc-collection.c
+++ b/src/libmongoc/tests/test-mongoc-collection.c
@@ -6674,20 +6674,24 @@ test_collection_install (TestSuite *suite)
                       NULL,
                       test_framework_skip_if_slow_or_live);
    TestSuite_AddLive (suite, "/Collection/many_return", test_many_return);
-   TestSuite_AddFull (suite,
-                      "/Collection/insert_one_validate",
-                      test_insert_one_validate,
-                      NULL,
-                      NULL,
-                      TestSuite_CheckLive,
-                      test_framework_skip_if_max_wire_version_more_than_9);
-   TestSuite_AddFull (suite,
-                      "/Collection/insert_many_validate",
-                      test_insert_many_validate,
-                      NULL,
-                      NULL,
-                      TestSuite_CheckLive,
-                      test_framework_skip_if_max_wire_version_more_than_9);
+   TestSuite_AddFull (
+      suite,
+      "/Collection/insert_one_validate",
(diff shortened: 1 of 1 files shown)


src/libmongoc/tests/test-mongoc-collection.c
```diff
@@ -4016,7 +4016,7 @@ _test_insert_validate (insert_fn_t insert_fn)
 }
 
 static void
-test_insert_bulk_validate (void)
+test_insert_bulk_validate (void *ctx)
```
jmikola, line 4019: These tests use [error API 2](http://mongoc.org/libmongoc/current/errors.html#setting-the-error-api-version), which means all server-side errors should be reported in the `MONGOC_ERROR_SERVER` domain. The expectations below use `MONGOC_ERROR_COMMAND` as the domain, which tells me these are client-side errors.

In that case, why should the server version have any impact on these tests? If anything, I would expect these tests to be removed entirely as part of [CDRIVER-3895](https://jira.mongodb.org/browse/CDRIVER-3895) since that will remove all client-side validation of dots and dollars. Alternatively, we could change the documents in these tests to violate some other validation rule that will be left in place (e.g. invalid UTF).

alcaeus: This test fails with an assertion error: https://evergreen.mongodb.com/task_log_raw/mongo_c_driver_windows_2017_test_latest_sharded_noauth_nosasl_nossl_patch_d08ef111e1a9cf7e71d916f562b46a9d9bf061c6_60a625487742ae6f40858e18_21_05_20_09_01_04/0?type=T#L2378. You are correct that the first bulk insert expects a command error because it uses no insert validation. There is a second insert that specifies `MONGOC_INSERT_NO_VALIDATE` which expectedly fails on 5.0 and expects a server error.

I was considering rewriting the tests but in the interest of fixing the build so I can get on with the language refactoring (which has the potential of breaking lots of things) I decided to skip these tests for now and revisit them when we implement changes for handling field names with dots and dollars. I've added a TODO to re-enable the tests (also fixed clang-format while I was at it)

jmikola (approved):

PR #796, opened 2021-05-24
Upgrade Ubuntu tasks Evergreen

benjirewis:
CDRIVER-3912
CDRIVER-3946

Runs OCSP tests against Ubuntu 20.04. Exports `LD_LIBRARY_PATH` accurately for each OCSP test to help with linking libraries.

Upgrades existing Ubuntu 16.04 tasks to use Ubuntu 18.04.


benjirewis:
There are still many system-failures for various tasks that I _think_ refer to failures from `download-mongodb.sh`. I've updated that script to its most recent version from `drivers-evergreen-tools`, so I'm not sure what's going on there; looking into it.

kevinAlbs (commented):
I still need to review, but I have a hypothesis on one of the observed compile failures on Ubuntu 18.04.

For the 32 bit failures (i386), I think there is no 32 bit compatible libzstd installed:

```
[2021/05/25 18:20:30.117] /usr/bin/ld: skipping incompatible //usr/lib/x86_64-linux-gnu/libzstd.a when searching for -lzstd
[2021/05/25 18:20:30.117] /usr/bin/ld: cannot find -lzstd
```

The `x86_64` indicates it is ignoring the 64 bit installation of libzstd, but cmake picks up the installation and sees it is a compatible version number (0.8.0).

On the older Ubuntu 16.04 host, the installed libzstd is `0.5.1` (see [logs](https://evergreen.mongodb.com/task_log_raw/mongo_c_driver_clang38_i386_debug_compile_nosasl_nossl_d88b9d17e6e3a9b530df90817902a6b71f7b7c87_21_05_27_13_28_00/0?type=T#L1312)). I think it was bypassing this problem (and would have failed the same way if the installed 64 bit version was > `0.8.0`).

Our cmake scripts are using pkg-config to determine if libzstd is installed [here](https://github.com/mongodb/mongo-c-driver/blob/d88b9d17e6e3a9b530df90817902a6b71f7b7c87/src/libmongoc/CMakeLists.txt#L104:L104), which is effectively the same as:

```
pkg-config --libs --cflags libzstd
```

@rcsanchez97  does this seem like a minor bug to you? If you are configuring for 32 bit via C flags like `CFLAGS="$CFLAGS -m32 -march=i386"` should cmake know that, and skip 64 bit dependencies? If so, I can file a CDRIVER ticket to track fixing how we look up the `zlib` dependency.

@benjirewis , I think a reasonable workaround would be to explicitly disable linking libzstd on 32 bit platforms with `-DENABLE_ZSTD=OFF` [in compile-unix.sh](https://github.com/mongodb/mongo-c-driver/blob/d88b9d17e6e3a9b530df90817902a6b71f7b7c87/.evergreen/compile-unix.sh#L165:L165)

benjirewis:
I've explicitly disable linking libsztd on 32 bit platforms, which did fix some of the failures.

There are still quite a few new system-failures that are happening because `download-mongodb.sh` cannot find the correct version for a few architectures/platforms. I think we need to test versions [3.2, 3.6] on Ubuntu 16.04, and the later versions on Ubuntu 18.04, since download links do not exist for earlier versions on Ubuntu 18.04. A new case needs to be added for `rhel-8.1-ppc64le` (if downloads for that platform exist). Also, the aarch64 architectures seem to be missing the `virtualenv` module for Python.

I'll look into resolving some of those system-failures, but let me know if you have any context on the other, non-system failures. Some are due to "network timeouts", others are due to a failing `_test_invalid_wc_server_error()`.

rcsanchez97:
@benjirewis Have a look at mongodb-labs/drivers-evergreen-tools#154 for an update to the download script that has links for RHEL 8.1 ppc64le.

rcsanchez97:
Here is the explanation for the zstd failures.  The C driver requires zstd 0.8.0.  The zstd installed on Ubuntu 16.04 is version 0.5.1.  The build finds that zstd, and determines that it is too old and so doesn't build against it.  Updating to Ubuntu 18.04 brings with it zstd 1.3.3.  When building on 18.04, CMake finds that the version satisfies the dependency and then tries to link against it.  However, the Evergreen images supplied by the build team only have 64-bit libzstd packages installed.  The build is able to locate the headers (which are not 64-bit specific), but the library itself is 64-bit specific and so the link step fails.  Disabling zstd for 32-bit builds works around this.  Another possible solution is to file a BUILD ticket, requesting installation the 32-bit libzstd packages on Ubuntu 18.04 images.

build/evergreen_config_lib/variants.py
```diff
@@ -274,8 +274,8 @@ def days(n):
              '.latest .nossl'],
             {'CC': 'gcc'}),
     Variant('gcc54-i386',
-            'GCC 5.4 (i386) (Ubuntu 16.04)',
-            'ubuntu1604-test',
+            'GCC 5.4 (i386) (Ubuntu 18.04)',
```
rcsanchez97, line 277: Ubuntu 18.04 has GCC 7.5

build/evergreen_config_lib/variants.py
```diff
@@ -287,8 +287,8 @@ def days(n):
              '.4.0 .nossl .nosasl'],
             {'CC': 'gcc', 'MARCH': 'i386'}),
     Variant('gcc54',
-            'GCC 5.4 (Ubuntu 16.04)',
-            'ubuntu1604-test',
+            'GCC 5.4 (Ubuntu 18.04)',
```
rcsanchez97, line 290: Ubuntu 18.04 has GCC 7.5

build/evergreen_config_lib/variants.py
```diff
@@ -167,8 +167,8 @@ def days(n):
              '.3.6 .nossl .nosasl'],
             {'CC': 'clang', 'MARCH': 'i386'}),
     Variant('clang38ubuntu',
-            'clang 3.8 (Ubuntu 16.04)',
-            'ubuntu1604-test',
+            'clang 3.8 (Ubuntu 18.04)',
```
rcsanchez97, line 170: Ubuntu 18.04 has clang 6.0

build/evergreen_config_lib/variants.py
```diff
@@ -152,8 +152,8 @@ def days(n):
              '.3.0 .nossl .nosasl !.auth'],
             {'CC': 'clang'}),
     Variant('clang38-i386',
-            'clang 3.8 (i386) (Ubuntu 16.04)',
-            'ubuntu1604-test',
+            'clang 3.8 (i386) (Ubuntu 18.04)',
```
rcsanchez97, line 155: Ubuntu 18.04 has clang 6.0

rcsanchez97 (changes requested):
Here are the recommendations for resolving the failures I was able to figure out:

'Release Archive Creator'/abi-compliance-check: file a BUILD ticket requesting installation of package abi-compliance-checker on 18.04 images

*-scan-build: Ubuntu 18.04 does not provide a default scan-build binary, so it is necessary to also try scan-build-3.9 if scan-build is not found as a valid command

test-*: file a BUILD ticket requesting installation of python-virtualenv and python3-virtualenv packages on Ubuntu 18.04 arm64 hosts

test-*: In variants that were updated 16.04 -> 18.04, drop pre-4.0 test tasks (e.g., test-asan-3.4-sharded-auth-nosasl-openssl), or at least the pre-4.0 SSL test tasks.  The download script only has 18.04 download links back to version 4.0.  Trying to download anything older gets you generic binaries that are not going to work for our SSL tests.

The other failing tasks seem to be failing for unrelated reasons.  I would start with the above changes, which should eliminate the majority of the failures, and then we can see what's left after that.

benjirewis:
@rcsanchez97 Thanks for all of the suggestions; I think I've addressed everything you've pointed out.

Filed [BUILD-13262](https://jira.mongodb.org/browse/BUILD-13262) and [BUILD-13263](https://jira.mongodb.org/browse/BUILD-13263) for requesting abi-compliance-checker and python-virtualenv installation on Ubuntu 18.04 (arm64) respectively. Also filed [BUILD-13266](https://jira.mongodb.org/browse/BUILD-13266) to request the installation of 32-bit libzstd packages on Ubuntu 18.04 images.

For now, I've simply disabled linking 32-bit libzstd, but if we can get 32-bit libztd installed on Ubuntu 18.04 images, I'll add the linking back in.

I added logic in `compile-unix` to use the command `scan-build-3.9` if `scan-build` doesn't exist.

I moved the pre-4.0 test tasks that were on Ubuntu 18.04 back to Ubuntu 16.04 and updated the `download-mongodb.sh` script one more time.

I also made sure the descriptions of the GCC and clang versions were accurate to the linux version.

It looks like these changes fix _almost_ all of the unexpected failures.

rcsanchez97:
@benjirewis The last commit looks good, except that you omitted the restoration of the RHEL 8.1 power8 task that you removed in the preceding commit.

benjirewis:
There are still a few failures introduced by this upgrade:

1. A few `test-latest-sharded` tasks are failing (on ASAN Ubuntu 18.04, clang 6.0 (i386) Ubuntu 18.04, GCC 7.5 (i386) Ubuntu 18.04, GCC 7.5 Ubuntu 18.04, and power8 RHEL 8.1) because of [`_test_invalid_wc_server_error`](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/tests/test-mongoc-write-commands.c#L600). The test expects an error code of [91](https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml#L126) but receives an error code of [6](https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml#L42).
2. `rpm-package-build` is failing because it cannot find `/home/ec2-user/rpmbuild/SOURCES/mongo-c-driver-1.19.0.tar.gz` when building.
3. `test-3.4-server-auth-sasl-winssl` [fails](https://evergreen.mongodb.com/lobster/evergreen/task/mongo_c_driver_windows_2015_test_3.4_server_auth_sasl_winssl_patch_0805656817fea6c0a0fee91e6928e2b9333f9564_60bf95d87742ae33e161b9b7_21_06_08_16_07_53/0/task#bookmarks=0%2C4767&l=1&shareLine=2397) [`test_tailable_alive`](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/tests/test-mongoc-cursor.c#L1729) because the cursor is unexpectedly empty.
4. `debug-compile` tasks fail on power8 RHEL 8.1 because OPENSSL cannot be found.
5. `abi-compliance-check` fails because that script does not exist on Ubuntu 18.04 (tracked by [BUILD-13262](https://jira.mongodb.org/browse/BUILD-13262)).
6. ARM Ubuntu 18.04 has system-failures because virtual-env is missing (tracked by [BUILD-13263](https://jira.mongodb.org/browse/BUILD-13262)).

src/libmongoc/tests/test-mongoc-write-commands.c
```diff
@@ -656,5 +656,6 @@ test_write_command_install (TestSuite *suite)
                       _test_invalid_wc_server_error,
                       NULL,
                       NULL,
-                      test_framework_skip_if_no_failpoint);
+                      test_framework_skip_if_no_failpoint,
+                      test_framework_skip_if_mongos);
```
benjirewis, line 660: This test was failing on latest against sharded configurations. Setting a failpoint for a write concern error and error code 91 does not work as expected with sharded clusters; the error code returned is 6 ([HostUnreachable](https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml#L42)) instead of 91 (see CDRIVER-4022 for the full failure message). 

I don't know exactly why this is happening, but I've disabled the test on sharded clusters which seems to fix quite a few failures.

rcsanchez97 (approved):
LGTM.  (We will need to follow-up and ensure that the BUILD tickets that have been filed end up fixing the failing tasks that they are intended to fix.)

build/evergreen_config_lib/variants.py
```diff
@@ -179,15 +189,10 @@ def days(n):
              '.debug-compile .stdflags !.c89',
              '.debug-compile !.sspi .openssl',
              '.debug-compile !.sspi .nossl',
-             '.authentication-tests .openssl',
-             '.authentication-tests .valgrind',
```
kevinAlbs, line 183: The `latest` tasks make sense to remove on this Ubuntu 16.04 variant, since there are no longer latest server builds published for Ubuntu 16.04.

The `authentication-tests` use the run-auth-tests.sh script, which connect to external servers (e.g. Atlas dev instances). I think those should ideally be passing, but they are not on the waterfall. Unless we know why, I would vote we leave them to investigate in case it indicates a bug.

build/evergreen_config_lib/variants.py
```diff
@@ -179,15 +189,10 @@ def days(n):
...
-             '.latest .openssl !.nosasl .server',
-             '.latest .nossl',
-             '.4.4 .openssl !.nosasl .server',
-             '.4.2 .openssl !.nosasl .server',
-             '.4.0 .openssl !.nosasl .server',
```
kevinAlbs, line 188: Why are 4.0, 4.2, and 4.4 tasks removed?

build/evergreen_config_lib/variants.py
```diff
@@ -313,6 +317,21 @@ def days(n):
              'test-dns-auth-openssl'
              ],
             {'CC': 'gcc'}),
+    Variant('gcc54',
```
kevinAlbs, line 320: Nice, testing compilation and running against one live server on this variant seems like a good idea so we validate the C driver continues to compile against GCC 5.4.

kevinAlbs (commented):
Looking great! Thank you for all of the investigative work and tracking of remaining task failures. Here are my inline responses.

> 1. A few `test-latest-sharded` tasks are failing (on ASAN Ubuntu 18.04, clang 6.0 (i386) Ubuntu 18.04, GCC 7.5 (i386) Ubuntu 18.04, GCC 7.5 Ubuntu 18.04, and power8 RHEL 8.1) because of [`_test_invalid_wc_server_error`](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/tests/test-mongoc-write-commands.c#L600). The test expects an error code of [91](https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml#L126) but receives an error code of [6](https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml#L42).

Tracked in [SERVER-57772](https://jira.mongodb.org/browse/SERVER-57772).

> 2. `rpm-package-build` is failing because it cannot find `/home/ec2-user/rpmbuild/SOURCES/mongo-c-driver-1.19.0.tar.gz` when building.

I am not sure how this is related to the current PR. `rpm-package-build` runs on `rhel80-test`, and that has not changed in this PR. If this fails on the waterfall we can track this in a CDRIVER ticket.

> 3. `test-3.4-server-auth-sasl-winssl` [fails](https://evergreen.mongodb.com/lobster/evergreen/task/mongo_c_driver_windows_2015_test_3.4_server_auth_sasl_winssl_patch_0805656817fea6c0a0fee91e6928e2b9333f9564_60bf95d87742ae33e161b9b7_21_06_08_16_07_53/0/task#bookmarks=0%2C4767&l=1&shareLine=2397) [`test_tailable_alive`](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/tests/test-mongoc-cursor.c#L1729) because the cursor is unexpectedly empty.

This is a known `flaky-test`, tracked in [CDRIVER-3567](https://jira.mongodb.org/browse/CDRIVER-3567).

> 4. `debug-compile` tasks fail on power8 RHEL 8.1 because OPENSSL cannot be found.

The `debug-compile-*-openssl-static` OpenSSL tasks on `Power8 (ppc64le) (RHEL 8.1)` apear to be failing to find static OpenSSL libraries. I filed [BUILD-13324](https://jira.mongodb.org/browse/BUILD-13324).

For the remaining failing tasks that are waiting on BUILD tickets, we can file CDRIVER tickets with the label `failing-on-waterfall` to make those easier to track on our end.

kevinAlbs (approved):
LGTM!

rcsanchez97 (approved):
LGTM

kevinAlbs (approved):
LGTM! I ran [this patch build](https://spruce.mongodb.com/version/60d37561c9ec4422045984c1/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&taskName=debug-compile) with this PR along with a revert of `d2617271`. It shows the mock server tests consistently passing. Let's proceed with this, and address the mock server tests separately.

PR #798, opened 2021-05-26
CDRIVER-3964 Append session to all operations in UTR

benjirewis:
CDRIVER-3964

Appends session to all possible operations in the unified test runner.

Not all operations seem to take the `session` option. I noticed that `operation_list_databases` and `bulk_op_append` cause errors if `session` is appended, and the `operation_find_one_and_*` operations cannot take a session option since they use `mongoc_find_and_modify_opts_t *` instead of `bson_t *` as their options param. Let me know if you think I missed an operation/some operation does not need the `session` option.


jmikola:
> I noticed that operation_list_databases and bulk_op_append cause errors if session is appended

[`mongoc_client_get_database_names_with_opts`](http://mongoc.org/libmongoc/current/mongoc_client_get_database_names_with_opts.html) is documented as taking a `sessionId` parameter, so it should certainly work with `mongoc_client_session_append`. If not, that sounds like a bug.

As for bulk writes, the session option should be added directly to the `mongoc_bulk_operation_t` object via [`mongoc_bulk_operation_set_client_session`](http://mongoc.org/libmongoc/current/mongoc_bulk_operation_set_client_session.html).

> and the operation_find_one_and_* operations cannot take a session option since they use mongoc_find_and_modify_opts_t * instead of bson_t * as their options param.

This looks like an oversight and should be reported as a new ticket. One possible workaround would be to forgo [`mongoc_collection_find_and_modify_with_opts`](http://mongoc.org/libmongoc/current/mongoc_collection_find_and_modify_with_opts.html) in favor of executing a raw command with `mongoc_database_read_write_command_with_opts`; however, that would defeat the point of the test, which is to exercise the actual API.

@kevinAlbs Is this something worth addressing for the 5.0-compat release? Otherwise, I suppose we'll need to skip any tests that use `findAndModify` with a session.

src/libmongoc/tests/unified/operation.c
```diff
@@ -499,6 +547,7 @@ operation_bulk_write (test_t *test,
    bson_error_t op_error = {0};
 
    parser = bson_parser_new ();
+   bson_parser_allow_extra (parser, true);
```
jmikola, line 550: Was `allow_extra` actually needed here? I don't see `bson_parser_get_extra` called in this function.

src/libmongoc/tests/unified/operation.c
```diff
@@ -764,7 +847,7 @@ operation_distinct (test_t *test,
                         "{",
                         &filter,
                         "}");
-   bson_concat (distinct, bson_parser_get_extra (parser));
+   bson_concat (distinct, opts);
```
jmikola, line 850: I don't think this works. `mongoc_client_session_append` is going to append a "sessionId" to `opts`, but that's only the internal libmongoc identifier and should not end up in the raw command document being sent to the server.

This test currently uses [`mongoc_collection_command_simple`](http://mongoc.org/libmongoc/current/mongoc_collection_command_simple.html), which doesn't support sessions. I think you'll need to change it to use [`mongoc_database_read_command_with_opts`](http://mongoc.org/libmongoc/current/mongoc_database_read_command_with_opts.html) instead. And in doing so, you should pass `opts` as its own parameter (removing `bson_concat` entirely).

FWIW, `operation_distinct` appears to be the only operation to use `mongoc_collection_command_simple` so I don't think this is a concern anywhere else.

src/libmongoc/tests/unified/operation.c
```diff
@@ -1897,17 +2015,26 @@ assert_collection_exists (test_t *test,
...
+   bson_parser_allow_extra (bp, true);
    bson_parser_utf8 (bp, "collectionName", &collection_name);
    bson_parser_utf8 (bp, "databaseName", &database_name);
    if (!bson_parser_parse (bp, op->arguments, error)) {
       goto done;
    }
 
+   opts = bson_copy (bson_parser_get_extra (bp));
+   if (op->session) {
+      if (!mongoc_client_session_append (op->session, opts, error)) {
```
jmikola, line 2030: What was the reason for adding this to `assert_collection_exists`? [`assertCollectionExists`](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#assertcollectionexists) doesn't take a session.

src/libmongoc/tests/unified/operation.c
```diff
@@ -2019,20 +2147,29 @@ assert_index_exists (test_t *test,
...
    bson_parser_utf8 (bp, "collectionName", &collection_name);
    bson_parser_utf8 (bp, "databaseName", &database_name);
    bson_parser_utf8 (bp, "indexName", &index_name);
    if (!bson_parser_parse (bp, op->arguments, error)) {
       goto done;
    }
 
+   opts = bson_copy (bson_parser_get_extra (bp));
+   if (op->session) {
+      if (!mongoc_client_session_append (op->session, opts, error)) {
```
jmikola, line 2163: Likewise, [`assertIndexExists`](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#assertindexexists) does not take a session.

src/libmongoc/tests/unified/operation.c
```diff
@@ -1546,13 +1656,20 @@ operation_upload (test_t *test,
...
```
jmikola, line 1661: GridFS does not support sessions. See [SPEC-921](https://jira.mongodb.org/browse/SPEC-921) and [DRIVERS-559](https://jira.mongodb.org/browse/DRIVERS-559).

As a rule of thumb, I think you should double-check that `mongoc_client_session_append` is not being introduced for any methods that aren't documented as taking a "sessionId" parameter. In this case, [`mongoc_gridfs_bucket_open_upload_stream`](http://mongoc.org/libmongoc/current/mongoc_gridfs_bucket_open_upload_stream.html) makes no mention of sessions.

src/libmongoc/tests/unified/operation.c
```diff
@@ -87,23 +88,27 @@ operation_create_change_stream (test_t *test,
       goto done;
    }
 
+   opts = bson_copy (bson_parser_get_extra (parser));
+   if (op->session) {
```
jmikola, line 92: Since I'm not familiar with libmongo's test runner, I want to confirm that `op->session` here is the `mongoc_client_session_t` that has already been dereferenced from the "session" entity ID in the operation arguments. Is that correct?

That is my understanding from reading the "Check for a "session" argument in all operations..." section in `operation_run`.

benjirewis: That was my understanding as well. 

From the "session" argument parsing logic [here](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/tests/unified/operation.c#L2321-L2341), it looks like `op->session_id` is set to the "session" entity ID in the operation arguments, and then `op->session` is set to `entity_map_get_session` of that `session_id`.

src/libmongoc/tests/unified/operation.c
```diff
@@ -242,19 +248,33 @@ operation_list_collections (test_t *test,
    bool ret = false;
    mongoc_database_t *db = NULL;
    mongoc_cursor_t *cursor = NULL;
+   bson_parser_t *parser = NULL;
+   bson_t *opts = NULL;
+
+   parser = bson_parser_new ();
```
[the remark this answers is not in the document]
benjirewis: Ah you're right, and there were multiple leaks locally from not calling `bson_parser_destroy_with_parsed_fields`. Added here, in `operation_list_databases`, and in `operation_list_collection_names` (where I added a `parser`).

It looks like you should use `bson_parser_destroy_with_parsed_fields` whenever the parser has actually been used. That way, the value (`entry->out`) of each entry in the parser is also destroyed.

src/libmongoc/tests/unified/operation.c
benjirewis: Nope! Removed.

src/libmongoc/tests/unified/operation.c
```diff
@@ -514,6 +563,11 @@ operation_bulk_write (test_t *test,
    if (ordered) {
       BSON_APPEND_BOOL (opts, "ordered", *ordered);
    }
+   if (op->session) {
+      if (!mongoc_client_session_append (op->session, opts, error)) {
```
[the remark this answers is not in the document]
benjirewis: Gotcha, looks like as long as the session is being appended in `mongoc_collection_create_bulk_operation_with_opts`, I don't need to additionally append it in `bulk_op_append`.

The `BulkWrite with mixed ordered operations` spec test fails when I append a session in `operation_bulk_write` _and_ `bulk_op_append`. The test output is not very enlightening, but it might be because two fields with `sessionId` are included in a single operation?

src/libmongoc/tests/unified/operation.c
benjirewis: Gotcha, thanks for the guidance. I believe you meant `mongoc_collection_read_command_with_opts`, so I switched to that with a separate `opts` parameter and no `bson_concat`.

src/libmongoc/tests/unified/operation.c
benjirewis: Yeah... I should have taken a look at the documentation for each. Removed the session appension here, in `assert_collection_exists` and `assert_index_exists`.

src/libmongoc/tests/unified/operation.c
benjirewis: Removed.

src/libmongoc/tests/unified/operation.c
benjirewis: Removed.

benjirewis (commented):
Appended session to `operation_list_databases`; I just needed to make sure `op->arguments` wasn't NULL before appending it to `opts`.

Per the discussion [here](https://github.com/mongodb/mongo-c-driver/pull/798#discussion_r640079606), `bulk_op_append` does not need a session appension since it's already appended in `operation_bulk_write`.

The `operation_find_and_*` tests still have no session appension, but I'll file a ticket to add a session option to `mongoc_find_and_modify_opts_t`.

src/libmongoc/tests/unified/operation.c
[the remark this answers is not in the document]
jmikola: The [BulkWrite with mixed ordered operations](https://github.com/mongodb/specifications/blob/5a8c8d774a0092f520a5ca5c2ef9b7b2dbca5f8a/source/unified-test-format/tests/valid-pass/poc-crud.yml#L53) test in `poc-crud.yml` doesn't use a session, so I'm not sure why you would encounter _any_ failure appending a session. What "two fields with `sessionId`" are you referring to?

src/libmongoc/tests/unified/operation.c
jmikola: I was definitely referring to `mongoc_database_read_command_with_opts`. I forgot there were similarly named collection methods, but if you look at both functions you'll see they're just syntactic sugar for `_mongoc_client_command_with_opts`.

In my mind, commands are scoped to a database, not a collection, so I don't think about having a helper at the collection level. The command itself _may_ refer to a collection name (e.g. `find: <name>`), but that's still something you'd have to manually specify in the command document.

Anywho, either method should work. And since `distinct` is a collection-level operation, it makes sense that you'll already have the necessary collection object on hand here.

src/libmongoc/tests/unified/operation.c
```diff
@@ -141,19 +141,36 @@ operation_list_databases (test_t *test,
...
+   bson_parser_allow_extra (parser, true);
+
+   opts = bson_copy (bson_parser_get_extra (parser));
+   if (op->session) {
+      if (!mongoc_client_session_append (op->session, opts, error)) {
+         goto done;
+      }
+   }
+   if (op->arguments) {
+      bson_concat (opts, op->arguments);
```
jmikola, line 157: Note: I think it makes sense for libmongoc to have a macro here to assert that `arguments` does not contain any unexpected keys before we end up passing it along to the API method's `opts` parameter (or concatenating it in this case).

As-is, it's quite possible that a spec test might introduce a new option that gets silently ignored by libmongoc (because it has yet to implement the functionality) and the test might still pass.

I do this in PHPLIB's test runner in a few places, but I don't believe it's done for method arguments. Going to make a ticket to track that todo item, but I'd suggest creating one for CDRIVER as well if @kevinAlbs agrees.

src/libmongoc/tests/unified/operation.c
```diff
@@ -242,19 +265,34 @@ operation_list_collections (test_t *test,
    bool ret = false;
    mongoc_database_t *db = NULL;
    mongoc_cursor_t *cursor = NULL;
+   bson_parser_t *parser = NULL;
```
jmikola, line 268: I noticed this PR is still missing some operations. For instance, `operation_drop_collection` was not modified. This is likely because a variant of `mongoc_collection_drop` was never introduced to take a "sessionId" option. Since you're going through the entire file here, would you mind creating a list of all of the operations that still _do not_ take a session operation (excluding things like the internal assertion and special test operations) and dump that list into a CDRIVER ticket? I think that would provide a good starting point for any future plans to create the necessary APIs so that all session-supporting operations _can_ solicit a "sessionId" option.

kevinAlbs:
> This looks like an oversight and should be reported as a new ticket. One possible workaround would be to forgo mongoc_collection_find_and_modify_with_opts in favor of executing a raw command with mongoc_database_read_write_command_with_opts; however, that would defeat the point of the test, which is to exercise the actual API.

> @kevinAlbs Is this something worth addressing for the 5.0-compat release? Otherwise, I suppose we'll need to skip any tests that use findAndModify with a session.

It's funky, but session options can be passed by first appending the session with `mongoc_client_session_append` to a `bson_t`. Then append the `bson_t` opts to the `mongoc_find_and_modify_opts_t` via `mongoc_find_and_modify_opts_append` ([example](https://github.com/kevinAlbs/mongo-c-driver/blob/30cb80dae2be30f8b62ab4069b994a9b7e365ed5/src/libmongoc/tests/test-mongoc-client-session.c#L1896:L1910)). This is documented in [mongoc_find_and_modify_opts_append](http://mongoc.org/libmongoc/current/mongoc_find_and_modify_opts_append.html#description)

src/libmongoc/tests/unified/operation.c
kevinAlbs: Extra keys can be checked with the `bson_parser_t`. By default will error on extra keys unless `bson_parser_allow_extra` is called. Some tests, but not all, already check the arguments and error on extra keys. Example from `operation_bulk_write`:

```
   parser = bson_parser_new ();
   bson_parser_array (parser, "requests", &requests);
   bson_parser_bool_optional (parser, "ordered", &ordered);
   if (!bson_parser_parse (parser, op->arguments, error)) {
      goto done;
   }
```

I agree it is preferable to error on unrecognized operation arguments. I think a minor follow-up ticket to make all cases explicit seems reasonable. But it seems low risk to me. If we do ignore an argument, a test should ideally fail.

src/libmongoc/tests/unified/operation.c
```diff
@@ -135,19 +141,36 @@ operation_list_databases (test_t *test,
    bool ret = false;
    mongoc_client_t *client = NULL;
    mongoc_cursor_t *cursor = NULL;
+   bson_parser_t *parser = NULL;
+   bson_t *opts = NULL;
+
+   parser = bson_parser_new ();
+   bson_parser_allow_extra (parser, true);
+
+   opts = bson_copy (bson_parser_get_extra (parser));
```
kevinAlbs, line 150: `parser` is not parsing any input BSON. This can be simplified to `opts = bson_new ()`.

src/libmongoc/tests/unified/operation.c
```diff
@@ -242,19 +265,34 @@ operation_list_collections (test_t *test,
    bool ret = false;
    mongoc_database_t *db = NULL;
    mongoc_cursor_t *cursor = NULL;
+   bson_parser_t *parser = NULL;
+   bson_t *opts = NULL;
+
+   parser = bson_parser_new ();
+   bson_parser_allow_extra (parser, true);
+
+   opts = bson_copy (bson_parser_get_extra (parser));
```
kevinAlbs, line 274: Similarly, since the parser is not parsing any input, this could be simplified to `opts = bson_new()`.

src/libmongoc/tests/unified/operation.c
```diff
@@ -269,21 +307,36 @@ operation_list_collection_names (test_t *test,
    mongoc_cursor_t *cursor = NULL;
    char **op_ret = NULL;
    bson_error_t op_error = {0};
+   bson_parser_t *parser = NULL;
+   bson_t *opts = NULL;
+
+   parser = bson_parser_new ();
+   bson_parser_allow_extra (parser, true);
+
+   opts = bson_copy (bson_parser_get_extra (parser));
```
kevinAlbs, line 316: Same comment about `opts = bson_new()`.

src/libmongoc/tests/unified/operation.c
```diff
@@ -303,9 +356,10 @@ operation_run_command (test_t *test,
    mongoc_read_concern_t *rc = NULL;
    bson_error_t op_error = {0};
    bson_t op_reply = BSON_INITIALIZER;
-   bson_t opts = BSON_INITIALIZER;
+   bson_t *opts = NULL;
 
    parser = bson_parser_new ();
+   bson_parser_allow_extra (parser, true);
```
kevinAlbs, line 362: Unless this is needed, I suggest removing the call to `bson_parser_allow_extra`. `mongoc_client_session_append` can be called on `opts` directly without calling `bson_parser_get_extra`. Since this was validating that no extra keys were included in the arguments before, `bson_parser_get_extra` should be an empty document.

src/libmongoc/tests/unified/operation.c
benjirewis: Ah gotcha; no need to add extra parsing if there wasn't any in the first place! Simplified to `opts = bson_new ()` and removed the `bson_parser_allow_extra`.

src/libmongoc/tests/unified/operation.c
benjirewis: Filed a follow-up ticket to ensure explicit cases for all arguments to all operations as in the example above (CDRIVER-4018).

src/libmongoc/tests/unified/operation.c
benjirewis: So I switched to using `mongoc_collection_drop_with_opts` which _does_ [take a session ID](http://mongoc.org/libmongoc/current/mongoc_collection_drop_with_opts.html). After changing that, and adding sessionId to all the `find_and_modify` operations, I think we're pretty close to everything. I don't think there are any operations represented here that are session-supporting but _cannot_ solicit a "sessionId" option.

The few that I'm wondering about are: `operation_failpoint`, `assert_index_exists`, and `assert_collection_exists`. I _can_ append a sessionId to all of these since they use `mongoc_client_command_simple`, [mongoc_collection_find_indexes_with_opts](http://mongoc.org/libmongoc/current/mongoc_collection_find_indexes_with_opts.html) and [mongoc_database_get_collection_names_with_opts](http://mongoc.org/libmongoc/current/mongoc_collection_find_indexes_with_opts.html) respectively.

However, I don't see any usages of `session` with those three operations in existing unified format tests. Thoughts?

src/libmongoc/tests/unified/operation.c
benjirewis: Simplified to `opts = bson_new()` and removed `allow_extra` call.

src/libmongoc/tests/unified/operation.c
benjirewis: Simplified to `opts = bson_new()` and removed `allow_extra` call.

src/libmongoc/tests/unified/operation.c
benjirewis: Removed!

src/libmongoc/tests/unified/operation.c
benjirewis: Ah gotcha! I'll just use the `mongoc_collection_read_command_with_opts` since I have the collection handle already.

src/libmongoc/tests/unified/operation.c
jmikola: Was this change not pushed? I still see `parser = bson_parser_new ();` and `bson_parser_allow_extra (parser, true);` here.

src/libmongoc/tests/unified/operation.c
jmikola: I don't see `mongoc_collection_drop_with_opts` in this PR. Was that not pushed?

If a [special test operation](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#special-test-operations) doesn't document a session option, those should not support a session at all.

src/libmongoc/tests/unified/operation.c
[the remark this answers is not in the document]
jmikola: Ping re: my last question.

src/libmongoc/tests/unified/operation.c
benjirewis: Ah 🤦 forgot to push. Should be all set now.

And great thanks for linking me to the spec there; it looks like none of `failPoint`, `assertIndex(Not)Exists` or `assertCollection(Not)Exists` are documented as taking a session option, so I will not add sessionId to those operations.

src/libmongoc/tests/unified/operation.c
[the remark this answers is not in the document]
benjirewis: TL;DR the "two fields with `sessionId`" was definitely not the reason for the failure. As you pointed out, the `BulkWrite` test in `poc-crud` doesn't even use a session...

To be clear, I'm no longer appending a session at all in `bulk_op_append` since it should already be added in `operation_bulk_write`.

I'm not familiar with the internals of `bson_parser_t`, but I believe the original error was because I had one line `opts = bson_parser_get_extra (parser)` [here](https://github.com/mongodb/mongo-c-driver/blob/724f1672f26018f0d23883595c212497a38b043b/src/libmongoc/tests/unified/operation.c#L413), before each of the types bulk operations. Then, I passed `opts` to the individual operations. It looks like each operation does its own [parsing](https://github.com/mongodb/mongo-c-driver/blob/724f1672f26018f0d23883595c212497a38b043b/src/libmongoc/tests/unified/operation.c#L416) before calling `bson_parser_get_extra (parser)`, so getting the extra fields before the operation-specific parsing meant that the operation was being parsed incorrectly. I think the malformed operation meant that the actual result was different than the expected.

src/libmongoc/tests/unified/operation.c
```diff
@@ -909,6 +1021,12 @@ operation_find_one_and_update (test_t *test,
    }
    mongoc_find_and_modify_opts_set_flags (opts, flags);
    mongoc_find_and_modify_opts_append (opts, bson_parser_get_extra (parser));
+   if (op->session) {
+      if (!mongoc_client_session_append (op->session, session_opts, error)) {
```
kevinAlbs, line 1025: `session_opts` is NULL. I think `session_opts = bson_new();` is needed above.

src/libmongoc/tests/unified/operation.c
```diff
@@ -1003,6 +1130,7 @@ operation_find_one_and_delete (test_t *test,
    mongoc_find_and_modify_flags_t flags = 0;
    bson_val_t *val = NULL;
    bson_iter_t iter;
+   bson_t *session_opts = NULL;
```
kevinAlbs, line 1133:
```suggestion
   bson_t *session_opts = bson_new();
```

src/libmongoc/tests/unified/operation.c
```diff
@@ -946,6 +1065,7 @@ operation_find_one_and_replace (test_t *test,
...
```
kevinAlbs, line 1068:
```suggestion
   bson_t *session_opts = bson_new();
```

src/libmongoc/tests/unified/operation.c
```diff
@@ -135,19 +141,35 @@ operation_list_databases (test_t *test,
    bool ret = false;
    mongoc_client_t *client = NULL;
    mongoc_cursor_t *cursor = NULL;
+   bson_parser_t *parser = NULL;
```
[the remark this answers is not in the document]
benjirewis: Ah good point, removed and removed all references to it in the function.

kevinAlbs (approved):
LGTM

src/libmongoc/tests/unified/operation.c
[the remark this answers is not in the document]
jmikola: Sounds like there's no longer anything to address here. I'll resolve.

src/libmongoc/tests/unified/operation.c
benjirewis: Good point. As it is, `operation_distinct` does not really test any C driver functionality (and I don't think it's documented that users should create their own distinct [in this manner](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/tests/unified/operation.c#L759-L766)).

The Go and CXX test runners also error on unsupported operations. I think it would be ok to skip `distinct` tests since the C driver has no helper, but always skipping on unsupported operations might make it harder to recognize when a new operation requires a new case in the test runner.

jmikola: I'd be OK with leaving this as-is and then opening a separate CDRIVER issue to follow-up. That issue can propose an explicit skip for tests utilizing `distinct` operations. Alternatively, you might decide to utilize the `is_test_skipped` function I'm adding in https://github.com/mongodb/mongo-c-driver/pull/801 and just explicitly add tests using `distinct` to that list.

Either way, I think it makes sense to figure out how we want to handle this _before_ the retryable read tests get ported and we need to skip its tests for unsupported operations (e.g. collection enumeration variants). In that case, the explicit list in `is_test_skipped` probably makes the most sense.

jmikola (approved):

src/libmongoc/tests/unified/operation.c
benjirewis: Filed CDRIVER-4024. Using something like `is_test_skipped` seems best to me; then we can explicitly list the tests with unsupported operations like `distinct` or collection enumeration variants. Leaving `operation_distinct` as-is for now, though.

PR #799, opened 2021-05-27
CDRIVER-3897 add tests for time-series collections

kevinAlbs:
Resyncs tests to include the new time-series spec tests and fixes a bug in the unified test runner's outcome matching. No code changes were required.

- `authentication-tests-openssl-nosasl` task failures are due to CDRIVER-3998
- `test-asan-latest-*` task failures will be resolved with the Ubuntu 18.04 upgrade to pull in latest server releases: CDRIVER-3946

### DIFF
--- a/src/libmongoc/tests/unified/runner.c
+++ b/src/libmongoc/tests/unified/runner.c
@@ -1153,7 +1153,7 @@ test_check_outcome_collection (test_t *test,
       expected_sorted = bson_copy_and_sort (&expected);
 
 
-      if (!bson_equal (&actual, &expected)) {
+      if (!bson_equal (actual_sorted, expected_sorted)) {
          test_set_error (error,
                          "expected %s, but got %s",
                          tmp_json (expected_sorted),
@@ -1493,4 +1493,6 @@ test_install_unified (TestSuite *suite)
    run_unified_tests (suite, JSON_DIR "/crud/unified");
 
    run_unified_tests (suite, JSON_DIR "/transactions/unified");
+
+   run_unified_tests (suite, JSON_DIR "/collection-management");
 }
--- a/src/libmongoc/doc/mongoc_database_create_collection.rst
+++ b/src/libmongoc/doc/mongoc_database_create_collection.rst
@@ -29,6 +29,8 @@ This function creates a :symbol:`mongoc_collection_t` from the given :symbol:`mo
 
 If no write concern is provided in ``opts``, the database's write concern is used.
 
+For a list of all options, see `the MongoDB Manual entry on the create command <https://docs.mongodb.org/manual/reference/command/create/>`_.
+
 Errors
 ------
 
--- a/src/libmongoc/tests/json/collection-management/timeseries-collection.json
+++ b/src/libmongoc/tests/json/collection-management/timeseries-collection.json
@@ -53,12 +53,11 @@
           "object": "database0",
           "arguments": {
             "collection": "test",
-            "expireAfterSeconds": {
-              "$numberLong": "604800"
-            },
+            "expireAfterSeconds": 604800,
             "timeseries": {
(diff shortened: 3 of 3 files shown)


src/libmongoc/tests/unified/runner.c
```diff
@@ -1153,7 +1153,7 @@ test_check_outcome_collection (test_t *test,
       expected_sorted = bson_copy_and_sort (&expected);
 
 
-      if (!bson_equal (&actual, &expected)) {
+      if (!bson_equal (actual_sorted, expected_sorted)) {
```
kevinAlbs, line 1156: This was a small bug fix, the outcome assertions are supposed to account for different order of keys.

https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#executing-a-test

> test runners MUST permit variations in document key order or otherwise normalize the documents before comparison.

benjirewis (commented):
Should we do some parameter validation for the new time-series options (i.e. ensure they are strings/ints as done for the other options [here](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/mongoc-database.c#L883-L971))?

kevinAlbs:
> Should we do some parameter validation for the new time-series options (i.e. ensure they are strings/ints as done for the other options here)?

Good question. IMO if the server errors for the wrong type, the existing driver-side checks do not provide value. I would rather remove the existing checks.

If the user currently uses the wrong type, they will get a `TypeMismatch` error from a 5.0 server for any of the new options (`timeseries.timeField`, `timeseries.metaField`, `granularity`, `expireAfterSeconds`).

```
MongoDB Enterprise rs0:PRIMARY> db.runCommand({create: "tcoll", timeseries: {timeField: 123}})
{
	"ok" : 0,
	"errmsg" : "BSON field 'create.timeseries.timeField' is the wrong type 'double', expected type 'string'",
	"code" : 14,
	"codeName" : "TypeMismatch",
	"$clusterTime" : {
		"clusterTime" : Timestamp(1622819137, 1),
		"signature" : {
			"hash" : BinData(0,"AAAAAAAAAAAAAAAAAAAAAAAAAAA="),
			"keyId" : NumberLong(0)
		}
	},
	"operationTime" : Timestamp(1622819137, 1)
}
```

This also aligns with the [Where possible, depend on server to return errors mantra](https://github.com/mongodb/specifications#where-possible-depend-on-server-to-return-errors). Relying on the server for validation helps with forward-compatibility. If the server later relaxes constraints, (e.g. if granularity later permits a numeric value, there are no required changes).

benjirewis (approved):
LGTM then; enforcing our own types in the test runner does seem harder and less stable than just relying on the server. Is there any ticket to track the work of removing type checks within the test runners?

kevinAlbs:
> LGTM then; enforcing our own types in the test runner does seem harder and less stable than just relying on the server. Is there any ticket to track the work of removing type checks within the test runners?

Ah, I was referring to the driver-side checks in the implementation, not the test runner. For the implementation, I think we should default to relying on server-side checks, and only add driver-side checks in cases we cannot rely on the server to error. For example – some CRUD options, like `UpdateOptions.hint` have associated wire version checks. The rationale is that the user will otherwise be uninformed if the server does not support the option: https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#:~:text=why%20are%20client-side%20errors%20raised%20for%20some%20unsupported%20options%3F.

I filed CDRIVER-4015 to cover relaxing all such client-side checks.

For the test runner type checks. The same logic applies for any option that is going to be added to the command sent to the server. But for fields that do not represent server-side options, e.g. a collection name, the validation could give more helpful up-front errors. The bigger concern I have is accidentally ignoring fields in the spec test (e.g. bypassing assertions).

PR #801, opened 2021-06-08
CDRIVER-3895: Remove dots/dollars validation for insert/replace docs

jmikola:
https://jira.mongodb.org/browse/CDRIVER-3895

----

Also addresses:

 * https://jira.mongodb.org/browse/CDRIVER-3960


src/libmongoc/tests/json/unified/poc-command-monitoring.json
```diff
@@ -149,7 +149,7 @@
                     ]
                   },
                   "collection": "test",
-                  "batchSize": 3
+                  "batchSize": 1
```
jmikola, line 152: This change reverts https://github.com/mongodb/mongo-c-driver/pull/793, which manually modified the test file to work around a deviation in libmongoc's behavior.

Since the original test fails, this is skipped in `runner.c`. It's not clear to me if this pertains more to [DRIVERS-1781](https://jira.mongodb.org/browse/DRIVERS-1781) or [DRIVERS-1448](https://jira.mongodb.org/browse/DRIVERS-1448), so I'd appreciate @alcaeus's input here. I did confirm that this test fails with _both_ MongoDB 4.4 and 5.0, so I don't think it's related to the slot-based execution changes in 5.0.

src/libmongoc/tests/unified/runner.c
```diff
@@ -26,6 +26,86 @@
 #include "util.h"
 
 
+typedef struct {
+   const char *file_description;
+   const char *test_description;
+} skipped_unified_test_t;
+
+#define SKIP_ALL_TESTS NULL
```
jmikola, line 34: While I could have used this for `unacknowledged-bulkWrite-delete-hint-clientError` and related tests, I decided not to as we know exactly what tests should be skipped there. This was really only necessary for cases where we can't even parse tests from a file due to an unsupported schema version.

src/libmongoc/tests/unified/runner.c
```diff
@@ -26,6 +26,86 @@
...
+   {"unacknowledged-findOneAndUpdate-hint-clientError", "Unacknowledged findOneAndUpdate with hint document fails with client-side error"},
+   {"unacknowledged-replaceOne-hint-clientError", "Unacknowledged ReplaceOne with hint string fails with client-side error"},
+   {"unacknowledged-replaceOne-hint-clientError", "Unacknowledged ReplaceOne with hint document fails with client-side error"},
+   {"unacknowledged-updateMany-hint-clientError", "Unacknowledged updateMany with hint string fails with client-side error"},
+   {"unacknowledged-updateMany-hint-clientError", "Unacknowledged updateMany with hint document fails with client-side error"},
+   {"unacknowledged-updateOne-hint-clientError", "Unacknowledged updateOne with hint string fails with client-side error"},
+   {"unacknowledged-updateOne-hint-clientError", "Unacknowledged updateOne with hint document fails with client-side error"},
+   /* CDRIVER-4001, DRIVERS-1781, and DRIVERS-1448: 5.0 cursor behavior */
+   {"poc-command-monitoring", "A successful find event with a getmore and the server kills the cursor"},
```
jmikola, line 62: This pertains to my earlier comment about reverting #793 and skipping the original test from the specs repo.

src/libmongoc/tests/unified/result.c
```diff
@@ -150,6 +150,11 @@ result_from_bulk_write (result_t *result, bson_t *reply, bson_error_t *error)
          rewritten_upserted_ids = rewrite_upserted_ids (upserted_ids);
          BSON_APPEND_DOCUMENT (
             write_result, "upsertedIds", rewritten_upserted_ids);
+      } else {
+         /* upsertedIds is a required field in BulkWriteResult, so append an
+          * empty document even if not documents were upserted. */
+         upserted_ids = bson_new ();
+         BSON_APPEND_DOCUMENT (write_result, "upsertedIds", upserted_ids);
```
jmikola, line 157: This appears to have been an oversight in the original test runner implementation.

src/libmongoc/tests/unified/operation.c
```diff
@@ -476,63 +476,53 @@ bulk_op_append (mongoc_bulk_operation_t *bulk,
       if (!bson_parser_parse (parser, &request_doc, error)) {
          goto done;
       }
-      if (!mongoc_bulk_operation_insert_with_opts (
-             bulk, document, bson_parser_get_extra (parser), error)) {
-         goto done;
-      }
+
+      mongoc_bulk_operation_insert_with_opts (
+         bulk, document, bson_parser_get_extra (parser), error);
```
jmikola, line 481: The previous code made it impossible to evaluate `expectError` for these methods, since any error interrupted the test runner's execution (like a parsing error or unsupported operation). This change is consistent with how other public API methods are invoked.

src/libmongoc/tests/test-mongoc-write-commands.c
```diff
@@ -576,6 +576,132 @@ test_w0_legacy_insert_many (void)
    mock_server_destroy (server);
 }
 
+static void
+test_w0_legacy_update_one (void)
```
jmikola, line 580: OP_UPDATE previously had no test coverage. This method is based after the earlier OP_INSERT test, and I added another test below to ensure update/replace validation still occurs (via `mongoc-util.c` since we removed it from the legacy write methods).

src/libmongoc/tests/test-mongoc-collection.c
```diff
@@ -559,15 +559,7 @@ test_insert (void)
...
-      collection, tmp_bson ("{'$hello': 1}"), NULL, NULL, &error);
-   ASSERT (!r);
-   ASSERT_ERROR_CONTAINS (error,
-                          MONGOC_ERROR_COMMAND,
-                          MONGOC_ERROR_COMMAND_INVALID_ARG,
-                          "invalid document");
-
-   r = mongoc_collection_insert_one (
-      collection, tmp_bson ("{'a.b': 1}"), NULL, NULL, &error);
+      collection, tmp_bson ("{'': 1}"), NULL, NULL, &error);
```
jmikola, line 562: Empty keys are one of the few client-side validation errors we can still rely upon. Invalid UTF-8 characters would also work in theory, but `tmp_bson` refuses to parse JSON with such characters.

src/libmongoc/tests/test-mongoc-collection.c
```diff
@@ -3946,14 +3921,14 @@ _test_insert_validate (insert_fn_t insert_fn)
...
+   BSON_ASSERT (!insert_fn (collection, tmp_bson ("{'': 1}"), NULL, &error));
    ASSERT_ERROR_CONTAINS (error,
                           MONGOC_ERROR_COMMAND,
                           MONGOC_ERROR_COMMAND_INVALID_ARG,
-                          "invalid document");
+                          "empty key");
 
    BSON_ASSERT (!insert_fn (collection,
-                            tmp_bson ("{'$': 1}"),
+                            tmp_bson ("{'_id': {'$a': 1}}"),
```
jmikola, line 3931: A dollar-prefixed field within an `_id` is one of the few server-side validation errors we can still rely upon regardless of server version. This is also used in the spec tests themselves.

src/libmongoc/tests/test-mongoc-collection.c
```diff
@@ -6674,24 +6652,10 @@ test_collection_install (TestSuite *suite)
...
-      test_insert_many_validate,
-      NULL,
-      NULL,
-      TestSuite_CheckLive,
-      /* TODO: remove checks when removing validation for dots and dollars */
-      test_framework_skip_if_max_wire_version_more_than_9);
+   TestSuite_AddLive (
+      suite, "/Collection/insert_one_validate", test_insert_one_validate);
+   TestSuite_AddLive (
+      suite, "/Collection/insert_many_validate", test_insert_many_validate);
```
jmikola, line 6658: These changes revert https://github.com/mongodb/mongo-c-driver/pull/795

src/libmongoc/tests/test-mongoc-bulk.c
```diff
@@ -476,30 +476,13 @@ test_insert_check_keys (void)
...
-   bulk = mongoc_collection_create_bulk_operation_with_opts (collection, NULL);
-   BSON_ASSERT (bulk);
-
-   mongoc_bulk_operation_insert (bulk, tmp_bson ("{'a.b': 1}"));
-   r = (bool) mongoc_bulk_operation_execute (bulk, &reply, &error);
-   BSON_ASSERT (!r);
-   ASSERT_ERROR_CONTAINS (error,
-                          MONGOC_ERROR_COMMAND,
-                          MONGOC_ERROR_COMMAND_INVALID_ARG,
-                          "keys cannot contain \".\": \"a.b\"");
```
jmikola, line 502: There was nothing to replace this test with, since we already use empty keys earlier and invalid UTF-8 isn't an option (due to how `tmp_bson` parses JSON).

src/libmongoc/src/mongoc/mongoc-util.c
```diff
@@ -28,13 +28,11 @@
...
    BSON_VALIDATE_UTF8 | BSON_VALIDATE_UTF8_ALLOW_NULL |
-   BSON_VALIDATE_EMPTY_KEYS | BSON_VALIDATE_DOT_KEYS |
-   BSON_VALIDATE_DOLLAR_KEYS;
+   BSON_VALIDATE_EMPTY_KEYS;
 
 const bson_validate_flags_t _mongoc_default_replace_vflags =
    BSON_VALIDATE_UTF8 | BSON_VALIDATE_UTF8_ALLOW_NULL |
-   BSON_VALIDATE_EMPTY_KEYS | BSON_VALIDATE_DOT_KEYS |
-   BSON_VALIDATE_DOLLAR_KEYS;
+   BSON_VALIDATE_EMPTY_KEYS;
```
jmikola, line 35: With this change all three vflags are identical. The vflags aren't part of the public API, so we could have consolidated them into one value; however, I think leaving them as-is is fine.

src/libmongoc/src/mongoc/mongoc-util.c
```diff
@@ -338,6 +338,27 @@ _mongoc_validate_replace (const bson_t *doc,
...
+   }
+
+   while (bson_iter_next (&iter)) {
+      key = bson_iter_key (&iter);
+      if (key[0] == '$') {
+         bson_set_error (error,
+                         MONGOC_ERROR_COMMAND,
+                         MONGOC_ERROR_COMMAND_INVALID_ARG,
+                         "Invalid key '%s': replace prohibits $ operators",
+                         key);
```
jmikola, line 356: This is basically the inverse of the existing logic in `_mongoc_validate_update`. We technically only need to check the first key since the server would raise an error (as [discussed in the CRUD spec](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#update-vs-replace-validation)), but I opted to keep this the same as `update` validation. If anyone feels strongly about this, feel free to open a new CDRIVER ticket to propose a future change to both validation methods.

src/libmongoc/src/mongoc/mongoc-write-command-legacy.c
```diff
@@ -406,45 +403,20 @@ _mongoc_write_command_update_legacy (mongoc_write_command_t *command,
...
    reader =
       bson_reader_new_from_data (command->payload.data, command->payload.len);
    while ((bson = bson_reader_read (reader, &eof))) {
+      /* ensure the document has "q" and "u" document fields */
+      r = (bson_iter_init (&subiter, bson) && bson_iter_find (&subiter, "q") &&
+           BSON_ITER_HOLDS_DOCUMENT (&subiter) &&
+           bson_iter_find (&subiter, "u") &&
+           BSON_ITER_HOLDS_DOCUMENT (&subiter));
+
+      BSON_ASSERT (r);
```
jmikola, line 418: Checking that both fields exist is necessary since we don't explicitly fetch the fields later. For whatever reason, `_mongoc_write_command_update_legacy` iterates through all fields and could fail to fully construct the command if one was absent. In practice, I doubt that would ever happen since the BSON for this is constructed elsewhere in libmongoc and doesn't originate from the user.

Note: I opted not to test for this assert failure, since it really isn't possible to trigger it from the public API.

src/libmongoc/tests/bsonutil/bson-parser.c
```diff
@@ -27,6 +27,7 @@ typedef enum {
    BSON_PARSER_BOOL,
    BSON_PARSER_DOC,
    BSON_PARSER_ARRAY,
+   BSON_PARSER_ARRAY_OR_DOC,
```
jmikola, line 30: This was necessary for accepting documents or pipeline arrays for update operations.

jmikola:
Evergreen looks good apart from the [known failures](https://jira.mongodb.org/issues/?jql=labels%20%3D%20failing-on-waterfall). I think this is ready for a full review.

src/libmongoc/tests/unified/runner.c
```diff
@@ -26,6 +26,86 @@
...
+typedef struct {
+   const char *file_description;
+   const char *test_description;
+} skipped_unified_test_t;
+
+#define SKIP_ALL_TESTS NULL
+
+/* clang-format off */
+skipped_unified_test_t SKIPPED_TESTS[] = {
+   /* CDRIVER-3630: libmongoc does not unconditionally raise an error when using
```
kevinAlbs, line 38: Ah right, IIRC we only raise an error if the server version indicates that hint is not supported.

src/libmongoc/tests/unified/runner.c
kevinAlbs: Having one place to list explicitly skipped tests, with an option to skip all tests in one file is helpful.

kevinAlbs (approved):
LGTM

src/libmongoc/src/mongoc/mongoc-write-command-legacy.c
```diff
@@ -406,45 +403,19 @@ _mongoc_write_command_update_legacy (mongoc_write_command_t *command,
...
-      }
-   }
-
    ns = bson_strdup_printf ("%s.%s", database, collection);
 
-   bson_reader_destroy (reader);
    reader =
       bson_reader_new_from_data (command->payload.data, command->payload.len);
    while ((bson = bson_reader_read (reader, &eof))) {
+      /* ensure the document has "q" and "u" document fields */
```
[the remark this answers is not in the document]
jmikola: Ah, good point. `bson_iter_init_find` would be needed if we expected them to possibly appear out of order, which isn't a concern here.

src/libmongoc/src/mongoc/mongoc-write-command-legacy.c
```diff
@@ -459,6 +430,10 @@ _mongoc_write_command_update_legacy (mongoc_write_command_t *command,
       while (bson_iter_next (&subiter)) {
          if (strcmp (bson_iter_key (&subiter), "u") == 0) {
             bson_iter_document (&subiter, &len, &data);
+
+            BSON_ASSERT (data);
+            BSON_ASSERT (len >= 5);
```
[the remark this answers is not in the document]
jmikola: Shamelessly copied from `_mongoc_write_command_delete_legacy`.

src/libmongoc/tests/test-mongoc-collection.c
```diff
@@ -1332,16 +1315,6 @@ test_update_pipeline (void *ctx)
       collection, b, replacement, NULL, NULL, &error);
    ASSERT_OR_PRINT (res, error);
 
-   /* ensure that pipeline updates sent to mongoc_collection_replace_one
```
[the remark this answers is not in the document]
jmikola: Where do you propose we document that `{"0": {"$foo": "x"}` is now considered a replacement? Should I replace the removed code with a comment explaining why we're no longer testing this?

src/libmongoc/tests/unified/runner.c
```diff
@@ -26,6 +26,86 @@
...
 
+typedef struct {
+   const char *file_description;
+   const char *test_description;
+} skipped_unified_test_t;
+
+#define SKIP_ALL_TESTS NULL
+
+/* clang-format off */
+skipped_unified_test_t SKIPPED_TESTS[] = {
```
alcaeus, line 37: This is great, thank you for adding this!

alcaeus (approved):

benjirewis (approved):
LGTM (just some opt nits). Helpful for the Go implementation 🧑‍🔧

src/libmongoc/tests/test-mongoc-collection.c
```diff
@@ -5567,10 +5536,11 @@ _test_update_and_replace (bool is_replace, bool is_multi)
...
       ASSERT (!ret);
-      ASSERT_ERROR_CONTAINS (err,
-                             MONGOC_ERROR_COMMAND,
-                             MONGOC_ERROR_COMMAND_INVALID_ARG,
-                             "invalid argument for replace");
+      ASSERT_ERROR_CONTAINS (
+         err,
+         MONGOC_ERROR_COMMAND,
+         MONGOC_ERROR_COMMAND_INVALID_ARG,
+         "Invalid key '$set': replace prohibits $ operators");
```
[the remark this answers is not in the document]
jmikola: The previous error messages originated in libbson (see: [bson.c](https://github.com/mongodb/mongo-c-driver/blob/1.17.6/src/libbson/src/bson/bson.c#L3262)). The message here is consistent with what this file already does for update validation:

> Invalid key '%s': update only works with $ operators and pipelines

src/libmongoc/tests/test-mongoc-collection.c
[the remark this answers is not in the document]
kevinAlbs: Sorry, I was not suggesting any change. I should have prefixed my comment with "Note:". This seems like the correct behavior.

jmikola:
Cherry-picked to r1.18 in https://github.com/mongodb/mongo-c-driver/commit/4069b7984d3b089195d4c7ff1fd58c3d6bc2aedc

PR #802, opened 2021-06-16
CDRIVER-3892 expose reason documents fail validation

jmikola:
https://jira.mongodb.org/browse/CDRIVER-3892

This implements the second prose test from the CRUD spec and relocates the first prose test, which was originally added in CDRIVER-3575 (https://github.com/mongodb/mongo-c-driver/commit/84635b39d4a3cdde133c3c8cfebd9c070a1943ab).


src/libmongoc/tests/test-mongoc-crud.c
```diff
@@ -51,8 +51,135 @@ test_all_spec_tests (TestSuite *suite)
       suite, resolved, &test_crud_cb, TestSuite_CheckLive);
 }
 
+static void
+prose_test_1 (void *ctx)
```
jmikola, line 55: FWIW, I actually wrote this _before_ I realized the prose test had already been implemented in https://github.com/mongodb/mongo-c-driver/commit/84635b39d4a3cdde133c3c8cfebd9c070a1943ab. Aside from some stylistic differences, it looks exactly the same. Seeing the original code did motivate me to use `get_test_collection` in both of these tests.

Let me know if you prefer to keep the original function, but either way I think we should relocate it to this file. That's consistent with how we organize prose tests for other specs (e.g. change streams, transactions).

src/libmongoc/tests/test-libmongoc.c
```diff
@@ -2477,6 +2477,7 @@ WIRE_VERSION_CHECKS (6)
 WIRE_VERSION_CHECKS (7)
 WIRE_VERSION_CHECKS (8)
 WIRE_VERSION_CHECKS (9)
+WIRE_VERSION_CHECKS (12)
```
jmikola, line 2480: 4.4 is wire version 9 and 5.0 jumps right to 12 (see: [wireversion-featurelist.rst](https://github.com/mongodb/specifications/blob/master/source/wireversion-featurelist.rst)). Let me know if this warrants a comment, but I imagine these only get created when they're actually needed by a test so I wouldn't anticipate this being an issue in the future.

alcaeus (approved):

src/libmongoc/tests/test-mongoc-crud.c
benjirewis: Relocating the test seems best to me (and this new version is nicely commented + formatted). I see prose tests for change streams, transactions and SDAM are in the right places; are there other prose tests that are scattered amongst unit tests?

benjirewis (approved):

benjirewis:
Can you restart the Evergreen patch? Looks like there's been an internal error.

src/libmongoc/tests/test-mongoc-crud.c
jmikola: I've no idea if we have other prose tests scattered about. I only found this original prose test because I happened to check the commit for the DRIVERS ticket for when this test was introduced. Nothing about the name of the test I deleted suggested that it was a spec prose test.

src/libmongoc/tests/test-libmongoc.c
kevinAlbs: Minor: [wire_version.h|https://github.com/mongodb/mongo/blob/2d0a51a52385cb01e4bba0170270daaa349cc99b/src/mongo/db/wire_version.h#L96] notes wire version 12 is for 4.9, and 13 for 5.0.
```suggestion
WIRE_VERSION_CHECKS (13)
```

Though [SERVER-20547](https://jira.mongodb.org/browse/SERVER-20547) was implemented in 4.9, I don't think we'll want to switch on whether a server is >= 4.9.

src/libmongoc/tests/test-libmongoc.h
```diff
@@ -188,6 +188,7 @@ WIRE_VERSION_CHECK_DECLS (6)
 WIRE_VERSION_CHECK_DECLS (7)
 WIRE_VERSION_CHECK_DECLS (8)
 WIRE_VERSION_CHECK_DECLS (9)
+WIRE_VERSION_CHECK_DECLS (12)
```
kevinAlbs, line 191:
```suggestion
WIRE_VERSION_CHECK_DECLS (13)
```

src/libmongoc/tests/test-mongoc-crud.c
```diff
@@ -51,8 +51,135 @@ test_all_spec_tests (TestSuite *suite)
...
+                      NULL, /* ctx */
+                      test_framework_skip_if_no_failpoint,
+                      test_framework_skip_if_max_wire_version_less_than_7);
+
+   TestSuite_AddFull (suite,
+                      "/crud/prose_test_2",
+                      prose_test_2,
+                      NULL, /* dtor */
+                      NULL, /* ctx */
+                      test_framework_skip_if_max_wire_version_less_than_12);
```
kevinAlbs, line 184:
```suggestion
                      test_framework_skip_if_max_wire_version_less_than_13);
```

src/libmongoc/tests/test-mongoc-crud.c
kevinAlbs: Moving and renaming to be more consistent SGTM. That should make it easier for readers to tie it back to the spec prose test.

src/libmongoc/tests/test-mongoc-crud.c
```diff
@@ -51,8 +51,135 @@ test_all_spec_tests (TestSuite *suite)
...
+   (void) mongoc_collection_drop (coll, NULL);
+
+   ret = mongoc_database_create_collection (
+      db,
+      mongoc_collection_get_name (coll),
+      tmp_bson ("{'validator': {'x': {'$type': 'string'}}}"),
+      &error);
+   ASSERT_OR_PRINT (ret, error);
+
+   callbacks = mongoc_apm_callbacks_new ();
```
kevinAlbs, line 140: This needs an associated `mongoc_apm_callbacks_destroy (callbacks)`.

src/libmongoc/tests/test-mongoc-crud.c
```diff
@@ -51,8 +51,135 @@ test_all_spec_tests (TestSuite *suite)
...
+   bson_error_t error = {0};
+
+   client = test_framework_new_default_client ();
+   db = get_test_database (client);
+   coll = get_test_collection (client, "coll");
+
+   /* don't care if ns not found. */
+   (void) mongoc_collection_drop (coll, NULL);
+
+   ret = mongoc_database_create_collection (
```
kevinAlbs, line 133: This returns a `mongoc_collection_t*` that must be destroyed.

kevinAlbs (commented):
I have minor comments to update wire version checks and suggestions to fix leaks reported by ASAN.

src/libmongoc/tests/test-mongoc-crud.c
jmikola: Oh, that might explain one of the odder test failures I came across earlier. I'll update this.

jmikola:
@kevinAlbs: I posted about the Evergreen CI failures in Slack. They appear unrelated to the PR but I couldn't match them all up with [failing-on-waterfall](https://jira.mongodb.org/issues/?jql=labels%20%3D%20failing-on-waterfall) issues. I'll wait for a final LGTM on this before merging.

kevinAlbs (approved):
I filed CDRIVER-4035 to track the failure observed on 4.4 replica set tasks. I do not think it is related. It is also failing on the waterfall. The remaining failures look related to the out-of-date latest server version obtained on Ubuntu 16.04.

PR #804, opened 2021-06-17
CDRIVER-4005 Test against 5.0

alcaeus:
CDRIVER-4005

This adds 5.0 to the build. The versioned API builds are now run on 5.0 and latest, which required some changes to the config generator files.


.evergreen/download-mongodb.sh
```diff
@@ -202,33 +223,49 @@ get_mongodb_download_url_for ()
...
          MONGODB_LATEST="http://downloads.10gen.com/linux/mongodb-linux-x86_64-enterprise-debian10-latest.tgz"
+             MONGODB_50="http://downloads.10gen.com/linux/mongodb-linux-x86_64-enterprise-debian10-${VERSION_50}.tgz"
              MONGODB_44="http://downloads.10gen.com/linux/mongodb-linux-x86_64-enterprise-debian10-${VERSION_44}.tgz"
       ;;
+      linux-ubuntu-20.04-aarch64)
+         MONGODB_LATEST="http://downloads.10gen.com/linux/mongodb-linux-aarch64-enterprise-ubuntu2004-latest.tgz"
+             MONGODB_50="http://downloads.10gen.com/linux/mongodb-linux-aarch64-enterprise-ubuntu2004-${VERSION_50}.tgz"
+             MONGODB_44="http://downloads.10gen.com/linux/mongodb-linux-aarch64-enterprise-ubuntu2004-${VERSION_44}.tgz"
+      ;;
+      linux-ubuntu-20.04*)
```
jmikola, line 242: Since "aarch64" is matched above, is the wildcard actually relevant here or is it just copying what was done for older distros? I assume the only other architecture at this point would be amd64, as 20.04 was the first release to drop i386 according to https://wiki.ubuntu.com/FocalFossa/ReleaseNotes.

.evergreen/download-mongodb.sh
```diff
@@ -40,18 +43,19 @@ get_distro ()
...
 get_mongodb_download_url_for ()
 {
    _DISTRO=$1
    _VERSION=$2
 
-   VERSION_44="v4.4-latest"
-   VERSION_42="4.2.5"
-   VERSION_40="4.0.17"
-   VERSION_36="3.6.17"
+   VERSION_50="5.0.0-rc0"
```
jmikola, line 54: Any reason not to use 5.0.0-rc2 here?

.evergreen/download-mongodb.sh
```diff
@@ -40,18 +43,19 @@ get_distro ()
    echo $DISTRO
 }
 
-# get_mongodb_download_url_for "linux-distro-version-architecture" "latest|40|36|34|32|30|28|26|24"
-# Sets EXTRACT to aproprate extract command
-# Sets MONGODB_DOWNLOAD_URL to the aproprate download url
+# get_mongodb_download_url_for "linux-distro-version-architecture" "latest|44|42|40|36|34|32|30|28|26|24"
```
jmikola, line 46: Should this include "5.0"?

Also, looking at `case "$_VERSION" in` down below, the second argument (`_VERSION=$2`) should be dotted (e.g. "4.4").

.evergreen/download-mongodb.sh
```diff
@@ -9,6 +9,9 @@ get_distro ()
    if [ -f /etc/os-release ]; then
       . /etc/os-release
       DISTRO="${ID}-${VERSION_ID}"
+   elif [ -f /etc/centos-release ]; then
+      version=$(cat /etc/centos-release | tr -dc '0-9.' | cut -d '.' -f1)
+      DISTRO="centos-${version}"
```
jmikola, line 14: Was this only required for `linux-centos-6*`? I assume RHEL is handled by `elif [ -f /etc/redhat-release ];` later on?

.evergreen/download-mongodb.sh
alcaeus: This file was copied over from drivers-evergreen-tools, since the c driver still uses its own files. I plan to revisit that separately so we don't need to maintain duplicate files.

jmikola (approved):
All of my comments were on `download-mongodb.sh`, which @alcaeus said was copied from [drivers-evergreen-tools](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/download-mongodb.sh). In that case, this LGTM but I think the comments are still applicable to address upstream.

PR #807, opened 2021-06-23
CDRIVER-4034 Allow snapshot reads through sessions

benjirewis:
CDRIVER-4034

Adds `MONGOC_SESSION_SNAPSHOT` flag option to `mongoc_session_opt_t`. Setting this flag will mean:

- The first read operation (find or aggregate) will be sent with read concern level "snapshot"
- The server will return `atClusterTime` in the reply cursor
- `atClusterTime` will be stored as new fields `snapshot_time_timestamp` and `snapshot_time_increment` on the `mongoc_client_session_t`
- Subsequent read operations (find or aggregate) will also be sent with read concern level "snapshot" _and_ `atClusterTime` set to the `snapshot_time` stored on the client session. 

These changes will allow users to read from a single point in time (a snapshot) during a session.

Adds new `at_cluster_time_timestamp` and `at_cluster_time_increment` fields to read concern (along with getters) and a new `mongoc_read_concern_set_at_cluster_time` setter. Allows setting `MONGOC_SESSION_SNAPSHOT` with `isSnapshot` in the unified test runner. Syncs sessions spec tests.


src/libmongoc/src/mongoc/mongoc-client-session-private.h
```diff
@@ -83,6 +84,9 @@ struct _mongoc_client_session_t {
    uint32_t client_generation;
    uint32_t server_id;
    bson_t *recovery_token;
+   uint32_t snapshot_time_timestamp;
+   uint32_t snapshot_time_increment;
+   bool snapshot_time_set;
```
benjirewis, line 89: This was my best guess at storing an optional timestamp 🤷 . It looks like timestamps are often stored as their `timestamp` and `increment` values (see `operation_timestamp` and `operation_increment` above), but I needed a `bool` as well (`snapshot_time_set`) to know when the snapshot time was set. I tried to use two `uint32_t *`  but the pointers lost the underlying value as I passed them around. Let me know what you think of this approach (I do something similar for `at_cluster_time` on `read_concern`).

benjirewis, line 89: This was my best guess at an optional timestamp 🤷 . I could have used pointers to `uint32_t`s, but the underlying values kept getting lost as I passed them around. Let me know what you think (I use the same structure for `at_cluster_time` in `read_concern`).

src/libmongoc/src/mongoc/mongoc-client-session-private.h
```diff
@@ -41,6 +41,7 @@ struct _mongoc_transaction_opt_t {
 typedef enum {
    MONGOC_SESSION_NO_OPTS = 0,
    MONGOC_SESSION_CAUSAL_CONSISTENCY = (1 << 0),
+   MONGOC_SESSION_SNAPSHOT = (2 << 0),
```
benjirewis, line 44: Snapshot and causal consistency are [mutually exclusive](https://github.com/mongodb/specifications/pull/1022/files#diff-6c9d82f6b269142ff8fe9f0ef8d7c90f9e2761c947b5caf92e6820124b81fc2bR138). Given that, this enum seemed like the best place for the new snapshot sessions options.

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -417,6 +424,11 @@ mongoc_session_opts_set_causal_consistency (mongoc_session_opt_t *opts,
    BSON_ASSERT (opts);
 
    if (causal_consistency) {
+      if (mongoc_session_opts_get_snapshot (opts)) {
+         MONGOC_WARNING (
+            "Cannot set causal consistency when session is snapshot");
+         EXIT;
```
benjirewis, line 430: I'm _sure_ this is not the right way to "error". Any ideas?

src/libmongoc/src/mongoc/mongoc-cursor.c
```diff
@@ -1002,7 +1023,8 @@ _mongoc_cursor_run_command (mongoc_cursor_t *cursor,
    /* we might use mongoc_cursor_set_hint to target a secondary but have no
     * read preference, so the secondary rejects the read. same if we have a
     * direct connection to a secondary (topology type "single"). with
-    * OP_QUERY we handle this by setting secondaryOk. here we use $readPreference.
+    * OP_QUERY we handle this by setting secondaryOk. here we use
+    * $readPreference.
```
benjirewis, line 1027: Random formatting from my `clang` 🧑‍🔧  here and above.

src/libmongoc/src/mongoc/mongoc-read-concern.h
```diff
@@ -46,6 +46,16 @@ mongoc_read_concern_get_level (const mongoc_read_concern_t *read_concern);
...
+MONGOC_EXPORT (uint32_t)
+mongoc_read_concern_get_at_cluster_time_timestamp (
+   const mongoc_read_concern_t *read_concern);
+MONGOC_EXPORT (uint32_t)
+mongoc_read_concern_get_at_cluster_time_increment (
+   const mongoc_read_concern_t *read_concern);
+MONGOC_EXPORT (bool)
+mongoc_read_concern_set_at_cluster_time (mongoc_read_concern_t *read_concern,
+                                         uint32_t t,
+                                         uint32_t i);
```
benjirewis, line 58: I made this API for setting and getting `at_cluster_time` public. Users _could_ feasibly send a read concern with level "snapshot" and "atClusterTime" specified to read from a particular point in time outside of a session. If this remains public, we'll have to document how this works and point to relevant docs in CDRIVER-4036.

src/libmongoc/tests/json/sessions/legacy/server-support.json
```diff
@@ -0,0 +1,181 @@
+{
```
benjirewis, line 1: This was a missing `sessions` test. It passes as-is, so I added it in while moving the other old test to `legacy`.

src/libmongoc/tests/unified/runner.c
```diff
@@ -60,6 +60,9 @@ skipped_unified_test_t SKIPPED_TESTS[] = {
    {"unacknowledged-updateOne-hint-clientError", "Unacknowledged updateOne with hint document fails with client-side error"},
    /* CDRIVER-4001, DRIVERS-1781, and DRIVERS-1448: 5.0 cursor behavior */
    {"poc-command-monitoring", "A successful find event with a getmore and the server kills the cursor"},
+   /* libmongoc does not have a distinct helper, so skip snapshot tests testing particular distinct functionality */
+   {"snapshot-sessions", "Distinct operation with snapshot"},
+   {"snapshot-sessions", "Mixed operation with snapshot"},
```
benjirewis, line 65: The new spec tests test that `distinct` behaves correctly during snapshot sessions. libmongoc has no `distinct` helper, so I thought it best to skip these new tests instead of adding snapshot functionality to the [`operation_distinct`](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/tests/unified/operation.c#L808-L865) test helper.

src/libmongoc/tests/unified/entity-map.c
```diff
@@ -551,6 +552,7 @@ session_opts_new (bson_t *bson, bson_error_t *error)
 
    bp = bson_parser_new ();
    bson_parser_bool_optional (bp, "causalConsistency", &causal_consistency);
+   bson_parser_bool_optional (bp, "isSnapshot", &snapshot);
```
alcaeus, line 555: I was confused by this change not being accompanied by a change in the supported schemas, but I forgot that the unified test format doesn't specify these options but rather accepts anything that the `startSession` helper supports. Adding this for the benefit of other reviewers.

src/libmongoc/tests/unified/entity-map.c
```diff
@@ -561,6 +563,11 @@ session_opts_new (bson_t *bson, bson_error_t *error)
    if (causal_consistency) {
       mongoc_session_opts_set_causal_consistency (opts, *causal_consistency);
    }
+   if (snapshot) {
+      // Set causal consistency to false to allow snapshot
+      mongoc_session_opts_set_causal_consistency (opts, false);
```
alcaeus, line 568: I don't think this is necessary. Since the combination of causal consistency and snapshot reads is invalid, we shouldn't quietly fix this scenario in the tests but forbid it elsewhere. A test specifying both causal consistency and snapshot reads should fail so the offending test is highlighted instead of silently being fixed.

src/libmongoc/tests/unified/runner.c
```diff
@@ -1585,10 +1588,11 @@ test_install_unified (TestSuite *suite)
...
    run_unified_tests (suite, JSON_DIR "/unified");
 
-   /* crud/unified spec tests also use the unified format. */
    run_unified_tests (suite, JSON_DIR "/crud/unified");
 
    run_unified_tests (suite, JSON_DIR "/transactions/unified");
 
    run_unified_tests (suite, JSON_DIR "/collection-management");
+
+   run_unified_tests (suite, JSON_DIR "/sessions/unified");
```
alcaeus, line 1597: Just a heads up, I've deviated from this already for the Versioned API, as I thought it'd be more sensible to add the unified spec tests for a spec in the same place where we would've added the legacy spec tests. As far as I can tell, this wouldn't affect the `SKIPPED_TESTS` behaviour, but it would keep tests for one spec in one place. No strong preference on this though, and I can move the Versioned API tests if we decide that this is the correct place to add new unified tests.

alcaeus (commented):
A few comments after an initial read. Functionality looks sound, just a few minor suggestions regarding the API.

src/libmongoc/src/mongoc/mongoc-client-session-private.h
benjirewis: Ah yeah that'd be useful here. These `snapshot_time` fields aren't exposed API, but the `at_cluster_time` fields are, so it might be worthwhile for me to implement `mongoc_optional_t` for timestamps if @kevinAlbs agrees.

src/libmongoc/tests/unified/entity-map.c
benjirewis: Ah yep, I don't think there were any unified test format changes here.

src/libmongoc/tests/unified/entity-map.c
benjirewis: There is an [ongoing discussion](https://github.com/mongodb/specifications/pull/1022#discussion_r657999964) in the new specification about whether to turn off `causalConsistency` when `isSnapshot` is set. `causalConsistency` is true by default, so without this line we'd technically error when attempting to `set_snapshot` to true.

src/libmongoc/tests/unified/runner.c
benjirewis: Ah I see, I'll leave it up to @kevinAlbs. The Go and CXX drivers do keep all the unified spec tests in one runner as is done here. Either way, I'll either move this to `test-mongoc-client-session` or move the versioned API tests here.

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -639,8 +678,29 @@ _mongoc_client_session_handle_reply (mongoc_client_session_t *session,
...
+
+         while (bson_iter_next (&cursor_iter)) {
+            /* If atClusterTime is in cursor and is a valid timestamp, use it to
+             * set snapshot_time_timestamp and increment. */
+            if (!strcmp (bson_iter_key (&cursor_iter), "atClusterTime") &&
+                BSON_ITER_HOLDS_TIMESTAMP (&cursor_iter)) {
+               bson_iter_timestamp (&cursor_iter, &snapshot_t, &snapshot_i);
+               session->snapshot_time_set = true;
+               session->snapshot_time_timestamp = snapshot_t;
+               session->snapshot_time_increment = snapshot_i;
```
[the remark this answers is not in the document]
benjirewis: Great idea. Added a private method `_mongoc_client_session_set_snapshot_time` and used it here instead of setting all three variables.

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -751,6 +811,9 @@ _mongoc_client_session_new (mongoc_client_t *client,
       session->opts.flags = MONGOC_SESSION_CAUSAL_CONSISTENCY;
    }
 
+   /* snapshot_time_set is false by default */
+   session->snapshot_time_set = false;
```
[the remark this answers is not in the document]
benjirewis: Sure! Added a private method `_mongoc_session_clear_snapshot_time` and used it here.

src/libmongoc/src/mongoc/mongoc-read-concern.c
```diff
@@ -113,6 +126,50 @@ mongoc_read_concern_set_level (mongoc_read_concern_t *read_concern,
    return true;
 }
 
+uint32_t
+mongoc_read_concern_get_at_cluster_time_timestamp (
+   const mongoc_read_concern_t *read_concern)
+{
+   BSON_ASSERT (read_concern);
+
+   return read_concern->at_cluster_time_timestamp;
```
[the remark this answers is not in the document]
benjirewis: Added a guard here and on `get_at_cluster_time_increment`, but wasn't sure what to do in the case of `at_cluster_time_set` being false. Chose to WARN and return 0, but let me know if that seems wrong.

src/libmongoc/src/mongoc/mongoc-read-concern.h
kevinAlbs: To play devil's advocate, I do not think this is a requirement of the spec and I do not see a compelling use case for the new API.

Snapshot reads setting `atClusterTime` in the read concern seems analogous to how causal consistency sets [`afterClusterTime` in the read concern](https://github.com/mongodb/specifications/blob/master/source/causal-consistency/causal-consistency.rst#server-commands).

There is no API to manually configure an `afterClusterTime` on a read concern. It is appended in the private [_mongoc_client_session_append_read_concern](https://github.com/mongodb/mongo-c-driver/blob/dad33550aa098229d395d2cf7c02f10a9f3f40c4/src/libmongoc/src/mongoc/mongoc-client-session.c#L1491:L1491).

The current draft of https://github.com/mongodb/specifications/pull/1022/files#diff-4a8e136fa35d82e76e2e53dd27491b4b06c1510164ade6ae65f545b6abf78709 does not indicate that new read concern API is needed to set `afterClusterTime`.

I do not think users can easily obtain the value of `atClusterTime` in `find` / `aggregate`. Users could use command monitoring or use the generic command helpers to parse the server reply. But we are not exposing a way to retrieve `atClusterTime` returned on a cursor. I think the expectation is that users will not be setting `atClusterTime` manually.

If there is not a compelling use case I would be inclined to add the API. But otherwise, I vote we leave it private.

src/libmongoc/tests/unified/runner.c
kevinAlbs: I think we should match the directory structure in the specifications repo. Keeping legacy and unified in separate directories makes sense since they go through different runners.

DRIVERS-1607 moves the existing sessions tests into:

`source/sessions/tests` => `source/sessions/tests/legacy`

And the new unified tests are put into:

`source/sessions/tests/unified`

I think that should correspond to these directories in the C driver (as proposed in this PR):

- `src/libmongoc/tests/json/sessions/legacy`
- `src/libmongoc/tests/json/sessions/unified`

@alcaeus Are you referring to the Versioned API tests in `src/libmongoc/tests/json/versioned_api`? That matches the specifications repo since there is no distinction between legacy and unified.

src/libmongoc/src/mongoc/mongoc-read-concern.h
kevinAlbs: Additionally, if `atClusterTime` is handled the same way as `afterClusterTime`, I think that may simplify the implementation.

There would no longer be a need to store it on a read concern at all.

We would not need to conditionally append a read concern in a cursor. The call to `_mongoc_client_session_append_read_concern` could append the read concern if the command name is `aggregate` / `find` / `distinct`. That may also be more correct, in that it would cover the case that a user is using a generic command helper (like `mongoc_client_command_with_opts`) with `aggregate` / `find` / `distinct`.

kevinAlbs (commented):
This is looking like a great start!

I have not done a line-by-line review, but I have high level questions about:
- Can we hide the API for setting `atClusterTime` on a read concern?
- Can we handle `atClusterTime` similarly to `afterClusterTime` to simplify how the read concern is appended?

src/libmongoc/src/mongoc/mongoc-client-session-private.h
benjirewis: Ah that makes sense, so they _can_ both be set on the flag; but we'll error in that case (for now).

src/libmongoc/src/mongoc/mongoc-client-session-private.h
benjirewis: No longer necessary since `at_cluster_time` fields are gone.

src/libmongoc/src/mongoc/mongoc-client-session.c
benjirewis: Perfect. I added an error with `MONGOC_ERROR_CLIENT` and `MONGOC_ERROR_CLIENT_SESSION_FAILURE` in `mongoc_client_start_session` instead.

src/libmongoc/src/mongoc/mongoc-read-concern.h
benjirewis: Thanks for the in-depth response here. Given the way `afterClusterTime` is currently added, and the simplicity + accuracy of just adding `snapshot` read concern in `_mongoc_client_session_append_read_concern`, I've decided to do just that and remove all new API on `mongoc_read_concern`. This really simplifies the whole PR 😆  

If @alcaeus still feels there's a need for users appending `atClusterTime` independently of the driver's sessions API, I can add the API back in (but I don't think it's needed to comply with the current spec).

src/libmongoc/tests/unified/runner.c
benjirewis: Unless I'm mistaken, @alcaeus was referring to where the tests are run from, not where the `json` files are. In this case, I've added `run_unified_tests` here in `runner.c`, whereas the versioned API tests (which are unified) are run from [`test-mongoc-versioned-api`](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/tests/test-mongoc-versioned-api.c#L179).

benjirewis (commented):
Refactored slightly to append snapshot to read_concern in `_mongoc_client_session_append_read_concern`. This seems simpler and more accurate to me; thanks for the suggestion @kevinAlbs.

Implemented the one [prose test](https://github.com/mongodb/specifications/pull/1022/files#diff-af22a0d4d8ce3472c49607d6d338dd368e55ee8b134cfd50717b44829233616fR88), too (CDRIVER-4037). I had mentioned offline there may be Evergreen changes for the test to set `minSnapshotHistoryWindowInSeconds` on the server, but I'm unsure if this is [strictly necessary](https://github.com/mongodb/specifications/pull/1022#discussion_r658226561).

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -639,8 +668,28 @@ _mongoc_client_session_handle_reply (mongoc_client_session_t *session,
...
+         while (bson_iter_next (&cursor_iter)) {
+            /* If atClusterTime is in cursor and is a valid timestamp, use it to
+             * set snapshot_time_timestamp and increment. */
+            if (!strcmp (bson_iter_key (&cursor_iter), "atClusterTime") &&
+                BSON_ITER_HOLDS_TIMESTAMP (&cursor_iter)) {
+               bson_iter_timestamp (&cursor_iter, &snapshot_t, &snapshot_i);
+               _mongoc_client_session_set_snapshot_time (
+                  session, snapshot_t, snapshot_i);
+            }
+         }
```
kevinAlbs, line 692: I think this should also inspect for a top-level `atClusterTime` in the server reply (not just within `cursor`).

This would support the use-case where a user runs `distinct` with a generic command helper.

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -639,8 +668,28 @@ _mongoc_client_session_handle_reply (mongoc_client_session_t *session,
          mongoc_client_session_advance_cluster_time (session, &cluster_time);
       } else if (!strcmp (bson_iter_key (&iter), "operationTime") &&
                  BSON_ITER_HOLDS_TIMESTAMP (&iter) && is_acknowledged) {
-         bson_iter_timestamp (&iter, &t, &i);
-         mongoc_client_session_advance_operation_time (session, t, i);
+         bson_iter_timestamp (&iter, &operation_t, &operation_i);
+         mongoc_client_session_advance_operation_time (
+            session, operation_t, operation_i);
+      } else if (!strcmp (bson_iter_key (&iter), "cursor") &&
```
kevinAlbs, line 674: Based on the spec:

> Saving the ``atClusterTime`` returned by 5.0+ servers for the first find/aggregate/distinct

I think we should limit the parsing to replies from `find` / `aggregate` / `distinct`. I think that `_mongoc_client_session_handle_reply` could accept a `cmd_name` argument, and the caller, `mongoc_cluster_run_opmsg`, can pass `cmd->command_name`.

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -1621,6 +1692,26 @@ _mongoc_client_session_pin (mongoc_client_session_t *session,
    session->server_id = server_id;
 }
 
+void
+_mongoc_client_session_set_snapshot_time (mongoc_client_session_t *session,
+                                          uint32_t t,
+                                          uint32_t i)
+{
+   BSON_ASSERT (session);
```
kevinAlbs, line 1700: Since I think calling this when a snapshot time is already set is a logic error:

```suggestion
   BSON_ASSERT (session);
   BSON_ASSERT (!session->snapshot_time_set);
```

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -1516,12 +1569,16 @@ _mongoc_client_session_append_read_concern (const mongoc_client_session_t *cs,
       (txn_state == MONGOC_INTERNAL_TRANSACTION_STARTING || is_read_command) &&
       mongoc_session_opts_get_causal_consistency (&cs->opts) &&
       cs->operation_timestamp;
+   is_snapshot =
+      is_read_command && mongoc_session_opts_get_snapshot (&cs->opts);
```
kevinAlbs, line 1573: `is_read_command` may be true on commands outside of `find` / `aggregate` / `distinct`. E.g. `listIndexes`, or any command run through the generic "read" helpers like `mongoc_client_read_command_with_opts`.

This function may also need to accept a `cmd_name` argument, which the caller `mongoc_cmd_parts_assemble`, has access to.

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -1531,18 +1588,32 @@ _mongoc_client_session_append_read_concern (const mongoc_client_session_t *cs,
    }
 
    if (txn_state == MONGOC_INTERNAL_TRANSACTION_STARTING) {
-      /* add transaction's read concern level unless user overrides */
-      if (txn_has_level && !user_rc_has_level) {
+      /* add transaction's read concern level unless user overrides or snapshot
+       * is enabled. */
+      if (txn_has_level && !user_rc_has_level && !is_snapshot) {
```
kevinAlbs, line 1593: Based on the spec, I think we do not want to set a snapshot read for the first operation in a transaction.

> Reads with read concern level ``snapshot`` that occur outside of transactions

This differs from causal consistency, which makes no mention of excluding transactions. The [transactions spec describes how read concern is applied](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#behavior-of-the-readconcern-field) on the first command of a transaction.

To accomplish that, this could set `is_snapshot` to false if `txn_state` is anything other than `MONGOC_INTERNAL_TRANSACTION_NONE`.

This may be good to suggest a spec test in the PR for DRIVERS-1607. I think this could be tested with the unified test runner.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1288,6 +1289,16 @@ mongoc_client_start_session (mongoc_client_t *client,
...
 
+   /* causal consistency and snapshot cannot both be set. */
+   if (opts && mongoc_session_opts_get_causal_consistency (opts) &&
+       mongoc_session_opts_get_snapshot (opts)) {
+      bson_set_error (
+         error,
+         MONGOC_ERROR_CLIENT,
+         MONGOC_ERROR_CLIENT_SESSION_FAILURE,
+         "Only one of causal consistency and snapshot can be enabled.");
+      RETURN (NULL);
```
kevinAlbs, line 1300: To prevent leaking `ss`:
```suggestion
      _mongoc_client_push_server_session (client, ss);
      RETURN (NULL);
```

kevinAlbs (commented):
I still need to review the prose test, but this is looking great. Modeling this from causal consistency seems much simpler!

src/libmongoc/src/mongoc/mongoc-client-session.c
benjirewis: Good point, otherwise we'll be doing looking at `cursor` replies for everything. Made this conditional on `is_find_aggregate_distinct` and added new `cmd_name` param to `_mongoc_client_session_handle_reply`.

src/libmongoc/src/mongoc/mongoc-client-session.c
benjirewis: Added. Didn't have this before since I was only appending in `_mongoc_cursor_new_with_opts`, so the use-case of `distinct` with a generic command helper was not supported.

src/libmongoc/src/mongoc/mongoc-client-session.c
benjirewis: Gotcha. Used `is_find_aggregate_distinct` instead of `is_read_command` here, and added a new `cmd_name` parameter to `_mongoc_client_session_append_read_concern`.

src/libmongoc/src/mongoc/mongoc-client-session.c
benjirewis: Great point; I think we should be testing this in the new spec tests (commented on it [here](https://github.com/mongodb/specifications/pull/1022/files#r658883220)). I changed the definition of `is_snapshot` to be
```
is_snapshot = is_find_aggregate_distinct &&
                 txn_state == MONGOC_INTERNAL_TRANSACTION_NONE &&
                 mongoc_session_opts_get_snapshot (&cs->opts);
```

src/libmongoc/src/mongoc/mongoc-client-session.c
benjirewis: Added.

src/libmongoc/src/mongoc/mongoc-client.c
benjirewis: Ah thanks, that should fix some of the test failures. Added.

src/libmongoc/tests/unified/runner.c
benjirewis: Gotcha, since both the `/sessions/unified` and `/versioned-api` tests are grep-able paths, I'll just leave them where they both are unless someone has a strong preference.

benjirewis (commented):
Looks like the prose test [no longer](https://github.com/mongodb/specifications/pull/1022/files#diff-af22a0d4d8ce3472c49607d6d338dd368e55ee8b134cfd50717b44829233616fR96) requires that we set `minSnapshotHistoryWindowInSeconds`, so it should be all set to review.

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -1055,6 +1124,15 @@ mongoc_client_session_start_transaction (mongoc_client_session_t *session,
       GOTO (done);
    }
 
+   if (mongoc_session_opts_get_snapshot (&session->opts)) {
+      bson_set_error (error,
+                      MONGOC_ERROR_TRANSACTION,
+                      MONGOC_ERROR_TRANSACTION_INVALID_STATE,
+                      "Transactions are not supported in snapshot sessions");
+      ret = false;
+      GOTO (done);
```
benjirewis, line 1133: This new error is an [alteration](https://github.com/mongodb/specifications/pull/1022/files#diff-6c9d82f6b269142ff8fe9f0ef8d7c90f9e2761c947b5caf92e6820124b81fc2bR140) to the spec accompanied by a new spec test "StartTransaction fails in snapshot session".

src/libmongoc/src/mongoc/mongoc-client-session.c
kevinAlbs: Awesome, thanks for confirming it is tested!

benjirewis: Now that there's a test for "StartTransaction fails in snapshot session", I think this is covered in spec tests.

src/libmongoc/tests/unified/entity-map.c
kevinAlbs: The alternative is that `mongoc_session_opts_set_snapshot` could disable causal consistency. But I think the current behavior aligns with the spec:

> If no value is provided for ``causalConsistency`` and snapshot reads are not requested a value of true is implied. See the ``causalConsistency`` section.

`mongoc_session_opts_new` always creates options with causal consistency enabled since it does not accept arguments.

src/libmongoc/tests/test-mongoc-transactions.c
```diff
@@ -1166,6 +1166,39 @@ test_max_commit_time_ms_is_reset (void *ctx)
...
+test_snapshot_session_prose_1 (void *ctx)
+{
+   mongoc_client_t *client = NULL;
+   mongoc_session_opt_t *session_opts = NULL;
+   mongoc_uri_t *uri = NULL;
+   bson_error_t error;
+   bool r;
+
+   uri = test_framework_get_uri ();
+   client = test_framework_client_new_from_uri (uri, NULL);
```
kevinAlbs, line 1179: The `test_framework_client_new_from_uri` will not set SSL options that are required when running against a server with SSL enabled. Other tests that use `test_framework_client_new_from_uri` also call `test_framework_set_ssl_opts` right after. 

Since you do not need to construct a URI in this test, you can simplify this to:

```suggestion
   client = test_framework_new_default_client ();
```

That also sets the SSL options.

src/libmongoc/tests/test-mongoc-transactions.c
```diff
@@ -1166,6 +1166,39 @@ test_max_commit_time_ms_is_reset (void *ctx)
...
+
+   session_opts = mongoc_session_opts_new ();
+   mongoc_session_opts_set_causal_consistency (session_opts, true);
+   mongoc_session_opts_set_snapshot (session_opts, true);
+
+   /* assert that starting session with causal consistency and snapshot enabled
+    * results in an error. */
+   r = mongoc_client_start_session (client, session_opts, &error);
+   ASSERT (!r);
+   ASSERT (error.domain == MONGOC_ERROR_CLIENT);
```
kevinAlbs, line 1190: Minor: the following can be simplified to:

```suggestion
   ASSERT_ERROR_CONTAINS (error,
                          MONGOC_ERROR_CLIENT, 
                          MONGOC_ERROR_CLIENT_SESSION_FAILURE,
                          "Only one of causal consistency and snapshot can be enabled.");
```

kevinAlbs (commented):
Fantastic work. LGTM with a minor test tweak. Let's wait until https://github.com/mongodb/specifications/pull/1022/files/dfb2472e96e1e3f4f374eb526c5190c8282239b0..e1e0672414ff7756de2a2fb9a5d4b7ddea7c1066 is merged to merge this in case there are any spec / test changes to add.

src/libmongoc/src/mongoc/mongoc-read-concern.h
alcaeus: No, I think there's no need to add that 👍

src/libmongoc/src/mongoc/mongoc-read-concern.c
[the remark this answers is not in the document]
alcaeus: That sounds fine for the moment 👍

src/libmongoc/tests/unified/runner.c
alcaeus: No preference here, I'll clean up the other tests when I get a chance

alcaeus (approved):
LGTM besides what @kevinAlbs mentioned

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -1565,12 +1593,17 @@ _mongoc_client_session_append_read_concern (const mongoc_client_session_t *cs,
...
+       !strcmp (cmd_name, "distinct"));
+
    has_timestamp =
       (txn_state == MONGOC_INTERNAL_TRANSACTION_STARTING || is_read_command) &&
       mongoc_session_opts_get_causal_consistency (&cs->opts) &&
       cs->operation_timestamp;
-   is_snapshot =
-      is_read_command && mongoc_session_opts_get_snapshot (&cs->opts);
+   is_snapshot = is_find_aggregate_distinct &&
+                 txn_state == MONGOC_INTERNAL_TRANSACTION_NONE &&
```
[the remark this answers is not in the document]
benjirewis: Yep! Now that we explicitly error in `mongoc_client_session_start_transaction` when snapshot is enabled on the session, I believe this check is pointless. Removed.

kevinAlbs (approved):

PR #809, opened 2021-06-25
CDRIVER-3895: Sync CRUD tests for dots/dollars

jmikola:
https://jira.mongodb.org/browse/CDRIVER-3895

Synced with mongodb/specifications@fc21cb7330925647e7961c324f1594096025c17e (see https://github.com/mongodb/specifications/pull/1028 for more context).

Note: this also pulls in test changes relevant to aggregate let support (CDRIVER-4010)

@rcsanchez97: The previous PR for CDRIVER-3895 (#801) introduced `aggregate-let.json` and this PR pulls in the latest updates from https://github.com/mongodb/specifications/commit/3a58fc612b3520030a97a1161a5651c451c40f8e. Per @kevinAlbs' comment in [CDRIVER-4010](https://jira.mongodb.org/browse/CDRIVER-4010?focusedCommentId=3881139&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-3881139) there may be no other work required for that ticket (perhaps beyond documentation if applicable) if the spec tests currently pass.

PR #810, opened 2021-06-25
CDRIVER-4010 support 'let' option for aggregate command

rcsanchez97:
Full patch build: https://spruce.mongodb.com/version/60d5ee4be3c33111e05b9faa/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

I checked quite a few of the failures in that patch build in every single case (with one exception*) the failure in the patch build was identical to the failure of the same task in the base commit.  *The one exception is a Valgrind test that had a hard failure in the base commit but which timed out in the patch build.

Based on this, it seems like the C driver already has the necessary support, so simply updating the CRUD JSON test is all we need to do.

### DIFF
--- a/build/generate-opts.py
+++ b/build/generate-opts.py
@@ -320,7 +320,8 @@ opts_structs = OrderedDict([
         bypass_option,
         collation_option,
         server_option,
-        ('batchSize', {'type': 'int32_t', 'help': 'An ``int32`` representing number of documents requested to be returned on each call to :symbol:`mongoc_cursor_next`', 'check_set': True})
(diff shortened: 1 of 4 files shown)


kevinAlbs (approved):
LGTM, thank you for inspecting test logs to verify these tests did not introduce errors.

kevinAlbs (changes requested):
Oops, I reviewed this before looking at https://github.com/mongodb/mongo-c-driver/pull/809. The test resync will be part of that PR.

Can you additionally update the documentation pages:
- http://mongoc.org/libmongoc/current/mongoc_database_aggregate.html
- http://mongoc.org/libmongoc/current/mongoc_collection_aggregate.html

To note the new option?

rcsanchez97:
Sure.  I am guessing that I will also need to drop the test sync since
#809 handled that.

kevinAlbs (approved):
LGTM

PR #811, opened 2021-06-29
CDRIVER-4028 Print correct error message when DNS resolution fails

alcaeus:
CDRIVER-4028

I'll backport the resulting commit to r1.17 after merging.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-client.c
+++ b/src/libmongoc/src/mongoc/mongoc-client.c
@@ -306,7 +306,6 @@ _mongoc_hstrerror (int code)
    case HOST_NOT_FOUND:
       return "The specified host is unknown.";
    case NO_ADDRESS:
-   case NO_DATA:
       return "The requested name is valid but does not have an IP address.";
    case NO_RECOVERY:
       return "A nonrecoverable name server error occurred.";
--- a/src/libmongoc/src/mongoc/mongoc-client.c
+++ b/src/libmongoc/src/mongoc/mongoc-client.c
@@ -299,6 +299,25 @@ typedef bool (*mongoc_rr_callback_t) (const char *service,
                                       mongoc_rr_data_t *rr_data,
                                       bson_error_t *error);
 
+static const char *
+_mongoc_hstrerror (int code)
+{
+   switch (code) {
+   case HOST_NOT_FOUND:
+      return "The specified host is unknown.";
+   case NO_ADDRESS:
+   case NO_DATA:
+      return "The requested name is valid but does not have an IP address.";
+   case NO_RECOVERY:
+      return "A nonrecoverable name server error occurred.";
+   case TRY_AGAIN:
+      return "A temporary error occurred on an authoritative name server. Try "
+             "again later.";
+   default:
+      return "An unknown error occurred.";
+   }
+}
+
 static bool
 srv_callback (const char *service,
               ns_msg *ns_answer,
@@ -328,7 +347,7 @@ srv_callback (const char *service,
    if (size < 1) {
       DNS_ERROR ("Invalid record in SRV answer for \"%s\": \"%s\"",
                  service,
-                 strerror (h_errno));
+                 _mongoc_hstrerror (h_errno));
    }
 
    if (!_mongoc_host_list_from_hostport_with_err (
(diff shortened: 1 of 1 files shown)


src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -299,6 +299,25 @@ typedef bool (*mongoc_rr_callback_t) (const char *service,
...
+   case NO_ADDRESS:
+   case NO_DATA:
+      return "The requested name is valid but does not have an IP address.";
+   case NO_RECOVERY:
+      return "A nonrecoverable name server error occurred.";
+   case TRY_AGAIN:
+      return "A temporary error occurred on an authoritative name server. Try "
+             "again later.";
+   default:
+      return "An unknown error occurred.";
```
jmikola, line 317: In practice, would we ever hit this case? If so, I wonder if it'd be helpful to somehow expose the integer value in the error message relayed up the chain. I realize we can't print the value of `code` here since it'd require allocating a string, but perhaps we can do this in the calling context.

If it's unlikely we'd hit this case, I'm fine leaving it as-is.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -328,7 +347,7 @@ srv_callback (const char *service,
    if (size < 1) {
       DNS_ERROR ("Invalid record in SRV answer for \"%s\": \"%s\"",
                  service,
-                 strerror (h_errno));
+                 _mongoc_hstrerror (h_errno));
```
jmikola, line 350: I'm trusting that `h_errno` is always set by this point. I do find it odd that we rely on it being declared and reference it here even if _both_ DNS APIs used above (which would set this) are unavailable. Or would libmongoc never compile without at least one of those available?

I left some more general comments about using `h_errno` in [CDRIVER-4028](https://jira.mongodb.org/browse/CDRIVER-4028?focusedCommentId=3907963&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-3907963), but that's outside the scope of this PR.

jmikola (approved):
LGTM. One question about the default `switch` case and another about relying on `h_errno` to be declared, but I don't expect the answers to those questions will warrant any changes here.

src/libmongoc/src/mongoc/mongoc-client.c
alcaeus: I was thinking of it but avoided it because it would require allocating memory. Either way, even if we hit this scenario, it would be difficult for us to do anything as the value given would not make any sense to us (since it isn't documented). Even with the code, we'd most likely have to resort to tracing or other means to find out what's going on so I opted for this message.

src/libmongoc/src/mongoc/mongoc-client.c
alcaeus: Quoting from the [resolver(3) man page](https://man7.org/linux/man-pages/man3/res_nsearch.3.html):
> In the case of an error return from res_nquery(), res_query(), res_nsearch(), res_search(), res_nquerydomain(), or res_querydomain(), the global variable h_errno (see gethostbyname(3)) can be consulted to determine the cause of the error.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -306,7 +306,6 @@ _mongoc_hstrerror (int code)
    case HOST_NOT_FOUND:
       return "The specified host is unknown.";
    case NO_ADDRESS:
-   case NO_DATA:
```
jmikola: Is this guaranteed to have the same value as `NO_ADDRESS`? I assume you caught this in some CI warning?

jmikola (approved):

src/libmongoc/src/mongoc/mongoc-client.c
alcaeus: Yes. I added both constants to ensure we're grabbing it, but CI complained about the case being duplicated. I'd argue that using one of these should be sufficient, and the 50:50 joker picked `NO_DATA` for elimination. I don't anticipate trouble from this, but I suppose a last resort to deal with either constant not being defined would be to hardcode the value (`4`) and be done with it.

alcaeus:
Cherry-picked to r1.17 in e1be09a.

PR #812, opened 2021-06-29
CDRIVER-3164 & CDRIVER-4049: Fix 'opts' usage in mongoc_gridfs_bucket_find

vector-of-bool:
GridFSBucket.find must forward the 'opts' to the underlying
collection query.

Also: Permit 'opts' to be NULL for mongoc_gridfs_bucket_find

### DIFF
--- a/src/libmongoc/tests/test-mongoc-gridfs-bucket.c
+++ b/src/libmongoc/tests/test-mongoc-gridfs-bucket.c
@@ -995,20 +995,36 @@ test_find (void *ctx)
    bool ok;
    bson_value_t const *found_id;
    const bson_t *const find_opts =
-      tmp_bson ("{'limit': 1, 'skip': 2, 'sort': {'uploadDate': -1}}");
+      tmp_bson ("{'limit': 1, 'skip': 2, 'sort': {'metadata.testOrder': -1}}");
 
-   _upload_file_from_str (gridfs, "file1", "First file", NULL, NULL);
-   _upload_file_from_str (gridfs, "file2", "Second file", NULL, NULL);
-   _upload_file_from_str (gridfs, "file3", "Third file", NULL, NULL);
-   _upload_file_from_str (gridfs, "file4", "Fourth file", NULL, NULL);
+   _upload_file_from_str (gridfs,
+                          "file1",
+                          "First file",
+                          tmp_bson ("{'metadata': {'testOrder': 1}}"),
+                          NULL);
+   _upload_file_from_str (gridfs,
+                          "file2",
+                          "Second file",
+                          tmp_bson ("{'metadata': {'testOrder': 2}}"),
+                          NULL);
+   _upload_file_from_str (gridfs,
+                          "file3",
+                          "Third file",
+                          tmp_bson ("{'metadata': {'testOrder': 3}}"),
+                          NULL);
+   _upload_file_from_str (gridfs,
+                          "file4",
+                          "Fourth file",
+                          tmp_bson ("{'metadata': {'testOrder': 4}}"),
+                          NULL);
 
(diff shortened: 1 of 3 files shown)


src/libmongoc/tests/test-mongoc-gridfs-bucket.c
```diff
@@ -952,6 +975,49 @@ test_find_w_session (void *ctx)
...
+   ASSERT_OR_PRINT (!mongoc_cursor_error (cursor, &error), error);
+
+   ok = mongoc_cursor_next (cursor, &found);
+   ASSERT (ok);
+   ok = bson_iter_init_find (&iter, found, "_id");
+   ASSERT (ok);
+   found_id = bson_iter_value (&iter);
+
+   _download_file_into_buf (gridfs, found_id, buffer, sizeof buffer);
+   ASSERT_CMPSTR (buffer, "Second file");
```
kevinAlbs, line 1010: This appears to be failing on the attached Evergreen patch build on this PR. Example: https://spruce.mongodb.com/task/mongo_c_driver_asan_ubuntu_test_asan_4.0_replica_set_auth_nosasl_openssl_patch_36189eb42f669e56c7b1d0620e005a9f93b3cb88_60dcb1bf30661514d7fa9a85_21_06_30_18_03_21/logs?execution=0

```
Assert Failure: "Third file" != "Second file"
```

Oddly, I checked this out locally and did not reproduce when running against a `5.0.0-alpha0-856-gf4e7955` sharded cluster.

kevinAlbs (commented):
Looking great so far!

src/libmongoc/tests/test-mongoc-gridfs-bucket.c
vector-of-bool: I noticed those failures, and I'm not sure what's happening either. This is what I would expect to see if the `sort` is not being honored while `skip` _is_.  It is failing identically for each u18 Asan task. I'm not sure how to reproduce this as I'm not yet familiar enough with the environments managed by Evergreen.

kevinAlbs: Ah, I think you were correct in that this test is interacting with others.

Running:
```
./cmake-build/src/libmongoc/test-libmongoc --no-fork -d -l "/gridfs/find*"
```

First runs the test `/gridfs/find_w_session`. I think you can drop the database first.

src/libmongoc/tests/test-mongoc-gridfs.c
```diff
@@ -1347,14 +1347,17 @@ test_set_id (void)
...
+   char *dbname;
    mongoc_gridfs_file_t *file;
    mongoc_gridfs_file_t *result;
    mongoc_gridfs_file_opt_t opt = {0};
 
    /* create new client and grab gridfs handle */
    client = test_framework_new_default_client ();
    ASSERT (client);
-   gridfs = mongoc_client_get_gridfs (client, "test", "fs", &error);
+   dbname = gen_collection_name ("test_set_id");
```
kevinAlbs, line 1358: Nice, good thinking.

src/libmongoc/tests/test-mongoc-gridfs-bucket.c
kevinAlbs: Nice find with CDRIVER-4050. Until CDRIVER-4050 is resolved, I suggest using another field aside from `uploadDate` to sort. This could set the `metadata` option: http://mongoc.org/libmongoc/current/mongoc_gridfs_bucket_open_upload_stream.html which will be applied to the file document.

kevinAlbs (changes requested):
Changes look good! I have a suggestion for an alternative of sorting by `uploadDate`.

kevinAlbs (approved):
LGTM! Feel free to "Squash and merge" and close the associated tickets.

PR #813, opened 2021-06-30
Sync SDAM spec tests

alcaeus:
This follows up CDRIVER-3997 with one last sync of the spec tests now that they have been merged. The main change is that the legacy_hello tests have been removed. This brings the tests up to date with mongodb/specifications@b07c26dc.

Note that since we don't have skipping logic for non-unified spec tests, I had to omit `monitoring/auth-*-error.json` and `integration/minPoolSize-error.json`. The spec changes required for these tests have not been implemented yet.


alcaeus:
The asan-ubuntu failures are due to CDRIVER-4035, the windows-2015 failure is due to CDRIVER-3998. This fixes a number of test failures that went undetected because we only added 5.0.0-rc0 in the meantime and I forgot to rebase my PR before merging it.

jmikola (approved):
Not about to read through the JSON files but the C diff looks correct as `legacy-hello/` is no longer a test directory.

PR #814, opened 2021-07-02
CDRIVER-4000 Test redaction of sensitive commands

alcaeus:
This PR takes care of CDRIVER-3797, CDRIVER-4000, and CDRIVER-4038.

Note that this PR only adds support for features introduced in the 1.5 version of unified test format. Changes for version 1.4 is part of the proxy testing work (CDRIVER-3886), while 1.3 is part of load balancer work (CDRIVER-3883).


src/libmongoc/src/mongoc/mongoc-apm.c
```diff
@@ -136,12 +191,23 @@ mongoc_apm_command_succeeded_init (mongoc_apm_command_succeeded_t *event,
                                    int64_t operation_id,
                                    const mongoc_host_list_t *host,
                                    uint32_t server_id,
+                                   bool force_redaction,
```
kevinAlbs, line 194: To check, is the rationale for `is_redacted` and `force_redaction` to avoid a needless call to `mongoc_apm_is_sensitive_reply`? I.e. if we know the command is redacted, we can redact the reply without having to inspect it.

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -579,6 +583,7 @@ mongoc_cluster_run_command_monitored (mongoc_cluster_t *cluster,
                                          cmd->operation_id,
                                          &server_stream->sd->host,
                                          server_id,
+                                         is_redacted,
```
kevinAlbs, line 586: Minor: I think there is one case where we over-redact. If the `hello` / `legacy hello` command includes `speculativeAuthenticate` but the server reply does not contain `speculativeAuthenticate`. We'll redact the reply regardless.

IMO it is very low impact. It would only affect users running `hello` with `speculativeAuthenticate` through the generic command helpers. Since it is low impact, I am OK with this behavior. I think it is worth noting that behavior in a comment here.

kevinAlbs: Edit: Nevermind... the spec seems clear that the reply should always be redacted if the command was redacted

> Some commands and replies will contain sensitive data and in order to not risk the leaking of this data to external sources or logs their commands AND replies MUST be redacted from the events.

src/libmongoc/src/mongoc/mongoc-apm-private.h
```diff
@@ -188,11 +193,25 @@ mongoc_apm_command_failed_init (mongoc_apm_command_failed_t *event,
...
 
 void
 mongoc_apm_command_failed_cleanup (mongoc_apm_command_failed_t *event);
 
+bool
+mongoc_apm_is_sensitive_command (const char *command_name,
+                                 const bson_t *command);
+
+void
+mongoc_apm_redact_command (bson_t *command);
```
kevinAlbs, line 207: I suggest moving the declarations of `mongoc_apm_redact_command` and `mongoc_apm_redact_reply` to the top of mongoc-apm.c and marking those static.

That makes it clearer to readers that these are helpers only needed in mongoc-apm.c, and not intended to be called elsewhere.

kevinAlbs (approved):
LGTM! I only have one question to verify my understanding and a request to move declarations.

src/libmongoc/src/mongoc/mongoc-cluster.c
alcaeus: Yeah, I had hoped to leave replies to `hello` unreacted if they don't contain any sensitive infos, but doing so would've required either changes in other driver, or having no test for it. I decided to always redact replies to sensitive hello commands here instead.

src/libmongoc/src/mongoc/mongoc-apm.c
alcaeus: That wasn't the implicit goal, but it's a nice side-effect. This was required to be able to redact replies to a `hello` command with speculative authentication even when the reply doesn't contain sensitive information (i.e., no `speculativeAuthenticate` field).

PR #818, opened 2021-07-06
Fix: CDRIVER-4050: Set the correct uploadDate on GridFS files

vector-of-bool:
Previously used the time from bson_get_monotonic_time(), which is
incorrect on most systems, as it is usually realtive to the system
uptime and is stored as microseconds, whereas uploadDate should be
stored as number of milliseconds since the Unix Epoch.

### DIFF
--- a/src/libbson/src/bson/bson-clock.c
+++ b/src/libbson/src/bson/bson-clock.c
@@ -101,32 +101,6 @@ bson_gettimeofday (struct timeval *tv) /* OUT */
 /*
  *--------------------------------------------------------------------------
  *
- * bson_get_real_time_ms --
- *
- * Returns:
- *       The current number of milliseconds that have elapsed since the
- *       Unix Epoch, or -1 on error
- *
- * Side effects:
- *       None.
- *
- *--------------------------------------------------------------------------
- */
-int64_t
-bson_get_real_time_ms (void)
-{
-   struct timeval tv;
-   const bool ok = bson_gettimeofday (&tv);
-   if (!ok) {
-      return -1;
-   }
-   return tv.tv_sec * 1000 + tv.tv_usec / 1000;
-}
-
-
-/*
- *--------------------------------------------------------------------------
- *
  * bson_get_monotonic_time --
  *
  *       Returns the monotonic system time, if available. A best effort is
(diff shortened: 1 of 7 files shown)


src/libbson/src/bson/bson-clock.c
```diff
@@ -98,6 +98,32 @@ bson_gettimeofday (struct timeval *tv) /* OUT */
...
+ *       The current number of milliseconds that have elapsed since the
+ *       Unix Epoch, or -1 on error
+ *
+ * Side effects:
+ *       None.
+ *
+ *--------------------------------------------------------------------------
+ */
+int64_t
+bson_get_real_time_ms (void)
```
kevinAlbs, line 116: I prefer not adding new API unless there is a known use case. We need to maintain the API until a future major release.

The C driver has two consumers: other wrapping drivers, and direct users. If this API helps either, I would be in favor of it. If not, I suggest moving this to internal API within libmongoc. This could be in mongoc-util-private.h or another (even new) private header.

If we do keep it, we should add RST documentation for the new symbol.

kevinAlbs (commented):
Looks good! I have a question about the new API.

kevinAlbs:
`/gridfs_old/create_from_stream` appears to be failing on the attached Evergreen patch build.

src/libmongoc/src/mongoc/mongoc-util.c
```diff
@@ -100,6 +100,17 @@ _mongoc_usleep (int64_t usec)
...
 
+int64_t
+_mongoc_get_real_time_ms (void)
+{
+   struct timeval tv;
+   const bool rc = bson_gettimeofday (&tv);
+   if (rc != 0) {
+      return -1;
+   }
+   return tv.tv_sec * 1000 + tv.tv_usec / 1000;
```
kevinAlbs, line 111: I suspect an overflow here is the root cause of `/unified/poc-gridfs` failing on Windows.

If so, an explicit cast `(int64_t) 1000` cast should resolve that.

Since it only appears on Windows, I suggest using the opportunity to set up a Windows development environment (either a non-expiring Windows spawn host, or a local Windows environment). It saves a lot of hassle when an issue only reproduces on Windows.

vector-of-bool: Locally tested and confirmed to fix.

vector-of-bool:
Rebased onto the recent master branch has test results in-sync with that branch. Should be ready to merge?

kevinAlbs (approved):
LGTM! Feel free to merge.

PR #820, opened 2021-07-08
Add missing features in unified test format

alcaeus:
CDRIVER-4066, CDRIVER-4060, CDRIVER-4064

While implementing the command redaction tests, I selectively implemented version 1.5 of the unified test format, ignoring 1.2, 1.3, and 1.4 as they weren't necessary. This PR fixes this.

For schema version 1.2, I've skipped implementing the `loop` operation and the `storeEventsAsEntities` option, as they are only required once we resume work on drivers atlas testing. We may decide to implement these earlier, but I didn't want to do so given the tight timeline on the other features.

In version 1.3 (load balancer support), the new `load-balanced` topology mode is accepted, but it obviously never matches. This will be added in a separate pull request related to load balancer.

### DIFF
--- a/src/libmongoc/tests/unified/entity-map.c
+++ b/src/libmongoc/tests/unified/entity-map.c
@@ -375,6 +375,8 @@ entity_client_new (entity_map_t *em, bson_t *bson, bson_error_t *error)
 
    if (store_events_as_entities) {
       /* TODO: CDRIVER-3867 Comprehensive Atlas Testing */
+      test_set_error (error, "storeEventsAsEntities is not supported");
+      goto done;
    }
 
    ret = true;
--- a/src/libmongoc/tests/unified/entity-map.c
+++ b/src/libmongoc/tests/unified/entity-map.c
@@ -257,6 +257,7 @@ entity_client_new (entity_map_t *em, bson_t *bson, bson_error_t *error)
    bson_t *uri_options = NULL;
    bool *use_multiple_mongoses = NULL;
    bson_t *observe_events = NULL;
+   bson_t *store_events_as_entities = NULL;
    bson_t *server_api = NULL;
    bool can_reduce_heartbeat = false;
    mongoc_server_api_t *api = NULL;
@@ -274,6 +275,8 @@ entity_client_new (entity_map_t *em, bson_t *bson, bson_error_t *error)
    bson_parser_doc_optional (parser, "serverApi", &server_api);
    bson_parser_bool_optional (
       parser, "observeSensitiveCommands", &entity->observe_sensitive_commands);
+   bson_parser_array_optional (
+      parser, "storeEventsAsEntities", &store_events_as_entities);
 
    if (!bson_parser_parse (parser, bson, error)) {
       goto done;
@@ -370,6 +373,10 @@ entity_client_new (entity_map_t *em, bson_t *bson, bson_error_t *error)
    }
    mongoc_client_set_apm_callbacks (client, callbacks, entity);
 
+   if (store_events_as_entities) {
+      /* TODO: CDRIVER-3867 Comprehensive Atlas Testing */
+   }
+
    ret = true;
 done:
    mongoc_uri_destroy (uri);
@@ -379,6 +386,7 @@ done:
    bson_destroy (uri_options);
    bson_free (use_multiple_mongoses);
    bson_destroy (observe_events);
+   bson_destroy (store_events_as_entities);
    bson_destroy (server_api);
    if (!ret) {
       entity_destroy (entity);
--- a/src/libmongoc/tests/unified/entity-map.c
+++ b/src/libmongoc/tests/unified/entity-map.c
@@ -883,6 +883,10 @@ entity_destroy (entity_t *entity)
       mongoc_gridfs_bucket_t *bucket = entity->value;
 
       mongoc_gridfs_bucket_destroy (bucket);
+   } else if (0 == strcmp ("cursor", entity->type)) {
+      mongoc_cursor_t *cursor = entity->value;
+
+      mongoc_cursor_destroy (cursor);
    } else {
       test_error ("Attempting to destroy unrecognized entity type: %s, id: %s",
                   entity->type,
@@ -916,6 +920,20 @@ entity_map_get (entity_map_t *entity_map, const char *id, bson_error_t *error)
 
    test_set_error (error, "Entity '%s' not found", id);
    return NULL;
+}
+
+bool
+entity_map_delete (entity_map_t *em, const char *id, bson_error_t *error)
+{
+   entity_t *entity = entity_map_get (em, id, error);
+   if (!entity) {
+      return false;
+   }
+
+   LL_DELETE (em->entities, entity);
+   entity_destroy (entity);
+
+   return true;
 }
 
 static entity_t *
@@ -990,6 +1008,18 @@ entity_map_get_changestream (entity_map_t *entity_map,
       return NULL;
    }
    return (mongoc_change_stream_t *) entity->value;
+}
+
+mongoc_cursor_t *
+entity_map_get_cursor (entity_map_t *entity_map,
+                       const char *id,
+                       bson_error_t *error)
+{
+   entity_t *entity = _entity_map_get_by_type (entity_map, id, "cursor", error);
+   if (!entity) {
+      return NULL;
+   }
+   return (mongoc_cursor_t *) entity->value;
 }
 
 bson_val_t *
@@ -1085,6 +1115,15 @@ entity_map_add_changestream (entity_map_t *em,
 {
    return _entity_map_add (
       em, id, "changestream", (void *) changestream, error);
+}
+
+bool
+entity_map_add_cursor (entity_map_t *em,
+                       const char *id,
+                       mongoc_cursor_t *cursor,
+                       bson_error_t *error)
+{
+   return _entity_map_add (em, id, "cursor", (void *) cursor, error);
 }
 
 bool
--- a/src/libmongoc/tests/unified/operation.c
+++ b/src/libmongoc/tests/unified/operation.c
@@ -28,6 +28,7 @@ typedef struct {
    bson_t *arguments;
    bson_t *expect_error;
    bson_val_t *expect_result;
+   bool *ignore_result_and_error;
    char *save_result_as_entity;
    bson_parser_t *parser;
    char *session_id;
@@ -55,6 +56,8 @@ operation_new (bson_t *bson, bson_error_t *error)
    bson_parser_doc_optional (op->parser, "arguments", &op->arguments);
    bson_parser_doc_optional (op->parser, "expectError", &op->expect_error);
    bson_parser_any_optional (op->parser, "expectResult", &op->expect_result);
+   bson_parser_bool_optional (
+      op->parser, "ignoreResultAndError", &op->ignore_result_and_error);
    bson_parser_utf8_optional (
       op->parser, "saveResultAsEntity", &op->save_result_as_entity);
    if (!bson_parser_parse (op->parser, bson, error)) {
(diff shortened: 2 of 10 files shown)


src/libmongoc/tests/unified/runner.c
```diff
@@ -716,6 +716,33 @@ check_run_on_requirement (test_runner_t *test_runner,
          continue;
       }
 
+      if (0 == strcmp (key, "serverless")) {
+         const char *serverless_mode = bson_iter_utf8 (&req_iter, NULL);
+
+         if (0 == strcmp (serverless_mode, "allow")) {
+            return true;
```
kevinAlbs, line 723: This return early may skip additional run on requirements to check.
```suggestion
            continue;
```

src/libmongoc/tests/unified/runner.c
```diff
@@ -716,6 +716,33 @@ check_run_on_requirement (test_runner_t *test_runner,
...
+            }
+
+            return true;
+         } else if (0 == strcmp (serverless_mode, "forbid")) {
+            if (test_runner->is_serverless) {
+               *fail_reason = bson_strdup_printf ("Running in serverless mode");
+               return false;
+            }
+
+            return true;
```
kevinAlbs, line 738:
```suggestion
            continue;
```

src/libmongoc/tests/unified/runner.c
```diff
@@ -716,6 +716,33 @@ check_run_on_requirement (test_runner_t *test_runner,
...
+         if (0 == strcmp (serverless_mode, "allow")) {
+            return true;
+         } else if (0 == strcmp (serverless_mode, "require")) {
+            if (!test_runner->is_serverless) {
+               *fail_reason =
+                  bson_strdup_printf ("Not running in serverless mode");
+               return false;
+            }
+
+            return true;
```
kevinAlbs, line 731:
```suggestion
            continue;
```

kevinAlbs (approved):
LGTM! I only have suggested tweaks to the runOnRequirement checks.

src/libmongoc/tests/unified/runner.c
jmikola: I can't resist ([PHPLIB-680](https://jira.mongodb.org/browse/PHPLIB-680)).

alcaeus: I blame years of writing early exit clauses due to https://github.com/slevomat/coding-standard#slevomatcodingstandardcontrolstructuresearlyexit-

On a separate note, I had half a mind of extracting checks to static functions to ensure such problems no longer happen. I'll do that in a separate PR as I've noticed it in several places.

src/libmongoc/tests/unified/entity-map.c
```diff
@@ -977,6 +1003,18 @@ entity_map_get_changestream (entity_map_t *entity_map,
    return (mongoc_change_stream_t *) entity->value;
 }
 
+mongoc_cursor_t *
+entity_map_get_cursor (entity_map_t *entity_map,
+                       const char *id,
+                       bson_error_t *error)
+{
+   entity_t *entity = _entity_map_get_by_type (entity_map, id, "cursor", error);
```
jmikola, line 1011: Seems odd to me that string literals are used for entity types. I filed [CDRIVER-4073](https://jira.mongodb.org/browse/CDRIVER-4073) for that, but it's surely outside the scope of this PR.

src/libmongoc/tests/unified/operation.c
```diff
@@ -1452,6 +1535,29 @@ operation_iterate_until_document_or_error (test_t *test,
    return ret;
 }
 
+static bool
+operation_close (test_t *test,
```
jmikola, line 1539: This operation is just for cursor types. Should the function be named `operation_cursor_close` instead?

Edit: I just realized that isn't typically done for other operation methods, so this is fine as-is.

src/libmongoc/tests/unified/operation.c
```diff
@@ -1452,6 +1535,29 @@ operation_iterate_until_document_or_error (test_t *test,
...
+{
+   bool ret = false;
+   mongoc_cursor_t *cursor = NULL;
+
+   cursor = entity_map_get_cursor (test->entity_map, op->object, error);
+   if (!cursor) {
+      goto done;
+   }
+
+   entity_map_delete (test->entity_map, op->object, error);
```
jmikola, line 1552: This borrowed logic from [PHPLIB-669](https://jira.mongodb.org/browse/PHPLIB-669) where deleting the Cursor and triggering garbage collection was the only way for us to invoke `mongoc_cursor_destroy`. libmongoc doesn't have the same technical limitation. Is this preferable to calling `_mongoc_client_kill_cursor` (since that's still a private API)?

I suppose there's no value in having the cursor entity stick around otherwise. Perhaps it'd be good to clarify this in the unified spec: [DRIVERS-1841](https://jira.mongodb.org/browse/DRIVERS-1841)

src/libmongoc/tests/unified/runner.c
```diff
@@ -64,14 +64,10 @@ skipped_unified_test_t SKIPPED_TESTS[] = {
    {"poc-crud", SKIP_ALL_TESTS},
    {"db-aggregate", SKIP_ALL_TESTS},
    {"mongos-unpin", SKIP_ALL_TESTS},
-   /* CDRIVER-3883: load balancer support (schema version 1.3) */
+   /* CMAP is not implemented */
```
jmikola, line 67: This can reference [CDRIVER-2871](https://jira.mongodb.org/browse/CDRIVER-2871)

src/libmongoc/tests/unified/operation.c
```diff
@@ -649,6 +652,65 @@ operation_count_documents (test_t *test,
...
+      }
+   }
+
+   cursor = mongoc_collection_find_with_opts (
+      coll, filter, opts, NULL /* read prefs */);
+
+
+   if (!op->save_result_as_entity) {
+      test_set_error (error,
+                      "unexpected createFindCursor does not save result");
```
jmikola, line 695: [createFindCursor](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#createfindcursor) doesn't actually require `saveResultAsEntity`, so I don't think this is necessary. Did you do this to avoid a potential leak if the entity wasn't going to be saved?

I think without `saveResultAsEntity`, you could just destroy the cursor immediately. A test might still want to do this to assert that a `find` command is issued.

src/libmongoc/tests/unified/operation.c
```diff
@@ -2485,6 +2605,9 @@ operation_run (test_t *test, bson_t *op_bson, bson_error_t *error)
          }
          if (!result_check (result,
                             test->entity_map,
+                            op->ignore_result_and_error
+                               ? *op->ignore_result_and_error
+                               : false,
```
jmikola, line 2610: Just to confirm, `result_check` is used for _both_ result _and_ error checks, correct? So this instructs it to be a no-op? Any reason you don't just avoid calling `result_check` altogether if `ignore_result_and_error` is set? Do we rely on `result_check` to catch other test format errors that aren't considered `expectError` (e.g. bad syntax)?

src/libmongoc/tests/unified/entity-map.c
```diff
@@ -370,6 +373,10 @@ entity_client_new (entity_map_t *em, bson_t *bson, bson_error_t *error)
    }
    mongoc_client_set_apm_callbacks (client, callbacks, entity);
 
+   if (store_events_as_entities) {
+      /* TODO: CDRIVER-3867 Comprehensive Atlas Testing */
```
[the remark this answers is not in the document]
alcaeus: That's a good point, I'll change this to errors.

src/libmongoc/tests/unified/operation.c
alcaeus: Yeah, this surprised me as well, as the operations don't seem to be tied to objects. @kevinAlbs want me to create a ticket for this?

kevinAlbs: The name of the operation functions should be 1:1 with the operation names from UTF, e.g. `operationFooBar` => `operation_foo_bar`.

I think the operation names must be unique in UTF. `operation_close` seems correct.

jmikola: > I think the operation names must be unique in UTF.

FWIW, this is not actually a requirement of the spec even if the names might appear unique in practice (if you're considering the variations of `aggregate` to be the same). I suppose we can punt on this until it actually becomes an issue.

src/libmongoc/tests/unified/operation.c
alcaeus: I've changed this to not call `result_check`. There was no reason for doing it that way except that I didn't think of any other way...

src/libmongoc/tests/unified/operation.c
alcaeus: I was considering this, but decided that `entity_map_delete` is the preferred way in C as well:
* We don't know what kind of cursor implementation backs the `mongoc_cursor_t` - they each have their own destruction logic, which can only be triggered through `mongoc_cursor_destroy`
* Calling `mongoc_cursor_destroy` destroys everything related to the cursor, meaning that we shouldn't be holding a reference to it anymore, hence the `entity_map_delete`.

If @kevinAlbs thinks it's safe to always call `killCursors` and leave the (now closed) cursor in the entity map, I suppose using `_mongoc_client_kill_cursor` would be fine.

jmikola: Good point about different cursor implementations. I'm fine with keeping this as-is.

src/libmongoc/tests/unified/operation.c
```diff
@@ -689,10 +689,13 @@ operation_create_find_cursor (test_t *test,
    cursor = mongoc_collection_find_with_opts (
       coll, filter, opts, NULL /* read prefs */);
 
+   ret = true;
+
+   mongoc_cursor_error_document (cursor, &op_error, &op_reply);
+   result_from_val_and_reply (result, NULL, (bson_t *) op_reply, &op_error);
```
jmikola, line 695: Is it significant that you swapped the order of `ret = true;` here vs. the original code?

src/libmongoc/tests/unified/operation.c
```diff
@@ -28,6 +28,7 @@ typedef struct {
    bson_t *arguments;
    bson_t *expect_error;
    bson_val_t *expect_result;
+   bool *ignore_result_and_error;
```
jmikola, line 31: Just noticed this is a pointer (based on how `bson_parser_bool_optional` works). Do these fields get freed anywhere? I don't see any other reference to this field beyond parsing and reading it, but I assume `bson_parser_destroy_with_parsed_fields` takes care of this?

src/libmongoc/tests/unified/result.c
```diff
@@ -564,7 +558,7 @@ test_resultfrombulkwrite (void)
       "'modifiedCount': 3, 'upsertedCount': 4}");
    result_from_bulk_write (result, reply, &empty);
    MONGOC_DEBUG ("rewritten to: %s", bson_val_to_json (result->value));
-   if (!result_check (result, false, NULL, expect, NULL, &error)) {
+   if (!result_check (result, NULL, expect, NULL, &error)) {
```
jmikola: I realize `false` was incorrect since the second param is a pointer to the entity map. Is `NULL` acceptable here because this isn't part of the general unified test runner (where an operator might need the entity map) and instead just an isolated unit test?

jmikola (approved):
Some questions but LGTM if the answers don't reveal anything.

src/libmongoc/tests/unified/result.c
alcaeus: This particular diff was from https://github.com/mongodb/mongo-c-driver/pull/820/commits/7679dc7085cbd178f00062f07fe9195724244db3 where I simplified the implementation. The `NULL` was there before, and I assumed the same that you do. @kevinAlbs can you confirm it's intentional that we're not passing the entity map?

jmikola: Oh, in that case this may be like what we do in [tests/UnifiedSpecTests/Constraint/MatchesTest.php](https://github.com/mongodb/mongo-php-library/blob/master/tests/UnifiedSpecTests/Constraint/MatchesTest.php). The constraint tests that don't actually need to resolve anything in the EntityMap will pass in null.

kevinAlbs: It is intentional. This is a self test of the `result_check` method. @jmikola is correct. The test assertion is not referencing anything in the entity map (e.g. `$$matchesEntity`)

src/libmongoc/tests/unified/operation.c
alcaeus: In this case, `bson_parser_entry_destroy` handles freeing the pointer: https://github.com/mongodb/mongo-c-driver/blob/d8684bed2f5d0763aceb8c03cbce1e2637bc515c/src/libmongoc/tests/bsonutil/bson-parser.c#L253..L258

src/libmongoc/tests/unified/operation.c
alcaeus: No particular reason - want me to move it for consistency?

jmikola: If it was done inadvertently, I think we should leave it in the original form. That would avoid a potential question down the line if someone comes across the diff again.

alcaeus: Updated in the last commit.

PR #821, opened 2021-07-08
CDRIVER-4054 support sessions in loadbalanced mode

kevinAlbs:
- Add MONGOC_TOPOLOGY_LOAD_BALANCED type
- Add MONGOC_SERVER_LOAD_BALANCER type
- Do not expire sessions in lb mode
- Always support sessions in lb mode


src/libmongoc/src/mongoc/mongoc-cmd.c
```diff
@@ -626,6 +626,7 @@ _mongoc_cmd_parts_assemble_mongod (mongoc_cmd_parts_t *parts,
          break;
       case MONGOC_TOPOLOGY_SHARDED:
       case MONGOC_TOPOLOGY_UNKNOWN:
+      case MONGOC_TOPOLOGY_LOAD_BALANCED:
```
kevinAlbs, line 629: This PR only covers updates to sessions behavior. But adding this case was necessary to avoid compiler warnings for a missing case on an enum value.

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -413,7 +413,10 @@ mongoc_topology_new (const mongoc_uri_t *uri, bool single_threaded)
       has_directconnection &&
       mongoc_uri_get_option_as_bool (uri, MONGOC_URI_DIRECTCONNECTION, false);
    hl = mongoc_uri_get_hosts (topology->uri);
-   if (service && !has_directconnection) {
+   if (mongoc_uri_get_option_as_bool (
+          topology->uri, MONGOC_URI_LOADBALANCED, false)) {
+      init_type = MONGOC_TOPOLOGY_LOAD_BALANCED;
```
kevinAlbs, line 418: Disabling SDAM monitoring will be in a follow-up PR. The initial type only applies until monitoring occurs. This seems sufficient to test the new sessions behavior. Starting and destroying a session does not kick off monitoring. Ending sessions does, but that behavior has not changed.

src/libmongoc/tests/test-mongoc-loadbalanced.c
```diff
@@ -0,0 +1,103 @@
...
+#include "mongoc/mongoc.h"
+#include "mongoc/mongoc-client-session-private.h"
+#include "test-conveniences.h"
+#include "test-libmongoc.h"
+#include "TestSuite.h"
+
+static char *loadbalanced_uri (void) {
+   /* TODO (CDRIVER-4062): This will need to add TLS and auth to the URI when
+    * run in evergreen. */
+   return test_framework_getenv ("SINGLE_MONGOS_LB_URI");
```
kevinAlbs, line 26: I used these to validate locally. I have noted in the description of CDRIVER-4062 to additionally run tests in the `/loadbalanced` test namespace.

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -1436,45 +1434,53 @@ _mongoc_topology_pop_server_session (mongoc_topology_t *topology,
...
-      /* if needed, connect and check for session timeout again */
-      if (!mongoc_topology_description_has_data_node (td)) {
-         bson_mutex_unlock (&topology->mutex);
-         if (!mongoc_topology_select_server_id (
-                topology, MONGOC_SS_READ, NULL, error)) {
-            RETURN (NULL);
-         }
+   /* When the topology type is LoadBalanced, sessions are always supported. */
+   if (!loadbalanced) {
+      if (timeout == MONGOC_NO_SESSIONS) {
```
jmikola, line 1449: Since we have two nested `if`s and no nested `else`, would it make sense to just change the condition here and minimize the diff?

```c
if (!loadbalanced && timeout == MONGOC_NO_SESSIONS) {
```

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -1436,45 +1434,53 @@ _mongoc_topology_pop_server_session (mongoc_topology_t *topology,
...
    while (topology->session_pool) {
       ss = topology->session_pool;
       CDL_DELETE (topology->session_pool, ss);
-      if (_mongoc_server_session_timed_out (ss, timeout)) {
-         _mongoc_server_session_destroy (ss);
-         ss = NULL;
-      } else {
-         break;
+      /* Sessions do not expire when the topology type is load balanced. */
+      if (!loadbalanced) {
```
jmikola, line 1477: To minimize the diff and conditional nesting, I'd suggest:

```c
if (!loadBalanced) {
   continue;
}
```

I assume you still want to clear the pool regardless of topology. It's just killing of sessions that you want to avoid.

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -1503,17 +1509,20 @@ _mongoc_topology_push_server_session (mongoc_topology_t *topology,
...
 
    timeout = topology->description.session_timeout_minutes;
+   loadbalanced = topology->description.type == MONGOC_TOPOLOGY_LOAD_BALANCED;
 
    /* start at back of queue and reap timed-out sessions */
    while (topology->session_pool && topology->session_pool->prev) {
       ss = topology->session_pool->prev;
-      if (_mongoc_server_session_timed_out (ss, timeout)) {
+      /* Sessions do not expire when the topology type is load balanced. */
+      if (!loadbalanced && _mongoc_server_session_timed_out (ss, timeout)) {
```
jmikola, line 1525: Why did you combine the `if` condition here but not in `_mongoc_topology_pop_server_session`?

In `_mongoc_topology_pop_server_session`, the original logic to kill the session or `break` _only_ happens when `loadbalanced` is false. When `loadbalanced` is true you just end up clearing the session pool locally.

Here, sessions never get killed if `loadbalanced` is true, but you're also leaving the session pool as-is and _not_ clearing it.

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -1524,8 +1533,10 @@ _mongoc_topology_push_server_session (mongoc_topology_t *topology,
    }
 
    /* If session is expiring or "dirty" (a network error occurred on it), do not
-    * return it to the pool. */
-   if (_mongoc_server_session_timed_out (server_session, timeout) ||
+    * return it to the pool. Sessions do not expire when the topology type is
+    * load balanced. */
+   if ((!loadbalanced &&
+        _mongoc_server_session_timed_out (server_session, timeout)) ||
```
jmikola, line 1539: Given how frequently you combine logic for `loadbalanced` and `_mongoc_server_session_timed_out`, would it make sense to have `_mongoc_server_session_timed_out` aware of the topology and just return `false` on its own for `loadbalanced` topologies?

I'm curious if `_mongoc_server_session_timed_out` is called from anywhere else that might need to care about this.

On the flip side, I understand doing so might violate a separation of concerns if sessions otherwise have no reason to know about the topology in which they were created.

src/libmongoc/tests/test-mongoc-loadbalanced.c
```diff
@@ -0,0 +1,103 @@
+/*
+ * Copyright 2020-present MongoDB, Inc.
```
jmikola, line 2: New file uses 2021.

src/libmongoc/tests/test-mongoc-loadbalanced.c
```diff
@@ -0,0 +1,103 @@
...
+   bson_error_t error;
+   bson_t *lsid1;
+   bson_t *lsid2;
+
+   client = mongoc_client_new (uristr);
+   session = mongoc_client_start_session (client, NULL /* opts */, &error);
+   ASSERT_OR_PRINT (session, error);
+
+   lsid1 = bson_copy (&session->server_session->lsid);
+   session->server_session->last_used_usec = 1;
```
jmikola, line 61: `bson_get_monotonic_time` is Unix time in microseconds, so `1` is always in the past, correct?

src/libmongoc/tests/test-mongoc-loadbalanced.c
```diff
@@ -0,0 +1,103 @@
...
+   mongoc_client_session_destroy (session);
```
jmikola, line 62: Is this related to you intentionally not clearing the pool during `_mongoc_topology_push_server_session` when `loadbalanced` is true? That in turn ensures the session _does_ get reused?

I may still be confused as to why the pool is cleared during the `pop` method, though.

src/libmongoc/tests/test-mongoc-loadbalanced.c
```diff
@@ -0,0 +1,103 @@
...
+   }
+
+   bson_destroy (lsid1);
+   bson_destroy (lsid2);
+   bson_free (uristr);
+   mongoc_client_destroy (client);
+}
+
+static int
+skip_if_not_loadbalanced (void)
```
jmikola, line 80: `skip_if` functions are all expected to return `int` instead of `bool`? This might be the first time I've noticed.

src/libmongoc/src/mongoc/mongoc-topology.c
kevinAlbs: I would prefer avoiding any circular dependency. The topology maintains the session pool, and the client session objects do not know about the topology object. `mongoc-topology-private.h` includes `mongoc-client-session-private.h`.

src/libmongoc/tests/test-mongoc-loadbalanced.c
kevinAlbs: Done.

src/libmongoc/tests/test-mongoc-loadbalanced.c
kevinAlbs: Correct.

src/libmongoc/tests/test-mongoc-loadbalanced.c
kevinAlbs: This test has been updated to ensure that the pool is not cleared in `pop`.

src/libmongoc/tests/test-mongoc-loadbalanced.c
kevinAlbs: Weirdly yes, to match the `typedef int (*CheckFunc) (void);` in `TestSuite.h`.

src/libmongoc/tests/test-mongoc-loadbalanced.c
```diff
@@ -0,0 +1,165 @@
...
+   session1_lsid = bson_copy (mongoc_client_session_get_lsid (session1));
+   session2_lsid = bson_copy (mongoc_client_session_get_lsid (session2));
+
+   /* Expire both sessions. */
+   session1->server_session->last_used_usec = 1;
+   session2->server_session->last_used_usec = 1;
+   mongoc_client_session_destroy (session1);
+   mongoc_client_session_destroy (session2);
+
+   /* Get a new session, it should reuse the most recently pushed session2. */
```
jmikola, line 76: This test makes more sense now :+1:

jmikola (approved):

PR #822, opened 2021-07-10
CDRIVER-3653 use handshake metadata for connection checks

kevinAlbs:
**Terms**
- **handshake description** is a server description constructed during connection handshake.
- **monitor description** is a server description constructed and updated in server monitoring.

**Motivation**
- libmongoc currently uses monitor descriptions when constructing commands (to check wire version, how to pass read preferences, etc.). This is racy, and the cause of bugs like CDRIVER-3404.
- Supporting load balancer requires setting the monitor description to an empty LoadBalancer description. A load balancer can front many backing servers. Connections must track the handshake description to work correctly with load balancers.

**Summary**
- Store handshake descriptions alongside their corresponding connections.
- Update change stream and transactions wire version checks to check the handshake description instead of the monitor description.
- Update auth commands to use the handshake description when constructing commands.
- Add `mongoc_client_get_handshake_description` for wrapping drivers to obtain a handshake description.

**Notes**
- As a happy consequence, change streams now send a killCursors command when resuming due to an error that marks a server unknown. Previously this would clear the wire version on the connection.

### DIFF
--- a/src/libmongoc/doc/mongoc_client_get_handshake_description.rst
+++ b/src/libmongoc/doc/mongoc_client_get_handshake_description.rst
@@ -21,21 +21,14 @@ Description
 
 :symbol:`mongoc_client_get_handshake_description` is distinct from :symbol:`mongoc_client_get_server_description`. :symbol:`mongoc_client_get_server_description` returns a server description constructed from monitoring, which may differ from the server description constructed from the connection handshake.
 
-Use this function only for building a language driver that wraps the C Driver. When writing applications in C, higher-level functions automatically select a suitable server.
+:symbol:`mongoc_client_get_handshake_description` will attempt to establish a connection to the server if a connection was not already established. It will perform the MongoDB handshake and authentication if required.
 
-:symbol:`mongoc_client_get_handshake_description` does not attempt to establish a connection to the server if a connection was not already established. If a connection has not been established, this returns ``NULL`` and sets ``error``.
+Use this function only for building a language driver that wraps the C Driver. When writing applications in C, higher-level functions automatically select a suitable server.
 
 Single-threaded client behavior
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-To ensure a connection has been established ensure that the server is selectable (e.g. by calling :symbol:`mongoc_client_select_server`) or that an operation has successfully run on the server (e.g. by sending "ping").
 
-If the established connection has not completed authentication, calling :symbol:`mongoc_client_get_handshake_description` will complete authentication on the connection.
-
-Single-threaded clients only have one active connection to each server. The one connection is used for both monitoring and application operations. However, the server description returned by :symbol:`mongoc_client_get_handshake_description` may still differ from :symbol:`mongoc_client_get_server_description`. Notably, if connected to a load balanced cluster, the :symbol:`mongoc_client_get_server_description` will describe the load balancer server (:symbol:`mongoc_server_description_type` will return "LoadBalancer"). And :symbol:`mongoc_client_get_handshake_description` will describe the backing server.
-
-Pooled client behavior
-^^^^^^^^^^^^^^^^^^^^^^
-To ensure a connection has been established ensure an operation has successfully run on a server (e.g. by sending "ping").
+Single-threaded clients only have one active connection to each server. The one connection is used for both monitoring and application operations. However, the server description returned by :symbol:`mongoc_client_get_handshake_description` may still differ from the server description returned by :symbol:`mongoc_client_get_server_description`. Notably, if connected to a load balanced cluster, the :symbol:`mongoc_client_get_server_description` will describe the load balancer server (:symbol:`mongoc_server_description_type` will return "LoadBalancer"). And :symbol:`mongoc_client_get_handshake_description` will describe the backing server.
 
 Parameters
 ----------
@@ -48,7 +41,7 @@ Parameters
 Returns
 -------
 
-A :symbol:`mongoc_server_description_t` that must be freed with :symbol:`mongoc_server_description_destroy`. If a connection has not been successfully established to a server, returns NULL and ``error`` is filled out.
+A :symbol:`mongoc_server_description_t` that must be freed with :symbol:`mongoc_server_description_destroy`. If a connection has not been successfully established to a server, returns ``NULL`` and ``error`` is filled out.
 
 
 See Also
--- a/src/libmongoc/doc/mongoc_client_get_handshake_description.rst
+++ b/src/libmongoc/doc/mongoc_client_get_handshake_description.rst
@@ -28,7 +28,7 @@ Use this function only for building a language driver that wraps the C Driver. W
 Single-threaded client behavior
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Single-threaded clients only have one active connection to each server. The one connection is used for both monitoring and application operations. However, the server description returned by :symbol:`mongoc_client_get_handshake_description` may still differ from the server description returned by :symbol:`mongoc_client_get_server_description`. Notably, if connected to a load balanced cluster, the :symbol:`mongoc_client_get_server_description` will describe the load balancer server (:symbol:`mongoc_server_description_type` will return "LoadBalancer"). And :symbol:`mongoc_client_get_handshake_description` will describe the backing server.
+Single-threaded clients only have one active connection to each server. The one connection is used for both monitoring and application operations. However, the server description returned by :symbol:`mongoc_client_get_handshake_description` may still differ from the server description returned by :symbol:`mongoc_client_get_server_description`. Notably, if connected to a load balanced cluster, the server description returned by :symbol:`mongoc_client_get_server_description` will describe the load balancer server (:symbol:`mongoc_server_description_type` will return "LoadBalancer"). And the server description returned by :symbol:`mongoc_client_get_handshake_description` will describe the backing server.
 
 Parameters
 ----------
--- a/src/libmongoc/doc/mongoc_client_get_handshake_description.rst
+++ b/src/libmongoc/doc/mongoc_client_get_handshake_description.rst
@@ -21,13 +21,21 @@ Description
 
 :symbol:`mongoc_client_get_handshake_description` is distinct from :symbol:`mongoc_client_get_server_description`. :symbol:`mongoc_client_get_server_description` returns a server description constructed from monitoring, which may differ from the server description constructed from the connection handshake.
 
-:symbol:`mongoc_client_get_handshake_description` does not attempt to establish a connection to the server if a connection was not already established. It will complete authentication on a single-threaded client if the connection has not yet been used by an operation.
-
-To establish a connection on a single-threaded client, ensure the server is selectable (the monitoring connection is the only connection made to the server). This can be done with a function like :symbol:`mongoc_client_select_server`, or any function which performs server selection.
-
-To establish a connection on a pooled client, send a command (e.g. ping) to the server.
-
 Use this function only for building a language driver that wraps the C Driver. When writing applications in C, higher-level functions automatically select a suitable server.
+
+:symbol:`mongoc_client_get_handshake_description` does not attempt to establish a connection to the server if a connection was not already established. If a connection has not been established, this returns ``NULL`` and sets ``error``.
+
+Single-threaded client behavior
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+To ensure a connection has been established ensure that the server is selectable (e.g. by calling :symbol:`mongoc_client_select_server`) or that an operation has successfully run on the server (e.g. by sending "ping").
+
+If the established connection has not completed authentication, calling :symbol:`mongoc_client_get_handshake_description` will complete authentication on the connection.
+
+Single-threaded clients only have one active connection to each server. The one connection is used for both monitoring and application operations. However, the server description returned by :symbol:`mongoc_client_get_handshake_description` may still differ from :symbol:`mongoc_client_get_server_description`. Notably, if connected to a load balanced cluster, the :symbol:`mongoc_client_get_server_description` will describe the load balancer server (:symbol:`mongoc_server_description_type` will return "LoadBalancer"). And :symbol:`mongoc_client_get_handshake_description` will describe the backing server.
+
+Pooled client behavior
+^^^^^^^^^^^^^^^^^^^^^^
+To ensure a connection has been established ensure an operation has successfully run on a server (e.g. by sending "ping").
 
 Parameters
 ----------
@@ -48,3 +56,4 @@ See Also
 
 - :symbol:`mongoc_client_select_server` To select a server from read preferences.
 - :symbol:`mongoc_client_get_server_description` To obtain the server description from monitoring for a server.
+- :symbol:`mongoc_server_description_type` To obtain the type of server from a server description.
--- a/src/libmongoc/doc/mongoc_client_get_handshake_description.rst
+++ b/src/libmongoc/doc/mongoc_client_get_handshake_description.rst
@@ -1,0 +1,50 @@
+:man_page: mongoc_client_get_handshake_description
+
+mongoc_client_get_handshake_description()
+=========================================
+
+Synopsis
+--------
+
+.. code-block:: c
+
+   mongoc_server_description_t *
+   mongoc_client_get_handshake_description (mongoc_client_t *client,
+                                            uint32_t server_id,
+                                            bson_t *opts,
+                                            bson_error_t *error);
+
+Returns a description constructed from the initial handshake response to a server.
+
+Description
+-----------
+
+:symbol:`mongoc_client_get_handshake_description` is distinct from :symbol:`mongoc_client_get_server_description`. :symbol:`mongoc_client_get_server_description` returns a server description constructed from monitoring, which may differ from the server description constructed from the connection handshake.
+
+:symbol:`mongoc_client_get_handshake_description` does not attempt to establish a connection to the server if a connection was not already established. It will complete authentication on a single-threaded client if the connection has not yet been used by an operation.
+
+To establish a connection on a single-threaded client, ensure the server is selectable (the monitoring connection is the only connection made to the server). This can be done with a function like :symbol:`mongoc_client_select_server`, or any function which performs server selection.
+
+To establish a connection on a pooled client, send a command (e.g. ping) to the server.
+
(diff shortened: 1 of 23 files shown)


src/libmongoc/tests/test-mongoc-change-stream.c
```diff
@@ -2490,6 +2490,13 @@ prose_test_17 (void)
    request_destroy (request);
 
    /* Resume occurs. */
```
kevinAlbs, line 2492: The "not primary" error above causes the monitor description to be marked Unknown. During the resume process `mongoc_cursor_destroy` would fail to construct a correct kill cursors command since the server description was Unknown.

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -3996,6 +3997,96 @@ test_mongoc_client_recv_network_error (void)
...
+test_mongoc_client_get_handshake_hello_response_pooled (void)
+{
+   mongoc_client_t *client;
+   mongoc_server_description_t *monitor_sd;
+   mongoc_server_description_t *invalidated_sd;
+   mongoc_server_description_t *handshake_sd;
+   bson_error_t error = {0};
+   bool ret;
+
+   client = test_framework_new_default_client ();
```
jmikola, line 4050: What about this test is pooled? You're fetching a `client` in the same manner that was done in `test_mongoc_client_get_handshake_hello_response_single` above.

src/libmongoc/tests/test-mongoc-cluster.c
```diff
@@ -1746,7 +1753,7 @@ _test_cmd_on_unknown_serverid (bool pooled)
    uri = test_framework_get_uri ();
    /* Set a high heartbeatFrequencyMS so subsequent topology scans do not
     * interfere with the test. */
-   mongoc_uri_set_option_as_int32 (uri, MONGOC_URI_HEARTBEATFREQUENCYMS, 99999);
+   mongoc_uri_set_option_as_int32 (uri, MONGOC_URI_HEARTBEATFREQUENCYMS, 5000);
```
jmikola, line 1756: Does the comment above need to be changed? The default is 10000 for pooled and 60000 for single-threaded, so this is no longer a "high heartbeatFrequencyMS". What was the reason for lowering this vs. removing it entirely?

src/libmongoc/tests/test-mongoc-cluster.c
```diff
@@ -1813,12 +1811,114 @@ _test_cmd_on_unknown_serverid (bool pooled)
...
+      client, true /* for writes */, NULL /* read prefs */, &error);
+   ASSERT_OR_PRINT (sd, error);
+
+   /* Test "clearing the pool". This should invalidate existing server streams.
+    */
+   stream = mongoc_cluster_stream_for_writes (
+      &client->cluster, NULL /* session */, NULL /* reply */, &error);
+   ASSERT_OR_PRINT (stream, error);
+   BSON_ASSERT (mongoc_cluster_stream_valid (&client->cluster, stream));
+   _mongoc_topology_clear_connection_pool (client->topology,
```
jmikola, line 1847: It's suitable to call this method in single-threaded mode? I suppose it doesn't care about how many streams are actually present and just wipes everything?

src/libmongoc/tests/test-mongoc-cluster.c
```diff
@@ -1813,12 +1811,114 @@ _test_cmd_on_unknown_serverid (bool pooled)
...
+static void
+test_cluster_stream_invalidation_pooled (void)
+{
+   mongoc_client_pool_t *pool;
+   mongoc_client_t *client;
+   mongoc_server_description_t *sd;
+   bson_error_t error;
+   mongoc_server_stream_t *stream;
+
+   pool = test_framework_new_default_client_pool ();
```
jmikola, line 1883: Unlike my previous comment in `test-mongoc-client.c`, you _are_ using `test_framework_new_default_client_pool` instead of `test_framework_new_default_client` here. So I assume the other test might have had a mistake.

src/libmongoc/tests/test-mongoc-server-stream.c
```diff
@@ -0,0 +1,227 @@
...
+                                          "admin",
+                                          tmp_bson ("{'ping': 1}"),
+                                          NULL /* read prefs */,
+                                          NULL /* reply */,
+                                          &error);
+   /* The first command on a pooled client creates a new connection. */
+   request = mock_server_receives_legacy_hello (server, NULL);
+   mock_server_replies_simple (request, HELLO_POST_OPMSG);
+   request_destroy (request);
+   /* Check that the mock server recieves an OP_MSG. */
```
jmikola, line 93:
```suggestion
   /* Check that the mock server receives an OP_MSG. */
```

src/libmongoc/tests/test-mongoc-server-stream.c
```diff
@@ -0,0 +1,227 @@
...
+                                          NULL /* read prefs */,
+                                          NULL /* reply */,
+                                          &error);
+   request = mock_server_receives_msg (server, 0, tmp_bson ("{'ping': 1}"));
+   ASSERT_CMPINT ((int) request->opcode, ==, (int) MONGOC_OPCODE_MSG);
+   mock_server_replies_ok_and_destroys (request);
+   ASSERT_OR_PRINT (future_get_bool (future), error);
+   future_destroy (future);
+
+   /* Check that selecting the server still returns the OP_MSG server */
```
jmikola, line 127: I don't follow this comment. If you're asserting below that the maxWireVersion is 6, isn't the OP_QUERY server being returned here?

Otherwise, I'm a bit confused as to why selecting a server with `client_opquery` would return a server pertaining to `client_opmsg`.

src/libmongoc/tests/test-mongoc-server-stream.c
```diff
@@ -0,0 +1,227 @@
...
+   /* Check that the mock server receives an OP_MSG. */
+   request = mock_server_receives_msg (server, 0, tmp_bson ("{'ping': 1}"));
+   ASSERT_CMPINT ((int) request->opcode, ==, (int) MONGOC_OPCODE_MSG);
+   mock_server_replies_ok_and_destroys (request);
+   ASSERT_OR_PRINT (future_get_bool (future), error);
+   future_destroy (future);
+
+   /* Muck with the topology description. */
+   /* Pass in an empty error. */
+   memset (&error, 0, sizeof (bson_error_t));
```
jmikola, line 181: Is there any reason `error` wouldn't still be empty? Is this just being defensive in case something _did_ set it earlier and you happened not to fail via an assertion? Just to confirm, this is the _only_ function in this test where `error` is an _input_ parameter, correct?

Alternatively, I suppose you could declare `bson_error_t empty_error = {0}` for this function to only be used for this purpose.

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -571,10 +571,22 @@ test_mongoc_client_speculative_auth_failure (bool pooled)
...
       r = mongoc_cursor_error (cursor, &error);
       BSON_ASSERT (r);
-      ASSERT_ERROR_CONTAINS (error,
-                             MONGOC_ERROR_CLIENT,
-                             MONGOC_ERROR_CLIENT_AUTHENTICATE,
-                             "Failed to send \"saslContinue\" command");
+      /* A client pool on servers supporting speculative auth (4.4+) will get an
+       * error on saslContinue and subsequently attempt to start auth from the
+       * beginning. The failpoint closes the stream causing an error from
+       * saslStart. */
```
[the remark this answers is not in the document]
kevinAlbs: The pooled connection speculative auth logic will re-use the closed connection to attempt to restart authentication. Single-threaded connections will bail out earlier due to the changes of CDRIVER-3934. 

I filed CDRIVER-4090 to investigate this on pooled connections, since that seems like it could be a bug.

Note, this was not surfacing before because the network error would cause the server to be marked Unknown in the topology description. The attempt to restart auth would error earlier when attempting to assemble a command on an Unknown server [here](https://github.com/kevinAlbs/mongo-c-driver/blob/bcf7da6489bdaf7c0adf295a350b5e0887c67fa1/src/libmongoc/src/mongoc/mongoc-cmd.c#L835:L839). That appended the previous error message containing `Failed to send \"saslContinue\"`.

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -817,19 +829,25 @@ test_mongoc_client_authenticate_timeout (void *context)
...
+/* Update: this test was changed after CDRIVER-3653 was fixed.
+ * Originally, the test used cursor operations, assuming changes in the
+ * min/maxWireVersion from the mock server would be re-evaluated on each cursor
+ * operation.
+ * After CDRIVER-3653 was fixed, this is no longer true. The cursor will examine
+ * the server description associated with the connection handshake. If the
+ * connection has not been closed, changes from monitoring will not affect the
+ * connection's server description.
+ * This test now uses mongoc_client_select_server to validate wire version
+ * checks. */
```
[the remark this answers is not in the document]
kevinAlbs: `mongoc_client_select_server` is used to initialize SDAM and to obtain a server description from the shared topology description, which will get updated (whereas the handshake server description associated with the cursor connections would be set and remain fixed).

src/libmongoc/tests/test-mongoc-client.c
kevinAlbs: 🤦 I totally missed that. Updated to use a client pool.

src/libmongoc/tests/test-mongoc-cluster.c
kevinAlbs: I updated the comment. This test was written before streamable hello. Marking the server as unknown manually ends up causing this test to take the full heartbeatFrequencyMS to rediscover a server to send endSessions.

src/libmongoc/tests/test-mongoc-cluster.c
kevinAlbs: Yes, it still invalidates the one stream. Though it is not particularly useful in single-threaded clients since there is only one stream.

src/libmongoc/tests/test-mongoc-cluster.c
kevinAlbs: Yes, the other test has been updated.

src/libmongoc/tests/test-mongoc-server-stream.c
kevinAlbs: That was a typo. I have updated it to say `OP_QUERY`.

> Otherwise, I'm a bit confused as to why selecting a server with client_opquery would return a server pertaining to client_opmsg.

To clarify, it does not matter which client is used. `mongoc_client_select_server` should return the server description from the shared topology description. I changed this to `client_opmsg` to ensure it does not matter which client is used.

src/libmongoc/tests/test-mongoc-server-stream.c
kevinAlbs: > Is this just being defensive in case something did set it earlier and you happened not to fail via an assertion?

No, this is so an uninitialized error is not passed in. I updated the comment to note that the error should be zeroed out.

> Just to confirm, this is the only function in this test where error is an input parameter, correct?

Correct.

src/libmongoc/doc/mongoc_client_get_handshake_description.rst
```diff
@@ -0,0 +1,50 @@
...
+                                            bson_error_t *error);
+
+Returns a description constructed from the initial handshake response to a server.
+
+Description
+-----------
+
+:symbol:`mongoc_client_get_handshake_description` is distinct from :symbol:`mongoc_client_get_server_description`. :symbol:`mongoc_client_get_server_description` returns a server description constructed from monitoring, which may differ from the server description constructed from the connection handshake.
+
+:symbol:`mongoc_client_get_handshake_description` does not attempt to establish a connection to the server if a connection was not already established. It will complete authentication on a single-threaded client if the connection has not yet been used by an operation.
```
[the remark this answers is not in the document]
jmikola: Ah, does this pertain to monitoring not using auth at all? But on single-threaded, since it's also the application socket auth ends up being completed?

I'm unsure if this current behavior would be problematic, but I think anything we can do to make it consistent between modes of operation would be welcome. It simplifies the documentation and can avoid a subtle impact (which we might not realize now) down the line.

src/libmongoc/doc/mongoc_client_get_handshake_description.rst
```diff
@@ -0,0 +1,50 @@
...
+
+To establish a connection on a single-threaded client, ensure the server is selectable (the monitoring connection is the only connection made to the server). This can be done with a function like :symbol:`mongoc_client_select_server`, or any function which performs server selection.
```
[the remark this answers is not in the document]
jmikola: > If this is going to be difficult for wrapping drivers to utilize. It may be simpler for this to always attempt to establish a connection if one does not exist. That way, callers do not need to do any special checking to ensure a connection is established.

I do think having `mongoc_client_get_handshake_description` establish a connection would be desirable. I can't think of a reason why we'd actually want an empty server description returned (unless the connection could not be established in which case it'd likely be an error).

src/libmongoc/doc/mongoc_client_get_handshake_description.rst
```diff
@@ -0,0 +1,50 @@
...
+
+To establish a connection on a pooled client, send a command (e.g. ping) to the server.
```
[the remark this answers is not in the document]
jmikola: Just to confirm, `mongoc_client_get_handshake_description` is all about returning handshake data for an _application_ stream, correct? In pooled mode, this has nothing to do with monitoring streams because those don't have server IDs and are entirely internal.

src/libmongoc/doc/mongoc_client_get_handshake_description.rst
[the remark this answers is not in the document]
kevinAlbs: That sounds good to me. I will proceed with that change. My original thinking was to have this function do no more than necessary. But ensuring a connection is established seems more awkward.

src/libmongoc/doc/mongoc_client_get_handshake_description.rst
```diff
@@ -21,13 +21,21 @@ Description
...
+To ensure a connection has been established ensure that the server is selectable (e.g. by calling :symbol:`mongoc_client_select_server`) or that an operation has successfully run on the server (e.g. by sending "ping").
 
-Use this function only for building a language driver that wraps the C Driver. When writing applications in C, higher-level functions automatically select a suitable server.
+If the established connection has not completed authentication, calling :symbol:`mongoc_client_get_handshake_description` will complete authentication on the connection.
+
+Single-threaded clients only have one active connection to each server. The one connection is used for both monitoring and application operations. However, the server description returned by :symbol:`mongoc_client_get_handshake_description` may still differ from :symbol:`mongoc_client_get_server_description`. Notably, if connected to a load balanced cluster, the :symbol:`mongoc_client_get_server_description` will describe the load balancer server (:symbol:`mongoc_server_description_type` will return "LoadBalancer"). And :symbol:`mongoc_client_get_handshake_description` will describe the backing server.
```
jmikola, line 34:
```
the :symbol:`mongoc_client_get_server_description` will...
```

Should this say:

```
the server description returned by :symbol:`mongoc_client_get_server_description` will...
```

It seems odd to refer to the function with "the" in the original text. Alternatively, maybe just remove "the" and call it a day.

jmikola (commented):
Left one comment about docs. I still agree with changing `mongoc_client_get_handshake_description` to always initialize the connection if possible, as I think that'd be easier to work with (and consistency across modes of operation is always a plus from my end).

Also, note the outstanding typo fix in https://github.com/mongodb/mongo-c-driver/pull/822#discussion_r670933137

src/libmongoc/doc/mongoc_client_get_handshake_description.rst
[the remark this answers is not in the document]
kevinAlbs: > Ah, does this pertain to monitoring not using auth at all?

Yes. Monitoring, even for single-threaded, will not authenticate connections. In single-threaded, a connection completes authentication the first time it is used for an operation when `mongoc_cluster_fetch_stream_single` is called.

> I'm unsure if this current behavior would be problematic, but I think anything we can do to make it consistent between modes of operation would be welcome. It simplifies the documentation and can avoid a subtle impact (which we might not realize now) down the line.

I agree it seems simpler to reason about `mongoc_client_get_handshake_description` if it also attempts to establish a connection.

src/libmongoc/doc/mongoc_client_get_handshake_description.rst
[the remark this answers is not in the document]
kevinAlbs: Correct, in pooled mode `mongoc_client_get_handshake_description` will not touch monitoring streams, or use a server description obtained by those monitoring streams.

In pooled mode, internally, monitoring streams they still map to server IDs. They are managed in `mongoc_server_monitor_t`, and are never used for application operations.

src/libmongoc/doc/mongoc_client_get_handshake_description.rst
kevinAlbs: Ah yes, done. I took the more verbose suggestion.

kevinAlbs (commented):
Updated `mongoc_client_get_handshake_description` to establish a connection and added tests to verify.

jmikola (approved):
A question about `mongoc_server_description_type` but LGTM.

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -4080,8 +4080,8 @@ test_mongoc_client_get_handshake_hello_response_pooled (void)
    handshake_sd = mongoc_client_get_handshake_description (
       client, monitor_sd->id, NULL /* opts */, &error);
    ASSERT_OR_PRINT (handshake_sd, error);
-   BSON_ASSERT (MONGOC_SERVER_UNKNOWN !=
-                mongoc_server_description_type (handshake_sd));
```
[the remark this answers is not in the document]
kevinAlbs: > mongoc_server_description_type always returned a string, so was this original code a bug since MONGOC_SERVER_UNKNOWN is an enum value?

Yes, this was a bug since it was comparing the string to the enum value.

> Separate question, is there any reason libmongoc doesn't define constants for these strings? I know the enum type is private. Is this related to just not wanting to declare possible server types in any sort of public ABI?

I am not aware of any reason since the API predates me. But that seems plausible. SDAM server types are entrenched and long lived enough to where I would not be opposed to defining string constants if that would be useful for wrapping drivers. Feel free to file a ticket or PR to make that change.

PR #824, opened 2021-07-13
continue snapshot reads implementation

benjirewis:
CDRIVER-4042 Add snapshot to all commands when enabled
CDRIVER-4043 Raise error for snapshot sessions on < 5.0

Syncs spec tests for snapshot sessions. Removes command name check in `_mongoc_client_session_append_read_concern` for appending "snapshot" level read concern. We'll now rely on the server to error if a user runs an unsupported command with "snapshot" level read concern. Raises a new error before potentially appending "snapshot" level read concern if wire version is less than 13 (5.0). Adds the missing `listIndexes` operation to the unified test runner.


src/libmongoc/src/mongoc/mongoc-cmd.c
```diff
@@ -981,12 +981,24 @@ mongoc_cmd_parts_assemble (mongoc_cmd_parts_t *parts,
...
          if (cs) {
+            /* Snapshot Sessions Spec: "Snapshot reads require MongoDB 5.0+."
+             * Throw an error if snapshot is enabled and wire version is less
+             * than 13 before potentially appending "snapshot" read concern. */
+            if (mongoc_session_opts_get_snapshot (&cs->opts) &&
+                server_stream->sd->max_wire_version <
+                   WIRE_VERSION_SNAPSHOT_READS) {
+               bson_set_error (error,
+                               MONGOC_ERROR_TRANSACTION,
+                               MONGOC_ERROR_TRANSACTION_INVALID_STATE,
```
benjirewis, line 992: Is there a better domain and/or code for this error?

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -1569,7 +1569,6 @@ void
 _mongoc_client_session_append_read_concern (const mongoc_client_session_t *cs,
                                             const bson_t *rc,
                                             bool is_read_command,
-                                            const char *cmd_name,
```
kevinAlbs, line 1572: Yay for simplifying :)

src/libmongoc/src/mongoc/mongoc-cmd.c
kevinAlbs: I think the `MONGOC_ERROR_TRANSACTION` domain should be avoided. This error scenario does not necessarily mean a transaction was started on this session.

The [definition](http://mongoc.org/libmongoc/current/errors.html) of the error code `MONGOC_ERROR_CLIENT_SESSION_FAILURE` seems fitting:
> Failure related to creating or using a logical session.

It is currently used in client side checks related to sessions:
- validating causal consistency is not enabled with snapshot
- checking the server indicates session support
- failure to generate a UUID locally

I suggest using `MONGOC_ERROR_CLIENT_SESSION_FAILURE` with the domain of `MONGOC_ERROR_CLIENT`.

kevinAlbs (approved):
LGTM with an update to the error code/domain for the wire version check!

src/libmongoc/tests/json/sessions/legacy/dirty-session-errors.json
```diff
@@ -21,171 +21,6 @@
     }
   ],
   "tests": [
-    {
-      "description": "Clean explicit session is not discarded",
```
benjirewis, line 25: I don't know when exactly these two tests were removed from the spec, but might as well update.

src/libmongoc/src/mongoc/mongoc-cmd.c
benjirewis: That sounds perfect; done.

PR #829, opened 2021-07-16
CDRIVER-4091 - Truncation in strncpy() causes build error

chardan:
https://jira.mongodb.org/browse/CDRIVER-4091

gcc 9.3.0 complains that because the strncpy() call uses the source's size as a bound,
the trailing NULL will not be copied.

Replaced with:
updated allocation size of destination to include room for the NULL character
eliminated redundant srlen()
changed strncpy() call to strcpy(), which includes the trailing NULL

### DIFF
--- a/src/libmongoc/tests/test-mongoc-mongohouse.c
+++ b/src/libmongoc/tests/test-mongoc-mongohouse.c
@@ -151,7 +151,6 @@ cmd_succeeded_cb (const mongoc_apm_command_succeeded_t *event)
 
    /* Store cursor information from our initial find. */
    if (strcmp (cmd, "find") == 0) {
-      const char *temp;
 
       BSON_ASSERT (!test->parsed_cursor);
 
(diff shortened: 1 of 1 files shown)


src/libmongoc/tests/test-mongoc-mongohouse.c
```diff
@@ -165,8 +166,9 @@ cmd_succeeded_cb (const mongoc_apm_command_succeeded_t *event)
       BSON_ASSERT (BSON_ITER_HOLDS_UTF8 (&child_iter));
 
       temp = bson_iter_utf8 (&child_iter, NULL);
-      test->cursor_ns = (char *) bson_malloc0 (strlen (temp));
-      strncpy (test->cursor_ns, temp, strlen (temp));
+      temp_size = 1 + strlen(temp);
+      test->cursor_ns = (char *) bson_malloc0 (temp_size);
```
kevinAlbs, line 170: This can be simplified with [bson_strdup](http://mongoc.org/libbson/current/bson_strdup.html):

```
test->cursor_ns = bson_strdup (temp);
```

kevinAlbs (commented):
Nice catch!

If it isn't too much trouble, can you include the version of gcc the build error message in the description of CDRIVER-4091?

Evergreen tests several gcc versions. Based on the variant names in [variants.py](https://github.com/mongodb/mongo-c-driver/blob/master/build/evergreen_config_lib/variants.py#L205:L205): 4.4, 4.8, 4.9, 5.4, 6.3, 8.2, 8.3.

If this occurs on a version of gcc we are not testing, it may be helpful to reference the exact error and gcc version in case other users run into this.

src/libmongoc/tests/test-mongoc-mongohouse.c
chardan: +1 bson_strdup() makes this much prettier! I'll update the commit with the version number (this is indeed a newer compiler, 9.3).

chardan:
@kevinAlbs It looks to me that the failing tests are the ones that are presently expected to fail. But as I'm quite green with Evergreen myself, if you could please sanity check that before I merge it, I'd appreciate it!

kevinAlbs (approved):
I took a look, and those all look expected. LGTM!

PR #816, opened 2021-07-03
Load balancer: SDAM and Server Selection

kevinAlbs:
- CDRIVER-4053 Load balancer: SDAM changes
  - Add a LoadBalanced topology type and LoadBalancer server type.
  - Manually set the topology description to contain one server of type LoadBalancer when opening.
  - Prevent updates to the topology description after opening.
  - Disable SDAM monitoring for a LoadBalanced topology.
  - Emit correct SDAM monitoring events for a LoadBalanced topology.
  - Bypass the [cooldown period](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#cooldownms) for the single-threaded topology scanner.
- CDRIVER-4055 Load balancer: Server Selection
  - Always return the one load balancer in server selection.

PR #831, opened 2021-07-22
CDRIVER-4058 Load balancer: Error handling

kevinAlbs:
Implements the change specified in the [Error Handling](https://github.com/mongodb/specifications/blob/master/source/load-balancers/load-balancers.rst#error-handling) section of the Load Balancer specification.

- Ignore SDAM error handling on errors before the handshake.
    - This was done in the SDAM error handling function `_mongoc_topology_handle_app_error` and updates to the function for obtaining an application-use connection: `_mongoc_cluster_stream_for_server`.
- Determine the connection generation based on the serviceId field in the handshake response.
    - This is done with a new data structure called a generation map. It maps a serviceID to a connection pool generation. In non-LB topologies, the serviceID is always the zero'ed ObjectID.

The remaining changes specified in the [Error Handling](https://github.com/mongodb/specifications/blob/master/source/load-balancers/load-balancers.rst#error-handling) section of the Load Balancer specification were already completed as part of #816:
- Process SDAM error handling for errors after the initial handshake, but do not mark the server Unknown.
- Do not make any changes to the TopologyDescription or ServerDescription on a state change error.

I added a refactoring change that was not in scope, but seemed small and helpful for maintainability:
- Remove redundant `generation` field from `mongoc_cluster_node_t`

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-cluster-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-cluster-private.h
@@ -44,7 +44,7 @@ BSON_BEGIN_DECLS
 typedef struct _mongoc_cluster_node_t {
    mongoc_stream_t *stream;
    char *connection_address;
-   uint32_t generation;
+   uint32_t generation; // LBTODO: should we remove this?
    /* handshake_sd is a server description created from the handshake on the stream. */
    mongoc_server_description_t *handshake_sd;
 } mongoc_cluster_node_t;
--- a/src/libmongoc/src/mongoc/mongoc-cluster-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-cluster-private.h
@@ -44,7 +44,6 @@ BSON_BEGIN_DECLS
 typedef struct _mongoc_cluster_node_t {
    mongoc_stream_t *stream;
    char *connection_address;
-   uint32_t generation; // LBTODO: should we remove this?
    /* handshake_sd is a server description created from the handshake on the stream. */
    mongoc_server_description_t *handshake_sd;
 } mongoc_cluster_node_t;
--- a/src/libmongoc/src/mongoc/mongoc-cluster.c
+++ b/src/libmongoc/src/mongoc/mongoc-cluster.c
@@ -2105,7 +2105,6 @@ _mongoc_cluster_add_node (mongoc_cluster_t *cluster,
    cluster_node =
       _mongoc_cluster_node_new (stream, host->host_and_port);
 
-   // LBTODO: note, this stream will have a 0 generation. That should be OK.
    handshake_sd = _mongoc_cluster_run_hello (cluster,
                                              cluster_node,
                                              server_id,
@@ -2116,12 +2115,6 @@ _mongoc_cluster_add_node (mongoc_cluster_t *cluster,
    if (!handshake_sd) {
       GOTO (error);
    }
-
-   // LBTODO: the generation cannot be determined until after handshake.
-   bson_mutex_lock (&cluster->client->topology->mutex);
-   cluster_node->generation = _mongoc_topology_get_connection_generation (cluster->client->topology, server_id, &handshake_sd->service_id);
-   MONGOC_DEBUG ("setting stream generation to: %d", (int) (cluster_node->generation));
-   bson_mutex_unlock (&cluster->client->topology->mutex);
 
    _mongoc_handshake_parse_sasl_supported_mechs (
       &handshake_sd->last_hello_response, &sasl_supported_mechs);
@@ -2146,10 +2139,12 @@ _mongoc_cluster_add_node (mongoc_cluster_t *cluster,
 
    /* Transfer ownership of the server description into the cluster node. */
    cluster_node->handshake_sd = handshake_sd;
-   /* Copy the generation from the cluster node.
+   /* Copy the latest connection pool generation.
     * TODO (CDRIVER-4078) do not store the generation counter on the server
     * description */
-   cluster_node->handshake_sd->generation = cluster_node->generation;
+   bson_mutex_lock (&cluster->client->topology->mutex);
+   cluster_node->handshake_sd->generation = _mongoc_topology_get_connection_pool_generation (cluster->client->topology, server_id, &handshake_sd->service_id);
+   bson_mutex_unlock (&cluster->client->topology->mutex);
 
    bson_destroy (&speculative_auth_response);
    mongoc_set_add (cluster->nodes, server_id, cluster_node);
@@ -2297,7 +2292,6 @@ _mongoc_cluster_stream_for_server (mongoc_cluster_t *cluster,
                                                               server_id);
       }
       bson_mutex_unlock (&topology->mutex);
-
       return NULL;
    }
 
@@ -2305,7 +2299,7 @@ _mongoc_cluster_stream_for_server (mongoc_cluster_t *cluster,
     * service id, disconnect and return an error. */
    bson_mutex_lock (&topology->mutex);
    if (topology->description.type == MONGOC_TOPOLOGY_LOAD_BALANCED) {
-      if (!mongoc_server_description_has_service_id(server_stream->sd)) {
+      if (!mongoc_server_description_has_service_id (server_stream->sd)) {
          bson_set_error (error,
                          MONGOC_ERROR_CLIENT,
                          MONGOC_ERROR_CLIENT_INVALID_LOAD_BALANCER,
@@ -2397,7 +2391,6 @@ mongoc_cluster_fetch_stream_single (mongoc_cluster_t *cluster,
                                     bson_error_t *error /* OUT */)
 {
    mongoc_topology_t *topology;
-   mongoc_server_description_t *monitor_sd;
    mongoc_server_description_t *handshake_sd;
    mongoc_topology_scanner_node_t *scanner_node;
    char *address;
@@ -2499,15 +2492,10 @@ mongoc_cluster_fetch_stream_single (mongoc_cluster_t *cluster,
       scanner_node->has_auth = true;
    }
 
-   /* Always copy the latest generation from the shared server description. */
-   monitor_sd = mongoc_topology_server_by_id (topology, server_id, error);
-   if (!monitor_sd) {
-      mongoc_server_description_destroy (handshake_sd);
-      return NULL;
-   }
-   /* TODO: (CDRIVER-4078) do not store the generation or generation map as part of the server description. */
-   handshake_sd->generation = mongoc_generation_map_get (monitor_sd->generation_map, &handshake_sd->service_id);
-   mongoc_server_description_destroy (monitor_sd);
+   /* Copy the latest connection pool generation.
+    * TODO (CDRIVER-4078) do not store the generation counter on the server
+    * description */
+   handshake_sd->generation = _mongoc_topology_get_connection_pool_generation (cluster->client->topology, server_id, &handshake_sd->service_id);
 
    return mongoc_server_stream_new (
       &topology->description, handshake_sd, scanner_node->stream);
@@ -2556,7 +2544,7 @@ mongoc_cluster_stream_valid (mongoc_cluster_t *cluster,
    bson_mutex_lock (&topology->mutex);
    sd = mongoc_topology_description_server_by_id (
       &topology->description, server_stream->sd->id, &error);
-   if (!sd || server_stream->sd->generation < _mongoc_topology_get_connection_generation (topology, server_stream->sd->id, &server_stream->sd->service_id)) {
+   if (!sd || server_stream->sd->generation < _mongoc_topology_get_connection_pool_generation (topology, server_stream->sd->id, &server_stream->sd->service_id)) {
       /* No server description, or the pool has been cleared. */
       bson_mutex_unlock (&topology->mutex);
       MONGOC_DEBUG ("stream invalid");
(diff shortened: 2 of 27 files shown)


src/libmongoc/src/mongoc/mongoc-cluster-private.h
```diff
@@ -44,7 +44,6 @@ BSON_BEGIN_DECLS
 typedef struct _mongoc_cluster_node_t {
    mongoc_stream_t *stream;
    char *connection_address;
-   uint32_t generation;
```
kevinAlbs, line 47: A `mongoc_cluster_node_t` stores an application connection on a pooled client. The `handshake_sd` already contained the `generation` associated with the connection. This was a redundant copy of that value.

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -102,7 +102,8 @@ _handle_not_primary_error (mongoc_cluster_t *cluster,
                                           reply,
                                           NULL,
                                           server_stream->sd->max_wire_version,
-                                          server_stream->sd->generation)) {
+                                          server_stream->sd->generation,
+                                          &server_stream->sd->service_id)) {
```
kevinAlbs, line 106: `_mongoc_topology_handle_app_error` processes SDAM errors as described in the SDAM specification's [Error Handling](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#error-handling). This now accepts the connection's `service_id` since connection pool clearing is now parameterized by serviceID.

src/libmongoc/src/mongoc/mongoc-generation-map.c
```diff
@@ -0,0 +1,151 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "mongoc-generation-map-private.h"
+
+#include "utlist.h"
+
+typedef struct _gm_node_t {
```
kevinAlbs, line 21: This is a simple map implemented as a singly-linked list. I err'ed on simplicity instead of performance for the generation map. This map will have at most one entry per unique backing server behind a load balancer.

src/libmongoc/src/mongoc/mongoc-server-description-private.h
```diff
@@ -184,11 +194,10 @@ void
...
 
-/* If a service_id is set on the topology description, copies it to @oid and
- * returns true. Otherwise returns false and zeros out oid.
- */
+extern const bson_oid_t kZeroServiceId;
+
 bool
-mongoc_server_description_service_id (
-   const mongoc_server_description_t *description, bson_oid_t *oid);
+mongoc_server_description_has_service_id (
```
kevinAlbs, line 200: This is now more consistent with the existing `mongoc_server_description_has_election_id`, which is another ObjectID field on the server description.

src/libmongoc/tests/test-mongoc-loadbalanced.c
```diff
@@ -549,6 +550,251 @@ test_loadbalanced_handshake_rejects_non_loadbalanced (void)
...
+                                          "admin",
+                                          tmp_bson ("{'ping': 1}"),
+                                          NULL /* read prefs */,
+                                          NULL /* reply */,
+                                          &error);
+   /* A new connection is opened. */
+   request =
+      mock_server_receives_legacy_hello (server, "{'loadBalanced': true}");
+   BSON_ASSERT (request);
+   capture_logs (true); /* hide Failed to buffer logs. */
```
kevinAlbs, line 608: Note, https://github.com/mongodb/mongo-c-driver/blob/4919d4f0e5ff7f4f7a18d433de88e5c16e1e3290/.evergreen/compile-unix.sh#L293:L307 checks the tests that are run alongside compilation (mock server and unit tests) to assert that nothing is logged to stderr. When `mock_server_hangs_up` closes the connection, libmongoc logs the failed socket write. This captures and ignores those logs.

src/libmongoc/src/mongoc/mongoc-generation-map.c
benjirewis: I've never seen a C map implemented as a singly-linked list. I do wonder about performance (particularly in `mongoc_generation_map_get/increment`). Is there precedence for map implementations in the C driver? Any thoughts @chardan?

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -1783,26 +1783,32 @@ _find_topology_version (const bson_t *reply, bson_t *topology_version)
...
 
+   BSON_ASSERT (service_id);
+
    sd = mongoc_topology_description_server_by_id (
       &topology->description, server_id, &error);
    if (!sd) {
       /* Server removed, ignore and ignore error. */
       return;
    }
-   TRACE ("clearing pool for server: %s", sd->host.host_and_port);
```
benjirewis, line 1797: Why remove the `TRACE`? I see this was moved to `_mongoc_topology_get_connection_pool_generation` but I'm not sure I understand why.

src/libmongoc/tests/test-mongoc-generation-map.c
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "mongoc-generation-map-private.h"
+
+#include "TestSuite.h"
+
+static void
+test_generation_map_basic (void) {
```
benjirewis, line 22: Thanks for adding associated tests for new type.

src/libmongoc/tests/test-mongoc-loadbalanced.c
```diff
@@ -549,6 +550,251 @@ test_loadbalanced_handshake_rejects_non_loadbalanced (void)
    mock_server_destroy (server);
 }
 
+/* Test that an error before the MongoDB handshake completes does NOT go through
```
benjirewis, line 553: Great tests here.

src/libmongoc/src/mongoc/mongoc-server-description-private.h
```diff
@@ -184,11 +194,10 @@ void
 mongoc_server_description_set_topology_version (mongoc_server_description_t *sd,
                                                 const bson_t *tv);
 
-/* If a service_id is set on the topology description, copies it to @oid and
- * returns true. Otherwise returns false and zeros out oid.
- */
+extern const bson_oid_t kZeroServiceId;
```
benjirewis, line 197: Why is this constant in lower camel case? And, what does `k` mean?

src/libmongoc/src/mongoc/mongoc-generation-map.c
```diff
@@ -0,0 +1,151 @@
...
+   if (!node) {
+      return;
+   }
+   bson_free (node);
+}
+
+static gm_node_t *
+gm_node_copy (const gm_node_t *node)
+{
+   gm_node_t *node_copy = gm_node_new ();
```
chardan, line 45: Is it ok that the result of this is not checked?
(Peeking at bson_oid_copy() below, it uses BSON_ASSERT() on both of its parameters; however, since the "key" member is being dereferenced, if that happens on a null pointer, I expect it will explode.)

src/libmongoc/src/mongoc/mongoc-generation-map.c
```diff
@@ -0,0 +1,151 @@
...
+   return gm;
+}
+
+mongoc_generation_map_t *
+mongoc_generation_map_copy (const mongoc_generation_map_t *gm)
+{
+   mongoc_generation_map_t *gm_copy;
+   gm_node_t *iter;
+
+   gm_copy = mongoc_generation_map_new ();
```
chardan, line 71: Ditto here, I suspect this could segfault below when one of its members is dereferenced.

src/libmongoc/src/mongoc/mongoc-generation-map.c
```diff
@@ -0,0 +1,151 @@
...
+   return gm_copy;
+}
+
+uint32_t
+mongoc_generation_map_get (const mongoc_generation_map_t *gm,
+                           const bson_oid_t *key)
+{
+   gm_node_t *iter = NULL;
+
+   BSON_ASSERT (key);
```
chardan, line 90: Do we also want to BSON_ASSERT(gm)?

src/libmongoc/src/mongoc/mongoc-generation-map.c
```diff
@@ -0,0 +1,151 @@
...
+}
+
+void
+mongoc_generation_map_increment (mongoc_generation_map_t *gm,
+                                 const bson_oid_t *key)
+{
+   gm_node_t *match;
+   gm_node_t *iter = NULL;
+
+   BSON_ASSERT (key);
```
chardan, line 113: Since we don't appear to BSON_ASSERT(gm), I'll assume there's a reason we don't worry about it, and won't mention it further.

src/libmongoc/src/mongoc/mongoc-generation-map.c
```diff
@@ -0,0 +1,151 @@
...
+   {
+      if (bson_oid_equal (key, &iter->key)) {
+         break;
+      }
+   }
+
+   if (iter) {
+      match = iter;
+   } else {
+      gm_node_t *new_node = gm_node_new ();
```
chardan, line 125: See previous note about dereferencing. (It could be that this library is ok with faulting, in which case please ignore my comments and LMK. Thanks!)

src/libmongoc/src/mongoc/mongoc-server-description-private.h
```diff
@@ -111,7 +112,16 @@ struct _mongoc_server_description_t {
    4. an app thread receives a "not primary" or "node is recovering" error from a
    pre-4.2 server.
    */
+
+   /* generation only applies to a server description tied to a connection.
+    * It represents the generation number for this connection. */
    uint32_t generation;
+
+   /* generation_map stores all generations for all service IDs associated with
```
chardan, line 120: +1, useful comments!

src/libmongoc/src/mongoc/mongoc-server-description.c
```diff
@@ -822,6 +826,8 @@ mongoc_server_description_new_copy (
    memcpy (&copy->error, &description->error, sizeof copy->error);
 
    copy->generation = description->generation;
+   copy->generation_map =
+      mongoc_generation_map_copy (description->generation_map);
    return copy;
```
chardan, line 831: If mongoc_generation_map_copy() is updated to return NULL on failure, remember to consider whatever needs to happen to invalidate "copy" or notify of error.

src/libmongoc/tests/test-mongoc-generation-map.c
```diff
@@ -0,0 +1,54 @@
...
+   /* The generation map increments to 1 for a key not found. */
+   mongoc_generation_map_increment (gm, &oid_b);
+   ASSERT_CMPUINT32 (1, ==, mongoc_generation_map_get (gm, &oid_b));
+
+   /* Test incrementing again. */
+   mongoc_generation_map_increment (gm, &oid_b);
+   ASSERT_CMPUINT32 (2, ==, mongoc_generation_map_get (gm, &oid_b));
+
+   /* Copying a generation map retains values. */
+   gm_copy = mongoc_generation_map_copy (gm);
```
chardan, line 44: Remember to update if mongoc_generation_map_copy() is changed to report errors.

src/libmongoc/src/mongoc/mongoc-generation-map.c
chardan: @benjirewis So, on one hand an associative array can be implemented any number of ways, which are all going to have different performance characteristics. @kevinAlbs has noted that this isn't intended for performance, with the key being that it will "have at most one entry per unique backing server". If this number is nearly always small, I see no real problem with using a linked list. 

I did have another couple questions, though, which is that it seems odd we'd have to do this afresh (although this is C land, not C++ land); does the library we're using (is it libcoap..?) for the LL_* macros maybe has something for associative arrays already? My other question appears on line 52, which is that it seems unusual to wrap the data and the link together inside another type, but it could be that @kevinAlbs is intending to make that whole implementation abstract? If so, it almost feels like it should be its own mini-library?

src/libmongoc/src/mongoc/mongoc-generation-map.c
```diff
@@ -0,0 +1,151 @@
...
+gm_node_copy (const gm_node_t *node)
+{
+   gm_node_t *node_copy = gm_node_new ();
+
+   bson_oid_copy (&node->key, &node_copy->key);
+   node_copy->val = node->val;
+   return node_copy;
+}
+
+struct _mongoc_generation_map {
```
chardan, line 52: Is the gm_node_t type being wrapped in here in order to allow us to replace the implementation of the map?

chardan (commented):
Looking good! There's a lot going on here, but generally it seems well organized and I think it's getting close-- I had just a few comments for your consideration.

Caveat: the tests appear to be thorough and good to me, but even though I read the error handling design it's a bit impenetrable and there's a lot that's new to me here, so I hope that other reviewers also are able to comment.

src/libmongoc/src/mongoc/mongoc-generation-map.c
chardan: +1 as per this and our offline discussioln, sounds entirely fine to me. Not only is N small enough, we have a centralized and internal way to change the implementation! :-)

chardan:
+1 LGTM-- thanks for the discussions!

src/libmongoc/src/mongoc/mongoc-generation-map.c
benjirewis: Thanks for the explanations of performance and implementation here. All sounds good to me.

benjirewis (approved):
LGTM! Great work.

PR #832, opened 2021-07-24
CDRIVER-3923 drop deprecated ppc64le build platform

rcsanchez97:
Patch build: https://spruce.mongodb.com/version/60fb72f7d6d80a4987281ff4/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


.evergreen/config.yml
```diff
@@ -27075,26 +27075,6 @@ buildvariants:
...
-  - debug-compile-no-align
-  - .debug-compile !.sspi .openssl
-  - .debug-compile !.sspi .nossl
-  - .authentication-tests .openssl !.sasl
-  - .latest .openssl !.nosasl .server
-  - .latest .nossl
-  - .5.0 .openssl !.nosasl .server
-  - .4.4 .openssl !.nosasl .server
-  - .4.2 .openssl !.nosasl .server
-  - .4.0 .openssl !.nosasl .server
```
kevinAlbs, line 27096: I think this is safe to drop. We do lose tests on PPC against 4.0 since the `Power8 (ppc64le) (RHEL 8.1)` variant only tests back to 4.2. That being said, that seems like a negligible loss of test coverage.

kevinAlbs (approved):
LGTM!

PR #834, opened 2021-07-27
CDRIVER-4059 service_id on command monitoring events

jmikola:
https://jira.mongodb.org/browse/CDRIVER-4059

It's not often that I add new API methods, so please let me know if I missed anything. AFAICT, the docs don't report the version number when a particular method was introduced.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-server-description-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-server-description-private.h
@@ -109,8 +109,8 @@ struct _mongoc_server_description_t {
    1. a monitor receives a network error
    2. an app thread receives any network error before completing a handshake
    3. an app thread receives a non-timeout network error after the handshake
-   4. an app thread receives a "not primary" or "node is recovering" error from a
-   pre-4.2 server.
+   4. an app thread receives a "not primary" or "node is recovering" error from
+   a pre-4.2 server.
    */
 
    /* generation only applies to a server description tied to a connection.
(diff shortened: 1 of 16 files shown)


src/libmongoc/src/mongoc/mongoc-server-description-private.h
```diff
@@ -184,7 +184,7 @@ void
 mongoc_server_description_set_topology_version (mongoc_server_description_t *sd,
                                                 const bson_t *tv);
 
-/* If a service_id is set on the topology description, copies it to @oid and
+/* If a service_id is set on the server description, copies it to @oid and
```
kevinAlbs, line 187: Good eye!

kevinAlbs (approved):
LGTM

> AFAICT, the docs don't report the version number when a particular method was introduced.

That is correct unfortunately. I dream of auto generating "Added in x.y.z" notes in the docs from ABI metadata or git history.

src/libmongoc/src/mongoc/mongoc-server-description-private.h
jmikola: Skipped this commit while rebasing as I saw you removed the comment in https://github.com/mongodb/mongo-c-driver/commit/635be2f7b105cdd62ae95c045b436f8837f2be83#diff-de5aea95eea3d2b786d33b8b694d349c445847e9cc506502c6824e05f177cc0cL187

PR #837, opened 2021-07-30
CDRIVER-4062 test load balancer on evg

kevinAlbs:
__Scope__
- Add tasks to the Evergreen configuration based to test against a loadbalanced cluster.
	- Tests with auth+SSL and with no auth+SSL on both server version 5.0 and latest.
	- Runs tests against: new load balanced tests, all unified tests, retryable reads, retryable writes, change streams, and the C driver specific `/loadbalanced/*` tests.
- Update test-libmongoc to support testing against a load balancer.
	- Support the environment variable `MONGOC_TEST_LOADBALANCED` to enable testing against load balancers in test-libmongoc.
	- Support the environment variables `SINGLE_MONGOS_LB_URI`, `MULTI_MONGOS_LB_URI` to set URIs for test clients.
	- Mock the `serviceId` when `MONGOC_TEST_LOADBALANCED` is true.
	- Support multiple occurrences of the `-l` flag in test-libmongoc and alias the flag as `--match`.
	- Remove unnecessary constraint of retryable writes tests for replica sets.
- CDRIVER-4060 add remaining support to the unified test runner for the new loadbalancer spec tests.
    - Iterate the first result of a `FindCursor` in the `createFindCursor` operation.
    - Support `appname` as a URI parameter.
    - Skip listening or comparing CMAP (Connection Monitoring and Pooling) events, since the C driver does not implement CMAP (per CDRIVER-3525).

__Notes for reviewers__
- 73 of the files changed are from updating the JSON files from the specifications repo. Those were reviewed when they were added to the mongodb/specifications repo, so there is no need to review them. The "File Filter" in the review can collapse all of the JSON files.


src/libmongoc/tests/TestSuite.c
```diff
@@ -137,6 +137,7 @@ TestSuite_Init (TestSuite *suite, const char *name, int argc, char **argv)
    suite->flags = 0;
    suite->prgname = bson_strdup (argv[0]);
    suite->silent = false;
+   _mongoc_array_init (&suite->match_patterns, sizeof (char *));
```
kevinAlbs, line 140: Note, now `test-libmongoc` can be invoked with multiple occurrences of `-l <pattern>`. This was necessary to invoke with the multiple subset of tests we need to run against a load balanced deployment.

I also added an alias of `--match`.

src/libmongoc/tests/test-mongoc-change-stream.c
```diff
@@ -2774,6 +2774,6 @@ test_change_stream_install (TestSuite *suite)
    TestSuite_AddMockServerTest (
       suite, "/change_streams/prose_test_18", prose_test_18);
 
-   test_framework_resolve_path (JSON_DIR "/change_streams", resolved);
+   test_framework_resolve_path (JSON_DIR "/change_streams/legacy", resolved);
```
kevinAlbs, line 2777: In resyncing the change_streams tests, there are now two subdirectories (`unified` and `legacy`). The `unified` tests are installed in the unified test runner: `runner.c`.

src/libmongoc/tests/TestSuite.c
```diff
@@ -857,36 +863,6 @@ TestSuite_PrintJsonFooter (FILE *stream) /* IN */
    fflush (stream);
 }
 
-
-static int
```
kevinAlbs, line 861: A distinct `TestSuite_RunSerial` is no longer needed. If there are no match patterns specified `TestSuite_RunAll` considers anything to match.

src/libmongoc/tests/unified/entity-map.c
```diff
@@ -31,6 +31,11 @@
 #define REDUCED_HEARTBEAT_FREQUENCY_MS 500
 #define REDUCED_MIN_HEARTBEAT_FREQUENCY_MS 50
 
+struct _entity_findcursor_t {
```
kevinAlbs, line 34: Cursors are lazy in libmongoc, so they do not send the initial command (in this case `find`) until the first iteration.

This updates the implementation of the `createFindCursor` test operation to adhere to the specification:

https://github.com/mongodb/specifications/blob/86197086c2290a67500321f008b7a0a0685f0da7/source/unified-test-format/unified-test-format.rst#createfindcursor

> Test runners for drivers that lazily execute the find command on the first iteration of the cursor MUST iterate the resulting cursor once.

src/libmongoc/src/mongoc/mongoc-server-description-private.h
```diff
@@ -200,4 +200,6 @@ bool
 mongoc_server_description_has_service_id (
    const mongoc_server_description_t *description);
 
+extern bool global_mock_service_id;
```
chardan, line 203: nit: This is a good and pretty unique identifier, but /should/ it be "mongoc_" or some such?
question: Any concerns about concurrent writes to this?

src/libmongoc/src/mongoc/mongoc-server-description.c
```diff
@@ -33,6 +33,8 @@ static bson_oid_t kObjectIdZero = {{0}};
 
 const bson_oid_t kZeroServiceId = {{0}};
 
+bool global_mock_service_id = false;
```
chardan, line 36: +1 yay, initialization!

src/libmongoc/tests/TestSuite.c
```diff
@@ -176,11 +177,14 @@ TestSuite_Init (TestSuite *suite, const char *name, int argc, char **argv)
...
-      } else if (0 == strcmp ("-l", argv[i])) {
+      } else if ((0 == strcmp ("-l", argv[i])) ||
+                 (0 == strcmp ("--match", argv[i]))) {
+         char *val;
          if (argc - 1 == i) {
-            test_error ("-l requires an argument.");
+            test_error ("%s requires an argument.", argv[i]);
          }
-         suite->testname = bson_strdup (argv[++i]);
+         val = bson_strdup (argv[++i]);
```
chardan, line 186: Is there a corresponding bson_free()?

note: bson_strdup() can return NULL if passed a NULL parameter, but I don't believe that's possible here.

src/libmongoc/tests/TestSuite.c
```diff
@@ -906,26 +882,48 @@ TestSuite_TestMatchesName (const TestSuite *suite,
...
+bool
+test_matches (TestSuite *suite, Test *test)
+{
+   int i;
+   bool matches;
+
+   /* If no match patterns were provided, then assume all match. */
+   if (suite->match_patterns.len == 0) {
+      matches = true;
+   } else {
```
chardan, line 894: nit: What you have here is fine, but the function is a bit simpler when the positive if-case just returns and this "else" is the rest of the function.

src/libmongoc/tests/TestSuite.c
```diff
@@ -906,26 +882,48 @@ TestSuite_TestMatchesName (const TestSuite *suite,
...
+      matches = false;
+      for (i = 0; i < suite->match_patterns.len; i++) {
+         char *pattern =
+            _mongoc_array_index (&suite->match_patterns, char *, i);
+         if (TestSuite_TestMatchesName (suite, test, pattern)) {
+            matches = true;
+            break;
```
chardan, line 901: nit: Ok, or just return-- variable "matches" might not be needed at all. But it also seems fine as-is.

src/libmongoc/tests/json-test.c
```diff
@@ -1693,6 +1695,23 @@ run_json_general_test (const json_test_config_t *config)
          continue;
       }
 
+      if (config->skips) {
+         test_skip_t *iter;
+         bool should_skip = false;
+
+         for (iter = config->skips; iter->description != NULL; iter++) {
```
chardan, line 1702: Tantalizingly close to a general find() on associative array keys, but here we have the complication of the special field names.

src/libmongoc/tests/json/change_streams/legacy/change-streams-resume-allowlist.json
```diff
@@ -777,7 +778,7 @@
       }
     },
     {
-      "description": "change stream resumes after NotMaster",
+      "description": "change stream resumes after NotWritablePrimary",
```
chardan, line 781: +1, I'm ok with this, although I thought that these kinds of changes were coming as part of a separate ticket. (Fine by me to fix it while we're here!)

src/libmongoc/tests/test-libmongoc.c
```diff
@@ -1395,6 +1403,24 @@ test_framework_get_uri ()
...
+   char *uri_str_no_auth;
+   char *uri_str;
+   mongoc_uri_t *uri;
+
+   uri_str_no_auth = _mongoc_getenv ("MULTI_MONGOS_LB_URI");
+   if (!uri_str_no_auth) {
+      test_error ("expected MULTI_MONGOS_LB_URI to be set");
+   }
+   uri_str = test_framework_add_user_password_from_env (uri_str_no_auth);
+   uri = mongoc_uri_new (uri_str);
```
chardan, line 1417: Note that this can return NULL.-- which may be fine with us, as we wind up just returning it in any case.

src/libmongoc/tests/unified/entity-map.c
```diff
@@ -322,9 +344,14 @@ entity_client_new (entity_map_t *em, bson_t *bson, bson_error_t *error)
     * has one host. If useMultipleMongoses unspecified, make no assertion.
     */
    if (use_multiple_mongoses != NULL) {
-      if (!test_framework_uri_apply_multi_mongos (
-             uri, *use_multiple_mongoses, error)) {
-         goto done;
+      if (*use_multiple_mongoses && test_framework_is_loadbalanced ()) {
```
chardan, line 347: A good time to meditate on "use_multiple_mongoses" and the nature of plurality. ;-)

src/libmongoc/tests/unified/entity-map.c
```diff
@@ -1119,13 +1151,43 @@ entity_map_add_changestream (entity_map_t *em,
...
+entity_map_add_findcursor (entity_map_t *em,
+                           const char *id,
+                           mongoc_cursor_t *cursor,
+                           const bson_t *first_result,
+                           bson_error_t *error)
 {
-   return _entity_map_add (em, id, "cursor", (void *) cursor, error);
+   entity_findcursor_t *findcursor;
+
+   findcursor =
```
chardan, line 1186: Ok, this bason_malloc() abort()s the process on failure.

src/libmongoc/tests/unified/runner.c
```diff
@@ -373,10 +379,13 @@ test_runner_new (void)
    callbacks = mongoc_apm_callbacks_new ();
    mongoc_apm_set_topology_changed_cb (callbacks, on_topology_changed);
    uri = test_framework_get_uri ();
-   /* Always use multiple mongos's if speaking to a mongos.
-    * Some test operations require communicating with all known mongos */
-   if (!test_framework_uri_apply_multi_mongos (uri, true, &error)) {
-      test_error ("error applying multiple mongos: %s", error.message);
+   /* In load balanced mode, the internal client must use the SINGLE_LB_MONGOS_URI. */
+   if (!test_framework_is_loadbalanced ()) {
+      /* Always use multiple mongos's if speaking to a mongos.
```
chardan, line 384: My suspicion is that "mongoses" is probably correct (though odd). There must be a definitive answer! ;-)
https://www.merriam-webster.com/words-at-play/what-happens-to-names-when-we-make-them-plural-or-possessive

src/libmongoc/tests/unified/runner.c
```diff
@@ -1106,14 +1135,30 @@ test_check_expected_events_for_client (test_t *test,
...
       goto done;
    }
 
+   if (event_type) {
+      if (0 == strcmp (event_type, "cmap")) {
+         /* TODO: (CDRIVER-3525) Explicitly ignore cmap events until CMAP is
+          * supported. */
+         ret = true;
+         goto done;
+      } else if (0 == strcmp (event_type, "command")) {
```
chardan, line 1154: Since we've already left via goto, the extra branching isn't needed.

src/libmongoc/tests/unified/util.c
```diff
@@ -163,3 +163,31 @@ test_bson_util_install (TestSuite *suite)
...
+      "connectionReadyEvent",
+      "connectionClosedEvent",
+      "connectionCheckOutStartedEvent",
+      "connectionCheckOutFailedEvent",
+      "connectionCheckedOutEvent",
+      "connectionCheckedInEvent",
+      NULL
+   };
+
+   char ** iter;
```
chardan, line 185: Consider declaring this in the loop, rather than at a distance.

chardan (commented):
This looks pretty good! There are a few things you might want to double-check. Please note that I don't have a great understanding of the test system and it's hard for me to usefully assess what's going on in the long stretches of this, so you may definitely want a second opinion, but they /look/ ok to me. Lots of effort, nice job!

src/libmongoc/tests/unified/runner.c
jmikola: "mongoses" is also used in specs. Consider the `useMultipleMongoses` test syntax.

src/libmongoc/tests/unified/entity-map.c
```diff
@@ -322,9 +344,14 @@ entity_client_new (entity_map_t *em, bson_t *bson, bson_error_t *error)
...
     */
    if (use_multiple_mongoses != NULL) {
-      if (!test_framework_uri_apply_multi_mongos (
-             uri, *use_multiple_mongoses, error)) {
-         goto done;
+      if (*use_multiple_mongoses && test_framework_is_loadbalanced ()) {
+         mongoc_uri_destroy (uri);
+         uri = test_framework_get_uri_multi_mongos_loadbalanced ();
+      } else {
+         if (!test_framework_uri_apply_multi_mongos (
```
jmikola, line 351: Is `test_framework_uri_apply_multi_mongos` ever called with `false` for its second param?

src/libmongoc/tests/unified/entity-map.c
```diff
@@ -365,6 +392,9 @@ entity_client_new (entity_map_t *em, bson_t *bson, bson_error_t *error)
             mongoc_apm_set_command_failed_cb (callbacks, command_failed);
          } else if (0 == strcmp (event_type, "commandSucceededEvent")) {
             mongoc_apm_set_command_succeeded_cb (callbacks, command_succeeded);
+         } else if (is_unsupported_event_type (event_type)) {
+            MONGOC_DEBUG ("Skipping observing unsupported event type: %s", event_type);
```
jmikola, line 396: Would it make sense to error here and require the test be explicitly skipped?

src/libmongoc/tests/unified/operation.c
```diff
@@ -2518,6 +2526,19 @@ operation_loop (test_t *test,
...
+operation_assert_number_connections_checked_out (test_t *test,
+                                                 operation_t *op,
+                                                 result_t *result,
+                                                 bson_error_t *error)
+{
+   /* "This operation only applies to drivers that implement connection pooling
+    * and should be skipped for drivers that do not."
+    * TODO: (CDRIVER-3525) add this assertion when CMAP is implemented. */
+   result_from_ok (result);
+   return true;
```
jmikola, line 2539: If the tests should be skipped, should this just raise an error?

src/libmongoc/tests/unified/runner.c
```diff
@@ -1086,6 +1099,22 @@ test_check_event (test_t *test,
       bson_val_destroy (actual_val);
    }
 
+   if (expected_has_service_id) {
+      char oid_str[25] = {0};
+      bool has_service_id = false;
+
+      bson_oid_to_string (&actual->service_id, oid_str);
+      has_service_id = 0 != bson_oid_compare (&actual->service_id, &kZeroServiceId);
```
jmikola, line 1107: `kZeroServiceId` comes from `mongoc-server-description-private.h`, correct?

src/libmongoc/tests/unified/runner.c
```diff
@@ -1106,14 +1135,30 @@ test_check_expected_events_for_client (test_t *test,
...
    if (!bson_parser_parse (bp, expected_events_for_client, error)) {
       goto done;
    }
 
+   if (event_type) {
+      if (0 == strcmp (event_type, "cmap")) {
+         /* TODO: (CDRIVER-3525) Explicitly ignore cmap events until CMAP is
+          * supported. */
+         ret = true;
+         goto done;
```
jmikola, line 1153: This may overlap with my previous question but would it be preferable to error here and require skipping the tests?

src/libmongoc/tests/unified/util.c
```diff
@@ -163,3 +163,31 @@ test_bson_util_install (TestSuite *suite)
    TestSuite_Add (
       suite, "/unified/selftest/util/copy_and_sort", test_copy_and_sort);
 }
+
+/* TODO (CDRIVER-3525) add test support for CMAP events once the C driver
+ * supports CMAP. */
+bool is_unsupported_event_type (const char* event_type) {
+   char * unsupported_event_types[] = {
```
jmikola, line 170: Is `char * unsupported_event_types` something `clang-format` suggests? I don't recall seeing isolated asterisks in any previous code. Likewise for `char ** iter` below.

src/libmongoc/tests/TestSuite.c
kevinAlbs: Updated and removed the `matches` variable. An earlier iteration of these changes supported a `--skip` and `--after` flag. I removed them because they were unnecessary, but did not clean up this function.

> @kevinAlbs wiki.c2.com/?ElseConsideredSmelly is a good read if you have some spare time

Oh thank you, I have added that to my reading list.

src/libmongoc/tests/TestSuite.c
kevinAlbs: Updated.

src/libmongoc/tests/json-test.c
kevinAlbs: I am not sure I understand. Are you referring to an associative array type in C?

src/libmongoc/tests/json-test.c
```diff
@@ -1702,12 +1721,17 @@ run_json_general_test (const json_test_config_t *config)
...
           bson_iter_as_bool (&uri_iter)) {
-         ASSERT_OR_PRINT (
-            mongoc_uri_upsert_host_and_port (uri, "localhost:27017", &error),
-            error);
-         ASSERT_OR_PRINT (
-            mongoc_uri_upsert_host_and_port (uri, "localhost:27018", &error),
-            error);
+         if (test_framework_is_loadbalanced ()) {
+            mongoc_uri_destroy (uri);
+            uri = test_framework_get_uri_multi_mongos_loadbalanced ();
```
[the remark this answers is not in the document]
kevinAlbs: >> If useMultipleMongoses is true or unset, the test runner MUST use the URI of the load balancer fronting multiple servers

Thank you for bringing this up! I misread the behavior of `useMultipleMongoses` in the test readme for the Unified Test Format. I thought if it was omitted it should default to the URI fronting a single server. I have updated entity-map.c.

> When `useMultipleMongoses` _isn't_ used, `test_framework_get_uri` returns the appropriate URI for fronting a single server, correct?

Correct.

For the the non-unified tests, I think the omission of `useMultipleMongoses` should default to the URI fronting a single mongos.

Both the [Retryable Reads test format](https://github.com/mongodb/specifications/tree/df6be82f865e9b72444488fd62ae1eb5fca18569/source/retryable-reads/tests#test-format) and [Retryable Writes test format](https://github.com/mongodb/specifications/tree/df6be82f865e9b72444488fd62ae1eb5fca18569/source/retryable-writes/tests#test-format) note:

> useMultipleMongoses (optional): If true, the MongoClient for this test should be initialized with multiple mongos seed addresses. If false or omitted, only a single mongos address should be specified. This field has no effect for non-sharded topologies.

The [Change Streams test format](https://github.com/mongodb/specifications/tree/df6be82f865e9b72444488fd62ae1eb5fca18569/source/change-streams/tests#spec-test-runner) has no such note.

If that sounds right, I will submit a PR to the specification repo to update the notes in the retryable reads/writes test READMEs to account for load balancer behavior.

src/libmongoc/tests/test-libmongoc.c
kevinAlbs: Thank you for the suggestion. Updating to `mongoc_uri_new_with_error` and `ASSERT_OR_PRINT` is a good idea. Done.

src/libmongoc/tests/test-mongoc-loadbalanced.c
```diff
@@ -798,11 +774,9 @@ test_post_handshake_error_clears_pool (void)
 static int
 skip_if_not_loadbalanced (void)
 {
-   char *val = loadbalanced_uri ();
-   if (!val) {
+   if (!test_framework_is_loadbalanced ()) {
```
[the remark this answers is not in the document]
kevinAlbs: Done.

src/libmongoc/tests/test-mongoc-retryable-writes.c
```diff
@@ -32,13 +32,20 @@ retryable_writes_test_run_operation (json_test_ctx_t *ctx,
 }
 
 
+static test_skip_t skips[] = {
+   {"InsertOne fails after multiple retryable writeConcernErrors",
+    "Waiting on CDRIVER-3790"},
+   {0, 0}};
```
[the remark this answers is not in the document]
kevinAlbs: That is feasible here. Done.

src/libmongoc/tests/unified/entity-map.c
kevinAlbs: Haha, I am following the specification's lead with `useMultipleMongoses`. I wouldn't mind `useMultipleMongeese`.

src/libmongoc/tests/unified/entity-map.c
kevinAlbs: Yes, in this case if a spec test specified `useMultipleMongoses: false` on a non load balanced cluster, this would be called with false. `test_framework_uri_apply_multi_mongos` would verify the host list is one host.

src/libmongoc/tests/unified/entity-map.c
kevinAlbs: I'd argue it still has some value to run the test even if we cannot make assertions on CMAP events.

src/libmongoc/tests/unified/entity-map.c
kevinAlbs: Correct.

src/libmongoc/tests/unified/operation.c
kevinAlbs: We can still run the test and make other assertions. I have a preference for keeping this as-is.

src/libmongoc/tests/unified/runner.c
kevinAlbs: Updating to mongoses.

src/libmongoc/tests/unified/runner.c
kevinAlbs: Correct.

src/libmongoc/tests/unified/runner.c
kevinAlbs: I have a preference for skipping the CMAP assertions and still running the test. They won't be testing as much, but it seems better to run what we can rather than skip the full test.

src/libmongoc/tests/unified/runner.c
kevinAlbs: Updated to remove extra `else`.

src/libmongoc/tests/unified/util.c
kevinAlbs: I forgot to clang-format. Updated.

src/libmongoc/tests/unified/util.c
kevinAlbs: C89 prohibits declaring within the loop unfortunately.

build/evergreen_config_lib/tasks.py
```diff
@@ -1074,4 +1074,71 @@ def _check_allowed(self):
 
 all_tasks = chain(all_tasks, OCSPTask.matrix())
 
+class LoadBalancedTask(MatrixTask):
+    axes = OD([
+        ('asan', [True]),
+        ('build_ssl', ['openssl']), # The SSL library the C driver is built with.
+        ('test_ssl', [True, False]), # Whether tests are run with SSL connections.
```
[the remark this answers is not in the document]
chardan: Ok, got it, thanks.

src/libmongoc/tests/unified/util.c
chardan: No C17 for us.

chardan (approved):
With cleanups, +1 from me! Nice work on this.

jmikola (approved):

src/libmongoc/tests/json/change_streams/legacy/change-streams-resume-allowlist.json
kevinAlbs: I missed this in my prior responses. This is a resync of these tests from the specifications repo. Testing change stream spec tests is a requirement for load balancer tests. They were updated for load balancer. The particular change of `NotMaster` => `NotWritablePrimary` comes from another spec test change unrelated to load balancers. But I resync the entire file up to the necessary commit. It may pull in unrelated changes like this, but the alternative is manually applying just the changes you are interested in, and diverging from the files from the specifications repo.

PR #838, opened 2021-07-30
CDRIVER-4110: Default for causal_consistency depends on snapshot

jmikola:
https://jira.mongodb.org/browse/CDRIVER-4110

 * Removes a work-around from the unified test runner originally introduced in https://github.com/mongodb/mongo-c-driver/commit/d5f73e18eac7acaa05f31c4ffde68c9cd565704a.
 * Relocates the snapshot prose test, since it's part of the session spec (not transactions)
 * Use `mongoc_optional_t` in `mongoc_session_opt_t` to track values for `causalConsistency` and `snapshot` options, and enforce the `causalConsistency` logic in the corresponding getter method.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-client-session-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-client-session-private.h
@@ -38,14 +38,9 @@ struct _mongoc_transaction_opt_t {
    int64_t max_commit_time_ms;
 };
 
-typedef enum {
-   MONGOC_SESSION_NO_OPTS = 0,
-   MONGOC_SESSION_CAUSAL_CONSISTENCY = (1 << 0),
-   MONGOC_SESSION_SNAPSHOT = (2 << 0),
-} mongoc_session_flag_t;
-
 struct _mongoc_session_opt_t {
-   mongoc_session_flag_t flags;
+   mongoc_optional_t causal_consistency;
+   mongoc_optional_t snapshot;
    mongoc_transaction_opt_t default_txn_opts;
 };
 
--- a/src/libmongoc/tests/test-mongoc-client-session.c
+++ b/src/libmongoc/tests/test-mongoc-client-session.c
@@ -34,7 +34,7 @@ test_session_opts_clone (void)
 
    opts = mongoc_session_opts_new ();
    clone = mongoc_session_opts_clone (opts);
-   /* causal is enabled by default */
+   /* causalConsistency is enabled by default if snapshot is not enabled */
    BSON_ASSERT (mongoc_session_opts_get_causal_consistency (clone));
    mongoc_session_opts_destroy (clone);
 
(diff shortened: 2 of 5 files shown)


src/libmongoc/src/mongoc/mongoc-client-session-private.h
```diff
@@ -38,14 +38,9 @@ struct _mongoc_transaction_opt_t {
...
-typedef enum {
-   MONGOC_SESSION_NO_OPTS = 0,
-   MONGOC_SESSION_CAUSAL_CONSISTENCY = (1 << 0),
-   MONGOC_SESSION_SNAPSHOT = (2 << 0),
-} mongoc_session_flag_t;
-
 struct _mongoc_session_opt_t {
-   mongoc_session_flag_t flags;
+   mongoc_optional_t causal_consistency;
+   mongoc_optional_t snapshot;
```
benjirewis, line 43: Seems better to have these as optionals instead of on the same flag, thanks. Should the type be `mongoc_optional_bool_t` or is that type name change API-breaking?

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -47,6 +47,31 @@ test_session_opts_clone (void)
 }
 
 
+static void
+test_session_opts_causal_consistency_and_snapshot (void)
```
benjirewis, line 51: Thanks for adding this test, we now have a similar one in the Go driver.

benjirewis (approved):
LGTM!

src/libmongoc/src/mongoc/mongoc-client-session-private.h
jmikola: I brought that up with @kevinAlbs in Slack last week and we concluded that the ship has sailed since `mongoc_optional_t` is already in the public API (shipped with Versioned API in 1.18.0). If we end up introducing an optional int type down the line, perhaps we can deprecate the original name and introduce `mongoc_optional_bool_t` alongside `mongoc_optional_int_t`, but I don't think there's anything to do presently.

kevinAlbs: Good catch @jmikola!

Here is my understanding.

Before snapshot reads:
- `mongoc_session_opts_new` sets `causalConsistency=true`.

After snapshot reads:
- `mongoc_session_opts_new` sets `causalConsistency=true` and `snapshot=false`.
- Validation occurs in `mongoc_client_start_session` (i.e. both being true is an error).

The new behavior proposed in this PR:
- `mongoc_session_opts_new` sets `causalConsistency=unset` and `snapshot=unset`
- Defaults are applied in `mongoc_client_start_session` if the values are unset. Validation occurs in `mongoc_client_start_session`.

Admittedly, when reviewing the snapshot reads changes, I had thought the behavior was correct since it built off of the previous behavior. I had not considered introducing an unset value for `causalConsistency` and `snapshot`. This seems correct.

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -2707,12 +2732,44 @@ test_session_dirty (void *unused)
    _test_session_dirty_helper (false /* retry succceeds */);
 }
 
+void
+test_sessions_snapshot_prose_test_1 (void *ctx)
```
kevinAlbs, line 2736: Thank you for moving this test.

kevinAlbs (approved):
LGTM!

src/libmongoc/src/mongoc/mongoc-client-session-private.h
jmikola: Your understanding is correct.

I think much of the confusion actually relates in some way to [SPEC-936](https://jira.mongodb.org/browse/SPEC-936), which added the technical equivalent of legalese to state that the `causalConsistency` session option doesn't get a default value (which would require modeling it as a trilean), but instead inherits some default. In practice, I think some drivers (libmongoc included) just defaulted it to true as that had a functionally equivalent effect.

PR #839, opened 2021-08-02
CDRIVER-4068 add connectivity tests for serverless

kevinAlbs:
Add a connectivity test to ensure we can connect and send a `ping` command to a live serverless instance.

Tests against the latest server are failing due to a recent server change. It is described in CDRIVER-4120 and resolved in #843.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -479,6 +479,8 @@ functions:
         export REQUIRE_TLS12='${require_tls12}'
         export OBSOLETE_TLS='${obsolete_tls}'
         export VALGRIND='${valgrind}'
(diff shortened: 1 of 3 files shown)


chardan (approved):
+1 LGTM. Please bear in mind that I'm new to the test system, so you may want another set of eyes on it-- but from what I'm gleaning, this looks good. :-)

kevinAlbs:
> +1 LGTM. Please bear in mind that I'm new to the test system, so you may want another set of eyes on it-- but from what I'm gleaning, this looks good. :-)

Thank you for the review! Given the small scope of the change, I am OK with one reviewer.

PR #842, opened 2021-08-06
Automatically compile with CCache and link using LLD if possible

vector-of-bool:


chardan (approved):
I think this looks good! That said, I must point out that although I've leveled up a bit lately, I'm fair from being especially good with CMake. Have you tried this (or  will the tests do so) on different target platforms?

build/cmake/CCache.cmake
```diff
@@ -0,0 +1,22 @@
...
+    CCache support will be automatically enabled if it is found on the system.
+    CCache can be forced on or off by setting the MONGO_USE_CCACHE CMake option to
+    ON or OFF.
+]]
+
+# Find and enable ccache for compiling
+find_program(CCACHE_EXECUTABLE ccache)
+if(CCACHE_EXECUTABLE)
+    message(STATUS "Found ccache: ${CCACHE_EXECUTABLE}")
+    option(MONGO_USE_CCACHE "Use CCache when compiling" ON)
```
kevinAlbs, line 15: Nit: Is naming options with the project name a common practice?

I grepped the cmake files for `option (` and all of the existing cmake options are named like `ENABLE_*`. I'd suggest renaming this to `ENABLE_CCACHE` or something similar for consistency.

build/cmake/LLDLinker.cmake
```diff
@@ -0,0 +1,60 @@
...
+
+if(NOT COMMAND add_link_options)
+    # This only works on new-enough versions of CMake that support LINK_OPTIONS as
+    # a separate configuration entity
+    return()
+endif()
+
+include(CMakePushCheckState)
+include(CheckCSourceCompiles)
+cmake_push_check_state(RESET)
```
kevinAlbs, line 21: Oh this is neat. I was not aware of `cmake_push_check_state` and `cmake_pop_check_state`.

build/cmake/LLDLinker.cmake
```diff
@@ -0,0 +1,60 @@
...
+            puts("Hello, world!");
+            return 0;
+        }
+    ]] HAVE_LLD_LINKER_SUPPORT)
+cmake_pop_check_state()
+
+
+if(HAVE_LLD_LINKER_SUPPORT)
+    # Expose an option to toggle usage of lld
+    option(MONGO_USE_LLD "Link runtime binaries using LLVM's lld linker" ON)
```
kevinAlbs, line 38: Nit: Same question about option naming.

build/cmake/LLDLinker.cmake
```diff
@@ -0,0 +1,60 @@
...
+    # the environment. Issue a *one-time* diagnostic telling the user if they *almost*
+    # have LLD support. This branch is only for diagnostic purposes.
+    find_program(_MONGO_LD_LLD_LINKER
+        NAMES ld.lld ld.lld-13 ld.lld-12 ld.lld-11 ld.lld-10 ld.lld-9 ld.lld-8 ld.lld-7 ld.lld-6 ld.lld-5
+        )
+    mark_as_advanced(_MONGO_LD_LLD_LINKER)
+    # If we found one, we are compiling with GCC, *and* the found lld has a version suffix, issue a message
+    # telling the user how they might be able to get lld work.
+    if(_MONGO_LD_LLD_LINKER AND CMAKE_C_COMPILER_ID STREQUAL "GNU" AND NOT _MONGO_LD_LLD_LINKER MATCHES "ld.lld$")
+        message(STATUS "NOTE: A GNU-compatible lld linker was found (${_MONGO_LD_LLD_LINKER}), but support from GCC requires that")
```
kevinAlbs, line 50: Great! That is helpful and actionable for the user.

kevinAlbs (approved):
Changes LGTM! I only have a question about the option names.

build/cmake/CMakeLists.txt
```diff
@@ -11,6 +11,8 @@ set (build_cmake_MODULES
    MongoCPackage.cmake
    ParseVersion.cmake
    SphinxBuild.cmake
+   CCache.cmake
+   LLDLinker.cmake
```
eramongodb, line 15: I suggest renaming `CCache.cmake` and `LLDLinker.cmake` to `FindCcache.cmake` and `FindLLD.cmake` respectively, following the convention used by `FindSnappy.cmake`, `FindSASL2.cmake`, etc.

build/cmake/LLDLinker.cmake
```diff
@@ -0,0 +1,60 @@
...
+include(CMakePushCheckState)
+include(CheckCSourceCompiles)
+cmake_push_check_state(RESET)
+    # Newer GNU compilers support lld with the '-fuse-ld=lld' flag.
+    set(CMAKE_REQUIRED_FLAGS "-fuse-ld=lld")
+    set(CMAKE_REQUIRED_LINK_OPTIONS "-fuse-ld=lld")
+    check_c_source_compiles([[
+        #include <stdio.h>
+
+        int main() {
```
eramongodb, line 28: `int main (void)` for C language?

rcsanchez97 (changes requested):
Apart from the comments made by the other members of the team, I have the following two observations:

- The general style of our `CMakeLists.txt` files is to have a space preceding an opening parenthesis (as we do in the C code): `if (CCACHE_EXECUTABLE)` rather than `if(CCACHE_EXECUTABLE)`.  Since we don't have a linter or tidy for CMake, you will need to trust me on this :-)
- These two options have the potential to impact the build in a significant enough way that if they do in fact become default, as your PR implements them, then they should be documented in `src/libmongoc/doc/installing.rst` (I recommend around line 172).

On the issue of making these new options the default, is there a possibility that making either or both of these the default will produce build behavior or build output which the user would otherwise not expect?

build/cmake/CCache.cmake
vector-of-bool: The reason I didn't go with `ENABLE_` names is because these options do not affect the build result, whereas the others do. There's no specific convention on qualifying variable names with the project name, but I've always had a habit of doing it.

build/cmake/CMakeLists.txt
eramongodb: I did not mean to suggest supporting `find_package()` usage. This was just about filename consistency. For example, `FindSnappy.cmake` is included as [`include (FindSnappy)`](https://github.com/mongodb/mongo-c-driver/blob/932e7e4861e28a719969cf716cdc4c10473e78ee/src/libmongoc/CMakeLists.txt#L353), and `FindSASL2` is included as [`include (FindSASL2)`](https://github.com/mongodb/mongo-c-driver/blob/932e7e4861e28a719969cf716cdc4c10473e78ee/src/libmongoc/CMakeLists.txt#L252); neither are used with, nor appear to designed to be used with, `find_package()`.

build/cmake/CCache.cmake
eramongodb: Unless we take this opportunity to establish a proper naming convention for CMake variables in the C Driver (and also the C++ Driver?), I suppose there is freedom of choice here.

My remarks about `ENABLE_*` naming are not motivated by their corresonding behavior, but from the perspective of users configuring CMake variables via tools such as `ccache` or CMake GUI, where (as far as I know) options the user may care to toggle are all grouped together with the `ENABLE_*` prefix.

I believe there is a UX and documentation argument to be made for consistency on this front, but this may be beyond the scope of this particular PR.

build/cmake/CMakeLists.txt
vector-of-bool: CMake modules with the magic form `FindFoo` are intended to be used with `find_package(Foo)`, which makes `include(FindSnappy)` and `include(FindSASL2)` to be "wrong", even if they incidentally "just work" as they currently are. It's likely an accident that they were given that name and then used with `include()`. Those should probably be patched up to use `find_package_handle_standard_args()` (and maybe imported libraries as well).

CMakeLists.txt
```diff
@@ -70,6 +70,11 @@ set (CMAKE_MODULE_PATH
 include (InstallRequiredSystemLibraries)
 include (GNUInstallDirs)
 
+# Enable CCache, if possible
+include(CCache)
```
rcsanchez97, line 74: Please add a space before the opening parenthesis.

CMakeLists.txt
```diff
@@ -70,6 +70,11 @@ set (CMAKE_MODULE_PATH
...
+# Link with LLD, if possible
+include(LLDLinker)
```
rcsanchez97, line 76: Please add a space before the opening parenthesis.

rcsanchez97 (commented):
That sounds good.  Based on the ability of these tools to act as drop-in replacements for the more commonly used tools, it seems reasonable to proceed with the expectation that there should be no user-visible changes.

As far as Evergreen, I don't think that there is a way to exercise this in an Evergreen build at this time.  If that is something that we want to do, for instance to ensure that we don't strange behavior in the unit tests resulting from the toolchain variation, then we would need to start by requesting that the Build team install the necessary additional components into one or more Evergreen host images.  Then we would be able to add a build task that does whatever we decide we need in order to be confident that our build works properly.  If that is something that we want to do, I would recommend handling it as a separate task from this initial change.

build/cmake/CMakeLists.txt
eramongodb: Fair points. Agreed that `Find<Name>.cmake` files are usually reserved for `find_package()` usage.

rcsanchez97 (approved):
LGTM!

PR #843, opened 2021-08-07
CDRIVER-4093 add versioned API strict examples 5-8

kevinAlbs:
This also includes a fix for tests against latest server builds (CDRIVER-4120).

### DIFF
--- a/src/libmongoc/tests/test-mongoc-primary-stepdown.c
+++ b/src/libmongoc/tests/test-mongoc-primary-stepdown.c
@@ -241,8 +241,7 @@ test_not_primary_keep_pool (mongoc_client_t *client)
       coll, tmp_bson ("{'test': 1}"), NULL, NULL, &error);
    ASSERT (!res);
    ASSERT_CMPINT (error.code, ==, 10107);
-   ASSERT_CONTAINS (error.message,
-                    "Failing command due to 'failCommand' failpoint");
+   ASSERT_CONTAINS (error.message, "failpoint");
 
    /* Execute a second insert, verify that it succeeds */
    res = mongoc_collection_insert_one (
@@ -308,8 +307,7 @@ test_not_primary_reset_pool (mongoc_client_t *client)
       coll, tmp_bson ("{'test': 1}"), NULL, NULL, &error);
    ASSERT (!res);
    ASSERT_CMPINT (error.code, ==, 10107);
-   ASSERT_CONTAINS (error.message,
-                    "Failing command due to 'failCommand' failpoint");
+   ASSERT_CONTAINS (error.message, "failpoint");
 
    /* Verify that the pool has been cleared */
    ASSERT_CMPINT ((conn_count + 1), ==, _connection_count (client, primary_id));
@@ -376,8 +374,7 @@ test_shutdown_reset_pool (mongoc_client_t *client)
       coll, tmp_bson ("{'test': 1}"), NULL, NULL, &error);
    ASSERT (!res);
    ASSERT_CMPINT (error.code, ==, 91);
-   ASSERT_CONTAINS (error.message,
-                    "Failing command due to 'failCommand' failpoint");
+   ASSERT_CONTAINS (error.message, "failpoint");
 
    /* Verify that the pool has been cleared */
    ASSERT_CMPINT ((conn_count + 1), ==, _connection_count (client, primary_id));
@@ -444,8 +441,7 @@ test_interrupted_shutdown_reset_pool (mongoc_client_t *client)
       coll, tmp_bson ("{'test': 1}"), NULL, NULL, &error);
    ASSERT (!res);
    ASSERT_CMPINT (error.code, ==, 11600);
-   ASSERT_CONTAINS (error.message,
-                    "Failing command due to 'failCommand' failpoint");
+   ASSERT_CONTAINS (error.message, "failpoint");
 
    /* Verify that the pool has been cleared */
    ASSERT_CMPINT ((conn_count + 1), ==, _connection_count (client, primary_id));
--- a/src/libmongoc/tests/test-mongoc-sample-commands.c
+++ b/src/libmongoc/tests/test-mongoc-sample-commands.c
@@ -3711,8 +3711,8 @@ static int64_t iso_to_unix (const char* iso_str) {
 
 static void _test_sample_versioned_api_example_5_6_7_8 (void) {
 #define N_DOCS 8
-   mongoc_client_t *client = NULL;
-   mongoc_server_api_t *server_api = NULL;
+   mongoc_client_t *client;
+   mongoc_server_api_t *server_api;
    mongoc_server_api_version_t server_api_version;
    bool ok;
    bson_error_t error;
--- a/src/libmongoc/tests/test-mongoc-sample-commands.c
+++ b/src/libmongoc/tests/test-mongoc-sample-commands.c
@@ -3703,6 +3703,138 @@ _test_sample_versioned_api_example_4 (void)
    mongoc_server_api_destroy (server_api);
 }
 
+static int64_t iso_to_unix (const char* iso_str) {
+   /* TODO (CDRIVER-2945) there is no convenient helper for converting ISO8601
+    * strings to Unix timestamps. This is not shown in the example. */
+   return 1628330345;
+}
+
+static void _test_sample_versioned_api_example_5_6_7_8 (void) {
+#define N_DOCS 8
+   mongoc_client_t *client = NULL;
+   mongoc_server_api_t *server_api = NULL;
+   mongoc_server_api_version_t server_api_version;
+   bool ok;
+   bson_error_t error;
+   mongoc_database_t *db;
+   mongoc_collection_t *sales;
+   bson_t *docs[N_DOCS];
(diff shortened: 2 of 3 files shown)


src/libmongoc/tests/test-mongoc-primary-stepdown.c
```diff
@@ -241,8 +241,7 @@ test_not_primary_keep_pool (mongoc_client_t *client)
       coll, tmp_bson ("{'test': 1}"), NULL, NULL, &error);
    ASSERT (!res);
    ASSERT_CMPINT (error.code, ==, 10107);
-   ASSERT_CONTAINS (error.message,
-                    "Failing command due to 'failCommand' failpoint");
+   ASSERT_CONTAINS (error.message, "failpoint");
```
kevinAlbs, line 244: SERVER-57040 changed the error message returned from errors triggered by failpoints.

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -3703,13 +3703,146 @@ _test_sample_versioned_api_example_4 (void)
...
+       * namespace does not exist. */
+      ASSERT_OR_PRINT (ok, error);
+   }
+
+   /* Start Versioned API Example 5 */
+   docs[0] = BCON_NEW ("_id", BCON_INT32 (1),
+                       "item", "abc",
+                       "price", BCON_INT32 (10),
+                       "quantity", BCON_INT32 (2),
+                       "date", BCON_DATE_TIME (iso_to_unix ("2021-01-01T08:00:00Z")));
```
kevinAlbs, line 3750: The `iso_to_unix` function is a no-op. It is not present in the documentation example (i.e. not within the `Start` / `End` comments. The intent is to convey that there should be some conversion to the unix timestamp so this can be expressed as a BSON datetime type.

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -3703,13 +3703,146 @@ _test_sample_versioned_api_example_4 (void)
...
+      db, cmd, NULL /* read_prefs */, &reply, &error);
+   ASSERT_ERROR_CONTAINS (error,
+                          MONGOC_ERROR_SERVER,
+                          323,
+                          "Provided apiStrict:true, but the command count is not in API Version 1");
+   ASSERT (!ok);
+   bson_destroy (&reply);
+#if 0
+   /* This block not evaluated, but is inserted into documentation to represent the above reply.
+    * Don't delete me! */
```
kevinAlbs, line 3803: Credit of this idea goes to the analogous C++ examples of CXX-2314 :)

src/libmongoc/tests/test-mongoc-primary-stepdown.c
chardan: General question: are the changes usually slow enough that all drivers and their tests can just play follow-up? Has there been thought put toward storing them in a shared resource file (or some other strategy to centralize them)? Is that valuable?

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -3703,13 +3703,146 @@ _test_sample_versioned_api_example_4 (void)
...
 }
 
+static int64_t iso_to_unix (const char* iso_str) {
+   /* TODO (CDRIVER-2945) there is no convenient helper for converting ISO8601
+    * strings to Unix timestamps. This is not shown in the example. */
+   return 1628330345;
+}
+
+static void _test_sample_versioned_api_example_5_6_7_8 (void) {
+#define N_DOCS 8
```
chardan, line 3713: It looks like this is only used once; do we need this macro?

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -3703,13 +3703,146 @@ _test_sample_versioned_api_example_4 (void)
...
+   mongoc_client_t *client = NULL;
```
chardan, line 3714: nit: Some of the variables below are initialized, and others aren't. Especially since this is test code (not that that suggests a lower standard, but rather that it means there may be specific reasons to do certain things) I'm not against it, but it does feel inconsistent.

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -3703,13 +3703,146 @@ _test_sample_versioned_api_example_4 (void)
...
+   docs[7] = BCON_NEW ("_id", BCON_INT32 (8),
+                       "item", "abc",
+                       "price", BCON_INT32 (10),
+                       "quantity",BCON_INT32 (5),
+                       "date", BCON_DATE_TIME (iso_to_unix ("2021-03-16T20:20:13Z")));
+   ok = mongoc_collection_insert_many (
+      sales, (const bson_t**) docs, N_DOCS, NULL /* opts */, &reply, &error);
+   /* End Versioned API Example 5 */
+   ASSERT_OR_PRINT (ok, error);
+   bson_destroy (&reply);
```
chardan, line 3790: Is "docs", etc. also released?

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -3703,13 +3703,146 @@ _test_sample_versioned_api_example_4 (void)
...
+   char *str = bson_as_json (&reply, NULL /* length */);
+   printf ("%s", str);
+   /* Prints the server reply:
+    * { "ok" : 0, "errmsg" : "Provided apiStrict:true, but the command count is not in API Version 1", "code" : 323, "codeName" : "APIStrictError" } */
+   bson_free (str);
+   /* End Versioned API Example 6 */
+#endif
+
+   /* Start Versioned API Example 7 */
+   filter = bson_new ();
```
chardan, line 3814: Ok, and looks like this is destroyed later, via bson_destroy(). +1

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -3703,13 +3703,146 @@ _test_sample_versioned_api_example_4 (void)
...
+   bson_destroy (&reply);
+
+   /* Start Versioned API Example 8 */
+   BSON_ASSERT (count == N_DOCS);
+   /* End Versioned API Example 8 */
+
+   bson_destroy (filter);
+   bson_destroy (cmd);
+   for (i = 0; i < N_DOCS; i++) {
+      bson_destroy (docs[i]);
```
chardan, line 3830: Ah! docs_entries are destroyed here, indeed!

chardan (approved):

vector-of-bool (approved):
Looks good to me 👍.

src/libmongoc/tests/test-mongoc-primary-stepdown.c
kevinAlbs: > are the changes usually slow enough that all drivers and their tests can just play follow-up?

In practice, it is rare that an error message change occurs and requires an update to tests. If I had to guess, once every few months.

> Has there been thought put toward storing them in a shared resource file (or some other strategy to centralize them)? Is that valuable?

That is an interesting idea. I do not know of any effort to centralize error messages. IMO drivers should prefer not checking error messages. There is no guarantee they remain stable server side. I believe the order of preference for driver error handling is:

1. Check the reply `errorLabels`.
2. Check the reply `code`.
3. Check the reply `message`.

An example of (1) being preferred to (2) is [retryable writes error handling](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst#retryablewriteerror-labels). Error labels convey the intent of handling from the server. It gives the server the flexibility to add / remove errors that are considered retryable. In pre 4.4 servers, there are no `errorLabels`, so there is a fallback list of error codes.

An example of (2) being preferred to (3) is the [SDAM](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#not-writable-primary-and-node-is-recovering) error handling. It checks for an error code first, but falls back to error message checking.

src/libmongoc/tests/test-mongoc-sample-commands.c
kevinAlbs: Yup!

src/libmongoc/tests/test-mongoc-sample-commands.c
kevinAlbs: Yes, the individual documents are released in the for loop on L3829.

src/libmongoc/tests/test-mongoc-sample-commands.c
kevinAlbs: I think so. `N_DOCS` is also used on the `mongoc_collection_insert_many` call on L3786, the assertions on L3818 and L3824, and the cleanup on L3829.

src/libmongoc/tests/test-mongoc-sample-commands.c
kevinAlbs: True, there is no reason to initialize `client` or `server_api`, as they are set immediately after. I removed the initialization for consistency.

kevinAlbs (commented):
Thank you for the review! Merging.

PR #844, opened 2021-08-10
CDRIVER-4104: Filter out events that have been redacted when observeSensitiveComands is false

vector-of-bool:
Closes CDRIVER-4104

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-apm-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-apm-private.h
@@ -206,11 +206,8 @@ void
 mongoc_apm_command_failed_cleanup (mongoc_apm_command_failed_t *event);
 
 bool
-mongoc_apm_is_sensitive_command (const char *command_name,
-                                 const bson_t *command);
-
-bool
-mongoc_apm_is_sensitive_reply (const char *command_name, const bson_t *reply);
+mongoc_apm_is_sensitive_command_message (const char *command_name,
+                                         const bson_t *command);
 
 BSON_END_DECLS
 
--- a/src/libmongoc/src/mongoc/mongoc-apm-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-apm-private.h
@@ -205,9 +205,16 @@ mongoc_apm_command_failed_init (mongoc_apm_command_failed_t *event,
 void
 mongoc_apm_command_failed_cleanup (mongoc_apm_command_failed_t *event);
 
+/**
+ * @brief Determine whether the given command-related message is a "sensitive
+ * command."
+ *
+ * @param command_name The name of the command being checked
+ * @param body The body of the command request, reply, or failure.
+ */
 bool
 mongoc_apm_is_sensitive_command_message (const char *command_name,
-                                         const bson_t *command);
+                                         const bson_t *body);
 
 BSON_END_DECLS
 
--- a/src/libmongoc/src/mongoc/mongoc-apm.c
+++ b/src/libmongoc/src/mongoc/mongoc-apm.c
@@ -110,7 +110,7 @@ mongoc_apm_command_started_init (mongoc_apm_command_started_t *event,
       event->command_owned = false;
    }
 
-   if (mongoc_apm_is_sensitive_command (command_name, command)) {
+   if (mongoc_apm_is_sensitive_command_message (command_name, command)) {
       if (!event->command_owned) {
          event->command = bson_copy (event->command);
          event->command_owned = true;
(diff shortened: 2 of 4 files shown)


src/libmongoc/src/mongoc/mongoc-apm.c
```diff
@@ -955,7 +955,9 @@ mongoc_apm_is_sensitive_command (const char *command_name,
       return false;
    }
 
-   return bson_has_field (command, "speculativeAuthenticate");
+   /* An empty command means it has been redacted */
+   return bson_count_keys (command) == 0 ||
+          bson_has_field (command, "speculativeAuthenticate");
```
kevinAlbs, line 960: `mongoc_apm_is_sensitive_command` is called in two places:

1. `mongoc_apm_command_started_init` to determine if the command should be redacted.
2. `is_sensitive_command` in entity-map.c, called by `should_ignore_event` to determine if the command should be ignored instead of captured.

(1) should not be calling this with an empty document.

Should we add the empty document check in `is_sensitive_command`?

src/libmongoc/src/mongoc/mongoc-apm.c
```diff
@@ -981,7 +983,9 @@ mongoc_apm_is_sensitive_reply (const char *command_name, const bson_t *reply)
       return false;
    }
 
-   return bson_has_field (reply, "speculativeAuthenticate");
+   /* An empty command means it has been redacted */
+   return bson_count_keys (reply) == 0 ||
```
kevinAlbs, line 987: Similar question about whether this check should be moved to `is_sensitive_command`.

src/libmongoc/tests/unified/entity-map.c
```diff
@@ -161,14 +161,12 @@ should_ignore_event (entity_t *client_entity, event_t *event)
       return true;
    }
 
-   if (!client_entity->ignore_command_monitoring_events) {
```
kevinAlbs, line 164: Oh good catch! It looks like this returned early if a client entity was not specified with `ignoreCommandMonitoringEvents`

src/libmongoc/src/mongoc/mongoc-apm.c
```diff
@@ -955,7 +955,9 @@ mongoc_apm_is_sensitive_command (const char *command_name,
       return false;
    }
 
-   return bson_has_field (command, "speculativeAuthenticate");
+   /* An empty command means it has been redacted */
+   return bson_count_keys (command) == 0 ||
```
kevinAlbs, line 959: Nit: `bson_empty` can also be used to check for an empty document.

kevinAlbs (commented):
Looks good! I have a suggestion for moving the empty document check.

src/libmongoc/src/mongoc/mongoc-apm.c
vector-of-bool: Probably, yes.

vector-of-bool: I went a different route to clean things up a bit.

src/libmongoc/src/mongoc/mongoc-apm-private.h
```diff
@@ -206,11 +206,8 @@ void
 mongoc_apm_command_failed_cleanup (mongoc_apm_command_failed_t *event);
 
 bool
-mongoc_apm_is_sensitive_command (const char *command_name,
-                                 const bson_t *command);
-
-bool
-mongoc_apm_is_sensitive_reply (const char *command_name, const bson_t *reply);
+mongoc_apm_is_sensitive_command_message (const char *command_name,
+                                         const bson_t *command);
```
kevinAlbs, line 210:
```suggestion
                                         const bson_t *body);
```

Can you note with a comment that `body` could be a command or a reply?

kevinAlbs (commented):
Can you investigate the test failures attached to this PR?

E.g. `test-5.0-server-noauth-sasl-winssl` has relevant logs (grepping for `"fail"`):
https://evergreen.mongodb.com/lobster/evergreen/task/mongo_c_driver_windows_2017_test_5.0_server_noauth_sasl_winssl_patch_b80d00f0993501b21193b1869f43949d3e80a044_611ac2941e2d170d5f5cf323_21_08_16_19_55_02/0/task#bookmarks=0%2C6011&l=1&shareLine=2223

vector-of-bool:
Looks like there were some goofs on some of the CI agents, but otherwise fixed those two failing tests.

kevinAlbs (approved):
LGTM!

PR #845, opened 2021-08-10
CDRIVER-4106 Update numbers for existing SRV polling prose tests.

eramongodb:
Only [prose test 9](https://github.com/mongodb/specifications/blob/df6be82f865e9b72444488fd62ae1eb5fca18569/source/polling-srv-records-for-mongos-discovery/tests/README.rst#9-test-that-srv-polling-is-not-done-for-load-balalanced-clusters) is [currently implemented](https://github.com/mongodb/mongo-c-driver/blob/0d6a8ba27ed5cd94a26e4db879f61b545799193f/src/libmongoc/tests/test-mongoc-dns.c#L646) according to spec. Behavior of [`test_srv_polling_mocked`](https://github.com/mongodb/mongo-c-driver/blob/0d6a8ba27ed5cd94a26e4db879f61b545799193f/src/libmongoc/tests/test-mongoc-dns.c#L500) does not conform to prose tests described by spec. Therefore, only names related to prose test 9 are updated with numbering.

Prose tests 1 through 8 are to be implemented separately; see CDRIVER-4125.

---

This resolves CDRIVER-4106.

Names of tests and test functions are modelled after existing numbered prose tests such as those in [test-mongoc-crud.c](https://github.com/mongodb/mongo-c-driver/blob/0d6a8ba27ed5cd94a26e4db879f61b545799193f/src/libmongoc/tests/test-mongoc-crud.c#L173).

### DIFF
--- a/src/libmongoc/tests/test-mongoc-dns.c
+++ b/src/libmongoc/tests/test-mongoc-dns.c
@@ -633,8 +633,8 @@ _prose_loadbalanced_ping (mongoc_client_t *client)
 }
 
 /*
-   Implements the following prose test described in the SRV polling test README:
-   Test that SRV polling is not done for load balalanced clusters. Connect to
+   Implements prose test 9 as described in the SRV polling test README:
+   Test that SRV polling is not done for load balanced clusters. Connect to
    mongodb+srv://test3.test.build.10gen.cc/?loadBalanced=true, mock the addition
    of the following DNS record, wait until 2*rescanSRVIntervalMS, and assert
    that the final topology description only contains one server
@@ -643,7 +643,7 @@ _prose_loadbalanced_ping (mongoc_client_t *client)
    localhost.test.build.10gen.cc.
 */
 static void
-_prose_loadbalanced_run (bool pooled)
+_prose_test_9 (bool pooled)
 {
    mongoc_client_pool_t *pool;
    mongoc_uri_t *uri;
@@ -717,15 +717,15 @@ _prose_loadbalanced_run (bool pooled)
 }
 
 static void
-test_prose_loadbalanced_single (void *unused)
+prose_test_9_single (void *unused)
 {
-   _prose_loadbalanced_run (false);
+   _prose_test_9 (false);
 }
 
 static void
-test_prose_loadbalanced_pooled (void *unused)
+prose_test_9_pooled (void *unused)
 {
-   _prose_loadbalanced_run (true);
+   _prose_test_9 (true);
 }
 
 void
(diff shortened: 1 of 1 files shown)


src/libmongoc/tests/test-mongoc-dns.c
```diff
@@ -643,7 +643,7 @@ _prose_loadbalanced_ping (mongoc_client_t *client)
    localhost.test.build.10gen.cc.
 */
 static void
-_prose_loadbalanced_run (bool pooled)
+_prose_test_9 (bool pooled)
```
chardan, line 646: nit: ...on the other hand, is something like "_prose_test_9_loadbalanced_run" more clear? If I were looking for loadbalanced tests, "prose_test_9" probably wouldn't show up in my search. I don't have too strong of an opinion on this, though.

chardan (commented):
I think this looks ok (I did have a question about naming), but I'm not yet very familiar with these tests, so it's tough for me to give solid direction beyond my impressions. But, you've got a few other reviewers who may have more concrete ideas. :-)

src/libmongoc/tests/test-mongoc-dns.c
eramongodb: I believe it more likely that one would search for this specific test by its prose test number (per SRV polling test specification) rather than by name. I believe the comments ("load balanced clusters") and referenced entities (`MONGOC_URI_LOADBALANCED`) should provide enough context for this test to be discoverable regardless.

src/libmongoc/tests/test-mongoc-dns.c
```diff
@@ -633,8 +633,8 @@ _prose_loadbalanced_ping (mongoc_client_t *client)
 }
 
 /*
-   Implements the following prose test described in the SRV polling test README:
-   Test that SRV polling is not done for load balalanced clusters. Connect to
+   Implements prose test 9 as described in the SRV polling test README:
+   Test that SRV polling is not done for load balanced clusters. Connect to
```
eramongodb, line 637: Added a minor typo fix to the set of changes which I had forgotten to include prior.

kevinAlbs: Good eye!

src/libmongoc/tests/test-mongoc-dns.c
kevinAlbs: The consistency with the naming of prose tests from test-mongoc-crud.c seems like additional rationale for the name `_prose_test_9`. I am OK with either proposed name.

kevinAlbs (approved):
LGTM! Thank you for including a detailed description.

src/libmongoc/tests/test-mongoc-dns.c
chardan: Okay, sounds good!

vector-of-bool (approved):
LGTM. My only questions/nits were already asked an answered in the associated conversations.

chardan (approved):
+1, great work getting this through! :-)

PR #847, opened 2021-08-11
CDRIVER-4113: Thread-safe pool for server session objects

vector-of-bool:


src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -1570,45 +1539,49 @@ _mongoc_topology_pop_server_session (mongoc_topology_t *topology,
...
-   while (topology->session_pool) {
-      ss = topology->session_pool;
-      CDL_DELETE (topology->session_pool, ss);
-      /* Sessions do not expire when the topology type is load balanced. */
-      if (loadbalanced) {
+   while (true) {
+      bool got_session;
+      ss = mongoc_ts_pool_try_pop (topology->session_pool, &got_session);
+      if (!got_session) {
+         /* The pool is empty */
```
kevinAlbs, line 1566: IIUC, `got_session` may be stale. We may end up creating a few more sessions in the event that many operations are concurrently popping/pushing sessions onto the pool.

Is that correct? If so, that seems like a trade-off to this approach.

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -1633,46 +1606,25 @@ _mongoc_topology_push_server_session (mongoc_topology_t *topology,
    ENTRY;
 
    bson_mutex_lock (&topology->mutex);
-
    timeout = topology->description.session_timeout_minutes;
    loadbalanced = topology->description.type == MONGOC_TOPOLOGY_LOAD_BALANCED;
+   bson_mutex_unlock (&topology->mutex);
 
-   /* start at back of queue and reap timed-out sessions */
```
kevinAlbs, line 1640: This does diverge from the driver specification's [Algorithm to return a ServerSession instance to the server session pool](https://github.com/mongodb/specifications/blob/df6be82f865e9b72444488fd62ae1eb5fca18569/source/sessions/driver-sessions.rst#algorithm-to-return-a-serversession-instance-to-the-server-session-pool)

> Before returning a server session to the pool a driver MUST first check the server session pool for server sessions at the back of the queue that are about to expire (meaning they will expire in less than one minute). A driver MUST stop checking server sessions once it encounters a server session that is not about to expire. Any server sessions found that are about to expire are removed from the end of the queue and discarded (or allowed to be garbage collected)

Would adding this functionality back nullify the performance improvement gained here?

src/libmongoc/src/mongoc/mongoc-ts-pool-private.h
```diff
@@ -0,0 +1,97 @@
...
+ * @brief A thread-safe pool.
+ *
+ * Items are allocated using `mongo_ts_pool_alloc_item` and should be deleted
+ * with `mongo_ts_pool_free_item`. Items are added to the pool with
+ * `mongo_ts_pool_push` and taken from the pool using `mongo_ts_pool_try_pop`.
+ */
+typedef struct _mongoc_ts_pool *mongoc_ts_pool;
+
+/**
+ * @brief Create a new thread-safe pool
```
kevinAlbs, line 17: Awesome, this seems generally useful!

Can you add unit tests of `mongoc_ts_pool` (akin to test-mongoc-generation-map.c)?

If we need to modify this data structure in the future, I think it would be easier to identify regressions in unit tests.

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -327,12 +327,7 @@ _test_session_pool_reap (bool pooled)
     */
    b->server_session->last_used_usec = bson_get_monotonic_time ();
    mongoc_client_session_destroy (b);
-   BSON_ASSERT (client->topology->session_pool);
-   ASSERT_SESSIONS_MATCH (&lsid_b, &client->topology->session_pool->lsid);
```
kevinAlbs, line 331: Instead of removing the assertion of lsid, can this use `mongoc_ts_pool_try_pop` to get session B, check the resulting `lsid`, then assert that the session pool is empty?

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -1243,36 +1238,6 @@ session_test_new (session_test_correct_t correct_client,
    return test;
 }
 
-
-static void
-check_session_returned (session_test_t *test, const bson_t *lsid)
```
kevinAlbs, line 1248: To avoid losing test coverage, could this instead pop everything from the pool, check for the `lsid`, then push it back in reverse order?

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -2280,7 +2234,7 @@ test_cursor_implicit_session (void *ctx)
    BSON_ASSERT (cursor->client_session);
    BSON_ASSERT (!cursor->explicit_session);
    bson_copy_to (&cursor->client_session->server_session->lsid, &find_lsid);
-   ASSERT_POOL_SIZE (topology, 0);
+   BSON_ASSERT (mongoc_ts_pool_size (topology->session_pool) == 0);
```
kevinAlbs, line 2237: I suggest using `ASSERT_CMPINT`, which prints the values being compared on failure.

kevinAlbs (commented):
Looks like a great start! I like the generally useful `mongoc_ts_pool` type.

src/libmongoc/src/mongoc/mongoc-topology.c
vector-of-bool: The `did_pop` bool is somewhat an artefact of an earlier design where you can push/pop arbitrary pointers (including `NULL`), but because of changes to have `alloc_item` and `free_item`, you can never push/pop a `NULL` from the pool, so it is entirely reliable to just check the returned pointer for `NULL` instead of the bool value. I can remove the `did_pop` bool parameter and just check the returned pointer instead.

src/libmongoc/src/mongoc/mongoc-ts-pool-private.h
```diff
@@ -0,0 +1,97 @@
+#ifndef MONGOC_TS_POOL_PRIVATE_H
+#define MONGOC_TS_POOL_PRIVATE_H
+
+#include <stddef.h>
+#include <stdbool.h>
```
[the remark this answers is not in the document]
vector-of-bool: I'm going to remove the bool parameters, so that's just going to disappear, but the comment reminds me that I would like to discuss the project's `#include`-hygiene, not specifically related to this PR, but in general.

src/libmongoc/src/mongoc/mongoc-ts-pool-private.h
```diff
@@ -0,0 +1,97 @@
...
+ */
+typedef struct _mongoc_ts_pool *mongoc_ts_pool;
+
+/**
+ * @brief Create a new thread-safe pool
+ *
+ * @param element_size The size of the queue elements
+ * @param element_dtor A destructor function to run when an item is removed
+ */
+extern mongoc_ts_pool
```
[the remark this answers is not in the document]
vector-of-bool: The `extern` is redundant for function declarations, but is just a force of habbit.

src/libmongoc/src/mongoc/mongoc-ts-pool.c
```diff
@@ -0,0 +1,117 @@
+#include "./mongoc-ts-pool-private.h"
```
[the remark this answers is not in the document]
vector-of-bool: Another force of habit related to the goofy way `#include` search paths are specified by the language (i.e. barely at all).

src/libmongoc/src/mongoc/mongoc-ts-pool.c
```diff
@@ -0,0 +1,117 @@
...
+#include "common-thread-private.h"
+
+#include "bson/bson.h"
+
+typedef struct pool_node {
+   struct pool_node *next;
+   max_align_t data[1];
```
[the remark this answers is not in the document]
vector-of-bool: `max_align_t` ensures that that data member is aligned to the maximum possible data alignment of the target CPU architecture, such that when I over-allocate the struct in `alloc_item` and then type-pun that data member to some `T`, it is guaranteed to be at-least as aligned as `T`. If I just made it `char` or something arbitrary, it could be under-aligned and then type-punning the data member would be UB. If there's another way to detect the max-alignment type, that would work in place of the `max_align_t`.

src/libmongoc/src/mongoc/mongoc-client-session-private.h
```diff
@@ -102,16 +101,17 @@ _mongoc_client_session_handle_reply (mongoc_client_session_t *session,
                                      const char *cmd_name,
                                      const bson_t *reply);
 
-mongoc_server_session_t *
-_mongoc_server_session_new (bson_error_t *error);
+extern bool
```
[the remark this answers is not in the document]
vector-of-bool: Nope

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -1701,26 +1653,24 @@ _mongoc_topology_push_server_session (mongoc_topology_t *topology,
...
-      _mongoc_server_session_destroy (ss);
-
-      if (++i == 10000) {
+   for (; i < 10000; ++i) {
+      char buf[16];
+      const char *key;
+      bool did_pop;
+      mongoc_server_session_t *ss =
+         mongoc_ts_pool_try_pop (topology->session_pool, &did_pop);
+      if (!did_pop) {
```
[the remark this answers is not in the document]
vector-of-bool: The `did_pop` parameter is redundant, so it's going away (See other comment). This can also be cleaned up somewhat.

src/libmongoc/src/mongoc/mongoc-topology.c
vector-of-bool: It would not entirely negate the performance improvement, since the pop operations are very fast and multiple threads can simultaneously pop.

However, I question the need for that to be in the spec, as it is more of an implementation concern than a behavioral change.

I presume it was specified that way to prevent the pool from growing with expired sessions ad infinitum, but I don't think the design of the pool as a stack would ever allow such unbounded growth.

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -216,6 +216,53 @@ _mongoc_topology_scanner_cb (uint32_t id,
...
+      return true;
+   }
+
+   /** If the session has never been used, it should be dropped */
+   if (session->last_used_usec == SESSION_NEVER_USED) {
+      return true;
+   }
+
+   /* Check for a timeout */
+   bson_mutex_lock (&topo->mutex);
```
aggrand, line 252: Just for my own understanding, this was the mutex in question that had all the contention, right?

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -216,6 +216,53 @@ _mongoc_topology_scanner_cb (uint32_t id,
...
+
+static void
+_server_session_destroy (mongoc_server_session_t *session,
+                         mongoc_topology_t *unused)
+{
+   _mongoc_server_session_destroy (session);
+}
+
+static int
+_server_session_should_prune (mongoc_server_session_t *session,
```
aggrand, line 235: If I read correctly, this is only being called on a particular session while retrieving it, right? If so, is there a risk of expired sessions hanging around indefinitely if they just happen to not get touched? If so, should there be a backstop to occasionally still go through and remove the expired threads?

Super cool if there's no sane place or time to do so; I know the point is to reduce contention on the below mutex anyway so an even occasional search-through might not work.

src/libmongoc/src/mongoc/mongoc-ts-pool.c
```diff
@@ -0,0 +1,263 @@
...
+{
+   pool_node *node;
+   {
+      bson_mutex_lock (&pool->mtx);
+      node = pool->head;
+      pool->head = NULL;
+      pool->size = 0;
+      bson_mutex_unlock (&pool->mtx);
+   }
+   while (node) {
```
aggrand, line 151: Would it be possible (or reasonable) to redefine this in terms of `visit_each`?

aggrand (commented):
I just had a few questions, but this is looking great! I'm excited for these changes!

src/libmongoc/src/mongoc/mongoc-ts-pool.c
vector-of-bool: Possible: yes, but not desirable, since visit_each holds a lock while it visits every element. This quickly clears the pool and then unlocks it before it starts deleting elements.

src/libmongoc/src/mongoc/mongoc-topology.c
vector-of-bool: This was discussed, yes. And indeed, a expired sessions can loom indefinitely in a pathological case. We also figure that the pool itself can never grow indefinitely, since creating a new element requires that all stale elements be deleted first. It *may* be the case that:

1. A burst of 500 simultaneous operations generate 500 session objects.
2. Those sessions are returned to the pool.
3. For an indefinite amount of time, only one session is used at a time, and the other 499 sit unused.
4. The bottom 499 sessions expire, but are not deallocated, as the top session is the only one ever inspected.
5. If at any attempt is made to use *two* sessions simultaneously, it will immediately prune the 499 sessions and the pool will be cleared.

This may have tradeoffs, but because of the LIFO order, the moment a single expired session is popped from the top, the entire pool will be cleared. We discussed in another comment on how to reconcile this with the spec that mandates the "bottom" of the pool be checked every time a session is returned. If we want to go this route, the pool will need to be tweaked into a deque, and that's very possible in the future as an upcoming change will require a thread-safe deque anyway.

src/libmongoc/src/mongoc/mongoc-topology.c
vector-of-bool: Yes, this is the troublesome mutex. Some of the usages (such as this one being commented on) can be replaced with atomic-loads, which will further alleviate pressure on this lock.

eramongodb (changes requested):

src/libmongoc/src/mongoc/mongoc-ts-pool.c
```diff
@@ -0,0 +1,263 @@
...
+   long double maxalign_2
+      __attribute__ ((__aligned__ (__alignof__(long double))));
+} _max_align_type;
+#endif
+
+typedef struct pool_node {
+   struct pool_node *next;
+   mongoc_ts_pool *owner_pool;
+
+   _max_align_type data[0];
```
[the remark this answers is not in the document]
eramongodb: Rather than emulating flexible array members, wouldn't simply using them be a viable alternative, given the C driver is already making use of C99 features?
```suggestion
   _max_align_type data[];
```
GCC's documentation for [Arrays of Length Zero](https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html) also recommends using C99 flexible array members instead of the extension:

> The preferred mechanism to declare variable-length types like `struct line` above is the ISO C99 _flexible array member_, with slightly different syntax and semantics: [...]

This also applies to `mongoc_ts_pool::outstanding_items`.

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -709,34 +709,26 @@ _mongoc_client_session_handle_reply (mongoc_client_session_t *session,
...
-   s = bson_malloc0 (sizeof (mongoc_server_session_t));
-   s->last_used_usec = SESSION_NEVER_USED;
-   s->prev = NULL;
-   s->next = NULL;
-   bson_init (&s->lsid);
-   bson_append_binary (
-      &s->lsid, "id", 2, BSON_SUBTYPE_UUID, uuid_data, sizeof uuid_data);
-
+   self->last_used_usec = SESSION_NEVER_USED;
+   bson_init (&self->lsid);
```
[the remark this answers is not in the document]
kevinAlbs: I am not sure about the cause of the reported ASAN leaks. I attempted to reproduce this with clang 11 on my mac.

I configured with the `-fsanitize=leaks`. Here is my full cmake command:
```
cmake \
    -G "Ninja" \
    -DCMAKE_BUILD_TYPE="Debug" \
    -DCMAKE_C_COMPILER="/Users/kevin.albertson/bin/llvm-11.0.0/bin/clang" \
    -DCMAKE_C_COMPILER_LAUNCHER="ccache" \
    -DENABLE_MAINTAINER_FLAGS=ON \
    -DBUILD_VERSION=1.18.0-pre \
    -DENABLE_EXTRA_ALIGNMENT=OFF \
    -DENABLE_AUTOMATIC_INIT_AND_CLEANUP=OFF \
    -DCMAKE_C_FLAGS="-fsanitize=address -fsanitize=leak -DBSON_MEMCHECK -fno-omit-frame-pointer" \
    -DCMAKE_VERBOSE_MAKEFILE=ON \
    -DCMAKE_INSTALL_PREFIX="/Users/kevin.albertson/install/mongo-c-driver-1.18.0-pre" \
    -DCMAKE_PREFIX_PATH="/Users/kevin.albertson/install/libmongocrypt" \
    -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
    -DENABLE_CLIENT_SIDE_ENCRYPTION=ON \
    -DENABLE_SSL=DARWIN \
    -DOPENSSL_ROOT_DIR="/usr/local/opt/openssl@1.1" \
    -DENABLE_TESTING=ON \
    -DENABLE_DEBUG_ASSERTIONS=ON \
    -S$(pwd) -B$(pwd)/cmake-build
```

The stack traces in the Evergreen logs included `background_mongoc_collection_count_documents`. That is only called in a few tests via `future_collection_count_documents`. I believe the following runs all of them:

```
ASAN_OPTIONS='detect_leaks=1' ./cmake-build/src/libmongoc/test-libmongoc --match "/inheritance/count*" --match "/Collection/count_documents"
```

But, no reported leaks.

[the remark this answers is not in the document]
vector-of-bool: I have been having trouble with ASan and LeakSanitizer repeatedly across all branches, and I am unsure if it is a bug in ASan, the compiler, or if there are actual issues in our codebase. It is frustrating as it renders ASan entirely useless on my system. I will need to do more investigation in the future.

src/libmongoc/src/mongoc/mongoc-ts-pool-private.h
```diff
@@ -1,97 +1,310 @@
...
  */
-extern void
-mongoc_ts_pool_free_item (mongoc_ts_pool pool, void *item);
+#define MONGOC_DECL_SPECIAL_TS_POOL(ElementType,                              \
+                                    PoolName,                                 \
+                                    UserDataType,                             \
+                                    Constructor,                              \
+                                    Destructor,                               \
+                                    PrunePredicate)                           \
+   typedef struct PoolName {                                                  \
```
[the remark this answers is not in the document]
vector-of-bool: Removing the tag name on the struct makes it invalid to use as an explicit tagged name. This isn't a huge problem, but is very confusing when for `typedef struct {} foo;`, then `foo` and `struct foo` no longer name the same type. Alternatively, it may be useful to remove the `typedef` and allow the macro-user to declare the `typedef` themselves.

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -2290,7 +2318,7 @@ test_cursor_implicit_session (void *ctx)
 
    /* lsid returned after last batch, doesn't wait for mongoc_cursor_destroy */
    check_session_returned (test, &find_lsid);
-   ASSERT_POOL_SIZE (topology, 2);
+   BSON_ASSERT (mongoc_server_session_pool_size (topology->session_pool) == 2);
```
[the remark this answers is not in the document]
vector-of-bool: I missed this one.

src/libmongoc/src/mongoc/mongoc-ts-pool.c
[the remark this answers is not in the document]
vector-of-bool: In this case, using an FAM is entirely reasonable, given that we already have some C99 peppered throughout.

For `outstanding_items`, I'd be sad to change it, because having that member be present/absent based on a `1/0` enum that can also be used in `if ()` conditions is so elegant (when compared to using an `#ifdef`) 😢. Alternatively, just bumping the size of of the `malloc()` based on `AUDIT_POOL_ENABLED` isn't a huge burden either 🙂. Or even just foregoing that trickery and letting the member be unused unless `AUDIT_POOL_ENABLED != 0`. (This is interesting enough that I might do a write-up on it.)

vector-of-bool:
Looks like there is a warning-as-error for `-Waggregate-return`. Some digging suggests this is an archaic warning. It's listed in `maintainer-flags.txt`, but I don't know why it was added.

src/libmongoc/src/mongoc/mongoc-ts-pool.c
```diff
@@ -1,117 +1,263 @@
...
-   bson_free (node);
+   /* Pointer to the pointer that must be updated in case of an item pruning */
+   pool_node **node_ptrptr;
+   /* The node we are looking at */
+   pool_node *node;
+   bson_mutex_lock (&pool->mtx);
+   node_ptrptr = &pool->head;
+   node = pool->head;
+   while (node) {
+      const bool should_remove =
```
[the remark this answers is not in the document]
vector-of-bool: I think this is more a stylistic question than anything. I prefer to `const` everything absolutely possible, but others may not.

src/libmongoc/src/mongoc/mongoc-ts-pool.c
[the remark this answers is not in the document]
vector-of-bool: I just removed the cleverness with `outstanding_items`. It's only a single integer, so not worth trying to be tricky.

vector-of-bool:
I cannot find any compelling reason to keep `-Waggregate-return`, so I simply removed it. Are there any objections?

src/libmongoc/src/mongoc/mongoc-ts-pool.c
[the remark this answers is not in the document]
kevinAlbs: In that case I am cool with it. I thought it was unintentional.

src/libmongoc/src/mongoc/mongoc-ts-pool-private.h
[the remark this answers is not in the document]
kevinAlbs: Ah I mistakenly assumed this was unintentional since `struct PoolName` was not referred to. Resolving.

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -709,34 +709,26 @@ _mongoc_client_session_handle_reply (mongoc_client_session_t *session,
...
    /* transaction number is a positive integer and will be incremented before
     * each use, so ensure it is initialized to zero. */
-   s->txn_number = 0;
+   self->txn_number = 0;
 
-   RETURN (s);
-}
+   if (!_mongoc_server_session_uuid (uuid_data, error)) {
```
kevinAlbs, line 724:
```suggestion
   if (!_mongoc_server_session_uuid (uuid_data, error)) {
      bson_destroy (&self->lsid);
```

src/libmongoc/src/mongoc/mongoc-client-session.c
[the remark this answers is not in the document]
kevinAlbs: I realized the `debug-compile-asan-clang` configures with `-DENABLE_SSL=OFF`. That disables support for sessions, so the call to `_mongoc_server_session_uuid` below fails.

I was able to reproduce this locally by configuring with:
```
cmake \
    -G "Ninja" \
    -DCMAKE_BUILD_TYPE="Debug" \
    -DCMAKE_C_COMPILER="/Users/kevin.albertson/bin/llvm-11.0.0/bin/clang" \
    -DCMAKE_C_COMPILER_LAUNCHER="ccache" \
    -DENABLE_MAINTAINER_FLAGS=ON \
    -DCMAKE_C_FLAGS="-fsanitize=address -fsanitize=leak -DBSON_MEMCHECK -fno-omit-frame-pointer" \
    -DENABLE_SSL=OFF \
    -S$(pwd) -B$(pwd)/cmake-build-847
```

And running with:
```
ASAN_OPTIONS='detect_leaks=1' ./cmake-build-847/src/libmongoc/test-libmongoc --match "/inheritance/count*" --match "/Collection/count_documents"
```

I suggested a fix below.

src/libmongoc/tests/test-mongoc-ts-pool.c
```diff
@@ -0,0 +1,109 @@
...
+static void
+_set_int_to_7 (int *v, void *unused, bson_error_t *unused2)
+{
+   *v = 7;
+}
+
+/* Declare a pool that contains `int`, sets each new int to seven, and drops
+ * integers that are equal to 42. */
+MONGOC_DECL_SPECIAL_TS_POOL (
+   int, int_pool, void, _set_int_to_7, NULL, _is_int_42)
```
kevinAlbs, line 61: Compiling locally with maintainer flags produces warnings for unused functions:

```
/Users/kevin.albertson/code/mongo-c-driver/src/libmongoc/tests/test-mongoc-ts-pool.c:60:1: warning: unused function 'int_pool_clear' [-Wunused-function]
MONGOC_DECL_SPECIAL_TS_POOL (
^
/Users/kevin.albertson/code/mongo-c-driver/src/libmongoc/src/mongoc/mongoc-ts-pool-private.h:286:28: note: expanded from macro 'MONGOC_DECL_SPECIAL_TS_POOL'
   static BSON_INLINE void PoolName##_clear (PoolName p)                      \
                           ^
<scratch space>:41:1: note: expanded from here
int_pool_clear
^
/Users/kevin.albertson/code/mongo-c-driver/src/libmongoc/tests/test-mongoc-ts-pool.c:60:1: warning: unused function 'int_pool_drop' [-Wunused-function]
/Users/kevin.albertson/code/mongo-c-driver/src/libmongoc/src/mongoc/mongoc-ts-pool-private.h:307:28: note: expanded from macro 'MONGOC_DECL_SPECIAL_TS_POOL'
   static BSON_INLINE void PoolName##_drop (ElementType *elem)                \
                           ^
<scratch space>:45:1: note: expanded from here
int_pool_drop
^
/Users/kevin.albertson/code/mongo-c-driver/src/libmongoc/tests/test-mongoc-ts-pool.c:60:1: warning: unused function 'int_pool_is_empty' [-Wunused-function]
/Users/kevin.albertson/code/mongo-c-driver/src/libmongoc/src/mongoc/mongoc-ts-pool-private.h:317:27: note: expanded from macro 'MONGOC_DECL_SPECIAL_TS_POOL'
   static BSON_INLINE int PoolName##_is_empty (PoolName p)                    \
                          ^
<scratch space>:47:1: note: expanded from here
int_pool_is_empty
^
/Users/kevin.albertson/code/mongo-c-driver/src/libmongoc/tests/test-mongoc-ts-pool.c:60:1: warning: unused function 'int_pool_visit_each' [-Wunused-function]
/Users/kevin.albertson/code/mongo-c-driver/src/libmongoc/src/mongoc/mongoc-ts-pool-private.h:322:28: note: expanded from macro 'MONGOC_DECL_SPECIAL_TS_POOL'
   static BSON_INLINE void PoolName##_visit_each (                            \
                           ^
<scratch space>:48:1: note: expanded from here
int_pool_visit_each
^
4 warnings generated.
```

Can you add casts for each function, like `(void) int_pool_clear;` to resolve these warnings?

src/libmongoc/src/mongoc/mongoc-ts-pool.c
```diff
@@ -0,0 +1,282 @@
...
+ * @brief Create a new pool node and contained element.
+ *
+ * @return pool_node* A pointer to a constructed element, or NULL if the
+ * constructor sets `error`
+ */
+static pool_node *
+_new_item (mongoc_ts_pool *pool, bson_error_t *error)
+{
+   pool_node *node =
+      bson_malloc0 (sizeof (pool_node) + pool->params.element_size);
```
eramongodb, line 65: I want to confirm: is this allocation size correct given the flexible data member in `pool_node` does not contribute to `sizeof (pool_node)` (does `pool->params.element_size` correctly account for possible access of `node->data`)?

src/libmongoc/src/mongoc/mongoc-client-session.c
vector-of-bool: Good find.

src/libmongoc/tests/test-mongoc-ts-pool.c
vector-of-bool: I believed that adding `BSON_INLINE` to these functions would erase these warnings. I'll have to dig deeper.

src/libmongoc/src/mongoc/mongoc-ts-pool.c
vector-of-bool: That is correct, yes.

src/libmongoc/tests/test-mongoc-ts-pool.c
vector-of-bool: Should now be fixed.

vector-of-bool:
I believe that everything should be happy now, sans flaky test failures.

src/libmongoc/tests/test-mongoc-ts-pool.c
```diff
@@ -0,0 +1,97 @@
...
+
+   mongoc_ts_pool_free (pool);
+}
+
+static int
+_is_int_42 (int *v, void *unused)
+{
+   (void) unused;
+   return *v == 42;
+}
```
eramongodb, line 51:
```suggestion
static int
_is_int_42 (int *v, BSON_MAYBE_UNUSED void *unused)
{
   return *v == 42;
}
```
Is silencing `-Wunused-parameter` warnings even necessary given there are currently +600 instances of said warning throughout the C driver? (I suppose no time like the present to start reducing them.)

vector-of-bool: I have a stashed changeset that marks every single unused parameter in the codebase. The "maybe unused" attribute doesn't help on MSVC, but my stashed changes will fix for MSVC as well. A good thing for a new PR, but I don't want to pile on _another_ PR on top of my four open ones already 🙂.

eramongodb: Fair enough! Will consider further modifications regarding unused warnings optional.

kevinAlbs (approved):
Fantastic work, LGTM!

eramongodb (commented):
Just some minor feedback and questions, but overall changes LGTM!

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -1570,45 +1588,29 @@ _mongoc_topology_pop_server_session (mongoc_topology_t *topology,
...
    }
-
    bson_mutex_unlock (&topology->mutex);
 
-   if (!ss) {
-      ss = _mongoc_server_session_new (error);
-   }
+   ss = mongoc_server_session_pool_get (topology->session_pool, error);
 
+done:
```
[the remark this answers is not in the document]
vector-of-bool: Yeah, that came from a tentative change that was then undone.

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -709,34 +709,27 @@ _mongoc_client_session_handle_reply (mongoc_client_session_t *session,
...
-_mongoc_server_session_new (bson_error_t *error)
+bool
+_mongoc_server_session_init (mongoc_server_session_t *self, bson_error_t *error)
 {
    uint8_t uuid_data[16];
-   mongoc_server_session_t *s;
-
    ENTRY;
 
-   if (!_mongoc_server_session_uuid (uuid_data, error)) {
```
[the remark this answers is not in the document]
vector-of-bool: Good question. The answer is: I don't remember. I'll rearrange this.

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -766,14 +759,9 @@ _mongoc_server_session_timed_out (const mongoc_server_session_t *server_session,
...
 void
-_mongoc_server_session_destroy (mongoc_server_session_t *server_session)
+_mongoc_server_session_destroy (mongoc_server_session_t *self)
 {
-   ENTRY;
-
-   bson_destroy (&server_session->lsid);
-   bson_free (server_session);
-
-   EXIT;
```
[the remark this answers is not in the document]
vector-of-bool: Not particularly, other than that the function was reduced to a single line.

eramongodb (approved):

PR #848, opened 2021-08-13
CDRIVER-4093 fix versioned API sample tests

kevinAlbs:
Addresses test failures introduced in #843 to the versioned API tests and 5.0 sharded tests.

### DIFF
--- a/src/libmongoc/tests/test-mongoc-sample-commands.c
+++ b/src/libmongoc/tests/test-mongoc-sample-commands.c
@@ -3826,7 +3826,6 @@ static void _test_sample_versioned_api_example_5_6_7_8 (void) {
       323,
       "Provided apiStrict:true, but the command count is not in API Version 1");
    ASSERT_CMPINT64 (-1, ==, count);
-   bson_destroy (&reply);
 #if 0
    /* This block not evaluated, but is inserted into documentation to represent the above reply.
     * Don't delete me! */
--- a/src/libmongoc/tests/test-mongoc-sample-commands.c
+++ b/src/libmongoc/tests/test-mongoc-sample-commands.c
@@ -3747,7 +3747,6 @@ static void _test_sample_versioned_api_example_5_6_7_8 (void) {
    bson_t *filter;
 
    client = get_client_for_version_api_example ();
-   test_framework_monitor_commands (client);
    mongoc_client_set_error_api (client, MONGOC_ERROR_API_VERSION_2);
    mongoc_server_api_version_from_string ("1", &server_api_version);
    server_api = mongoc_server_api_new (server_api_version);
--- a/src/libmongoc/tests/test-mongoc-sample-commands.c
+++ b/src/libmongoc/tests/test-mongoc-sample-commands.c
@@ -3743,6 +3743,7 @@ static void _test_sample_versioned_api_example_5_6_7_8 (void) {
    bson_t *docs[N_DOCS];
    int i;
    bson_t reply;
+   bson_t *cmd;
    int64_t count;
    bson_t *filter;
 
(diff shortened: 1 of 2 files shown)


src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -3592,12 +3607,13 @@ _test_sample_versioned_api_example_1 (void)
     * client = mongoc_client_new (uri_sharded);
     */
 
-   client = get_client ();
+   client = get_client_for_version_api_example ();
```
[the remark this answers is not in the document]
chardan: Is it worth adding a comment in the code about the API options note?

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -3435,6 +3435,20 @@ get_client (void)
...
 
+/* Returns a test client without version API options configured. */
+static mongoc_client_t *
+get_client_for_version_api_example (void) {
+   mongoc_client_t *client;
+   mongoc_uri_t *uri;
+
+   uri = test_framework_get_uri ();
+   client = mongoc_client_new_from_uri (uri);
+   ASSERT (client);
```
chardan, line 3446: I'm assuming that since this is test code that we expect to just terminate, there's no need to release the "client" resource?

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -3690,13 +3711,14 @@ _test_sample_versioned_api_example_4 (void)
...
 
-   client = get_client ();
+   client = get_client_for_version_api_example ();
 
    mongoc_server_api_version_from_string ("1", &server_api_version);
    server_api = mongoc_server_api_new (server_api_version);
    mongoc_server_api_deprecation_errors (server_api, true);
 
-   mongoc_client_set_server_api (client, server_api, &error);
+   ok = mongoc_client_set_server_api (client, server_api, &error);
```
chardan, line 3720: nit: I'm ok with "ok", but it might be okay-er to just check the call with BSON_ASSERT directly, if this is the only place we're using it. But, really, I'm ok either way!

chardan (approved):
+1, a couple of things you might peek at but I think this looks good!

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -3592,12 +3607,13 @@ _test_sample_versioned_api_example_1 (void)
...
 
-   client = get_client ();
+   client = get_client_for_version_api_example ();
 
    mongoc_server_api_version_from_string ("1", &server_api_version);
    server_api = mongoc_server_api_new (server_api_version);
 
-   mongoc_client_set_server_api (client, server_api, &error);
+   ok = mongoc_client_set_server_api (client, server_api, &error);
+   BSON_ASSERT (ok);
```
vector-of-bool, line 3616: Do we care much that `BSON_ASSERT` is appearing in the API examples (as opposed to the regular C `assert()`)?

vector-of-bool (approved):

src/libmongoc/tests/test-mongoc-sample-commands.c
kevinAlbs: Callers of `get_client_for_version_api_example` are expected to call `mongoc_client_destroy` to release the `client`.

src/libmongoc/tests/test-mongoc-sample-commands.c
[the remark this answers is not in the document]
kevinAlbs: Good idea, I added a note to say that `get_client_for_version_api_example` returns a client without server API options configured.

src/libmongoc/tests/test-mongoc-sample-commands.c
kevinAlbs: Good question. Though `BSON_ASSERT` is part of the public libbson API, and is defined in bson-macros.h, I think it makes sense to use `assert` since readers are likely to be familiar with `assert`.

I looked through other examples code for reference. There aren't many assertions, but there are instances of both uses of `BSON_ASSERT` and `assert`.

https://github.com/mongodb/mongo-c-driver/blob/3a41062e397638399a6e491ec94bd6c861ec212c/src/libmongoc/examples/example-update.c#L52:L52 uses `BSON_ASSERT`.

https://github.com/mongodb/mongo-c-driver/blob/3a41062e397638399a6e491ec94bd6c861ec212c/src/libmongoc/examples/example-gridfs.c#L56:L56 uses `assert`.

src/libmongoc/tests/test-mongoc-sample-commands.c
kevinAlbs: Done, here and in the other changes!

PR #849, opened 2021-08-16
CDRIVER-4074 - Update documentation for Time Series

chardan:
Adds a note with a link to Time Series Collections from database
creation.

It may not be immediately clear to users that time series collections
can be created through mongoc_database_create_collection().


kevinAlbs (approved):
LGTM

PR #850, opened 2021-08-16
CDRIVER-3632 Address pedantic warnings for function pointer conversions

eramongodb:
Conversion between object pointers and function pointers is undefined behavior, though permitted as a "common extension" by the C standard (see [N1570 J.5.7 "Function pointer casts"](http://port70.net/~nsz/c/c11/n1570.html#J.5.7)).

This commit defines a new TestSuite_AddFullWithTestFn macro as a convenient utility that defines an intermediate TestFuncCtx object through which a given function pointer is passed via the ctx parameter without incurring function pointer <-> object pointer conversions.

---

Converting between object pointers and function pointers is undefined behavior per [N1570](http://port70.net/~nsz/c/c11/n1570.html) (C11 standard draft). It is however specified as a ["common extension"](http://port70.net/~nsz/c/c11/n1570.html#J.5). As also noted by [POSIX.1-2008](https://pubs.opengroup.org/onlinepubs/9699919799/functions/dlsym.html):

> Note that conversion from a `void *` pointer to a function pointer as in:
> ```c
>     fptr = (int (*)(int))dlsym(handle, "my_function");
> ```
> is not defined by the ISO C standard.

This PR introduces a new `TestSuite_AddFullWithTestFn` macro as a conforming workaround. This utility stores the provided test function pointer `_test_fn` in an intermediate statically-allocated `TestFnCtx` object. A pointer to this intermediate object is then passed as the `ctx` parameter to `_TestSuite_AddFull` instead of the original test function pointer:
```c
{
   static TestFnCtx ctx; // Note: variable is scoped to avoid local conflicts.
   ctx.test_fn = _test_fn;
   _TestSuite_AddFull (_suite, _name, _func, _dtor, &ctx, __VA_ARGS__, NULL);
}
```

An alternative solution that was considered is dynamically allocating the `TestFuncCtx` objects instead:
```c
TestFnCtx *ctx = malloc (sizeof (TestFnCtx));

ctx->test_fn = (TestFunc) (_test_fn);
ctx->dtor = _dtor;

_TestSuite_AddFull (_suite, _name, _func, _TestSuite_TestFnCtxDtor, ctx, __VA_ARGS__, NULL);
```
This would necessitate an additional data member and function to correctly invoke the user-provided `_dtor` function when provided and free the `TestFnCtx` object:
```c
void _TestSuite_TestFnCtxDtor (void *ctx) {
   TestFuncDtor dtor = ((TestFnCtx *) ctx)->dtor;
   if (dtor) { dtor (ctx); }
   free (ctx);
}
```

The binary size of `test-libmongoc` when built with the static solution vs. dynamic solution does not seem to make a compelling argument to favor one solution over the other. Some numbers when building on WSL Ubuntu 20.04 x86_64 using CMake-provided (version 3.16.3) default compiler arguments for Debug and Release mode:

| Size (Bytes) | Debug (GCC 9) | Release (GCC 9) | Debug (Clang 10) | Release (Clang 10) |
| --- | --- | --- | --- | --- |
| Static | 8,666,040 | 3,585,632 | 8,063,480 | 3,253,640 |
| Dynamic | 8,666,680 | 3,586,520 | 8,062,272 | 3,252,328 |

Therefore I am in favor of the simpler "static" solution as currently proposed over the alternative "dynamic" solution.


rcsanchez97 (approved):
LGTM!

Thank for figuring out how to address the last few warnings that I didn't manage to get to.  I appreciate learning this interesting bit about how function pointers work.

src/libmongoc/tests/TestSuite.h
```diff
@@ -710,6 +716,14 @@ _TestSuite_AddFull (TestSuite *suite,
                     ...);
 #define TestSuite_AddFull(_suite, _name, _func, _dtor, _ctx, ...) \
    _TestSuite_AddFull (_suite, _name, _func, _dtor, _ctx, __VA_ARGS__, NULL)
+#define TestSuite_AddFullWithTestFn(                            \
```
kevinAlbs, line 719: Nit: Would `TestSuite_AddFullWithCtxFn` or `TestSuite_AddFullWithArgFn` be a more apt name?

I interpret `_func` as the test function. And `_test_fn` is the context argument passed to `_func`.

kevinAlbs (approved):
LGTM. I have a nit question about the name `TestSuite_AddFullWithTestFn`, but nothing to block merging. If you have a counter proposal, feel free to re-request review.

This makes reading the output of compiles much easier!

> Converting between object pointers and function pointers is undefined behavior per [N1570](http://port70.net/~nsz/c/c11/n1570.html) (C11 standard draft). It is however specified as a ["common extension"](http://port70.net/~nsz/c/c11/n1570.html#J.5). As also noted by [POSIX.1-2008](https://pubs.opengroup.org/onlinepubs/9699919799/functions/dlsym.html):
> 
> > Note that conversion from a `void *` pointer to a function pointer as in:
> > ```c
> >     fptr = (int (*)(int))dlsym(handle, "my_function");
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is not defined by the ISO C standard.

Thank you for the references. I had not realized use of `dlsym` requires an undefined cast.
 
> The binary size of `test-libmongoc` when built with the static solution vs. dynamic solution does not seem to make a compelling argument to favor one solution over the other. Some numbers when building on WSL Ubuntu 20.04 x86_64 using CMake-provided (version 3.16.3) default compiler arguments for Debug and Release mode:
> 
> Size (Bytes)	Debug (GCC 9)	Release (GCC 9)	Debug (Clang 10)	Release (Clang 10)
> Static	8,666,040	3,585,632	8,063,480	3,253,640
> Dynamic	8,666,680	3,586,520	8,062,272	3,252,328
> Therefore I am in favor of the simpler "static" solution as currently proposed over the alternative "dynamic" solution.

Oh very cool! Thank you for the thorough investigation. For some context, AFAIK the binary size of `test-libmongoc` (or `libbson` / `libmongoc`) has not been a concern when evaluating changes. I see value in tracking the size of `libbson` / `libmongoc` since it would directly impact users. I have not seen user reports about increase of binary size.

kevinAlbs (changes requested):
Oops, I spoke too soon before checking the attached Evergreen tasks. Can you investigate the ASAN related failures. They appear related to these changes: https://spruce.mongodb.com/task/mongo_c_driver_asan_ubuntu_test_asan_latest_sharded_auth_nosasl_openssl_patch_d2838038aa7d1808b872ea7bd39df33db5bba364_611adc29a4cf4779ee240418_21_08_16_21_44_11/logs?execution=0

src/libmongoc/tests/TestSuite.h
eramongodb: I do not think so.

`WithArgFn` suggests an "argument function" is being provided, which I do not think is a meaningful/useful description ("Argument to what?" "A function that provides an argument?").

`WithCtxFn` suggests the argument being provided is a "context function", but the argument provided _is_ the context ("context" referring to the argument given to a test), not a function to obtain one.

`WithTestFn` was selected because `test_fn` is [the name of the parameter](https://github.com/mongodb/mongo-c-driver/blob/96ed1d70210ec5096aaeab0adbb5ae4323291f20/src/libmongoc/tests/test-mongoc-client-session.c#L1452) used by relevant tests that expect a function pointer as an argument, paralleling how parameters are [usually named `ctx`](https://github.com/mongodb/mongo-c-driver/blob/96ed1d70210ec5096aaeab0adbb5ae4323291f20/src/libmongoc/tests/test-mongoc-client-session.c#L76) by other tests. For full symmetry, I would also rename `TestSuite_AddFull` to `TestSuite_AddFullWithCtx` rather than being implicit (as an example, not as a suggestion).

eramongodb:
Preliminary results of [Evergreen tests](https://spruce.mongodb.com/version/611d62e8d6d80a2095a0dc56/) seem to suggest the relevant ASan test failures have been addressed. The failures were identified as due to [a missing cast](https://github.com/mongodb/mongo-c-driver/pull/850/commits/ab2adc1caefbe72685766ab21ec1d260e3d0c1bb) in a test now registered with `TestSuite_AddFullWithTestFn`.

kevinAlbs (approved):
LGTM! Offline we looked through test failures that and believe that these are not introduced by changes in this PR related to CDRIVER-4111 and CDRIVER-4134.

@eramongodb is rerunning tests that are believed to be flaky; if the result of that uncovers anything that requires additional changes, feel free to re-request review from me.

eramongodb:
Rerun of failed tests appears to have flushed out unexpected test failures relative to base commit, confirming they were indeed flaky.

src/libmongoc/tests/TestSuite.c
```diff
@@ -263,12 +251,8 @@ TestSuite_Add (TestSuite *suite, /* IN */
...
                TestFunc func)    /* IN */
 {
-   TestSuite_AddFull (suite,
-                      name,
-                      TestSuite_AddHelper,
-                      NULL,
-                      (void *) func,
-                      TestSuite_CheckDummy);
+   TestSuite_AddFullWithTestFn (
+      suite, name, TestSuite_AddHelper, NULL, func, TestSuite_CheckDummy);
```
vector-of-bool, line 255: This and the below call within `TestSuite_AddLive` are broken. The `static` test context instance is only instantiated within the scope of this function, so each test that calls `TestSuite_Add()` will shared the static instance, causing only the final caller to actually have its test function registered. Each test that was defined using this function will instead that final test function instead of their own test function.

It may be enough to change `TestSuite_Add` and `TestSuite_AddLive` to also be macros, but that might just be kicking the can further down the road.

eramongodb: Good spot! I had a feeling opting for this "simpler" approach might come back to bite me. It seems the principle of keeping scope (and lifetime) as limited to necessary scope as possible proves itself again here.

Thankfully I already explored and documented an alternative solution that does exactly that, which should directly address this issue you identified. I will document this issue on JIRA and post a corresponding PR shortly.

eramongodb: PR posted: https://github.com/mongodb/mongo-c-driver/pull/865

PR #852, opened 2021-08-25
Fix CDRIVER3441 - document mongoc_handshake_data_append and alias as …

chardan:
Fix CDRIVER3441 - document mongoc_handshake_data_append and alias as mongoc_add_driver_info

I took a slightly different tack on this and made the older form an alias to the preferred,
newer one.

https://jira.mongodb.org/browse/CDRIVER-3441

### DIFF
--- a/src/libmongoc/doc/mongoc_handshake.rst
+++ b/src/libmongoc/doc/mongoc_handshake.rst
@@ -1,0 +1,89 @@
+:man_page: mongoc_add_driver_info
+
+mongoc_add_driver_info()
+=========================================
+
+Synopsis
+--------
+
+.. code-block:: c
+
+   bool
+   mongoc_add_driver_info (const char *driver_name,
+                           const char *driver_version,
+                           const char *platform);
+
+   bool
+   mongoc_handshake_data_append (const char *driver_name,
+                                 const char *driver_version,
+                                 const char *platform);
+
+Appends the given strings to the handshake data for the underlying C Driver.
+
+Description
+-----------
+
+This function is intended for use by drivers which wrap the C Driver.
+Calling this function will store the given strings as handshake data about
+the system and driver by appending them to the handshake data for the
+underlying C Driver. These values, along with other handshake data collected
+during mongoc_init will be sent to the server as part of the initial
+connection handshake in the "client" document. This function must not be
+called more than once, or after a handshake has been initiated.
+
+The passed in strings are copied, and don't have to remain valid after the
+call to mongoc_add_driver_info(). The driver may store truncated
+versions of the passed in strings.
+
+Note:
+  This function allocates memory, and therefore caution should be used when
+  using this in conjunction with bson_mem_set_vtable. If this function is
+  called before bson_mem_set_vtable, then bson_mem_restore_vtable must be
+  called before calling mongoc_cleanup. Failure to do so will result in
+  memory being freed by the wrong allocator.
+
+Note:
+  mongoc_handshake_data_append() is the previously undocumented, now-deprecated 
+  name for this function. It calls mongoc_add_driver_info().
+
+Parameters
+----------
+
+* ``driver_name``: An optional string storing the name of the wrapping driver
+* ``driver_version``: An optional string storing the version of the wrapping driver.
+* ``platform``: An optional string storing any information about the current platform, for example configure options or compile flags.
+
+Returns
+-------
+
+``true`` if the given fields are set successfully. Otherwise, it returns ``false`` and logs an error.
+
+The default handshake data the driver sends with "hello" looks something
+like:
+
+.. code-block:: c
+
+ client: {
+   driver: {
+     name: "mongoc",
+     version: "1.5.0"
+   },
+   os: {...},
+   platform: "CC=gcc CFLAGS=-Wall -pedantic"
+ }
+
+If we call
+  mongoc_add_driver_info ("phongo", "1.1.8", "CC=clang")
+and it returns true, the driver sends handshake data like:
+
+.. code-block::
+ client: {
+   driver: {
+     name: "mongoc / phongo",
+     version: "1.5.0 / 1.1.8"
+   },
+   os: {...},
+   platform: "CC=gcc CFLAGS=-Wall -pedantic / CC=clang"
+ }
+
+
--- a/src/libmongoc/doc/mongoc_handshake.rst
+++ b/src/libmongoc/doc/mongoc_handshake.rst
@@ -37,10 +37,6 @@ Note:
   called before calling mongoc_cleanup. Failure to do so will result in
   memory being freed by the wrong allocator.
 
-Note:
-  mongoc_handshake_data_append() was previously previously undocumented, 
-  and is now deprecated.
-
 Parameters
 ----------
 
--- a/src/libmongoc/doc/mongoc_handshake.rst
+++ b/src/libmongoc/doc/mongoc_handshake.rst
@@ -1,17 +1,12 @@
-:man_page: mongoc_add_driver_info
+:man_page: mongoc_handshake_data_append
 
-mongoc_add_driver_info()
+mongoc_handshake_data_append()
 =========================================
 
 Synopsis
 --------
 
 .. code-block:: c
-
-   bool
-   mongoc_add_driver_info (const char *driver_name,
-                           const char *driver_version,
-                           const char *platform);
 
    bool
    mongoc_handshake_data_append (const char *driver_name,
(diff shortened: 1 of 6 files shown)


src/libmongoc/doc/mongoc_handshake.rst
```diff
@@ -0,0 +1,89 @@
...
+mongoc_add_driver_info()
+=========================================
+
+Synopsis
+--------
+
+.. code-block:: c
+
+   bool
+   mongoc_add_driver_info (const char *driver_name,
```
kevinAlbs, line 12: Apologies for back-tracking. But given @alcaeus's recent feature request of CDRIVER-4142, would it be preferable to hold off on the alias, and reduce the scope of CDRIVER-3441 to documenting `mongoc_handshake_data_append`?

chardan: That's a good question! I actually was wondering a bit about those semantics, myself. I'm inclined to agree with your intuition as well: the motivation for /this/ PR was to provide a name that fits in better with our extant conventions, as well as document a previously undocumented function. If we'd like the "new" name to have the "new" semantics, I think that would make sense: it doesn't break old code, and the funky name should actually partially alert callers to some funky behavior. Perhaps @alcaeus' cautionary notes should be adapted into the documentation as well. 

If you both think this is a good idea, I'm for it.

chardan: Note: The changes in this PR still seem generally useful to me-- I the change would be that I remove mongo_add_driver_info() and change the references to it back to mongoc_handshake_data_append().

chardan:
Question: I mention that the "old" function name is deprecated. Is that actually true?

src/libmongoc/doc/mongoc_handshake.rst
chardan: Updated as per discussion; OR we can revert the change entirely.

chardan:
Updated docs to remove reference to being deprecated.

src/libmongoc/doc/mongoc_handshake.rst
```diff
@@ -0,0 +1,80 @@
...
+     name: "mongoc",
+     version: "1.5.0"
+   },
+   os: {...},
+   platform: "CC=gcc CFLAGS=-Wall -pedantic"
+ }
+
+If we call
+  mongoc_handshake_data_append ("phongo", "1.1.8", "CC=clang")
+and it returns true, the driver sends handshake data like:
```
kevinAlbs, line 68: This produces a warning triggering a failure in the `make-release-archive` task: https://spruce.mongodb.com/task/mongo_c_driver_releng_make_release_archive_patch_d2838038aa7d1808b872ea7bd39df33db5bba364_613294890305b906ebc5c3df_21_09_03_21_32_59/logs?execution=0

You can test the documentation build locally with:
```
$ cd /path/to/mongo-c-driver
$ cd src/libmongoc/doc
$ sphinx-build -W -b html . html
```

src/libmongoc/doc/mongoc_handshake.rst
```diff
@@ -0,0 +1,80 @@
+:man_page: mongoc_handshake_data_append
+
+mongoc_handshake_data_append()
+=========================================
```
kevinAlbs, line 4: IIRC sphinx warns on mismatched underlines:

```suggestion
==============================
```

src/libmongoc/doc/mongoc_handshake.rst
```diff
@@ -0,0 +1,80 @@
...
+Description
+-----------
+
+This function is intended for use by drivers which wrap the C Driver.
+Calling this function will store the given strings as handshake data about
+the system and driver by appending them to the handshake data for the
+underlying C Driver. These values, along with other handshake data collected
+during mongoc_init will be sent to the server as part of the initial
+connection handshake in the "client" document. This function must not be
+called more than once, or after a handshake has been initiated.
```
kevinAlbs, line 27: I suggest clarifying the handshake is initiated when server monitoring begins. The [connection pooling](http://mongoc.org/libmongoc/current/connection-pooling.html) page gives additional background into the distinction of singled / pooled monitoring. Feel free to reword the suggestion.

```suggestion
called more than once, or after server monitoring begins. For a single-threaded :symbol:`mongoc_client_t`, server monitoring begins on the first operation requiring a server. For a :symbol:`mongoc_client_pool_t`, server monitoring begins on the first call to `:symbol:`mongoc_client_pool_pop`.
```

src/libmongoc/doc/mongoc_handshake.rst
```diff
@@ -0,0 +1,80 @@
...
+
+The passed in strings are copied, and don't have to remain valid after the
+call to mongoc_handshake_data_append(). The driver may store truncated
```
kevinAlbs, line 30:
```suggestion
call to :symbol:`mongoc_handshake_data_append()`. The driver may store truncated
```

To add a link.

src/libmongoc/doc/mongoc_handshake.rst
```diff
@@ -0,0 +1,80 @@
...
+versions of the passed in strings.
+
+Note:
```
kevinAlbs, line 33: Oh, good idea noting this! 

RST provides a `note` directive to indent and format. This may be applicable here.

```suggestion
.. note::
```

src/libmongoc/doc/mongoc_handshake.rst
```diff
@@ -0,0 +1,80 @@
...
+  This function allocates memory, and therefore caution should be used when
+  using this in conjunction with bson_mem_set_vtable. If this function is
```
kevinAlbs, line 35: I suggest adding `:symbol:` links here on the references functions listed here.

src/libmongoc/doc/mongoc_handshake.rst
```diff
@@ -0,0 +1,80 @@
...
+and it returns true, the driver sends handshake data like:
+
+.. code-block::
+ client: {
+   driver: {
+     name: "mongoc / phongo",
+     version: "1.5.0 / 1.1.8"
+   },
+   os: {...},
+   platform: "CC=gcc CFLAGS=-Wall -pedantic / CC=clang"
```
kevinAlbs, line 77:
```suggestion
   platform: "CC=clang / CC=gcc CFLAGS=-Wall -pedantic"
```

From [this test](https://github.com/kevinAlbs/c-bootstrap/blob/90f99b57487339c3c8ac75fad259d6da1ade5e94/investigations/handshake_append.3441/main.c#L28), the current behavior prefixes and does not prepend any additional text.

src/libmongoc/doc/mongoc_handshake.rst
```diff
@@ -0,0 +1,80 @@
...
+   driver: {
+     name: "mongoc",
+     version: "1.5.0"
+   },
+   os: {...},
+   platform: "CC=gcc CFLAGS=-Wall -pedantic"
+ }
+
+If we call
+  mongoc_handshake_data_append ("phongo", "1.1.8", "CC=clang")
```
kevinAlbs, line 67:
```suggestion
  mongoc_handshake_data_append ("phongo", "1.1.8", "CC=clang / ")
```

See my comment below, but I believe the `platform` argument is prepended without additional text.

src/libmongoc/src/mongoc/mongoc-handshake.h
```diff
@@ -28,64 +28,7 @@ BSON_BEGIN_DECLS
 
 #define MONGOC_HANDSHAKE_APPNAME_MAX 128
 
-/**
- * mongoc_handshake_data_append:
```
kevinAlbs, line 32: Ah, I now realize much of the RST text comes from the comment here.

src/libmongoc/doc/mongoc_handshake.rst
kevinAlbs: I agree with removing `mongoc_add_driver_info()` in favor of only documenting `mongoc_handshake_data_append()`. I have updated the description of CDRIVER-3441.

src/libmongoc/src/mongoc/mongoc-handshake.h
chardan: Right, there wasn't a lot else to go on!

src/libmongoc/doc/mongoc_handshake.rst
chardan: The document (from the code comments) does say "something like", but it's better to be more rather than less accurate!

src/libmongoc/doc/mongoc_handshake_data_append.rst
```diff
@@ -0,0 +1,84 @@
...
+Appends the given strings to the handshake data for the underlying C Driver.
+
+Description
+-----------
+
+This function is intended for use by drivers which wrap the C Driver.
+Calling this function will store the given strings as handshake data about
+the system and driver by appending them to the handshake data for the
+underlying C Driver. These values, along with other handshake data collected
+during mongoc_init will be sent to the server as part of the initial
```
kevinAlbs, line 25:
```suggestion
during :symbol:`mongoc_init` will be sent to the server as part of the initial
```

kevinAlbs (approved):
LGTM!

PR #854, opened 2021-08-27
CDRIVER-4114: Reduce contention on the topology description

vector-of-bool:
This is a fairly substantial changeset, and may require an equally substantial
explanation.

The main mutable state of a `mongoc_topology_t` is all contained within the
description, which contains the known state of the associated server topology.
Other pieces of state might also be updated in parallel, but they can be managed
through regular atomic operations.

The purpose of opening this as a draft is to get eyes on the work to ensure I
have a proper understanding of how the topology's state is meant to be
maintained, and that the correct understanding is reflected in the code.

This work builds upon changes in #840 (new atomics) and #847 (new session
pooling).

## Important draft TODOs and notes

- ~~This is still a draft. It may have speeling ,and   grammar error .~~
- Several new internal APIs have abbreviated names, which is uncommon in the
  rest of the codebase. This was purely to save my hands the strain of typing
  verbose names all the time 🙂. Anything can be renamed before this PR is
  merged.
- ~~Several doc comments are now outdated.~~
- ~~Very few usages of `topology::mutex` have been removed yet. These will be
  pruned later on once the topology-update-queue is fully implemented.~~ Done.
- ~~Various C99 usage throughout.~~

## A summary of the changes

- The introduction of a "shared pointer" type. This is analogous to a C++
  `shared_ptr` in design and behavior:
  - A shared_ptr is created with a pointer than was previously allocated, and a
    destructor+deallocator function.
  - The shared pointer maintains auxiliary data that tell it how many live
    references remain, and the destructor+deallocator.
  - Copying and destroying different shared pointers that refer to the same
    object is thread-safe.
  - Copying a shared pointer object that is also being updated is *not*
    thread-safe, and must be done using the `_atomic()` functions.
  - When the reference count of a shared pointer reaches zero, the
    destructor+deallocator function is immediately invoked with the managed
    pointer (this occurs in the thread that decremented the reference count to
    zero).
  - Shared pointer objects are layout-compatible with a regular pointer. This is
    used in `union mc_shared_tpl_descr`.
- The `mongoc_topology_t::description` member has been replaced with a
  `union mc_shared_tpl_descr _shared_descr_` member. Users that wish to access
  the description must do so by calling an API that will return them a shared
  pointer to the description, which the caller must then release when it is
  finished. Acquiring this shared pointer is "taking" the description, and
  releasing it is "dropping" the description.
- In the current code, the macro `MC_DECL_TD_TAKE (VarName, Topology)` is used
  to simultaneous declare a define `mc_shared_tpl_descr VarName` as _taking_ the
  shared description of `Topology`. `MC_TD_DROP(VarName)` should be used to
  _drop_ that pointer. A shared pointer can be rebound using
  `tdptr = mc_tpld_take_ref(topology)`
- The topology referred-to by `_shared_descr_` is never updated in-place.
  Instead, a copy is made of the description, updates are applied to that copy,
  and `_shared_descr_` is atomically swapped with the modified copy. This
  prevents someone attempting to take the shared description from seeing a torn
  value.
- Finding out when to use `mc_tpld_take_ref()` and `mc_tpld_drop_ref()` is a current
  main difficulty of this changeset.
  - When `mc_tpld_take_ref()` are called, the
    instantaneous state of the topology is captured into the resulting variable,
    and is guaranteed to remain unchanged when accessed through that shared
    pointer instance. If another shared pointer is taken, then there is no
    guarantee that the new shared pointer refers to the same topology.
  - Referring to points that call `bson_mutex_{un}lock` is a good guide on where
    to take and drop these references. Within a locked section, it can be
    assumed that the code expects a single view of the topology from the
    beginning to the end.
  - Functions that remark to "expect the caller to hold the topology mutex"
    means that they intend to see a consistent view of the topology during their
    execution (_or_ it means the function is intending to modify the topology,
    but that's discussed later).
    - Because these functions need to refer to the topology description it would
      _almost_ be correct to just take a shared pointer instead of assuming the
      caller holds a mutex, except this is **wrong**: The function expects a
      consistent view of the topology before _and_ after it is finished
      execution. A caller may hold a shared pointer `td1`, call one of these
      functions, which will then take a shared pointer `td2`. There is no
      guarantee that `td1` and `td2` point to the same topology description, so
      the callee and caller may be working with different descriptions, whereas
      with the mutex they would have been able to assume they are equivalent.
    - The solution is that functions that "expect the topology mutex to be held"
      need to instead be given a handle to the topology description state that
      the caller is expecting them to work on. These functions have had a
      `mongoc_topology_description_t*` parameter added, and each caller will
      pass-down the topology description that they are keeping alive using a
      shared pointer. These functions that "expect the topology mutex to be
      held" will not refer to the shared pointer of the topology, and defer
      instead to the topology that was given to them by their caller.
- At some points, functions are invoked that are expected to update the topology
  (e.g. topology scanning). In this case, the caller takes a shared pointer `td`
  and checks if it is valid/stale. If the `td` is okay, it continues with `td`.
  If not, it will invoke a scan operation and block. After the scanning
  completes, the caller must drop `td` and then _rebind_ `td` by taking the
  topology description again. This allows the callers to see the updated
  topology after scanning has completed.
- In the case of multithreaded scanning/monitoring, the topology
  description updates need to be queued and processed _one at a time_. If
  multiple threads were allowed to concurrently modify the topology, some
  topology updates might be lost. For example: If thread `th1` clones the
  topology description and begins updating it, and then thread `th2` takes
  _another_ clone, quickly modifies it, and then swaps it back in place, then
  `th1` might finish its modifications and write back the topology description
  based on the copy it took before `th2` did its modifications, thus losing the
  modifications done by `th2`. **(Solved below)**

# Update:

## `const` for all-the-things

There was trouble in deciding where to take a copy of the shared topology
description, versus when we would need to duplicate the description so that it
could be modified and later swapped out. The solution was to make the
`_shared_descr_` be a pointer-to-`const`. Thus, any caller of
`mc_tpld_take_ref()` would obtain a
shared-pointer-to-`const mongoc_topology_description_t`, and be unable to modify
the description or bind it to a pointer-to-non-`const` (I'll abbreviate as
`mutable*`). From this point, it was just a matter of finding everywhere that a
we attempt to create a `mutable*` and from a `const*` deciding:

1. Should the `mutable*` become a `const*`?
2. Are we actually attempting to modify the topology description?

In the case of [1], it was simply a matter of adding more `const` at the
use-site and in the function signatures where we are trying to pass the
`const*`. After modifying the callee signatures to take `const*`, it would often
lead to *more* sites where we attempt to form `mutable*` from a `const*`. In
every case, this would eventually result in:

- The entire call-tree becomes `const`-correct, and we can know that any time we
  see `const mongoc_topology_description_t*`, we know that we are not going to
  attempt to modify it.
- Eventually, adding a `const` would be impossible, because we are attempting to
  modify the topology. In this case, we backtrack and find where we originally
  took the shared-`const*` and mark that as a point where we need to perform a
  topology modification. That is, these sites become an instance of [2] above.

Additionally, `mongoc_topology_description_t` contains (*contained*) a member
`mongoc_set_t* servers` which contains a set of `mongoc_server_description_t`.
The `servers` pointer member is an owning-pointer, and the server descriptions
are themselves a part of the topology. Thus, when we have a
`mongoc_topology_description_t const*`, we should only access it's
`mongoc_server_description_t` instances through `const*` as well.

Because pointer members are shallow-`const`, the `servers` member was renamed to
`_servers_`, and any attempt to access the server descriptions of a topology
description now must go through two getter functions:

- ```c
  mc_tpld_servers(mongoc_topology_description_t*)
    -> mongoc_set_t*
  ```
- ```c
  mc_tpld_servers_const(mongoc_topology_description_t const*)
    -> mongoc_set_t const*
  ```

Because it is invalid to pass a `const*` to `mc_tpld_servers`, it is not
possible to obtain a `mutable*` for a `mongoc_server_description_t` when all you
have is a `mongoc_topology_description_t const*` (unless you do a cast. Don't do
a cast).

> Note: There were not yet any `const`-aware `mongoc_set` functions, so those
> were also added as part of this PR.

## Modifying the Topology

Now we need to define the procedure to update the topology. There are a few
requirements:

- Only one thread may update a topology description at any time. Everyone else
  needs to wait their turn.
- Any number of threads can still inspect the topology while another thread is
  performing an update.
- Threads should not see any "partial" updates to the topology. The topology
  update must appear atomic threads other than the one doing the updates.
- After an update is committed, other threads that still have a reference to the
  old topology should be safe to keep using that old reference.

All of the above are addressed by a few new APIs:

```c++
struct mc_tpld_modification {
  mongoc_topology_description_t* new_td;
  mongoc_topology_t* topology;
};

mc_tpld_modification
mc_tpld_modify_begin(mongoc_topology_t* topo);

void
mc_tpld_modify_commit(mc_tpld_modification tdmod);

void
mc_tpld_modify_drop(mc_tpld_modification tdmod);
```

They work as such:

- `mc_tpld_modification`: A composite object that contains state for the
  topology update. The most important member is `new_td`, which is a
  *non-`const`* pointer to a topology description. Through this pointer we
  should apply updates to the topology description. `topology` is pointer to
  the topology that owns the descriptions.
- `mc_tpld_modify_begin(tpl)`: Locks a mutex in `tpl` that is used to serialize
  modifications to the topology description. The current topology description is
  cloned and put into `mc_tpld_modification::new_td` on the returned object.
- `mc_tpld_modify_drop(tdmod)`: Drops any pending modifications defined by
  `tdmod`. Unlocks the topology modification mutex. Destroys and deallocates
  `new_td`.
- `mc_tpld_modify_commit(tdmod)`: Perform the atomic update of the topology
  description. After this function returns, the topology description from
  `new_td` is the new description for the whole topology. Unlocks the
  modification mutex. `new_td` is bound to the
  shared-pointer stored in `topology`.

A call to `mc_tpld_modify_begin()` *must* be later followed by a call to
*either* `mc_tpld_modify_drop()` or `mc_tpld_modify_commit()` (*never both*).

For example:

```c
mc_shared_tpl_descr td = mc_tpld_take_ref(tpl); // [1]
if (needs_to_update(td.ptr)) {  // [2]
  mc_tpld_modification tdmod = mc_tpld_modify_begin(tpl);  // [3]
  if (needs_to_update(tdmod.new_td)) {   // [4]
    do_update_topology(tdmod.new_td);    // [5]
    mc_tpld_modification_commit(tdmod);  // [6]
  } else {
    mc_tpld_modification_drop(tdmod);    // [7]
  }
}
mc_tpld_drop_ref(&td);  // [8]
```

Explained:

1. Take a reference to the topology description at that exact moment. `td.ptr`
   will remain valid until [8], and the pointed-to-description will remain
   unchanged regardless of any intermediate code.
2. Check whether we need to update the topology. This may come in the form of an
   error during normal execution, rather than an explicit check for
   outdated-ness.
3. Begin an update transaction.
4. *Check again!:* The call to `mc_tpld_modify_begin()` may have been preempted
   by another thread that was in the middle of performing an update. The other
   thread may have already performed the update that we expected ourselves to
   do. It is not guaranteed that the `td` in the outer scope is the same as the
   `tdmod.new_td`, since `new_td` was cloned after securing the exclusive lock
   to perform a topology update.
5. Pass the `mutable*` to the topology description to a function that will
   perform the updates. No other thread will have access to this pointer, so no
   further synchronization is necessary.
6. Commit the update. Now other threads will see our updates the next time they
   pull the topology description.
7. If `needs_to_update(tdmod.new_td)` returned `false`, abandon the update.
8. `td` is still holding a reference to the topology description at the moment
   of the call to `mc_tpld_take_ref` in `[1]`. We always need to drop those
   references.


src/libmongoc/src/mongoc/mongoc-shared.c
```diff
@@ -0,0 +1,163 @@
...
+   aux->dtor (aux->managed);
+   bson_free (aux);
+}
+
+static bson_mutex_t g_shared_ptr_spin_mtx = PTHREAD_MUTEX_INITIALIZER;
+
+static void
+_shared_ptr_spin_lock ()
+{
+   bson_mutex_lock (&g_shared_ptr_spin_mtx);
```
kevinAlbs, line 42: I see this was changed from a spin lock in a prior commit. I have not thought about spin locks in a while, so please excuse the silly question.

IIRC spin locks loop and fully utilize the CPU. A mutex will make a system call into the kernel, and put the calling thread to sleep.

When the expectation is that the lock will not need to wait for very long, a spin lock is preferable to avoid the overhead of the system call of a mutex.

I am not sure which is preferable for shared pointer. Did the original spin lock run into problems?

src/libmongoc/src/mongoc/mongoc-shared.h
```diff
@@ -0,0 +1,195 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef MONGOC_SHARED_H
+#define MONGOC_SHARED_H
```
kevinAlbs, line 18: `mongoc_shared_ptr` seems like a really useful abstraction, but I vote to make this header private unless there is a use-case to expose it. That gives us the freedom to make backwards breaking changes to struct layouts or function declarations.

src/libmongoc/src/mongoc/mongoc-shared.h
```diff
@@ -0,0 +1,195 @@
...
+/** Get the managed pointer owned by the given shared pointer */
+#define MONGOC_Shared_Pointee(Type, Pointer) ((Type *) ((Pointer).ptr))
+
+/**
+ * @brief Create a new shared pointer instance from 'Pointer' and bind a regular
+ * pointer 'Type *VarName' in the calling scope.
+ *
+ * @param Type The pointed-to type of the shared pointer
+ * @param VarName A name for a regular pointer to declare in the enclosing scope
+ * @param Pointer A shared ppointer instance to take from
```
kevinAlbs, line 163:
```suggestion
 * @param Pointer A shared pointer instance to take from
```

src/libmongoc/src/mongoc/mongoc-shared.c
```diff
@@ -0,0 +1,163 @@
...
+void
+mongoc_shared_ptr_rebind_atomic (mongoc_shared_ptr *const out,
+                                 mongoc_shared_ptr const from)
+{
+   struct _mongoc_shared_ptr_aux *prev_aux = NULL;
+   size_t prevcount = 0;
+   assert (out &&
+           "NULL given as output argument to mongoc_shared_ptr_rebind_atomic");
+
+   {
```
kevinAlbs, line 100: Is the additional `{` for scope necessary?

src/libmongoc/src/mongoc/mongoc-shared.c
```diff
@@ -0,0 +1,163 @@
...
+         prevcount = bson_atomic_int_fetch_sub (
+            &out->_aux->refcount, 1, bson_memorder_seqcst);
+      }
+      *out = from;
+      bson_atomic_int_fetch_add (&out->_aux->refcount, 1, bson_memorder_seqcst);
+      _shared_ptr_spin_unlock ();
+   }
+
+   if (prevcount == 1) {
+      _release_aux (prev_aux);
```
kevinAlbs, line 113: Does the previous `out->ptr` need to be freed in this case?

src/libmongoc/src/mongoc/mongoc-shared.c
```diff
@@ -0,0 +1,163 @@
...
+                                 mongoc_shared_ptr const from)
+{
+   struct _mongoc_shared_ptr_aux *prev_aux = NULL;
+   size_t prevcount = 0;
+   assert (out &&
+           "NULL given as output argument to mongoc_shared_ptr_rebind_atomic");
+
+   {
+      _shared_ptr_spin_lock ();
+      prev_aux = out->_aux;
```
kevinAlbs, line 102: Could `mongoc_shared_ptr_rebind_raw` be called while holding the lock to simplify this implementation?

src/libmongoc/src/mongoc/mongoc-shared.h
```diff
@@ -0,0 +1,195 @@
...
+
+/**
+ * @brief A ref-counted thread-safe shared pointer to arbitrary data
+ */
+typedef struct mongoc_shared_ptr {
+   /** Pointed-to data */
+   void *ptr;
+   /** Auxilary book-keeping. Do not touch. */
+   struct _mongoc_shared_ptr_aux *_aux;
+} mongoc_shared_ptr;
```
kevinAlbs, line 32: To check, can a pointer to the same `mongoc_shared_ptr` be safely shared between threads? Or do users of `mongoc_shared_ptr` only be sharing copies of the struct between threads?

Based on the comment above `mongoc_shared_ptr_rebind_atomic` I believe a pointer to `mongoc_shared_ptr` can be shared between threads, as long as the `*_atomic` functions are used as appropriate.

kevinAlbs (commented):
This is looking great so far. I've taken a look at the shard pointer abstraction.

I am going to go back to https://github.com/mongodb/mongo-c-driver/pull/847 and https://github.com/mongodb/mongo-c-driver/pull/840 before proceeding.

eramongodb (changes requested):
There is a major concern I have with the design as currently proposed.

As proposed, a read-only reference to a given topology description object is obtained via `mc_tpld_take_ref()`, which creates a copy of the owning shared pointer via `mongoc_shared_ptr_take_atomic()`. This copy requires synchronization via a global mutex `g_shared_ptr_spin_mtx`. Contrary to the PR description, this behavior is _not_ analogous to C++ `std::shared_ptr<T>`, which does _not_ provide thread-safe operations on shared pointer objects themselves. Only the reference count should be thread-safe; synchronizing access of the managed object is not the responsibility of `std::shared_ptr<T>`. The presence of this global mutex complicates the `mongoc_shared_ptr` interface and misleads the reader into believing `mongoc_shared_ptr` provides more thread-safety than it actually does.

Global thread synchronization was not necessary prior to these proposed changes. Synchronization was limited to a given topology object via the `mongoc_topology_t::mutex` data member. Requiring global synchronization on every guarded access to a topology description (even when shared pointers completely unrelated to topology descriptions must be copied or rebound) does not seem advisable when the goal of this PR is to reduce contention. The scope of potential contention has expanded from "per topology object" to the unbounded global scope, regardless if a given shared pointer is being used for topology descriptions or not. Synchronization of access to `mongoc_topology_t::_shared_descr_` should remain limited to a given `mongoc_topology_t` object, not global scope.

Reiterating the requirements for topology description modification listed:

* Only one thread may update a topology description at any time. Everyone else
needs to wait their turn.
* Any number of threads can still inspect the topology while another thread is
performing an update.
* Threads should not see any "partial" updates to the topology. The topology
update must appear atomic threads other than the one doing the updates.
* After an update is committed, other threads that still have a reference to the
old topology should be safe to keep using that old reference.

This requirements suggest to me that access of the topology description object should be guarded by a reader-writer lock per `mongoc_topology_t` object. The proposed `mc_tpld_take_ref()` already identifies locations where a reader lock would be appropriate. Similarly, the proposed `mc_tpld_modify_commit()` identifies locations where a writer lock would be appropriate.

Using `_mongoc_topology_get_type()` as example of read-only access (relative to master):

```diff
mongoc_topology_description_type_t
_mongoc_topology_get_type (mongoc_topology_t *topology)
{
   mongoc_topology_description_type_t td_type;

-  bson_mutex_lock (&topology->mutex);
+  MC_TPLD_READER_LOCK (topology, td);

-  td_type = topology->description.type;
+  td_type = td->type;

-  bson_mutex_unlock (&topology->mutex);
+  MC_TPLD_READER_UNLOCK (td);

   return td_type;
}
```

And `mongoc_topology_invalidate_server ()` as example of write access (relative to master):

```diff
void
mongoc_topology_invalidate_server (mongoc_topology_t *topology,
                                   uint32_t id,
                                   const bson_error_t *error)
{
   BSON_ASSERT (error);

-  bson_mutex_lock (&topology->mutex);
-  mongoc_topology_description_invalidate_server (
-     &topology->description, id, error);
-  bson_mutex_unlock (&topology->mutex);
+  MC_TPLD_WRITER_LOCK (topology, td);
+  mongoc_topology_description_invalidate_server (
+     td, id, error);
+  MC_TPLD_WRITER_UNLOCK (topology);
}
```

This can be implemented using already-existing mutex and condition variable utilities, but it can just as well be implemented using the proposed atomic functions (as semaphores). This keeps potential contention limited strictly to a given `_mongoc_topology_t` object and its topology description shared pointer (no global mutex), while also freeing up contention between read-only access to the topology description object. Determining when and how locks of `topology->mutex` should be substituted with a reader lock or writer lock should be much more straightforward compared to the state of the current proposal.

As already proposed in the PR, contention can be reduced further by limiting reader lock duration to copying the shared pointer when the corresponding topology description need not be "up-to-date" (this is good):

```c
MC_TPLD_READER_LOCK (topology, td)
mongoc_shared_ptr sptr = mongoc_shared_ptr_take (topology->_shared_descr_);
td = sptr.ptr;
MC_TPLD_READER_UNLOCK (topology)

use(td);

mongoc_shared_ptr_release (sptr);
```

The proposed changes to pass a topology description object via parameters would still be necessary to enforce a "consistency view" of the topology (also good). The addition of `const`-qualifiers is consistent with and encouraged by the reader-writer lock pattern, but should not be strictly necessary given modifications are limited to replacing `mongoc_topology_t::_shared_descr_` (readers holding shared pointers to the old topology description would see no modifications); nevertheless, I welcome their addition (the C driver could use more `const` in general).

In summary, the global `g_shared_ptr_spin_mtx` seems poorly motivated. `mongoc_shared_ptr` should accurately model `std::shared_ptr<T>` without behavioral extensions. Access to `mongoc_topology_t::_shared_descr_` should be guarded on a per `mongoc_topology_t` object basis by means of a reader-writer lock, not a global mutex.

src/libmongoc/src/mongoc/mongoc-ts-pool.c
```diff
@@ -0,0 +1,268 @@
...
+      pool->params.constructor (node->data, pool->params.userdata, error);
+      if (error->code != 0) {
+         /* Constructor reported an error. Deallocate and drop the node. */
+         bson_free (node);
+         node = NULL;
+      }
+   }
+   if (node) {
+      bson_atomic_int32_fetch_add (
+         &pool->outstanding_items, 1, bson_memorder_relaxed);
```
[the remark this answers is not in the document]
vector-of-bool: These atomic operations are not used for synchronization, so their operations do not require specific ordering.

src/libmongoc/src/mongoc/mongoc-topology-scanner-private.h
```diff
@@ -85,15 +86,33 @@ typedef struct mongoc_topology_scanner_node {
...
+   /**
+    * The handshake command could not be constructed because it would be too
+    * large. The handshake_cmd pointer will be NULL.
+    */
+   HANDSHAKE_CMD_TOO_BIG,
+   /**
+    * The handshake command is valid and ready to be copied-from.
+    */
+   HANDSHAKE_CMD_OKAY,
+} handshake_state_t;
```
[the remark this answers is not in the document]
vector-of-bool: The state of `handshake_cmd` being `NULL` ambiguous alone (it was either uninitialized or "too big"), and thus necessitated the `handshake_ok` member variable. Synchronizing these two variables independently required more code than just representing the state in a three-state flag of `hanshake_state_t`.

src/libmongoc/src/mongoc/mongoc-shared.c
vector-of-bool: Yes, the original spin lock ran into problems. The issue is that it breaks the forward-progress guarantees of the lock-holder. If there are three threads on a 2-hw-thread system, and one thread is holding the spin lock and gets preempted, and then the other two threads attempt to take the spin lock, there is no indication to the kernel that they are spinning waiting for the sleeping thread to release the lock. If there are *hundreds* of threads blocking on the spinlock (as was happening in my benchmarking), it can almost appear to deadlock since all threads are waiting on the lock holder to get woken up by the kernel, which can take an arbitrary amount of time since the kernel doesn't know that it needs to do so. This can be fixed by yielding the thread when you spin, but is less than optimal because the scheduler doesn't know why you're doing that.

I believe the best tradeoff is to use a limited number of `try_lock()` spins, which will never actually syscall, followed by a `lock()` that might syscall and actually sleep the thread. This is the approach taken by LLVM's libc++.

src/libmongoc/src/mongoc/mongoc-shared.c
vector-of-bool: No. I just use it as habit to denote the scope of the lock.

src/libmongoc/src/mongoc/mongoc-shared.c
vector-of-bool: Yes, because the rebind-target might have been pointing to something that needs to be released.

src/libmongoc/src/mongoc/mongoc-shared.c
vector-of-bool: That would simplify it, but it would move the `_release_aux (out->_aux)` into the scope of the lock, which may execute `out->_aux->dtor()`, leading to arbitrary code executing while the lock is being held.

src/libmongoc/src/mongoc/mongoc-shared.h
vector-of-bool: There's three different objects when dealing with a shared_ptr: A `shared_ptr` instance, the shared state (stored in `_aux`), and the pointed-to data. Only the shared state is thread-safe. A `shared_ptr` object is not itself thread-safe, which is why we must duplicate the shared_ptr via `take_atomic(&p)` if `p` is accessible between threads and one of those threads might rebind `p` (as is the case with `topology->_shared_descr_`).

In the case of `_shared_descr_`, the pointee is `const`, and thus requires no synchronization as no other thread can modify it while we use it (as long as no one tries anything tricky). The only shared mutable state is the `_shared_descr_` object itself, thus requiring the synchronization.

src/libmongoc/src/mongoc/mongoc-topology-private.h
```diff
@@ -95,14 +121,20 @@ typedef struct _mongoc_topology_t {
    bson_thread_t srv_polling_thread;
    mongoc_cond_t srv_polling_cond;
 
-   bson_mutex_t mutex;
```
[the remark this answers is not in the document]
vector-of-bool: Looks like there's a bit more work to do in the case of CSE. I'll have to take a deeper look here, since there may be more shared state that I didn't notice.

eramongodb:
@vector-of-bool Thank you for the thorough response.

Your reference to atomic shared_ptr operations cleared up a lot my previously stated concerns. The global mutex remains surprising, but I will trust your reference to libc++'s implementation and your own benchmarks. You make compelling arguments that it is not as severe a concern as I had initially assumed. If updates to the topology description can be committed as an atomic exchange of the `_shared_descr_` shared pointer, and contention is measurably infrequent, I can see the argument for why it would be preferable to a reader-writer lock. I believe renaming several of the associated functions to better align with their C11 / C++11 counterparts will help avoid much of the confusion I fell into (I had misunderstood the purpose of the global mutex). I will follow up on this in more detail in the new PR https://github.com/mongodb/mongo-c-driver/pull/858.

src/libmongoc/src/mongoc/mongoc-shared.c
eramongodb: I am also in favor of using `{}` to denote critical sections or similarly distinct local blocks of code, even if technically unnecessary.

src/libmongoc/src/mongoc/mongoc-shared.h
kevinAlbs: Thank you for the explanation. I now see that the `mongoc_shared_ptr` struct cannot safely be accessed between threads, and must be copied.

src/libmongoc/src/mongoc/mongoc-shared.c
kevinAlbs: Using `{}` to denote the scope of the lock SGTM.

src/libmongoc/src/mongoc/mongoc-shared.c
kevinAlbs: Good point. Resolving.

src/libmongoc/src/mongoc/mongoc-shared.c
kevinAlbs: > If there are three threads on a 2-hw-thread system, and one thread is holding the spin lock and gets preempted, and then the other two threads attempt to take the spin lock, there is no indication to the kernel that they are spinning waiting for the sleeping thread to release the lock. If there are hundreds of threads blocking on the spinlock (as was happening in my benchmarking), it can almost appear to deadlock since all threads are waiting on the lock holder to get woken up by the kernel, which can take an arbitrary amount of time since the kernel doesn't know that it needs to do so

Interesting, thank you for the thorough explanation. That sounds like a priority inversion problem – the thread holding the spin lock should be the highest priority to schedule but the kernel is not aware.

> I believe the best tradeoff is to use a limited number of try_lock() spins, which will never actually syscall, followed by a lock() that might syscall and actually sleep the thread. This is the approach taken by LLVM's libc++.

That sounds close to the implementation proposed in https://github.com/mongodb/mongo-c-driver/pull/840/files#diff-7591f9551d8904ff97695bc90f2d6b819afe82b783c4e798b046d597acd3880fR57 for `_lock_64bit_atomic`. That might be interesting to explore if we need to fine tune this.

kevinAlbs (commented):
I will continue reviewing the new shared pointer API in #858.

src/libbson/src/bson/bson-macros.h
```diff
@@ -196,18 +196,48 @@
...
+ */
+#define BSON_ASSERT_INLINE(Assertion, Value)                              \
+   ((void) ((Assertion) ? (0)                                             \
+                        : ((fprintf (stderr,                              \
+                                     "%s:%d %s(): Assertion '%s' failed", \
+                                     __FILE__,                            \
+                                     __LINE__,                            \
+                                     BSON_FUNC,                           \
+                                     #Assertion),                         \
+                            abort ()),                                    \
```
chardan, line 212: Since we're calling abort() rather than exit(), does the abbove fprintf() need fflush()?

vector-of-bool: Whether files are flushed is implementation-defined, it appears. In all of my testing with the various `assert` macros, I've not found that files are not implicitly flushed before the program terminates. None of the other assertion macros flush, so I have omitted that here as well.

chardan: +1, it does appear that this is implementation-defined (and currently at least glibc does not appear to flush specifically because of contention issues, although it has done so in the past).

eramongodb (changes requested):
Big 👏  for improving const correctness of access of topology description, topology description servers, etc.

There are many instances (especially in test code) where `client->topology->_shared_descr_.ptr` is accessed directly. Should these be changed to access via `mc_tpld_unsafe_get_mutable` instead?

There are many instances (especially in test code) where `topology->tpld_modification_mtx` is locked manually. Should these be changed to calls to `mc_tpld_modify_begin` instead?

There is inconsistency in how the identifiers `td`, `tdmod`, `tpld`, and `topology` are being used throughout the codebase. It would be preferable if the choice of identifier was made consistent to avoid confusion (e.g. is a local variable or parameter named `topology` referring to a `mongoc_topology_t`, `mongoc_topology_description_t`, ` mc_shared_tpld`, or `mc_tpld_modification`?).

chardan (commented):
I think the only other big question mark for me is that, as has already been pointed out, it may be that the mutexes lock too broadly; otherwise, don't have a ton to add. Nice work on this!

src/libmongoc/tests/test-mongoc-dns.c
```diff
@@ -701,11 +699,9 @@ _prose_test_9 (bool pooled)
       _prose_loadbalanced_ping (client);
    }
 
-   bson_mutex_lock (&topology->tpld_modification_mtx);
    expected_hosts = MAKE_HOSTS ("localhost.test.build.10gen.cc:27017");
    check_topology_description (mc_tpld_unsafe_get_mutable (client->topology),
                                expected_hosts);
-   bson_mutex_unlock (&topology->tpld_modification_mtx);
```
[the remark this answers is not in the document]
vector-of-bool: Possibly, although `check_topology_description` is expecting `add_server()` to be a no-op (because the server description should already be registered). But also, you are right that another thread managing the pool might modify it. This can be handled by just taking a `tpld_modification` and using that to restore thread-safety, even though we don't intend to modify the topology.

src/libmongoc/tests/test-mongoc-exhaust.c
```diff
@@ -33,17 +33,15 @@ get_generation (mongoc_client_t *client, mongoc_cursor_t *cursor)
 {
    uint32_t server_id;
    uint32_t generation;
-   mongoc_server_description_t *sd;
+   mongoc_server_description_t const *sd;
    bson_error_t error;
 
    server_id = mongoc_cursor_get_hint (cursor);
 
-   bson_mutex_lock (&client->topology->tpld_modification_mtx);
```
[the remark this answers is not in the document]
vector-of-bool: Not necessarily a lock, but grabbing the topology the proper way will do the trick.

src/libmongoc/tests/test-mongoc-topology-description.c
```diff
@@ -47,24 +47,18 @@ _test_has_readable_writable_server (bool pooled)
       client, "admin", tmp_bson ("{'ping': 1}"), NULL, NULL, &error);
    ASSERT_OR_PRINT (r, error);
 
-   bson_mutex_lock (&topology->tpld_modification_mtx);
```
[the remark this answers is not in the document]
vector-of-bool: Updated with synchronization

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -1227,233 +1221,161 @@ mongoc_topology_select_server_id (mongoc_topology_t *topology,
...
+      }
 
-         TRACE ("server selection about to wait for %" PRId64 "ms",
-                (expire_at - loop_start) / 1000);
-         r = mongoc_cond_timedwait (&topology->cond_client,
-                                    &topology->mutex,
-                                    (expire_at - loop_start) / 1000);
-         TRACE ("%s", "server selection awake");
-         _topology_collect_errors (topology, &scanner_error);
+      bson_mutex_lock (&topology->tpld_modification_mtx);
```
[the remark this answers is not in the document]
eramongodb: Would the addition of a dedicate mutex for update synchronization be appropriate here (as done similarly for `srv_polling_mtx`), i.e. a `tpld_update_mtx` paired with `cond_client` (which should probably be renamed to something more descriptive like `tpld_update_cond`)?

src/libmongoc/src/mongoc/mongoc-set.c
```diff
@@ -27,7 +27,7 @@ mongoc_set_new (size_t nitems, mongoc_set_item_dtor dtor, void *dtor_ctx)
 {
    mongoc_set_t *set = (mongoc_set_t *) bson_malloc (sizeof (*set));
 
-   set->items_allocated = nitems;
+   set->items_allocated = BSON_MAX (nitems, 1);
```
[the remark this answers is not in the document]
eramongodb: > It is valid to create a set for 0 and then later insert more items

If that is the case, I am fine with the changes as proposed.

src/libmongoc/tests/TestSuite.c
```diff
@@ -118,6 +118,7 @@ _handle_signal (int signum)
 {
    const char *s = "\nProcess was interrupted by the delivery of a signal.\n";
    const char *sigstr;
+   size_t n;
```
eramongodb, line 121: Why the addition of this write-only variable? It seems as if it is to address nodiscard warnings on calls to `write()`, but if that is the case, a cast to `(void)` should suffice.

src/libmongoc/tests/test-mongoc-counters.c
```diff
@@ -26,31 +26,36 @@
...
-                    .slots[COUNTER_##id % SLOTS_PER_CACHELINE]; \
-      }                                                         \
-      return _sum;                                              \
+#define COUNTER(id, category, name, description)                               \
+   int32_t prev_##id;                                                          \
+   int32_t count_##id (void)                                                   \
+   {                                                                           \
+      int32_t _sum = 0;                                                        \
+      int32_t _i;                                                              \
+      for (_i = 0; _i < _mongoc_get_cpu_count (); _i++) {                      \
```
eramongodb, line 35: > warning: comparison of integers of different signs: 'int32_t' (aka 'int') and 'unsigned int' [-Wsign-compare]

I believe the `_i` counter should remain `uint32_t`.

eramongodb (changes requested):
👍 to renaming of functions such as `mongoc_topology_server_by_id` -> `mongoc_topology_description_server_by_id_const`.

👍 for addition of `srv_polling_mtx` to clear up responsibility of `tpld_modification_mtx`.

There is a mix of east and west const throughout proposed changes. I recommend making the const formatting consistent (AFAIK west const is the current format standard in the C driver).

src/libmongoc/src/mongoc/mongoc-topology-description-private.h
```diff
@@ -41,7 +41,7 @@ struct _mongoc_topology_description_t {
    bool opened;
    mongoc_topology_description_type_t type;
    int64_t heartbeat_msec;
-   mongoc_set_t *servers;
+   mongoc_set_t *_servers_;
```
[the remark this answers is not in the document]
kevinAlbs: Resolving. I am OK with no comment since there is no established convention.

kevinAlbs (approved):
LGTM. I do not think open comments are blockers.

> Several new internal APIs have abbreviated names, which is uncommon in the
rest of the codebase. This was purely to save my hands the strain of typing
verbose names all the time 🙂. Anything can be renamed before this PR is
merged.

I defer artistic freedom of naming to you. But I vote for consistency in the abbreviation. There are three different abbreviations for "topology description", consider choosing one. Here is one option:

- `MC_DECL_TD_TAKE` => `MC_DECL_TPLD_TAKE`
- `_topo_descr_destroy_and_free` => `_tpld_destroy_and_free`
- `mc_shared_tpld` => (no change)

src/libmongoc/tests/test-mongoc-topology-reconcile.c
```diff
@@ -39,24 +39,23 @@ get_node (mongoc_topology_t *topology, const char *host_and_port)
 
 
 static bool
-has_server_description (mongoc_topology_t *topology, const char *host_and_port)
+has_server_description (const mongoc_topology_t *topology,
+                        const char *host_and_port)
 {
-   mongoc_set_t *servers = topology->description.servers;
+   const mongoc_set_t *servers =
+      mc_tpld_servers_const (mc_tpld_unsafe_get_const (topology));
```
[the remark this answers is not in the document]
vector-of-bool: You're probably correct. No reason not to make this thread-safe.

src/libmongoc/src/mongoc/mongoc-topology-description.c
```diff
@@ -2328,5 +2320,29 @@ mongoc_topology_description_reconcile (mongoc_topology_description_t *td,
...
+   sd = mongoc_topology_description_server_by_id (td, server_id, &error);
+   if (!sd) {
+      /* Server removed, ignore and ignore error. */
+      return;
+   }
+
+   TRACE ("clearing pool for server: %s", sd->host.host_and_port);
+
+   mongoc_generation_map_increment (sd->generation_map, service_id);
+}
```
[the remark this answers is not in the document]
vector-of-bool: I think we could add a clang-format rule for this one.

src/libmongoc/src/mongoc/mongoc-topology-private.h
```diff
@@ -62,14 +64,38 @@ typedef struct _mongoc_rr_data_t {
...
+static const mc_shared_tpl_descr MC_SHARED_TPL_DESCR_NULL = {
+   MONGOC_SHARED_PTR_NULL};
+
 typedef struct _mongoc_topology_t {
-   mongoc_topology_description_t description;
+   /**
+    * @brief The topology description. Do not access directly. Instead, use
+    * mc_tpld_take_ref()
+    */
+   mc_shared_tpl_descr _shared_descr_;
```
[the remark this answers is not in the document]
vector-of-bool: I was thinking about it, but I then thought it would be better to keep it as strongly-typed as often as possible, since shared_ptr just contains a `void*`. I don't think it's a huge difference, and I'm still open to be convinced otherwise again.

src/libmongoc/tests/test-mongoc-topology.c
```diff
@@ -512,8 +499,9 @@ _test_topology_invalidate_server (bool pooled)
 
    bson_set_error (
       &error, MONGOC_ERROR_STREAM, MONGOC_ERROR_STREAM_SOCKET, "error");
-   mongoc_topology_invalidate_server (client->topology, id, &error);
-   sd = (mongoc_server_description_t *) mongoc_set_get (td->servers, id);
+   td = mc_tpld_unsafe_get_mutable (client->topology);
```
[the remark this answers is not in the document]
vector-of-bool: This is a good idea

src/libmongoc/tests/test-mongoc-topology.c
```diff
@@ -1711,6 +1699,7 @@ test_compatible_null_error_pointer (void)
       error, MONGOC_ERROR_PROTOCOL, MONGOC_ERROR_PROTOCOL_BAD_WIRE_VERSION, "");
 
    /* null error pointer is ok */
+   td = mc_tpld_unsafe_get_mutable (client->topology);
```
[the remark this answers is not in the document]
vector-of-bool: Updating docs on `mongoc_client_command_simple` is a larger task that updating doc-comments on internal APIs. The topology (AFAIK) is not part of the public API, so this would be a doc-comment on an internal component, but attached to an external API, which is confusion.

I have some orthogonal opinions on our current documentation process that I would like to bring up eventually.

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -1093,13 +1075,14 @@ mongoc_topology_select_server_id (mongoc_topology_t *topology,
...
    int64_t local_threshold_ms;
-   mongoc_server_description_t *selected_server = NULL;
+   const mongoc_server_description_t *selected_server = NULL;
    bool try_once;
    int64_t sleep_usec;
    bool tried_once;
    bson_error_t scanner_error = {0};
    int64_t heartbeat_msec;
    uint32_t server_id;
+   mc_shared_tpld td = mc_tpld_take_ref (topology);
```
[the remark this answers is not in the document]
vector-of-bool: Yeah, that's kind of a code-smelly comment anyway, and was only relevant to the prior implementation.

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -1120,26 +1102,30 @@ mongoc_topology_select_server_id (mongoc_topology_t *topology,
...
    expire_at =
       loop_start + ((int64_t) topology->server_selection_timeout_msec * 1000);
 
    if (topology->single_threaded) {
-      _mongoc_topology_description_monitor_opening (&topology->description);
+      if (!td.ptr->opened) {
+         mc_tpld_modification tdmod = mc_tpld_modify_begin (topology);
+         _mongoc_topology_description_monitor_opening (tdmod.new_td);
+         mc_tpld_modify_commit (tdmod);
+         mc_tpld_renew_ref (&td, topology);
```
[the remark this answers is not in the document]
vector-of-bool: I think this is a good question to debate on whether it is preferable to have uniformity, or to just use the direct method in the single-threaded case.

This case is fairly benign, as `monitor_opening` doesn't add/remove/tweak any server descriptions, so it is safe to do the "unsafe" thing. In other cases it may be that a caller has a live pointer to a server description which might be disappeared out from under them.

src/libmongoc/src/mongoc/mongoc-topology-private.h
```diff
@@ -107,7 +135,13 @@ typedef struct _mongoc_topology_t {
    bson_thread_t srv_polling_thread;
    mongoc_cond_t srv_polling_cond;
 
-   bson_mutex_t mutex;
+   /**
+    * @brief This lock is held in order to serialize operations that modify the
+    * topology description. It *should not* be held while performing read-only
+    * operations on the topology.
```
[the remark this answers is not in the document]
vector-of-bool: I updated the comment and added comments throughout that describe how the condition variable and this mutex are used. I hope they will be helpful for `$future_developer` (probably me, after I've forgotten what I was doing here).

src/libmongoc/src/mongoc/mongoc-cluster-cyrus.c
```diff
@@ -83,12 +83,7 @@ _mongoc_cluster_auth_node_cyrus (mongoc_cluster_t *cluster,
       TRACE ("SASL: authenticating (step %d)", sasl.step);
 
       server_stream = _mongoc_cluster_create_server_stream (
-         cluster->client->topology, sd, stream, error);
-
-      if (!server_stream) {
-         bson_destroy (&cmd);
-         goto failure;
-      }
+         cluster->client->topology, sd, stream);
```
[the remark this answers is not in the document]
vector-of-bool: Yikes. This is something definitely `-Werror` worthy. I didn't have SASL enabled on my local box, so I missed this, and CI doesn't run with that warning (probably should).

src/libmongoc/src/mongoc/mongoc-topology-background-monitoring.c
```diff
@@ -36,21 +36,16 @@ static BSON_THREAD_FUN (srv_polling_run, topology_void)
    mongoc_topology_t *topology;
 
    topology = topology_void;
-   bson_mutex_lock (&topology->mutex);
-   while (true) {
+   while (bson_atomic_int_fetch ((int *) &topology->scanner_state,
```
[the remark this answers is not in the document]
vector-of-bool: Ouch. Good find. I had wrongly assumed that `enum` was layout-compatible with `int` in all cases. Hmm...

src/libmongoc/tests/TestSuite.c
vector-of-bool: I thought so too. It seems like GCC (8.x) doesn't consider the `(void)write(...)` to be enough 😕.

src/libmongoc/tests/test-mongoc-counters.c
vector-of-bool: Good find. I don't know why I changed that.

src/libmongoc/src/mongoc/mongoc-topology-background-monitoring.c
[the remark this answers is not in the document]
vector-of-bool: I've changed `scanner_state` to an int, and rely on enum-int convertibility. It's not ideal, but it is well-defined in all cases.

src/libmongoc/src/mongoc/mongoc-topology-background-monitoring.c
```diff
@@ -69,39 +64,42 @@ static BSON_THREAD_FUN (srv_polling_run, topology_void)
...
 
       /* Check for shutdown again here. mongoc_topology_rescan_srv unlocks the
-       * topology mutex for the scan. The topology may have shut down in that
-       * time. */
-      if (topology->scanner_state != MONGOC_TOPOLOGY_SCANNER_BG_RUNNING) {
-         bson_mutex_unlock (&topology->mutex);
+       * topology srv_polling_mtx for the scan. The topology may have shut
+       * down in that time. */
+      bson_mutex_lock (&topology->srv_polling_mtx);
+      if (bson_atomic_int_fetch ((int *) &topology->scanner_state,
```
eramongodb, line 70:
```suggestion
      if (bson_atomic_int_fetch (&topology->scanner_state,
```

src/libmongoc/tests/test-mongoc-topology-reconcile.c
```diff
@@ -407,7 +407,9 @@ test_topology_reconcile_from_handshake (void *ctx)
    ASSERT_CMPINT (count, ==, 0);
 
    /* allow pool to start scanner thread */
-   topology->scanner_state = MONGOC_TOPOLOGY_SCANNER_OFF;
+   bson_atomic_int_exchange ((int *) &topology->scanner_state,
```
eramongodb, line 410:
```suggestion
   bson_atomic_int_exchange (&topology->scanner_state,
```

eramongodb (changes requested):

eramongodb (approved):

PR #856, opened 2021-08-28
CDRIVER-4083 + CDRIVER-4138

kevinAlbs:


src/libbson/src/bson/bson.c
```diff
@@ -415,6 +415,20 @@ _bson_append (bson_t *bson,              /* IN */
    return ok;
 }
 
+#define HANDLE_KEY_LENGTH(key, key_length)                                \
+   do {                                                                   \
+      if (key_length < 0) {                                               \
+         key_length = (int) strlen (key);                                 \
+      } else {                                                            \
+         /* Necessary to validate embedded NULL is not present in key. */ \
```
jmikola, line 423: Would it be simpler to return false if `key_length != strlen(key)`?

src/libbson/src/bson/bson.c
```diff
@@ -1573,12 +1555,18 @@ bson_append_regex_w_len (bson_t *bson,
...
 
-   if (key_length < 0) {
-      key_length = (int) strlen (key);
-   }
+   HANDLE_KEY_LENGTH (key, key_length);
 
    if (regex_length < 0) {
       regex_length = (int) strlen (regex);
+   } else {
+      /* Necessary to validate embedded NULL is not present in key. */
```
jmikola, line 1563: Likewise: `if (regex_length != strlen(regex)) return false;`

src/libbson/tests/json/bson_corpus/top-mongoc.json
```diff
@@ -0,0 +1,16 @@
+{
+    "description": "Top-level document validity - libmongoc-specific tests",
```
jmikola, line 2: Noted that this file will eventually be relocated for [CDRIVER-4016](https://jira.mongodb.org/browse/CDRIVER-4016).

jmikola (approved):

src/libbson/src/bson/bson-json.c
```diff
@@ -978,13 +978,28 @@ _bson_json_read_string (bson_json_reader_t *reader, /* IN */
...
+       * $regularExpression: { pattern: "pattern", options: "options" } } */
+      allow_null = false;
+   } else if (rs == BSON_JSON_IN_BSON_TYPE &&
+              (bs == BSON_JSON_LF_REGEX || bs == BSON_JSON_LF_OPTIONS)) {
+      /* Prohibit embedded NULL bytes for legacy regex: { $regex: "pattern",
+       * $options: "options" } */
+      allow_null = false;
+   }
+
+   if (!bson_utf8_validate ((const char *) val, vlen, allow_null)) {
```
vector-of-bool, line 1002: I always get worried about very complex boolean logic dances. Could this be extracted into a function that checks a `bson_json_reader_bson_t ` object specifically to calculate the `allow_null` value? i.e.

```c
static bool
_bson_json_allow_embedded_nulls (bson_json_reader_t const *reader)
{
   const bson_json_read_state_t read_state = reader->bson.read_state;
   const bson_json_read_bson_state_t bson_state = reader->bson.bson_state;

   if (read_state == BSON_JSON_IN_BSON_TYPE_REGEX_VALUES) {
      if (bson_state == BSON_JSON_LF_REGULAR_EXPRESSION_PATTERN ||
          bson_state == BSON_JSON_LF_REGULAR_EXPRESSION_OPTIONS) {
         /* Prohibit embedded NULL bytes for canonical extended regex:
          * { $regularExpression: { pattern: "pattern", options: "options" } }
          */
         return false;
      }
   }

   if (read_state == BSON_JSON_IN_BSON_TYPE) {
      if (bson_state == BSON_JSON_LF_REGEX |
          bson_state == BSON_JSON_LF_OPTIONS) {
         /* Prohibit embedded NULL bytes for legacy regex:
          * { $regex: "pattern", $options: "options" } */
         return false;
      }
   }

   /* Embedded nulls are okay in any other context */
   return true;
}
```

And then `allow_null` can be `const`-initialized:

```c
const bool allow_null = _bson_json_allow_embedded_nulls(reader);
```

or even just omitted entirely:

```c
if (!bson_utf8_validate((const char*)val,
                        vlen, 
                        _bson_json_allow_embedded_nulls(reader)) {
```

This is mostly just a stylistic choice.

src/libbson/src/bson/bson.c
vector-of-bool: I think that would assume that `key` is NULL-terminated, which it might not be.

---

Although I think moving this loop into a dedicated function, i.e.

```c
bool _string_contains_null(const char* str, size_t len) {
  for (; len; ++str, --len)
    if (*str == 0) return true;
  return false;
}
```

might be useful.

vector-of-bool (approved):
This code looks all correct. I just left two style-related comments.

src/libbson/tests/json/bson_corpus/top-mongoc.json
kevinAlbs: Thanks for the link! Updated the description to note CDRIVER-4016.

src/libbson/src/bson/bson.c
kevinAlbs: Done.

src/libbson/src/bson/bson.c
kevinAlbs: The check has been simplified to `if (key_length != strlen (key))`. I remember you mentioned this to me before @jmikola but this slipped my mind 🤦.

src/libbson/src/bson/bson-json.c
kevinAlbs: I like it, and thank you for the code snippet. Done.

PR #860, opened 2021-09-04
CDRIVER-3636 exclude PossiblePrimary from MSS check

kevinAlbs:
Do not check PossiblePrimary servers in the maxWireVersion check
for topology support of maxStalenessSeconds


eramongodb (approved):

chardan (approved):
+1, LGTM

PR #866, opened 2021-09-10
CDRIVER-4155 address deadStore warnings

kevinAlbs:


src/libbson/src/bson/bson-decimal128.c
```diff
@@ -709,8 +709,8 @@ bson_decimal128_from_string_w_len (const char *string,     /* IN */
    }
 
    /* Encode significand */
-   significand_high = 0,   /* The high 17 digits of the significand */
-      significand_low = 0; /* The low 17 digits of the significand */
+   significand_high = 0;   /* The high 17 digits of the significand */
+   /* significand_low is the low 17 digits of the significand */
```
eramongodb, line 713: Both `significant_high` and `significant_low` are already initialized to `0` on lines 503 and 504. Preserving the comments does not seem necessary either given similar the variable declarations are labeled with similar descriptions. Therefore, I believe lines 712-714 should be removed.

src/libmongoc/tests/test-mongoc-rpc.c
```diff
@@ -508,7 +508,7 @@ test_mongoc_rpc_reply_scatter2 (void)
    ASSERT (rpc.reply.documents_len == 16200);
    reader =
       bson_reader_new_from_data (rpc.reply.documents, rpc.reply.documents_len);
-   while ((b = bson_reader_read (reader, &eof))) {
+   while (bson_reader_read (reader, &eof)) {
```
eramongodb, line 511: Declaration of `const bson_t *b;` on line 484 should be removed as the variable is no longer used.

eramongodb (changes requested):

src/libbson/src/bson/bson-decimal128.c
kevinAlbs: Good point, done.

src/libmongoc/tests/test-mongoc-rpc.c
kevinAlbs: Done.

eramongodb (approved):

PR #867, opened 2021-09-12
CDRIVER-4156 use Python from toolchain on Ubuntu 14.04 Evergreen hosts

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/613df8d52a60ed7292b98b30/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

In the process of looking for specific failures on the waterfall, I stumbled across a pattern of failures on Ubuntu 14.04.  This  change addresses the reason for the failure: the system Python is too old and the helper script was pointing at the incorrect location for the toolchain-provided Python interpreter.  This corrects all of the `System Failed` tasks on Ubuntu 14.04 variants.  The majority turned green, but a few turned red as a result of legitimate test failures which are documented in existing tickets.

@kevinAlbs Note that I could not find a ticket that describes the particular failures I am correcting with this PR (I searched for tickets with the `failing-on-waterfall` tag).  If a ticket is needed, let me know and I can file the ticket and then amend the commit message to reference the new ticket.  Let me know if this is something I should do.

### DIFF
--- a/.evergreen/integration-tests.sh
+++ b/.evergreen/integration-tests.sh
@@ -117,6 +117,9 @@ case "$OS" in
       if [ -f /opt/python/2.7/bin/python ]; then
          # Python toolchain installation.
          PYTHON=/opt/python/2.7/bin/python
+      elif [ "x$(lsb_release -cs)" = "xtrusty" -a -f /opt/mongodbtoolchain/v2/bin/python ]; then
+         # Python toolchain installation.
+         PYTHON=/opt/mongodbtoolchain/v2/bin/python
       else
          PYTHON=python
       fi


kevinAlbs (approved):
Thank you for the fix! LGTM

> @kevinAlbs Note that I could not find a ticket that describes the particular failures I am correcting with this PR (I searched for tickets with the `failing-on-waterfall` tag). If a ticket is needed, let me know and I can file the ticket and then amend the commit message to reference the new ticket. Let me know if this is something I should do.

I do not see a ticket either. I think it would be good to backport this to the r1.19 branch so tests on the release branch can pass. I suggest filing a ticket so both commits can be linked to the same ticket.

acmorrow:
The `mongodbtoolchain` is for the server project, and not intended for use in other projects. Is the C driver (or C++ driver for that matter) already using `mongodbtoolchain` components?

rcsanchez97:
@acmorrow apart from the change I am proposing (which I think is not
introducing a new reference, but updating a stale reference), the C
driver contains toolchain references to:

`/opt/mongodbtoolchain/v3/bin/llvm-symbolizer` (in the test helper
and in the compilation helper scripts)

`/opt/mongodbtoolchain/v3/bin/clang` (in the TSAN build variant)

`/opt/mongodbtoolchain/v3/bin/python3` (in the OCSP build variant helper
script)

Is this something that needs to be changed?

acmorrow:
@rcsanchez97 - It's less than ideal for any project other than the server to use `mongodbtoolchain` because the only intended/acknowledged consumer is the server project. Since the server team is both the producer and the sole intended consumer, we basically consider ourselves free to change `mongodbtooolchain` if and when it suits the server team. Non-server projects that use it are exposed to the risk of it changing without notice, and we would be unlikely to revert changes that break things in non-server projects that have become dependent.

rcsanchez97:
@acmorrow thanks for the insight.  I have spoken with Kevin and I am
going to merge this PR (and also #868) so that we can address the system
failures now.  I will also write tickets so that we review and remove
any dependencies from the C driver and C++ driver to the server
toolchain.

PR #870, opened 2021-09-16
Allow CTest as a test driver

vector-of-bool:
Previously, a single CTest test was defined for test-libmongoc. This change allows CTest to see the individual test cases contained within test-libmongoc. These can be selected and executed with the CTest CLI, including with some parallelism.

Changes are only visible with CMake 3.10 or newer. Earlier CMake versions should be unaffected.

This changeset does not modify how the tests are executed in CI. In the future, if we wish to adopt CTest as a test runner, the CTest-emitted results could be transformed via simple scripting back into Evergreen test results.

CTest can be executed with parallelism with the `-j`/`--parallel` options. Tests execute well in parallel except when they attempt to modify a shared resource. This can be fixed by attaching concurrency properties to tests as appropriate, or by ensuring that tests each receive and use a unique copy of the resource under contention, but such changes are beyond the scope of this changeset.

These changes relate-to and partially-address CXX-2301 and CXX-2302, but do not complete them.

(The choice of `@@ctest-skipped@@` as a skip-marker is arbitrary and insignificant.)


chardan (approved):
+1. Bearing in mind my basic CMake know-how and not-solid understanding of how that command-line tooling goes, this looks good. Hopefully it will help us make many new tests! :-)

eramongodb (changes requested):
There are two tests that fail to be registered with CTest (shows up in `test-libmongoc --list-tests` but not in `ctest -N`):
```
/bson/oid/after_fork
/Client/exhaust_cursor/after_reset
```
Total tests registered should total 2212, but currently total 2210 according to `ctest -N`. I could not figure out why these two tests are being dropped from examining the current set of changes.

LoadTests.cmake
```diff
@@ -0,0 +1,47 @@
...
+    # Define the test. Use `--ctest-run` to tell it that CTest is in control.
+    add_test ("${test}" "${TEST_LIBMONGOC_EXE}" --ctest-run "${line}")
+    set_tests_properties ("${test}" PROPERTIES
+        # test-libmongoc expects to execute in the root of the source directory
+        WORKING_DIRECTORY "${SRC_ROOT}"
+        # If a test emits '@@ctest-skipped@@', this tells us that the test is
+        # not run, but did not fail.
+        SKIP_REGULAR_EXPRESSION "@@ctest-skipped@@"
+        # 15 seconds of timeout on each test.
+        TIMEOUT 15
```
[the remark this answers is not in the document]
vector-of-bool: Indeed, I've found one test that takes 20 seconds exactly.

src/libmongoc/tests/TestSuite.c
```diff
@@ -909,6 +937,14 @@ TestSuite_RunAll (TestSuite *suite /* IN */)
       }
    }
 
+   if (suite->ctest_run) {
+      /* We should have matched *at most* one test */
+      ASSERT (count <= 1);
```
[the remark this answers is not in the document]
vector-of-bool: Yes, and the double-registered tests is actually a mistake that this assertion catches. I have some changes that will address those.

LoadTests.cmake
```diff
@@ -0,0 +1,47 @@
...
+string (REPLACE "\n" ";" lines "${tests_out}")
+
+# Generate the test definitions
+foreach (line IN LISTS lines)
+    if (NOT line MATCHES "^/")
+        # Only generate if the line begins with `/`, which all tests should.
+        continue ()
+    endif ()
+    # The new test name is prefixed with 'mongoc'
+    set (test "mongoc/${CONFIG}${line}")
```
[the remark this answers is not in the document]
vector-of-bool: Qualifying them by the configuration is required in the case of multiconf builds, in which case tests can be registered for each build type simultaneously. CTest with multiconf builds is a bit goofy. Multiconf builds are goofy in general.

src/libmongoc/tests/TestSuite.h
```diff
@@ -100,6 +100,23 @@ bson_open (const char *filename, int flags, ...)
 #define CERT_PASSWORD_PROTECTED CERT_TEST_DIR "/password_protected.pem"
 
 
+#ifdef ASSERT
+#undef ASSERT
+#endif
+#define ASSERT(Cond)                                              \
```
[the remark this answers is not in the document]
vector-of-bool: That may be a better design choice.

CMakeLists.txt
```diff
@@ -485,3 +488,30 @@ if (CMAKE_GENERATOR_TOOLSET)
...
+   ]=] code @ONLY)
+   file (GENERATE
+         OUTPUT "${PROJECT_BINARY_DIR}/LoadTests-inc-$<CONFIG>.cmake"
+         CONTENT "${code}")
+   file (WRITE "${PROJECT_BINARY_DIR}/LoadTests-inc.cmake" [[
+      file (GLOB files "./LoadTests-inc-*.cmake")
+      foreach (inc IN LISTS files)
+         include ("${inc}")
+      endforeach ()
+   ]])
```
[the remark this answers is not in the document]
vector-of-bool: I considered this goof. I'll have to address it. Unfortunately, multiconf support is required for any Visual Studio generators, not just the Ninja one.

rcsanchez97 (approved):
LGTM

src/libmongoc/tests/TestSuite.c
```diff
@@ -165,6 +166,16 @@ TestSuite_Init (TestSuite *suite, const char *name, int argc, char **argv)
...
                  (0 == strcmp ("--silent", argv[i]))) {
          suite->silent = true;
+      } else if ((0 == strcmp ("--ctest-run", argv[i]))) {
+         if (suite->ctest_run) {
+            test_error ("'--ctest-run' can only be specified once");
+         }
+         if (argc - 1 == 0) {
+            test_error ("'--ctest-run' requires an argument");
+         }
+         suite->flags |= TEST_NOFORK;
```
kevinAlbs, line 176: To check, ctest runs each test within a separate process to isolate, right? I assume this is necessary to avoid an extra process spawn, or for ctest to capture output.

src/libmongoc/tests/test-mongoc-server-selection.c
```diff
@@ -63,10 +63,6 @@ test_all_spec_tests (TestSuite *suite)
    test_framework_resolve_path (JSON_DIR "/server_selection/server_selection",
                                 resolved);
    install_json_test_suite (suite, resolved, &test_server_selection_logic_cb);
-
-   test_framework_resolve_path (
-      JSON_DIR "/server_selection/server_selection/LoadBalanced", resolved);
-   install_json_test_suite (suite, resolved, &test_server_selection_logic_cb);
```
kevinAlbs, line 69: Ah, it looks like this is resolved on L63 since `install_json_test_suite` recurses.

kevinAlbs (approved):
I have not used ctest much. But it looks like an improvement over using test-libmongoc.

The `-I` option lets me specify an indexed range of tests to run. I have hacked test-libmongoc to support that when running a sequence of time consuming tests during development.

I like the consolidated output of skipped tests.

LGTM

src/libmongoc/tests/TestSuite.c
eramongodb: Yes. Similarly to how `test-libmongoc` (without `-f`) spawns a given test in its own process (via `fork ()`) to handle process termination and capture output without interrupting the runner, ctest does the same for each registered test case.

LoadTests.cmake
```diff
@@ -0,0 +1,47 @@
+# This file is include()'d by CTest. It executes test-libmongoc to get a list
+# of all tests that are registered. Each test is then defined as a CTest test,
+# allowing CTest to control the execution, parallelization, and collection of
+# test results.
```
eramongodb, line 4: This file feels out of place being in project root directory. Should it be moved into `build/cmake`?

vector-of-bool: Probably a better place, yes.

LoadTests.cmake
```diff
@@ -0,0 +1,47 @@
...
+    # The new test name is prefixed with 'mongoc'
+    set (test "mongoc/${CONFIG}${line}")
+    # Define the test. Use `--ctest-run` to tell it that CTest is in control.
+    add_test ("${test}" "${TEST_LIBMONGOC_EXE}" --ctest-run "${line}")
+    set_tests_properties ("${test}" PROPERTIES
+        # test-libmongoc expects to execute in the root of the source directory
+        WORKING_DIRECTORY "${SRC_ROOT}"
+        # If a test emits '@@ctest-skipped@@', this tells us that the test is
+        # skipped.
+        SKIP_REGULAR_EXPRESSION "@@ctest-skipped@@"
```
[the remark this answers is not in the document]
vector-of-bool: These changes are currently only concerned with dev workflow. Older CMake will just see it as a passed test.

CMakeLists.txt
```diff
@@ -57,7 +57,7 @@ project (mongo-c-driver C)
 # Optionally enable C++ to do some C++-specific tests
 enable_language (CXX OPTIONAL)
 
-if (NOT CMAKE_BUILD_TYPE)
+if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
```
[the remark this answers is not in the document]
vector-of-bool: Specifying `CMAKE_BUILD_TYPE` while using a multi-config generator doesn't change the behavior of CMake, but is Almost Always Wrong™, not in that it changes behavior of CMake (it doesn't), but because other code starts to assume that it is meaningful. Writing code that works on both single-config and multi-config is somewhat tricky.

src/libmongoc/tests/TestSuite.c
vector-of-bool: This is correct.

LoadTests.cmake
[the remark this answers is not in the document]
vector-of-bool: You made me curious, as VS's `RUN_TESTS` "target" isn't something I have inspected before. It looks to be that `RUN_TEST` simply runs CTest with `--build-config $(Configuration)`, which will match what was specified in VS or MSBuild. If you try to run `ctest` in a VS-generated build directory without specifying the build type, CTest will not attempt to define any of the tests that are based on multi-conf generated executables, and will instead generate invalid test definitions. I've updated the code to mimic this behavior with multi-conf generators and require a `-C` to be specified. This will thus only register each test once, and we can omit the build-type qualifier from the names. (Previously it would try to register a test for each available config, thus needing the qualified names).

Aside: The `-R` argument takes a regex, not a glob, and it searches for a submatch. `-R '/bson/*` only just happens to also match the same tests 🙂. I actually find this a bit easier since I can just write a sub-string of the test name to `-R`, no quoting required.

---

Not really relevant here, but based on how Catch2 does it, I hypothesize that it actually contains a subtle bug in an uncommon situation where you build multiple configurations simultaneously (which made possible with the `Ninja Multi-Config` generator). The test registration runs as a post-build step on each target, and writes those tests to a file. The file it generates is not qualified by the build type, so whichever configuration was the last to build will end up writing the test registration file. In most situations this won't matter, but in a simultaneous multi-conf build each configuration will end up racing to write the test listing, and the last one to execute will end up being the one that it registered.

But that's just a theory: a _CTest_ theory.

build/cmake/LoadTests.cmake
```diff
@@ -0,0 +1,48 @@
...
+string (REPLACE "\n" ";" lines "${tests_out}")
+
+# Generate the test definitions
+foreach (line IN LISTS lines)
+    if (NOT line MATCHES "^/")
+        # Only generate if the line begins with `/`, which all tests should.
+        continue ()
+    endif ()
+    # The new test name is prefixed with 'mongoc'
+    set (test "mongoc${line}")
```
eramongodb, line 36:
```suggestion
    set (test "${line}")
```
Note: I still do not see the need to break symmetry with `test-libmongoc` by adding a prefix to test names, especially given that distinct test names with the configuration included are proven no longer necessary. I would prefer the `"mongoc"` prefix be removed as well.

src/libmongoc/tests/TestSuite.c
```diff
@@ -125,8 +162,14 @@ TestSuite_Init (TestSuite *suite, const char *name, int argc, char **argv)
    suite->flags = 0;
    suite->prgname = bson_strdup (argv[0]);
    suite->silent = false;
+   suite->ctest_run = NULL;
    _mongoc_array_init (&suite->match_patterns, sizeof (char *));
 
+   suite->prev_sigabrt = signal (SIGABRT, _handle_signal);
+   suite->prev_sigint = signal (SIGINT, _handle_signal);
+   suite->prev_sigterm = signal (SIGTERM, _handle_signal);
+   suite->prev_sigsegv = signal (SIGSEGV, _handle_signal);
```
eramongodb, line 171: Is registering signal handlers for `SIGINT`, `SIGTERM`, and `SIGSEGV` necessary? This goes beyond the initial suggestion to use a signal handler to replace existing calls to `abort ()` with `exit (1)` behavior, which only applies to the `SIGABRT` signal.

src/libmongoc/tests/TestSuite.c
```diff
@@ -112,6 +113,42 @@ static BSON_ONCE_FUN (_test_suite_ensure_mutex_once)
...
+   write (STDERR_FILENO, s, strlen (s));
+   write (STDERR_FILENO, "  ", 2);
+   write (STDERR_FILENO, sigstr, strlen (sigstr));
+   fsync (STDERR_FILENO);
+#else
+   /* On Windows these APIs are signal-safe */
+   fprintf (stderr, "\n%s\n  %s\n", s, sigstr);
+   fflush (stderr);
+#endif
+   _Exit (signum);
```
eramongodb, line 148:
```suggestion
   exit (signum);
```
What is the motivation for explicitly calling file flush functions and `_Exit ()` rather than delegating to `exit ()`?

eramongodb (approved):
Other than a couple remaining questions, LGTM.

Output of attempts to run tests when none are registered (e.g. built `Debug`, attempt to run `Release` tests) can be rather noisy, but I prefer the noise over little/no feedback. Great work getting both single and multi-config generators to work properly. I hope this paves the way to a parallel-friendly test suite. 👍

src/libmongoc/tests/TestSuite.c
vector-of-bool: It isn't *necessary*, but it is helpful to see why a particular test is exiting abnormally.

src/libmongoc/tests/TestSuite.c
vector-of-bool: `exit()` is not signal-safe, but `_Exit()` is signal-safe. `quick_exit()` is also signal-safe (and preferable), but isn't available until C11. (`exit()` will execute arbitrary code that might not be safe in the current process condition e.g. there's an `assert()` that failed and the program invariants are broken, or `SIGINT` might be delivered in the middle of a code block that is holding a global lock). It would be more sophisticated and complete to restore the prior signal handlers and re-`raise()`, but I was experiencing a lot of trouble getting that to work and it isn't yet needed in our test suite.

build/cmake/LoadTests.cmake
vector-of-bool: At least in my VSCode setup, the leading `/` adds an extraneous blank node at the root of the test suite tree. This isn't a blocking issue, but it is confusing to look at, and is mostly aesthetic.

src/libmongoc/tests/TestSuite.c
eramongodb: It seems `exit()` is signal-safe, but not _async_-signal-safe (in large part due to the invocation of an arbitrary set of `atexit` functions as you described). Regardless, the reasons you have given are justified.

CMakeLists.txt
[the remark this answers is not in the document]
kevinAlbs: Thank you for the explanation.

kevinAlbs (approved):
LGTM

PR #876, opened 2021-09-19
CDRIVER-4035 use runAdminCommand.arguments.readPreference

kevinAlbs:
Fixes the test failure associated with the [rediscover-quickly-after-stepdown](https://github.com/mongodb/specifications/blob/c48d7f416e51dce8fd4c8d05bdb05790508ca3bb/source/server-discovery-and-monitoring/tests/integration/rediscover-quickly-after-step-down.yml#L34-L42) SDAM specification test.

The test runner was not passing a read preference. No read preference defaults to a primary read preference. The test operation requires sending to a secondary.


vector-of-bool (approved):
A simple change. LGTM.

PR #877, opened 2021-09-27
CDRIVER-4166 permit NULL platform argument to mongoc_handshake_data_append()

eramongodb:
`mongoc_handshake_data_append()` [special-cases](https://github.com/mongodb/mongo-c-driver/blob/130938ad44a6516e6c399b9d25e5fe1b0464f14f/src/libmongoc/src/mongoc/mongoc-handshake.c#L638) the first write to `mongoc_handshake_t::platform` to avoid inserting a leading delimiter. This special-case branch does not protect against a `NULL` value for the `platform` parameter; other branches use the [`NULL` check](https://github.com/mongodb/mongo-c-driver/blob/130938ad44a6516e6c399b9d25e5fe1b0464f14f/src/libmongoc/src/mongoc/mongoc-handshake.c#L590) within the `_append_and_truncate()` helper function. However, this bug was likely not detected due to both `vsnprintf()` (Unix/macOS) and `_vsnprintf_s()` (Windows), [used internally](https://github.com/mongodb/mongo-c-driver/blob/130938ad44a6516e6c399b9d25e5fe1b0464f14f/src/libbson/src/bson/bson-string.c#L585) by `bson_strdup_printf()`, outputting `"(null)"` given a null argument corresponding to a `"%s"` string specifier across libstdc++, libc++, and STL. This is unfortunate but permitted given this condition falls under undefined behavior:

```c
#include <stdio.h>

int main(void) {
  printf("%s\n", NULL);
}
```
```
(null)
```

This PR therefore:
1. makes the optional nature of parameters explicit in `mongoc_handshake_data_append()` by lifting the `NULL` check out of `_append_and_truncate()`,
2. ensures all branches, including the special-case mentioned above, are protected by a corresponding null check, and
3. adds a new test `/MongoDB/handshake/null_args` explicitly asserting for the absense of a `"null"` string in the result given a `NULL` platform argument.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-handshake.c
+++ b/src/libmongoc/src/mongoc/mongoc-handshake.c
@@ -572,7 +572,6 @@ _mongoc_handshake_freeze (void)
 /*
  * free (*s) and make *s point to *s concated with suffix.
  * If *s is NULL it's treated like it's an empty string.
- * If suffix is NULL, nothing happens.
  */
 static void
 _append_and_truncate (char **s, const char *suffix, int max_len)
@@ -582,13 +581,10 @@ _append_and_truncate (char **s, const char *suffix, int max_len)
    const int delim_len = (int) strlen (" / ");
    int space_for_suffix;
 
-   BSON_ASSERT (s);
+   BSON_ASSERT_PARAM (s);
+   BSON_ASSERT_PARAM (suffix);
 
    prefix = old_str ? old_str : "";
-
-   if (!suffix) {
-      return;
-   }
 
    space_for_suffix = max_len - (int) strlen (prefix) - delim_len;
 
@@ -627,29 +623,37 @@ mongoc_handshake_data_append (const char *driver_name,
       return false;
    }
 
+   BSON_ASSERT (_mongoc_handshake_get ()->platform);
+
    /* allow practically any size for "platform", we'll trim it down in
     * _mongoc_handshake_build_doc_with_application */
    platform_space =
       HANDSHAKE_MAX_SIZE - (int) strlen (_mongoc_handshake_get ()->platform);
 
-   /* we check for an empty string as a special case to avoid an unnecessary
-    * delimiter being added in front of the string by _append_and_truncate */
-   if (strcmp (_mongoc_handshake_get ()->platform, "") == 0) {
-      bson_free (_mongoc_handshake_get ()->platform);
-      _mongoc_handshake_get ()->platform =
-         bson_strdup_printf ("%.*s", platform_space, platform);
-   } else {
-      _append_and_truncate (
-         &_mongoc_handshake_get ()->platform, platform, HANDSHAKE_MAX_SIZE);
+   if (platform) {
+      /* we check for an empty string as a special case to avoid an unnecessary
+       * delimiter being added in front of the string by _append_and_truncate */
+      if (strcmp (_mongoc_handshake_get ()->platform, "") == 0) {
+         bson_free (_mongoc_handshake_get ()->platform);
+         _mongoc_handshake_get ()->platform =
+            bson_strdup_printf ("%.*s", platform_space, platform);
+      } else {
+         _append_and_truncate (
+            &_mongoc_handshake_get ()->platform, platform, HANDSHAKE_MAX_SIZE);
+      }
    }
 
-   _append_and_truncate (&_mongoc_handshake_get ()->driver_name,
-                         driver_name,
-                         HANDSHAKE_DRIVER_NAME_MAX);
+   if (driver_name) {
+      _append_and_truncate (&_mongoc_handshake_get ()->driver_name,
+                            driver_name,
+                            HANDSHAKE_DRIVER_NAME_MAX);
+   }
 
-   _append_and_truncate (&_mongoc_handshake_get ()->driver_version,
-                         driver_version,
-                         HANDSHAKE_DRIVER_VERSION_MAX);
+   if (driver_version) {
+      _append_and_truncate (&_mongoc_handshake_get ()->driver_version,
+                            driver_version,
+                            HANDSHAKE_DRIVER_VERSION_MAX);
+   }
 
    _mongoc_handshake_freeze ();
    bson_mutex_unlock (&gHandshakeLock);
--- a/src/libmongoc/src/mongoc/mongoc-handshake.c
+++ b/src/libmongoc/src/mongoc/mongoc-handshake.c
@@ -633,7 +633,7 @@ mongoc_handshake_data_append (const char *driver_name,
    if (platform) {
       /* we check for an empty string as a special case to avoid an unnecessary
        * delimiter being added in front of the string by _append_and_truncate */
-      if (strcmp (_mongoc_handshake_get ()->platform, "") == 0) {
+      if (_mongoc_handshake_get()->platform[0] == '\0') {
          bson_free (_mongoc_handshake_get ()->platform);
          _mongoc_handshake_get ()->platform =
             bson_strdup_printf ("%.*s", platform_space, platform);
--- a/src/libmongoc/src/mongoc/mongoc-handshake.c
+++ b/src/libmongoc/src/mongoc/mongoc-handshake.c
@@ -473,8 +473,7 @@ _append_platform_field (bson_t *doc, const char *platform)
                             1 +
 
                             /* key size */
-                            (int) strlen (HANDSHAKE_PLATFORM_FIELD) +
-                            1 +
+                            (int) strlen (HANDSHAKE_PLATFORM_FIELD) + 1 +
 
                             /* 4 bytes for length of string */
                             4);
--- a/src/libmongoc/src/mongoc/mongoc-handshake.c
+++ b/src/libmongoc/src/mongoc/mongoc-handshake.c
@@ -633,7 +633,7 @@ mongoc_handshake_data_append (const char *driver_name,
    if (platform) {
       /* we check for an empty string as a special case to avoid an unnecessary
        * delimiter being added in front of the string by _append_and_truncate */
-      if (_mongoc_handshake_get()->platform[0] == '\0') {
+      if (_mongoc_handshake_get ()->platform[0] == '\0') {
          bson_free (_mongoc_handshake_get ()->platform);
          _mongoc_handshake_get ()->platform =
             bson_strdup_printf ("%.*s", platform_space, platform);
--- a/src/libmongoc/tests/test-mongoc-handshake.c
+++ b/src/libmongoc/tests/test-mongoc-handshake.c
@@ -182,7 +182,6 @@ test_mongoc_handshake_data_append_success (void)
    server = mock_server_new ();
    mock_server_run (server);
    uri = mongoc_uri_copy (mock_server_get_uri (server));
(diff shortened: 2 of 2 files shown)


src/libmongoc/src/mongoc/mongoc-handshake.c
```diff
@@ -628,29 +623,37 @@ mongoc_handshake_data_append (const char *driver_name,
       return false;
    }
 
+   BSON_ASSERT (_mongoc_handshake_get ()->platform);
+
    /* allow practically any size for "platform", we'll trim it down in
     * _mongoc_handshake_build_doc_with_application */
    platform_space =
       HANDSHAKE_MAX_SIZE - (int) strlen (_mongoc_handshake_get ()->platform);
```
chardan, line 631: note: I think this is correct, since /at most/ one NULL character is embedded in the string at a time.

src/libmongoc/src/mongoc/mongoc-handshake.c
```diff
@@ -628,29 +623,37 @@ mongoc_handshake_data_append (const char *driver_name,
...
-      bson_free (_mongoc_handshake_get ()->platform);
-      _mongoc_handshake_get ()->platform =
-         bson_strdup_printf ("%.*s", platform_space, platform);
-   } else {
-      _append_and_truncate (
-         &_mongoc_handshake_get ()->platform, platform, HANDSHAKE_MAX_SIZE);
+   if (platform) {
+      /* we check for an empty string as a special case to avoid an unnecessary
+       * delimiter being added in front of the string by _append_and_truncate */
+      if (strcmp (_mongoc_handshake_get ()->platform, "") == 0) {
```
chardan, line 636: note: I'm ok with this, although I wonder if it might have been more immediately clear to see a direct check for the first byte being NULL, rather than strcmp().

src/libmongoc/src/mongoc/mongoc-handshake.c
```diff
@@ -628,29 +623,37 @@ mongoc_handshake_data_append (const char *driver_name,
       return false;
    }
 
+   BSON_ASSERT (_mongoc_handshake_get ()->platform);
+
    /* allow practically any size for "platform", we'll trim it down in
```
chardan, line 628: Ok, it looks like this is idiomatic WRT how handshakes are assembled, and explains the frequent bson_free() calls, rather than just overwriting into already-allocated memory.

src/libmongoc/src/mongoc/mongoc-handshake.c
```diff
@@ -628,29 +623,37 @@ mongoc_handshake_data_append (const char *driver_name,
...
+   if (platform) {
+      /* we check for an empty string as a special case to avoid an unnecessary
+       * delimiter being added in front of the string by _append_and_truncate */
+      if (strcmp (_mongoc_handshake_get ()->platform, "") == 0) {
+         bson_free (_mongoc_handshake_get ()->platform);
+         _mongoc_handshake_get ()->platform =
+            bson_strdup_printf ("%.*s", platform_space, platform);
+      } else {
+         _append_and_truncate (
+            &_mongoc_handshake_get ()->platform, platform, HANDSHAKE_MAX_SIZE);
```
chardan, line 642: Ok-- I think HANDSHAKE_MAX_SIZE is correct. Other functions, like _append_platform_field(), go through some pains to adjust for the length of the other fields, but that's because they're dealing with individual fields and this is dealing with the whole handshake.

src/libmongoc/tests/test-mongoc-handshake.c
```diff
@@ -276,6 +276,122 @@ test_mongoc_handshake_data_append_success (void)
...
+   mongoc_client_t *client;
+   mongoc_client_pool_t *pool;
+   request_t *request;
+   const bson_t *request_doc;
+   bson_iter_t iter;
+   bson_iter_t md_iter;
+   bson_iter_t inner_iter;
+   const char *val;
+
+   char big_string[HANDSHAKE_MAX_SIZE];
```
kevinAlbs, line 293: `big_string` is not used. I suggest removing the declaration and L295-296.

src/libmongoc/tests/test-mongoc-handshake.c
```diff
@@ -276,6 +276,122 @@ test_mongoc_handshake_data_append_success (void)
...
+   big_string[HANDSHAKE_MAX_SIZE - 1] = '\0';
+
+   _reset_handshake ();
+   /* Make sure setting the handshake works */
+   ASSERT (mongoc_handshake_data_append (NULL, NULL, NULL));
+
+   server = mock_server_new ();
+   mock_server_run (server);
+   uri = mongoc_uri_copy (mock_server_get_uri (server));
+   mongoc_uri_set_option_as_int32 (uri, MONGOC_URI_HEARTBEATFREQUENCYMS, 500);
```
kevinAlbs, line 305: I suggest removing L305. I see `MONGOC_URI_HEARTBEATFREQUENCYMS` is set in `test_mongoc_handshake_data_append_success`. `test_mongoc_handshake_data_append_success` looks like the template of this test. But setting `MONGOC_URI_HEARTBEATFREQUENCYMS` is not necessary.

The test needs to observe the MongoDB handshake from server monitoring. `MONGOC_URI_HEARTBEATFREQUENCYMS` determines the frequency server monitoring checks. But the first check will occur shortly after the call to `mongoc_client_pool_pop`.

src/libmongoc/tests/test-mongoc-handshake.c
```diff
@@ -276,6 +276,122 @@ test_mongoc_handshake_data_append_success (void)
...
+   request_t *request;
+   const bson_t *request_doc;
+   bson_iter_t iter;
+   bson_iter_t md_iter;
+   bson_iter_t inner_iter;
+   const char *val;
+
+   char big_string[HANDSHAKE_MAX_SIZE];
+
+   memset (big_string, 'a', HANDSHAKE_MAX_SIZE - 1);
```
chardan, line 295: note: Ok, fill /up to/ the last byte, then the next line, then NULL terminate. (I'm not sure if newer C standards have nicer ways to do this; I believe in C99 you can use a designated initializer for the last byte, and get away with something like:
big_string[HANDSHAKE_MAX_SIZE] = { [0... HANDSHAKE_MAX_SIZE - 1] 'a', NULL };
...but I'm not entirely sure if we're actually supporting C99 yet.

src/libmongoc/tests/test-mongoc-handshake.c
```diff
@@ -276,6 +276,122 @@ test_mongoc_handshake_data_append_success (void)
...
+   big_string[HANDSHAKE_MAX_SIZE - 1] = '\0';
+
+   _reset_handshake ();
+   /* Make sure setting the handshake works */
+   ASSERT (mongoc_handshake_data_append (NULL, NULL, NULL));
```
chardan, line 300: Ok. This works because the function is operating on the global context.

kevinAlbs (approved):
LGTM with small tweaks to the test!

src/libmongoc/tests/test-mongoc-handshake.c
```diff
@@ -276,6 +276,122 @@ test_mongoc_handshake_data_append_success (void)
...
+   const bson_t *request_doc;
+   bson_iter_t iter;
+   bson_iter_t md_iter;
+   bson_iter_t inner_iter;
+   const char *val;
+
+   char big_string[HANDSHAKE_MAX_SIZE];
+
+   memset (big_string, 'a', HANDSHAKE_MAX_SIZE - 1);
+   big_string[HANDSHAKE_MAX_SIZE - 1] = '\0';
```
chardan, line 296: ...is big_string ever used?

src/libmongoc/tests/test-mongoc-handshake.c
eramongodb: Should the line be removed from `test_mongoc_handshake_data_append_success()` as well?

src/libmongoc/tests/test-mongoc-handshake.c
eramongodb: Good catch, thank you.

src/libmongoc/src/mongoc/mongoc-handshake.c
```diff
@@ -628,29 +623,37 @@ mongoc_handshake_data_append (const char *driver_name,
...
+      }
    }
 
-   _append_and_truncate (&_mongoc_handshake_get ()->driver_name,
-                         driver_name,
-                         HANDSHAKE_DRIVER_NAME_MAX);
+   if (driver_name) {
```
chardan, line 646: ...just to make sure I understand the intent: we're only worried about NULL embedding in the "platform" field, and not in any of the others?

chardan (changes requested):
If I'm following along with the plot, I think this is ok, with maybe a cleanup or two, see notes. Thanks for doing this!

src/libmongoc/src/mongoc/mongoc-handshake.c
eramongodb: I considered this transformation as well. Since you pointed it out, I think I will include it in this PR (might as well).

src/libmongoc/src/mongoc/mongoc-handshake.c
eramongodb: `_append_and_truncate()` previously handled `NULL` arguments for `driver_name` and `driver_version` correctly. The only branch that was not being null-tested correctly is where `bson_strdup_printf()` is called directly. All three parameters are optional and must be null-checked; elevating the null check out of `_append_and_truncate()` was in favor of making this more apparent within the function definition.

src/libmongoc/tests/test-mongoc-handshake.c
eramongodb: I have not looked too deeply into the contents of this test function (which was modelled after the existing `test_mongoc_handshake_data_append_success()` test function).

src/libmongoc/tests/test-mongoc-handshake.c
eramongodb: It is not. Addressed in comment by @kevinAlbs.

src/libmongoc/src/mongoc/mongoc-handshake.c
```diff
@@ -628,29 +623,37 @@ mongoc_handshake_data_append (const char *driver_name,
...
-   } else {
-      _append_and_truncate (
-         &_mongoc_handshake_get ()->platform, platform, HANDSHAKE_MAX_SIZE);
+   if (platform) {
+      /* we check for an empty string as a special case to avoid an unnecessary
+       * delimiter being added in front of the string by _append_and_truncate */
+      if (strcmp (_mongoc_handshake_get ()->platform, "") == 0) {
+         bson_free (_mongoc_handshake_get ()->platform);
+         _mongoc_handshake_get ()->platform =
+            bson_strdup_printf ("%.*s", platform_space, platform);
```
chardan, line 639: note: Ok. The unchecked re-allocation makes my spine tingle, however this winds up indirectly relying on bson_malloc0(), which aborts the program on allocation failure.

kevinAlbs (approved):

chardan (approved):
+1

PR #879, opened 2021-10-04
CDRIVER-3927 add KMS TLS tests for client side encryption

eramongodb:
This PR comes in three parts:

1. Ensure that all four SSL libraries (OpenSSL, Secure Channel, Secure Transport, LibreSSL) report descriptive errors for each certificate verification failure.

The TLS handshake implementation for OpenSSL and Secure Channel require some modification to report descriptive certificate verification errors. The TLS handshake implementation for Secure Transport and LibreSSL do not require any modification to obtain descriptive error messages.

2. Add tests that verify correct error (or lack thereof) against mock KMS servers.

Client Side Encryption tests were unconditionally being included by the test suite even when SSL is not enabled, despite CSE [requiring SSL to be enabled](https://github.com/mongodb/mongo-c-driver/blob/a4c8c66f3638e59a6acc6733dc8e36b71062fcca/src/libmongoc/CMakeLists.txt#L414). Therefore, the relevant source file and test functions are now moved into appropriate conditionals so they are only built and run when `MONGOC_ENABLE_SSL` is defined by the build system.

Existing unit test helper functions are modified to enable reuse of AWS KMS provider setup helpers for KMS TLS tests separately from other KMS providers. The test function `test_kms_tls_cert_valid()`, which confirms a valid TLS handshake on valid server certificate, is not required by CDRIVER-3927, but nevertheless useful to confirm proper CA certificate registration in the environment. It explicitly calls `mongoc_stream_tls_handshake_block()` and relevant setup functions rather `mongoc_client_encryption_create_datakey()` to avoid unnecessary errors on datakey creation due to running against the mock KMS server.

3. Enable KMS TLS tests on Evergreen.

Evergreen config scripts are updated to start mock KMS servers in the background only for tasks with Client Side Encryption enabled. The run_tests.sh script also requires an update to ensure proper CA certificate registration on Windows and Red Hat Linux, which use different commands and directories.

### DIFF
--- a/.evergreen/run-tests.sh
+++ b/.evergreen/run-tests.sh
@@ -37,11 +37,20 @@ if [ "$SSL" != "nossl" ]; then
          certutil.exe -addstore "Root" "src\libmongoc\tests\x509gen\ca.pem"
          ;;
       *)
-         sudo cp -v src/libmongoc/tests/x509gen/ca.pem /usr/local/share/ca-certificates/cdriver.crt || true
-         if [ -f /usr/local/share/ca-certificates/cdriver.crt ]; then
-            sudo update-ca-certificates
+         if [ -f /etc/redhat-release ]; then
+            sudo cp -v src/libmongoc/tests/x509gen/ca.pem /usr/share/pki/ca-trust-source/anchors/cdriver.crt || true
+            if [ -f /usr/share/pki/ca-trust-source/anchors/cdriver.crt ]; then
+               sudo update-ca-trust
+            else
+               export MONGOC_TEST_SSL_CA_FILE="src/libmongoc/tests/x509gen/ca.pem"
+            fi
          else
-            export MONGOC_TEST_SSL_CA_FILE="src/libmongoc/tests/x509gen/ca.pem"
+            sudo cp -v src/libmongoc/tests/x509gen/ca.pem /usr/local/share/ca-certificates/cdriver.crt || true
+            if [ -f /usr/local/share/ca-certificates/cdriver.crt ]; then
+               sudo update-ca-certificates
+            else
+               export MONGOC_TEST_SSL_CA_FILE="src/libmongoc/tests/x509gen/ca.pem"
+            fi
          fi
          ;;
    esac
--- a/.evergreen/run-tests.sh
+++ b/.evergreen/run-tests.sh
@@ -31,12 +31,20 @@ fi
 if [ "$SSL" != "nossl" ]; then
    export MONGOC_TEST_SSL_WEAK_CERT_VALIDATION="off"
    export MONGOC_TEST_SSL_PEM_FILE="src/libmongoc/tests/x509gen/client.pem"
-   sudo cp src/libmongoc/tests/x509gen/ca.pem /usr/local/share/ca-certificates/cdriver.crt || true
-   if [ -f /usr/local/share/ca-certificates/cdriver.crt ]; then
-      sudo update-ca-certificates
-   else
-      export MONGOC_TEST_SSL_CA_FILE="src/libmongoc/tests/x509gen/ca.pem"
-   fi
+
+   case "$OS" in
+      cygwin*)
+         certutil.exe -addstore "Root" "src\libmongoc\tests\x509gen\ca.pem"
+         ;;
+      *)
+         sudo cp -v src/libmongoc/tests/x509gen/ca.pem /usr/local/share/ca-certificates/cdriver.crt || true
+         if [ -f /usr/local/share/ca-certificates/cdriver.crt ]; then
+            sudo update-ca-certificates
+         else
+            export MONGOC_TEST_SSL_CA_FILE="src/libmongoc/tests/x509gen/ca.pem"
+         fi
+         ;;
+   esac
 fi
 
 export MONGOC_ENABLE_MAJORITY_READ_CONCERN=on
--- a/src/libmongoc/src/mongoc/mongoc-secure-channel.c
+++ b/src/libmongoc/src/mongoc/mongoc-secure-channel.c
@@ -111,7 +111,7 @@ mongoc_secure_channel_setup_certificate_from_file (const char *filename)
       NULL, /* phCertStore, OUT, HCERTSTORE.., unused, for now */
       NULL, /* phMsg, OUT, HCRYPTMSG, only for PKC7, unused */
       (const void **) &cert /* ppvContext, OUT, the Certificate Context */
-      );
+   );
 
    if (!cert) {
       MONGOC_ERROR ("Failed to extract public key from '%s'. Error 0x%.8X",
@@ -429,7 +429,7 @@ mongoc_secure_channel_setup_crl (
       NULL, /* phCertStore, OUT, HCERTSTORE.., unused, for now */
       NULL, /* phMsg, OUT, HCRYPTMSG, only for PKC7, unused */
       (const void **) &cert /* ppvContext, OUT, the Certificate Context */
-      );
+   );
    bson_free (str);
 
    if (!cert) {
@@ -609,7 +609,7 @@ mongoc_secure_channel_handshake_step_1 (mongoc_stream_tls_t *tls,
       &outbuf_desc,                       /* pOutput OUT param */
       &secure_channel->ret_flags,         /* pfContextAttr OUT param */
       &secure_channel->ctxt->time_stamp   /* ptsExpiry OUT param */
-      );
+   );
 
    if (sspi_status != SEC_I_CONTINUE_NEEDED) {
       MONGOC_ERROR ("initial InitializeSecurityContext failed: %ld",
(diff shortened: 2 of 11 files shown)


src/libmongoc/src/mongoc/mongoc-secure-channel.c
```diff
@@ -569,9 +570,16 @@ _mongoc_secure_channel_init_sec_buffer_desc (SecBufferDesc *desc,
 }
 
 
+#define MONGOC_LOG_AND_SET_ERROR(ERROR, DOMAIN, CODE, ...) \
+   do {                                                    \
+      MONGOC_ERROR (__VA_ARGS__);                          \
+      bson_set_error (ERROR, DOMAIN, CODE, __VA_ARGS__);   \
+   } while (0)
```
kevinAlbs, line 577: Oh nice! This seems much better than logging and returning a generic error.

src/libmongoc/src/mongoc/mongoc-secure-channel.c
```diff
@@ -695,8 +710,12 @@ mongoc_secure_channel_handshake_step_2 (mongoc_stream_tls_t *tls,
                return true;
             }
 
-            MONGOC_ERROR (
+            MONGOC_LOG_AND_SET_ERROR (
+               error,
+               MONGOC_ERROR_STREAM,
+               MONGOC_ERROR_STREAM_SOCKET,
                "failed to receive handshake, SSL/TLS connection failed");
+
```
kevinAlbs, line 718: Optional: If the expectation is that these `mongoc_secure_channel_handshake_step_2` sets `error` when returning false, I suggest adding a new error message to L684. IIUC that indicates a logic error in the caller. So perhaps the message could be something like:

```
MONGOC_LOG_AND_SET_ERROR (error,
                          MONGOC_ERROR_STREAM,
                          MONGOC_ERROR_STREAM_SOCKET,
                          "Invalid state: credentials or context unset");
```

Noting this is optional, since `mongoc_stream_tls_secure_channel_handshake` sets the generic "TLS handshake failed" in the event that the error is unset.

src/libmongoc/src/mongoc/mongoc-secure-channel.c
```diff
@@ -929,7 +976,10 @@ mongoc_secure_channel_handshake_step_3 (mongoc_stream_tls_t *tls,
 
    /* check if the required context attributes are met */
    if (secure_channel->ret_flags != secure_channel->req_flags) {
-      MONGOC_ERROR ("Failed handshake");
+      MONGOC_LOG_AND_SET_ERROR (error,
+                                MONGOC_ERROR_STREAM,
+                                MONGOC_ERROR_STREAM_SOCKET,
+                                "Failed handshake");
```
kevinAlbs, line 982: Optional: Similar comment. I suggest adding an error above L974 like the following:
```
MONGOC_LOG_AND_SET_ERROR (error,
                          MONGOC_ERROR_STREAM,
                          MONGOC_ERROR_STREAM_SOCKET,
                          "Invalid state: credentials unset");
```

src/libmongoc/src/mongoc/mongoc-stream-tls-secure-channel.c
```diff
@@ -884,12 +883,14 @@ mongoc_stream_tls_secure_channel_handshake (mongoc_stream_t *stream,
...
    }
 
-
    *events = 0;
-   bson_set_error (error,
-                   MONGOC_ERROR_STREAM,
-                   MONGOC_ERROR_STREAM_SOCKET,
-                   "TLS handshake failed");
+
+   if (error && !error->code) {
```
kevinAlbs, line 888: The caller, `mongoc_stream_tls_handshake`, may not have initialized `error`. I suggest adding:
```
if (error) {
    memset (error, 0, sizeof (bson_error_t));
}
```
To the start of this function.

src/libmongoc/tests/test-mongoc-client-side-encryption.c
```diff
@@ -135,93 +135,139 @@ _check_bypass (mongoc_auto_encryption_opts_t *opts)
...
-      }
+   aws_secret_access_key =
+      test_framework_getenv ("MONGOC_TEST_AWS_SECRET_ACCESS_KEY");
+   aws_access_key_id = test_framework_getenv ("MONGOC_TEST_AWS_ACCESS_KEY_ID");
+   if (!aws_secret_access_key || !aws_access_key_id) {
+      fprintf (stderr,
+               "Set MONGOC_TEST_AWS_SECRET_ACCESS_KEY and "
+               "MONGOC_TEST_AWS_ACCESS_KEY_ID environment "
+               "variables to run Client Side Encryption tests.");
+      abort ();
```
kevinAlbs, line 157: Optional: `test_error` can simplify tests wanting to print and abort:

```
test_error ("Set MONGOC_TEST_AWS_SECRET_ACCESS_KEY and "
            "MONGOC_TEST_AWS_ACCESS_KEY_ID environment "
            "variables to run Client Side Encryption tests.");
```

This is a refactor, so no need to change. Noting for future reference.

src/libmongoc/tests/test-mongoc-client-side-encryption.c
```diff
@@ -135,93 +140,139 @@ _check_bypass (mongoc_auto_encryption_opts_t *opts)
...
-         abort ();
-      }
+   aws_secret_access_key =
+      test_framework_getenv ("MONGOC_TEST_AWS_SECRET_ACCESS_KEY");
+   aws_access_key_id = test_framework_getenv ("MONGOC_TEST_AWS_ACCESS_KEY_ID");
+   if (!aws_secret_access_key || !aws_access_key_id) {
+      fprintf (stderr,
+               "Set MONGOC_TEST_AWS_SECRET_ACCESS_KEY and "
+               "MONGOC_TEST_AWS_ACCESS_KEY_ID environment "
+               "variables to run Client Side Encryption tests.");
```
kevinAlbs, line 156:
```suggestion
               "variables to enable CSFLE tests..");
```

For consistency with error messages below.

src/libmongoc/tests/test-mongoc-client-side-encryption.c
```diff
@@ -2261,6 +2266,147 @@ test_bypass_spawning_via_bypassAutoEncryption (void *unused)
...
+static void
+test_kms_tls_cert_valid (void *unused)
+{
+   bson_error_t error;
+   const int32_t connecttimeoutms = MONGOC_DEFAULT_CONNECTTIMEOUTMS;
+   const int is_client = 1;
+
+   mongoc_host_list_t host;
+   ASSERT_OR_PRINT (
+      _mongoc_host_list_from_string_with_err (&host, "127.0.0.1:7999", &error),
```
kevinAlbs, line 2307: `test_kms_tls_cert_valid`, `test_kms_tls_cert_expired`, and `test_kms_tls_cert_wrong_host` rely on three instances of the mock KMS server running, on ports 7999, 8000, and 8001.

I suggest noting that assumption in CONTRIBUTING.md. It contains notes about which environment variables are required to run CSFLE tests.

.evergreen/run-tests.sh
```diff
@@ -31,12 +31,20 @@ fi
...
-   sudo cp src/libmongoc/tests/x509gen/ca.pem /usr/local/share/ca-certificates/cdriver.crt || true
-   if [ -f /usr/local/share/ca-certificates/cdriver.crt ]; then
-      sudo update-ca-certificates
-   else
-      export MONGOC_TEST_SSL_CA_FILE="src/libmongoc/tests/x509gen/ca.pem"
-   fi
+
+   case "$OS" in
+      cygwin*)
+         certutil.exe -addstore "Root" "src\libmongoc\tests\x509gen\ca.pem"
```
kevinAlbs, line 37: Oh nice! TIL how to install CA certificates on Windows from the command line.

kevinAlbs (changes requested):
Great work! I appreciate the detailed description. Splitting up the work into meaningful commits made reviewing much easier.

src/libmongoc/src/mongoc/mongoc-stream-tls-openssl.c
```diff
@@ -543,6 +543,28 @@ _mongoc_stream_tls_openssl_check_closed (mongoc_stream_t *stream) /* IN */
 }
 
 
+static bool
+_mongoc_stream_tls_openssl_check_cert_failure (SSL *ssl, bson_error_t *error)
```
rcsanchez97, line 547: Minor nit: I wonder if the name `_mongoc_stream_tls_openssl_cert_verify_failed` might be more consistent with the other functions that return `bool` based on the state of a parameter.

rcsanchez97 (approved):
LGTM, modulo the one minor nit.

eramongodb:
@kevinAlbs Per external discussion, this PR will now disable CSE tests on RHEL upon merge. This change will be reverted on resolution of BUILD-14068, which will obviate the need to manually register the CA certificate via Evergreen build scripts. This followup task is recorded as CDRIVER-4183.

kevinAlbs (approved):

rcsanchez97 (approved):

PR #881, opened 2021-10-16
CDRIVER-4206 KMIP support

kevinAlbs:
# Summary

Implements the specification changes of DRIVERS-1353. Please see the comments of DRIVERS-1353 for the linked specification change commits.

- Add `mongoc_auto_encryption_opts_set_tls_opts` and `mongoc_client_encryption_opts_set_tls_opts` to set custom TLS options on KMS connections.
- Update specification tests and prose tests.
- Update CSFLE documentation:
	- Add KMIP KMS provider.
	- Add missing documentation for Azure and GCP KMS providers (CDRIVER-4087).
	- Fix documentation for local KMS provider (CDRIVER-4088).

# Rejected alternatives

## Passing TLS options with mongoc_ssl_opt_t

The specification is not prescriptive in the API for passing TLS options:

> Drivers SHOULD provide API that is consistent with configuring TLS options for MongoDB server TLS connections. New API to support the options MUST be independent of the KMS provider to permit future extension. The following is an example:

I considered an alternative API for passing TLS options:

```
void
mongoc_auto_encryption_opts_set_tls_opts (
   mongoc_auto_encryption_opts_t *opts,
   const char* provider,
   const mongoc_ssl_opt_t *ssl_opt
);
```

It is more consistent with `mongoc_client_set_ssl_opts`. But I chose to reject it because:
- The `mongoc_ssl_opt_t` does not permit configuring all TLS options supported in the [URI specification](https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst#why-use-tls-as-the-prefix-instead-of-ssl-for-related-options) (`tlsDisableOCSPEndpointCheck`, `tlsDisableCertificateRevocationCheck`).
- `mongoc_ssl_opt_t` includes insecure options (`allow_invalid_hostname` and `weak_cert_validation`) which are prohibited in the specification, as well as C specific options `ca_dir` and `crl_file`, which are not required and are not supported on all TLS implementations.
- SSL is a deprecated term in favor of TLS. E.g. `mongoc_uri_get_ssl` is deprecated for `mongoc_uri_get_tls`.

# Caveats

- Error messages in the "no client certificate" cases are empty. Commented inline.


src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -658,6 +675,179 @@ _state_machine_run (_state_machine_t *state_machine,
...
+      } else if (0 == bson_strcasecmp (key, MONGOC_URI_TLSCERTIFICATEKEYFILE)) {
+         continue;
+      } else if (0 == bson_strcasecmp (key, MONGOC_URI_TLSCAFILE)) {
+         continue;
+      }
+
+      bson_set_error (error,
+                      MONGOC_ERROR_CLIENT_SIDE_ENCRYPTION,
+                      MONGOC_ERROR_CLIENT_INVALID_ENCRYPTION_ARG,
+                      "Error setting TLS options for %s: %s is prohibited",
```
vector-of-bool, line 734: Is "prohibited" the right word to use here?

vector-of-bool (approved):

chardan (changes requested):
Lots of work in here! I especially appreciate the careful documentation. Please take a look at just a few things in here.

src/libmongoc/src/mongoc/mongoc-crypt.c
kevinAlbs: Oops. "prohibited" matches the intent, but not the wording required in the specification:

> Drivers SHOULD raise an error if insecure TLS options are set. The error MUST contain the message "Insecure TLS options prohibited".

Will update.

src/libmongoc/doc/mongoc_uri_t.rst
```diff
@@ -141,16 +141,16 @@ Deprecated SSL Options
...
-MONGOC_URI_SSL                             ssl                               MONGOC_URI_TLS                          tls
-MONGOC_URI_SSLCLIENTCERTIFICATEKEYFILE     sslclientcertificatekeyfile       MONGOC_URI_TLSCERTIFICATEKEYFILE        tlscertificatekeyfile
-MONGOC_URI_SSLCLIENTCERTIFICATEKEYPASSWORD sslclientcertificatekeypassword   MONGOC_URI_TLSCERTIFICATEKEYPASSWORD    tlscertificatekeypassword
-MONGOC_URI_SSLCERTIFICATEAUTHORITYFILE     sslcertificateauthorityfile       MONGOC_URI_TLSCAFILE                    tlscafile
-MONGOC_URI_SSLALLOWINVALIDCERTIFICATES     sslallowinvalidcertificates       MONGOC_URI_TLSALLOWINVALIDCERTIFICATES  tlsallowinvalidcertificates
-MONGOC_URI_SSLALLOWINVALIDHOSTNAMES        sslallowinvalidhostnames          MONGOC_URI_TLSALLOWINVALIDHOSTNAMES     tlsallowinvalidhostnames
-========================================== ================================= ======================================= =================================
+========================================== ================================= =========================================== =================================
```
eramongodb, line 144: Note: Not necessary to address, but this formatting... could be better, lol.

image

src/libmongoc/doc/mongoc_client_encryption_opts_set_kms_providers.rst
```diff
@@ -18,23 +18,52 @@ Parameters
...
+      secretAccessKey: String
    }
 
 The format for "local" is as follows:
 
 .. code-block:: javascript
 
    local: {
-      key: <96 byte BSON binary of subtype 0> // The master key used to encrypt/decrypt data keys.
+      key: byte[96] or String // The master key used to encrypt/decrypt data keys. May be passed as a base64 encoded string.
```
eramongodb, line 37: C++-style comments for C code? It's fine if we're writing C99, but if our minimal support is C89, then perhaps we shouldn't...? 🤔

.evergreen/run-tests.sh
```diff
@@ -138,6 +138,8 @@ if [ "$CLIENT_SIDE_ENCRYPTION" = "on" ]; then
    wait_for_kms_server 7999
    wait_for_kms_server 8000
    wait_for_kms_server 8001
+   wait_for_kms_server 5698
+   wait_for_kms_server 8002
```
eramongodb, line 142:
```suggestion
   wait_for_kms_server 8002
   wait_for_kms_server 5698
```
Minor: order not consistent with scripts and documentation.

eramongodb (approved):
Other than some minor formatting, LGTM. 👍

src/libmongoc/tests/test-libmongoc.c
```diff
@@ -2611,18 +2621,31 @@ test_framework_skip_if_no_failpoint (void)
...
+      "MONGOC_TEST_GCP_EMAIL",
+      "MONGOC_TEST_GCP_PRIVATEKEY",
+      "MONGOC_TEST_CSFLE_TLS_CA_FILE",
+      "MONGOC_TEST_CSFLE_TLS_CERTIFICATE_KEY_FILE",
+      NULL};
+   const char **iter;
+   bool has_creds = true;
+
+   for (iter = required_env_vars; *iter != NULL; iter++) {
+      char* val;
```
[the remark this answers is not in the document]
kevinAlbs: Gah, again I forgot to run clang-format. Thank you.

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
```diff
@@ -117,6 +119,20 @@ mongoc_auto_encryption_opts_set_kms_providers (
...
 
+void
+mongoc_auto_encryption_opts_set_tls_opts (mongoc_auto_encryption_opts_t *opts,
+                                          const bson_t *tls_opts)
+{
+   if (!opts) {
+      return;
+   }
+   bson_destroy (opts->tls_opts);
+   opts->tls_opts = NULL;
```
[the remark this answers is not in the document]
chardan: This is ok, but since it's getting repeated I *wonder* if:

opts_tls_opts = bson_copy_or_null(tls_opts); // makes a copy, or returns NULL
or
bson_copy_or_null(tls_opts, opts->tls_opts); // destroys opts->tls_opts if non-null, then as per above

...might not be a nice simplification-- perhaps with a more catchy name. May or may not be worth doing, but is worth thinking about.

src/libmongoc/tests/test-libmongoc.c
```diff
@@ -499,6 +499,16 @@ test_framework_getenv (const char *name)
    return _mongoc_getenv (name);
 }
 
+char *
+test_framework_getenv_required (const char *name)
```
chardan, line 503: +1, nice! (Although, do we ever call it?)

chardan (approved):
Strong improvements! Just a couple of things you may want to consider, but +1 from me.

src/libmongoc/tests/test-libmongoc.c
kevinAlbs: Yes, this is called in `test-mongoc-client-side-encryption.c` when reading required environment variables.

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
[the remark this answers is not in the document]
kevinAlbs: I am not confident `bson_copy_or_null` is generically useful enough to add to public API, but it seems reasonable to add as a private helper. Added `_bson_copy_or_null` in this file.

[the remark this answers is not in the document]
kevinAlbs: I am not confident `bson_copy_or_null` is generically useful enough to add to public API, but it seems reasonable to add as a private helper. Added `_bson_copy_or_null` in this file.

PR #883, opened 2021-10-29
When running with CTest, don't register all tests.

vector-of-bool:
Only register tests that match the name of the expression. This logic is
still somewhat fuzzy, but provides significant speedup in the common
case.

This reduces test startup overhead for an individual run from ~100ms
to <10ms.

Before:

```sh
~/D/d/m/cmake-build> time ctest -R bson/ -C Debug -j10 -Q
32.97user 4.78system 0:06.57elapsed 574%CPU (0avgtext+0avgdata 62128maxresident)k
0inputs+672outputs (0major+450824minor)pagefaults 0swaps
```

After:

```sh
~/D/d/m/cmake-build> time ctest -R bson/ -C Debug -j10 -Q
1.61user 0.44system 0:00.55elapsed 373%CPU (0avgtext+0avgdata 58548maxresident)k
0inputs+664outputs (0major+186847minor)pagefaults 0swaps
```


src/libmongoc/tests/test-mongoc-change-stream.c
```diff
@@ -2774,6 +2773,6 @@ test_change_stream_install (TestSuite *suite)
    TestSuite_AddMockServerTest (
       suite, "/change_streams/prose_test_18", prose_test_18);
 
-   test_framework_resolve_path (JSON_DIR "/change_streams/legacy", resolved);
-   install_json_test_suite (suite, resolved, &test_change_stream_spec_cb);
+   install_json_test_suite (
+      suite, JSON_DIR, "change_streams", &test_change_stream_spec_cb);
```
kevinAlbs, line 2777: I believe this is the cause of the test failures. This should include the `legacy` directory appended.

kevinAlbs (commented):
The ctest speedup is great. Changes look good with a fix to new failing tests. I commented with a possible cause.

src/libmongoc/tests/test-mongoc-change-stream.c
vector-of-bool: Good find. I don't know how I missed that one.

kevinAlbs (approved):
LGTM!

PR #884, opened 2021-10-30
CDRIVER-4184 don't emit closed on invalid topology

kevinAlbs:


src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -617,7 +617,12 @@ mongoc_topology_destroy (mongoc_topology_t *topology)
       bson_mutex_destroy (&topology->apm_mutex);
       mongoc_cond_destroy (&topology->srv_polling_cond);
    }
-   _mongoc_topology_description_monitor_closed (&topology->description);
+
+   if (topology->valid) {
```
chardan, line 621: +1 I'm assuming that this is okay via spec.

src/libmongoc/tests/test-mongoc-dns.c
```diff
@@ -728,6 +728,117 @@ prose_test_9_pooled (void *unused)
...
 }
 
+typedef struct {
+   int num_topology_opening;
+   int num_topology_closed;
+   int num_server_opening;
+   int num_server_closed;
+} cb_stats_t;
+
+static void
```
chardan, line 738: A comment about why this name would be really helpful: is it creating an invalid state for purposes of testing? (It's not obvious that it is.)

src/libmongoc/tests/test-mongoc-dns.c
```diff
@@ -728,6 +728,117 @@ prose_test_9_pooled (void *unused)
...
+test_invalid_topology_pooled (void *unused) {
+   mongoc_client_pool_t *pool;
+   mongoc_client_t *client;
+   mongoc_uri_t *uri;
+   mongoc_apm_callbacks_t *cbs;
+   cb_stats_t stats = {0};
+
+   /* TXT record for test20.test.build.10gen.cc resolves to "loadBalanced=true". */
+   uri = mongoc_uri_new ("mongodb+srv://test20.test.build.10gen.cc/?replicaSet=rs0");
```
chardan, line 773: In many environments, a unit test against a resources like this is considered an invalid strategy. It's not clear to me that we're ok with this inside of 10gen-- that is, if this is a known-ok technique, so I'm pointing it out.

src/libmongoc/tests/test-mongoc-dns.c
```diff
@@ -728,6 +728,117 @@ prose_test_9_pooled (void *unused)
...
+   pool = mongoc_client_pool_new (uri);
+   cbs = mongoc_apm_callbacks_new ();
```
chardan, line 775: note: It's surprising to me that now single setup function for instantiating with all of these bindings in a single call? (Are they expected to normally not be set together, or left uninitialized..?)

src/libmongoc/tests/test-mongoc-dns.c
```diff
@@ -728,6 +728,117 @@ prose_test_9_pooled (void *unused)
...
+   ASSERT (!sd);
+
+   mongoc_apm_callbacks_destroy (cbs);
+   mongoc_client_destroy (client);
+   mongoc_uri_destroy (uri);
+
+   ASSERT_CMPINT(stats.num_topology_opening, ==, 0);
+   ASSERT_CMPINT(stats.num_server_opening, ==, 0);
+   ASSERT_CMPINT(stats.num_server_closed, ==, 0);
+   ASSERT_CMPINT(stats.num_topology_closed, ==, 0);
```
chardan, line 839: ...and this is the actual subject of the test? Is there a situation when it should be non-zero?

chardan (changes requested):
Looks like a good start! Since it's Halloween today, best check for gremlins, too!

kmahar (approved):
thanks @kevinAlbs! I ran a [patch](https://spruce.mongodb.com/task/mongo_swift_driver_load_balancer_all__os_fully_featured~ubuntu_18.04_swift_version~5.5_ssl_auth~ssl_auth_test_latest_load_balancer_patch_2e03abf05500c3a8c464fb004ced89d037c8a071_617f0ac93e8e866e12d10ba6_21_10_31_21_29_54/tests?execution=0&sortBy=STATUS&sortDir=ASC) with the Swift driver and your branch here vendored in and the DNS seedlist test that was crashing due to this issue now passes.

src/libmongoc/tests/test-mongoc-dns.c
chardan: Ok, @kmahar clarified this in her comment I'm with you now. +1.

src/libmongoc/src/mongoc/mongoc-topology.c
kevinAlbs: [The SDAM monitoring spec](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst#specification) defines the `TopologyOpening` event as:

> When a topology description is initialized - this MUST be the first SDAM event fired

I interpret that to mean it is OK to omit a `TopologyOpening` event if the topology fails to initialize. That is the current behavior. And I think it follows that a `TopologyClosed` event must not be omitted either.

src/libmongoc/tests/test-mongoc-dns.c
kevinAlbs: The names are intended to separate note these callbacks are applied to the `test_invalid_topology_pooled` and `test_invalid_topology_single` tests. Added a comment on the callbacks.

src/libmongoc/tests/test-mongoc-dns.c
kevinAlbs: Yes, this is the subject of the test. A `TopologyClosed` event is emitted in `mongoc_client_destroy` for a valid topology.

src/libmongoc/tests/test-mongoc-dns.c
kevinAlbs: There are 12 callbacks defined in `mongoc-apm.h` to implement the callbacks defined in the SDAM monitoring and Command Monitoring specifications.

The API permits only registering callbacks for individual events. At least in tests, I often do not need all events.

src/libmongoc/tests/test-mongoc-dns.c
chardan: +1 thanks-- naming things is hard!!

src/libmongoc/tests/test-mongoc-dns.c
chardan: Aaah, ok. That's the part that I wasn't so sure about-- sometimes it's good to see them together and optional:
new_sdam_foo(uri, cbs, NULL, ...)
...but that can be just as or honestly more confusing without work going into it (as the frequent guest of foo(/* expl. */ NULL) in our code attests), thanks for the clarification.

src/libmongoc/tests/test-mongoc-dns.c
chardan: Ok, I'm with you. This is is the only accepted result.

chardan (approved):
Thanks for your work on this! I appreciate your clarifications.

PR #889, opened 2021-11-08
CDRIVER-4215 use legacy atomic built-ins on gcc (>= 4.1, < 4.9)

rcsanchez97:
This work is not yet complete.  However, given the scope of the change just in getting the compile to succeed on RHEL 6, it seemed prudent to push my first set of changes as a WIP PR.

Initial patch build: https://spruce.mongodb.com/version/61874ce932f4177a1ad61efb/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

@vector-of-bool In the patch build, you will want to focus on the compile task in the `rhel62` variant.  The test tasks are still failing, but for unrelated reasons.  At this point I am specifically interested to know if you find any errors or deficiencies in my initial changes.

After I get the compile working on RHEL 7 I will take a look at all the warnings that are still being produced.

PR #890, opened 2021-11-08
Set Certain Warnings as Errors Unconditionally

vector-of-bool:
Some constructs that are allowed in C are well-established to be virtually always erroneous, e.g. calling a function that hasn't been declared, implicitly binding a pointer to a different type, and implicitly converting an integral to a pointer. Implicitly converting a pointer-to-`const` to a pointer-to-`mutable` is not _always_ an error, but is problematic enough that an explicit cast should be required.

Many of these errors will be caught at a later time, but only due to test failures, linker failures, or on a build on CI that enables some warnings-as-errors (currently never on Windows).

This change instead makes such constructs always-errors, thus adding a barrier to any changes that produce such errors that might otherwise have been missed. The selected warnings are well-established and detected by all modern compilers, and are very unlikely to appear anew do to compiler upgrades.

Additional warnings that are not so definitely-wrong (etc. sign conversion/compare, unused variables) are not addressed by these changes.


build/cmake/MongoC-Warnings.cmake
```diff
@@ -0,0 +1,46 @@
...
+
+# Warnings that should always be unconditional hard errors, as the code is
+# almost definitely broken
+mongoc_add_platform_compile_options (
+     # Implicit function or variable declarations
+     gnu-like:$<${is_c_lang}:-Werror=implicit> msvc:/we4013 msvc:/we4431
+     # Missing return types/statements
+     gnu-like:-Werror=return-type msvc:/we4716
+     # Incompatible pointer types
+     gnu-like:$<${is_c_lang}:-Werror=incompatible-pointer-types> msvc:/we4113
```
rcsanchez97, line 36: This one might be problematic on older compilers with the new atomics.  For instance, here is an excerpt of build output from RHEL 6.2 (gcc 4.4.7):
```
/root/mongo-c-driver.git/src/libbson/src/bson/bson-atomic.c: In function 'bson_atomic_int_add':
/root/mongo-c-driver.git/src/libbson/src/bson/bson-atomic.c:28: warning: passing argument 1 of 'bson_atomic_int32_fetch_add' from incompatible pointer type
/root/mongo-c-driver.git/src/libbson/src/bson/bson-atomic.h:323: note: expected 'volatile long int *' but argument is of type 'volatile int32_t *'
/root/mongo-c-driver.git/src/libbson/src/bson/bson-atomic.c: In function '_lock_64bit_atomic':
/root/mongo-c-driver.git/src/libbson/src/bson/bson-atomic.c:61: warning: passing argument 1 of 'bson_atomic_int_compare_exchange_weak' from incompatible pointer type
/root/mongo-c-driver.git/src/libbson/src/bson/bson-atomic.h:324: note: expected 'volatile long int *' but argument is of type 'int *'
```

We might want to wait on this PR until after CDRIVER-4215 is done, as that might clear some of the warnings which would turn into errors with this change.

vector-of-bool: There's another open ticket about some related implicit conversions in the atomics code. I plan to inspect it, because it is very suspicious.

build/maintainer-flags.txt
```diff
@@ -10,7 +10,6 @@
 -Wno-strict-aliasing
 -Wno-uninitialized
 -Wredundant-decls
--Wreturn-type
```
chardan, line 13: I'm curious as to why we're removing this one, it seems to fall into the "almost always an error" category?

build/cmake/MongoC-Warnings.cmake
```diff
@@ -0,0 +1,46 @@
...
+     # Integral/pointer conversions
+     gnu-like:$<${is_c_lang}:-Werror=int-conversion> msvc:/we4047
```
chardan, line 38: I've seen some (usually pretty old) code where this can make things noisy-- but it is indeed "almost always an error".

chardan (approved):
+1. I think these will generally only improve the code, although as has been pointed out we should check how they behave across all the compiler versions we support; I'm less familiar with flags on the Microsoft compiler, but from what I've read I think it's also ok.

build/maintainer-flags.txt
vector-of-bool: It was moved into the `MongoC-Warnings.cmake` file. This just removes the redundancy.

build/cmake/MongoC-Warnings.cmake
```diff
@@ -0,0 +1,46 @@
...
+     # Discarding qualifiers
+     gnu-like:$<${is_c_lang}:-Werror=discarded-qualifiers> msvc:/we4090
```
eramongodb, line 40: > warning: unknown warning option '-Werror=discarded-qualifiers'; did you mean '-Werror=ignored-qualifiers'? [-Wunknown-warning-option]

GCC expects `discarded-qualifiers`, but Clang expects `ignored-qualifiers`.

build/cmake/MongoC-Warnings.cmake
```diff
@@ -0,0 +1,46 @@
...
+function (mongoc_add_platform_compile_options)
+   foreach (opt IN LISTS ARGV)
+      if (NOT opt MATCHES "^(gnu-like|msvc):(.*)")
+         message (SEND_ERROR "Invalid option '${opt}' (Should be prefixed by 'msvc:' or 'gnu-like:'")
+         continue ()
+      endif ()
+      set (is_gnu_like "$<C_COMPILER_ID:GNU,Clang,AppleClang>")
+      set (is_msvc "$<C_COMPILER_ID:MSVC>")
+      if (CMAKE_MATCH_1 STREQUAL "gnu-like")
+         add_compile_options ("$<${is_gnu_like}:${CMAKE_MATCH_2}>")
```
eramongodb, line 17:
```
CMake Error at build/cmake/MongoC-Warnings.cmake:17 (add_compile_options):
   Error evaluating generator expression:
     $<C_COMPILER_ID:GNU,Clang,AppleClang>
   $<C_COMPILER_ID> expression requires one or zero parameters.
```

Evergreen tasks appear to be failing during CMake configuration due to invalid generator expression evaluation given definition of `is_gnu_like` on line 14. `$<C_COMPILER_ID:compiler_ids>` (plural "ids") appears to be a [CMake 3.15 and newer feature](https://cmake.org/cmake/help/v3.15/manual/cmake-generator-expressions.7.html#variable-queries). Either the `CMAKE_MINIMUM_REQUIRED` version should be bumped to `3.15`, or the condition should be rewritten, e.g. something like:

```cmake
set (is_gnu_like "$<IN_LIST:$<C_COMPILER_ID>,GNU;Clang;AppleClang>")
```

eramongodb (changes requested):

build/cmake/MongoC-Warnings.cmake
vector-of-bool: oof.

rcsanchez97:
@vector-of-bool Now that #889 is merged, you might consider rebasing these changes and starting a new patch build.  It will give us useful information about the effect of these changes on the build for older platforms.

build/cmake/MongoC-Warnings.cmake
eramongodb: Addressed by https://github.com/mongodb/mongo-c-driver/pull/890/commits/5c1f02724b4dde8fd46f639272b6181cf226b019.

eramongodb (approved):
The workaround for CMake versions less than 3.3 (assuming C language even for C++ targets) leaves me with an unsatisfied feeling. I feel the situation could be improved by using `target_compile_options` rather than `add_compile_options`, but I do not consider it to be a blocker for this PR.

src/libmongoc/tests/test-mongoc-topology.c
```diff
@@ -2430,13 +2430,13 @@ _test_hello_ok (bool pooled)
 }
 
 static void
-test_hello_ok_single ()
+test_hello_ok_single (void)
 {
    _test_hello_ok (false);
```
kevinAlbs, line 2435: Can you fix the `maybe-uninitialized` warnings in the `_test_hello_ok` function? The abi-compliance-check task ([example]( https://spruce.mongodb.com/task/mongo_c_driver_releng_abi_compliance_check_patch_bebe063a42c87cabd8b628e61e9901da1161a1a7_619430bf56234348fc2946dc_21_11_16_22_29_21/logs?execution=1)) with gcc 7.5.0 appears to fail from those.

build/cmake/MongoC-Warnings.cmake
```diff
@@ -0,0 +1,61 @@
...
+   Define additional compile options, conditional on the compiler being used.
+   Each option should be prefixed by `gnu-like:` or `msvc:`. Those options will be
+   enabled for GCC/Clang or MSVC respectively.
+
+   These options are attached to the source directory and its children.
+]]
+function (mongoc_add_platform_compile_options)
+   foreach (opt IN LISTS ARGV)
+      if (NOT opt MATCHES "^(gnu-like|gnu|clang|msvc):(.*)")
+         message (SEND_ERROR "Invalid option '${opt}' (Should be prefixed by 'msvc:' or 'gnu-like:'")
```
kevinAlbs, line 11:
```suggestion
         message (SEND_ERROR "Invalid option '${opt}' (Should be prefixed by 'msvc:', 'gnu-like:', `gnu:`, or `clang:`")
```

CMakeLists.txt
```diff
@@ -79,6 +79,8 @@ set (CMAKE_MODULE_PATH
 include (InstallRequiredSystemLibraries)
 include (GNUInstallDirs)
 
+include(MongoC-Warnings)
```
kevinAlbs, line 82: I suggest commenting to note the intent of these warnings as compared to the warnings enabled with `ENABLE_MAINTAINER_FLAGS`. My understanding is:
- `MongoC-Warnings` are always enabled. They should be a subset that should never produce warnings.
- `ENABLE_MAINTAINER_FLAGS` are less critical, but of interest to maintainers.

src/kms-message/src/kms_port.c
```diff
@@ -15,8 +15,13 @@
...
 
 #include "kms_port.h"
+
+#include <stdlib.h>
+#include <string.h>
+
 #if defined(_WIN32)
-char * kms_strndup (const char *src, size_t len)
+char *
+kms_strndup (const char *src, size_t len)
```
kevinAlbs, line 24: `src/kms-message` is a vendored copy from libmongocrypt.

Should the new warnings only be applied to `src/libbson`, `src/libmongoc`, and `src/common`?

kevinAlbs (commented):
Nice work! The generator expressions make it simpler to declare equivalent warnings across compilers.

I suggest running a full patch build to check all compilers on the Evergreen waterfall. That can be done with the Evergreen CLI: `evergreen patch --project=mongo-c-driver`

src/libmongoc/CMakeLists.txt
```diff
@@ -45,9 +45,6 @@ if (NOT ENABLE_ZSTD MATCHES "ON|AUTO|OFF")
    message (FATAL_ERROR "ENABLE_ZSTD option must be ON, AUTO, or OFF")
 endif ()
 
-# Disable warnings on bundled zlib source files.
```
kevinAlbs, line 48: The debug-compile-with-warnings task is warning on [zlib sources](https://spruce.mongodb.com/task/mongo_c_driver_releng_debug_compile_with_warnings_patch_bebe063a42c87cabd8b628e61e9901da1161a1a7_61a56fb7d6d80a0f59f623ac_21_11_30_00_26_34/logs?execution=0). Should this be added back?

kevinAlbs (commented):
Looks good, I think some tasks are failing with warnings on zlib sources.

src/libmongoc/CMakeLists.txt
vector-of-bool: Fixed now

kevinAlbs (approved):
LGTM! The effort towards fixing compiler warnings and improving developer workflow is much appreciated.

src/libmongoc/CMakeLists.txt
```diff
@@ -78,7 +75,7 @@ endif ()
 if ( (ENABLE_ZLIB STREQUAL "BUNDLED")
    OR (ENABLE_ZLIB STREQUAL "AUTO" AND NOT ZLIB_FOUND) )
    message (STATUS "Enabling zlib compression (bundled)")
-   set (SOURCES ${SOURCES} ${ZLIB_SOURCES})
+   set (SOURCES ${SOURCES} $<TARGET_OBJECTS:zlib_obj>)
```
rcsanchez97, line 78: It's not clear to me that this change is necessary or correct.  From the `cmake-generator-expressions` man page:

```
$<TARGET_OBJECTS:objLib>
              List of objects resulting from build of objLib. objLib must be an object of type OBJECT_LIBRARY.
```

That seems to indicate that the generator expression will actually provide a list of `.o` files.  Am I mistaken in thinking this?

build/cmake/MongoC-Warnings.cmake
```diff
@@ -0,0 +1,74 @@
...
+set (__is_gnu_like "$<OR:${__is_gnu},${__is_clang}>")
+set (__is_msvc "$<C_COMPILER_ID:MSVC>")
+
+# "Old" GNU is GCC < 5, which is missing several warning options
+set (__is_old_gnu "$<AND:${__is_gnu},$<VERSION_LESS:$<C_COMPILER_VERSION>,5>>")
+set (__not_old_gnu "$<NOT:${__is_old_gnu}>")
+
+#[[
+   Define additional compile options, conditional on the compiler being used.
+   Each option should be prefixed by `gnu-like:` or `msvc:`. Those options will be
```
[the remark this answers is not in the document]
vector-of-bool: I forgot to update this comment, yeah.

src/libmongoc/CMakeLists.txt
vector-of-bool: This is correct. `TARGET_OBJECTS` is designed to be used in the source lists of other targets. Using `TARGET_OBJECTS` will have the object files from zlib directly included into the library, but they will compiled under the rules of the `zlib_obj` library. This prevents the zlib sources from being compiled with the flags that we use on the mongoc targets (e.g. the warnings flags), but they'll still be included in the mongoc libraries directly.

rcsanchez97: Cool!  TIL

rcsanchez97 (approved):
LGTM

PR #891, opened 2021-11-08
Bump WIRE_VERSION_MAX to 14

chardan:
Updates the wire protocol max version to 14.

Added comments to help when searching for maxWireVersion.

See: https://jira.mongodb.org/browse/CDRIVER-4165

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-client-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-client-private.h
@@ -42,8 +42,8 @@
 BSON_BEGIN_DECLS
 
 /* protocol versions this driver can speak */
-#define WIRE_VERSION_MIN 3	/* a.k.a. minWireVersion */
-#define WIRE_VERSION_MAX 14	/* a.k.a. maxWireVersion */
+#define WIRE_VERSION_MIN 3  /* a.k.a. minWireVersion */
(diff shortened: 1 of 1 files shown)


src/libmongoc/src/mongoc/mongoc-client-private.h
```diff
@@ -42,8 +42,8 @@
 BSON_BEGIN_DECLS
 
 /* protocol versions this driver can speak */
-#define WIRE_VERSION_MIN 3
-#define WIRE_VERSION_MAX 13
+#define WIRE_VERSION_MIN 3	/* a.k.a. minWireVersion */
+#define WIRE_VERSION_MAX 14	/* a.k.a. maxWireVersion */
```
kevinAlbs, line 46: Please run clang-format. I think the tab will be replaced with spaces.

kevinAlbs (approved):
LGTM with a run of clang-format.

src/libmongoc/src/mongoc/mongoc-client-private.h
chardan: Sorry, I meant to make this a draft PR-- I am wondering if there are tests that need changing. I only see a couple that reference maxWireVersion 13, I'll see if it causes a regression.

PR #892, opened 2021-11-10
Fix Win32 Warnings

vector-of-bool:
There are some warnings that appear in code, including one major incompatible-pointer-type that only appears on Windows. This also fixes an incompatible pointer type used in bson-atomic.c appearing in some configuration of MinGW on Windows (CDRIVER-4221).


kevinAlbs (approved):
LGTM!

PR #893, opened 2021-11-11
use legacy atomics on gcc < 5

rcsanchez97:
So, it turns out that we need to use legacy atomics on all gcc 4.x versions.  The original change in #889 still results in build failures on Debian 8 (gcc 4.9).  It seems that using the new `__atomic` built-ins requires setting `-std=c11`.  It is not clear that we want to require that for legacy platforms, so expanding the check in this way seems like the better way.

Full patch build (filtered down to the Debian 8 variant): https://spruce.mongodb.com/version/618d9bcd0305b93ba3051410/tasks?page=0&variant=%5Egcc49%24

### DIFF
--- a/src/libbson/src/bson/bson-atomic.h
+++ b/src/libbson/src/bson/bson-atomic.h
@@ -47,15 +47,15 @@ enum bson_memory_order {
 #define MSVC_MEMORDER_SUFFIX(X)
 #endif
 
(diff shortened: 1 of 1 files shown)


vector-of-bool (approved):
Ouch. Well, a simple fix. LGTM.

kevinAlbs (approved):
LGTM, thank you for the quick fix!

PR #894, opened 2021-11-12
SHM counter on mac m1 arm64 implementation.

devnexen:


src/libmongoc/src/mongoc/mongoc-counters-private.h
```diff
@@ -101,6 +101,16 @@ _mongoc_sched_getcpu (void)
 }
 #elif defined(MONGOC_HAVE_SCHED_GETCPU)
 #define _mongoc_sched_getcpu sched_getcpu
+#elif defined(__APPLE__) && defined(__aarch64__)
+static BSON_INLINE unsigned
+_mongoc_sched_getcpu (void)
+{
+   uintptr_t tls;
+   unsigned core_id;
+   __asm__ volatile("mrs %x0, tpidrro_el0" : "=r"(tls));
```
kevinAlbs, line 110: https://developer.arm.com/documentation/den0024/a/The-Memory-Management-Unit/Context-switching?lang=en notes:

> Additionally, the ARMv8-A Architecture provides Thread ID registers for use by operating system software. These have no hardware significance and are typically used by threading libraries as a base pointer to per-thread data.

Would reading `MPIDR_EL1` be more appropriate? References:
- https://developer.arm.com/documentation/den0024/a/Multi-core-processors/Multi-processing-systems/Determining-which-core-the-code-is-running-on?lang=en
- https://community.arm.com/support-forums/f/architectures-and-processors-forum/6978/armv8-a-instruction-for-getting-cpu-number

src/libmongoc/src/mongoc/mongoc-counters-private.h
```diff
@@ -101,6 +101,16 @@ _mongoc_sched_getcpu (void)
 }
 #elif defined(MONGOC_HAVE_SCHED_GETCPU)
 #define _mongoc_sched_getcpu sched_getcpu
+#elif defined(__APPLE__) && defined(__aarch64__)
```
kevinAlbs, line 104: Is it possible that `tipdrro_el0` is not available on less privileged modes of execution?

If so, I suggest wrapping this in a conditional define, similar to `MONGOC_ENABLE_RDTSCP`.

kevinAlbs (changes requested):
Thank you for the contribution @devnexen! I included questions and suggestions. I filed CDRIVER-4241 for the proposed improvement.

src/libmongoc/src/mongoc/mongoc-counters-private.h
devnexen: What you re referring to is considered illegal in non privileged mode.

src/libmongoc/src/mongoc/mongoc-counters-private.h
devnexen: Thankfully it is accessible in non privileged mode (e.g. some memory allocators, etc use it).

src/libmongoc/src/mongoc/mongoc-counters-private.h
```diff
@@ -101,6 +101,16 @@ _mongoc_sched_getcpu (void)
...
+static BSON_INLINE unsigned
+_mongoc_sched_getcpu (void)
+{
+   uintptr_t tls;
+   unsigned core_id;
+   __asm__ volatile("mrs %x0, tpidrro_el0" : "=r"(tls));
+   core_id = tls & ((1 << 3) - 1);
```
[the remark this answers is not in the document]
devnexen: the first 8 manageable cores. in arm 16 cores is in fact 8+8.

kevinAlbs (approved):
Thank you again for the contribution @devnexen. Changes LGTM. I am requesting another reviewer. Filed CDRIVER-4240 to add macOS m1 to our test matrix.

src/libmongoc/src/mongoc/mongoc-counters-private.h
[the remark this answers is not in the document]
eramongodb: I would prefer use of unsigned integer literal `0b0111u` or `0x07u` over bitwise arithmetic, but I believe the code should work fine as proposed.

src/libmongoc/src/mongoc/mongoc-counters-private.h
eramongodb: I would prefer use of [ARM C Language Extensions](https://gcc.gnu.org/onlinedocs/gcc/ARM-C-Language-Extensions-_0028ACLE_0029.html) (test for `__ARM_ACLE` -> `#include <arm_acle.h>` -> `__arm_rsr64()` compiler intrinsic?) over inline ASM, but I believe the code should work fine as proposed.

eramongodb (approved):
I would have preferred use of compiler intrinsics (`<arm_acle.h>`?) in place of inline ASM, but I do not consider it to be a blocker. LGTM.

kevinAlbs:
Thanks again for the contribution @devnexen. This PR was rebased atop CDRIVER-4240 to test building on a macOS m1 host. Merging.

PR #897, opened 2021-11-16
CDRIVER-4229 emulate int32 and int atomic operations on zSeries

kevinAlbs:
**Summary**
Emulate int32 and int atomic operations on zSeries to resolve CDRIVER-4229.

This is a full patch build with the fix: https://spruce.mongodb.com/version/619412e6e3c33129bc14e09a

**Background & Motivation**
This [patch build](https://spruce.mongodb.com/task/mongo_c_driver_zseries_rhel72_debug_compile_nosasl_nossl_patch_2078ab2272ad72df091a757fba9d45285221477d_6193f615d1fe072144f38c53_21_11_16_18_19_02/tests?execution=0&sortBy=STATUS&sortDir=ASC) tested the integral types individually and narrowed the test failure to the `int32` and `int` atomic types.

I do not know why the legacy atomic intrinsics do not function correctly on zSeries. But this seemed like a straightforward fix to unblock the 1.20.0 release.


src/libbson/src/bson/bson-atomic.h
```diff
@@ -67,6 +67,13 @@ enum bson_memory_order {
 #define BSON_IF_GNU_LEGACY_ATOMICS(...)
 #endif
 
+/* CDRIVER-4229 zSeries with gcc 4.8.4 produces illegal instructions for int and
+ * int32 atomic intrinsics. */
+#if defined(__s390__) || defined(__s390x__) || defined(__zarch__)
```
kevinAlbs, line 72: These macros were determined from https://sourceforge.net/p/predef/wiki/Architectures/. `gcc` on the `zseries-rhel72` distro defines all three.

vector-of-bool (approved):
Ouch. Good to get a fix. Do we have a zSeries target in CI?

kevinAlbs:
> Ouch. Good to get a fix. Do we have a zSeries target in CI?

Yes. The zSeries variant is named `zseries-rhel72` in `.evergreen/config.yml`. It is not part of the subset of tasks that runs on GitHub PRs. Though I am thinking it might be useful to have compile tasks on all variants run as part of the subset of tasks that are run from GitHub PRs.

rcsanchez97 (approved):
LGTM

PR #898, opened 2021-11-16
CDRIVER-4097 Implement srvMaxHosts for initial DNS seedlist and SRV polling

eramongodb:
This PR implements CDRIVER-4097, which introduces the [`srvMaxHosts` URI option](https://github.com/mongodb/specifications/blob/b508f6d2a1819882d40c4e7e73b95fc8ae3bdfe5/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst#srvmaxhosts). This option limits number of hosts during [initial DNS seedlist discovery](https://github.com/mongodb/specifications/blob/b508f6d2a1819882d40c4e7e73b95fc8ae3bdfe5/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst#querying-dns) and [SRV polling for mongos discovery](https://github.com/mongodb/specifications/blob/b508f6d2a1819882d40c4e7e73b95fc8ae3bdfe5/source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst#implementation).

Any changes involving the `srvServiceName` URI option were deliberately skipped; this feature should be addressed separately as CDRIVER-4086.

This PR comes in three major parts:

1. URI Options

First, a [bug](https://github.com/mongodb/mongo-c-driver/commit/29e1ec913127d3acbdfe835c7267c44f1734b5ef) in the validation of load balancer URI options had to be addressed to permit valid execution of `srvMaxHosts`-related tests.

Then, the necessary URI option parsing features [were implemented](https://github.com/mongodb/mongo-c-driver/commit/2a67283dde3cb41f242a65262f0dbf1ad3c2a140) and corresponding spec tests [added](https://github.com/mongodb/mongo-c-driver/commit/1406b6ffa7896e052ae0ae4d990a53718d62d12b) to the uri-options directory. Note that some test cases had to be [added to the warning filters](https://github.com/mongodb/mongo-c-driver/commit/1406b6ffa7896e052ae0ae4d990a53718d62d12b#diff-de273c6f735cdbe82c59b0bd26c6157ba527804980ccdf42bbcb6b2988379e3bR104-R106) as CDRIVER-3167 has not yet been addressed.

2. Initial DNS Seedlist Discovery

There were two drive-by improvements during this step. The first was the addition of a `const`-qualifier to a [read-only host list function](https://github.com/mongodb/mongo-c-driver/commit/82e32bb294c0c463cca8f6a4619f1e52477762cf). The second was fixing [an infinite loop bug](https://github.com/mongodb/mongo-c-driver/commit/e995831720c9a89c137f99efa90de51287fb2d0f) in `dump_hosts()` (only triggered by certain test failures) due to an incorrect induction step.

As the Initial Seedlist Discovery Spec requires [random shuffling and selection](https://github.com/mongodb/specifications/blob/b508f6d2a1819882d40c4e7e73b95fc8ae3bdfe5/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst#querying-dns) of hosts, I took the opportunity to introduce a set of [uniform random integer functions](https://github.com/mongodb/mongo-c-driver/commit/bec90ab95c46f3fdfc35a302a88b24c29ef33947). These are `_mongoc_rand_uint32_t`, `_mongoc_rand_uint64_t`, and `_mongoc_rand_size_t` respectively, each returning a random integer in the range `[min, max]` with uniform distribution. The algorithms used are as described in ["Fast Random Integer Generation in an Interval" by Daniel Lemire](https://arxiv.org/abs/1805.10941). The "nearly divisionless" algorithm proposed by the paper is used for 32-bit integers. The "Java algorithm" is used for 64-bit integers as the nearly divisionless algorithm requires using 128-bit integers during computation; I did not want to attempt to introduce `__int128` into the C driver due to consistent cross-platform compatibility being [effectively non-existent](https://quuxplusone.github.io/blog/2019/02/28/is-int128-integral/).

With the new random number generators, implementing the filtering of initial DNS seedlist hosts was [relatively straightforward](https://github.com/mongodb/mongo-c-driver/commit/0480de8fb1ad6434ad168914516f7fcae067ce2e).

The new spec tests for Initial Seedlist Discovery also required [some modification to test code](https://github.com/mongodb/mongo-c-driver/commit/adf0c12f3fb68cf250414e76d88b1dc6bb8bfd68#diff-c80a6c3e231807c673e7da7a1f5db8c468ba7027084fa9fc5d7cc2489365823d) to accomodate the new `numHosts` parameter. Evergreen/CI scripts did not have to be updated to accomodate the new `sharded` spec tests, as the `load-balancer` spec test setup also happens to [satisfy the conditions required](https://github.com/mongodb/specifications/tree/b508f6d2a1819882d40c4e7e73b95fc8ae3bdfe5/source/initial-dns-seedlist-discovery/tests#test-setup) to correctly run `sharded` spec tests (a mongos server on ports 27017 and 27018). These new `sharded` spec tests can therefore be run alongside existing `load-balancer` tests on Evergreen.

3. SRV Polling for mongos Discovery

There was some [awkward math required](https://github.com/mongodb/mongo-c-driver/commit/976aaa58312ea98c14a7edf75520638fdb58469b#diff-6eef3e5d332b14e922ca887b8a8ba5b39efd0a109ebc9b27923cdcae15575061R2378) to preserve the correct order of operations, where the [addition of new valid hosts](https://github.com/mongodb/mongo-c-driver/commit/976aaa58312ea98c14a7edf75520638fdb58469b#diff-6eef3e5d332b14e922ca887b8a8ba5b39efd0a109ebc9b27923cdcae15575061R2360-R2406) (within the limit as described by `srvMaxHosts`) must come _before_ [removal of missing hosts](https://github.com/mongodb/mongo-c-driver/commit/976aaa58312ea98c14a7edf75520638fdb58469b#diff-6eef3e5d332b14e922ca887b8a8ba5b39efd0a109ebc9b27923cdcae15575061R2408-R2419) to avoid generating [a misleading warning](https://github.com/mongodb/mongo-c-driver/blob/2078ab2272ad72df091a757fba9d45285221477d/src/libmongoc/src/mongoc/mongoc-topology-description.c#L995). If the warning can be disabled or removed, the implementation of `mongoc_topology_description_reconcile` could be made simpler.

Finally, prose tests 10, 11, and 12 were [added](https://github.com/mongodb/mongo-c-driver/commit/2444d18e1a4301678b90d113bfbe5d35def6e738) according to the [SRV Polling Tests Spec](https://github.com/mongodb/specifications/tree/b508f6d2a1819882d40c4e7e73b95fc8ae3bdfe5/source/polling-srv-records-for-mongos-discovery/tests). However, due to the verbosity and repetition of the existing patterns used by prose tests, I took the opportunity to perform some [dependency inversion](https://github.com/mongodb/mongo-c-driver/commit/435e15daca3ab13503a3167f339968f97858bd95) (or is it ["Inversion of Control"](https://en.wikipedia.org/wiki/Inversion_of_control)? Either way, some kind of "inversion" is involved) to simplify the implementation of the prose test code. To better understand the intent of this refactor, I recommend comparing the state of a single prose test before and after the change in separate views, rather than side-by-side or inline as a diff.

The SRV Polling prose tests have been given the [`test_dns_check_srv_polling()` skip condition](https://github.com/mongodb/mongo-c-driver/commit/2444d18e1a4301678b90d113bfbe5d35def6e738#diff-c80a6c3e231807c673e7da7a1f5db8c468ba7027084fa9fc5d7cc2489365823dR388-R392) such that they are only run if the `MONGOC_TEST_DNS_SRV_POLLING` environment variable is set to `on`. This is due to current Evergreen scripts not yet being configured to create the required sharded cluster configuration with mongos instances running on 4 ports. Supporting this will take additional work as described by CDRIVER-4230. Note, this means that the Evergreen tasks being run for this PR do _not_ currently evaluate the SRV Polling prose tests, unliked the sharded tests for Initial DNS Seedlist Discovery, although they have been verified locally as passing given the correct sharded cluster configuration.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-util.c
+++ b/src/libmongoc/src/mongoc/mongoc-util.c
@@ -727,7 +727,7 @@ uint32_t
 _mongoc_rand_uint32_t (uint32_t min, uint32_t max)
 {
    BSON_ASSERT (min <= max);
-   BSON_ASSERT (min != 0u && max != UINT32_MAX);
+   BSON_ASSERT (min != 0u || max != UINT32_MAX);
 
    return _mongoc_rand_nduid32 (max - min + 1u, _mongoc_simple_rand_uint32_t) +
           min;
@@ -737,7 +737,7 @@ uint64_t
 _mongoc_rand_uint64_t (uint64_t min, uint64_t max)
 {
    BSON_ASSERT (min <= max);
-   BSON_ASSERT (min != 0u && max != UINT64_MAX);
+   BSON_ASSERT (min != 0u || max != UINT64_MAX);
 
    return _mongoc_rand_java64 (max - min + 1u, _mongoc_simple_rand_uint64_t) +
           min;
--- a/src/libmongoc/src/mongoc/mongoc-util.c
+++ b/src/libmongoc/src/mongoc/mongoc-util.c
@@ -21,6 +21,7 @@
 #include <string.h>
 
 #include "common-md5-private.h"
+#include "mongoc-rand-private.h"
 #include "mongoc-util-private.h"
 #include "mongoc-client.h"
 #include "mongoc-client-session-private.h"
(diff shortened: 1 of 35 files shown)


src/libmongoc/src/mongoc/mongoc-util.c
```diff
@@ -626,3 +627,154 @@ _mongoc_getenv (const char *name)
...
+{
+   uint64_t res;
+
+   (void) _mongoc_rand_bytes ((uint8_t *) &res, sizeof (res));
+
+   return res;
+}
+
+#else
+/* Otherwise, fallback to rudimentary use of `rand()`. */
```
kevinAlbs, line 700: I think `rand()` needs to be preceded with a call to `srand()` to seed. Otherwise, multiple processes may select the same initial subset of hosts. From [rand(3) - Linux manual page](https://man7.org/linux/man-pages/man3/srand.3.html):

> If no seed value is provided, the rand() function is automatically seeded with a value of 1.

test-libmongoc calls `srand` from [TestSuite.c](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/tests/TestSuite.c#L656:L656). And `_bson_context_init_random` conditionally calls `srand`.

I suggest adding a call to srand in a `BSON_ONCE` function with a seed similar to the one generated in `_bson_context_init_random`.

src/libmongoc/src/mongoc/mongoc-util.c
```diff
@@ -626,3 +627,154 @@ _mongoc_getenv (const char *name)
...
+   while ((x - r) > (limit - s)) {
+      x = rand64 ();
+      r = x % s;
+   }
+
+   return r;
+}
+
+#if defined(MONGOC_ENABLE_CRYPTO)
+/* Attempt to use `_mongoc_rand_bytes` if available. */
```
kevinAlbs, line 677: What is the motivation for preferring `_mongoc_rand_bytes` instead of `rand`?

One concern I have is that is mixes intention. Callers of `_mongoc_rand_uint32_t`, `_mongoc_rand_uint64_t`, and `_mongoc_rand_size_t` are not guaranteed a random number derived from a secure random source depending on the platform. 

Secure random generation may also be significantly slower on some platforms.

kevinAlbs (commented):
This looks great. Thank you for the detailed description. The prose test refactoring of pooled/single functions is a nice pattern. The only substantial comments are about the new random functions.

> If the warning can be disabled or removed, the implementation of mongoc_topology_description_reconcile could be made simpler.

The warning is motivated by an SDAM specification recommendation in CDRIVER-3107.

kevinAlbs (approved):
LGTM, fantastic work.

vector-of-bool (approved):
Sorry for the very slow review. This looks good. The thorough inline comments are very helpful!

PR #901, opened 2021-11-29
CDRIVER 3992 deprecate legacy wire protocols

chardan:


src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -823,16 +866,15 @@ _mongoc_stream_run_hello (mongoc_cluster_t *cluster,
...
-    * known hello indicates the server supports a newer wire protocol.
-    */
    memset (&hello_cmd, 0, sizeof (hello_cmd));
    hello_cmd.db_name = "admin";
    hello_cmd.command = command;
    hello_cmd.command_name = _mongoc_get_command_name (command);
    hello_cmd.query_flags = MONGOC_QUERY_SECONDARY_OK;
    hello_cmd.server_stream = server_stream;
 
+/* JFW what about hello_cmd SASL? Looks like they never performed encryption when called via
```
chardan, line 876: I think I see the reason for this-- will remove comment.

src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -28,6 +28,7 @@
 #include "../test-conveniences.h"
 #include "../test-libmongoc.h"
 #include "../TestSuite.h"
+#include "../mongoc-test-settings.h"
```
chardan, line 31: This may not be needed at all; the intent is to provide a central location for transporting settings across the client/mock-server/harness boundaries via global access (my real idea is a lightweight symbol table, not a bunch of free variables).

src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -1016,13 +1017,23 @@ mock_server_receives_legacy_hello (mock_server_t *server,
...
    if (strcasecmp (request->command_name, "hello") &&
        strcasecmp (request->command_name, HANDSHAKE_CMD_LEGACY_HELLO)) {
+fprintf(stderr, "JFW: mock_server_receives_legacy_hello(): strings did NOT match; command name \"%s\"\n", request->command_name), fflush(stderr);
       request_destroy (request);
       return NULL;
    }
 
+/* JFW: */
+fprintf(stderr, "JFW: mongoc_test_settings_legacy_hello_expect_ApiVersion = %d\n", mongoc_test_settings_legacy_hello_expect_ApiVersion);
+if(mongoc_test_settings_legacy_hello_expect_ApiVersion) {
```
chardan, line 1027: In my tests, this is /never/ true. So I'm slightly confused as to how it ever works, TBH, and seem to remember this being needed-- but that was before I left for vacation, so now that I'm back it's possible that the phase of the moon has changed and it's always been fine. (Or perhaps I was linking against old code?)

src/libmongoc/tests/mongoc-test-settings.h
```diff
@@ -0,0 +1,21 @@
...
+/* 
+A place for settings that are shared between the test harness, tests, and mock server. 
+
+TODO: rather than free globals, a simple symbol table API:
+
+void *mongoc_test_settings_get(mongoc_test_settings_ctx *ctx, const char *test_name); 
+
+Returns true if old settings were replaced:
+bool mongoc_test_settings_register(mongoc_test_settings_ctx *ctx, const char *test_name, void **settings); 
+
```
chardan, line 13: (ETC.)

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -641,6 +641,53 @@ mongoc_cluster_run_command_monitored (mongoc_cluster_t *cluster,
...
+ *
+ *--------------------------------------------------------------------------
+ */
+static bool _mongoc_should_use_op_msg(const mongoc_cluster_t *cluster, const mongoc_cmd_t *cmd)
+{
+ /* If we're running a newer wire protocol OR have manually selected a server protocol,
+ always select OP_MSG: */
+
+ return (WIRE_VERSION_OP_MSG <= cmd->server_stream->sd->max_wire_version) || 
+        mongoc_client_uses_server_api(cluster->client);
```
kevinAlbs, line 663: Is the intended change is to use OP_MSG with `hello` on the first command if a server API version is set as described in CDRIVER-4193?

If so, [mongoc-async-cmd.c](https://github.com/mongodb/mongo-c-driver/commit/aa76669d60e3bdc9670eedcf64bc12472bd03ec8#diff-d35f0d63a4ad8bfb630b204735bbc00dab445deed4a03e0e23793671c5fb4152R26) will need to be updated. That is used by a single threaded `mongoc_client_t` for scanning servers. This can be tested with a mock server test (noted below).

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -679,14 +726,10 @@ mongoc_cluster_run_command_private (mongoc_cluster_t *cluster,
...
-      retval = mongoc_cluster_run_opmsg (cluster, cmd, reply, error);
-   } else {
-      retval =
-         mongoc_cluster_run_command_opquery (cluster, cmd, -1, reply, error);
-   }
-   _handle_not_primary_error (cluster, server_stream, reply);
+
+   retval = _mongoc_cluster_run_command_selected_mech_private(cluster, cmd, reply, error);
+
+    _handle_not_primary_error (cluster, server_stream, reply);
```
kevinAlbs, line 732: One possible way to validate these changes is to use a mock server test. Mock server tests are a blunt instrument, but I do not see another easy alternative to test.

Here is an example of a mock server test validating that `isMaster` is the first command sent:
https://github.com/mongodb/mongo-c-driver/commit/aa76669d60e3bdc9670eedcf64bc12472bd03ec8#diff-d35f0d63a4ad8bfb630b204735bbc00dab445deed4a03e0e23793671c5fb4152R26

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -641,6 +641,53 @@ mongoc_cluster_run_command_monitored (mongoc_cluster_t *cluster,
...
+ *
+ * Returns:
+ *       true if successful; otherwise false and @error is set.
+ *
+ * Side effects:
+ *  as per mongoc_cluster_run_command_opquery() or mongoc_cluster_run_opmsg()
+ *
+ *--------------------------------------------------------------------------
+ */
+static bool _mongoc_cluster_run_command_selected_mech_private(
```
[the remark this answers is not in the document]
chardan: Yes, for "mechanism". This isn't the most inspired function name. I suppose with the leading underscore implying private, I could give "_mongoc_cluster_run_command_selected_wire_protocol()" a try, but I'm all ears for ideas!

src/libmongoc/src/mongoc/mongoc-cluster.c
chardan: Yes, that's right. Basically, the only time we should now speak OP_QUERY is with an old wire version on the server and no specified API (which would imply the higher version anyway). I see _mongoc_async_cmd_init_send() in mongoc-async-cmd.c, I'll see what I can do.

src/libmongoc/tests/test-mongoc-hello.c
```diff
@@ -0,0 +1,164 @@
...
+   TestSuite_AddMockServerTest (
+      suite,
+      "/hello",
+      test_mongoc_hello,
+      test_framework_skip_if_max_wire_version_less_than_5);
+   TestSuite_AddMockServerTest (
+      suite,
+      "/hello/client_pool",
+      test_mongoc_hello_client_pool,
+      test_framework_skip_if_max_wire_version_less_than_5);
```
kevinAlbs, line 163: `test_framework_skip_if_max_wire_version_less_than_5` checks the wire version of a live server running for tests. Since these tests are using the mock server, the `test_framework_skip_if_max_wire_version_less_than_5` can be removed.

src/libmongoc/tests/test-mongoc-hello.c
```diff
@@ -0,0 +1,164 @@
...
+bool
+_test_mongoc_hello_client_pool_impl ()
+{
+   mongoc_client_t
+      *cl3; /* the command client, which we'll push back into the pool */
+   bson_t *cmd;
+   bson_error_t err;
+   bool cmd_result;
+
+   mongoc_server_api_t *requested_server_api = NULL;
```
kevinAlbs, line 100: IIUC the proposed behavior change is to always use `OP_MSG` if a server API version is set. 

I think `requested_server_api` will need to be set to test the new behavior.

src/libmongoc/tests/test-mongoc-hello.c
chardan: Ah, I meant to do this! Thank you for catching it.

src/libmongoc/tests/test-mongoc-hello.c
chardan: This has been a bit confounding: it sounds like:
* if we are requested to use a specific API version, we always use OP_MSG;
* if a specific version is not requested, we try OP_QUERY;
* if OP_MSG has been negotiated, subsequent commands use it.

This makes this a bit tricky to test, as I don't know of a way to directly see what was actually negotiated. (There is no was I see so far to get messages back and forthvrom the mock server.) I'll extend this a bit more and at least be sure the client side is trying both ways.

src/libmongoc/src/mongoc/mongoc-cluster.c
chardan: As discussed offline, this PR reverts async work.

src/libmongoc/tests/test-mongoc-hello.c
```diff
@@ -0,0 +1,200 @@
...
+
+   test_framework_set_pool_ssl_opts (pool);
+
+   client = mongoc_client_pool_pop (pool);
+   ASSERT(client);
+
+   // Popping the client should have triggered a connection:
+   if(-1 != requested_server_api_version) {
+        // a specific API version has been requested:
+   	request = mock_server_receives_hello(server);
```
kevinAlbs, line 159: I think these tests pass without additional changes behavior changes.

`mock_server_receives_hello` uses `mock_server_receives_command`.
`mock_server_receives_command` expects OP_QUERY.

I suggest taking a step back. Start on a fresh branch.

The proposed behavior change is to use OP_MSG for the first hello if a server API version is requested. Try the red/green/refactor method:

1. Add failing tests.
2. Make the behavior change to get passing tests.
3. Refactor as needed.

kevinAlbs (changes requested):

src/libmongoc/src/mongoc/mongoc-cluster.c
chardan: Ah, my misunderstanding! I'll revisit changes to mongo-async-cmd.c; getting the information about the requested API into that layer may require changing some of the signatures, I will look again but last time I didn't see the client settings exposed to them.

chardan:
Superceeded-by: https://github.com/mongodb/mongo-c-driver/pull/930

PR #902, opened 2021-11-29
CDRIVER-4225 Explicitly set CA and cert file for use by KMS TLS connections

eramongodb:
Thanks to CDRIVER-4206, the KMS TLS connections can now use user-provided CA and certificate files rather than depending on system certificate registration. This PR removes system cert registration and associated test skip condition checks, replacing them with use of the new `tls_opts` fields whose values are set by the `MONGOC_TEST_CSFLE_TLS_CA_FILE` and `MONGOC_TEST_CSFLE_TLS_CERTIFICATE_KEY_FILE` environment variables.


kevinAlbs (approved):

rcsanchez97 (approved):
LGTM!

PR #904, opened 2021-11-30
CDRIVER-3893: Conditionally support $merge and $out execution on secondaries

vector-of-bool:
This changeset is for CDRIVER-3893, which allows sending of aggregate commands to secondary servers, even if those aggregate operations contain writing stages (`$out` and  `$merge`).

~~TODO: This changeset works fully with recent servers, but when server selection falls back to the prior behavior of selecting a primary regardless of `readPreference`, the spec mandates that we must not send `$readPreference` with the aggregate command. This is surprisingly difficult given the current architecture of the code as these actions do not communicate with each other.~~ Done.


eramongodb (commented):
The PR contains changes that are shared with / in conflict with https://github.com/mongodb/mongo-c-driver/pull/890/. Can these changes be synchronized or removed from this PR?

vector-of-bool:
I forgot that I had started from that branch in this one. I thought I would be able to get it merged sooner, but the holidays and unexpected compiler errors have held that one back. It will definitely be consolidated before this one will be fully ready.

src/libmongoc/src/mongoc/mongoc-topology-description.c
```diff
@@ -1874,9 +1931,9 @@ transition_t gSDAMTransitionTable
  *--------------------------------------------------------------------------
  */
 static const char *
-_mongoc_topology_description_type (mongoc_topology_description_t *topology)
+_tpld_type_str (mongoc_topology_description_type_t type)
```
eramongodb, line 1934: Comment documentation should be updated to reflect new function name.

src/libmongoc/src/mongoc/mongoc-cursor.c
```diff
@@ -1070,11 +1073,17 @@ _mongoc_cursor_run_command (mongoc_cursor_t *cursor,
 
       mongoc_server_stream_cleanup (server_stream);
 
-      server_stream = mongoc_cluster_stream_for_reads (&cursor->client->cluster,
-                                                       cursor->read_prefs,
-                                                       cursor->client_session,
-                                                       reply,
-                                                       &cursor->error);
+      BSON_ASSERT (cursor->is_aggr_with_write_stage ==
```
eramongodb, line 1076: Confirming: is it appropriate to be asserting this condition vs. treating it as a recoverable command failure?

eramongodb (commented):
Refactoring the JSON test code to reduce repetition of path resolution checks is very nice. 👍

src/libmongoc/src/mongoc/mongoc-cursor.c
```diff
@@ -662,11 +663,13 @@ _mongoc_cursor_fetch_stream (mongoc_cursor_t *cursor)
                                            &reply,
                                            &cursor->error);
    } else {
-      server_stream = mongoc_cluster_stream_for_reads (&cursor->client->cluster,
-                                                       cursor->read_prefs,
-                                                       cursor->client_session,
-                                                       &reply,
-                                                       &cursor->error);
+      server_stream =
```
kevinAlbs, line 666: > TODO: This changeset works fully with recent servers, but when server selection falls back to the prior behavior of selecting a primary regardless of readPreference, the spec mandates that we must not send $readPreference with the aggregate command. This is surprisingly difficult given the current architecture of the code as these actions do not communicate with each other.

I am not sure if it sending `"$readPreference":{"mode":"secondary"}` in a command to a primary after overriding the read preference is an acceptable caveat. [This test](https://github.com/kevinAlbs/go-bootstrap/blob/a8a0bf43d9932f1c35c0ad4ff63f82c36f7279c7/investigations/sending_dollar_secondary_to_a_primary/main.go#L32-L37) sends a `ping` with secondary read preference. `"$readPreference":{"mode":"secondary"}` is in the command started events. The command succeeds in server versions back to 3.6.

cc @jmikola in case you are aware of other side-effects.

@vector-of-bool if this is not an acceptable caveat, would adding an out parameter, like `mongoc_read_prefs_t *effective_read_pref` to `_mongoc_cursor_fetch_stream` and passing that down to `mongoc_topology_description_suitable_servers` be a possible (though gross) solution?

jmikola: Even if an RS primary accepts `{ $readPreference: { mode: "secondary" }}`, that could be problematic if sent to a mongos or load balancer.

In https://github.com/mongodb/mongo-php-library/pull/876, I opted to check the list of servers in the topology first to decide whether or not to force a primary read preference. I imagine there's a more efficient way to do this in libmongoc (I had to resort to an extra server selection attempt if the topology was not already initialized), such as using an output parameter on a new server selection function.

src/libmongoc/tests/json-test.c
```diff
@@ -1895,9 +1896,21 @@ _install_json_test_suite_with_check (TestSuite *suite,
...
    va_list ap;
+   char joined[PATH_MAX];
+   char resolved[PATH_MAX];
+
+   snprintf (joined, PATH_MAX, "%s/%s", base, subdir);
+   ASSERT (realpath (joined, resolved));
+
+   if (suite->ctest_run) {
+      const char *found = strstr (suite->ctest_run, subdir);
+      if (found != suite->ctest_run && found != suite->ctest_run + 1) {
```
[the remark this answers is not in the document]
vector-of-bool: This came from the #883 changeset

src/libmongoc/src/mongoc/mongoc-cursor.c
vector-of-bool: I put in the recommended changes to pass around the effective read mode and store it on the cursor and server stream. It's definitely not the prettiest solution, but it is the most un-intrusive one that I can think of.

src/libmongoc/src/mongoc/mongoc-cmd.c
```diff
@@ -474,6 +474,13 @@ _mongoc_cmd_parts_assemble_mongos (mongoc_cmd_parts_t *parts,
       hedge = mongoc_read_prefs_get_hedge (parts->read_prefs);
    }
 
+   if (server_stream->effective_read_mode != (mongoc_read_mode_t) 0) {
+      /* Server selection may have overriden the read mode used to generate this
+       * server stream. This has effects on the body of the message that we send
+       * to the server */
+      mode = server_stream->effective_read_mode;
```
jmikola, line 481: `mongoc_read_mode_t` is not a complete read preference on its own. Isn't this problematic if the "effective read preference" is more complex (e.g. includes tags) and we need to send _that_ in the outgoing command?

src/libmongoc/src/mongoc/mongoc-cursor-private.h
```diff
@@ -131,6 +132,17 @@ struct _mongoc_cursor_t {
    mongoc_read_prefs_t *read_prefs;
    mongoc_write_concern_t *write_concern;
 
+   /** If the cursor was created for an operation that might have overridden the
+    * user's read preferences' read mode, the actually-used read mode will be
+    * stored here. Otherwise, this is zero. This value should be restored to
+    * server_stream objects that are created from this cursor when the stream
+    * was created using the memoized 'server_id' of the cursor. */
+   mongoc_read_mode_t effective_read_mode;
```
jmikola, line 140: This relates to my earlier comment, but I think you need to track the entire read preference here.

src/libmongoc/src/mongoc/mongoc-topology-description.c
```diff
@@ -649,6 +654,57 @@ _mongoc_topology_description_validate_max_staleness (
    return true;
 }
 
+static bool
+_check_any_server_less_than_wire_version_13 (const void *sd_,
+                                             void *any_too_old_)
```
jmikola, line 659: I assume this signature is necessary for the `mongoc_set_for_each` API? Each callback must return true to indicate success and allow iteration to continue?

src/libmongoc/src/mongoc/mongoc-topology-description.c
```diff
@@ -649,6 +654,57 @@ _mongoc_topology_description_validate_max_staleness (
...
+ * requested and what is available in the topology.
+ *
+ * Per the CRUD spec, if the requested read mode is *not* primary, and *any*
+ * server in the topology has a wire version <13, we must override the read
+ * mode preference with "primary." Wire version 13 indicates support on a
+ * secondary server for using aggregate pipelines that contain writing stages
+ * (i.e. '$out' and '$merge').
+ */
+static mongoc_read_mode_t
+_calc_effective_read_mode (const mongoc_topology_description_t *td,
```
jmikola, line 680: Per my earlier comment, this may need to calculate the effective read _preference_ (not just mode).

src/libmongoc/src/mongoc/mongoc-topology-description.c
vector-of-bool: This is correct

src/libmongoc/src/mongoc/mongoc-cmd.c
vector-of-bool: The read mode is only tracked here so that we can later decide whether to send the read preference at all. The command builder will have access to the full read preference data, but it will only need to know whether server selection ignored the read mode or not.

kevinAlbs (commented):
Looking good!

src/libmongoc/src/mongoc/mongoc-cmd.c
jmikola: Ah, thanks for clearing that up. I'd be in favor of changing this to a boolean in the interest of clarity/maintainability if that's not too much trouble.

src/libmongoc/src/mongoc/mongoc-topology-description.c
```diff
@@ -649,6 +654,57 @@ _mongoc_topology_description_validate_max_staleness (
...
+                           mongoc_ss_optype_t optype,
+                           mongoc_read_mode_t requested_read_mode)
+{
+   if (requested_read_mode == MONGOC_READ_PRIMARY ||
+       requested_read_mode == MONGOC_READ_PRIMARY_PREFERRED) {
+      return requested_read_mode;
+   }
+   switch (optype) {
+   case MONGOC_SS_WRITE:
+      return (mongoc_read_mode_t) 0;
```
eramongodb, line 690: What clause in the [CRUD spec](https://github.com/mongodb/specifications/blob/3a4a6e0f318545c9f050a0000fbd53fc4d9a2790/source/crud/crud.rst#read-preferences-and-server-selection) does this condition ("unset" result) correspond to? My understanding is that this function should either return either `primary` (for pre-5.0 server and etc.) or the "available read preference" (`requested_read_mode`) only. Shouldn't this function be equivalent to the following?
```c
static mongoc_read_mode_t
_calc_effective_read_mode (const mongoc_topology_description_t *td,
                           mongoc_ss_optype_t optype,
                           mongoc_read_mode_t requested_read_mode)
{
   /* This section is only applicable if an explicit (i.e. per-operation) or
    * inherited (e.g. from a Collection) read preference is available and it is
    * not a primary read preference. */
   if (requested_read_mode == MONGOC_READ_PRIMARY) {
      return requested_read_mode;
   }

   /* If there are one or more available servers and one or more of those
    * servers is pre-5.0, drivers MUST NOT use the available read preference and
    * MUST instead select a server using a primary read preference. */
   {
      bool any_too_old = false;
      mongoc_set_for_each_const (mc_tpld_servers_const (td),
                                 _check_any_server_less_than_wire_version_13,
                                 &any_too_old);
      if (any_too_old) {
         return MONGOC_READ_PRIMARY;
      }
   }

   /* Otherwise, if there are either no available servers, all available servers
    * are 5.0+ (i.e. wire version >= 13), or the topology type is LoadBalanced
    * (we can assume the backing mongos is 5.0+), drivers MUST use the available
    * read preference. */
   BSON_ASSERT (mongoc-set-size (mc_tpld_servers_const (td))));
   BSON_ASSERT (!any-too-old); /* Addressed in above block. */
   BSON_ASSERT (is-loadbalanced (td)); /* Per LoadBalanced assumption. */
   return requested_read_mode;
}
```

It is also not clear to me from the spec what the behavior should be if the topology type is _not_ LoadBalanced (which seems like a possibility that is not ruled out by prior conditions). Possible spec wording defect?

src/libmongoc/src/mongoc/mongoc-topology-description.c
```diff
@@ -678,11 +735,18 @@ mongoc_topology_description_suitable_servers (
...
+      _calc_effective_read_mode (topology, optype, given_read_mode);
+
+   if (chosen_read_mode) {
+      *chosen_read_mode = effective_read_mode;
+   }
 
    candidates = bson_malloc0 (sizeof (*candidates) * td_servers->items_len);
 
-   data.read_mode = read_mode;
+   data.read_mode = effective_read_mode;
```
eramongodb, line 749: There is mixed usage of both `data.read_mode` and `effective_read_mode` in this function following this assignment, despite both being guaranteed to have the same value. This creates a misleading impression that it matters which is being used. For clarity, can the function be updated to use only one or the other following this assignment?

src/libmongoc/src/mongoc/mongoc-topology-description.c
```diff
@@ -649,6 +654,57 @@ _mongoc_topology_description_validate_max_staleness (
    return true;
 }
 
+static bool
+_check_any_server_less_than_wire_version_13 (const void *sd_,
+                                             void *any_too_old_)
+{
+   const mongoc_server_description_t *sd = sd_;
+   bool *any_too_old = any_too_old_;
+   if (sd->max_wire_version < 13) {
```
eramongodb, line 663:
```suggestion
   if (sd->max_wire_version < WIRE_VERSION_5_0) {
```

src/libmongoc/src/mongoc/mongoc-cursor-private.h
```diff
@@ -131,6 +132,17 @@ struct _mongoc_cursor_t {
...
+
+   /** Whether this cursor corresponds to an aggregate command that contains a
+    * writing-stage */
+   aggr_with_write_stage_flag is_aggr_with_write_stage;
```
eramongodb, line 144: Both the name and the binary choice of enumerators available suggest this should be a `bool`.

src/libmongoc/src/mongoc/mongoc-cmd.c
eramongodb: > if and only if the `effective_read_mode` is primary or 0 [...] only need to know whether server selection ignored the read mode or not

I am also in favor of the boolean.

Suggestion: I recommend `must_use_primary` as the name per wording in the [Server Selection Spec](https://github.com/mongodb/specifications/blob/3a4a6e0f318545c9f050a0000fbd53fc4d9a2790/source/server-selection/server-selection.rst#use-of-read-preferences-with-commands):
> For pre-5.0 servers, an aggregate command is "must-use-primary" if its pipeline contains a write stage (e.g. `$out`, `$merge`) ...

src/libmongoc/src/mongoc/mongoc-cursor.c
vector-of-bool: No, this is not corresponding to a test failure, but the enforcement of being retryable is spread throughout the codebase enough that I feel it should be guarded against with an assert to prevent future modifications from accidentally retrying an aggregate+write operation. The assert isn't to check that the user doesn't do something wrong, but to assert that the driver code itself doesn't do something wrong. My reasoning is that retrying a write could cause a double-write, which may corrupt user data, so it would be better to halt than allow it to pass through.

src/libmongoc/src/mongoc/mongoc-cmd.c
vector-of-bool: There is more than just a true/false state on how server selection behaved. I don't know how to narrow this down to a single bit of information and keep the same behavior. It may be possible, but I'm not sure what it would look like.

src/libmongoc/src/mongoc/mongoc-cmd.c
```diff
@@ -474,6 +474,13 @@ _mongoc_cmd_parts_assemble_mongos (mongoc_cmd_parts_t *parts,
       hedge = mongoc_read_prefs_get_hedge (parts->read_prefs);
    }
```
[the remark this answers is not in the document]
vector-of-bool: I don't think so, but I could be wrong. The guard is around the `_mongoc_cmd_parts_add_read_prefs` function, which is called by `_mongoc_cmd_parts_assemble_mongos` and `mongoc_cmd_parts_assemble`. Since `..._mongod` doesn't call this function, I didn't add any logic there.

src/libmongoc/src/mongoc/mongoc-cursor.c
```diff
@@ -654,22 +655,32 @@ _mongoc_cursor_fetch_stream (mongoc_cursor_t *cursor)
    ENTRY;
 
    if (cursor->server_id) {
+      /* We already did server selection once before. Reuse the prior
+       * selection to create a new stream in a new stream. */
```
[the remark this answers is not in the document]
vector-of-bool: Just a typo.

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -1151,6 +1156,7 @@ uint32_t
 mongoc_topology_select_server_id (mongoc_topology_t *topology,
                                   mongoc_ss_optype_t optype,
                                   const mongoc_read_prefs_t *read_prefs,
+                                  mongoc_read_mode_t *chosen_read_mode,
```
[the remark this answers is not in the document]
vector-of-bool: Removed a duplicate doc comment block.

src/libmongoc/src/mongoc/mongoc-server-stream.c
```diff
@@ -37,6 +37,7 @@ mongoc_server_stream_new (const mongoc_topology_description_t *td,
    bson_copy_to (&td->cluster_time, &server_stream->cluster_time);
    server_stream->sd = sd;         /* becomes owned */
    server_stream->stream = stream; /* merely borrowed */
+   server_stream->effective_read_mode = (mongoc_read_mode_t) 0;
```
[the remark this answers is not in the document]
vector-of-bool: A better idea that fiddling around with a magic zero.

src/libmongoc/src/mongoc/mongoc-topology-description.c
```diff
@@ -649,6 +654,57 @@ _mongoc_topology_description_validate_max_staleness (
...
+ * secondary server for using aggregate pipelines that contain writing stages
+ * (i.e. '$out' and '$merge').
+ */
+static mongoc_read_mode_t
+_calc_effective_read_mode (const mongoc_topology_description_t *td,
+                           mongoc_ss_optype_t optype,
+                           mongoc_read_mode_t requested_read_mode)
+{
+   if (requested_read_mode == MONGOC_READ_PRIMARY ||
+       requested_read_mode == MONGOC_READ_PRIMARY_PREFERRED) {
```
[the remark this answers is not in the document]
vector-of-bool: This is a good question. My intuition was to treat them as equivalent cases, but the literal text says otherwise. @jmikola do you know what the correct behavior would be for 'primaryPreferred'? Is it even applicable to this code path?

src/libmongoc/src/mongoc/mongoc-topology-description.c
vector-of-bool: This is another good question. The purpose of calculating the effective read mode here is to send that information back to the caller so that it can modify its behavior when it builds the command to send to the server. Importantly, the overriding of the read mode should only apply if the intended operation is an aggregate operation with write stages, so that branch _must_ remain (A regular old read operation should stick to a secondary if the user requested a secondary). 

Whether to care about the read mode when the intended operation is a write (as with `MONGOC_SS_WRITE`) is interesting, since it should have no bearing on server selection. Using the zero "unset" value when the operation is a write is just working under the assumption that the user won't be giving use a readPreference for a write operation, since that doesn't seem to make sense.

src/libmongoc/src/mongoc/mongoc-topology-description.c
vector-of-bool: Yeah, this is confusing. I can clean things up by moving most information into `data`.

src/libmongoc/src/mongoc/mongoc-cursor-private.h
vector-of-bool: It may be a binary value, but at call sites it a plain `bool` is opaque and untyped, e.g.:

```c
retry_server_stream =
   mongoc_cluster_stream_for_reads (&client->cluster,
                                    parts->read_prefs,
                                    parts->assembled.session,
                                    NULL,
                                    NOT_AGGR_WITH_WRITE_STAGE,
                                    &ignored_error);
```

versus

```c
retry_server_stream =
   mongoc_cluster_stream_for_reads (&client->cluster,
                                    parts->read_prefs,
                                    parts->assembled.session,
                                    NULL,
                                    false,
                                    &ignored_error);
```

src/libmongoc/src/mongoc/mongoc-topology-description.c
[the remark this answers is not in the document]
jmikola: Sorry for the delayed response. `primary` should definitely be distinguished from `primaryPreferred` when we're concerned with whether to support `$out`/`$merge` on secondaries or not. I'm not entirely sure about the context of this function, but if it's intended to determine fall back behavior then we should consider `primary` separately.

There are ramifications to the read preference beyond what libmongoc itself will select. In the case of a load balanced or sharded cluster, we're potentially forwarding the read preference on to mongos for evaluation.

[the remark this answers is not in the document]
eramongodb: @vector-of-bool I don't believe this has been addressed yet?

src/libmongoc/src/mongoc/mongoc-cursor-private.h
eramongodb: All call sites use `NOT_AGGR_WITH_WRITE_STAGE`; `AGGR_WITH_WRITE_STAGE` is in fact completely unused in the code as proposed. Couldn't this parameter (and similar parameters in corresponding functions) be removed entirely? (This should alleviate concerns regarding the type/name of the data member on grounds of function call argument readability.)

src/libmongoc/src/mongoc/mongoc-topology-description.c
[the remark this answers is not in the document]
vector-of-bool: Darn. I had one last commit that was un-pushed. Pushed now.

src/libmongoc/src/mongoc/mongoc-cursor-private.h
vector-of-bool: > `AGGR_WITH_WRITE_STAGE` is in fact completely unused in the code

This completely surprised me as "how could the code possibly work if no one ever passes this flag??" I dug through and realized that I was accidentally already implicitly relying on a `bool`-to-`enum` conversion in `mongoc-aggregate.c` to set the `mongoc_cursor_t::is_aggr_with_write_stage` attribute, which is then passed to `mongoc_cluster_stream_for_reads`, which is the single place where we _do_ actually pass `AGGR_WITH_WRITE_STAGE` after it is implicitly created from a `true` value. It's just a happy accident that the enum flag aligns with `true`/`false`, as inverting it would cause the logic to invert in the conversion. Oops. I was hoping that a compiler warning would have caught such a conversion, but it appears that no compiler implements such a warning ([and there doesn't seem to be much push to add such a warning](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=7654)). C enums hurt.

I'll fix the bool-to-enum conversion, but I suppose it can still be debated whether the binary `enum` is worth using.

src/libmongoc/src/mongoc/mongoc-cmd.c
kevinAlbs: > There is more than just a true/false state on how server selection behaved. I don't know how to narrow this down to a single bit of information and keep the same behavior. It may be possible, but I'm not sure what it would look like.

IIUC the information needed in `mongoc_cmd_parts_assemble` is "Can the user's read preference be applied, or should it be overridden with a primary read preference".

This is what I have in mind for the logic below:

```
   mode = mongoc_read_prefs_get_mode (prefs_ptr);
   if (server_stream->must_use_primary) {
      /* Server selection determined the user's read preference must be overridden with primary */
      mode = MONGOC_READ_PRIMARY;
   }
```

src/libmongoc/src/mongoc/mongoc-cmd.c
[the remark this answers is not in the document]
kevinAlbs: `_mongoc_cmd_parts_assemble_mongod` uses the `parts->read_prefs->mode` to determine how to set OP_QUERY flags. 

For background, `_mongoc_cmd_parts_assemble_mongod` applies to OP_QUERY. OP_MSG was introduced to supersede OP_QUERY in MongoDB 3.6 (wire version 6).

OP_QUERY is still used on MongoDB 3.6+ servers for monitoring servers with the `hello` / legacy hello commands.

`_mongoc_cmd_parts_assemble_mongod` and `_mongoc_cmd_parts_assemble_mongos` may be able to be removed when the minimum wire version is set to 6 (server version 3.6) in CDRIVER-4098. @eramongodb may be able to confirm or deny that.

If `_mongoc_cmd_parts_assemble_mongod` and `_mongoc_cmd_parts_assemble_mongos` are being removed, I think it is OK to leave them unchanged.

Otherwise, I suggest passing the `mode`  as an argument to `_mongoc_cmd_parts_assemble_mongod`.

kevinAlbs (commented):
Looking good pending the discussion about replacing `effective_read_mode` with a boolean `must_use_primary` and updating `_mongoc_cmd_parts_assemble_mongod`.

src/libmongoc/src/mongoc/mongoc-cluster-private.h
```diff
@@ -105,6 +105,16 @@ mongoc_cluster_try_recv (mongoc_cluster_t *cluster,
                          mongoc_server_stream_t *server_stream,
                          bson_error_t *error);
 
+/**
+ * @brief Flag telling whether an aggregate operation has a write stage or not
+ */
+typedef enum aggr_with_write_stage_flag {
+   /** The operation is not an aggregate with a write stage */
+   NOT_AGGR_WITH_WRITE_STAGE,
```
jmikola, line 113: I assume the enum starts at 0 but should we explicitly define that? A number of other enums in libmongoc also define the first value, but it's possible that's only done when a non-zero value is used.

A separate observation was that it seemed a bit odd to introduce a mandatory parameter referring to aggregation on a common method used by things other than aggregation. I'm not too concerned either way as `mongoc_cluster_stream_for_reads` is private so we could always address this down the line if desired/possible. This is really a complaint about how annoying it is to implement this feature as opposed to a criticism on the PR itself.

Edit: this may relate to the other discussion in https://github.com/mongodb/mongo-c-driver/pull/904/files#r780560871

src/libmongoc/tests/test-mongoc-dns.c
```diff
@@ -441,7 +441,7 @@ test_all_spec_tests (TestSuite *suite)
    install_json_test_suite_with_check (
       suite,
       JSON_DIR,
-      "/initial_dns_seedlist_discovery/sharded",
+      "initial_dns_seedlist_discovery/sharded",
```
jmikola, line 444: This looks unrelated. Is it fixing an older issue not tracked by a ticket?

jmikola:
I don't have any major input to provide on this PR, so I'd like to remove myself and defer to Ezra and Kevin if that's OK. Let me know if there are any spec-related issues you'd still like me to provide feedback on, though.

src/libmongoc/src/mongoc/mongoc-cursor.c
```diff
@@ -666,7 +666,7 @@ _mongoc_cursor_fetch_stream (mongoc_cursor_t *cursor)
                                            &cursor->error);
       /* Also restore the effective read mode that was used on that prior
        * selection */
```
eramongodb, line 668: Update comment to reflect `effective_read_mode` -> `must_use_primary`?

src/libmongoc/src/mongoc/mongoc-cursor.c
```diff
@@ -680,7 +680,7 @@ _mongoc_cursor_fetch_stream (mongoc_cursor_t *cursor)
          /* Remember the selected server_id and the effective read mode so that
           * we can re-create an equivalent server_stream at a later time */
```
eramongodb, line 681: Update comment to reflect `effective_read_mode` -> `must_use_primary`?

src/libmongoc/src/mongoc/mongoc-topology-description.c
```diff
@@ -655,15 +655,17 @@ _mongoc_topology_description_validate_max_staleness (
...
 static bool
-_check_any_server_less_than_wire_version_v5_0 (const void *sd_,
-                                               void *any_too_old_)
+_check_any_server_less_than_wire_version_13 (const void *sd_,
+                                             void *any_too_old_)
 {
    const mongoc_server_description_t *sd = sd_;
    bool *any_too_old = any_too_old_;
-   if (sd->max_wire_version < WIRE_VERSION_5_0) {
+   if (sd->max_wire_version < 13) {
```
eramongodb, line 663:
```suggestion
   if (sd->max_wire_version < WIRE_VERSION_5_0) {
```
Reverted; should still be applied.

src/libmongoc/src/mongoc/mongoc-topology-description.c
```diff
@@ -676,10 +678,10 @@ _check_any_server_less_than_wire_version_v5_0 (const void *sd_,
  * secondary server for using aggregate pipelines that contain writing stages
  * (i.e. '$out' and '$merge').
  */
-static mongoc_read_mode_t
-_calc_effective_read_mode (const mongoc_topology_description_t *td,
-                           mongoc_ss_optype_t optype,
-                           mongoc_read_mode_t requested_read_mode)
+static bool
+_force_read_primary (const mongoc_topology_description_t *td,
```
eramongodb, line 682:
```suggestion
_must_use_primary (const mongoc_topology_description_t *td,
```
Consistency with name of data member.

src/libmongoc/src/mongoc/mongoc-topology-description.c
```diff
@@ -690,29 +692,28 @@ _calc_effective_read_mode (const mongoc_topology_description_t *td,
...
       }
       /* We're okay to send an aggr-with-write to a secondary server, so permit
        * the caller's read mode preference */
-      return requested_read_mode;
+      return false;
    }
    default:
-      BSON_UNREACHABLE (
-         "Invalid mongoc_ss_optype_t for _calc_effective_read_mode()");
+      BSON_UNREACHABLE ("Invalid mongoc_ss_optype_t for _force_read_primary()");
```
eramongodb, line 716:
```suggestion
      BSON_UNREACHABLE ("Invalid mongoc_ss_optype_t for _must_use_primary()");
```
Consistency with name of data member.

eramongodb (approved):
Minor feedback, but no requests for functional changes left. LGTM.

src/libmongoc/src/mongoc/mongoc-cmd.c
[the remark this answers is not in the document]
kevinAlbs: After CDRIVER-4098 is merged, I think an aggregate will always go through OP_MSG. I do not think changes to `_mongoc_cmd_parts_assemble_mongod` or `_mongoc_cmd_parts_assemble_mongos` are required.

I suggest keeping both `_mongoc_cmd_parts_assemble_mongod` or `_mongoc_cmd_parts_assemble_mongos` consistent, either:
1. Revert the changes to `_mongoc_cmd_parts_assemble_mongos` and and comment noting that an aggregate command can only use OP_MSG.
2. Or change L629 in `_mongoc_cmd_parts_assemble_mongod` from:
```c
         if (parts->read_prefs &&
             parts->read_prefs->mode != MONGOC_READ_PRIMARY) {
            parts->assembled.query_flags |= MONGOC_QUERY_SECONDARY_OK;
         }
```

To:

```c
         if (parts->read_prefs &&
             parts->read_prefs->mode != MONGOC_READ_PRIMARY &&
             !server_stream->must_use_primary) {
            parts->assembled.query_flags |= MONGOC_QUERY_SECONDARY_OK;
         }
```

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -2808,7 +2814,7 @@ mongoc_cluster_stream_for_reads (mongoc_cluster_t *cluster,
    }
 
    return _mongoc_cluster_stream_for_optype (
-      cluster, MONGOC_SS_READ, prefs_override, cs, reply, error);
+      cluster, has_write_stage, prefs_override, cs, reply, error);
```
kevinAlbs, line 2817: `has_write_stage` is being converted to `mongoc_ss_optype_t`. I think this may be related to the cause of test failures on pre-5.0 servers.

Perhaps `MONGOC_SS_READ` and `has_write_stage` should be separate arguments to `_mongoc_cluster_stream_for_optype`?

kevinAlbs (changes requested):

src/libmongoc/src/mongoc/mongoc-cluster.c
vector-of-bool: Good find. Implicit int-to-enum conversion strikes again.

src/libmongoc/tests/test-mongoc-dns.c
vector-of-bool: I believe this is a just a change from a manual merge resolution, not a fix for any issue.

src/libmongoc/src/mongoc/mongoc-topology-description.c
```diff
@@ -895,8 +989,12 @@ mongoc_topology_description_select (
    _mongoc_array_init (&suitable_servers,
```
kevinAlbs, line 989: Tasks on pre 5.0 sharded servers are failing: https://spruce.mongodb.com/version/61e76013a4cf4705407cfa1f/tasks?page=0&sorts=STATUS%3AASC

This reproduces locally with this environment:
```
mlaunch init --replicaset --nodes=1 --sharded=1 --mongos=2 --priority --binarypath $(m bin 4.2-ent) --hostname localhost --enableMajorityReadConcern --setParameter enableTestCommands=1 --dir .menv_sh2_42
```

I was unable to fully track down the cause, but I have an idea.

libmongoc sets the initial topology type to `MONGOC_TOPOLOGY_SINGLE` when given a URI with one host and no replicaSet option (e.g. `mongodb://foo.example.com/`) . That behavior is driver dependent and noted in [this spreadsheet](https://docs.google.com/spreadsheets/d/1sddTIxDtaB_x6z8c2vp6QbTwJSHm5nEPCRHbhUBLbPs/edit#gid=0).

I applied this diff:
```patch
diff --git a/src/libmongoc/src/mongoc/mongoc-topology-description.c b/src/libmongoc/src/mongoc/mongoc-topology-description.c
index 2ba90466c..0f9a5fcb7 100644
--- a/src/libmongoc/src/mongoc/mongoc-topology-description.c
+++ b/src/libmongoc/src/mongoc/mongoc-topology-description.c
@@ -978,6 +978,11 @@ mongoc_topology_description_select (
    if (topology->type == MONGOC_TOPOLOGY_SINGLE) {
       sd = mongoc_set_get_item_const (mc_tpld_servers_const (topology), 0);
 
+      if (optype == MONGOC_SS_AGGREGATE_WITH_WRITE &&
+          sd->max_wire_version < WIRE_VERSION_5_0) {
+         *must_use_primary = true;
+      }
+
       if (sd->has_hello_response) {
          RETURN (sd);
       } else {
```

Which resolves the failure locally.

kevinAlbs (changes requested):

src/libmongoc/src/mongoc/mongoc-topology-description.c
vector-of-bool: This is a good catch. I had wrongly assumed those failures were spurious. I'll add those changes.

kevinAlbs (approved):
LGTM!

The task failures on `test-asan-4.2-sharded-auth-nosasl-openssl`, `test-asan-4.4-sharded-auth-nosasl-openssl`, and `test-asan-5.0-sharded-auth-nosasl-openssl` are failing on the waterfall in the same test `/transactions/legacy/mongos-recovery-token`. I filed CDRIVER-4268 to track those.

PR #906, opened 2021-12-03
fix build of bundled zlib

rcsanchez97:
Ensure that the target for the bundled zlib is only built when the
bundled target is needed.  Also ensure that the build always correctly
finds the necessary headers.  For instance, even if there is a zlib
installed on the system and the build specifies the bundled zlib, ensure
that bundled headers are located before the system headers.


vector-of-bool (approved):
LGTM.

I didn't think about the search path ordering issue. Good find.

kevinAlbs (approved):

PR #915, opened 2021-12-17
CDRIVER-4207 handshake on new stream in scanner

kevinAlbs:


jmikola (approved):
Tested this patch with PHPLIB's test suite ([PHPLIB-717](https://jira.mongodb.org/browse/PHPLIB-717)) and noted that we no longer receive the following errors throughout:

> Failed to get handshake server description: Driver attempted to initialize in load balancing mode, but the server does not support this mode.

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -4145,6 +4145,65 @@ test_mongoc_client_get_handshake_establishes_connection_pooled (void)
...
+
+   /* Send a "ping" command. */
+   future = future_client_command_simple (
+      client, "db", ping, NULL /* read_prefs */, NULL /* reply */, &error);
+   /* The first command on the new connection is handshake. It uses the legacy
+    * hello and includes the client.application.name. */
+   request = mock_server_receives_legacy_hello (
+      server,
+      "{'" HANDSHAKE_CMD_LEGACY_HELLO
+      "': 1, 'client': {'application': {'name': 'foo'}}}");
```
chardan, line 4172: Ok, other fields are optional.

src/libmongoc/src/mongoc/mongoc-topology-scanner.c
```diff
@@ -946,11 +947,17 @@ mongoc_topology_scanner_node_setup_tcp (mongoc_topology_scanner_node_t *node,
    }
 
    if (node->successful_dns_result) {
-      _begin_hello_cmd (node, NULL, false, node->successful_dns_result, 0);
+      _begin_hello_cmd (node,
+                        NULL,
+                        false,
```
chardan, line 952: Since you've done it elsewhere, it might also be nice to annotate the field, YMMV.

chardan (approved):
+1, this looks good! (In my current changes, I'm also having to add more parameters to this function, I'm starting to wonder if it's worth an "options" parameter or something.)

src/libmongoc/src/mongoc/mongoc-topology-scanner.c
kevinAlbs: Done.

PR #916, opened 2021-12-17
CDRIVER-4086 Implement srvServiceName for initial DNS seedlist and SRV polling

eramongodb:
This PR implements support for the `srvServiceName` URI option.

Excluding a minor drive-by update to the `MONGOC_URI_SRVMAXHOSTS` macro (to lowercase for consistency with all other URI option macros), the implementation was relatively straightforward thanks to setup of relevant codepaths in CDRIVER-4097. Most of the work involved [implementing RFC 6335 compliance checks](https://github.com/mongodb/mongo-c-driver/commit/08fd175b340496c01b9f236fba62b4b1be135771#diff-dbe89a1dfeb43b78de3fd2a755697ea837b1e6f6cc8fd589476e64d4fb873700R615) during URI option parsing.

Spec tests were [updated](https://github.com/mongodb/mongo-c-driver/commit/0b0d4b3cdd1465f4d91f4a2cb349ea7607fb98ae) with srvServiceName tests (deliberately omitted during implementation of CDRIVER-4097), including another addition to the [error filter list](https://github.com/mongodb/mongo-c-driver/commit/0b0d4b3cdd1465f4d91f4a2cb349ea7607fb98ae#diff-de273c6f735cdbe82c59b0bd26c6157ba527804980ccdf42bbcb6b2988379e3bR105-R106) pending resolution of CDRIVER-3167.

The existing `mongoc_uri_get_service()` function is named and documented as yielding the "SRV service name", but this is incorrect: it yields the [`{hostname}.{domainname}` portion](https://github.com/mongodb/mongo-c-driver/blob/d67527c6b733e11e88e69240c57cc39471942054/src/libmongoc/src/mongoc/mongoc-topology.c#L465) of the SRV URI. The function name was not updated to preserve the existing public interface (although the documentation was fixed). If permitted, I would rename it to `mongoc_uri_get_srv_host_and_domain()`.

Instead, a new [`mongoc_uri_get_srv_service_name()` function](https://github.com/mongodb/mongo-c-driver/commit/08fd175b340496c01b9f236fba62b4b1be135771#diff-1e21269b0bb676e4ff0db3699dbce0f1f656d924cc49f16a78c36d2ccc63985aR115) was added for the purpose of obtaining the service name. This function is used to [implement `srvServiceName` support](https://github.com/mongodb/mongo-c-driver/commit/9ec9d8a16d35b62875a7680f00fa8d1313c1c5cf) during initial DNS seedlist discovery and SRV polling.


jmikola:
> The function name was not updated to preserve the existing public interface (although the documentation was fixed). If permitted, I would rename it to mongoc_uri_get_srv_host_and_domain().

Is it feasible to introduce a new alias and deprecate the old function name? I believe that has been done elsewhere (e.g. recent changes to libbson's atomic API).

eramongodb:
@jmikola That's a good idea. It would also be a good opportunity to clean up some naming issues internally. Will add relevant changes to PR (can easily be removed if decided not worth).

kevinAlbs (approved):
Great work! Thank you for the thorough description, clear commits, and clarifications (argument names and adding the alias `mongoc_uri_get_srv_hostname`).

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -595,7 +595,7 @@ _mongoc_get_rr_search (const char *service,
  */
 
 bool
-_mongoc_client_get_rr (const char *service,
+_mongoc_client_get_rr (const char *hostname,
```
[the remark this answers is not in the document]
eramongodb: Good spot. Will fix.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -610,6 +610,67 @@ mongoc_uri_parse_auth_mechanism_properties (mongoc_uri_t *uri, const char *str)
    return true;
 }
 
+
+static bool
+mongoc_uri_check_srv_service_name (mongoc_uri_t *uri, const char *str)
+{
+   /* 63 character DNS query limit, excluding prepended underscore. */
+   const size_t mongoc_srv_service_name_max = 62;
```
jmikola, line 618: Is there any reason you're using the `u` suffix (unsigned) on literals below but not here?

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -2197,6 +2265,30 @@ mongoc_uri_get_service (const mongoc_uri_t *uri)
...
+
+   BSON_ASSERT_PARAM (uri);
+
+   if (bson_iter_init_find_case (
+          &iter, &uri->options, MONGOC_URI_SRVSERVICENAME) &&
+       BSON_ITER_HOLDS_UTF8 (&iter)) {
+      return bson_iter_utf8 (&iter, NULL);
+   }
+
+   return "mongodb";
```
jmikola, line 2288: Should we use a private constant for this?

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -610,6 +610,67 @@ mongoc_uri_parse_auth_mechanism_properties (mongoc_uri_t *uri, const char *str)
...
+   /* Initial DNS Seedlist Discovery Spec: This option specifies a valid SRV
+    * service name according to RFC 6335, with the exception that it may exceed
+    * 15 characters as long as the 63rd (62nd with prepended underscore)
+    * character DNS query limit is not surpassed. */
+   if (length > mongoc_srv_service_name_max) {
+      return false;
+   }
+
+   /* RFC 6335: MUST be at least 1 character. */
+   if (length == 0) {
```
jmikola, line 639: Also noticed you're using `0` here but `0u` in another comparison below.

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -462,8 +463,9 @@ mongoc_topology_new (const mongoc_uri_t *uri, bool single_threaded)
          MONGOC_TOPOLOGY_MIN_RESCAN_SRV_INTERVAL_MS;
 
       /* a mongodb+srv URI. try SRV lookup, if no error then also try TXT */
-      prefixed_service = bson_strdup_printf ("_mongodb._tcp.%s", service);
-      if (!topology->rr_resolver (prefixed_service,
+      prefixed_hostname = bson_strdup_printf (
+         "_%s._tcp.%s", mongoc_uri_get_srv_service_name (uri), srv_hostname);
```
jmikola, line 467: This is the more significant change, correct? I'm glancing past a lot of the diff pertaining to the renaming of "service" to "hostname" and wanted to confirm this is more significant.

src/libmongoc/src/mongoc/mongoc-uri.c
eramongodb: In an attempt to keep sign consistency. It is not as important in initializers and comparison expressions (particularly when the value is clearly within range of representable values), but lack-of may result in unexpected well-defined overflow (or lack thereof) in arithmetic expressions. However, for full/proper consistency, I should indeed be using the `u` suffix here as well.

src/libmongoc/src/mongoc/mongoc-uri.c
eramongodb: Good spot. Will update/fix.

src/libmongoc/src/mongoc/mongoc-uri.c
eramongodb: I do not believe it is being used anywhere else, but it would not hurt to give it an explicit name. Will add.

Note: `"mongodb"` also happens to be the default "service name" for [GSSAPI](https://github.com/mongodb/specifications/blob/3a4a6e0f318545c9f050a0000fbd53fc4d9a2790/source/auth/auth.rst#gssapi), but I do not believe these to be relevant enough to justify sharing a constant variable.

src/libmongoc/src/mongoc/mongoc-topology.c
eramongodb: Correct. This is the actual implementation of the `srvServiceName` feature for initial DNS seedlist discovery. Similarly for SRV polling on line line 819. The rest of the "service" -> "hostname" edits are just syntactic changes for consistency.

Note: if the diffs are overwhelming, I recommend comparing the individual commit diffs for the [feature implementation](https://github.com/mongodb/mongo-c-driver/pull/916/commits/9ec9d8a16d35b62875a7680f00fa8d1313c1c5cf) and [renaming](https://github.com/mongodb/mongo-c-driver/pull/916/commits/b03ccd8163a7df911587b7a07177fd7786d046ad) separately, as I deliberately made sure to keep them in separate commits.

src/libmongoc/src/mongoc/mongoc-uri.c
jmikola: Ah, I didn't realize `"mongodb"` was also hard-coded in `mongoc_uri_finalize_auth`. I suppose if you want to be consistent with that you can declare the constant within this function instead of statically within the file -- but I won't press the issue.

I originally assumed this would warrant `#define` in the private header, but it looks like that simply isn't done anywhere.

jmikola (approved):

PR #917, opened 2021-12-17
CDRIVER-4247 Replace cmakedefine01 with autotools-compatible equivalent

eramongodb:
Per CDRIVER-4247 description, the `#cmakedefine01` preprocessor directive breaks PHPC compatibility, which compiles the C driver using autotools rather than CMake. This PR substitutes the CMake-specific directive with an equivalent and autotools-compatible pattern already being used by similar preprocessor macros.


vector-of-bool (approved):
Ouch. LGTM.

jmikola (approved):
Thanks for the quick response on this!

kevinAlbs (approved):
LGTM, thank you for the quick fix.

PR #918, opened 2021-12-20
CDRIVER-4246 Disable CSE tasks on Power8 variant

eramongodb:
Build of the Python package `cryptography`, required by Python package `pykmip`, required by the [drivers-evergreen-tools mock KMIP server](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/013e525aa9d5167184a6a46f875c1fc19cce582e/.evergreen/csfle/kms_kmip_server.py), required by CSE tests in the C driver, fails due to [introduction of Rust toolchain dependency in the cryptography library](https://cryptography.io/en/latest/faq/#why-does-cryptography-require-rust). Several workarounds were attempted (updating `pip` before package upgrade, pinning `cryptography` package, and `CRYPTOGRAPHY_DONT_BUILD_RUST=1`) but unsuccessful on the Power8 variant. Installing the Rust toolchain was considered, but due to FLE support on Power8 not being a tested, supported, or requested combination to begin with, it was decided that dropping this combination on Evergreen was an acceptable workaround.


kevinAlbs (approved):
Thanks for the quick fix!

rcsanchez97 (approved):
LGTM

PR #919, opened 2021-12-21
CDRIVER-4250: Expose internal validation as mongoc_uri_finalize_options

jmikola:
https://jira.mongodb.org/browse/CDRIVER-4250

I originally considered using a name like `is_valid` or `validate` but realized these do more than just _check_ the options. `mongoc_uri_finalize_tls` and `mongoc_uri_finalize_auth` actually modify the URI in certain cases.

I was fairly liberal with the `seealso` links and placed them on any function that can be used to alter the URI after it is constructed/parsed. Finalization technically isn't necessary after setting something like read prefs on a URI, but I felt it didn't hurt to be consistent (esp. if logic changes down the line). I also decided _not_ to reference the new function on the client and pool constructors since it wouldn't apply to all (or even the most common) use cases.

### DIFF
--- a/src/libmongoc/doc/mongoc_uri_t.rst
+++ b/src/libmongoc/doc/mongoc_uri_t.rst
@@ -283,6 +283,7 @@ MONGOC_URI_SAFE                            safe                              {tr
 
     mongoc_uri_copy
     mongoc_uri_destroy
+    mongoc_uri_finalize_options
     mongoc_uri_get_auth_mechanism
     mongoc_uri_get_auth_source
     mongoc_uri_get_compressors
--- a/src/libmongoc/src/mongoc/mongoc-uri.c
+++ b/src/libmongoc/src/mongoc/mongoc-uri.c
@@ -1620,9 +1620,8 @@ mongoc_uri_parse (mongoc_uri_t *uri, const char *str, bson_error_t *error)
       }
    }
 
-   /* TODO (CDRIVER-3723) Consider moving all "finalize" calls into one function
-    * that is additionally called after initial SRV and TXT records are applied
-    * in mongoc_topology_new. */
+   /* TODO (CDRIVER-3723) Consider calling mongoc_uri_finalize_options after 
+    * initial SRV and TXT records are applied in mongoc_topology_new. */
    if (!mongoc_uri_finalize_options (uri, error)) {
       goto error;
    }
@@ -3306,7 +3305,7 @@ mongoc_uri_finalize_srv (const mongoc_uri_t *uri, bson_error_t *error)
 bool
 mongoc_uri_finalize_options (mongoc_uri_t *uri, bson_error_t *error)
 {
-   BSON_ASSERT (uri);
+   BSON_ASSERT_PARAM (uri);
 
    if (!mongoc_uri_finalize_tls (uri, error)) {
       return false;
--- a/src/libmongoc/src/mongoc/mongoc-uri.c
+++ b/src/libmongoc/src/mongoc/mongoc-uri.c
@@ -1404,7 +1404,7 @@ mongoc_uri_finalize_auth (mongoc_uri_t *uri,
 {
    bson_iter_t iter;
    const char *source = NULL;
-   bool require_auth = uri->username != NULL;
+   const bool require_auth = uri->username != NULL;
 
    if (bson_iter_init_find_case (
           &iter, &uri->credentials, MONGOC_URI_AUTHSOURCE)) {
--- a/src/libmongoc/src/mongoc/mongoc-uri.c
+++ b/src/libmongoc/src/mongoc/mongoc-uri.c
@@ -1400,11 +1400,11 @@ mongoc_uri_finalize_tls (mongoc_uri_t *uri, bson_error_t *error)
 
 static bool
 mongoc_uri_finalize_auth (mongoc_uri_t *uri,
-                          bson_error_t *error,
-                          bool require_auth)
+                          bson_error_t *error)
 {
    bson_iter_t iter;
    const char *source = NULL;
+   bool require_auth = uri->username != NULL;
 
    if (bson_iter_init_find_case (
           &iter, &uri->credentials, MONGOC_URI_AUTHSOURCE)) {
@@ -1569,7 +1569,6 @@ mongoc_uri_parse (mongoc_uri_t *uri, const char *str, bson_error_t *error)
 {
    char *before_slash = NULL;
    const char *tmp;
-   bool require_auth = false;
 
    if (!bson_utf8_validate (str, strlen (str), false /* allow_null */)) {
       MONGOC_URI_ERROR (error, "%s", "Invalid UTF-8 in URI");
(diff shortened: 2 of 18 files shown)


src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -3320,3 +3302,31 @@ mongoc_uri_finalize_srv (const mongoc_uri_t *uri, bson_error_t *error)
 
    return true;
 }
+
+bool
+mongoc_uri_finalize_options (mongoc_uri_t *uri, bson_error_t *error)
+{
+   BSON_ASSERT (uri);
```
kevinAlbs, line 3309:
```suggestion
   BSON_ASSERT_PARAM (uri);
```

src/libmongoc/doc/mongoc_uri_t.rst
```diff
@@ -283,6 +283,7 @@ MONGOC_URI_SAFE                            safe                              {tr
 
     mongoc_uri_copy
     mongoc_uri_destroy
+    mongoc_uri_finalize_options
```
kevinAlbs, line 286: I think this `mongoc_uri_finalize_options.rst` is missing from the PR.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -1624,24 +1623,7 @@ mongoc_uri_parse (mongoc_uri_t *uri, const char *str, bson_error_t *error)
    /* TODO (CDRIVER-3723) Consider moving all "finalize" calls into one function
     * that is additionally called after initial SRV and TXT records are applied
     * in mongoc_topology_new. */
```
kevinAlbs, line 1625:
```suggestion
   /* TODO (CDRIVER-3723) Consider calling mongoc_uri_finalize_options after 
    * initial SRV and TXT records are applied in mongoc_topology_new. */
```

src/libmongoc/doc/mongoc_uri_t.rst
jmikola: Just added. Thanks for catching that.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -1621,27 +1620,9 @@ mongoc_uri_parse (mongoc_uri_t *uri, const char *str, bson_error_t *error)
...
-   if (!mongoc_uri_finalize_directconnection (uri, error)) {
-      goto error;
-   }
-
-   if (!mongoc_uri_finalize_loadbalanced (uri, error)) {
-      goto error;
-   }
-
-   if (!mongoc_uri_finalize_srv (uri, error)) {
+   /* TODO (CDRIVER-3723) Consider calling mongoc_uri_finalize_options after
```
eramongodb, line 1623: Can this update also be applied to [`mongoc_topology_new()`](https://github.com/mongodb/mongo-c-driver/blob/a203bafab7d412ed05f33b99229bacbc043c2268/src/libmongoc/src/mongoc/mongoc-topology.c#L515) to resolve CDRIVER-3723 and address/remove this comment?
```diff
-   if (!mongoc_uri_finalize_loadbalanced (topology->uri,
-                                          &topology->scanner->error)) {
-      topology->valid = false;
-   }
-
-   if (!mongoc_uri_finalize_srv (topology->uri, &topology->scanner->error)) {
-      topology->valid = false;
-   }
+   if (!mongoc_uri_finalize_options (topology->uri,
+                                     &topology->scanner->error)) {
+      topology->valid = false;
+   }
```

eramongodb (commented):
> I originally considered using a name like `is_valid` or `validate` but realized these do more than just check the options. `mongoc_uri_finalize_tls` and `mongoc_uri_finalize_auth` actually modify the URI in certain cases.

I would prefer properly separating the "finalization" of TLS/auth options from the "validation" of option correctness so we can expose a non-modifying `mongoc_uri_validate_options()` in the public interface (`const mongoc_uri_t* uri` parameter) instead of the currently-proposed modifying one. Thoughts on making the addition effort to refactor out the modifying/finalizing code from read-only validation?

Alternatively, we could defer the work of refactoring and adding a proper non-modifying `validate_options()` function as followup work (new ticket), but at the cost of ending up with a deprecated `finalize_options()` function in the public interface.

src/libmongoc/src/mongoc/mongoc-uri.c
jmikola: I considered that, but erred on the side of minimizing changes for this PR. Two concerns:

 * Swapping to `mongoc_uri_finalize_options` could technically change behavior. The `loadbalanced` and `srv` functions are read-only and take a `const mongoc_uri_t*` but `mongoc_uri_finalize_options` could possibly modify the struct via the `tls` and `auth` functions.
 * [CDRIVER-3723](https://jira.mongodb.org/browse/CDRIVER-3723) is still going to require more comprehensive changes, so this would be incomplete.

jmikola:
> I would prefer properly separating the "finalization" of TLS/auth options from the "validation" of option correctness so we can expose a non-modifying mongoc_uri_validate_options() in the public interface (const mongoc_uri_t* uri parameter) instead of the currently-proposed modifying one. Thoughts on making the addition effort to refactor out the modifying/finalizing code from read-only validation?

I'm not sure there's any benefit to a `mongoc_uri_validate_options` function in the public API, as that would still put the onus on wrapping drivers (or applications) using that function to manually apply the modifications done by `finalize_tls` and `finalize_auth` (e.g. implicitly enabling TLS for SRV URIs, forcing `authSource=$external` for GSSAPI and X509).

If we all agree that this `mongoc_uri_finalize_options` will become obsolete once [CDRIVER-3723](https://jira.mongodb.org/browse/CDRIVER-3723) is implemented, then perhaps I could abandon this PR and redirect my efforts towards that ticket instead. Since `mongoc_client_new_from_uri` takes a `const mongoc_uri_t*`, we could never have CDRIVER-3723 finalize the `mongoc_uri_t` parameter the same way we're doing here; however, it's probably sufficient to just apply that finalization to the internal URI copies. I think that would pertain to `topology->uri` and `pool->uri` (`client->uri` is just a copy of `topology->uri`).

Thoughts?


 the same way we're doing here; however, I suppose it would only require  theoretically

eramongodb (approved):
> that would still put the onus on wrapping drivers (or applications) using that function to manually apply the modifications

I (again) mistakenly assumed the means and methods for applying individual URI options were/can be accompanied by corresponding immediate modifications as being applied in `finalize()`. Consider my comments above to be a product of misunderstanding the requirements and disregard.

> `mongoc_uri_finalize_options` will become obsolete once CDRIVER-3723 is implemented

Similarly, disregard; I misunderstood the requirements. The justifications you provided are sound. LGTM as proposed.

kevinAlbs (approved):
> If we all agree that this `mongoc_uri_finalize_options` will become obsolete once [CDRIVER-3723](https://jira.mongodb.org/browse/CDRIVER-3723) is implemented, then perhaps I could abandon this PR and redirect my efforts towards that ticket instead. Since `mongoc_client_new_from_uri` takes a `const mongoc_uri_t*`, we could never have [CDRIVER-3723](https://jira.mongodb.org/browse/CDRIVER-3723) finalize the `mongoc_uri_t` parameter the same way we're doing here; however, it's probably sufficient to just apply that finalization to the internal URI copies. I think that would pertain to `topology->uri` and `pool->uri` (`client->uri` is just a copy of `topology->uri`).
> 
> Thoughts?

I agree that `mongoc_uri_finalize_options` will become obsolete once CDRIVER-3723 is complete. The proposed `mongoc_client_new_with_error` / `mongoc_client_pool_new_with_error` could call a private `mongoc_uri_finalize_options` on the topology's copy of the URI after applying options through TXT records.

I slightly prefer closing this in favor of CDRIVER-3723 to limit public API.

Changes LGTM. I will leave it to you whether you would like to address CDRIVER-3723.

jmikola:
@kevinAlbs I'll let this PR sit for a moment and attempt to implement [CDRIVER-3723](https://jira.mongodb.org/browse/CDRIVER-3723) today in a separate PR. We can decide later, but I'm all for avoiding the introduction of a public API that would soon become obsolete.

jmikola:
Closing this in favor of https://github.com/mongodb/mongo-c-driver/pull/921 for [CDRIVER-3723](https://jira.mongodb.org/browse/CDRIVER-3723).

PR #920, opened 2021-12-23
CDRIVER-4249 Apply _DEFAULT_SOURCE to all subdirectories

eramongodb:
Per CDRIVER-4249, this PR cleans up feature test macros used to enable non-conforming C language extensions such as POSIX and BSD library functions.

The `_BSD_SOURCE` macro is a [deprecated alias](https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/include/features.h#L178) of `_DEFAULT_SOURCE`, so it was removed in favor of the already-present `_DEFAULT_SOURCE`.

There was a choice available as to whether to use `_DEFAULT_SOURCE` or `_GNU_SOURCE`. The difference being (according to [GNU documentation](https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html)):
- `_GNU_SOURCE`: everything is included: ISO C89, ISO C99, POSIX.1, POSIX.2, BSD, SVID, X/Open, LFS, and GNU extensions. In the cases where POSIX.1 conflicts with BSD, the POSIX definitions take precedence.
- `_DEFAULT_SOURCE`: most features are included apart from X/Open, LFS and GNU extensions: the effect is to enable features from the 2008 edition of POSIX, as well as certain BSD and SVID features without a separate feature test macro to control them.

The documentation recommends `_GNU_SOURCE` for new programs. However, `_GNU_SOURCE` preferring POSIX definitions over BSD seems to cause [`strerror_r()` compatibility/detection issues on Darwin](https://evergreen.mongodb.com/lobster/evergreen/task/mongo_c_driver_darwin_debug_compile_aws_patch_ab65efbfe5442289d6f9b4180a79cb76f6979bf6_61c4970e9ccd4e2173dfc6d5_21_12_23_15_34_57/0/task#bookmarks=0%2C272&l=1&shareLine=219) that are not present with `_DEFAULT_SOURCE`. Because `_DEFAULT_SOURCE` does not prompt these issues while still enabling the set of features required and used by the C driver, `_DEFAULT_SOURCE` was selected and `_GNU_SOURCE` removed.

To permit validating changes on Darwin, this PR includes a drive-by fix of build failures on Darwin due to #890. Since this fix requires modifying the zlib bundle (last updated [in 2018](https://github.com/mongodb/mongo-c-driver/commit/9cfe2ea2a6025b6b0eada34a4782886e84def594)), I would appreciate some confirmation as to whether the changes made are appropriate. If we wish to leave the zlib sources unmodified, we can choose to selectively exclude `_DEFAULT_SOURCE` during compilation of the zlib bundle instead.


eramongodb:
@jmikola Regarding PHPC, the changes in this PR suggest that defining `_DEFAULT_SOURCE` alone should (hopefully) suffice for building the entire C driver via autotools. Please let me know if you observe otherwise.

jmikola (approved):
> Regarding PHPC, the changes in this PR suggest that defining _DEFAULT_SOURCE alone should (hopefully) suffice for building the entire C driver via autotools. Please let me know if you observe otherwise.

I can confirm that replacing `-D_XOPEN_SOURCE=700` with `-D_DEFAULT_SOURCE` in our bundled CFLAGS allows libmongoc to once again compile without error. I'll defer to Kevin on the rest of this PR, but this LGTM.

src/zlib-1.2.11/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ check_include_file(stddef.h    HAVE_STDDEF_H)
 #
 # Check to see if we have large file support
 #
+cmake_push_check_state()
```
kevinAlbs, line 30: > To permit validating changes on Darwin, this PR includes a drive-by fix of build failures on Darwin due to #890. Since this fix requires modifying the zlib bundle (last updated in 2018), I would appreciate some confirmation as to whether the changes made are appropriate. If we wish to leave the zlib sources unmodified, we can choose to selectively exclude _DEFAULT_SOURCE during compilation of the zlib bundle instead.

I prefer leaving the zlib sources unmodified. It more easily enables updating the bundled zlib. E.g. if zlib has a priority patch release (e.g. to fix a security bug) it may simplify updating the bundled zlib.

kevinAlbs (changes requested):
Thank you for the drive-by fix for Darwin tasks with zlib! I request not modifying the zlib source directly, but otherwise LGTM.

kevinAlbs (approved):

src/libmongoc/CMakeLists.txt
```diff
@@ -99,6 +99,12 @@ if (NOT ENABLE_ZLIB STREQUAL "OFF")
    include (CheckIncludeFiles)
    check_include_files ("unistd.h" HAVE_UNISTD_H)
    check_include_files ("stdarg.h" HAVE_STDARG_H)
+   if (HAVE_UNISTD_H)
+      add_definitions (-DHAVE_UNISTD_H)
+   endif ()
+   if (HAVE_STDARG_H)
+      add_definitions (-DHAVE_STDARG_H)
+   endif ()
```
jmikola, line 107: Offhand, PHP was [already defining `HAVE_UNISTD_H` for bundled zlib](https://github.com/mongodb/mongo-php-driver/blob/1.12.0/config.m4#L434). It was trivial to make the change for `HAVE_STDARG_H`, although I didn't notice any build failure omitting it on Ubuntu 20.04.

LGTM!

jmikola (approved):

src/libmongoc/CMakeLists.txt
eramongodb: (Lack of) `<stdarg.h>` was not causing issues, as far as I could observe. `HAVE_STDARG_H` was just added to compliment the addition of `HAVE_UNISTD_H` for consistency. The (lack of) `<unistd.h>` was the relevant issue being addressed.

PR #921, opened 2021-12-23
CDRIVER-3723 finalize URI during client/pool construction

jmikola:
https://jira.mongodb.org/browse/CDRIVER-3723
https://jira.mongodb.org/browse/CDRIVER-4251

Introduces new `mongoc_client_new_from_uri_with_error` and `mongoc_client_pool_new_with_error` methods. Also addresses topology leak in mongoc_client_new_from_uri (CDRIVER-4251).

As a behavioral change, clients and pools can no longer be constructed with an invalid topology. This means that failures during SRV lookup and/or URI finalization will result in the client/pool constructors returning null.

This PR is an alternative to https://github.com/mongodb/mongo-c-driver/pull/919.


src/libmongoc/doc/mongoc_client_pool_new.rst
```diff
@@ -21,5 +21,13 @@ Parameters
 Returns
 -------
 
-A newly allocated :symbol:`mongoc_client_pool_t` that should be freed with :symbol:`mongoc_client_pool_destroy()` when no longer in use.
+A newly allocated :symbol:`mongoc_client_pool_t` that should be freed with :symbol:`mongoc_client_pool_destroy()` when no longer in use. On error, ``NULL`` is returned and an error may be logged.
```
jmikola, line 24: `mongoc_client_new` and `mongoc_client_new_from_uri` both returned `NULL` on error; however, `mongoc_client_pool_new` always returned a pool. I suppose we can't start returning `NULL` on error here since that would introduce a BC break.

Hopefully we _can_ do so for `mongoc_client_pool_new_with_error` since that'd be a new method. Let me know your thoughts.


Returning `NULL` for This may need to be changed

src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -98,6 +98,21 @@ _mongoc_client_pool_set_internal_tls_opts (
...
 mongoc_client_pool_t *
 mongoc_client_pool_new (const mongoc_uri_t *uri)
+{
+   mongoc_client_pool_t *pool;
+   bson_error_t error = {0};
+
+   if (!(pool = mongoc_client_pool_new_with_error (uri, &error)) &&
+       0 != strcmp (error.message, "")) {
+      MONGOC_ERROR ("%s", error.message);
+   }
```
jmikola, line 108: Per my comment on `mongoc_client_pool_new.rst`, I may need to change this such that `mongoc_client_pool_new` _always` returns a non-NULL pointer to avoid a BC break. Either way, I think it'd be sensible to keep the behavior that logs an error if we have something to report.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -982,48 +982,36 @@ _mongoc_client_recv (mongoc_client_t *client,
...
- *
- * Returns:
- *       A newly allocated mongoc_client_t or NULL if @uri_string is
- *       invalid.
- *
- * Side effects:
- *       None.
- *
- *--------------------------------------------------------------------------
- */
```
jmikola, line 1004: AFAIK, these doc blocks are obsolete. I didn't see any reason to update this, especially for a public method that has its own docs page.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1084,54 +1072,49 @@ mongoc_client_set_ssl_opts (mongoc_client_t *client,
...
    mongoc_topology_t *topology;
 
    topology = mongoc_topology_new (uri, true);
 
+   if (!topology->valid) {
+      if (error) {
+         memcpy (error, &topology->scanner->error, sizeof (bson_error_t));
+      }
+
+      mongoc_topology_destroy (topology);
```
jmikola, line 1103: This addresses the previous memory leak ([CDRIVER-4251](https://jira.mongodb.org/browse/CDRIVER-4251)).

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1084,54 +1072,49 @@ mongoc_client_set_ssl_opts (mongoc_client_t *client,
...
- *
- * Side effects:
- *       None.
- *
- *--------------------------------------------------------------------------
- */
 
 mongoc_client_t *
-_mongoc_client_new_from_uri (mongoc_topology_t *topology)
+_mongoc_client_new_from_topology (mongoc_topology_t *topology)
```
jmikola, line 1117: I'm not sure why this originally had the name `_mongoc_client_new_from_uri`, but this seems more clear. It's also private and was only called in a few places so there should be no risk to renaming it.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1140,6 +1123,7 @@ _mongoc_client_new_from_uri (mongoc_topology_t *topology)
    const char *appname;
 
    BSON_ASSERT (topology);
+   BSON_ASSERT (topology->valid);
```
jmikola, line 1126: I can remove this if there's a concern it may break something; however, I've added `topology->valid` checks in the various functions that would construct a topology before calling this method and I don't believe we'd ever enter this function with an invalid topology.

That said, I'm not sure what the code path is for `mongoc_client_pool_new` constructing a pool with an invalid topology. If the application then tries to pop a client from that pool, is it possible that `_mongoc_client_new_from_topology` (previously `_mongoc_client_new_from_uri`) gets called with an invalid topology?

src/libmongoc/src/mongoc/mongoc-uri-private.h
```diff
@@ -76,27 +76,8 @@ _mongoc_uri_init_scram (const mongoc_uri_t *uri,
...
-bool
-mongoc_uri_finalize_loadbalanced (const mongoc_uri_t *uri, bson_error_t *error);
-
-/* mongoc_uri_finalize_srv validates constraints for SRV URI options.
- * For example, it is invalid to have loadBalanced=true with srvMaxHosts=1.
- * This is expected to be called whenever URI options may change (e.g.
- * parsing a new URI or applying TXT records).
- * Returns false and sets @error on failure.
- * Returns true and does not modify @error on success.
- */
```
jmikola, line 97: This comments were adapted into the new comment for `mongoc_uri_finalize` in `mongoc-uri.c`.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -982,48 +982,36 @@ _mongoc_client_recv (mongoc_client_t *client,
...
 mongoc_client_t *
 mongoc_client_new (const char *uri_string)
 {
-   mongoc_topology_t *topology;
    mongoc_client_t *client;
-   mongoc_uri_t *uri;
+   bson_error_t error = {0};
+
+   if (!(client = mongoc_client_new_with_error (uri_string, &error)) &&
+       0 != strcmp (error.message, "")) {
```
eramongodb, line 992:
```suggestion
   if (!(client = mongoc_client_new_with_error (uri_string, &error))) {
```
`mongoc_client_new_with_error ()` should return `NULL` on error, implying `error.code != 0`, so the `strcmp()` seems unnecessary.

src/libmongoc/src/mongoc/mongoc-client.c
eramongodb:
```suggestion
   
   if (!topology->valid) {
      MONGOC_ERROR ("cannot construct client from invalid topology");
      return NULL;
   }
```
In this case, I think a warning + `NULL` result would be preferable to an assertion. If `mongoc_topology_new ()` had guaranteed `NULL` result on invalid topology, then I would have been in favor of this assertion, but alas.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1084,54 +1072,49 @@ mongoc_client_set_ssl_opts (mongoc_client_t *client,
...
- */
-
 mongoc_client_t *
 mongoc_client_new_from_uri (const mongoc_uri_t *uri)
+{
+   mongoc_client_t *client;
+   bson_error_t error = {0};
+
+   if (!(client = mongoc_client_new_from_uri_with_error (uri, &error)) &&
+       0 != strcmp (error.message, "")) {
```
eramongodb, line 1082:
```suggestion
   if (!(client = mongoc_client_new_from_uri_with_error (uri, &error))) {
```
`mongoc_client_new_from_uri_with_error ()` should return `NULL` on error, implying `error.code != 0`, so the `strcmp()` seems unnecessary.

src/libmongoc/doc/mongoc_client_pool_new.rst
eramongodb: I do not think that is the case. There is enough evidence throughout both prior definition and usage sites that support users being expected to test for `NULL`. I think this is a case of outdated/misleading documentation that needs to be updated, rather than backwards-compatibility we need to maintain.

src/libmongoc/src/mongoc/mongoc-client-pool.c
eramongodb: I think this looks good as written.

eramongodb (changes requested):

src/libmongoc/doc/mongoc_client_new.rst
```diff
@@ -21,9 +21,14 @@ Parameters
 Returns
 -------
 
-A newly allocated :symbol:`mongoc_client_t` if the URI parsed successfully, otherwise ``NULL``.
+A newly allocated :symbol:`mongoc_client_t` that should be freed with :symbol:`mongoc_client_destroy()` when no longer in use. On error, ``NULL`` is returned and an error will be logged.
+
+.. warning::
+
+  Failure to handle the result of this function is a programming error.
```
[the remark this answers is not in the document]
jmikola: I copied this from other doc pages for functions annotated with `BSON_GNUC_WARN_UNUSED_RESULT`. I'm honestly not sure if this text is relevant or not. Happy to remove if we have consensus to do so.

src/libmongoc/doc/mongoc_client_pool_new.rst
kevinAlbs: I agree the documentation does not match the current behavior. `mongoc_client_pool_new` currently returns `NULL` if [attempting to set TLS options without TLS enabled](](https://github.com/mongodb/mongo-c-driver/blob/a203bafab7d412ed05f33b99229bacbc043c2268/src/libmongoc/src/mongoc/mongoc-client-pool.c#L113-L119)).

src/libmongoc/doc/mongoc_client_new.rst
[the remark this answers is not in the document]
kevinAlbs: I agree the warning is redundant.

Between consistency and less redundancy I prefer consistency. They can all be removed later in a follow-up PR.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1084,54 +1071,48 @@ mongoc_client_set_ssl_opts (mongoc_client_t *client,
...
+
+      return NULL;
+   }
+
    /* topology->uri may be different from uri: if this is a mongodb+srv:// URI
     * then mongoc_topology_new has fetched SRV and TXT records and updated its
     * uri from them.
     */
-   return _mongoc_client_new_from_uri (topology);
+   return _mongoc_client_new_from_topology (topology);
```
kevinAlbs, line 1110: I suggest storing the return value and freeing `topology` if `_mongoc_client_new_from_topology` returns `NULL`. Otherwise this may leak `topology`.

src/libmongoc/doc/mongoc_client_new_with_error.rst
```diff
@@ -0,0 +1,35 @@
...
+mongoc_client_new_with_error()
+==============================
+
+Synopsis
+--------
+
+.. code-block:: c
+
+  mongoc_client_t *
+  mongoc_client_new_with_error (const char *uri_string, bson_error_t *error)
```
kevinAlbs, line 12: I vote to not including `mongoc_client_new_with_error`.

IMO it does not provide much convenience over `mongoc_client_new_from_uri_with_error`.

`mongoc_client_pool_new` accepts a `mongoc_uri_t`. There are no requests I know of to add a variant that accepts a string.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -982,48 +982,35 @@ _mongoc_client_recv (mongoc_client_t *client,
...
- */
 mongoc_client_t *
 mongoc_client_new (const char *uri_string)
 {
-   mongoc_topology_t *topology;
    mongoc_client_t *client;
-   mongoc_uri_t *uri;
+   bson_error_t error = {0};
 
+   if (!(client = mongoc_client_new_with_error (uri_string, &error))) {
```
kevinAlbs, line 991: The [/Client/max_staleness](https://spruce.mongodb.com/task/mongo_c_driver_asan_ubuntu_debug_compile_asan_clang_openssl_patch_a203bafab7d412ed05f33b99229bacbc043c2268_61c6018d850e6143937a0bfb_21_12_24_17_21_19/logs?execution=0) test is failing due to this change. The test expects a log warning log on an invalid URI string with `mongoc_client_new`.

I do not think changing the log level is particularly harmful. I suggest updating the test to expect an error log level.

kevinAlbs (changes requested):
Thank you for making this change. I think this is preferable to #919.

I suggest removing `mongoc_client_new_with_error` and have a proposed test fix.

src/libmongoc/src/mongoc/mongoc-client.c
jmikola: Thanks. The two cases where `NULL` is returned would be from an invalid topology (addressed above) and TLS not being supported (not yet addressed). Will fix this now.

src/libmongoc/doc/mongoc_client_new_with_error.rst
jmikola: Removed.

src/libmongoc/src/mongoc/mongoc-client.c
jmikola: `mongoc-uri.c` already logs warnings for unsupported values, so I think this may highlight a case where we're actually logging a redundant error message at a different log level. I'm not familiar with how `ASSERT_CAPTURED_LOG` works but I'll look into it and report back.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -982,48 +982,27 @@ _mongoc_client_recv (mongoc_client_t *client,
...
+   bson_error_t error = {0};
 
    if (!uri_string) {
       uri_string = "mongodb://127.0.0.1/";
    }
 
-   if (!(uri = mongoc_uri_new (uri_string))) {
+   if (!(uri = mongoc_uri_new_with_error (uri_string, &error))) {
+      /* Log URI errors as a warning for consistency with mongoc_uri_new */
+      MONGOC_WARNING ("Error parsing URI: '%s'", error.message);
```
jmikola, line 998: Per the [Connection String spec](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#values), URI errors should be logged as warnings. This change keeps `mongoc_client_new` consistent with previous versions of libmongoc, where the function only relied on behavior within `mongoc_uri_new`. Additionally, this fixes the "/Client/max_staleness" test failure @kevinAlbs noted in https://github.com/mongodb/mongo-c-driver/pull/921#discussion_r775611909

src/libmongoc/doc/mongoc_client_new.rst
```diff
@@ -21,9 +21,10 @@ Parameters
 Returns
 -------
 
-A newly allocated :symbol:`mongoc_client_t` if the URI parsed successfully, otherwise ``NULL``.
+A newly allocated :symbol:`mongoc_client_t` that should be freed with :symbol:`mongoc_client_destroy()` when no longer in use. On error, ``NULL`` is returned and an error or warning will be logged.
```
jmikola, line 24: Added "or warning" here for the case where a URI error is emitted. See https://github.com/mongodb/mongo-c-driver/pull/921/files#r775994888 for context.

src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -112,12 +126,27 @@ mongoc_client_pool_new (const mongoc_uri_t *uri)
 
 #ifndef MONGOC_ENABLE_SSL
    if (mongoc_uri_get_tls (uri)) {
-      MONGOC_ERROR ("Can't create SSL client pool,"
-                    " SSL not enabled in this build.");
+      bson_set_error (error,
+                      MONGOC_ERROR_COMMAND,
+                      MONGOC_ERROR_COMMAND_INVALID_ARG,
+                      "Can't create SSL client pool, SSL not enabled in this "
+                      "build.");
```
jmikola, line 133: I missed this in a previous review, but there is no reason for `mongoc_client_pool_new_with_error` to log an error directly. `mongoc_client_pool_new` will already do so (preserving behavior) so this should merely set the error on the `bson_error_t` to allow the caller to interpret it as they wish.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1141,6 +1119,14 @@ _mongoc_client_new_from_uri (mongoc_topology_t *topology)
 
    BSON_ASSERT (topology);
 
+   if (!topology->valid) {
+      /* This should not be reached if mongoc_client_new_from_uri_with_error and
+       * mongoc_client_pool_new_with_error raise errors when mongoc_topology_new
+       * returns an invalid topology. */
```
jmikola, line 1125: This comment relates to @eramongodb's previous comment in https://github.com/mongodb/mongo-c-driver/pull/921#discussion_r774763909. Asserting `topology->valid` may be too strong, even though I don't believe this will ever be reached. I assume there's no harm in the comment -- perhaps someone can come back to this down the line for some refactoring.

src/libmongoc/tests/test-mongoc-max-staleness.c
```diff
@@ -57,9 +57,15 @@ test_mongoc_client_max_staleness (void)
    capture_logs (true);
    ASSERT (!test_framework_client_new (
       "mongodb://a/?" MONGOC_URI_MAXSTALENESSSECONDS "=120", NULL));
+   ASSERT_CAPTURED_LOG (MONGOC_URI_MAXSTALENESSSECONDS "=120",
+                        MONGOC_LOG_LEVEL_WARNING,
+                        "Invalid readPreferences");
+
+   capture_logs (true);
```
jmikola, line 64: The test was previously using `ASSERT_CAPTURED_LOG` once for two separate calls to `test_framework_client_new`. This ensures each is tested independently, since `capture_logs()` resets the log each time.

src/libmongoc/tests/test-mongoc-max-staleness.c
```diff
@@ -57,9 +57,15 @@ test_mongoc_client_max_staleness (void)
...
    ASSERT (!test_framework_client_new (
       "mongodb://a/?" MONGOC_URI_READPREFERENCE
-      "=primary&" MONGOC_URI_MAXSTALENESSSECONDS "=120", NULL));
+      "=primary&" MONGOC_URI_MAXSTALENESSSECONDS "=120",
+      NULL));
```
jmikola, line 68: This and the diff below are the result of `clang-format`.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1084,54 +1063,53 @@ mongoc_client_set_ssl_opts (mongoc_client_t *client,
...
    topology = mongoc_topology_new (uri, true);
 
+   if (!topology->valid) {
+      if (error) {
+         memcpy (error, &topology->scanner->error, sizeof (bson_error_t));
+      }
+
+      mongoc_topology_destroy (topology);
+
+      return NULL;
```
kevinAlbs, line 1096: Returning `NULL` on failure to apply SRV records looks like the cause of `/initial_dns_seedlist_discovery/null_error_pointer` test failures in test-mongoc-dns.c. [Evergreen link](https://evergreen.mongodb.com/lobster/evergreen/task/mongo_c_driver_asan_ubuntu_debug_compile_asan_clang_openssl_patch_a203bafab7d412ed05f33b99229bacbc043c2268_61cb42c1d1fe0720f5f9a13f_21_12_28_17_00_50/0/task#bookmarks=0%2C11782&l=1&shareLine=11769).

I suggest removing the ``/initial_dns_seedlist_discovery/null_error_pointer` test. It tests calls on an invalid topology. But an invalid topology is no longer accessible to a user.

src/libmongoc/doc/mongoc_client_new.rst
```diff
@@ -21,9 +21,10 @@ Parameters
...
 
 .. seealso::
 
+  | :symbol:`mongoc_client_new_with_error()`
```
kevinAlbs, line 28:
```suggestion

```

src/libmongoc/doc/mongoc_client_t.rst
```diff
@@ -73,6 +73,8 @@ Example
     mongoc_client_get_write_concern
     mongoc_client_new
     mongoc_client_new_from_uri
+    mongoc_client_new_from_uri_with_error
+    mongoc_client_new_with_error
```
kevinAlbs, line 77:
```suggestion

```

src/libmongoc/src/mongoc/mongoc-client.c
kevinAlbs: Good catch, I was not aware the Connection String spec was noted the log level.

src/libmongoc/tests/test-mongoc-loadbalanced.c
```diff
@@ -230,23 +230,16 @@ test_loadbalanced_client_uri_validation (void *unused)
...
-                                       tmp_bson ("{'ping': 1}"),
-                                       NULL /* read prefs */,
-                                       NULL /* reply */,
-                                       &error);
+   BSON_ASSERT (!client);
    ASSERT_ERROR_CONTAINS (error,
-                          MONGOC_ERROR_SERVER_SELECTION,
-                          MONGOC_ERROR_SERVER_SELECTION_FAILURE,
+                          MONGOC_ERROR_COMMAND,
+                          MONGOC_ERROR_COMMAND_INVALID_ARG,
```
jmikola, line 238: This addresses a separate [test failure](https://spruce.mongodb.com/task/mongo_c_driver_asan_ubuntu_debug_compile_asan_clang_openssl_patch_a203bafab7d412ed05f33b99229bacbc043c2268_61cb7099850e6143937d4c52_21_12_28_20_16_28/logs?execution=0).

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -1394,17 +1394,23 @@ mongoc_uri_finalize_tls (mongoc_uri_t *uri, bson_error_t *error)
       return false;
    }
 
+#ifndef MONGOC_ENABLE_SSL
+   if (mongoc_uri_get_tls (uri)) {
+      MONGOC_URI_ERROR (error, "SSL is not enabled in this build");
+      return false;
+   }
+#endif
```
jmikola, line 1402: Is there any concern moving this check from `mongoc_client_pool_new` and `_mongoc_client_new_from_uri` to `mongoc_uri_finalize_tls` (called via `mongoc_uri_finalize`)? It did require changing a test that previously expected `mongoc_client_pool_new` to return `NULL` (that test wasn't asserting a logged error but probably should have).

This also feels more consistent because now the error message is relayed via `bson_error_t` instead of being logged directly via the `MONGOC_ERROR` macro.

jmikola: This seems to introduce some test failures, so I'm going to revert the change. One particular failure is in `test_srv_polling_mocked`, which expects to construct a `mongoc_uri_t` but never a client or pool. There's really no reason that test should require SSL to be enabled.

jmikola:
Regarding the two most recent build failures:

 * [debug-compile-sasl-openssl](https://spruce.mongodb.com/task/mongo_c_driver_windows_2015_debug_compile_sasl_openssl_patch_a203bafab7d412ed05f33b99229bacbc043c2268_61ccad50c9ec4410a96dea83_21_12_29_18_47_46/logs?execution=0) does not fail locally, and this is a mock server test. I wonder if this is a flaky test specific to Windows.
 * [test-asan-4.4-sharded-auth-nosasl-openssl](https://spruce.mongodb.com/task/mongo_c_driver_asan_ubuntu_test_asan_4.4_sharded_auth_nosasl_openssl_patch_a203bafab7d412ed05f33b99229bacbc043c2268_61ccad50c9ec4410a96dea83_21_12_29_18_47_46) is a new flaky test being tracked in [CDRIVER-4252](https://jira.mongodb.org/browse/CDRIVER-4252).

src/libmongoc/doc/mongoc_client_new.rst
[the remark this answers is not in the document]
jmikola: These lines have been removed. Feel free to resolve.

src/libmongoc/src/mongoc/mongoc-client.c
jmikola: That test and another that were relying on returning a client/pool with an invalid topology have been removed. Feel free to resolve.

src/libmongoc/src/mongoc/mongoc-uri.c
kevinAlbs: > There's really no reason that test should require SSL to be enabled.

Agreed. Removing `?tls=true` from the test's URI still passes locally for me.

kevinAlbs:
> Regarding the two most recent build failures:
> 
> * [debug-compile-sasl-openssl](https://spruce.mongodb.com/task/mongo_c_driver_windows_2015_debug_compile_sasl_openssl_patch_a203bafab7d412ed05f33b99229bacbc043c2268_61ccad50c9ec4410a96dea83_21_12_29_18_47_46/logs?execution=0) does not fail locally, and this is a mock server test. I wonder if this is a flaky test specific to Windows.

It looks unrelated to these changes. Filed CDRIVER-4257.

kevinAlbs (approved):
Changes LGTM

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1084,54 +1063,69 @@ mongoc_client_set_ssl_opts (mongoc_client_t *client,
...
+      RETURN (NULL);
+   }
+
    /* topology->uri may be different from uri: if this is a mongodb+srv:// URI
     * then mongoc_topology_new has fetched SRV and TXT records and updated its
     * uri from them.
     */
-   return _mongoc_client_new_from_uri (topology);
+   if (!(client = _mongoc_client_new_from_topology (topology))) {
+      mongoc_topology_destroy (topology);
```
eramongodb, line 1120: May be worth adding a comment/documentation to `_mongoc_client_new_from_topology()` that ownership of `topology` remains with the caller on failure.

src/libmongoc/doc/mongoc_client_t.rst
```diff
@@ -73,6 +73,7 @@ Example
     mongoc_client_get_write_concern
     mongoc_client_new
     mongoc_client_new_from_uri
+    mongoc_client_new_from_uri_with_error
```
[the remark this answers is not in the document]
jmikola: Why are you suggesting this be removed? `mongoc_client_new_from_uri_with_error.rst` and `mongoc_client_pool_new_with_error.rst` are both being added in this PR.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1084,54 +1063,69 @@ mongoc_client_set_ssl_opts (mongoc_client_t *client,
...
+
+      mongoc_topology_destroy (topology);
+
+      RETURN (NULL);
+   }
+
    /* topology->uri may be different from uri: if this is a mongodb+srv:// URI
     * then mongoc_topology_new has fetched SRV and TXT records and updated its
     * uri from them.
     */
```
jmikola, line 1118: @kevinAlbs @eramongodb: Does this comment still provide any value? It'd didn't seem very relevant _before_ I started this PR and I reckon it is even less relevant now.

In the case of this function `uri` is just a parameter and its ownership is always retained by the caller.

src/libmongoc/doc/mongoc_client_t.rst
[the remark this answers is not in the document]
eramongodb: Disregard; confused `mongoc_client_new_from_uri_with_error` with `mongoc_client_new_with_error`.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1084,54 +1063,64 @@ mongoc_client_set_ssl_opts (mongoc_client_t *client,
...
+         memcpy (error, &topology->scanner->error, sizeof (bson_error_t));
+      }
+
+      mongoc_topology_destroy (topology);
+
+      RETURN (NULL);
+   }
+
+   client = _mongoc_client_new_from_topology (topology);
+   BSON_ASSERT (client);
```
jmikola, line 1116: @eramongodb: I added this assertion for consistency with those you added in `mongoc-client-pool.c`. Since we already check `topology->valid` earlier in this function, there is no case where `_mongoc_client_new_from_topology` should actually fail.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1140,13 +1129,7 @@ _mongoc_client_new_from_uri (mongoc_topology_t *topology)
...
 
    BSON_ASSERT (topology);
-
-#ifndef MONGOC_ENABLE_SSL
-   if (mongoc_uri_get_tls (topology->uri)) {
-      MONGOC_ERROR ("Can't create SSL client, SSL not enabled in this build.");
-      return NULL;
-   }
-#endif
+   BSON_ASSERT (topology->valid);
```
jmikola, line 1132: After @eramongodb suggested some assertions on this function's return value, I realized that the code is always checking `topology->valid` before invoking `_mongoc_client_new_from_topology`. There should be no case where we call this with an invalid topology, and asserting a non-null return value is equivalent to asserting the one condition that would have resulted in such a return value.

It seemed simpler to just remove the error condition and assert directly. I don't see the harm in retaining asserts on the returned `client` pointer, but I'm open to removing those if you like.

@eramongodb: This change also avoids any need to document ownership of the `topology` on failure, since this function will no longer fail (short of a `SIGABRT`),

eramongodb: @jmikola Sweet. I think it would then be appropriate to add a comment/documentation to `_mongoc_client_new_from_topology()` that the provided topology must be valid as a precondition.

kevinAlbs (approved):

src/libmongoc/src/mongoc/mongoc-client.c
jmikola: Precondition is now documented.

Another point to mention: whether or not `client` ends up _owning_ the `topology` still depends on whether the client is single-threaded (see: `mongoc_client_destroy`). So I'm glad we don't have to add any comment about `topology` ownership.

src/libmongoc/src/mongoc/mongoc-client.c
jmikola: I ended up removing this comment.

PR #924, opened 2022-01-05
update NEWS for 1.20.1

kevinAlbs:


eramongodb (approved):

chardan (approved):
+1

NEWS
```diff
@@ -1,3 +1,16 @@
...
+
+Bug fixes:
+
+  * Fix bug where first hello command on a single-threaded client may not include full handshake.
+
+Thanks to everyone who contributed to the development of this release.
+
+  * Kevin Albertson
+
+
```
chardan, line 12: This differs a little from the previous versions, in case that's an issue for tooling. (I'm completely fine with it.)

kevinAlbs: Fixed.

PR #927, opened 2022-01-18
CDRIVER-4237 - Address incompatible pointer warnings on GCC 4.8.5

eramongodb:
This PR aims to resolve CDRIVER-4237, depended on by [PHP-2013](https://jira.mongodb.org/browse/PHPC-2013).

~Four~ Three types of warnings are addressed by this PR (as described by GCC 4.8.5):
- ISO C does not allow extra `;` outside of a function [-Wpedantic]
- passing argument `{N}` of `{func}` from incompatible pointer type [enabled by default]
- pointer targets in passing argument `{N}` of `{func}` differ in signedness [-Wpointer-sign]
- ~empty macro arguments are undefined in ISO C90 and ISO C++98 [enabled by default]~

Followup changes in addition to the current set may need to be added to fully resolve [PHP-2013](https://jira.mongodb.org/browse/PHPC-2013). These will be added as demanded in response to code review.

### DIFF
--- a/src/libbson/src/bson/bson-atomic.h
+++ b/src/libbson/src/bson/bson-atomic.h
@@ -343,7 +343,7 @@ enum bson_memory_order {
 
 #if defined(_MSC_VER) || defined(BSON_USE_LEGACY_GCC_ATOMICS)
 /* MSVC expects precise types for their atomic intrinsics. */
-DECL_ATOMIC_INTEGRAL (int8, char, 8);
+DECL_ATOMIC_INTEGRAL (int8, char, 8)
 DECL_ATOMIC_INTEGRAL (int16, short, 16)
 #if !defined(BSON_EMULATE_INT32)
 DECL_ATOMIC_INTEGRAL (int32, long, )
--- a/src/libbson/src/bson/bson-atomic.h
+++ b/src/libbson/src/bson/bson-atomic.h
@@ -342,14 +342,24 @@ enum bson_memory_order {
    DECL_ATOMIC_INTEGRAL (Name, Name##_t, VCSuffix)
 
 #if defined(_MSC_VER) || defined(BSON_USE_LEGACY_GCC_ATOMICS)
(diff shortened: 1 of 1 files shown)


eramongodb:
Reverted "fixes" to empty macro arguments that were not sufficiently tested for correctness.

vector-of-bool (approved):
The changes look simple enough. Is it really that the compilers mishandle typedefs for their intrinsics? That surprises me.

My only question is about formatting: The prior (wonky) formatting came from my clang-format frantically trying to enforce the 80 column limit (which is much too small, imo), so the the prior formatting satisfied the .clang-format in the repository. What version of clang-format are you using? Or did you do some other formatting formatting?

kevinAlbs (approved):
Thank you for the fix. The current changes LGTM. Please re-request review if there are followup changes.

eramongodb:
> What version of clang-format are you using? Or did you do some other formatting formatting?

@vector-of-bool I am using `.clang-format` as provided by the C driver repository. Normally I use the clang-format executable as provided via [VS Code's C/C++ extension](https://code.visualstudio.com/docs/cpp/cpp-ide#_code-formatting). Running the system `clang-format` executable (clang version 10.0.0-4ubuntu1) from project root with the command `clang-format --style=file -i src/libbson/src/bson/bson-atomic.h` yields the same result.

jmikola (approved):
This patch fixed PHP builds on zSeries. PPC is still failing for unrelated reasons ([tasks](https://spruce.mongodb.com/version/61e854925623437c62b492ee/tasks)), but I've noted the successful PPC builds for libmongoc ([tasks](https://spruce.mongodb.com/version/61e73fb0d6d80a6488f0c070/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&variant=%5Epower8-rhel81%24)) so I have no objections to merging.

PR #930, opened 2022-01-20
CDRIVER-3992 - Use OP_MSG if a server API version is requested

chardan:
When a client has selected a specific server API, OPCODE_MSG should be used instead of OPCODE_QUERY (legacy) for the initial handshake.

https://jira.mongodb.org/browse/CDRIVER-3992

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-async-cmd.c
+++ b/src/libmongoc/src/mongoc/mongoc-async-cmd.c
@@ -146,9 +146,10 @@ _mongoc_async_cmd_init_send (const mongoc_opcode_t cmd_opcode,
 
    acmd->ns = bson_strdup_printf ("%s.$cmd", dbname);
 
-   if (MONGOC_OPCODE_QUERY == cmd_opcode) {
 /* JFW: DELETEME: this appears to be needed by both OPCODE_QUERY and OPCODE_MSG:
        acmd->ns = bson_strdup_printf ("%s.$cmd", dbname); */
+   if (MONGOC_OPCODE_QUERY == cmd_opcode) {
+      acmd->ns = bson_strdup_printf ("%s.$cmd", dbname);
 
       acmd->rpc.header.opcode = MONGOC_OPCODE_QUERY;
       acmd->rpc.query.flags = MONGOC_QUERY_SECONDARY_OK;
--- a/src/libmongoc/src/mongoc/mongoc-async-cmd.c
+++ b/src/libmongoc/src/mongoc/mongoc-async-cmd.c
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-present MongoDB, Inc.
+ * Copyright 2014 MongoDB, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -23,7 +23,6 @@
 #include "mongoc-error.h"
 #include "mongoc-opcode.h"
 #include "mongoc-rpc-private.h"
-#include "mongoc-flags-private.h"
 #include "mongoc-stream-private.h"
 #include "mongoc-server-description-private.h"
 #include "mongoc-topology-scanner-private.h"
@@ -143,12 +142,9 @@ _mongoc_async_cmd_init_send (const mongoc_opcode_t cmd_opcode,
    acmd->rpc.header.request_id = ++acmd->async->request_id;
    acmd->rpc.header.response_to = 0;
 
-// JFW: DELETEME: this appears to be needed by both OPCODE_QUERY and OPCODE_MSG:
-// JFW: Evergreen shows this leaking, but we do destroy it in mongoc_async_cmd_destroy(). Never reached?
    acmd->ns = bson_strdup_printf ("%s.$cmd", dbname);
 
    if (MONGOC_OPCODE_QUERY == cmd_opcode) {
-
       acmd->rpc.header.opcode = MONGOC_OPCODE_QUERY;
       acmd->rpc.query.flags = MONGOC_QUERY_SECONDARY_OK;
       acmd->rpc.query.collection = acmd->ns;
@@ -157,7 +153,6 @@ _mongoc_async_cmd_init_send (const mongoc_opcode_t cmd_opcode,
       acmd->rpc.query.query = bson_get_data (&acmd->cmd);
       acmd->rpc.query.fields = NULL;
    }
-
 
    if (MONGOC_OPCODE_MSG == cmd_opcode) {
       acmd->rpc.header.opcode = MONGOC_OPCODE_MSG;
@@ -506,39 +501,3 @@ _mongoc_async_cmd_phase_recv_rpc (mongoc_async_cmd_t *acmd)
 
    return MONGOC_ASYNC_CMD_IN_PROGRESS;
 }
-/*
- * Copyright 2014-2022 MongoDB, Inc.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *   http://www.apache.org/licenses/LICENSE-2.0
- *
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-
-#include <bson/bson.h>
-
-#include "mongoc-client.h"
-#include "mongoc-async-cmd-private.h"
-#include "mongoc-async-private.h"
-#include "mongoc-error.h"
-#include "mongoc-opcode.h"
-#include "mongoc-rpc-private.h"
-#include "mongoc-flags-private.h"
-#include "mongoc-stream-private.h"
-#include "mongoc-server-description-private.h"
-#include "mongoc-topology-scanner-private.h"
-#include "mongoc-log.h"
-#include "utlist.h"
-
-#ifdef MONGOC_ENABLE_SSL
-#include "mongoc-stream-tls.h"
-#endif
--- a/src/libmongoc/src/mongoc/mongoc-async-cmd.c
+++ b/src/libmongoc/src/mongoc/mongoc-async-cmd.c
@@ -222,6 +222,11 @@ mongoc_async_cmd_new (mongoc_async_t *async,
    acmd->connect_started = bson_get_monotonic_time ();
    bson_copy_to (cmd, &acmd->cmd);
 
+   if (MONGOC_OPCODE_MSG == cmd_opcode) {
+      // If we're sending an OPCODE_MSG, we need to add the "db" field:
+      bson_append_utf8 (&acmd->cmd, "$db", 3, "admin", 5);
+   }
+
    _mongoc_array_init (&acmd->array, sizeof (mongoc_iovec_t));
    _mongoc_buffer_init (&acmd->buffer, NULL, 0, NULL, NULL);
 
--- a/src/libmongoc/src/mongoc/mongoc-async-cmd.c
+++ b/src/libmongoc/src/mongoc/mongoc-async-cmd.c
@@ -142,9 +142,8 @@ _mongoc_async_cmd_init_send (const mongoc_opcode_t cmd_opcode,
    acmd->rpc.header.request_id = ++acmd->async->request_id;
    acmd->rpc.header.response_to = 0;
 
-   acmd->ns = bson_strdup_printf ("%s.$cmd", dbname);
-
    if (MONGOC_OPCODE_QUERY == cmd_opcode) {
+      acmd->ns = bson_strdup_printf ("%s.$cmd", dbname);
       acmd->rpc.header.opcode = MONGOC_OPCODE_QUERY;
       acmd->rpc.query.flags = MONGOC_QUERY_SECONDARY_OK;
       acmd->rpc.query.collection = acmd->ns;
--- a/src/libmongoc/src/mongoc/mongoc-async-cmd.c
+++ b/src/libmongoc/src/mongoc/mongoc-async-cmd.c
@@ -223,7 +223,7 @@ mongoc_async_cmd_new (mongoc_async_t *async,
    bson_copy_to (cmd, &acmd->cmd);
 
    if (MONGOC_OPCODE_MSG == cmd_opcode) {
-      // If we're sending an OPCODE_MSG, we need to add the "db" field:
+      /* If we're sending an OPCODE_MSG, we need to add the "db" field: */
       bson_append_utf8 (&acmd->cmd, "$db", 3, "admin", 5);
    }
 
--- a/src/libmongoc/src/mongoc/mongoc-async-cmd.c
+++ b/src/libmongoc/src/mongoc/mongoc-async-cmd.c
@@ -121,7 +121,6 @@ mongoc_async_cmd_run (mongoc_async_cmd_t *acmd)
       return true;
    }
 
-
    duration_usec = bson_get_monotonic_time () - acmd->cmd_started;
 
    if (result == MONGOC_ASYNC_CMD_SUCCESS) {
@@ -144,12 +143,11 @@ _mongoc_async_cmd_init_send (const mongoc_opcode_t cmd_opcode,
    acmd->rpc.header.request_id = ++acmd->async->request_id;
    acmd->rpc.header.response_to = 0;
 
+// JFW: DELETEME: this appears to be needed by both OPCODE_QUERY and OPCODE_MSG:
+// JFW: Evergreen shows this leaking, but we do destroy it in mongoc_async_cmd_destroy(). Never reached?
    acmd->ns = bson_strdup_printf ("%s.$cmd", dbname);
 
-/* JFW: DELETEME: this appears to be needed by both OPCODE_QUERY and OPCODE_MSG:
-       acmd->ns = bson_strdup_printf ("%s.$cmd", dbname); */
    if (MONGOC_OPCODE_QUERY == cmd_opcode) {
-      acmd->ns = bson_strdup_printf ("%s.$cmd", dbname);
 
       acmd->rpc.header.opcode = MONGOC_OPCODE_QUERY;
       acmd->rpc.query.flags = MONGOC_QUERY_SECONDARY_OK;
@@ -171,11 +169,6 @@ _mongoc_async_cmd_init_send (const mongoc_opcode_t cmd_opcode,
       acmd->rpc.msg.sections[0].payload.bson_document =
          bson_get_data (&acmd->cmd);
    }
-
-/* JFW: DELETEME looks like this is already initialized in
-  mongoc_async_cmd_new(): 
-	_mongoc_array_init (&acmd->array, sizeof (mongoc_iovec_t));
-*/
 
    /* This will always be hello, which are not allowed to be compressed */
    _mongoc_rpc_gather (&acmd->rpc, &acmd->array);
--- a/src/libmongoc/src/mongoc/mongoc-async-cmd.c
+++ b/src/libmongoc/src/mongoc/mongoc-async-cmd.c
@@ -144,8 +144,11 @@ _mongoc_async_cmd_init_send (const mongoc_opcode_t cmd_opcode,
    acmd->rpc.header.request_id = ++acmd->async->request_id;
    acmd->rpc.header.response_to = 0;
 
+   acmd->ns = bson_strdup_printf ("%s.$cmd", dbname);
+
    if (MONGOC_OPCODE_QUERY == cmd_opcode) {
-      acmd->ns = bson_strdup_printf ("%s.$cmd", dbname);
+/* JFW: DELETEME: this appears to be needed by both OPCODE_QUERY and OPCODE_MSG:
+       acmd->ns = bson_strdup_printf ("%s.$cmd", dbname); */
 
       acmd->rpc.header.opcode = MONGOC_OPCODE_QUERY;
       acmd->rpc.query.flags = MONGOC_QUERY_SECONDARY_OK;
@@ -168,7 +171,10 @@ _mongoc_async_cmd_init_send (const mongoc_opcode_t cmd_opcode,
          bson_get_data (&acmd->cmd);
    }
 
-   _mongoc_array_init (&acmd->array, sizeof (mongoc_iovec_t));
+/* JFW: DELETEME looks like this is already initialized in
+  mongoc_async_cmd_new(): 
+	_mongoc_array_init (&acmd->array, sizeof (mongoc_iovec_t));
+*/
 
    /* This will always be hello, which are not allowed to be compressed */
    _mongoc_rpc_gather (&acmd->rpc, &acmd->array);
--- a/src/libmongoc/src/mongoc/mongoc-async-cmd.c
+++ b/src/libmongoc/src/mongoc/mongoc-async-cmd.c
@@ -136,38 +136,39 @@ mongoc_async_cmd_run (mongoc_async_cmd_t *acmd)
 }
 
 void
-_mongoc_async_cmd_init_send (const mongoc_opcode_t cmd_opcode, mongoc_async_cmd_t *acmd, const char *dbname)
+_mongoc_async_cmd_init_send (const mongoc_opcode_t cmd_opcode,
+                             mongoc_async_cmd_t *acmd,
+                             const char *dbname)
 {
-   acmd->rpc.header.msg_len = 0; 
+   acmd->rpc.header.msg_len = 0;
    acmd->rpc.header.request_id = ++acmd->async->request_id;
    acmd->rpc.header.response_to = 0;
 
-   if(MONGOC_OPCODE_QUERY == cmd_opcode) {
+   if (MONGOC_OPCODE_QUERY == cmd_opcode) {
+      acmd->ns = bson_strdup_printf ("%s.$cmd", dbname);
 
-   acmd->ns = bson_strdup_printf ("%s.$cmd", dbname);
-
-   acmd->rpc.header.opcode = MONGOC_OPCODE_QUERY;    
-   acmd->rpc.query.flags = MONGOC_QUERY_SECONDARY_OK;
-   acmd->rpc.query.collection = acmd->ns;
-   acmd->rpc.query.skip = 0;
-   acmd->rpc.query.n_return = -1;
-   acmd->rpc.query.query = bson_get_data (&acmd->cmd);
-   acmd->rpc.query.fields = NULL;
+      acmd->rpc.header.opcode = MONGOC_OPCODE_QUERY;
+      acmd->rpc.query.flags = MONGOC_QUERY_SECONDARY_OK;
+      acmd->rpc.query.collection = acmd->ns;
+      acmd->rpc.query.skip = 0;
+      acmd->rpc.query.n_return = -1;
+      acmd->rpc.query.query = bson_get_data (&acmd->cmd);
+      acmd->rpc.query.fields = NULL;
    }
 
 
-if(MONGOC_OPCODE_MSG == cmd_opcode) { 
+   if (MONGOC_OPCODE_MSG == cmd_opcode) {
+      acmd->rpc.header.opcode = MONGOC_OPCODE_MSG;
 
-   acmd->rpc.header.opcode = MONGOC_OPCODE_MSG;    
+      acmd->rpc.msg.msg_len = 0;
+      acmd->rpc.msg.flags = 0;
+      acmd->rpc.msg.n_sections = 1;
+      acmd->rpc.msg.sections[0].payload_type = 0;
+      acmd->rpc.msg.sections[0].payload.bson_document =
+         bson_get_data (&acmd->cmd);
+   }
 
-   acmd->rpc.msg.msg_len = 0;
-   acmd->rpc.msg.flags = 0;
-   acmd->rpc.msg.n_sections = 1;
-   acmd->rpc.msg.sections[0].payload_type = 0;
-   acmd->rpc.msg.sections[0].payload.bson_document = bson_get_data (&acmd->cmd);
-}
-
-_mongoc_array_init(&acmd->array, sizeof(mongoc_iovec_t)); 
+   _mongoc_array_init (&acmd->array, sizeof (mongoc_iovec_t));
 
    /* This will always be hello, which are not allowed to be compressed */
    _mongoc_rpc_gather (&acmd->rpc, &acmd->array);
@@ -541,4 +542,3 @@ _mongoc_async_cmd_phase_recv_rpc (mongoc_async_cmd_t *acmd)
 #ifdef MONGOC_ENABLE_SSL
 #include "mongoc-stream-tls.h"
 #endif
-
--- a/src/libmongoc/src/mongoc/mongoc-client-pool.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-pool.c
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2022 MongoDB, Inc.
+ * Copyright 2013 MongoDB, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/libmongoc/src/mongoc/mongoc-client-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-client-private.h
@@ -244,6 +244,13 @@ mongoc_client_connect (bool buffered,
                        const mongoc_uri_t *uri,
                        const mongoc_host_list_t *host,
                        bson_error_t *error);
+
+
+/* Returns true if a versioned server API has been selected,
+otherwise returns false: */
+bool
+mongoc_client_uses_server_api (const mongoc_client_t *client);
+
 BSON_END_DECLS
 
 #endif /* MONGOC_CLIENT_PRIVATE_H */
--- a/src/libmongoc/src/mongoc/mongoc-client-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-client-private.h
@@ -244,7 +244,6 @@ mongoc_client_connect (bool buffered,
                        const mongoc_uri_t *uri,
                        const mongoc_host_list_t *host,
                        bson_error_t *error);
-
 BSON_END_DECLS
 
 #endif /* MONGOC_CLIENT_PRIVATE_H */
--- a/src/libmongoc/src/mongoc/mongoc-client-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-client-private.h
@@ -247,7 +247,7 @@ mongoc_client_connect (bool buffered,
 
 
 /* Returns true if a versioned server API has been selected,
-otherwise returns false: */
+ * otherwise returns false. */
 bool
 mongoc_client_uses_server_api (const mongoc_client_t *client);
 
--- a/src/libmongoc/src/mongoc/mongoc-client.c
+++ b/src/libmongoc/src/mongoc/mongoc-client.c
@@ -1969,7 +1969,6 @@ _mongoc_client_command_with_opts (mongoc_client_t *client,
 
    BSON_ASSERT (client);
    BSON_ASSERT (db_name);
-   BSON_ASSERT (command);
 
    command_name = _mongoc_get_command_name (command);
    cluster = &client->cluster;
--- a/src/libmongoc/src/mongoc/mongoc-client.c
+++ b/src/libmongoc/src/mongoc/mongoc-client.c
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2022 MongoDB, Inc.
+ * Copyright 2013 MongoDB, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/libmongoc/src/mongoc/mongoc-client.c
+++ b/src/libmongoc/src/mongoc/mongoc-client.c
@@ -1962,6 +1962,7 @@ _mongoc_client_command_with_opts (mongoc_client_t *client,
 
    BSON_ASSERT (client);
    BSON_ASSERT (db_name);
+   BSON_ASSERT (command);
 
    command_name = _mongoc_get_command_name (command);
    cluster = &client->cluster;
@@ -3171,4 +3172,3 @@ mongoc_client_uses_server_api (const mongoc_client_t *client)
 {
    return mongoc_topology_uses_server_api (client->topology);
 }
-
--- a/src/libmongoc/src/mongoc/mongoc-client.c
+++ b/src/libmongoc/src/mongoc/mongoc-client.c
@@ -1132,13 +1132,6 @@ _mongoc_client_new_from_topology (mongoc_topology_t *topology)
    BSON_ASSERT (topology);
    BSON_ASSERT (topology->valid);
 
-#ifndef MONGOC_ENABLE_SSL
-   if (mongoc_uri_get_tls (topology->uri)) {
-      MONGOC_ERROR ("Can't create SSL client, SSL not enabled in this build.");
-      return NULL;
-   }
-#endif
-
    client = (mongoc_client_t *) bson_malloc0 (sizeof *client);
    client->uri = mongoc_uri_copy (topology->uri);
    client->initiator = mongoc_client_default_stream_initiator;
--- a/src/libmongoc/src/mongoc/mongoc-client.c
+++ b/src/libmongoc/src/mongoc/mongoc-client.c
@@ -3165,3 +3165,10 @@ mongoc_client_get_handshake_description (mongoc_client_t *client,
    mongoc_server_stream_cleanup (server_stream);
    return sd;
 }
+
+bool
+mongoc_client_uses_server_api (const mongoc_client_t *client)
+{
+   return mongoc_topology_uses_server_api (client->topology);
+}
+
--- a/src/libmongoc/src/mongoc/mongoc-cluster-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-cluster-private.h
@@ -228,6 +228,10 @@ bool
 mongoc_topology_scanner_uses_server_api (
    const mongoc_topology_scanner_t *topology_scanner);
 
+bool
+mongoc_client_session_uses_server_api (
+   const mongoc_client_session_t *client_session);
+
 #ifdef MONGOC_ENABLE_CRYPTO
 void
 _mongoc_cluster_init_scram (const mongoc_cluster_t *cluster,
--- a/src/libmongoc/src/mongoc/mongoc-cluster-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-cluster-private.h
@@ -228,10 +228,6 @@ bool
 mongoc_topology_scanner_uses_server_api (
    const mongoc_topology_scanner_t *topology_scanner);
 
-bool
-mongoc_client_session_uses_server_api (
-   const mongoc_client_session_t *client_session);
-
 #ifdef MONGOC_ENABLE_CRYPTO
 void
 _mongoc_cluster_init_scram (const mongoc_cluster_t *cluster,
--- a/src/libmongoc/src/mongoc/mongoc-cluster-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-cluster-private.h
@@ -216,17 +216,7 @@ _mongoc_cluster_get_auth_cmd_x509 (const mongoc_uri_t *uri,
                                    bson_error_t *error /* OUT */);
 
 bool
-mongoc_topology_uses_server_api (const mongoc_topology_t *topology);
-
-bool
-mongoc_client_uses_server_api (const mongoc_client_t *client);
-
-bool
 mongoc_cluster_uses_server_api (const mongoc_cluster_t *cluster);
-
-bool
-mongoc_topology_scanner_uses_server_api (
-   const mongoc_topology_scanner_t *topology_scanner);
 
 #ifdef MONGOC_ENABLE_CRYPTO
 void
--- a/src/libmongoc/src/mongoc/mongoc-cluster-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-cluster-private.h
@@ -215,6 +215,8 @@ _mongoc_cluster_get_auth_cmd_x509 (const mongoc_uri_t *uri,
                                    bson_t *cmd /* OUT */,
                                    bson_error_t *error /* OUT */);
 
+/* Returns true if a versioned server API has been selected,
+ * otherwise returns false. */
 bool
 mongoc_cluster_uses_server_api (const mongoc_cluster_t *cluster);
 
--- a/src/libmongoc/src/mongoc/mongoc-cluster.c
+++ b/src/libmongoc/src/mongoc/mongoc-cluster.c
@@ -1982,7 +1982,7 @@ _mongoc_cluster_node_new (mongoc_stream_t *stream,
    node->stream = stream;
    node->connection_address = bson_strdup (connection_address);
 
-   /* Note that the node->sd field is set to NULL by bson_malloc(),
+   /* Note that the node->sd field is set to NULL by bson_malloc0(),
    rather than being explicitly initialized. */
 
    return node;
--- a/src/libmongoc/src/mongoc/mongoc-cluster.c
+++ b/src/libmongoc/src/mongoc/mongoc-cluster.c
@@ -836,7 +836,7 @@ _stream_run_hello (mongoc_cluster_t *cluster,
    /* Use OP_QUERY for the handshake, unless the user has specified an
     * API version; the correct hello_cmd has already been selected: */
    if (!mongoc_cluster_uses_server_api (cluster)) {
-      // Complete OPCODE_QUERY setup:
+      /* Complete OPCODE_QUERY setup: */
       hello_cmd.query_flags = MONGOC_QUERY_SECONDARY_OK;
    } else {
       /* We're using OP_MSG, and require some additional doctoring: */
@@ -1982,8 +1982,8 @@ _mongoc_cluster_node_new (mongoc_stream_t *stream,
    node->stream = stream;
    node->connection_address = bson_strdup (connection_address);
 
-   /* Note that the node->sd field is not explicitly initialized
-   by parameter in this function. */
+   /* Note that the node->sd field is set to NULL by bson_malloc(),
+   rather than being explicitly initialized. */
 
    return node;
 }
@@ -2111,7 +2111,7 @@ _cluster_add_node (mongoc_cluster_t *cluster,
        * generation. */
    }
 
-   /* take critical fields from a fresh hello (except the ->sd field): */
+   /* take critical fields from a fresh hello */
    cluster_node = _mongoc_cluster_node_new (stream, host->host_and_port);
 
    handshake_sd = _cluster_run_hello (cluster,
--- a/src/libmongoc/src/mongoc/mongoc-cluster.c
+++ b/src/libmongoc/src/mongoc/mongoc-cluster.c
@@ -2297,6 +2297,10 @@ _mongoc_cluster_stream_for_server (mongoc_cluster_t *cluster,
 
    td = mc_tpld_take_ref (topology);
 
+   fprintf (stderr,
+            "_mongoc_cluster_stream_for_server about to call "
+            "_try_get_server_stream()\n"),
+      fflush (stderr);
    ret_server_stream = _try_get_server_stream (
       cluster, td.ptr, server_id, reconnect_ok, err_ptr);
 
--- a/src/libmongoc/src/mongoc/mongoc-cluster.c
+++ b/src/libmongoc/src/mongoc/mongoc-cluster.c
@@ -821,20 +821,54 @@ _stream_run_hello (mongoc_cluster_t *cluster,
    mongoc_server_description_init (&empty_sd, address, server_id);
    server_stream =
       _mongoc_cluster_create_server_stream (td.ptr, &empty_sd, stream);
+
    mongoc_server_description_cleanup (&empty_sd);
 
    /* Set up the shared parts of the mongo_cmd_t, which will later be converted
    to either an op_msg or op_query: */
    memset (&hello_cmd, 0, sizeof (hello_cmd));
 
-   hello_cmd.server_stream = server_stream;
-   hello_cmd.command_name = _mongoc_get_command_name (&handshake_command);
-   hello_cmd.command = &handshake_command;
 
-   /* Always use OP_QUERY for the handshake, unless the user has specified an
-    * API version; the correct hello_cmd has already been selected. For
-    * OPCODE_QUERY, we have to set up a few additional fields: */
+   hello_cmd.command = &handshake_command;
+   hello_cmd.command_name = _mongoc_get_command_name (&handshake_command);
+   hello_cmd.server_stream = server_stream;
+
+/* JFW: 
+  typedef struct _mongoc_cmd_t {
+     const char *db_name; 
+     mongoc_query_flags_t query_flags;
+     const bson_t *command;
+     const char *command_name;
+     const uint8_t *payload;
+     int32_t payload_size;
+     const char *payload_identifier;
+     mongoc_server_stream_t *server_stream;
+     int64_t operation_id;
+     mongoc_client_session_t *session;
+     mongoc_server_api_t *api;
+     bool is_acknowledged;
+     bool is_txn_finish;
+  } mongoc_cmd_t;
+*/
+
+/* JFW: neither of these seem to matter here:
+ bson_append_utf8 (&handshake_command, "$db", 3, "admin", 5); // JFW: experiment 
+      hello_cmd.db_name = "admin"; //JFW: experiment
+*/
+
+   /* Use OP_QUERY for the handshake, unless the user has specified an
+    * API version; the correct hello_cmd has already been selected: */
    if (!mongoc_cluster_uses_server_api (cluster)) {
+      // Complete OP_MSG setup:
+      hello_cmd.db_name = "admin";
+
+/* JFW: there are some tests which check for this-- is QUERY here meant for OP_QUERY or does it also apply
+ * to OP_MSG? */
+      hello_cmd.query_flags = MONGOC_QUERY_SECONDARY_OK;
+   }
+  else // i.e. OPCODE_QUERY
+   {
+      // Complete OPCODE_QUERY setup:
       hello_cmd.db_name = "admin";
       hello_cmd.query_flags = MONGOC_QUERY_SECONDARY_OK;
    }
@@ -941,15 +975,19 @@ _cluster_run_hello (mongoc_cluster_t *cluster,
                            error);
 
    if (!sd) {
+fprintf(stderr, "JFW: !sd\n");
       return NULL;
    }
 
+fprintf(stderr, "JFW: HAVE sd\n");
    if (sd->type == MONGOC_SERVER_UNKNOWN) {
+fprintf(stderr, "JFW: server type UNKNOWN\n");
       memcpy (error, &sd->error, sizeof (bson_error_t));
       mongoc_server_description_destroy (sd);
       return NULL;
    }
 
+fprintf(stderr, "JFW: sd return *OK*\n");
    return sd;
 }
 
@@ -2006,6 +2044,8 @@ _mongoc_cluster_node_new (mongoc_stream_t *stream,
    node->stream = stream;
    node->connection_address = bson_strdup (connection_address);
 
+fprintf(stderr, "JFW: _mongoc_cluster_node_new(): ->sd field not assigned-to\n"), fflush(stderr);
+
    return node;
 }
 
@@ -2132,7 +2172,7 @@ _cluster_add_node (mongoc_cluster_t *cluster,
        * generation. */
    }
 
-   /* take critical fields from a fresh hello */
+   /* take critical fields from a fresh hello (except the ->sd field): */
    cluster_node = _mongoc_cluster_node_new (stream, host->host_and_port);
 
    handshake_sd = _cluster_run_hello (cluster,
@@ -2143,6 +2183,7 @@ _cluster_add_node (mongoc_cluster_t *cluster,
                                       &speculative_auth_response,
                                       error);
    if (!handshake_sd) {
+fprintf(stderr, "JFW: handshake_sd null\n");
       GOTO (error);
    }
 
@@ -2592,11 +2633,15 @@ _mongoc_cluster_create_server_stream (
    const mongoc_server_description_t *handshake_sd,
    mongoc_stream_t *stream)
 {
+/* JFW: does server_stream_new() actualy copy all the fields..? */
    mongoc_server_description_t *const sd =
       mongoc_server_description_new_copy (handshake_sd);
    /* can't just use mongoc_topology_server_by_id(), since we must hold the
     * lock while copying topology->shared_descr.ptr->logical_time below */
-   return mongoc_server_stream_new (td, sd, stream);
+/*JFW:   return mongoc_server_stream_new (td, sd, stream); */
+mongoc_server_stream_t *ss = mongoc_server_stream_new (td, sd, stream);
+fprintf(stderr, "JFW: HERE\n");
+return ss;
 }
 
 
--- a/src/libmongoc/src/mongoc/mongoc-cluster.c
+++ b/src/libmongoc/src/mongoc/mongoc-cluster.c
@@ -839,6 +839,10 @@ _stream_run_hello (mongoc_cluster_t *cluster,
       // Complete OPCODE_QUERY setup:
       hello_cmd.query_flags = MONGOC_QUERY_SECONDARY_OK;
    }
+   else {
+	/* We're using OP_MSG, and require some additional doctoring: */
+        bson_append_utf8 (&handshake_command, "$db", 3, "admin", 5);	
+   }
 
    if (!mongoc_cluster_run_command_private (
           cluster, &hello_cmd, &reply, error)) {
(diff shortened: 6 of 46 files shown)


src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -305,6 +305,7 @@ mongoc_client_pool_pop (mongoc_client_pool_t *pool)
 again:
    if (!(client = (mongoc_client_t *) _mongoc_queue_pop_head (&pool->queue))) {
       if (pool->size < pool->max_pool_size) {
+         // JFW: client = _mongoc_client_new_from_uri (pool->topology);
```
chardan, line 308: It looks like _mongoc_client_new_from_topology() is correct to me.

src/libmongoc/src/mongoc/mongoc-client-private.h
```diff
@@ -240,6 +240,10 @@ mongoc_client_connect (bool buffered,
                        const mongoc_uri_t *uri,
                        const mongoc_host_list_t *host,
                        bson_error_t *error);
+
+MONGOC_EXPORT (bool)
+mongoc_client_uses_server_api (mongoc_client_t *client);
```
chardan, line 245: Should probably have prefix '_'.

chardan:
Note that I haven't run clang-format on this yet.

src/libmongoc/src/mongoc/mongoc-async-cmd.c
```diff
@@ -134,20 +136,38 @@ mongoc_async_cmd_run (mongoc_async_cmd_t *acmd)
 }
 
 void
-_mongoc_async_cmd_init_send (mongoc_async_cmd_t *acmd, const char *dbname)
+_mongoc_async_cmd_init_send (const mongoc_opcode_t cmd_opcode, mongoc_async_cmd_t *acmd, const char *dbname)
 {
+   acmd->rpc.header.msg_len = 0; 
+   acmd->rpc.header.request_id = ++acmd->async->request_id; /* in _server_monitor_awaitable_hello_send(), this is postfix++ */
```
kevinAlbs, line 142: I think the `postfix++` / `++prefix` is arbitrary. As long as the request_id is unique for each message on a single connection.

src/libmongoc/src/mongoc/mongoc-async-cmd.c
```diff
@@ -134,20 +136,38 @@ mongoc_async_cmd_run (mongoc_async_cmd_t *acmd)
...
    acmd->rpc.query.fields = NULL;
+   }
+
+
+if(MONGOC_OPCODE_MSG == cmd_opcode) { 
+
+   acmd->rpc.header.opcode = MONGOC_OPCODE_MSG;    
+
+   acmd->rpc.msg.msg_len = 0;
+   acmd->rpc.msg.flags = 0; /* JFW: other flags, like MONGOC_MSG_EXHAUST_ALLOWED, as per _server_monitor_awaitable_hello_send(), fail validation */
```
kevinAlbs, line 164: `MONGOC_MSG_EXHAUST_ALLOWED` does not apply for single-threaded monitoring. 

`MONGOC_MSG_EXHAUST_ALLOWED` corresponds to `exhaustAllowed` in [OP_MSG flags](https://docs.mongodb.com/manual/reference/mongodb-wire-protocol/#op_msg)

> The client is prepared for multiple replies to this request using the moreToCome bit. The server will never produce replies with the moreToCome bit set unless the request has this bit set.
>
> This ensures that multiple replies are only sent when the network layer of the requester is prepared for them.

For single-threaded monitoring, the requester is not prepared for multiple replies. After `_mongoc_async_cmd_phase_recv_rpc` completes reading one `OP_REPLY` or `OP_MSG`, it does not continue reading replies. Nor should it. Single-threaded monitoring blocks the main thread to send a `hello` / legacy hello to all servers. 

By "fail validation", I assume the server returned an error when this flag was set. That is likely due to the server expecting `topologyVersion` to be included in the [awaitable hello command](https://github.com/mongodb/specifications/blob/6dc6f80026f0f8d99a8c81f996389534b14f6602/source/server-discovery-and-monitoring/server-monitoring.rst#awaitable-hello-or-legacy-hello-server-specification).

src/libmongoc/src/mongoc/mongoc-client-pool.c
kevinAlbs: Agreed.

src/libmongoc/src/mongoc/mongoc-client-private.h
kevinAlbs: All of these functions are private since they are in a private header. There is already inconsistency for use of `_` prefix on private functions. Either way seems OK to me.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1132,6 +1132,13 @@ _mongoc_client_new_from_topology (mongoc_topology_t *topology)
    BSON_ASSERT (topology);
    BSON_ASSERT (topology->valid);
 
+#ifndef MONGOC_ENABLE_SSL
+   if (mongoc_uri_get_tls (topology->uri)) {
+      MONGOC_ERROR ("Can't create SSL client, SSL not enabled in this build.");
+      return NULL;
+   }
+#endif
```
kevinAlbs, line 1140: This added block looks unintended.

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -677,13 +678,16 @@ mongoc_cluster_run_command_private (mongoc_cluster_t *cluster,
    if (!reply) {
       reply = &reply_local;
    }
+
    server_stream = cmd->server_stream;
-   if (server_stream->sd->max_wire_version >= WIRE_VERSION_OP_MSG) {
+
```
kevinAlbs, line 683: `mongoc_cluster_t` contains a pointer to `mongoc_client_t`. Instead of having a `force_op_msg` bool as a parameter, consider:

```c
bool force_op_msg = mongoc_client_uses_server_api (cluster->client);
```

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -3126,6 +3133,14 @@ mongoc_client_set_server_api (mongoc_client_t *client,
    return true;
 }
 
+/* Returns true if a specific server API has been set: */
+bool mongoc_client_uses_server_api(mongoc_client_t *client)
+{
+ BSON_ASSERT_PARAM(client);
+
+ return NULL != client->api;
```
kevinAlbs, line 3141: `mongoc_client_set_server_api` sets `client->api` and calls `_mongoc_topology_scanner_set_server_api`. `mongoc_client_set_server_api` only applies to single-threaded clients.

`mongoc_client_pool_set_server_api` sets `pool->api` and calls `_mongoc_topology_scanner_set_server_api`. `mongoc_client_pool_set_server_api` only applies to pooled clients. One `mongoc_topology_t` is shared among all pooled clients.

Consider having a single accessor, like `mongoc_topology_uses_server_api`. It can apply to both pooled and single threaded clients. It can check if the `topology->scanner->api` is set.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -319,11 +319,56 @@ _server_monitor_polling_hello (mongoc_server_monitor_t *server_monitor,
...
+   if (!_mongoc_stream_writev_full (server_monitor->stream,
+                                    iovec,
+                                    niovec,
+                                    server_monitor->connect_timeout_ms,
+                                    error)) {
+      MONITOR_LOG_ERROR (
+         server_monitor, "failed to write awaitable hello: %s", error->message);
+      _mongoc_array_destroy (&array_to_write);
+      return false;
+   }
```
kevinAlbs, line 367: This also needs to receive the reply. Compare with `_server_monitor_send_and_recv_opquery`.

src/libmongoc/src/mongoc/mongoc-client.c
chardan: I think that's an artifact of me removing it and then replacing it during a rebase-- it does appear in master.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
chardan: +1 done-- good catch!

src/libmongoc/src/mongoc/mongoc-cluster.c
chardan: I like that function name-- and I agree that it's clear now that parameter doesn't need to come from a distance. Once tests are passing I'll implement this (and your related suggestion to consolidate the various tests for the client having specified a server API).

chardan: I've added a few functions (in the style of overloads) to do this, as sometimes the right path through the nested structs was tricky to find.

src/libmongoc/src/mongoc/mongoc-async-cmd.c
```diff
@@ -484,3 +506,39 @@ _mongoc_async_cmd_phase_recv_rpc (mongoc_async_cmd_t *acmd)
 
    return MONGOC_ASYNC_CMD_IN_PROGRESS;
 }
+/*
+ * Copyright 2014-2022 MongoDB, Inc.
```
chardan, line 510: Should we use XXXX-present, or XXXX-<CURRENT_YEAR>?

src/libmongoc/src/mongoc/mongoc-cluster-private.h
```diff
@@ -215,6 +215,23 @@ _mongoc_cluster_get_auth_cmd_x509 (const mongoc_uri_t *uri,
                                    bson_t *cmd /* OUT */,
                                    bson_error_t *error /* OUT */);
 
+bool
```
chardan, line 218: It would be nice to make these accessors instead of predicates-- although this is a little tider, the ability to /also/ fetch the server API in one step would be pleasant. Maybe something to do after this is "done"!

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -768,9 +870,19 @@ _server_monitor_setup_connection (mongoc_server_monitor_t *server_monitor,
    _mongoc_topology_dup_handshake_cmd (server_monitor->topology, &cmd);
    _server_monitor_append_cluster_time (server_monitor, &cmd);
    bson_destroy (hello_response);
-   if (!_server_monitor_send_and_recv_opquery (
-          server_monitor, &cmd, hello_response, error)) {
-      GOTO (fail);
+
+   /* If the user has select a versioned API, we'll assume OPCODE_MSG;
+   otherwise, we'll use the legacy OPCODE_QUERY: */
+   if (NULL != server_monitor->topology->scanner->api) {
```
chardan, line 876: Should use our shiny new functions here...

src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -1160,6 +1141,92 @@ mock_server_receives_query (mock_server_t *server,
...
+mock_server_receives_hello_op_msg (mock_server_t *server)
+{
+   /*   bson_t *msg = tmp_bson("{'hello': 1, 'helloOk': true, 'maxAwaitTimeMS':
+    * { '$exists': false }}");  */
+   bson_t *msg =
+      tmp_bson ("{'hello': 1, 'maxAwaitTimeMS': { '$exists': false }}");
+
+   return _mock_server_receives_single_msg (server, 0, msg);
+
+   /*
```
chardan, line 1171: I haven't seen this happen in quite a while, now.

src/libmongoc/src/mongoc/mongoc-async-cmd.c
```diff
@@ -134,20 +135,40 @@ mongoc_async_cmd_run (mongoc_async_cmd_t *acmd)
...
-   acmd->rpc.header.opcode = MONGOC_OPCODE_QUERY;
-   acmd->rpc.query.flags = MONGOC_QUERY_SECONDARY_OK;
-   acmd->rpc.query.collection = acmd->ns;
-   acmd->rpc.query.skip = 0;
-   acmd->rpc.query.n_return = -1;
-   acmd->rpc.query.query = bson_get_data (&acmd->cmd);
-   acmd->rpc.query.fields = NULL;
+
+// JFW: DELETEME: this appears to be needed by both OPCODE_QUERY and OPCODE_MSG:
+// JFW: Evergreen shows this leaking, but we do destroy it in mongoc_async_cmd_destroy(). Never reached?
```
chardan, line 147: I think this is a red herring and is ok. (At least, I don't see it with memcheck.)

src/libmongoc/src/mongoc/mongoc-async-cmd.c
kevinAlbs: IIRC `XXXX-present` from DRIVERS-409. IMO no need to update in this PR.

src/libmongoc/src/mongoc/mongoc-async-cmd.c
kevinAlbs: Do you have a link to the failing task? The task failures linked to the PR [ASAN variant](https://spruce.mongodb.com/version/61f47cd43e8e86482507bd11/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&variant=asan-ubuntu) do not indicate leaks and look like occurrences of CDRIVER-4268.

src/libmongoc/src/mongoc/mongoc-async-cmd.c
```diff
@@ -484,3 +506,39 @@ _mongoc_async_cmd_phase_recv_rpc (mongoc_async_cmd_t *acmd)
 
    return MONGOC_ASYNC_CMD_IN_PROGRESS;
 }
+/*
```
kevinAlbs, line 509: I think this copyright notice and included headers are not intended to be included.

kevinAlbs (commented):
If this PR is not meant to be ready for merge, please move it to Draft and leave GitHub comments with open questions.

If this is ready for merge, please do a self-review of this PR to clean up the following:
- Address or remove `JFW:` comments. If you have open questions for reviewers, leave them comments on GitHub.
- `//` comments are removed or updated to `/* ... */` comments. `//` comments break the [debug-compile-c89](https://spruce.mongodb.com/task/mongo_c_driver_releng_debug_compile_c89_patch_66bca4e613c971e328f5299505bd6257dc6bba83_61f47cd43e8e86482507bd11_22_01_28_23_31_34/logs?execution=0) task.
- Run clang-format on the updated files.

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -674,17 +675,25 @@ mongoc_cluster_run_command_private (mongoc_cluster_t *cluster,
...
+/* JFW: BUG?: legacy helloOk should be sent by OPCODE_QUERY no matter what the wire version is,
+ * however: */
+   if (mongoc_cluster_uses_server_api (cluster) ||
```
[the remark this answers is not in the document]
kevinAlbs: The proposed changes of CDRIVER-3992 are:
> Use OP_MSG for the first hello if a server API version is requested.

A legacy hello is not used when server API version is not requested. If a legacy hello is sent over OP_QUERY, that seems outside of scope. The logic here satisfies the requirement of the ticket.

I do not think there is a scenario in which a maxWireVersion indicates OP_MSG is not supported, but the server does not support legacy hello.

src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -981,49 +982,25 @@ _mock_server_receives_msg (mock_server_t *server, uint32_t flags, ...)
...
+                                  uint32_t flags,
+                                  const bson_t *doc)
 {
    request_t *request;
    bool r;
 
+/* JFW: DELETEME unclear to me if this constructs a valid document or not:
+   const bson_t *docs[] = {doc, NULL};
+*/
+
```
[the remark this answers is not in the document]
kevinAlbs: I think that declares an array of pointers to `const bson_t*`. `bson_t` manages a buffer underneath. A `bson_t` has 120 bytes within the struct to stack allocate small BSON documents, and will switch to using the heap when the document exceeds that size.

src/libmongoc/src/mongoc/mongoc-topology-scanner.c
```diff
@@ -295,6 +298,15 @@ _build_handshake_cmd (const bson_t *basis_cmd,
       BSON_APPEND_BOOL (doc, "loadBalanced", true);
    }
 
+   /* JFW: RESOLVEME question: should we be searching the document we're copying for
+    * this field and using any extant value, or is it always "admin"? Also, the check
+    * below doesn't actually detect OPCODE_MSG, just user-selected server API. Should
+    * we just always add it? Is there another way to do this? */
+   // If we're sending an OPCODE_MSG, we need to add the "db" field:
+   if (mongoc_topology_scanner_uses_server_api (ts)) {
```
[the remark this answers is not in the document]
kevinAlbs: > 1) is this always "admin", or should we search the document we're copying for a value?

Good question. I think the answer is "yes". But the only evidence I see is from https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.rst#example-implementation

> reply = conn.send_command("admin", cmd)

--

> 2) this check doesn't really detect OPCODE_MSG, just whether the user has selected an API: should we always assume OP_MSG here? Is there another way to do this?

The proposed changes are to use OP_MSG if a server API version is requested. I think it is OK to assume OP_MSG is supported if `mongoc_topology_scanner_uses_server_api (ts)` returns true.

> (I think we pretty much always need the $db field regardless of OP_MSG or OP_QUERY.)

I disagree. $db is a [global argument](https://github.com/mongodb/specifications/blob/master/source/message/OP_MSG.rst#global-command-arguments) for OP_MSG.

For OP_QUERY, the database is indicated by the `admin.$cmd` value in the `fullCollectionName` field of the protocol: https://docs.mongodb.com/manual/reference/mongodb-wire-protocol/#std-label-wire-op-query. I think the legacy hello or hello ignores extra fields. So a server will not error if `$db` is not recognized as a field in legacy hello.

src/libmongoc/src/mongoc/mongoc-handshake-private.h
```diff
@@ -36,6 +36,9 @@ BSON_BEGIN_DECLS
 #define HANDSHAKE_DRIVER_NAME_MAX 64
 #define HANDSHAKE_DRIVER_VERSION_MAX 32
 
+#define HANDSHAKE_CMD_HELLO "hello"
```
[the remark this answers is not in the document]
kevinAlbs: I do not have strong feelings either way. The `HANDSHAKE_CMD_LEGACY_HELLO` and `HANDSHAKE_RESPONSE_LEGACY_HELLO` were introduced to limit the usage of the terms themselves.

src/libmongoc/tests/mock_server/mock-server.c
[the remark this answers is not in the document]
chardan: That would explain the behavior I saw when I was using it! Thanks.

src/libmongoc/src/mongoc/mongoc-cluster.c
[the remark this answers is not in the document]
chardan: Ah, because maxWireVersion is a /runtime/ value set after handshake. Agreed.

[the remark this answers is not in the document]
chardan: Agreed, that wouldn't be possible. And this is ok as it is /not/ dealing with the initial hello.

src/libmongoc/src/mongoc/mongoc-handshake-private.h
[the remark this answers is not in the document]
vector-of-bool: I think it would be useful to have macros/constants defined for most all of our "magic strings" just to reduce cognitive overhead and chance of introducing typos or spelling errors that can't be caught by a compiler.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -210,6 +210,110 @@ _server_monitor_append_cluster_time (mongoc_server_monitor_t *server_monitor,
...
+   /* Done sending! Now, receive the reply: */
+
+   _mongoc_buffer_init (&buffer, NULL, 0, NULL, NULL);
+
+   if (!_mongoc_buffer_append_from_stream (&buffer,
+                                           server_monitor->stream,
+                                           4,
+                                           server_monitor->connect_timeout_ms,
+                                           error)) {
+      GOTO (fail);
```
vector-of-bool, line 271: The `GOTO` macro is only applicable for the trace logging when the same function uses `ENTRY`, `EXIT`, and `RETURN`.

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -2254,7 +2254,7 @@ test_mongoc_client_mismatched_me (void)
    mock_server_run (server);
    uri = mongoc_uri_copy (mock_server_get_uri (server));
    mongoc_uri_set_option_as_utf8 (uri, "replicaSet", "rs");
-   client = test_framework_client_new_from_uri (uri, NULL);
+   client = mongoc_client_new_from_uri (uri);
```
vector-of-bool, line 2257: Why this change?

src/libmongoc/tests/test-mongoc-versioned-api.c
```diff
@@ -174,6 +174,37 @@ _test_mongoc_server_api_client_pool_once (void)
...
+static void
+_test_mongoc_client_uses_server_api (void)
+{
+   mongoc_client_t *client0; /* versioned */
+   mongoc_client_t *client1; /* not versioned */
+   mongoc_server_api_t *api;
+   bson_error_t error;
+
+   client0 = mongoc_client_new ("mongodb://localhost");
+   client1 = mongoc_client_new ("mongodb://localhost");
```
vector-of-bool, line 186: This should use to `test_framework` APIs to create a client, so that the client behavior can be tweaked by the test runner.

vector-of-bool (commented):
Some comments on the WIP, nothing glaring to change.

src/libmongoc/tests/test-mongoc-topology.c
```diff
@@ -2321,12 +2321,20 @@ _test_hello_versioned_api (bool pooled)
...
    if (pooled) {
       pool = test_framework_client_pool_new_from_uri (uri, api);
+
       client = mongoc_client_pool_pop (pool);
    } else {
       client = test_framework_client_new_from_uri (uri, api);
    }
 
-   hello = bson_strdup_printf ("{'ok': 1,"
+   /* For client pools, the first handshake happens when the client is popped.
```
kevinAlbs, line 2330: Did this section need to move? If not, I suggest moving back to reduce the diff.

src/libmongoc/tests/test-mongoc-versioned-api.c
```diff
@@ -174,6 +174,37 @@ _test_mongoc_server_api_client_pool_once (void)
...
+
+   /* Neither client should have an API set: */
+   BSON_ASSERT (!client0->api);
+   BSON_ASSERT (!client1->api);
```
kevinAlbs, line 190:
```suggestion
   BSON_ASSERT (!mongoc_client_uses_server_api(client0->api));
   BSON_ASSERT (!mongoc_client_uses_server_api(client1->api));
```

src/libmongoc/src/mongoc/mongoc-async-cmd.c
```diff
@@ -134,20 +134,36 @@ mongoc_async_cmd_run (mongoc_async_cmd_t *acmd)
...
-   acmd->rpc.query.collection = acmd->ns;
-   acmd->rpc.query.skip = 0;
-   acmd->rpc.query.n_return = -1;
-   acmd->rpc.query.query = bson_get_data (&acmd->cmd);
-   acmd->rpc.query.fields = NULL;
+
+   acmd->ns = bson_strdup_printf ("%s.$cmd", dbname);
+
+   if (MONGOC_OPCODE_QUERY == cmd_opcode) {
+      acmd->rpc.header.opcode = MONGOC_OPCODE_QUERY;
```
kevinAlbs, line 148:
```suggestion
   if (MONGOC_OPCODE_QUERY == cmd_opcode) {
      acmd->ns = bson_strdup_printf ("%s.$cmd", dbname);
      acmd->rpc.header.opcode = MONGOC_OPCODE_QUERY;
```

`acmd->ns` is only applied to the OP_QUERY `fullCollectionName`. It should be OK to leave `acmd->ns` as NULL for OP_MSG. In `mongoc_async_cmd_destroy`, the call to `bson_free (acmd->ns);` will ignore a NULL value.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1962,7 +1969,6 @@ _mongoc_client_command_with_opts (mongoc_client_t *client,
 
    BSON_ASSERT (client);
    BSON_ASSERT (db_name);
-   BSON_ASSERT (command);
```
kevinAlbs, line 1965: Is removing this assertion necessary?

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -1327,6 +1347,39 @@ _mongoc_cluster_auth_node_x509 (mongoc_cluster_t *cluster,
 #endif
 }
 
+/* True if the client user has requested a specific wire protocol version: */
```
kevinAlbs, line 1350:
```suggestion
/* True if the client user has requested a specific server API version: */
```

I suggest moving these functions definitions to the files associated with the type. Namely:

`mongoc_topology_scanner_uses_server_api` to mongoc-topology-scanner.c
`mongoc_topology_uses_server_api` to mongoc-topology.c
`mongoc_client_uses_server_api` to mongoc-client.c

I suggest removing `mongoc_client_session_uses_server_api` since it is not called.

src/libmongoc/src/mongoc/mongoc-rpc.c
```diff
@@ -993,6 +997,29 @@ _mongoc_rpc_get_first_document (mongoc_rpc_t *rpc, bson_t *reply)
    return false;
 }
 
+/* Reply to a first OP_MSG with another OP_MSG: */
```
kevinAlbs, line 1000: The comment suggests that this function does the reply. I suggest clarifying this parses a BSON document from a reply. 

```suggestion
/* Get the first BSON document from an OP_MSG reply: */
```

src/libmongoc/src/mongoc/mongoc-server-description.c
```diff
@@ -45,7 +45,7 @@ mongoc_server_description_cleanup (mongoc_server_description_t *sd)
 {
    BSON_ASSERT (sd);
```
kevinAlbs, line 47: Please revert changes to this file.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -210,6 +210,110 @@ _server_monitor_append_cluster_time (mongoc_server_monitor_t *server_monitor,
...
+                "sending with timeout %" PRId64,
+                server_monitor->connect_timeout_ms);
+
+   if (!_mongoc_stream_writev_full (server_monitor->stream,
+                                    iovec,
+                                    niovec,
+                                    server_monitor->connect_timeout_ms,
+                                    error)) {
+      MONITOR_LOG_ERROR (
+         server_monitor, "failed to write awaitable hello: %s", error->message);
```
kevinAlbs, line 257:
```suggestion
         server_monitor, "failed to write polling hello: %s", error->message);
```

src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -1160,6 +1135,83 @@ mock_server_receives_query (mock_server_t *server,
...
+{
+   bson_t *msg =
+      tmp_bson ("{'hello': 1, 'maxAwaitTimeMS': { '$exists': false }}");
+
+   return _mock_server_receives_single_msg (server, 0, msg);
+}
+
+/*--------------------------------------------------------------------------
+ *
+ * mock_server_receives_bulk_msg --
```
kevinAlbs, line 1170: I suggest moving `mock_server_receives_bulk_msg` up back to the original location to reduce the diff.

src/libmongoc/tests/mock_server/request.c
```diff
@@ -517,7 +519,7 @@ request_matches_kill_cursors (const request_t *request, int64_t cursor_id)
  *
  * request_matches_msg --
  *
- *       Test that a client OP_MSG matches a pattern. The OP_MSG consists
+ *       Test that client OP_MSGs match a pattern. The OP_MSGs consists
```
kevinAlbs, line 522: I suggest reverting this comment.  This expects one OP_MSG. The OP_MSG may consist of a sequence of documents in multiple sections. See https://docs.mongodb.com/manual/reference/mongodb-wire-protocol/#op_msg.

src/libmongoc/tests/mock_server/request.c
```diff
@@ -556,8 +559,20 @@ request_matches_msg (const request_t *request,
 
       /* make sure the pattern is reasonable, e.g. that we didn't pass a string
        * instead of a bson_t* by mistake */
-      BSON_ASSERT (bson_validate (
-         pattern, BSON_VALIDATE_EMPTY_KEYS | BSON_VALIDATE_UTF8, NULL));
+      if (!bson_validate_with_error (pattern,
+                                     BSON_VALIDATE_EMPTY_KEYS |
+                                        BSON_VALIDATE_UTF8,
+                                     &bson_error)) {
+         fprintf (stderr,
```
kevinAlbs, line 566: Capturing the error from `bson_validate` is a good idea. I suggest using `test_error`, which prints the filename and line number before aborting
```suggestion
         test_error (
```

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -3416,12 +3445,12 @@ test_client_appname (bool pooled, bool use_uri)
 
    mock_server_replies_simple (request, server_reply);
    if (!pooled) {
-      _respond_to_ping (future, server);
+       _respond_to_ping (future, server, false); // JFW: mongoc_client_uses_server_api(client));
```
eramongodb, line 3448: Running the test locally suggests the 3rd argument should be `true`. `mongoc_client_uses_server_api(client)` returns `false`.

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -3341,7 +3370,7 @@ _test_client_sends_handshake (bool pooled)
    request_destroy (request);
 
    if (!pooled) {
-      _respond_to_ping (future, server);
+      _respond_to_ping (future, server, false);
```
eramongodb, line 3373: Running the test locally suggests the 3rd argument should be `true`. `mongoc_client_uses_server_api(client)` returns `false`.

src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -1103,22 +1081,23 @@ mock_server_receives_legacy_hello (mock_server_t *server,
...
+ *       Logs if the current request is a hello command using OP_QUERY.
  *
  *--------------------------------------------------------------------------
  */
 
 request_t *
 mock_server_receives_hello (mock_server_t *server)
 {
+   /* Note that mock_server_receives_command() expects JSON documents, not BSON
+   documents, unlike _mock_server_receives_msg(): */
```
eramongodb, line 1093: This comment is misleading, please remove. `mock_server_receives_command` expects a null-terminated byte string corresponding to a single document, whereas `mock_server_receives_msg` expects a series of `bson_t` objects corresponding to one or more documents. The content of the documents are specified in exactly the same way (no distinction between "JSON" and "BSON" documents between these two functions).

src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -1160,6 +1139,83 @@ mock_server_receives_query (mock_server_t *server,
...
+ *--------------------------------------------------------------------------
+ */
+
+request_t *
+mock_server_receives_hello_op_msg (mock_server_t *server)
+{
+   bson_t *msg =
+      tmp_bson ("{'hello': 1, 'maxAwaitTimeMS': { '$exists': false }}");
+
+   return _mock_server_receives_single_msg (server, 0, msg);
```
eramongodb, line 1165:
```suggestion
   return mock_server_receives_msg (
      server,
      MONGOC_MSG_NONE,
      tmp_bson ("{'hello': 1, 'maxAwaitTimeMS': { '$exists': false }}"));
```

src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -981,49 +982,19 @@ _mock_server_receives_msg (mock_server_t *server, uint32_t flags, ...)
...
- *
- *--------------------------------------------------------------------------
- */
 request_t *
-mock_server_receives_bulk_msg (mock_server_t *server,
-                               uint32_t flags,
-                               const bson_t *msg_pattern,
-                               const bson_t *doc_pattern,
-                               size_t n_docs)
+_mock_server_receives_single_msg (mock_server_t *server,
```
eramongodb, line 986: This function is just a special case of `mock_server_receives_msg` and does not seem necessary. Please remove and replace calls to `_mock_server_receives_single_msg (server, flags, doc)` with `mock_server_receives_msg (server, flags, doc)`.

src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -868,6 +868,7 @@ request_assert_no_duplicate_keys (request_t *request)
 }
 
 
+/* Note that mock-rs.c has another, similar function of the same name: */
```
eramongodb, line 871: This comment is misleading, please remove; the function in `mock-rs.c` is named `mock_rs_receives_request`, as are other mock replica-set counterparts to mock server functions (`mock_server_*` vs. `mock_rs_*`).

src/libmongoc/tests/mock_server/mock-rs.c
```diff
@@ -697,7 +697,6 @@ _mock_rs_receives_msg (mock_rs_t *rs, uint32_t flags, ...)
    return request;
 }
 
-
```
eramongodb, line 700: Unnecessary diff, please remove.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -313,12 +417,12 @@ _server_monitor_polling_hello (mongoc_server_monitor_t *server_monitor,
    bson_t cmd;
    const bson_t *hello;
    bool ret;
-
    hello = _mongoc_topology_scanner_get_monitoring_cmd (
       server_monitor->topology->scanner, hello_ok);
    bson_copy_to (hello, &cmd);
 
    _server_monitor_append_cluster_time (server_monitor, &cmd);
+
```
eramongodb, line 425: Please remove unnecessary diffs in this file.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -210,6 +210,110 @@ _server_monitor_append_cluster_time (mongoc_server_monitor_t *server_monitor,
    }
    mc_tpld_drop_ref (&td);
 }
+static bool
+_server_monitor_send_and_recv_hello_opmsg (
```
eramongodb, line 214: `_server_monitor_send_and_recv_hello_opmsg` and `_server_monitor_send_and_recv_opquery` are completely identical in all respects except for the initialization of RPC data. Consider refactoring the two functions such that initialization of RPC data for OP_MSG vs. OP_QUERY is handled separately from the actual "send and recv" implementation. (Related: see implementation of `_mongoc_async_cmd_init_send`.)

src/libmongoc/src/mongoc/mongoc-handshake-private.h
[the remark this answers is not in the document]
eramongodb: There are several instances of string literal `"hello"` and `"helloOk"` being used throughout both implementation and test code that could be replaced with these new macros. Regardless whether or not these existing literals are substituted with these new macros, I think it may be worth introducing them regardless for symmetry with their legacy counterparts.

src/libmongoc/src/mongoc/mongoc-client.c
eramongodb: This block is _not_ present in the [master branch](https://github.com/mongodb/mongo-c-driver/blob/f6e6a7025d3c65a24e3c32c2763dadf3ef5b804a/src/libmongoc/src/mongoc/mongoc-client.c#L1134), please remove.

src/libmongoc/src/mongoc/mongoc-handshake-private.h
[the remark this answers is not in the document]
chardan: Agreed that it would be good to use these consistently-- it was /very/ confusing (and still easily so). We also have a number of different ways that we "freehand" construct or piecewise construct handshakes-- an audit cataloguing them and consolidating them into straightforward functions would have made this MUCH easier: a great deal of my effort went into figuring out what was slightly wrong, and it was not easy to determine which examples were "right". I think that's a worthwhile project, but I want to avoid touching anything right now simply because this is feature is much behind schedule.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
chardan: I've used it here because _server_monitor_send_and_recv_opquery() also uses it (as far as I can tell also without the other macros!). I can rework that function as well as this one (the op_msg version) to use them, or just bare goto (a slight preference, but I can see the value of of trace logging also). Any preference?

src/libmongoc/tests/test-mongoc-client.c
chardan: This is because even if no server API is specified to test_framework_client_new_from_uri() (i.e.the second parameter is NULL), it still picks up the MONGOC_SERVER_API environment variable. Unfortunately, there are MANY tests only check for legacy hello and fail or abort() if an OP_MSG hello is sent instead. This function, and its char* and pool conterparts, doesn't check for that variable, so we get legacy hello-- as expected.

src/libmongoc/tests/test-mongoc-versioned-api.c
chardan: The next section of code in the test asserts that no server API has been set, and then the test goes on to set the API on one and only one client. Is there a test_framework_* function that will still let me do that?

src/libmongoc/src/mongoc/mongoc-cluster-private.h
```diff
@@ -215,6 +215,23 @@ _mongoc_cluster_get_auth_cmd_x509 (const mongoc_uri_t *uri,
...
+
+bool
+mongoc_cluster_uses_server_api (const mongoc_cluster_t *cluster);
+
+bool
+mongoc_topology_scanner_uses_server_api (
+   const mongoc_topology_scanner_t *topology_scanner);
+
+bool
+mongoc_client_session_uses_server_api (
```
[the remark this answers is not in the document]
chardan: +1, done. (Although I could easily be convinced to leave it there, and add tests for all of these. But it does appear there was no need for it.)

src/libmongoc/tests/mock_server/request.c
```diff
@@ -22,6 +22,8 @@
 #include "../test-conveniences.h"
 #include "../TestSuite.h"
 
+#include "../bsonutil/bson-val.h"
```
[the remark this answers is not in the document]
chardan: +1, done.

src/libmongoc/tests/test-mongoc-async.c
```diff
@@ -153,6 +153,7 @@ test_hello_impl (bool with_ssl)
                             setup_ctx,
                             "admin",
                             &q,
+                            MONGOC_OPCODE_QUERY, /* i.e. legacy hello */
```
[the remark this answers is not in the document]
chardan: +1, that is more exact.

src/libmongoc/src/mongoc/mongoc-server-monitor.c
chardan: Good catch, thank you! Done.

src/libmongoc/tests/mock_server/request.c
chardan: +1

src/libmongoc/tests/mock_server/request.c
chardan: That's a good idea! Done.

src/libmongoc/tests/test-mongoc-versioned-api.c
```diff
@@ -189,4 +220,6 @@ test_client_versioned_api_install (TestSuite *suite)
    TestSuite_Add (suite, "/VersionedApi/copy", _test_mongoc_server_api_copy);
    TestSuite_Add (
       suite, "/VersionedApi/setters", _test_mongoc_server_api_setters);
+   TestSuite_Add (
+      suite, "VersionedApi/private/client_uses_server_api", _test_mongoc_client_uses_server_api);
```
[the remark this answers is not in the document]
chardan: clang-format applied! Thank you.

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -3173,16 +3181,32 @@ _force_hello_with_ping (mongoc_client_t *client, int heartbeat_ms)
...
+_respond_to_ping (future_t *future, mock_server_t *server, bool check_for_op_msg)
 {
    request_t *request;
 
    ASSERT (future);
 
+   if(check_for_op_msg) {
+
+     request = mock_server_receives_msg (
+        server, MONGOC_QUERY_NONE, tmp_bson ("{'ping': 1}"));
```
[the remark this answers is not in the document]
chardan: This should be resolved now.

src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -3173,16 +3181,32 @@ _force_hello_with_ping (mongoc_client_t *client, int heartbeat_ms)
...
+
+/*JFW:
```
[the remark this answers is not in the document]
chardan: This appears to be resolved.

src/libmongoc/tests/test-mongoc-handshake.c
```diff
@@ -284,6 +286,9 @@ test_mongoc_handshake_data_append_null_args (void)
    bson_iter_t inner_iter;
    const char *val;
 
+/* JFW: this test crashes, generating OP_MSG */
```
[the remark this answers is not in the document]
chardan: No, should no longer apply! Thank you.

src/libmongoc/tests/mock_server/mock-rs.c
chardan: Looks like sometimes there's one newline, sometimes two! Reverted.

src/libmongoc/src/mongoc/mongoc-handshake-private.h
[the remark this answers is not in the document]
chardan: I've created a new ticket to address this and related concerns:
https://jira.mongodb.org/browse/CDRIVER-4280

src/libmongoc/src/mongoc/mongoc-server-monitor.c
chardan: I entirely agree. I also really, really don't want to touch anything that risks breaking this right now. But it definitely isn't needed and possibly may even cause extra confusion-- ditto the bare literals vs. macros. I have filed a ticket for this:
https://jira.mongodb.org/browse/CDRIVER-4280

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -3754,9 +3754,16 @@ _test_sample_versioned_api_example_5_6_7_8 (void)
    ASSERT_OR_PRINT (ok, error);
    db = mongoc_client_get_database (client, "db");
    sales = mongoc_database_get_collection (db, "sales");
-   /* Drop db.sales in case the collection exists. */
+   /* Ignore an "ns not found" error on dropping the db.sales collection in case
+   it exists. */
    ok = mongoc_collection_drop (sales, &error);
    if (!ok && NULL == strstr (error.message, "ns not found")) {
+
+      fprintf (stderr,
```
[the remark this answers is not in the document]
chardan: I found it useful when trying to figure out why I was or was not entering that block; as it's information that is already available there, I see no harm in leaving it (but I've made it test_print()); but if you would rather it be removed, I don't feel too strongly about it.

src/libmongoc/src/mongoc/mongoc-topology-scanner.c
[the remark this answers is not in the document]
chardan: I think you're right about $db vs. admin.$cmd. It can be confusing because they both wind up with the value "admin". Thank you for the clarification!

chardan:
When the environment variable MONGODB_API_VERSION is set, there are still some tests that can break, especially "/retryable_writes/*". This suite has a different implementation style than most of the others, I'm working on a fix.

I have filed a new ticket to address important clarification issues that I would prefer to defer from this changeset in order to do the further work from a working feature:
https://jira.mongodb.org/browse/CDRIVER-4280

chardan:
(Sorry for the force push, I did some rebasing!)

src/libmongoc/src/mongoc/mongoc-cluster-private.h
```diff
@@ -215,6 +215,23 @@ _mongoc_cluster_get_auth_cmd_x509 (const mongoc_uri_t *uri,
                                    bson_t *cmd /* OUT */,
                                    bson_error_t *error /* OUT */);
 
+bool
+mongoc_topology_uses_server_api (const mongoc_topology_t *topology);
```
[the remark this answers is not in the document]
chardan: I'm of mixed minds with this. I like your suggestion because it keeps the relevant types together. However, the implementations actually call each other, so they need to "see" each other's prototypes. Another idea is a new header/module. I will try it and see how it goes; if there's no dependency hell that seems reasonable.

src/libmongoc/tests/mock_server/mock-server.c
eramongodb: Ping: not yet addressed.

src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -885,6 +886,58 @@ mock_server_receives_request (mock_server_t *server)
...
+
+   if (!r) {
+      request_destroy (request);
+      return NULL;
+   }
+
+   return request;
+}
+
+
```
eramongodb, line 940:
```suggestion

```
Clang Format.

src/libmongoc/tests/mock_server/mock-server.c
eramongodb: Ping: not yet addressed.

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -1955,6 +2001,9 @@ _mongoc_cluster_node_new (mongoc_stream_t *stream,
    node->stream = stream;
    node->connection_address = bson_strdup (connection_address);
 
+   /* Note that the node->sd field is not explicitly initialized 
+   by parameter in this function. */
```
eramongodb, line 2005:
```suggestion
   /* Note that the node->sd field is not explicitly initialized by parameter in
    * this function. */
```
Clang Format.

src/libmongoc/src/mongoc/mongoc-cluster-private.h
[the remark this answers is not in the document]
eramongodb: Given declarations in header files and definitions in source files, there should be no issue with circular dependencies in header file inclusion.

src/libmongoc/tests/test-mongoc-versioned-api.c
```diff
@@ -189,4 +221,7 @@ test_client_versioned_api_install (TestSuite *suite)
    TestSuite_Add (suite, "/VersionedApi/copy", _test_mongoc_server_api_copy);
    TestSuite_Add (
       suite, "/VersionedApi/setters", _test_mongoc_server_api_setters);
+   TestSuite_Add (suite,
+                  "VersionedApi/private/client_uses_server_api",
+                  _test_mongoc_client_uses_server_api);
```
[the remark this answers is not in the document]
chardan: Good catch! Especially because the test harness is exceptionally picky about this and it's error handling essentially consists of segfaulting...

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -3754,9 +3754,14 @@ _test_sample_versioned_api_example_5_6_7_8 (void)
...
+   it exists. */
    ok = mongoc_collection_drop (sales, &error);
    if (!ok && NULL == strstr (error.message, "ns not found")) {
+
+      test_error("error.domain == %d, error.code == %d\n",
+               error.domain, error.code);
+
       /* Ignore an "ns not found" error on dropping the collection in case the
        * namespace does not exist. */
       ASSERT_OR_PRINT (ok, error);
```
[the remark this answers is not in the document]
chardan: Will do-- thanks, I wasn't familiar with ASSERT_WITH_MSG().

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -2347,7 +2396,6 @@ mongoc_cluster_stream_for_server (mongoc_cluster_t *cluster,
    server_stream = _mongoc_cluster_stream_for_server (
       cluster, server_id, reconnect_ok, cs, reply, error);
 
-
```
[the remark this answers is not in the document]
chardan: I may have lost the scent on this one during my rebase.

src/libmongoc/tests/test-mongoc-sdam-monitoring.c
```diff
@@ -840,10 +840,11 @@ _test_heartbeat_fails_dns (bool pooled)
    uri = mongoc_uri_new (
       "mongodb://doesntexist.foobar/?serverSelectionTimeoutMS=3000");
    if (pooled) {
-      pool = test_framework_client_pool_new_from_uri (uri, NULL);
+      pool = mongoc_client_pool_new (uri);
       pool_set_heartbeat_event_callbacks (pool, &context);
       client = mongoc_client_pool_pop (pool);
    } else {
+      /* This does not appear to fail with a legacy hello error: */
```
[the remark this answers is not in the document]
chardan: Comment removed. I don't quite understand what you're asking for for in the second clause.

src/libmongoc/tests/test-mongoc-versioned-api.c
chardan: +1

chardan (commented):
Made a pass going through comments-- there are a couple (which are a bit more involved) that I haven't gotten to yet. I'll make another sweep shortly.

src/libmongoc/tests/mock_server/mock-server.c
chardan: Interestingly, this causes test failure in "/Topology/hello/versioned_api/single"; going through it again, I am not seeing the bug yet.

src/libmongoc/tests/test-libmongoc.h
```diff
@@ -210,6 +210,8 @@ test_framework_set_debug_stream (mongoc_client_t *client,
 
 typedef int64_t server_version_t;
 
+server_version_t
+test_framework_get_server_version_with_client(mongoc_client_t *client);
```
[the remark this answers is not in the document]
chardan: test_framework_get_server_version() implicitly creates a client, which will pick up any environment variables, etc.; I wanted a version that would use a specific, already-constructed client. I believe I used this earlier on, I may have dropped the code that called it; in any case, it's probably useful to have a bit more granularity available.

src/libmongoc/tests/mock_server/mock-server.c
chardan: I've added this to:
https://jira.mongodb.org/browse/CDRIVER-4280

src/libmongoc/tests/test-mongoc-client.c
chardan: Okay, these have been revised. Thank you both for the suggestion.

src/libmongoc/tests/test-mongoc-background-monitoring.c
```diff
@@ -41,7 +41,8 @@ typedef struct {
 typedef enum {
    TF_FAST_HEARTBEAT = 1 << 0,
    TF_FAST_MIN_HEARTBEAT = 1 << 1,
-   TF_AUTO_RESPOND_POLLING_HELLO = 1 << 2
+   TF_AUTO_RESPOND_POLLING_HELLO = 1 << 2,
+   TF_NO_MONGODB_API_VERSION = 1 << 3, /* if set, do not pick up the MONGODB_API_VERSION environment variable */
```
kevinAlbs, line 45: I suggest removing `TF_NO_MONGODB_API_VERSION` and replacing the calls to `mock_server_receives_legacy_hello` with `mock_server_receives_any_hello`.

That will enable `tf_new` to use `test_framework_client_pool_new_from_uri`.

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -3754,9 +3754,16 @@ _test_sample_versioned_api_example_5_6_7_8 (void)
    ASSERT_OR_PRINT (ok, error);
    db = mongoc_client_get_database (client, "db");
    sales = mongoc_database_get_collection (db, "sales");
-   /* Drop db.sales in case the collection exists. */
```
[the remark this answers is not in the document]
kevinAlbs: This is still not addressed. You can revert these changes with:
```
git co master -- src/libmongoc/tests/test-mongoc-sample-commands.c
```

src/libmongoc/tests/test-mongoc-handshake.c
```diff
@@ -183,7 +183,7 @@ test_mongoc_handshake_data_append_success (void)
    mock_server_run (server);
    uri = mongoc_uri_copy (mock_server_get_uri (server));
    mongoc_uri_set_option_as_utf8 (uri, MONGOC_URI_APPNAME, "testapp");
-   pool = test_framework_client_pool_new_from_uri (uri, NULL);
+   pool = mongoc_client_pool_new (uri);
```
kevinAlbs, line 186: Please update this file to use `test_framework_client_pool_new_from_uri` and replace the calls to `mock_server_receives_legacy_hello` with `mock_server_receives_any_hello`.

src/libmongoc/tests/test-mongoc-sdam-monitoring.c
```diff
@@ -699,11 +699,11 @@ _test_heartbeat_events (bool pooled, bool succeeded)
    start = bson_get_monotonic_time ();
 
    if (pooled) {
-      pool = test_framework_client_pool_new_from_uri (uri, NULL);
+      pool = mongoc_client_pool_new (uri);
```
kevinAlbs, line 702: Please update this file to use test_framework_client_pool_new_from_uri and replace the calls to mock_server_receives_legacy_hello with mock_server_receives_any_hello.

src/libmongoc/tests/test-mongoc-speculative-auth.c
```diff
@@ -163,7 +163,7 @@ _test_mongoc_speculative_auth (bool pooled,
    }
 
    if (pooled) {
-      pool = test_framework_client_pool_new_from_uri (uri, NULL);
+      pool = mongoc_client_pool_new (uri);
```
kevinAlbs, line 166: Please update this file to use test_framework_client_pool_new_from_uri and replace the calls to mock_server_receives_legacy_hello with mock_server_receives_any_hello.

src/libmongoc/tests/test-mongoc-streamable-hello.c
```diff
@@ -52,7 +52,7 @@ test_topology_version_update (void)
    mock_server_run (server);
 
    client =
-      test_framework_client_new_from_uri (mock_server_get_uri (server), NULL);
+      mongoc_client_new_from_uri (mock_server_get_uri (server));
```
kevinAlbs, line 55: Please update this file to use test_framework_client_pool_new_from_uri and replace the calls to mock_server_receives_legacy_hello with mock_server_receives_any_hello.

src/libmongoc/tests/test-mongoc-topology-scanner.c
```diff
@@ -179,7 +179,7 @@ test_topology_scanner_discovery (void)
 
    uri_str = bson_strdup_printf ("mongodb://%s/?" MONGOC_URI_REPLICASET "=rs",
                                  mock_server_get_host_and_port (primary));
-   client = test_framework_client_new (uri_str, NULL);
+   client = mongoc_client_new (uri_str);
```
kevinAlbs, line 182: Please update this file to use test_framework_client_pool_new_from_uri and replace the calls to mock_server_receives_legacy_hello with mock_server_receives_any_hello.

src/libmongoc/tests/test-mongoc-versioned-api.c
kevinAlbs: I do not think there is one. I suggest defining a `test_framework_client_pool_new_no_server_api`.

src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -1160,6 +1307,32 @@ mock_server_receives_query (mock_server_t *server,
...
+ *
+ *       Pop a client non-streaming hello call if one is enqueued,
+ *       or wait up to request_timeout_ms for the client to send a request.
+ *
+ * Returns:
+ *       A request you must request_destroy, or NULL if the current
+ *       request is not a hello command.
+ *
+ * Side effects:
+ *       Logs if the current request is a hello command using OP_MSG.
```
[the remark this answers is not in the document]
chardan: Eventually it calls request_matches_msg(), which can call test_error(); this is documented as "Side effects: None", so I'll go with that and a note to see-also.

src/libmongoc/tests/test-mongoc-handshake.c
chardan: +1

src/libmongoc/tests/test-mongoc-topology.c
chardan: It does not need to move; I've put it back (but ironically it will still show up in the diff after clang-format; c'est la vie).

src/libmongoc/tests/test-libmongoc.h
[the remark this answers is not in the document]
kevinAlbs: Ping.

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -1954,6 +1982,9 @@ _mongoc_cluster_node_new (mongoc_stream_t *stream,
    node->stream = stream;
    node->connection_address = bson_strdup (connection_address);
 
+   /* Note that the node->sd field is set to NULL by bson_malloc(),
```
kevinAlbs, line 1985:
```suggestion
   /* Note that the node->sd field is set to NULL by bson_malloc0(),
```

src/libmongoc/src/mongoc/mongoc-topology-scanner.c
```diff
@@ -106,16 +106,15 @@ _jumpstart_other_acmds (mongoc_topology_scanner_node_t *node,
...
-
    BSON_APPEND_INT32 (&ts->hello_cmd, "hello", 1);
    BSON_APPEND_BOOL (&ts->hello_cmd, "helloOk", true);
 
    BSON_APPEND_INT32 (&ts->legacy_hello_cmd, HANDSHAKE_CMD_LEGACY_HELLO, 1);
    BSON_APPEND_BOOL (&ts->legacy_hello_cmd, "helloOk", true);
 
-   if (api) {
-      _mongoc_cmd_append_server_api (&ts->hello_cmd, api);
+   /* Append appropriate server API fields (such as "serverApi") if selected: */
```
kevinAlbs, line 115:
```suggestion
   /* Append appropriate server API fields if selected: */
```

"serverApi" is not a field.

eramongodb (approved):
Feedback regarding PR handling: due to rebasing, several comments are outdated and unresolved, yet unreachable (commit history lost). Suggestion: next time, close draft PR (keep branch) and, instead of rebasing, create a new branch and open a new (rebased) PR as "ready for review".

There are several calls to `mock_server_receives_any_hello_with_match` where the match pattern is identical for both legacy and non-legacy cases. Consider DRY and use a single variable for both arguments (makes visual diff-checking unnecessary).

My comments in this review cycle are relatively minor and non-blocking, so marking as LGTM pending further updates.

src/libmongoc/tests/mock_server/mock-server.c
```diff
@@ -1035,29 +1056,90 @@ mock_server_receives_bulk_msg (mock_server_t *server,
...
+ const char *hello_str = "{'hello': 1, 'maxAwaitTimeMS': { '$exists': false }}";
+ 
+ bool success = false;
+
+ if(!request) {
+     return NULL;
+ }
+
+ /* We check the opcode separately because request_matches_msg() and friends like
+ to abort the program when checks fail: */
```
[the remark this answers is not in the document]
kevinAlbs: OK. No need to test.

src/libmongoc/src/mongoc/mongoc-rpc.c
```diff
@@ -993,6 +997,29 @@ _mongoc_rpc_get_first_document (mongoc_rpc_t *rpc, bson_t *reply)
...
 
+/* Get the first BSON document from an OP_MSG reply: */
+bool
+_mongoc_rpc_reply_get_first_msg (mongoc_rpc_msg_t *reply_msg,
+                                 bson_t *bson_reply)
+{
+   /* Note that mongo_rpc_reply_t is a union, with a mongo_rpc_msg_t field; see
+   the *.def files and MongoDB Wire Protocol documentation for details. */
+
+   int32_t document_len;
```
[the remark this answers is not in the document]
chardan: Good catch! Although now there's a question of which is right-- all the .def files use int32, but below we see:

   document_len = BSON_UINT32_FROM_LE (document_len);

...while I think uint32 is /logical/, in _mongoc_rpc_reply_get_first() there's the same behavior! Is this a bug?

@kevinAlbs Any thoughts here?

src/libmongoc/src/mongoc/mongoc-topology-scanner.c
```diff
@@ -252,7 +251,8 @@ _mongoc_topology_scanner_parse_speculative_authentication (
 }
 
 static bson_t *
-_build_handshake_cmd (const bson_t *basis_cmd,
+_build_handshake_cmd (const mongoc_topology_scanner_t *ts,
+                      const bson_t *basis_cmd,
```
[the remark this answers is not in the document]
chardan: +1, thank you

src/libmongoc/tests/test-mongoc-versioned-api.c
```diff
@@ -189,4 +227,7 @@ test_client_versioned_api_install (TestSuite *suite)
    TestSuite_Add (suite, "/VersionedApi/copy", _test_mongoc_server_api_copy);
    TestSuite_Add (
       suite, "/VersionedApi/setters", _test_mongoc_server_api_setters);
+   TestSuite_Add (suite,
+                  "VersionedApi/private/client_uses_server_api",
```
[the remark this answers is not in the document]
chardan: +1

src/libmongoc/src/mongoc/mongoc-rpc.c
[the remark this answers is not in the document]
kevinAlbs: On [bsonspec.org](https://bsonspec.org/spec.html) the four byte document length is specified as a signed int32. I think that is why the .def files use int32. AFAIK the expectation is that the length is non-negative. Otherwise it is invalid BSON.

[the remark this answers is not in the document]
chardan: It probably is better to fix them in a batch. Agreed that signed to unsigned is defined (although can still be "fun"!). I will note this for CDRIVER-4280. Thanks everyone for the comments!

src/libmongoc/tests/test-mongoc-cluster.c
```diff
@@ -1357,7 +1365,7 @@ _test_cluster_hello_fails (bool hangup)
    uri = mongoc_uri_copy (mock_server_get_uri (mock_server));
    /* increase heartbeatFrequencyMS to prevent background server selection. */
    mongoc_uri_set_option_as_int32 (uri, "heartbeatFrequencyMS", 99999);
-   pool = test_framework_client_pool_new_from_uri (uri, NULL);
+   pool = mongoc_client_pool_new (uri);
```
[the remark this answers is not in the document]
chardan: +1 indeed, it was a test I missed while reverting and converting to *any_hello(). Thanks!

src/libmongoc/tests/test-mongoc-sdam-monitoring.c
```diff
@@ -941,7 +941,7 @@ test_no_duplicates (void)
     * from interfering. */
    uri = mongoc_uri_copy (mock_server_get_uri (server));
    mongoc_uri_set_option_as_int32 (uri, MONGOC_URI_HEARTBEATFREQUENCYMS, 99999);
-   pool = test_framework_client_pool_new_from_uri (uri, NULL);
+   pool = mongoc_client_pool_new (uri);
```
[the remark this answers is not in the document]
chardan: Yes, indeed!

src/libmongoc/tests/test-mongoc-topology.c
```diff
@@ -2401,7 +2404,7 @@ _test_hello_ok (bool pooled)
    mongoc_uri_set_option_as_int32 (uri, MONGOC_URI_HEARTBEATFREQUENCYMS, 500);
 
    if (pooled) {
-      pool = test_framework_client_pool_new_from_uri (uri, NULL);
+      pool = mongoc_client_pool_new (uri);
```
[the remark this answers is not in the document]
chardan: +1!

src/libmongoc/tests/test-mongoc-sample-commands.c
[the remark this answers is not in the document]
chardan: This looks a bit wierd in my local history because I had done a merge-- I believe it should be sorted out now.

kevinAlbs (approved):
LGTM

PR #931, opened 2022-01-20
CDRIVER-4231: New random generation of OID values

vector-of-bool:
bson-context initializes five bytes with randomness. Previously, this was done by XOR'ing certain machine attribute into an integral seed value, and then calling `rand_r(seed)` to get a random value. Using plain XOR on those attributes would very often result in a seed collision, and `rand_r()` would only return four bytes of randomness, with the fifth byte being generated by another call to `rand_r()`, and the two sequence counters initialized with further calls to `rand_r()`. This resulted in _at most_ 32 bits of randomness being used for nine bytes of random output, and the seed was very likely to collide in the first place.

This changeset instead generates a struct containing those same machine attributes, and then performs a SipHash of the bit representation of that struct, resulting in 16 bytes of psuedorandomness. These bytes are then stored as the random data and used to initialize the sequence counters.

The changeset has tweaked a few aspects of the context, which may need be un-tweaked.


src/libbson/src/bson/bson-context.c
```diff
@@ -43,49 +43,24 @@
  * Globals.
  */
 static bson_context_t gContextDefault;
+static int64_t gRandCounter = INT64_MIN;
```
eramongodb, line 46: Can this variable be moved into `_bson_context_init_random()` as a static local variable?

I also recommend adding a comment to documentation its purpose.

src/libbson/src/bson/bson-context.c
```diff
@@ -264,79 +142,236 @@ _bson_context_get_hostname (char *out)
...
+
+   /* Hash the param struct */
+   _siphash (&rand_params, sizeof rand_params, key, digest);
+
+   /** Initialize the rand and sequence counters with our random digest */
+   memcpy (context->randomness, digest, sizeof context->randomness);
+   if (init_seq) {
+      memcpy (&context->seq32, digest + 1, sizeof context->seq32);
+      memcpy (&context->seq64, digest + 1, sizeof context->seq64);
+      /* Chop off some initial bits for nicer counter behavior */
```
eramongodb, line 319: What is "nicer counter behavior"? Is it "so that the last digit of the OID will start at zero" (per older comments)? If so, suggest this description is preserved.

src/libbson/src/bson/bson-context.c
```diff
@@ -264,79 +142,236 @@ _bson_context_get_hostname (char *out)
...
+      /* User has requested that we check if our PID has changed. This can occur
+       * after a call to fork() */
+      uint64_t now_pid = _bson_getpid ();
+      if (now_pid != context->pid) {
+         _bson_context_init_random (
+            context, false /* Do not update the sequence counters */);
+      }
+   }
+   /* Copy the stored randomness into the OID */
+   memcpy (&oid->bytes[4], &context->randomness, 5);
```
eramongodb, line 372:
```suggestion
   memcpy (&oid->bytes[4], &context->randomness, BSON_CONTEXT_RANDOMNESS_SIZE);
```
Magic number; replace with named constant?

src/libbson/src/bson/bson-context-private.h
```diff
@@ -31,10 +31,10 @@ BSON_BEGIN_DECLS
 struct _bson_context_t {
    /* flags are defined in bson_context_flags_t */
    int flags;
-   int32_t seq32;
-   int64_t seq64;
-   uint8_t rand[5];
-   uint16_t pid;
+   uint32_t seq32;
+   uint64_t seq64;
+   uint8_t randomness[5];
```
eramongodb, line 36:
```suggestion
   uint8_t randomness[BSON_CONTEXT_RANDOMNESS_SIZE];
```
Magic number; replace with named constant?

src/libbson/src/bson/bson-context.c
```diff
@@ -264,79 +142,236 @@ _bson_context_get_hostname (char *out)
...
+}
 
-#ifndef BSON_HAVE_RAND_R
-   srand (seed);
-#endif
+void
+_siphash (const void *in,
+          const size_t inlen,
+          const uint64_t key[2],
+          uint64_t digest[2])
```
eramongodb, line 186: Was a reference used to implement this function (e.g. https://eprint.iacr.org/2012/351.pdf)? If so, please add a source as comment/documentation.

src/libbson/src/bson/bson-context.c
```diff
@@ -264,79 +142,236 @@ _bson_context_get_hostname (char *out)
...
+   /* Init each part of the randomness source: */
+   memset (&rand_params, 0, sizeof rand_params);
+   bson_gettimeofday (&rand_params.time);
+   rand_params.pid = _bson_getpid ();
+   context->gethostname (rand_params.hostname);
+   rand_params.rand_call_counter =
+      bson_atomic_int64_fetch_add (&gRandCounter, 1, bson_memory_order_seq_cst);
+
+   memcpy (key, &rand_params, sizeof key);
+   key[1] = ~key[0];
```
eramongodb, line 309: What is the rationale behind this method to determine the key value `k` to use with the SipHash function?

eramongodb (commented):
> The changeset has tweaked a few aspects of the context, which may need be un-tweaked.

Given it is declared in a private header, I think modifications should be fine.

> then performs a SipHash of the bit representation of that struct

Please document the rationale behind the choice of SipHash for this purpose.

src/libbson/src/bson/bson-oid.h
```diff
@@ -214,8 +214,8 @@ bson_oid_init_from_string_unsafe (bson_oid_t *oid, const char *str)
    int i;
 
    for (i = 0; i < 12; i++) {
-      oid->bytes[i] = ((bson_oid_parse_hex_char (str[2 * i]) << 4) |
-                       (bson_oid_parse_hex_char (str[2 * i + 1])));
+      oid->bytes[i] = ((uint8_t) (bson_oid_parse_hex_char (str[2 * i]) << 4) |
+                       (uint8_t) (bson_oid_parse_hex_char (str[2 * i + 1])));
```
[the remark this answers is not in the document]
vector-of-bool: This is correct. It seems that my compiler was happy with either change to silence the warnings. These are warnings with `-Wconversion`, which I was using the validate there were no accidental narrowing casts while adding SipHash.

src/libbson/src/bson/bson-context.c
vector-of-bool: Yes

src/libbson/src/bson/bson-context.c
```diff
@@ -264,79 +139,261 @@ _bson_context_get_hostname (char *out)
...
 
 static void
 _bson_context_init (bson_context_t *context, bson_context_flags_t flags)
 {
    context->flags = (int) flags;
-   context->oid_set_seq32 = _bson_context_set_oid_seq32;
-   context->oid_set_seq64 = _bson_context_set_oid_seq64;
    context->gethostname = _bson_context_get_hostname;
 
-   if ((flags & BSON_CONTEXT_THREAD_SAFE)) {
```
kevinAlbs, line 330: I do not know why they were added originally. But are the thread unsafe variants of `_bson_context_set_oid_seq32` and `_bson_context_set_oid_seq64` removed because they provide no benefit?

src/libbson/src/bson/bson-context.c
```diff
@@ -264,79 +139,261 @@ _bson_context_get_hostname (char *out)
...
-_bson_context_init_random (bson_context_t *context, bool init_sequence)
+/*** ========================================
+ * The below SipHash implementation is based on the original public-domain
+ * reference implementation from Jean-Philippe Aumasson and DJB
+ * (https://github.com/veorq/SipHash).
+ */
+
+/* in-place rotate a 64bit number */
+void
+_rotl64 (uint64_t *p, int nbits)
```
kevinAlbs, line 150:
```suggestion
_bson_rotl64 (uint64_t *p, int nbits)
```

The linked [patch build](https://spruce.mongodb.com/task/mongo_c_driver_windows_2017_debug_compile_sasl_openssl_cse_patch_7a35cf865629122267c3d59cca0f922c5e455f0d_61e9fd0ad1fe074fa7152a7c_22_01_21_00_23_40/logs?execution=0) errors from redefinition:

```
c:\data\mci\a7a6ccecd92f7a5293b119f04a16cb5c\mongoc\src\libbson\src\bson\bson-context.c(151): error C2169: '_rotl64': intrinsic function, cannot be defined [C:\data\mci\a7a6ccecd92f7a5293b119f04a16cb5c\mongoc\src\libbson\bson_static.vcxproj]
```

kevinAlbs (changes requested):
Confirmed with [this repro](https://github.com/kevinAlbs/c-bootstrap/tree/master/investigations/help30695) that duplicates ObjectIDs are no longer produced an Ubuntu 18.04 spawn host.

Changes look good pending remaining questions, a fix to replace `BSON_CONTEXT_DISABLE_HOST_CACHE` with `BSON_CONTEXT_DISABLE_PID_CACHE`, and passing tests.

Since we are close to a release, I suggest configuring a full patch build to test `debug-compile-*` tasks on each platform once this is ready to merge.

src/libbson/src/bson/bson-context.c
```diff
@@ -264,79 +142,236 @@ _bson_context_get_hostname (char *out)
...
+ *
+ *--------------------------------------------------------------------------
+ */
+void
+_bson_context_set_oid_rand (bson_context_t *context, bson_oid_t *oid)
+{
+   BSON_ASSERT (context);
+   BSON_ASSERT (oid);
+
+   if (context->flags & BSON_CONTEXT_DISABLE_HOST_CACHE) {
```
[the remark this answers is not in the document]
vector-of-bool: Oh, that's an auto-complete typo. Good catch.

src/libbson/src/bson/bson-context.c
vector-of-bool: I believe so. The first introductory changeset they were still only differing in atomic operations versus regular operations. IMO, the cost of an atomic increment over a regular increment is so vanishingly small in the context of OID generation that it isn't worth having a non-threadsafe option. This is the main "context tweak" I mentioned. If we really want the non-threadsafe option to remain, I'm open to restoring it, but I think it's not worth the additional complexity. Removing the option will require docs updates and allows for further simplification by removing some function pointers in the context.

Thoughts?

src/libbson/src/bson/bson-context.c
vector-of-bool: Oops. That's a fun coincidence.

src/libbson/src/bson/bson-context.c
vector-of-bool: Replaced with named constants

src/libbson/src/bson/bson-context.c
vector-of-bool: Updated with a reference

src/libbson/src/bson/bson-context.c
vector-of-bool: It is entirely arbitrary. Added a comment.

src/libbson/src/bson/bson-context.c
kevinAlbs: I do not know the motivation for the thread unsafe variants. If the atomic increment likely will not impact the default behavior, removing SGTM.

src/libbson/src/bson/bson-context.c
```diff
@@ -249,7 +92,7 @@ _get_rand (unsigned int *pseed)
  * --------------------------------------------------------------------------
  */
 static void
-_bson_context_get_hostname (char *out)
+_bson_context_get_hostname (char out[HOST_NAME_MAX])
```
eramongodb, line 95:
```suggestion
_bson_context_get_hostname (char* out)
```
Suggestion: avoid misleading array type parameters. Alternatively, deliberately opt into C99 "static array indices" (note: may not be supported by older compilers, e.g. VS 2013 and/or GCC 4.8):
```suggestion
_bson_context_get_hostname (char out[static HOST_NAME_MAX])
```

src/libbson/src/bson/bson-context.c
```diff
@@ -264,111 +107,242 @@ _bson_context_get_hostname (char *out)
 }
 
 
-static void
-_bson_context_init_random (bson_context_t *context, bool init_sequence)
+/*** ========================================
+ * The below SipHash implementation is based on the original public-domain
```
eramongodb, line 111: Comment seems misplaced; should be moved to immediately before the `_siphash()` function definition.

eramongodb (changes requested):

src/libbson/src/bson/bson-context.c
vector-of-bool: The only difference with `static` is the introduction of additional UB consideration (and a helpful compiler warning). I switched from a plain pointer to an array to represent that the pointee must be an array of the given size. I'd add the `static`, but we aren't on C99 yet. Given that this is an internal-only API I think it is safe to use the array-decl as hints to the expected array-ness (and required bounds).

src/libbson/src/bson/bson-context.c
vector-of-bool: Everything between this comment and the end of `_siphash` are all based on the original impl. I'd like to reverse their decl order, but we're in C, so they must be declared bottom-up.

kevinAlbs (approved):
Thank you for the additional documentation and cleanup. LGTM!

src/libbson/src/bson/bson-context.c
eramongodb: I am generally opposed to syntax that differs from its actual semantics (for reasons [colorfully described by Linus Torvalds](https://lkml.org/lkml/2015/9/3/428)). However, given it is not user-facing interface, its scope is limited enough that I do not consider its removal a blocker (I also see there are a few instances of it scattered throughout the C driver already, so there is precedent).

eramongodb (approved):

PR #932, opened 2022-01-21
CDRIVER-4264 Address build failures on VS 2013 variants on Evergreen

eramongodb:
Per CDRIVER-4264, there are three compilation errors preventing successful compilation [on VS 2013 variants](https://evergreen.mongodb.com/version/mongo_c_driver_7a35cf865629122267c3d59cca0f922c5e455f0d). This PR addresses each of these warnings:

- `'_Exit' undefined; assuming extern returning int`

VS 2013 only provides partial support for the C99 standard. This support apparently does not extend to providing the `_Exit` function, which is a C99 feature. Given that its use in the test suite (to report the type of signal triggering failure) is [not strictly necessary](https://github.com/mongodb/mongo-c-driver/pull/870#discussion_r712657428), I have [opted for its removal](https://github.com/mongodb/mongo-c-driver/commit/5b0dbb33a76eb0c1d837de790fc313c963b5bf53).

- `'snprintf' undefined; assuming extern returning int`

`snprintf` is also a C99 feature. Some files in libmongoc were directly invoking `snprintf` despite libbson providing `bson_snprintf` to address standard / platform compatibility issues. These instances [were substituted with](https://github.com/mongodb/mongo-c-driver/commit/9f9f48b761c4d4f3bebae9ea41f8a5a19727dbf6) `bson_snprintf`.

- `'_InterlockedExchangePointer': identifier not found`

This issue only manifested when building with VS 2013 targeting x86 (i686). I do not yet fully understand the reason behind this issue. According to [a Boost patch](https://github.com/boostorg/winapi/commit/bb91190323166485a338b416446020a47be8beff) (note, "MSVC-12" == "VS 2013") in response to [a report of similar build failures](https://lists.boost.org/Archives/boost/2019/07/246563.php), the issue is apparently due to a conflict between the `<intrin.h>` and `<winnt.h>` libraries when targeting x86 with VS 2013. This is in spite of [Microsoft documentation](https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2013/853x471w(v=vs.120)) indicating support for the intrinsic when targeting x86. Consequently, I have opted to [emulating the intrinsic](https://github.com/mongodb/mongo-c-driver/commit/77bbfee6a927af39a8040f18a6a2c37624b5791c) given VS 2013 and x86.

### DIFF
--- a/src/libbson/src/bson/bson-atomic.c
+++ b/src/libbson/src/bson/bson-atomic.c
@@ -48,7 +48,7 @@ bson_memory_barrier (void)
 }
 
 /**
- * Some platforms do not support compiler intrinsics for atomic operations.
+ * Some platforms do not support compiler ics for atomic operations.
  * We emulate that here using a spin lock and regular arithmetic operations
  */
 static int8_t gEmulAtomicLock = 0;
@@ -244,4 +244,17 @@ _bson_emul_atomic_int_compare_exchange_weak (volatile int *p,
    /* We're emulating. We can't do a weak version. */
    return _bson_emul_atomic_int_compare_exchange_strong (
       p, expect_value, new_value, order);
+}
+
+void *
+_bson_emul_atomic_ptr_exchange (void *volatile *p,
+                                void *n,
+                                enum bson_memory_order _unused)
+{
+   void *ret;
+   _lock_emul_atomic ();
+   ret = *p;
+   *p = n;
+   _unlock_emul_atomic ();
+   return ret;
 }
--- a/src/libbson/src/bson/bson-atomic.c
+++ b/src/libbson/src/bson/bson-atomic.c
@@ -195,8 +195,8 @@ _bson_emul_atomic_int32_compare_exchange_weak (volatile int32_t *p,
 
 int
 _bson_emul_atomic_int_fetch_add (volatile int *p,
-                                   int n,
-                                   enum bson_memory_order _unused)
+                                 int n,
+                                 enum bson_memory_order _unused)
 {
    int ret;
    _lock_emul_atomic ();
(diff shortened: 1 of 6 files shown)


vector-of-bool (approved):
LGTM. Do we not have VS 2013 set as a default build task?

eramongodb:
It appears not. A patch build verifying these changes (with minor discrepancies in the diff compared to this PR) are [available here](https://spruce.mongodb.com/version/61eaf9050305b960387dcdfa/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) for reference.

src/libbson/src/bson/bson-atomic.c
```diff
@@ -48,7 +48,7 @@ bson_memory_barrier (void)
 }
 
 /**
- * Some platforms do not support compiler intrinsics for atomic operations.
+ * Some platforms do not support compiler ics for atomic operations.
```
kevinAlbs, line 51: Is this a typo?

src/libbson/src/bson/bson-atomic.h
```diff
@@ -73,6 +73,12 @@ enum bson_memory_order {
 #define BSON_EMULATE_INT
 #endif
 
+/* CDRIVER-4264 Contrary to documentation, VS 2013 targeting x86 does not
+ * correctly/consistently provide _IntrinsicPointerExchange. */
```
kevinAlbs, line 77:
```suggestion
 * correctly/consistently provide _InterlockedExchangePointer. */
```

kevinAlbs (approved):
LGTM

src/libbson/src/bson/bson-atomic.c
eramongodb: Yes. Thank you for the catch.

src/libbson/src/bson/bson-atomic.h
eramongodb: Thank you. I keep mistakening "Interlocked" as "Intrinsic". Will fix. 😑

PR #935, opened 2022-01-31
Remove use COMMON_PREFIX for decorating common lib names

vector-of-bool:
The use of `COMMON_PREFIX` is to ensure that the decorated names are different when compiled into libbson versus libmongoc, but make things quite cluttered and strangely formatted at the call site. While I don't like heavy use on the preprocessor, `#define`-ing the names to a more readable alias makes it much easier to use and read those decorated names. I chose an arbitrary `mcommon` prefix for the names, but any sufficiently unique name will do.

(The guard on `__INTELLISENSE__` is to hide the decorated names from autocomplete tools.)

(AFAIK, this has no effect on any public API, nor visible symbols in any dynamic libraries.)

### DIFF
--- a/build/future_function_templates/future.c.template
+++ b/build/future_function_templates/future.c.template
@@ -56,9 +56,9 @@ void
 future_start (future_t *future,
               BSON_THREAD_FUN_TYPE (start_routine))
 {
-   int r = COMMON_PREFIX (thread_create) (&future->thread,
-                               start_routine,
-                               (void *) future);
+   int r = mcommon_thread_create (&future->thread,
+                                  start_routine,
+                                  (void *) future);
 
    BSON_ASSERT (!r);
 }
@@ -97,7 +97,7 @@ future_wait_max (future_t *future, int64_t timeout_ms)
       future->awaited = true;
 
       /* free memory */
-      COMMON_PREFIX (thread_join) (future->thread);
+      mcommon_thread_join (future->thread);
    }
 
    return resolved;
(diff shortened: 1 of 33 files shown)


eramongodb (commented):
I am generally opposed to this PR.

Use of the `COMMON_PREFIX` macro may uglify call sites, but it also serves as _the_ (singular) maintenance point through which the common prefix may be applied and/or modified for all associated entities. The proposed change significantly increases maintenance burden by requiring all such entities to be modified individually, rather than via the `COMMON_PREFIX` utility. Therefore, I do not find these changes sufficiently motivated.

eramongodb (commented):
Disregard prior comment, I did not fully comprehend the purpose of the `mcommon_` macros, which maintain the use of the `COMMON_NAME` utility, albeit hiding it behind a more palatable series of macro names by which entities may be referred to.

I'm not sure how I feel about `mcommon` as the corresponding entity name prefixes. I think I would have proposed `mongoc_common` (C driver common utilities), though it may be more verbose.

kevinAlbs (approved):
Thank you for the proposed QoL improvement and for doing the refactor as a separate PR.
I do not have a strong opinion. I think it is an improvement to readability of the `COMMON_PREFIX` calls.

> (AFAIK, this has no effect on any public API, nor visible symbols in any dynamic libraries.)

Agreed.

> I'm not sure how I feel about mcommon as the corresponding entity name prefixes. I think I would have proposed mongoc_common (C driver common utilities), though it may be more verbose.

Either `mcommon` or `mongoc_common` SGTM.

eramongodb (approved):

PR #943, opened 2022-02-08
CDRIVER-4284 Avoid empty OpenSSL error condition on handshake failure

eramongodb:
CDRIVER-4284 reports a scenario where `ERR_get_error` is used to construct failure information despite the possibility that the OpenSSL library may not contain error info to be reported. This PR ensures that there is a meaningful error message is guaranteed to be relayed even if OpenSSL does not provide error info.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-stream-tls-openssl.c
+++ b/src/libmongoc/src/mongoc/mongoc-stream-tls-openssl.c
@@ -544,7 +544,7 @@ _mongoc_stream_tls_openssl_check_closed (mongoc_stream_t *stream) /* IN */
 
 
 static bool
-_mongoc_stream_tls_openssl_cert_verify_failed (SSL *ssl, bson_error_t *error)
+_mongoc_stream_tls_openssl_set_verify_cert_error (SSL *ssl, bson_error_t *error)
 {
    BSON_ASSERT_PARAM (ssl);
    BSON_ASSERT_PARAM (error);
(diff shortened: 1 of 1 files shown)


src/libmongoc/src/mongoc/mongoc-stream-tls-openssl.c
```diff
@@ -631,14 +634,32 @@ _mongoc_stream_tls_openssl_handshake (mongoc_stream_t *stream,
...
-                      "TLS handshake failed: %s",
-                      ERR_error_string (ERR_get_error (), NULL));
+   /* Try to relay certificate failure reason from OpenSSL library if any. */
+   if (_mongoc_stream_tls_openssl_cert_verify_failed (ssl, error)) {
+      RETURN (false);
    }
 
+   /* Otherwise, try to relay error info from OpenSSL. */
+   {
+      const unsigned long code = ERR_get_error ();
```
jmikola, line 644: Noted that [`ERR_get_error`](https://www.openssl.org/docs/man3.0/man3/ERR_get_error.html) pops an error code off of the queue, so you must capture its return value in lieu of calling it twice.

jmikola (approved):
Changes LGTM.

One suggestion to either rename or add a doc blurb above `_mongoc_stream_tls_openssl_cert_verify_failed`, but that's not a deal breaker.

jmikola (approved):

kevinAlbs (approved):
LGTM

PR #944, opened 2022-02-09
CDRIVER-4274 Increase wait time for mock KMS server startup to 5 minutes

eramongodb:
Flaky mock KMS server startup failures as reported in CDRIVER-4274 seem to be addressed by raising the wait duration up from 1 minute to 5 minutes per [patch results](https://spruce.mongodb.com/version/6201961761837d0dfe173e88/) (none of the failing tasks are due to mock KMS server startup failure). It seems like the background processes launching the mock KMS servers were not being given enough time (1 minute was too short).


kevinAlbs (approved):
LGTM, thanks for the quick fix!

PR #954, opened 2022-03-13
CDRIVER-3923 switch to new RHEL8 s390x Evergreen hosts

rcsanchez97:
Full patch build: https://spruce.mongodb.com/version/622e50b457e85a4b73b784fd/tasks

A few items to note:
- the new variant is named `zseries-rhel83`, but retains the same display name as before, `*zSeries`
- the `authentication-tests-openssl-nosasl` and `authentication-tests-openssl` tasks in the new variant fail, but they were also failing on the waterfall before this change
- all of the `test-*` tasks in the new variant are failing in the `SYSTEM FAILED` state; this is a result of the server not yet publishing builds for RHEL8 zSeries (the ticket for that is [SERVER-44074](https://jira.mongodb.org/browse/SERVER-44074)) and the tasks should turn green once the server builds are published
- based on the Slack discussion I have added a skip for the `/server_discovery_and_monitoring/monitoring/heartbeat/pooled/dns`, verifying that it is skipped on RHEL8 zSeries and not skipped on other platforms

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -22022,11 +22022,11 @@ buildvariants:
   - .debug-compile !.sspi .nossl
   - .4.0 .openssl !.nosasl .server
   batchtime: 1440
-- name: zseries-rhel72
+- name: zseries-rhel83
   display_name: '*zSeries'
   expansions:
     CC: gcc
-  run_on: rhel72-zseries-test
+  run_on: rhel83-zseries-small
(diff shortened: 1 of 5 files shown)


kevinAlbs (approved):
LGTM! Apologies for the late review. I was not aware of the `_mongoc_linux_distro_scanner_get_distro` function.

PR #959, opened 2022-03-24
eliminate dead store operations that caused scan-build failures

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/623cd28b61837d50964ea4e1/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&taskName=scan-build (it is a complete patch build, but the link has it filtered down to only scan-build tasks)

@vector-of-bool I requested your review as well since this required tweaking code that was committed by you in the last few months.

### DIFF
--- a/src/libmongoc/tests/test-mongoc-topology.c
+++ b/src/libmongoc/tests/test-mongoc-topology.c
@@ -1747,7 +1747,6 @@ test_compatible_null_error_pointer (void)
    mock_server_run (server);
    client =
       test_framework_client_new_from_uri (mock_server_get_uri (server), NULL);
-   mc_tpld_unsafe_get_const (client->topology);
 
    /* trigger connection, fails due to incompatibility */
    ASSERT (!mongoc_client_command_simple (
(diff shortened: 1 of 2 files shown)


src/libmongoc/tests/test-mongoc-topology.c
```diff
@@ -1747,7 +1747,7 @@ test_compatible_null_error_pointer (void)
    mock_server_run (server);
    client =
       test_framework_client_new_from_uri (mock_server_get_uri (server), NULL);
-   td = mc_tpld_unsafe_get_const (client->topology);
+   mc_tpld_unsafe_get_const (client->topology);
```
vector-of-bool, line 1750: This line can be deleted outright.

vector-of-bool (approved):
LGTM, but one note.

kevinAlbs (approved):

src/libmongoc/tests/test-mongoc-topology.c
rcsanchez97: @vector-of-bool Thanks for pointing that out.  I assumed that the call to
`mc_tpld_unsafe_get_const` produced a side-effect that might not be okay
to eliminate.  I just looked at the implementation and I can see that it
does not have any side-effects.

rcsanchez97:
Thanks for pointing that out.  I assumed that the call to
`mc_tpld_unsafe_get_const` produced a side-effect that might not be okay
to eliminate.  I just looked at the implementation and I can see that it
does not have any side-effects.

PR #960, opened 2022-03-24
Reducing Warnings - Safe Integral Comparisons

eramongodb:
This is part of a series of PRs intended to reduce the set of warnings currently being observed when building the C driver.

This PR introduces [a set of utilities](https://github.com/mongodb/mongo-c-driver/commit/1c007a3bd54328a59b5ec08df58acdcb2cc5d762) modeled after the [Safe Integral Comparisons](https://wg21.link/p0586r2) proposal for the C++ standard library, which was accepted and merged into the C++20 standard. These utilities are extremely useful in addressing warnings involving comparisons or conversions between integers with different signedness and sizes.

The interface proposed in the original paper relies on type deduction to select appropriate behavior according to the signedness and size of the given arguments:

```cpp
A a = ...
B b = ...
// no need for any cast since std::cmp_less is taking care of everything
if(std::cmp_less(a,b)){
  // do X
} else {
  // do Y
}
```

As C does not have type deduction required to support this interface, two major compromises are made instead:
1. All functions must explicitly specify the signedness of the provided argument(s).
2. All arguments are promoted to either `int64_t` or `uint64_t` as specified by (1) before further analysis.

Therefore, the equivalent code to the above example given `A = int16_t` and `B = uint32_t` looks like the following:
```c
int16_t a = ...
uint32_t b = ...

if (bson_cmp_less_su (a, b)) {
  /* ... */
} else {
  /* ... */
}
```

where the `_su` suffix indicates `a` is signed and `b` is unsigned.

The set of supported types is also limited to the following:
* signed and unsigned char, short, int, long, and long long (but not plain char).
* signed and unsigned fixed width integers of size 8, 16, 32, and 64.
* `ssize_t` and `size_t`.

More types may be added if needed, so long as for the given type `T`, `sizeof (T) <= sizeof (uint64_t)` is true and its representable range is known and defined with corresponding `*_MIN` and `*_MAX` macros.

To support the `std::in_range` set of functions even on C90 (where the necessary `<limits.h>` and `<stdint.h>` macros may not be defined), the corresponding set of numeric limit macros have been [added to bson-compat.h](https://github.com/mongodb/mongo-c-driver/commit/7d32b8e500abbf7ea3959f14d10d6ed7496269f4), including those for `ssize_t` (note: `SSIZE_MIN` is normally not defined even by the POSIX standard). The preprocessor block [conditioned on](https://github.com/mongodb/mongo-c-driver/commit/7d32b8e500abbf7ea3959f14d10d6ed7496269f4#diff-1af8fdef56993ceee568101fbf55209a7290a59e5ed92a5b9150f1af5c98a7bdR148) `!defined(__STDC_VERSION__) || __STDC_VERSION__ < 199901L` may be removed once the minimum C standard for the C driver is bumped to C99 or newer.

### DIFF
--- a/src/libbson/tests/test-bson-cmp.c
+++ b/src/libbson/tests/test-bson-cmp.c
@@ -97,7 +97,7 @@ test_bson_cmp_less (void)
    BSON_ASSERT (!bson_cmp_less_ss (1, -1));
    BSON_ASSERT (!bson_cmp_less_ss (1, 1));
 
-   BSON_ASSERT (!bson_cmp_less_uu (0u, 0));
+   BSON_ASSERT (!bson_cmp_less_uu (0u, 0u));
    BSON_ASSERT (bson_cmp_less_uu (0u, 1u));
    BSON_ASSERT (!bson_cmp_less_uu (1u, 0u));
    BSON_ASSERT (!bson_cmp_less_uu (1u, 1u));
@@ -130,7 +130,7 @@ test_bson_cmp_greater (void)
    BSON_ASSERT (bson_cmp_greater_ss (1, -1));
    BSON_ASSERT (!bson_cmp_greater_ss (1, 1));
 
-   BSON_ASSERT (!bson_cmp_greater_uu (0u, 0));
+   BSON_ASSERT (!bson_cmp_greater_uu (0u, 0u));
    BSON_ASSERT (!bson_cmp_greater_uu (0u, 1u));
    BSON_ASSERT (bson_cmp_greater_uu (1u, 0u));
    BSON_ASSERT (!bson_cmp_greater_uu (1u, 1u));
@@ -163,7 +163,7 @@ test_bson_cmp_less_equal (void)
    BSON_ASSERT (!bson_cmp_less_equal_ss (1, -1));
    BSON_ASSERT (bson_cmp_less_equal_ss (1, 1));
 
-   BSON_ASSERT (bson_cmp_less_equal_uu (0u, 0));
+   BSON_ASSERT (bson_cmp_less_equal_uu (0u, 0u));
    BSON_ASSERT (bson_cmp_less_equal_uu (0u, 1u));
    BSON_ASSERT (!bson_cmp_less_equal_uu (1u, 0u));
    BSON_ASSERT (bson_cmp_less_equal_uu (1u, 1u));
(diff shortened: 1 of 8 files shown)


src/libbson/tests/test-bson-cmp.c
```diff
@@ -0,0 +1,309 @@
...
+
+   BSON_ASSERT (!bson_in_range_signed (int32_t, int32_min - 1));
+   BSON_ASSERT (bson_in_range_signed (int32_t, int32_min));
+   BSON_ASSERT (bson_in_range_signed (int32_t, 0));
+   BSON_ASSERT (bson_in_range_signed (int32_t, int32_max));
+   BSON_ASSERT (!bson_in_range_signed (int32_t, int32_max + 1));
+
+   BSON_ASSERT (bson_in_range_unsigned (int32_t, 0u));
+   BSON_ASSERT (bson_in_range_unsigned (int32_t, (uint64_t) int32_max));
+   BSON_ASSERT (!bson_in_range_unsigned (int32_t, (uint64_t) (int32_max + 1)));
```
rcsanchez97, line 267:
```suggestion
   BSON_ASSERT (!bson_in_range_unsigned (int32_t, (uint64_t) (int32_max + 1u)));
```

rcsanchez97 (approved):
I have made some suggested changes based on what appear to be missing unsigned specifiers.  Though, I may be mistaken so I am going to go ahead and approve the PR and leave it up to you to decide whether my suggestions are appropriate.

src/libbson/tests/test-bson-cmp.c
```diff
@@ -0,0 +1,309 @@
...
+   const ssize_t ssize_max = SSIZE_MAX;
+
+   BSON_ASSERT (!bson_in_range_signed (int8_t, int8_min - 1));
+   BSON_ASSERT (bson_in_range_signed (int8_t, int8_min));
+   BSON_ASSERT (bson_in_range_signed (int8_t, 0));
+   BSON_ASSERT (bson_in_range_signed (int8_t, int8_max));
+   BSON_ASSERT (!bson_in_range_signed (int8_t, int8_max + 1));
+
+   BSON_ASSERT (bson_in_range_unsigned (int8_t, 0u));
+   BSON_ASSERT (bson_in_range_unsigned (int8_t, (uint64_t) int8_max));
```
kevinAlbs, line 244: Suggest removing unnecessary cast to `uint64_t` for consistency with other calls. Here and below.

kevinAlbs (approved):
LGTM. Nicely done. I think `bson-cmp.h` is going to be useful. I have two comments, but nothing to block merge.

PR #968, opened 2022-04-15
CDRIVER-4272: Add options and code to load the csfle dynamic library

vector-of-bool:
This is an initial basic implementation of the `csfle` options from the new CSE spec changes to support the dynamic library. This also adds mlib as a support library, which now requires some C99 support.

TODO: A a testing scenario that injects the `csfle` library. Ensure tests do not load a `csfle` from the system, only from the isolated test copy or relies on mongocrypd.


src/libmongoc/src/mongoc/mongoc-topology-private.h
```diff
@@ -187,6 +189,11 @@ typedef struct _mongoc_topology_t {
    bson_t *mongocryptd_spawn_args;
 #endif
 
+   // Corresponds to extraOptions.csflePath
+   mstr csfle_override_path;
+   // Corresponds to extraOptions.csfleRequired
+   bool csfle_required;
```
kevinAlbs, line 195: The new options will need to be documented in `mongoc_auto_encryption_opts_set_extra.rst`.

kevinAlbs (commented):
The new options and code to load csfle LGTM.

If mlib is only used for `csfle_override_path`, I suggest leaving it out for now. It may be better to add mlib as a separate PR to expedite unblocking other drivers testing with `csfle`.

build/mongodl.py
```diff
@@ -0,0 +1,665 @@
...
+        for version, target, arch, edition, comp_key, comp_data in matching:
+            found_any = True
+            print('Download: {}\n'
+                  ' Version: {}\n'
+                  '  Target: {}\n'
+                  '    Arch: {}\n'
+                  ' Edition: {}\n'
+                  '    Info: {}\n'.format(comp_key, version, target, arch,
+                                          edition, comp_data))
+    arches, targets, editions, versions, components = next(
```
kevinAlbs, line 273: IIUC the intent is to not print all unique values if there are any matches.
```suggestion
    if found_any:
        return
    arches, targets, editions, versions, components = next(
```

build/mongodl.py
```diff
@@ -0,0 +1,665 @@
...
+                0o655,
+                test=test,
+            )
+    return n_extracted
+
+
+def _maybe_extract_member(out, relpath, pattern, strip, is_dir, opener,
+                          modebits, test):
+    """
+    Try to extract an archive member according to the given argumetns.
```
kevinAlbs, line 527:
```suggestion
    Try to extract an archive member according to the given arguments.
```

build/mongodl.py
```diff
@@ -0,0 +1,665 @@
...
+                        '-T',
+                        help='The target platform for which to download. '
+                        'Use "--list" to list available targets.')
+    dl_grp.add_argument('--arch',
+                        '-A',
+                        help='The architecture for which to download')
+    dl_grp.add_argument(
+        '--edition',
+        '-E',
+        help='The edition of the product to download (Default is "targeted"). '
```
kevinAlbs, line 580: I am not sure what "targeted" means. Since this script is intended for DBX engineers, I suggest setting a default of "enterprise".

build/mongodl.py
```diff
@@ -0,0 +1,665 @@
...
+        'intermediate directories.')
+    dl_grp.add_argument(
+        '--strip-path-components',
+        '-p',
+        dest='strip_components',
+        metavar='N',
+        default=0,
+        type=int,
+        help=
+        'Stip the given number of path components from archive members before '
```
kevinAlbs, line 611:
```suggestion
        'Strip the given number of path components from archive members before '
```

src/libmongoc/CMakeLists.txt
```diff
@@ -856,35 +856,56 @@ if (ENABLE_SHM_COUNTERS STREQUAL "ON")
    )
 endif ()
 
+find_package (Python3 COMPONENTS Interpreter)
+if (TARGET Python3::Interpreter)
```
kevinAlbs, line 860:
```suggestion
if (ENABLE_TESTS AND TARGET Python3::Interpreter)
```

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
```diff
@@ -1281,8 +1323,13 @@ _mongoc_cse_client_enable_auto_encryption (mongoc_client_t *client,
...
 
-   client->topology->crypt = _mongoc_crypt_new (
-      opts->kms_providers, opts->schema_map, opts->tls_opts, error);
+   client->topology->crypt =
+      _mongoc_crypt_new (opts->kms_providers,
+                         opts->schema_map,
+                         opts->tls_opts,
+                         client->topology->csfle_override_path,
+                         client->topology->csfle_required,
+                         error);
```
kevinAlbs, line 1332: `opts->bypass_auto_encryption` may also need to be passed to disable loading `csfle`. That is new behavior added in DRIVERS-2287.

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
```diff
@@ -1514,8 +1565,12 @@ mongoc_client_encryption_new (mongoc_client_encryption_opts_t *opts,
...
 
    client_encryption->kms_providers = bson_copy (opts->kms_providers);
-   client_encryption->crypt = _mongoc_crypt_new (
-      opts->kms_providers, NULL /* schema_map */, opts->tls_opts, error);
+   client_encryption->crypt = _mongoc_crypt_new (opts->kms_providers,
+                                                 NULL /* schema_map */,
+                                                 opts->tls_opts,
+                                                 NULL /* No csfle path */,
+                                                 false /* csfle not requried */,
+                                                 error);
```
kevinAlbs, line 1573: This should also pass information through to disable loading `csfle`. It is not needed for a `mongoc_client_encryption_t`. This is new behavior from DRIVERS-2287.

src/libmongoc/CMakeLists.txt
```diff
@@ -1058,6 +1079,10 @@ mongoc_add_test (test-libmongoc FALSE ${test-libmongoc-sources})
 mongoc_add_test (test-mongoc-gssapi FALSE ${PROJECT_SOURCE_DIR}/tests/test-mongoc-gssapi.c)
 mongoc_add_test (test-mongoc-cache FALSE ${PROJECT_SOURCE_DIR}/tests/test-mongoc-cache.c)
 
+if (TARGET get-csfle AND TARGET test-libmongoc)
+   target_compile_definitions (test-libmongoc PRIVATE "TESTING_CSFLE_OVERRIDE_PATH=\"$<TARGET_PROPERTY:get-csfle,CSFLE_PATH>\"")
```
kevinAlbs, line 1083: Is `get-csfle` expected to run in Evergreen? The ASAN `*-cse` task failures look unrelated (I think due to [`BSON_MEMCHECK`](http://mongoc.org/libbson/current/valgrind.html) shenanigans with libmongocrypt.). I do not see logs for the get-csfle task in [the VS 2017 task](https://spruce.mongodb.com/task/mongo_c_driver_windows_2017_test_latest_server_auth_sasl_openssl_cse_patch_28d66415d35bec28dba76c054bde1a0c9d69f56c_62606bcd32f41728ef1faa7c_22_04_20_20_23_43/logs?execution=0).

kevinAlbs (commented):
Looking good overall.

The mongodl.py script looks robust and generally useful.

Great catch with the exit and allocation unsafety within the child process.

src/libmongoc/CMakeLists.txt
vector-of-bool: Yes, it should execute by default. I'll have to inspect further. This will not run if `get-csfle` is not defined, which will not be defined unless `Python3::Interpreter` is defined. [In these build logs](https://spruce.mongodb.com/task/mongo_c_driver_windows_2017_debug_compile_sasl_openssl_cse_patch_28d66415d35bec28dba76c054bde1a0c9d69f56c_62606bcd32f41728ef1faa7c_22_04_20_20_23_43?execution=0) it says that `Python3` is not found, which means it is likely too old of a CMake version to have the `FindPython3.cmake` module.

We may or may not be okay with this, as the tests should always pass regardless of whether `csfle` is injected into them. Having some builds that use `mongocryptd` and some that use `csfle` may be useful, but we may want to make that more explicit.

build/mongodl.py
vector-of-bool: I'm not sure what it means either. I just chose it as an arbitrary default. I'm finding that "`enterprise`" will be better since one needs it for `csfle` on `mongocryptd` anyway.

build/mongodl.py
```diff
@@ -0,0 +1,665 @@
...
+    ''')
+    db.executescript(r'''
+        CREATE TABLE IF NOT EXISTS past_downloads (
+            url TEXT NOT NULL UNIQUE,
+            etag TEXT,
+            last_modified TEXT
+        )
+    ''')
+    changed, full_json = _download_file(
+        db, 'https://downloads.mongodb.org/full.json')
```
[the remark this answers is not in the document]
vector-of-bool: Sorry, I'm not really sure what you mean in this comment.

As for "latest" URLs: They are _not_ contained within `full.json` (unless you do a scan of the versions and pick the highest version yourself). Except for one-off quick testing, I'm wary of any automated system that automatically pulls a moving target as a dependency.

src/libmongoc/CMakeLists.txt
kevinAlbs: > We may or may not be okay with this, as the tests should always pass regardless of whether csfle is injected into them. Having some builds that use mongocryptd and some that use csfle may be useful, but we may want to make that more explicit.

Having at least one task use `csfle` and one task use `mongocryptd` seems desirable. I prefer it be made explicit. Otherwise, Evergreen host changes may change which is being used, and we may not notice that.

Perhaps a simple option is to add a cmake flag: `DISABLE_CSFLE_DOWNLOAD`? One task in Evergreen can set that flag to guarantee it uses mongocryptd?

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -945,11 +948,44 @@ _mongoc_crypt_new (const bson_t *kms_providers,
...
+   }
+
    if (!mongocrypt_init (crypt->handle)) {
       _crypt_check_error (crypt->handle, error, true);
       goto fail;
    }
 
+   if (csfle_required) {
+      const char *s = mongocrypt_csfle_version_string (crypt->handle, NULL);
+      if (!s || strlen (s) == 0) {
```
eramongodb, line 973:
```suggestion
      uint32_t len = 0u;
      const char *s = mongocrypt_csfle_version_string (crypt->handle, &len);
      if (!s || len == 0u) {
```
Minor suggestion to avoid `strlen ()`.

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
```diff
@@ -888,6 +888,18 @@ _do_spawn (const char *path, char **args, bson_error_t *error)
 }
 #else
 
+/**
+ * @brief Exit the calling process *immediately* without running any per-process
+ * cleanup handlers.
+ *
+ * @param status The exist status for the process.
```
eramongodb, line 895:
```suggestion
 * @param status The exit status for the process.
```

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
```diff
@@ -888,6 +888,18 @@ _do_spawn (const char *path, char **args, bson_error_t *error)
...
+ */
+static void
+fast_exit (int status)
```
eramongodb, line 898: Why `fast_exit ()` instead of directly calling `_exit ()`? Is it for platform-compatibility reasons?

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
```diff
@@ -888,6 +888,18 @@ _do_spawn (const char *path, char **args, bson_error_t *error)
...
+{
+   _exit (status);
```
eramongodb, line 900: Why `_exit ()` (POSIX only) and not `_Exit ()` ([C99 equivalent](https://man7.org/linux/man-pages/man2/exit.2.html))?

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
```diff
@@ -1181,6 +1198,31 @@ _parse_extra (const bson_t *extra,
...
+
+      if (bson_iter_init_find (&iter, extra, "csfleRequired")) {
+         if (!BSON_ITER_HOLDS_BOOL (&iter)) {
+            bson_set_error (error,
+                            MONGOC_ERROR_CLIENT,
+                            MONGOC_ERROR_CLIENT_INVALID_ENCRYPTION_ARG,
+                            "Expected a bool for 'csfleRequired'");
+            GOTO (fail);
+         }
+         topology->csfle_required = bson_iter_bool (&iter);
```
eramongodb, line 1224:
```suggestion
         topology->csfle_required = bson_iter_bool_unsafe (&iter);
```

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
```diff
@@ -1181,6 +1198,31 @@ _parse_extra (const bson_t *extra,
...
+      if (bson_iter_init_find (&iter, extra, "csflePath")) {
+         if (!BSON_ITER_HOLDS_UTF8 (&iter)) {
+            bson_set_error (error,
+                            MONGOC_ERROR_CLIENT,
+                            MONGOC_ERROR_CLIENT_INVALID_ENCRYPTION_ARG,
+                            "Expected a string for 'csflePath'");
+            GOTO (fail);
+         }
+         uint32_t len;
+         const char *ptr = bson_iter_utf8 (&iter, &len);
```
eramongodb, line 1211:
```suggestion
         const char *ptr = bson_iter_utf8_unsafe (&iter, &len);
```

src/libmongoc/doc/mongoc_auto_encryption_opts_set_extra.rst
```diff
@@ -25,6 +25,34 @@ Parameters
 * ``mongocryptdBypassSpawn`` set to true to prevent the driver from spawning the mongocryptd process (default behavior is to spawn).
 * ``mongocryptdSpawnPath`` set to a path (with trailing slash) to search for mongocryptd (defaults to empty string and uses default system paths).
 * ``mongocryptdSpawnArgs`` set to an array of string arguments to pass to ``mongocryptd`` when spawning (defaults to ``[ "--idleShutdownTimeoutSecs=60" ]``).
+* ``csflePath`` - Set a filepath string referring to a ``csfle`` dynamic library
+  file. Default is undefined.
```
eramongodb, line 29:
```suggestion
  file. Unset by default.
```
To avoid confusion with "undefined behavior".

src/libmongoc/doc/mongoc_auto_encryption_opts_set_extra.rst
```diff
@@ -25,6 +25,34 @@ Parameters
...
+
+  * If not specified (the default), ``libmongocrypt`` will attempt to load
```
eramongodb, line 31:
```suggestion
  * If not set (the default), ``libmongocrypt`` will attempt to load
```
Wording consistency with suggestion above.

src/libmongoc/doc/mongoc_auto_encryption_opts_set_extra.rst
```diff
@@ -25,6 +25,34 @@ Parameters
...
+    ``csfle`` using the host system's default dynamic-library-search system.
+  * If used, should be the path to the ``csfle`` dynamic library file itself,
+    not just the directory that contains it.
```
eramongodb, line 34:
```suggestion
  * If set, the pathshould be to the ``csfle`` dynamic library file itself,
    not the directory that contains it.
```

src/libmongoc/doc/mongoc_auto_encryption_opts_set_extra.rst
```diff
@@ -25,6 +25,34 @@ Parameters
...
+  * If the given path is a relative path and the first path component is
+    ``$ORIGIN``, the ``$ORIGIN`` component will be replaced with the absolute
+    path to the directory containing the ``libmongocrypt`` library in use by the
+    application.
+
+    .. note:: No other ``RPATH``/``RUNPATH``-style substitutions are availbale.
```
eramongodb, line 40:
```suggestion
    .. note:: No other ``RPATH``/``RUNPATH``-style substitutions are available.
```

src/libmongoc/doc/mongoc_auto_encryption_opts_set_extra.rst
```diff
@@ -25,6 +25,34 @@ Parameters
...
+
+  * If the given path is a relative path, the path will be resolves relative to
+    the working directory of the operating system process.
```
eramongodb, line 43:
```suggestion
  * If the given path is a relative path, the path will be resolved relative to
    the working directory of the operating system process.
```

src/libmongoc/doc/mongoc_auto_encryption_opts_set_extra.rst
```diff
@@ -25,6 +25,34 @@ Parameters
...
+  * If this option is specified and ``libmongocrypt`` fails to load ``csfle``
+    from that filepath, ``libmongocrypt`` will fail to initialize with a
+    hard-error, and will not attempt to search for ``csfle`` in any other
+    locations.
```
eramongodb, line 47:
```suggestion
  * If this option is set and ``libmongocrypt`` fails to load ``csfle``
    from the given filepath, ``libmongocrypt`` will fail to initialize with a
    hard-error and will not attempt to search for ``csfle`` in any other
    locations.
```

src/libmongoc/doc/mongoc_auto_encryption_opts_set_extra.rst
```diff
@@ -25,6 +25,34 @@ Parameters
...
+
+* ``csfleRequired`` - If set to ``true``, and ``libmongocrypt`` fails to load a
+  ``csfle`` dynamic library, generate a hard-error and do not attempt to spawn
```
eramongodb, line 50: Does a "hard-error" constitute something more than libmongocrypt initialization failure?

build/mongodl.py
```diff
@@ -0,0 +1,666 @@
...
+def infer_target():
+    if sys.platform == 'win32':
+        return 'windows'
+    if sys.platform == 'darwin':
+        return 'macos'
+    # Now the tricky bit
+    if Path('/etc/os-release').is_file():
+        return _infer_target_os_rel()
+    raise RuntimeError("Don't know yet how to find the default download "
+                       "'--target' for this system. Please contribute!")
```
eramongodb, line 80:
```suggestion
    raise RuntimeError(
        "We don't know how to find the default download '--target' on this system. Please contribute!")
```
Minor wording improvement suggestion for consistency with other error messages.

build/mongodl.py
```diff
@@ -0,0 +1,666 @@
...
+    cached = _download_file(db, data['url']).path
+    return _expand_archive(cached,
+                           out_dir,
+                           pattern,
+                           strip_components,
+                           test=test)
+
+
+def pathjoin(items):
+    'Return a path formed by joining the given path components'
```
eramongodb, line 398:
```suggestion
    """
    Return a path formed by joining the given path components
    """
```

.gitignore
```diff
@@ -16,6 +16,7 @@ CMakeFiles
 CPackConfig.cmake
 CPackSourceConfig.cmake
 CTestTestfile.cmake
+_build/
```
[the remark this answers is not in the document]
vector-of-bool: My own defaults. I was just getting annoyed of resetting it or adding it to my own `.gitignore` without committing it.

build/maintainer-flags.txt
```diff
@@ -1,11 +1,9 @@
 -Wall
--Wdeclaration-after-statement
```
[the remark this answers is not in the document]
vector-of-bool: Correct, and we are already phasing-out C89 compatibility incrementally.

build/maintainer-flags.txt
```diff
@@ -1,11 +1,9 @@
...
 -Wempty-body
 -Wformat
 -Wformat-nonliteral
 -Wformat-security
 -Winit-self
--Winline
```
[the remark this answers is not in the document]
vector-of-bool: This warnings is issued when GCC decides not to inline an `extern inline` function. It's inherently meaningless in modern code where the inliner (usually) knows better. This can be reverted for now because only `mlib` contained `extern inline` functions.

build/mongodl.py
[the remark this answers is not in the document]
vector-of-bool: Since this script is entirely based around `full.json`, I'll postpone adding "latest" options, although it may be useful in the future. It would be handy if Evergreen posted a machine-parseable listing (Does it?)

src/libmongoc/doc/mongoc_auto_encryption_opts_set_extra.rst
vector-of-bool: The "hard-error" means that initialization fails rather than error-and-continue (as with the case of searching). This should be made more clear.

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
vector-of-bool: Initially yes, as `_exit()` is part of POSIX whereas `_Exit` is part of C99, and some platforms do not have `_Exit`. Since this segment of code is conditionally compiled on unix-like systems, calling `_exit()` directly will be fine.

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
vector-of-bool: (See other comment)

src/libmongoc/src/mongoc/mongoc-crypt.c
vector-of-bool: I wrote the API and didn't even recognize this. Oops.

src/libmongoc/tests/test-mongoc-client-side-encryption.c
```diff
@@ -1887,6 +1887,9 @@ _reset (mongoc_client_pool_t **pool,
    *opts = mongoc_auto_encryption_opts_new ();
    extra = BCON_NEW ("mongocryptdBypassSpawn", BCON_BOOL (true));
    mongoc_auto_encryption_opts_set_extra (*opts, extra);
+#ifdef TESTING_CSFLE_OVERRIDE_PATH
```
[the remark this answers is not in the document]
vector-of-bool: Good question. Because I want the acquisition of `csfle` to be fully automatable and human-error-free, and `csfle` is managed by another team entirely (i.e. driver devs won't likely be trying to iterate on it directly), it is built directly into the build process rather than needing to be specified manually. Of course, specifying an environment variable could also be done via scripting (e.g. with the `ENVIRONMENT` CTest property). This was the easiest way to get it working for now. In the future, one will likely want to support injecting alternative `csfle` libraries. I put this as YAGNI for now.

build/mongodl.py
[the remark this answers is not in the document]
kevinAlbs: SGTM. I am not aware of machine-parseable listing of latest builds.

vector-of-bool:
@kevinAlbs I've found that I need to add an option for a test to completely disable `csfle` as the test will rely on mongocryptd behaviors. Take a look at the latest few commits.

src/libmongoc/CMakeLists.txt
```diff
@@ -856,35 +856,67 @@ if (ENABLE_SHM_COUNTERS STREQUAL "ON")
    )
 endif ()
 
+cmake_dependent_option (
+   MONGOC_TEST_USE_CSFLE "If TRUE, download and use a csfle dynamic library for client-side-encryption testing. Otherwise will rely on mongocryptd"
+   TRUE ENABLE_TESTS FALSE
+)
+
+if (MONGOC_TEST_USE_CSFLE)
+   find_program (_PYTHON3_EXE NAMES python3 py python)
```
eramongodb, line 865: If I understand correctly, if `python3` is not found, then `py` or `python` may be used instead, either of which may be `python2` rather than `python3`. Is this acceptable?

src/libmongoc/CMakeLists.txt
```diff
@@ -856,35 +856,67 @@ if (ENABLE_SHM_COUNTERS STREQUAL "ON")
...
-      target_include_directories (${test} PRIVATE ${PROJECT_SOURCE_DIR}/tests)
+      target_link_libraries (${test}
+         PRIVATE
+            $<$<C_COMPILER_ID:MSVC>:DbgHelp.dll>
+            $<$<PLATFORM_ID:Windows>:shlwapi>
+            $<$<BOOL:${use_shared}>:mongoc_shared>
+            $<$<NOT:$<BOOL:${use_shared}>>:mongoc_static>
+            ${LIBRARIES}
+            ${RESOLV_LIBRARIES}
+         )
```
eramongodb, line 902:
```suggestion
      )
```

src/libmongoc/CMakeLists.txt
```diff
@@ -856,35 +856,67 @@ if (ENABLE_SHM_COUNTERS STREQUAL "ON")
...
+   find_program (_PYTHON3_EXE NAMES python3 py python)
+   mark_as_advanced (_PYTHON3_EXE)
+   if (NOT _PYTHON3_EXE)
+      message (SEND_ERROR
+         "MONGOC_TEST_USE_CSFLE is enabled, but we were unable to find a Python 3 interpreter to use mongodl.py. "
+         "Install a Python 3 or set MONGOC_TEST_USE_CSFLE to FALSE"
+      )
+   endif ()
+
+   get_filename_component (csfle_dest "${CMAKE_CURRENT_BINARY_DIR}/_csfle/mongo_csfle_v1${CMAKE_SHARED_LIBRARY_SUFFIX}" ABSOLUTE)
```
eramongodb, line 874: Confirming, is this line setting `CMAKE_SHARED_LIBRARY_SUFFIX`? Otherwise, where and how is `CMAKE_SHARED_LIBRARY_SUFFIX` defined?

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
```diff
@@ -1181,6 +1198,39 @@ _parse_extra (const bson_t *extra,
...
+            GOTO (fail);
+         }
+         topology->csfle_required = bson_iter_bool_unsafe (&iter);
+      }
+
+      /** The __csfleDisabled option is a private API option only used for
+       * testing purposes. */
+      if (bson_iter_init_find (&iter, extra, "__csfleDisabled")) {
+         BSON_ASSERT (BSON_ITER_HOLDS_BOOL (&iter) &&
+                      "__csfleDisabled is for testing purposes only");
```
eramongodb, line 1231: This error message seems a little oddly worded given the assertion being tested. It seems to suggest `__csfleDisabled` shouldn't be present at all, rather than that it should contain a `bool` value. Should it be worded something like, "__csfleDisabled should only have been defined by X for testing purposes"?

src/libmongoc/tests/json/client_side_encryption/explain.json
```diff
@@ -123,6 +123,9 @@
         "autoEncryptOpts": {
           "kmsProviders": {
             "aws": {}
+          },
+          "extra": {
+            "__csfleDisabled": true
```
eramongodb, line 128: Confirming, has the spec test this test file corresponds to been updated in the spec repo? If not, this test might not belong under the `json` directory.

src/libmongoc/tests/test-mongoc-client-side-encryption.c
```diff
@@ -2321,16 +2324,17 @@ test_bypass_spawning_via_mongocryptdBypassSpawn (void *unused)
 
    /* Create a MongoClient with encryption enabled */
    client_encrypted = test_framework_new_default_client ();
-   extra =
-      BCON_NEW ("mongocryptdBypassSpawn",
+   extra = BCON_NEW ("mongocryptdBypassSpawn",
```
eramongodb, line 2327: Missing ClangFormat changes.

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -945,11 +949,47 @@ _mongoc_crypt_new (const bson_t *kms_providers,
...
    if (!mongocrypt_init (crypt->handle)) {
       _crypt_check_error (crypt->handle, error, true);
       goto fail;
    }
 
+   if (csfle_required) {
+      BSON_ASSERT (!csfle_disabled &&
+                   "Both csfleRequired and __csfleDisabled were set 'true'");
+      uint32_t len = 0;
+      const char *s = mongocrypt_csfle_version_string (crypt->handle, &len);
```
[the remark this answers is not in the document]
vector-of-bool: IIUC, we're lifting that restriction since all compilers support it and we've removed the c89-check task.

src/libmongoc/CMakeLists.txt
vector-of-bool: That's correct. That will end up causing an error and the fix would be similarly disabling csfle testing, although the error message won't be as pretty as when Python is completely absent. If we could use a newer CMake with a FindPython3 module, this would be a non-issue. Maybe someday…

Barely related aside pain: Windows now ships with a `python` command that just launches the Windows store page, so it's not even really possible to reliably tell whether the `python` executable you just found is even the real thing without trying to run a script with it, in which can you might accidentally open the Windows store page. Maybe they have some environment variable to suppress it, but still: pain.

src/libmongoc/CMakeLists.txt
vector-of-bool: That's a built-in variable defined by CMake (or a toolchain file) that tells you the default file extension of runtime libraries (e.g. `.so`, `.dll`, or `.dylib`)

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
vector-of-bool: Probably, yes

src/libmongoc/tests/json/client_side_encryption/explain.json
vector-of-bool: True. This is somewhat experimenting on how to "disable" csfle for testing. We don't want to publish an "official" way to disable csfle, but this spec test also changed behavior depending on whether csfle is present. It's a catch-22.

src/libmongoc/tests/test-mongoc-client-side-encryption.c
vector-of-bool: What do you mean?

eramongodb: Lines immediately following line 2327 are not indented to match recent changes.

vector-of-bool: Oh gross.

I have my VSCode set to hide whitespace in the working diff. When I selected the range and and staged it, it must have not staged the whitespace changes because they were hidden.

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
```diff
@@ -1181,6 +1198,39 @@ _parse_extra (const bson_t *extra,
...
+      if (bson_iter_init_find (&iter, extra, "csflePath")) {
+         if (!BSON_ITER_HOLDS_UTF8 (&iter)) {
+            bson_set_error (error,
+                            MONGOC_ERROR_CLIENT,
+                            MONGOC_ERROR_CLIENT_INVALID_ENCRYPTION_ARG,
+                            "Expected a string for 'csflePath'");
+            GOTO (fail);
+         }
+         size_t len;
+         const char *ptr = bson_iter_utf8_unsafe (&iter, &len);
```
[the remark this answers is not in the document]
eramongodb: Deemed inapplicable due to recent removal of compile-c89 Evergreen task.

src/libmongoc/tests/json/client_side_encryption/explain.json
eramongodb: I think it is fine to include the field in spec tests so long as it is clearly documented as for testing purposes only. However, if the specification repo will not updated to match proposed test file contents, I would like the `__csfleDisabled` test to be moved out of the `json` directory and tested as a prose test instead.

eramongodb (approved):
Other than the issue with `__csfleDisabled` in a modified unified spec test, LGTM.

src/libmongoc/tests/test-mongoc-client-side-encryption.c
```diff
@@ -2321,16 +2324,17 @@ test_bypass_spawning_via_mongocryptdBypassSpawn (void *unused)
...
+                     BCON_BOOL (true),
+                     "mongocryptdSpawnArgs",
+                     "[",
+                     "--pidfilepath=bypass-spawning-mongocryptd.pid",
+                     "--port=27021",
+                     "]",
+                     "mongocryptdURI",
+                     "mongodb://localhost:27021/?serverSelectionTimeoutMS=1000",
+                     "__csfleDisabled",
```
kevinAlbs, line 2336: All drivers will need a solution to skip this prose test when running with `csfle`. I prefer not adding a hidden option to the API. I think it adds small risk that users rely on it.

I suggest only run this test if the test runner expects to use `csfle`. That could be passed through an environment variable. There is precedence: (e.g. `MONGOC_TEST_IS_SERVERLESS`, `MONGOC_TEST_LOADBALANCED`.

kevinAlbs (commented):
I left a suggestion to skip tests based on test environment rather than a `__csfleDisabled` option.

kevinAlbs (approved):
Latest changes and tests look good. Skipping  Windows is not ideal, but seems low risk given that it has been validated locally, and libmongocrypt tests with the stub library in CI.

PR #973, opened 2022-04-25
fix CMake version check for OpenSSL 1.1+ on Windows

TheDorkKnight:
As it stands, this check does not function as intended on the CMake versions that I tried (3.22 and 3.23).

I suspect it would not work from CMake 3.10 until CMake 3.70 (were such a version ever to exist); the comparison appears to a floating-point comparison instead of a version comparison.

I took the error message

> `requires CMake 3.7 or later`

at face value and assumed that `${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.7` is correct instead of `${CMAKE_VERSION} VERSION_GREATER 3.7` which would be the direct translation from the previous version of the script.

### DIFF
--- a/src/libmongoc/CMakeLists.txt
+++ b/src/libmongoc/CMakeLists.txt
@@ -196,8 +196,8 @@ if (NOT ENABLE_SSL STREQUAL OFF)
 endif ()
 
 if (OPENSSL_FOUND)
-   if (WIN32 AND OPENSSL_VERSION GREATER 1.1 AND NOT
-         ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} GREATER 3.7)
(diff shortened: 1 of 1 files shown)


rcsanchez97 (changes requested):
Based on the CMake docs, the `VERSION_GREATER_EQUAL` test was introduced in CMake 3.7.  However, our build only requires CMake 3.1.  This will need to be updated to be functional CMake 3.1 before it can be merged.

TheDorkKnight:
Sounds good. Here's a link to the version comparisons [documentation link](https://cmake.org/cmake/help/latest/command/if.html#version-comparisons) for posterity.

As you said:
> if(<variable|string> VERSION_GREATER_EQUAL <variable|string>)
New in version 3.7:

vector-of-bool (approved):

rcsanchez97 (approved):

PR #972, opened 2022-04-21
CDRIVER-4085 add server connectionId to command monitoring events

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/625f6c3057e85a68b11bea04/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

I've reviewed the tasks that are green on the base commit and red on the patch build.  The few failures are unrelated (e.g., the longstanding ccache bug on Ubuntu, the failure to build with static libssl on RHEL, etc.).

PR #976, opened 2022-04-27
CDRIVER-4085 fix docs for server_connection_id getters

jmikola:
This removes copypasta that was missed in https://github.com/mongodb/mongo-c-driver/commit/ba73833f18191b34acc990bdca91bb9e68c90dfb (#972)

### DIFF
--- a/src/libmongoc/doc/mongoc_apm_command_failed_get_server_connection_id.rst
+++ b/src/libmongoc/doc/mongoc_apm_command_failed_get_server_connection_id.rst
@@ -8,11 +8,16 @@ Synopsis
 
 .. code-block:: c
 
-  void *
+  int32_t
   mongoc_apm_command_failed_get_server_connection_id (
      const mongoc_apm_command_failed_t *event);
 
 Returns this event's context.
+
+Returns the server connection ID for the command. The server connection ID is
+distinct from the server ID (:symbol:`mongoc_apm_command_failed_get_server_id`)
+and is returned by the hello or legacy hello response as "connectionId" from the
+server on 4.2+.
 
 Parameters
 ----------
@@ -22,7 +27,7 @@ Parameters
 Returns
 -------
 
-The pointer passed with :symbol:`mongoc_client_set_apm_callbacks` or :symbol:`mongoc_client_pool_set_apm_callbacks`.
+The server connection ID as a positive integer or -1 if it is not available.
 
 .. seealso::
 
--- a/src/libmongoc/doc/mongoc_apm_command_started_get_server_connection_id.rst
+++ b/src/libmongoc/doc/mongoc_apm_command_started_get_server_connection_id.rst
@@ -8,11 +8,14 @@ Synopsis
 
 .. code-block:: c
 
-  void *
+  int32_t
   mongoc_apm_command_started_get_server_connection_id (
      const mongoc_apm_command_started_t *event);
 
-Returns this event's context.
+Returns the server connection ID for the command. The server connection ID is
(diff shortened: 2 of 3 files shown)


src/libmongoc/doc/mongoc_apm_command_failed_get_server_connection_id.rst
```diff
@@ -22,7 +27,7 @@ Parameters
 Returns
 -------
 
-The pointer passed with :symbol:`mongoc_client_set_apm_callbacks` or :symbol:`mongoc_client_pool_set_apm_callbacks`.
+The server connection ID as a positive integer or -1 if it is not available.
```
jmikola, line 30: https://github.com/mongodb/mongo-c-driver/commit/ba73833f18191b34acc990bdca91bb9e68c90dfb#diff-de5aea95eea3d2b786d33b8b694d349c445847e9cc506502c6824e05f177cc0cR48 defined `MONGOC_NO_SERVER_CONNECTION_ID` internally as `-1`, which is similar to what was done for `MONGOC_RTT_UNSET`.

I previously opened a ticket to document the unset RTT return value in [CDRIVER-4176](https://jira.mongodb.org/browse/CDRIVER-4176) and @kevinAlbs proposed just making `MONGOC_RTT_UNSET` public. Should we just consider doing so for `MONGOC_NO_SERVER_CONNECTION_ID`?

If so, would we want to keep the same name or change it to `MONGOC_SERVER_CONNECTION_ID_UNSET`?

rcsanchez97: So, I initially considered using `MONGOC_SERVER_CONNECTION_ID_UNSET`.  However, it seemed to me like "unset" was something different from "not present".  Based on the naming of other nearby constants, like `MONGOC_NO_SESSIONS` and `MONGOC_NO_SET_VERSION`, the better name appeared to be `MONGOC_NO_SERVER_CONNECTION_ID`.  That said, I'm not opposed to your suggested change.  I simply wanted to state my rationale more thoroughly for your consideration.

rcsanchez97 (approved):
LGTM

Thanks for catching these oversights.  I was so focused on making sure that I documented each function and that I named the files correctly, that I completely forgot to read the documentation after copying it to make any necessary changes.  :facepalm:

src/libmongoc/doc/mongoc_apm_command_failed_get_server_connection_id.rst
eramongodb: I am slightly in favor of `MONGOC_NO_SERVER_CONNECTION_ID` but am not opposed to the `UNSET` naming scheme.

eramongodb (approved):

src/libmongoc/doc/mongoc_apm_command_failed_get_server_connection_id.rst
jmikola: Ah, that makes much more sense and I agree `MONGOC_NO_SERVER_CONNECTION_ID` is the better choice. I agree that at the time we call this method, we either know that their's an ID or not, which is a different situation than RTT.

Any thoughts on whether we should make the constant public? I'm happy to leave this as-is and just add it as a todo item in [CDRIVER-4176](https://jira.mongodb.org/browse/CDRIVER-4176).

rcsanchez97: I would lean toward adding it to CDRIVER-4176 and changing the visibility of all of them at the same time.

PR #977, opened 2022-04-28
Fixes for csfle downloading on some Linux platforms

vector-of-bool:
This fixes some CI failures that appeared following the merge of #968 : In particular, some older Linux platforms are missing Python features that mongodl used, and some platforms do not have `csfle` available at all (For these it is conditionally disabled). These changes have been tested on all erring platforms.


kevinAlbs (approved):

build/mongodl.py
```diff
@@ -84,12 +89,15 @@ def infer_target():
...
-    content = Path('/etc/os-release').read_text()
+    with Path('/etc/os-release').open('r', encoding='utf-8') as f:
+        content = f.read()
     id_re = re.compile(r'\bID=("?)(.*)\1')
     mat = id_re.search(content)
     assert mat, 'Unable to detect ID from [/etc/os-release] content:\n{}'.format(
         content)
     os_id = mat.group(2)
+    if os_id == 'arch':
+        return 'rhel80'
```
eramongodb, line 100: Why is `'arch'` special-cased here instead of using `DISTRO_ID_MAP` and `DISTRO_ID_TO_TARGET`? Git blame says Arch is a special case but does not explain why (would be nice if rationale is also documented as a comment).

build/mongodl.py
```diff
@@ -214,9 +222,21 @@ def _import_json_data(db, json_file):
                 )
 
 
+def _mkdir(dirpath):
```
eramongodb, line 225: Was `parents=True` among the unsupported Python features? If so, I would like that to be documented here (reason for deliberately not using an otherwise preferable feature).

eramongodb (approved):

PR #978, opened 2022-04-29
CDRIVER-4373 do not set bulk let option unnecessarily

jmikola:
https://jira.mongodb.org/browse/CDRIVER-4373

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-collection.c
+++ b/src/libmongoc/src/mongoc/mongoc-collection.c
@@ -3228,7 +3228,10 @@ mongoc_collection_create_bulk_operation_with_opts (
                                       write_flags,
                                       wc);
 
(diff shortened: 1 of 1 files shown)


src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -3228,7 +3228,10 @@ mongoc_collection_create_bulk_operation_with_opts (
                                       write_flags,
                                       wc);
 
-   mongoc_bulk_operation_set_let (bulk, &bulk_opts.let);
+   if (!bson_empty (&bulk_opts.let)) {
```
jmikola, line 3231: `_mongoc_bulk_opts_parse` already initializes `bulk_opts.let`, so we need not use `bson_empty0`

vector-of-bool (approved):
A simple fix. LGTM.

PR #980, opened 2022-05-02
Fix fallback for missing stdbool.h

AgostinoSturaro:
I've noticed that if `stdbool.h` is missing (not supported by the compiler), the existing fallback is not correctly called.
This used to work in version 1.14.1.

The first part of the issue is in `src/libbson/CMakeLists.txt`.
If the check finds no `stdbool.h`, the `BSON_HAVE_STDBOOL_H` macro was not initialized to 0.

The other part of the issue is that `stdbool.h` is being included directly rather than via `bson-compat.h`, which provides a fallback if `stdbool.h` is missing, along with other commonly used headers.

### DIFF
--- a/src/kms-message/src/kms_crypto.h
+++ b/src/kms-message/src/kms_crypto.h
@@ -17,8 +17,7 @@
 #ifndef KMS_MESSAGE_KMS_CRYPTO_H
 #define KMS_MESSAGE_KMS_CRYPTO_H
 
-#include <stdbool.h>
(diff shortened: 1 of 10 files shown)


AgostinoSturaro:
@AgostinoSturaro I can't see the failed checks because it requires a login.

kevinAlbs:
Thank you for the PR. As noted in CDRIVER-4376, the C driver does not support versions of older than Visual Studio 2013.

PR #982, opened 2022-05-03
CDRIVER-4199: Add support for the comment field to all helpers

jmikola:
https://jira.mongodb.org/browse/CDRIVER-4199
https://jira.mongodb.org/browse/CDRIVER-4374


src/libmongoc/src/mongoc/mongoc-bulk-operation.c
```diff
@@ -394,15 +401,23 @@ mongoc_bulk_operation_insert_with_opts (mongoc_bulk_operation_t *bulk,
       }
    }
 
+   if (bulk->comment.value_type != BSON_TYPE_EOD) {
+      bson_append_value (&cmd_opts, "comment", 7, &bulk->comment);
+   }
+
+   /* TODO: mongoc_bulk_insert_opts_t specifies allow_extra=False, so there was
+    * no reason to previously pass &insert_opts.extra as cmd_opts */
```
jmikola, line 409: This TODO exists for the benefit of code review. Please confirm that my understanding is correct and I'll remove this accordingly.

src/libmongoc/src/mongoc/mongoc-bulk-operation.c
```diff
@@ -973,6 +992,25 @@ mongoc_bulk_operation_set_bypass_document_validation (
...
+{
+   BSON_ASSERT (bulk);
+   BSON_ASSERT (comment);
+   BSON_ASSERT (comment->value_type != BSON_TYPE_EOD);
+
+   /* This method cannot be called after appending operations, as the CRUD spec
+    * states the option should apply to all commands. Since commands are
+    * initialized as operations are added, allowing "comment" to be changed at
+    * any time could violate that contract. */
+   BSON_ASSERT (bulk->commands.len == 0);
```
jmikola, line 1007: Same logic as the `let` setter.

src/libmongoc/src/mongoc/mongoc-change-stream.c
```diff
@@ -243,6 +243,10 @@ _make_cursor (mongoc_change_stream_t *stream)
    bson_init (&command);
    bson_copy_to (&(stream->opts.extra), &command_opts);
 
+   if (stream->opts.comment.value_type != BSON_TYPE_EOD) {
+      bson_append_value (&command_opts, "comment", 7, &stream->opts.comment);
+   }
```
jmikola, line 248: This is not currently tested as the change stream spec tests are currently skipped. More work is required to update the unified test runner ([CDRIVER-3973](https://jira.mongodb.org/browse/CDRIVER-3973)) and that seemed like too much for this PR.

I'd propose scheduling those tickets immediately after this work (or at least in the same fixVersion).

src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -1952,12 +1965,18 @@ mongoc_collection_insert_many (mongoc_collection_t *collection,
       return false;
    }
 
+   if (insert_many_opts.crud.comment.value_type != BSON_TYPE_EOD) {
+      bson_append_value (
+         &cmd_opts, "comment", 7, &insert_many_opts.crud.comment);
+   }
+
+   if (!bson_empty (&insert_many_opts.extra)) {
+      bson_concat (&cmd_opts, &insert_many_opts.extra);
```
jmikola, line 1974: I realize that in `mongoc_collection_insert_one` I did this backwards and appended `comment` before concatenating `extra`. Does that matter at all? I assume `extra` will never include any known options so there's no concern with a possible key conflict here (were that to happen, I expect the server would just use the first instance of `comment` in the command).

Happy to swap the order here if we'd like to be consistent though.

src/libmongoc/src/mongoc/mongoc-cursor.c
```diff
@@ -62,6 +62,21 @@ _mongoc_cursor_set_opt_int64 (mongoc_cursor_t *cursor,
...
 
+static const bson_value_t *
+_mongoc_cursor_get_opt (const mongoc_cursor_t *cursor, const char *option)
+{
+   bson_iter_t iter;
+
+   if (bson_iter_init_find (&iter, &cursor->opts, option)) {
+      /* TODO: bson_iter_value's return value is invalidated if the iter is
+       * modified. We may need to copy into a bson_value_t* dst instead. */
+      return bson_iter_value (&iter);
```
jmikola, line 73: This is an outstanding question. The tests seem OK as-is, but I'm curious if we should address this.

Does `bson_iter_value` actually depend on the stack-allocated `bson_iter_t iter` or is it really depending on `cursor->opts`?

src/libmongoc/src/mongoc/mongoc-cursor.c
```diff
@@ -1755,6 +1771,28 @@ _mongoc_cursor_prepare_getmore_command (mongoc_cursor_t *cursor,
...
+      mongoc_server_stream_t *server_stream;
+
+      /* CRUD spec: If a comment is provided, drivers MUST attach this comment
+       * to all subsequent getMore commands run on the same cursor for server
+       * versions 4.4 and above. For server versions below 4.4 drivers MUST NOT
+       * attach a comment to getMore commands. */
+      server_stream = _mongoc_cursor_fetch_stream (cursor);
+
+      /* TODO: is no-op on server_stream sensible? This function has no error
+       * reporting so it's unclear can be done if we fail to fetch a stream. */
```
jmikola, line 1786: Outstanding question. There really isn't a viable way to report an error here, but I believe `_mongoc_cursor_fetch_stream` already sets some error state within the cursor if it fails internally.

Presumably, a stream would already have been fetched for the cursor to decide whether to use a a getMore command or opcode, so perhaps we need not worry about an error here. If so, should we consider using an assertion on `server_stream`?

src/libmongoc/tests/bsonutil/bson-match.c
```diff
@@ -498,7 +500,7 @@ bson_matcher_match (bson_matcher_t *matcher,
                                   expected_val,
                                   actual_val,
                                   path_child,
-                                  allow_extra,
+                                  (is_root && array_of_root_docs),
```
jmikola, line 503: Sanity check appreciated. I believe with this logic `array_of_root_docs` only gets inherited when we recurse from handling a root BSON array. And per the code above, the we ignore extra keys if the root is either a document or a child document of a root array.

src/libmongoc/tests/json/crud/unified/bulkWrite-deleteMany-hint-unacknowledged.json
```diff
@@ -0,0 +1,269 @@
+{
+  "description": "bulkWrite-deleteMany-hint-unacknowledged",
```
jmikola, line 2: These `hint-unacknowledged` tests pertain to [CDRIVER-4127](https://jira.mongodb.org/browse/CDRIVER-4127), which is sitting in the backlog.

In the past, I think "sync" commits only pulled in the subset of test files relevant to the DRIVERS ticket vs. updating _all_ files in a directory to the target specifications commit. That seems like an easy way for spec tests to fall through the cracks, so I think it'd be preferable to keep these here.

Since we have the ability to skip tests, I see no reason to start syncing _everything_ in a directory and adding necessary skips and TODO comments as needed (as I did in `runner.c` for change stream spec tests).

These tests already seem to pass, but whenever [CDRIVER-4127](https://jira.mongodb.org/browse/CDRIVER-4127) gets scheduled we can double-check and resolve it accordingly if there's truly no work to be done.

src/libmongoc/tests/json/crud/unified/find-allowdiskuse.json
```diff
@@ -32,7 +32,7 @@
   ],
   "tests": [
     {
-      "description": "Find does not send allowDiskuse when value is not specified",
+      "description": "Find does not send allowDiskUse when value is not specified",
```
jmikola, line 35: Also pulled in while syncing the entire `crud/unified` directory. I'm not sure why this wasn't included in [CDRIVER-4358](https://jira.mongodb.org/browse/CDRIVER-4358) (since resolved), but that would have been the relevant issue. Likely just an oversight, as it was only a description change (`aggregate-allowdiskuse.json` was the actual new test file there).

src/libmongoc/tests/unified/result.c
```diff
@@ -361,10 +363,13 @@ result_check (result_t *result,
...
-         test_set_error (error,
-                         "checking expectResult: %s",
-                         bson_val_to_json (expect_result));
+      if (!entity_map_match (em,
+                             expect_result,
+                             result->value,
+                             result->array_of_root_docs,
+                             error)) {
+         test_diagnostics_error_info ("checking expectResult: %s",
+                                      bson_val_to_json (expect_result));
```
jmikola, line 372: This ensures that the `entity_map_match` error with detailed info (e.g. path info) actually gets reported in the `test-libmongoc` output. Previously, it would have been discarded in favor of this less informative message.

src/libmongoc/tests/unified/result.c
```diff
@@ -527,7 +532,12 @@ result_check (result_t *result,
             goto done;
          }
 
-         if (!bson_match (error_expect_result, result->value, true, error)) {
+         /* TODO: expectError.expectResult is not used for cursor-bearing
+          * operations, so array_of_root_docs should always be false */
```
jmikola, line 536: Outstanding question. Perhaps we want to assert this?

src/libmongoc/tests/unified/runner.c
```diff
@@ -79,6 +79,16 @@ skipped_unified_test_t SKIPPED_TESTS[] = {
...
    /* libmongoc does not pin connections to cursors. It cannot force an error from waitQueueTimeoutMS by creating cursors in load balanced mode. */
    {"wait queue timeout errors include details about checked out connections", SKIP_ALL_TESTS},
+   /* CDRIVER-4277: Change streams support for user-facing PIT pre- and post-images */
+   {"change-streams-pre_and_post_images", SKIP_ALL_TESTS},
+   /* CDRIVER-3973, CDRIVER-4305, CDRIVER-4279, CDRIVER-4321: unified change stream tests */
+   {"change-streams", SKIP_ALL_TESTS},
+   {"change-streams-errors", SKIP_ALL_TESTS},
+   {"change-streams-resume-allowlist", SKIP_ALL_TESTS},
+   {"change-streams-resume-errorlabels", SKIP_ALL_TESTS},
+   /* CDRIVER-4309: Use the count command instead of collStats to implement estimatedDocumentCount */
```
jmikola, line 89: I originally included this change in the same PR but decided to pull it out. In any event, the code is ready to go (mostly just some deletions) so I'll plan to take this ticket on immediately after we merge this.

src/libmongoc/tests/unified/runner.c
```diff
@@ -1093,7 +1104,7 @@ test_check_event (test_t *test,
       bson_val_t *expected_val = bson_val_from_bson (expected_reply);
       bson_val_t *actual_val = bson_val_from_bson (actual->reply);
       if (!entity_map_match (
-             test->entity_map, expected_val, actual_val, true, error)) {
+             test->entity_map, expected_val, actual_val, false, error)) {
```
jmikola, line 1107: This technically fixes [CDRIVER-4329](https://jira.mongodb.org/browse/CDRIVER-4329), which I folded into [CDRIVER-4374](https://jira.mongodb.org/browse/CDRIVER-4374).

src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -3418,6 +3451,10 @@ mongoc_collection_find_and_modify_with_opts (
       bson_append_document (&parts.extra, "let", 3, &appended_opts.let);
    }
 
+   if (&appended_opts.comment.value_type != BSON_TYPE_EOD) {
```
kevinAlbs, line 3454:
```suggestion
   if (appended_opts.comment.value_type != BSON_TYPE_EOD) {
```

This may resolve errors reported on Windows builds.

kevinAlbs (commented):
This looks good. Left two suggested test fixes. Thank you for calling out questions with comments and the organized commits.

src/libmongoc/tests/bsonutil/bson-match.c
jmikola: Without `array_of_root_docs` coming from `bson_matcher_match`, we might consider _any_ array of BSON values to be an array of docs. Consider the case where we're matching an array of embedded documents returned by `distinct`. Those should _not_ allow extra fields within.

build/generate-opts.py
```diff
@@ -156,11 +156,18 @@ def __init__(self, items, **defaults):
     'help': 'A BSON document consisting of any number of parameter names, each followed by definitions of constants in the MQL Aggregate Expression language.'
 })
 
+comment_option = ('comment', {
+    'type': 'bson_value_t',
+    'convert': '_mongoc_convert_bson_value_t',
+    'help': 'Comment to attach to this command, which will appear in log messages, profiler output, and currentOp output.'
+})
```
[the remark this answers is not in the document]
jmikola: Defined separate tuples and added docs to `count_documents`, `estimated_document_count`, and `find`. The option docs for each method should now clarify server version support and type requirements (where applicable).

src/libmongoc/src/mongoc/mongoc-bulk-operation.c
```diff
@@ -973,6 +992,25 @@ mongoc_bulk_operation_set_bypass_document_validation (
 }
 
 
+void
+mongoc_bulk_operation_set_comment (mongoc_bulk_operation_t *bulk,
+                                   const bson_value_t *comment)
+{
+   BSON_ASSERT (bulk);
+   BSON_ASSERT (comment);
```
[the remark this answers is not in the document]
jmikola: Will add a separate commit to apply `BSON_ASSERT_PARAM` generally throughout this file.

src/libmongoc/src/mongoc/mongoc-bulk-operation.c
```diff
@@ -973,6 +992,25 @@ mongoc_bulk_operation_set_bypass_document_validation (
...
+   BSON_ASSERT (comment->value_type != BSON_TYPE_EOD);
+
+   /* This method cannot be called after appending operations, as the CRUD spec
+    * states the option should apply to all commands. Since commands are
+    * initialized as operations are added, allowing "comment" to be changed at
+    * any time could violate that contract. */
+   BSON_ASSERT (bulk->commands.len == 0);
+
+   bson_value_destroy (&bulk->comment);
+   bson_value_copy (comment, &bulk->comment);
```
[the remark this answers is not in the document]
jmikola: Ah, thanks for catching that. Totally missed it.

src/libmongoc/tests/unified/runner.c
jmikola: [CDRIVER-4309](https://jira.mongodb.org/browse/CDRIVER-4309) was merged in #984 and I've since rebased this PR on master.

src/libmongoc/src/mongoc/mongoc-cursor.c
jmikola: Function is now inlined. I debated adding a comment to consider adding `_mongoc_cursor_get_opt` down the line if needed, but there's no guarantee we'd remember that comment should another case come up.

src/libmongoc/src/mongoc/mongoc-cursor.c
jmikola: `_mongoc_cursor_prepare_getmore_command` is called from quiet a few places, some of which don't have any error reporting, so this sounds like a more ambitious change. Do you think it's worth tracking this todo in another ticket for future work?

jmikola: @kevinAlbs: I think this is the only outstanding comment thread.

kevinAlbs: >  Do you think it's worth tracking this todo in another ticket for future work?

IMO no. I think the likelihood it being addressed in the near future is low. If `_mongoc_cursor_fetch_stream` returns an error, I think it is likely the next call to `_mongoc_cursor_fetch_stream` will fail.

src/libmongoc/src/mongoc/mongoc-cursor.c
```diff
@@ -1734,6 +1734,8 @@ _mongoc_cursor_prepare_getmore_command (mongoc_cursor_t *cursor,
    const char *collection;
    int collection_len;
    int64_t batch_size;
+   bson_iter_t iter;
+   const bson_value_t *comment;
```
kevinAlbs, line 1738:
```suggestion
```

Remove unused declaration. That may fix the debug-compile-with-warnings and compile-tracing tasks.

kevinAlbs (approved):
LGTM with a warning fix.

src/libmongoc/src/mongoc/mongoc-opts.c
```diff
@@ -1563,6 +1652,15 @@ _mongoc_change_stream_opts_parse (
...
       }
+      else if (!strcmp (bson_iter_key (&iter), "comment")) {
+         if (!_mongoc_convert_bson_value_t (
+               client,
+               &iter,
+               &mongoc_change_stream_opts->comment,
+               error)) {
+            return false;
+         }
+      }
```
vector-of-bool, line 1663: No a comment on this PR in particular, but the repetition of this pattern over and over and over again looks like a good opportunity to bring out the blunt hammer of C preprocessor macros.

vector-of-bool (approved):
LGTM. Nothing sticks out, but the repetition in `mongoc-opts.c` makes me curious about refactoring that file in particular.

src/libmongoc/src/mongoc/mongoc-opts.c
jmikola: Oh, these files are all generated from templates in `generate-opts.py`. Did you mean that this pattern is used throughout libmongoc or just the generated code? If the latter, I don't think it's worth creating a macro since this files isn't intended to be manually modified anyway.

jmikola:
@kevinAlbs: Can you help me discern the actual build failure in [debug-compile-nosasl-winssl](https://spruce.mongodb.com/task/mongo_c_driver_windows_2015_debug_compile_nosasl_winssl_patch_99e9893286767737d2e0cb5d55c7906c3f768656_627db0e361837d34963b517d_22_05_13_01_14_13)? This is the only task whose build status is inconsistent with the base commit in https://spruce.mongodb.com/version/627db0e361837d34963b517d/tasks.

PR #983, opened 2022-05-05
CDRIVER-4365 support encryptedFields in collection create and drop

kevinAlbs:
# Summary

- Add `mongoc_auto_encryption_opts_set_encrypted_fields_map`
- Add `encryptedFields` required behavior to create and drop collection helpers.
- Run "*-cse" tasks on replica set for server 6.0+.
    - Add "majority" write concern to key vault operations in CSFLE tests.
- Add `_mongoc_iter_document_as_bson` utility to convert a `bson_iter_t` to `bson_t`.

# Background & Motivation

Please see https://github.com/mongodb/specifications/pull/1183/ for the specification changes.

FLE 2.0 requires a non-standalone server topology. Tests need to apply write concern majority on key vault operations, as noted in the [specification test README](https://github.com/mongodb/specifications/blob/c5067286cf05f741660c05e1fe5fc66144fef76d/source/client-side-encryption/tests/README.rst#prose-tests)

> Perform all applicable operations on key vault collections (e.g. inserting an example data key, or running a find command) with readConcern/writeConcern "majority".

That is because the key vault is queried with readConcern "majority". If keys are inserted without writeConcern "majority", some tests fail to find the key if it the insert is not majority committed before it is queried.

Because the name of FLE 2.0 is still [being decided](https://jira.mongodb.org/browse/WRITING-10860), the documentation for `mongoc_auto_encryption_opts_set_encrypted_fields_map` is brief.

### DIFF
--- a/build/evergreen_config_lib/tasks.py
+++ b/build/evergreen_config_lib/tasks.py
@@ -606,7 +606,7 @@ class IntegrationTask(MatrixTask):
             require(self.version == 'latest' or parse_version(self.version) >= parse_version("4.2"))
             if self.version == 'latest' or parse_version(self.version) >= parse_version("6.0"):
                 # FLE 2.0 Client-Side Encryption tasks on 6.0 require a non-standalone topology.
-                require(self.topology == 'server' or self.topology == 'replica_set')
+                require(self.topology in ('server', 'replica_set'))
             else:
                 require(self.topology == 'server')
             if self.sanitizer != "asan":
--- a/build/evergreen_config_lib/tasks.py
+++ b/build/evergreen_config_lib/tasks.py
@@ -604,7 +604,11 @@ class IntegrationTask(MatrixTask):
 
         if self.cse:
             require(self.version == 'latest' or parse_version(self.version) >= parse_version("4.2"))
-            require(self.topology == 'server')
+            if self.version == 'latest' or parse_version(self.version) >= parse_version("6.0"):
+                # FLE 2.0 Client-Side Encryption tasks on 6.0 require a non-standalone topology.
+                require(self.topology == 'server' or self.topology == 'replica_set')
+            else:
+                require(self.topology == 'server')
             if self.sanitizer != "asan":
                 # limit to SASL=AUTO to reduce redundant tasks.
                 require(self.sasl)
--- a/src/libmongoc/src/mongoc/mongoc-database.c
+++ b/src/libmongoc/src/mongoc/mongoc-database.c
@@ -853,9 +853,8 @@ mongoc_database_get_collection_names_with_opts (mongoc_database_t *database,
    return ret;
 }
 
-
-mongoc_collection_t *
-mongoc_database_create_collection (mongoc_database_t *database,
+static mongoc_collection_t *
+create_collection (mongoc_database_t *database,
                                    const char *name,
                                    const bson_t *opts,
                                    bson_error_t *error)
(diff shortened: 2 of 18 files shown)


build/evergreen_config_lib/tasks.py
```diff
@@ -604,7 +604,11 @@ def _check_allowed(self):
 
         if self.cse:
             require(self.version == 'latest' or parse_version(self.version) >= parse_version("4.2"))
-            require(self.topology == 'server')
+            if self.version == 'latest' or parse_version(self.version) >= parse_version("6.0"):
+                # FLE 2.0 Client-Side Encryption tasks on 6.0 require a non-standalone topology.
+                require(self.topology == 'server' or self.topology == 'replica_set')
```
vector-of-bool, line 609: Fun shorthand tip, Python lets you use `in` with an iterable on the right-hand side:

```py
require(self.topology in ('server', 'replica_set'))
```

kevinAlbs: Neat, done.

src/libmongoc/src/mongoc/mongoc-database.c
```diff
@@ -853,9 +853,8 @@ mongoc_database_get_collection_names_with_opts (mongoc_database_t *database,
...
 }
 
-
-mongoc_collection_t *
-mongoc_database_create_collection (mongoc_database_t *database,
+static mongoc_collection_t *
+create_collection (mongoc_database_t *database,
                                    const char *name,
                                    const bson_t *opts,
                                    bson_error_t *error)
```
vector-of-bool, line 860: Ensure you're running a formatting pass over the sources before merging, as later unrelated changes to the same file might re-format the previously-unformatted code and glutter the diff. I use VSCode "Format on save" to ensure I never forget and don't need to bother fiddling with whitespace. Although this often causes headaches when an unrelated change in the file will rewrite some other lines that hadn't been given a formatting pass.

We could also take advantage of `.gitblameignore` to do a formatting pass on all-the-things (and also maybe increase our column limit? 👀)

src/libmongoc/src/mongoc/mongoc-database.c
```diff
@@ -996,6 +995,293 @@ mongoc_database_create_collection (mongoc_database_t *database,
...
+   escName = _mongoc_get_encryptedField_state_collection (
+      encryptedFields, name, "esc", error);
+   if (!escName) {
+      goto fail;
+   }
+
+   escCollection = create_collection (database, escName, NULL /* opts */, error);
+   if (!escCollection) {
+      goto fail;
+   }
```
vector-of-bool, line 1066: This pair of repeated ops could be made into a separate function to reduce redundancy and help clarity, as then:

```c
bool okay = 
       (escCollection = _create_encField_state_collection(database, encryptedFields, name, "esc", error)
    && (eccCollection = _create_encField_state_collection(database, encryptedFields, name, "ecc", error)
    && (ecocCollection = _create_encField_state_collection(database, encryptedFields, name, "ecoc", error);
if (!okay) {
    // Failed to create one or more field collections
    goto fail;
}
```

src/libmongoc/src/mongoc/mongoc-database.c
```diff
@@ -996,6 +995,293 @@ mongoc_database_create_collection (mongoc_database_t *database,
...
+                                                    NULL /* reply */,
+                                                    error)) {
+         bson_destroy (create_indexes);
+         bson_free (index_name);
+         bson_destroy (keys);
+         goto fail;
+      }
+      bson_destroy (create_indexes);
+      bson_free (index_name);
+      bson_destroy (keys);
```
vector-of-bool, line 1137: These cleanups are the same. I would try:

```c
ok = mongoc_database_write_command_with_opts(...);
// [cleanup]
```

No `goto` needed since the return value is effectively stored in `ok`

src/libmongoc/src/mongoc/mongoc-database.c
```diff
@@ -996,6 +995,293 @@ mongoc_database_create_collection (mongoc_database_t *database,
...
+                                      const char *collName,
+                                      bson_t *encryptedFields,
+                                      bson_error_t *error)
+{
+   const bson_t *efMap = client->topology->encrypted_fields_map;
+
+   bson_init (encryptedFields);
+
+   if (bson_empty0 (efMap)) {
+      return true;
```
vector-of-bool, line 1169: I'd recommend comments on reasons for early returns and `goto fail`s. What conditions are being satisfied or violated? e.g. "Empty efMap will have no encrypted fields", "No efMap entry for this database+collection", or "The efMap entry should always be a document", to help a future reader quickly understand the purpose of a branch or jump.

src/libmongoc/src/mongoc/mongoc-database.c
```diff
@@ -996,6 +995,293 @@ mongoc_database_create_collection (mongoc_database_t *database,
...
+      bson_free (ns);
+      return true;
+   }
+
+   if (!_mongoc_iter_document_as_bson (&iter, encryptedFields, error)) {
+      bson_free (ns);
+      return false;
+   }
+
+   bson_free (ns);
```
vector-of-bool, line 1184: `ns` can be freed earlier (after the `init_find` branch).

vector-of-bool (approved):
There are a few stylistic nits, but LGTM.

PR #988, opened 2022-05-10
[CDRIVER-3380] Fixup JSON specials and decimal128 parsing

vector-of-bool:
Certain token sequences leave the bson-from-json parser unhappy. This addresses those.

### DIFF
--- a/src/libbson/src/bson/bson-decimal128.c
+++ b/src/libbson/src/bson/bson-decimal128.c
@@ -542,7 +542,7 @@ bson_decimal128_from_string_w_len (const char *string,     /* IN */
          continue;
       }
 
-      if (ndigits_stored < 34) {
+      if (ndigits_stored < BSON_DECIMAL128_MAX_DIGITS) {
          if (*str_read != '0' || found_nonzero) {
             if (!found_nonzero) {
                first_nonzero = ndigits_read;
@@ -634,7 +634,7 @@ bson_decimal128_from_string_w_len (const char *string,     /* IN */
       /* Shift exponent to significand and decrease */
       last_digit++;
 
-      if (last_digit - first_digit > BSON_DECIMAL128_MAX_DIGITS) {
+      if (last_digit - first_digit >= BSON_DECIMAL128_MAX_DIGITS) {
          /* The exponent is too great to shift into the significand. */
          if (significant_digits == 0) {
             /* Value is zero, we are allowed to clamp the exponent. */
--- a/src/libbson/src/bson/bson-json.c
+++ b/src/libbson/src/bson/bson-json.c
@@ -1186,7 +1186,6 @@ _bson_json_read_start_map (bson_json_reader_t *reader) /* IN */
           * expected a legacy Binary format. now we see the second "{", so
           * backtrack and parse $type query operator. */
          bson->read_state = BSON_JSON_IN_START_MAP;
-         bson->read_state = BSON_JSON_IN_START_MAP;
          STACK_PUSH_DOC (bson_append_document_begin (
             STACK_BSON_PARENT, key, len, STACK_BSON_CHILD));
          _bson_json_save_map_key (bson, (const uint8_t *) "$type", 5);
@@ -1200,13 +1199,21 @@ _bson_json_read_start_map (bson_json_reader_t *reader) /* IN */
       case BSON_JSON_LF_MINKEY:
       case BSON_JSON_LF_OID:
       case BSON_JSON_LF_OPTIONS:
-      case BSON_JSON_LF_REGEX: // $regex lands here, but $regularExpression does
-                               // not
+      case BSON_JSON_LF_REGEX:
+      /**
+       * NOTE: A read_state of BSON_JSON_IN_BSON_TYPE is used when "$regex" is
+       * found, but BSON_JSON_IN_BSON_TYPE_REGEX_STARTMAP is used for
+       * "$regularExpression", which will instead go to a below 'if else' branch
+       * instead of this switch statement. They're both called "regex" in their
+       * respective enumerators, but they behave differently when parsing.
+       */
+      // fallthrough
       case BSON_JSON_LF_REGULAR_EXPRESSION_OPTIONS:
       case BSON_JSON_LF_REGULAR_EXPRESSION_PATTERN:
       case BSON_JSON_LF_SYMBOL:
       case BSON_JSON_LF_UNDEFINED:
       case BSON_JSON_LF_UUID:
+         // These special keys do not expect objects as their values. Fail.
          _bson_json_read_set_error (
             reader,
             "Unexpected nested object value for \"%s\" key",
@@ -1216,6 +1223,7 @@ _bson_json_read_start_map (bson_json_reader_t *reader) /* IN */
       case BSON_JSON_LF_SCOPE:
       case BSON_JSON_LF_TIMESTAMP_I:
       case BSON_JSON_LF_TIMESTAMP_T:
+         // These special LF keys aren't handled with BSON_JSON_IN_BSON_TYPE
          BSON_UNREACHABLE (
             "These LF values are handled with a different read_state");
       }
--- a/src/libbson/src/bson/bson-json.c
+++ b/src/libbson/src/bson/bson-json.c
@@ -1174,18 +1174,50 @@ _bson_json_read_start_map (bson_json_reader_t *reader) /* IN */
    BASIC_CB_PREAMBLE;
 
    if (bson->read_state == BSON_JSON_IN_BSON_TYPE) {
-      if (bson->bson_state == BSON_JSON_LF_DATE) {
+      switch (bson->bson_state) {
+      case BSON_JSON_LF_DATE:
          bson->read_state = BSON_JSON_IN_BSON_TYPE_DATE_NUMBERLONG;
-      } else if (bson->bson_state == BSON_JSON_LF_BINARY) {
+         break;
+      case BSON_JSON_LF_BINARY:
          bson->read_state = BSON_JSON_IN_BSON_TYPE_BINARY_VALUES;
-      } else if (bson->bson_state == BSON_JSON_LF_TYPE) {
+         break;
+      case BSON_JSON_LF_TYPE:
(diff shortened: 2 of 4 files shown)


src/libbson/src/bson/bson-decimal128.c
```diff
@@ -542,7 +542,7 @@ bson_decimal128_from_string_w_len (const char *string,     /* IN */
          continue;
       }
 
-      if (ndigits_stored < 34) {
+      if (ndigits_stored < BSON_DECIMAL128_MAX_DIGITS) {
```
vector-of-bool, line 545: No functional change, just replace a magic number.

src/libbson/src/bson/bson-decimal128.c
```diff
@@ -634,7 +634,7 @@ bson_decimal128_from_string_w_len (const char *string,     /* IN */
       /* Shift exponent to significand and decrease */
       last_digit++;
 
-      if (last_digit - first_digit > BSON_DECIMAL128_MAX_DIGITS) {
+      if (last_digit - first_digit >= BSON_DECIMAL128_MAX_DIGITS) {
```
vector-of-bool, line 637: If `last_digit - first_digit` is exactly `34` (`BSON_DECIMAL128_MAX_DIGITS`), we want to treat that as out-of-bounds as well.

src/libbson/src/bson/bson-json.c
```diff
@@ -1174,18 +1174,50 @@ _bson_json_read_start_map (bson_json_reader_t *reader) /* IN */
...
+      case BSON_JSON_LF_REGULAR_EXPRESSION_OPTIONS:
+      case BSON_JSON_LF_REGULAR_EXPRESSION_PATTERN:
+      case BSON_JSON_LF_SYMBOL:
+      case BSON_JSON_LF_UNDEFINED:
+      case BSON_JSON_LF_UUID:
+         _bson_json_read_set_error (
+            reader,
+            "Unexpected nested object value for \"%s\" key",
+            reader->bson.unescaped.buf);
+         break;
```
vector-of-bool, line 1214: In all of these cases, we are parsing a special `"$dollar"` object key that *does not* want a nested sub-object value. Previously, this would fall through and it would later treat the closing brace `}` of the sub-object as the closing of the entire special object itself. This fixes [CDRIVER-3380](https://jira.mongodb.org/browse/CDRIVER-3380).

src/libbson/src/jsonsl/jsonsl.c
```diff
@@ -457,13 +457,17 @@ jsonsl_feed(jsonsl_t jsn, const jsonsl_char_t *bytes, size_t nbytes)
...
                       state->special_flags &= ~JSONSL_SPECIALf_NULL;
                    }
-
-                   if (tolower(CUR_CHAR) != 'a') {
+                   if (not_a) {
                       state->special_flags &= ~JSONSL_SPECIALf_NAN;
                    }
+                   if (not_u && not_a) {
+                      VERIFY_SPECIAL ("null", 4);
+                   }
```
vector-of-bool, line 470: This fixes a bug in `NaN`/`null` parsing in `jsonsl`. Previously: Any byte with value `'n'` would then be expected to be followed by a `u` or an `A`/`a` character, but this check did not error if `'n'` was followed by anything else, and the contiguous sequence of characters would be thrown away and ignored. That is, the string `[nope]` would parse as an empty array, because the `nope` would be erroneously ignored. This had fun effects as in `{"foo": nope}`, where we generate a key with no associated value. Woops.

src/libbson/tests/test-json.c
```diff
@@ -2727,6 +2727,7 @@ test_bson_json_errors (void)
        "Invalid read of boolean in state IN_BSON_TYPE"},
       {"{\"x\": {\"$dbPointer\": true}}",
        "Invalid read of boolean in state IN_BSON_TYPE_DBPOINTER_STARTMAP"},
+      {"[{\"$code\": {}}]", "Unexpected nested object value for \"$code\" key"},
```
vector-of-bool, line 2730: A test case to reject `{"$code": {}}`, which does not accept an object value. (Tests [CDRIVER-3380](https://jira.mongodb.org/browse/CDRIVER-3380))

src/libbson/src/bson/bson-json.c
```diff
@@ -1174,18 +1174,50 @@ _bson_json_read_start_map (bson_json_reader_t *reader) /* IN */
...
+      case BSON_JSON_LF_BINARY:
          bson->read_state = BSON_JSON_IN_BSON_TYPE_BINARY_VALUES;
-      } else if (bson->bson_state == BSON_JSON_LF_TYPE) {
+         break;
+      case BSON_JSON_LF_TYPE:
          /* special case, we started parsing {$type: {$numberInt: "2"}} and we
           * expected a legacy Binary format. now we see the second "{", so
           * backtrack and parse $type query operator. */
          bson->read_state = BSON_JSON_IN_START_MAP;
+         bson->read_state = BSON_JSON_IN_START_MAP;
```
eramongodb, line 1189:
```suggestion
```
Accidental line duplication?

src/libbson/src/bson/bson-json.c
```diff
@@ -1174,18 +1174,50 @@ _bson_json_read_start_map (bson_json_reader_t *reader) /* IN */
...
+      case BSON_JSON_LF_DECIMAL128:
+      case BSON_JSON_LF_DOUBLE:
+      case BSON_JSON_LF_INT32:
+      case BSON_JSON_LF_INT64:
+      case BSON_JSON_LF_MAXKEY:
+      case BSON_JSON_LF_MINKEY:
+      case BSON_JSON_LF_OID:
+      case BSON_JSON_LF_OPTIONS:
+      case BSON_JSON_LF_REGEX: // $regex lands here, but $regularExpression does
+                               // not
```
eramongodb, line 1204: What is the significance of this comment?

src/libbson/src/jsonsl/jsonsl.c
```diff
@@ -457,13 +457,17 @@ jsonsl_feed(jsonsl_t jsn, const jsonsl_char_t *bytes, size_t nbytes)
                 } else if (state->special_flags & JSONSL_SPECIALf_NULL ||
                            state->special_flags & JSONSL_SPECIALf_NAN) {
                    /* previous char was "n", are we parsing null or nan? */
-                   if (CUR_CHAR != 'u') {
+                   bool not_u = CUR_CHAR != 'u';
+                   bool not_a = tolower (CUR_CHAR) != 'a';
```
eramongodb, line 461:
```suggestion
                   const bool not_u = CUR_CHAR != 'u';
                   const bool not_a = tolower (CUR_CHAR) != 'a';
```

eramongodb, line 461: Minor suggestion: rename and redefine `not_u` and `not_a` as `is_u` and `is_a` respectively to avoid possibility of double-negatives in conditions (e.g. `!not_u`).

src/libbson/src/jsonsl/jsonsl.c
eramongodb, line 470: Confirming: is `not_u` really supposed to be in this conditional? Given we are verifying the value to be `"null"`, the check being conditioned on `not_u` seems contradictory at first glance.

src/libbson/src/bson/bson-json.c
vector-of-bool: This is to clear up confusion for future reader, as I was initially confused by this. `read_state` of `BSON_JSON_IN_BSON_TYPE` is used when `$regex` is found, but `BSON_JSON_IN_BSON_TYPE_REGEX_STARTMAP` is used for `$regularExpression`, which will instead go to below on line ~1228. They're both called "regex" in their respective enumerators, but they behave differently when parsing.

src/libbson/src/jsonsl/jsonsl.c
vector-of-bool: I forgot to put a comment explaining this. The `VERIFY_SPECIAL` here will unconditionally generate an error about an invalid "special" token. This is to halt JSON tokenization if a bare `n` is followed by anything other than a `u` or an `a`. The choice of `"null"` as the string here is just to match what the user more likely meant to type (as NaN is an extension in jsonsl).

src/libbson/src/jsonsl/jsonsl.c
vector-of-bool: I used the negation here to maintain the intent of the original code, which is "if the char *is not* `c`, unset the bit that corresponds to `c`". Inverting that would be "if the char *is* `c`, do not unset the bit that corresponds to `c`", which also contains a double negative "do not unset". Storing them as variables is just to dedupe the later `&&` check.

src/libbson/src/jsonsl/jsonsl.c
```diff
@@ -457,13 +457,23 @@ jsonsl_feed(jsonsl_t jsn, const jsonsl_char_t *bytes, size_t nbytes)
...
+                   if (not_u) {
                       state->special_flags &= ~JSONSL_SPECIALf_NULL;
                    }
-
-                   if (tolower(CUR_CHAR) != 'a') {
+                   if (not_a) {
                       state->special_flags &= ~JSONSL_SPECIALf_NAN;
                    }
+                   if (not_u && not_a) {
+                      /* This veryify will always fail, as we have an 'n'
```
eramongodb, line 469:
```suggestion
                      /* This verify will always fail, as we have an 'n'
```

eramongodb (approved):

src/libbson/src/jsonsl/jsonsl.c
kevinAlbs: Great catch. This may be worth submitting a PR to jsonsl.

kevinAlbs (approved):
LGTM

src/libbson/src/jsonsl/jsonsl.c
vector-of-bool: I considered that, but I also discovered Roberto's existing open PR that has been sitting open for a good while 🙂.

PR #989, opened 2022-05-10
CDRIVER-4369 Add FLE 2 API to AutoEncryptionOpts

kevinAlbs:
# Summary

- Do not return error on "ns not found" error in drop with encryptedFields.
- Pass `encrypted_fields_map` to libmongocrypt.
- Add `add mongoc_auto_encryption_opts_set_bypass_query_analysis`.
- Require libmongocrypt version 1.5.0.
- Update csfle test dependency to 6.0.0-rc4.
- Do not spawn mongocryptd if `bypass_query_analysis` is true.

# Background & Motivation

The "ns not found" error is returned by the "drop" command if the collection does not exist. All drivers excluding Node and C ignore this error in their drop helpers. See [Do drivers ignore "ns not found" in collection drop?](https://docs.google.com/spreadsheets/d/1TVgZTynQTmoVsmPdSxOAmGzg8OvVSCVZEKig34ZaxAg/edit#gid=0). Ignoring the "ns not found" error in the drop operation with `encryptedFields` is consistent with [Node's implementation](https://github.com/mongodb/node-mongodb-native/blob/c12fef3bf7c51d6d8023ffd3dcda687b4ee250d3/src/operations/drop.ts#L62)..

Using `csfle` version 5.3.1 resulted in the following error:
> Expected success, got error "csfle "analyze_query" failed: jsonSchema is a required command field [Error 2, code 51073]"

That error suggests 5.3.1 does not have all FLE 2 features. `jsonSchema` is only required for FLE 1. Updating to 6.0.0-rc4 resolved this error.


vector-of-bool (approved):
LGTM

PR #992, opened 2022-05-10
Update mongoc-socket.c

MonkeybreadSoftware:
added #pragma, so the Microsoft C Compiler can add the Advapi32.lib as import library and the library compiles here for Windows.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-socket.c
+++ b/src/libmongoc/src/mongoc/mongoc-socket.c
@@ -27,6 +27,11 @@
 #ifdef _WIN32
 #include <Mstcpip.h>
 #include <process.h>
+
+#if (defined (_MSCVER)
+#pragma comment(lib, "Advapi32.lib")
+#endif
+
 #endif
 
 #undef MONGOC_LOG_DOMAIN


MonkeybreadSoftware:
The screenshot from the original problem I had when building this with VS 2019 and cmake in Terminal:
Bildschirmfoto 2022-05-10 um 22 48 10

So registry functions are used in Advapi32, but the lib is not in the list to link. The #pragma is a convenient way to fix this.

vector-of-bool:
Thank you for your contribution!

We're hesitant to use `#pragma comment(linker)` as it usurps control of linking from the build system. It would be best to fix this within the build system itself. Before merging any changes, I'd like to reproduce the issue locally to discover under what conditions this link error occurs, as this isn't something we see on our own systems. Base on my reading of our current CMake project, this might occur if you are linking to `mongoc_static` but does not occur when linking to `mongoc_shared`, and there are missing library link directives for the static-linking case.

MonkeybreadSoftware:
Well, just build with command line and Visual Studio 2019 for Windows ARM 64-bit target.
Maybe the cmake files somewhere only specify to include the advapi32.lib when using Windows with Intel?

kevinAlbs:
@MonkeybreadSoftware is this issue still relevant? We have not been able to reproduce and may close this PR.

MonkeybreadSoftware:
Well, we added the #pragma and then it worked for us.
If nobody cares, close it.

PR #993, opened 2022-05-11
CDRIVER-4389 Update FLE 2 Create Collection

kevinAlbs:
# Summary

- Add clusteredIndex option to state collections.
- Drop the data collection after state collections.

# Background & Motivation 
See DRIVERS-2326 for the corresponding spec change.


vector-of-bool (approved):
LGTM

PR #996, opened 2022-05-16
Allow for building on Alpine Linux systems

sigmaseven:
This change allows for mongo-c-driver to be built on Alpine Linux-based systems.

### DIFF
--- a/build/mongodl.py
+++ b/build/mongodl.py
@@ -96,7 +96,7 @@ def _infer_target_os_rel():
     assert mat, 'Unable to detect ID from [/etc/os-release] content:\n{}'.format(
         content)
     os_id = mat.group(2)
-    if os_id == 'arch':
+    if os_id == 'arch' or os_id == 'alpine':
         # There are no Archlinux-specific MongoDB downloads, so we'll just use
         # the build for RHEL8, which is reasonably compatible with other modern
         # distributions (including Arch).


vector-of-bool:
Any errors from `mongodl.py` would be non-fatal errors, as the script is only applicable for integration testing purposes, additionally the MongoDB binaries that would be downloaded by this changeset would be incompatible with Alpine as the they would require GNU glibc rather than musl. What errors are you seeing that prevent the code from compiling and running on Alpine?

sigmaseven:
The error this specifically deals with was the following:

`[ 94%] Linking C executable mongoc-stat
[ 94%] Built target mongoc-stat
[ 94%] Downloading csfle
Downloading [https://downloads.mongodb.org/full.json] ...
Refreshing downloads manifest ...
  File "/root/mongo-c-driver/build/mongodl.py", line 700, in <module>
    sys.exit(main())
  File "/root/mongo-c-driver/build/mongodl.py", line 679, in main
    target = args.target or infer_target()
  File "/root/mongo-c-driver/build/mongodl.py", line 86, in infer_target
    return _infer_target_os_rel()
  File "/root/mongo-c-driver/build/mongodl.py", line 121, in _infer_target_os_rel
    raise RuntimeError("We don't know how to map '{}' to a distribution "
RuntimeError: We don't know how to map 'alpine' to a distribution download target. Please contribute!
make[2]: *** [src/libmongoc/CMakeFiles/get-csfle.dir/build.make:73: src/libmongoc/mongo_csfle_v1.so] Error 1
make[1]: *** [CMakeFiles/Makefile2:2198: src/libmongoc/CMakeFiles/get-csfle.dir/all] Error 2
make: *** [Makefile:166: all] Error 2`

I'm simply adding a 'mapping' that doesn't download any additional binaries so that it actually goes through the entire build process. If you'd prefer to take a crack at it I'm all for it, but it's kind of silly to hold up a build because a string isn't in a python script.

PR #998, opened 2022-05-17
CDRIVER-4376 note that Visual Studio 2013 is required

kevinAlbs:
The C driver has not tested on Visual Studio < 2013 or since Visual Studio versions older than 2013 were removed from Evergreen (refer: CDRIVER-3593).

There is an incorrect note in [the public documentation](http://mongoc.org/libmongoc/current/installing.html#build-environment-on-windows-with-visual-studio) claiming to support back to 2010.


rcsanchez97 (approved):
LGTM

PR #999, opened 2022-05-18
Fix SANITIZE arg shell quoting

vector-of-bool:
The current build is not building with sanitizers enabled. The escaped quotes around `MONGO_SANITIZE` are passed through literally, and CMake just discards the argument and doesn't mention it. The quotes are not required as the argument should be shell-safe.


rcsanchez97 (approved):
LGTM

PR #1003, opened 2022-05-19
CDRIVER-4380 Update maxWireVersion for 6.0

vector-of-bool:
Addresses CDRIVER-4380 

I've added a set of tasks for `6.0` integration testing in this PR. Also we seem to have jumped over wire version 16, which I believe corresponds to server version 5.3 based on digging in the code.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -2202,6 +2202,144 @@ tasks:
       AUTH: noauth
       SSL: nossl
       VALGRIND: 'on'
+- name: test-valgrind-6.0-server-auth-nosasl-openssl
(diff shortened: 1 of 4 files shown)


kevinAlbs:
> I've added a set of tasks for `6.0` integration testing in this PR. Also we seem to have jumped over wire version 16, which I believe corresponds to server version 5.3 based on digging in the code.

https://github.com/mongodb/specifications/blob/master/source/wireversion-featurelist.rst also is a quick reference. Though it needs to be updated for 6.0.

kevinAlbs (approved):

PR #1011, opened 2022-05-24
CDRIVER-4321: Support a "rename" operation for collections

vector-of-bool:
This adds support for the "rename" operation in the unified test format
this will rename the collection in the server as well as the local
copy in the entity map for later operations.

This changeset does not include test updates, but was manually validated
locally against the updated changestreams tests, which cannot be enabled
without additional changes that are beyond the scope of this PR.

### DIFF
--- a/src/libmongoc/tests/unified/operation.c
+++ b/src/libmongoc/tests/unified/operation.c
@@ -2799,9 +2799,10 @@ operation_rename (test_t *test,
       goto done;
    }
    result_from_ok (result);
+   ret = true;
 done:
    bson_free (new_name);
-   return true;
+   return ret;
 }
 
 typedef struct {
--- a/src/libmongoc/tests/unified/operation.c
+++ b/src/libmongoc/tests/unified/operation.c
@@ -2800,6 +2800,7 @@ operation_rename (test_t *test,
    if (!mongoc_collection_rename (coll, coll->db, new_name, false, error)) {
       return false;
    }
+   result_from_ok (result);
    return true;
 }
 
--- a/src/libmongoc/tests/unified/operation.c
+++ b/src/libmongoc/tests/unified/operation.c
@@ -2770,18 +2770,19 @@ operation_rename (test_t *test,
    // First validate the arguments
    const char *object = op->object;
    bson_parser_t *bp = bson_parser_new ();
-   char *new_name;
+   bool ret = false;
+   char *new_name = NULL;
    bson_parser_utf8 (bp, "to", &new_name);
    bool parse_ok = bson_parser_parse (bp, op->arguments, error);
    bson_parser_destroy (bp);
    if (!parse_ok) {
-      return false;
+      goto done;
    }
 
    // Now get the entity
    entity_t *ent = entity_map_get (test->entity_map, object, error);
    if (!ent) {
-      return false;
+      goto done;
    }
    // We only support collections so far
    if (0 != strcmp (ent->type, "collection")) {
@@ -2790,14 +2791,16 @@ operation_rename (test_t *test,
          "'rename' is only supported for collection objects '%s' has type '%s'",
          object,
          ent->type);
-      return false;
+      goto done;
    }
    // Rename the collection in the server,
    mongoc_collection_t *coll = ent->value;
    if (!mongoc_collection_rename (coll, coll->db, new_name, false, error)) {
-      return false;
+      goto done;
    }
    result_from_ok (result);
+done:
+   bson_free (new_name);
    return true;
 }
 
--- a/src/libmongoc/tests/unified/operation.c
+++ b/src/libmongoc/tests/unified/operation.c
@@ -2761,6 +2761,48 @@ operation_assert_number_connections_checked_out (test_t *test,
    return true;
 }
 
+static bool
+operation_rename (test_t *test,
+                  operation_t *op,
+                  result_t *result,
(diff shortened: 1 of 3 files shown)


kevinAlbs (commented):
I think this may not be correct or necessary. The entity map maps from ID to collection object. The ID does not need to match collection name. I think "rename" operation is only required to rename the collection name on the server.

See [PHP](https://github.com/mongodb/mongo-php-library/blob/62ffb4dd52db82da23faaf4138577979dd0c41d1/tests/UnifiedSpecTests/Operation.php#L490-L494) and [Go](https://github.com/mongodb/mongo-go-driver/blob/aa32249b4cdbd7d5c3e6a851b376b061849efbf8/mongo/integration/unified/collection_operation_execution.go#L879) implementations for reference.

vector-of-bool:
I was unsure of that. I'll investigate.

vector-of-bool:
@kevinAlbs The original changeset was missing the `"rename"` operation code entirely, so it wasn't even correct. I've removed the entity rename coe, as we won't need it.

src/libmongoc/tests/unified/operation.c
```diff
@@ -2761,6 +2761,48 @@ operation_assert_number_connections_checked_out (test_t *test,
...
+static bool
+operation_rename (test_t *test,
+                  operation_t *op,
+                  result_t *result,
+                  bson_error_t *error)
+{
+   // First validate the arguments
+   const char *object = op->object;
+   bson_iter_t it;
+   bool okay = bson_iter_init_find (&it, op->arguments, "to");
```
kevinAlbs, line 2773: I suggest parsing the arguments with the `bson_parser_t`. Example:

```c
   bson_parser_t *bp;
   char* to = NULL;

   bp = bson_parser_new ();
   bson_parser_utf8 (bp, "to", &to);
   if (!bson_parser_parse (bp, op->arguments, error)) {
      goto done;
   }

   /* other code */
done:
   bson_parser_destroy_with_parsed_fields (bp);
```

It will error on extra unrecognized fields by default. That may make a clearer error if we sync tests that add arguments to the "rename" operation.

src/libmongoc/tests/unified/operation.c
```diff
@@ -2761,6 +2761,49 @@ operation_assert_number_connections_checked_out (test_t *test,
...
+      test_set_error (error, "'rename' operation is missing a 'to' argument");
+      return false;
+   }
+   if (!BSON_ITER_HOLDS_UTF8 (&it)) {
+      test_set_error (error, "'rename' argument is not a utf8 string");
+      return false;
+   }
+
+   // Now get the entity
+   entity_t *ent = entity_map_get (test->entity_map, object, error);
```
kevinAlbs, line 2784: Variable declarations may need to be hoisted to satisfy Visual Studio 2013 builds.

vector-of-bool: VS2013 supports intermixed variable declarations

kevinAlbs: Oh, nice. TIL.

src/libmongoc/tests/unified/operation.c
```diff
@@ -2761,6 +2761,46 @@ operation_assert_number_connections_checked_out (test_t *test,
...
+                  operation_t *op,
+                  result_t *result,
+                  bson_error_t *error)
+{
+   // First validate the arguments
+   const char *object = op->object;
+   bson_parser_t *bp = bson_parser_new ();
+   char *new_name;
+   bson_parser_utf8 (bp, "to", &new_name);
+   bool parse_ok = bson_parser_parse (bp, op->arguments, error);
```
kevinAlbs, line 2775: The parser copies the fields it parses. `new_name` will need to be freed with `bson_free (new_name);`. Or, the parser can be freed with `bson_parser_destroy_with_parsed_fields`.

kevinAlbs (approved):
LGTM with a free of `new_name`.

src/libmongoc/tests/unified/operation.c
```diff
@@ -2761,6 +2761,50 @@ operation_assert_number_connections_checked_out (test_t *test,
...
+   // Now get the entity
+   entity_t *ent = entity_map_get (test->entity_map, object, error);
+   if (!ent) {
+      goto done;
+   }
+   // We only support collections so far
+   if (0 != strcmp (ent->type, "collection")) {
+      test_set_error (
+         error,
+         "'rename' is only supported for collection objects '%s' has type '%s'",
```
eramongodb, line 2791: Suggest rewording this error message to:
```
"'rename' is only supported for collection objects, but entity '%s' is a '%s'"
```

src/libmongoc/tests/unified/operation.c
```diff
@@ -2761,6 +2761,50 @@ operation_assert_number_connections_checked_out (test_t *test,
...
+   }
+   // Rename the collection in the server,
+   mongoc_collection_t *coll = ent->value;
+   if (!mongoc_collection_rename (coll, coll->db, new_name, false, error)) {
+      goto done;
+   }
+   result_from_ok (result);
+   ret = true;
+done:
+   bson_free (new_name);
```
eramongodb, line 2804: Given all parsed fields are destroyed at end of this function, I would suggest using `bson_parser_destroy_with_parsed_fields` instead of manually destroying the fields and parser individually.

PR #1014, opened 2022-05-25
CDRIVER-4199: Add comment to change stream getMore commands

vector-of-bool:
This adds a missing "comment" value to getMore for changestreams. This
fixes test failures in the changestreams tests, which are not yet
enabled in this branch but has been manually validated locally.


jmikola (approved):

jmikola:
Note: I renamed the PR, as I think this should be associated with CDRIVER-4199 (unless you had another ticket for it).

PR #1015, opened 2022-05-25
CDRIVER-3973 + CDRIVER-4366 : Update and use change-streams tests

vector-of-bool:
This changeset addresses both CDRIVER-3973 and CDRIVER-4366. If pulls the updated tests for 4366 while also finishing support for the test cases required to complete 3973. This PR includes changes from #1011 and #1014: Check those PRs to see those changes in isolation.

PR #1016, opened 2022-05-26
CDRIVER-4277: Pre-point-in-time support

vector-of-bool:
This change implements [DRIVERS-2189](https://jira.mongodb.org/browse/DRIVERS-2189). This also includes a generate-opts change that was missed as part of #1010. This changeset builds upon #1015, which enables and fixes the change-streams unified tests.


vector-of-bool:
Just noticed that this also includes spec test updates for CDRIVER-4368, since I pulled the JSON files from HEAD.

kevinAlbs (approved):

jmikola (approved):

PR #1021, opened 2022-05-31
Sync collection-management unified tests

eramongodb:
### Description

This PR resolves CDRIVER-4359, CDRIVER-4368, and CDRIVER-4392.

### Implementation

No changes were required to support the new `changeStreamPreAndPostImages` or `clusteredIndex` options, as options are specified as fields of a `bson_t` document that is forwarded directly to the server. Supporting the new options simply constituted adding a `clusteredIndex` field to the options document. Similarly, no changes were required to support the new `clustered` field in the result of `listIndexes`, as it is simply forwarding the result from the server as a `bson_t`.

### Unified Test Runner

The only major change required was adding support for the new `modifyCollection` operation.

However, the new unified tests exposed two bugs in the unified test runner implementation that are addressed in this PR.

- The `error` parameter may not be initialized or assigned to before being tested during result checks after the first test is executed, leading to undefined behavior manifest as a test failure due to an unexpected error with junk error message.
- The implementation of the `listIndexes` operation assumed one or more arguments, which is no longer the case with the new unified tests. A simple check was added to guard the call to `bson_concat()` to permit zero arguments.

### DIFF
--- a/src/libmongoc/tests/json/collection-management/clustered-indexes.json
+++ b/src/libmongoc/tests/json/collection-management/clustered-indexes.json
@@ -10,14 +10,17 @@
   "createEntities": [
     {
       "client": {
-        "id": "client0"
+        "id": "client0",
+        "observeEvents": [
+          "commandStartedEvent"
+        ]
       }
     },
     {
       "database": {
         "id": "database0",
         "client": "client0",
-        "databaseName": "ts-tests"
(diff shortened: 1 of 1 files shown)


vector-of-bool (approved):
LGTM.

kevinAlbs (approved):
Nice, LGTM

PR #1024, opened 2022-06-03
CDRIVER-4401 decrypt after emitting events

kevinAlbs:
# Summary

- Decrypt after emitting events

# Background & Motivation

Please see CDRIVER-4401 for more details.

Logic try to decrypt an error reply was considered. I do not think an error reply can contain ciphertext to decrypt.

Adding specification tests was considered. The legacy CSE test format, based on the legacy Transactions test format, [only supports assertions on command started events](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#test-format):

> expectations: Optional list of command-started events.

### DIFF
--- a/.evergreen/compile-unix.sh
+++ b/.evergreen/compile-unix.sh
@@ -228,7 +228,7 @@ pkg-config --modversion libssl || true
 if [ "$COMPILE_LIBMONGOCRYPT" = "ON" ]; then
    # Build libmongocrypt, using the previously fetched installed source.
    # TODO (CDRIVER-4397): add "--branch 1.5.0-rc0" in git clone once libmongocrypt 1.5.0-rc0 is released.
-   git clone https://github.com/mongodb/libmongocrypt
+   git clone https://github.com/kevinAlbs/libmongocrypt --branch fix-leak
    mkdir libmongocrypt/cmake-build
    cd libmongocrypt/cmake-build
    $CMAKE -DENABLE_SHARED_BSON=ON -DCMAKE_BUILD_TYPE="Debug" -DCMAKE_INSTALL_PREFIX="$INSTALL_DIR" -DCMAKE_PREFIX_PATH="$CMAKE_PREFIX_PATH" ../
--- a/.evergreen/compile-unix.sh
+++ b/.evergreen/compile-unix.sh
@@ -228,7 +228,7 @@ pkg-config --modversion libssl || true
 if [ "$COMPILE_LIBMONGOCRYPT" = "ON" ]; then
    # Build libmongocrypt, using the previously fetched installed source.
    # TODO (CDRIVER-4397): add "--branch 1.5.0-rc0" in git clone once libmongocrypt 1.5.0-rc0 is released.
-   git clone https://github.com/kevinAlbs/libmongocrypt --branch fix-leak
+   git clone https://github.com/mongodb/libmongocrypt
    mkdir libmongocrypt/cmake-build
    cd libmongocrypt/cmake-build
    $CMAKE -DENABLE_SHARED_BSON=ON -DCMAKE_BUILD_TYPE="Debug" -DCMAKE_INSTALL_PREFIX="$INSTALL_DIR" -DCMAKE_PREFIX_PATH="$CMAKE_PREFIX_PATH" ../
--- a/src/libmongoc/src/mongoc/mongoc-cluster.c
+++ b/src/libmongoc/src/mongoc/mongoc-cluster.c
@@ -553,18 +553,6 @@ mongoc_cluster_run_command_monitored (mongoc_cluster_t *cluster,
          cluster, cmd, compressor_id, reply, error);
    }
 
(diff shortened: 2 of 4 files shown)


.evergreen/compile-unix.sh
```diff
@@ -228,7 +228,7 @@ pkg-config --modversion libssl || true
 if [ "$COMPILE_LIBMONGOCRYPT" = "ON" ]; then
    # Build libmongocrypt, using the previously fetched installed source.
    # TODO (CDRIVER-4397): add "--branch 1.5.0-rc0" in git clone once libmongocrypt 1.5.0-rc0 is released.
-   git clone https://github.com/mongodb/libmongocrypt
+   git clone https://github.com/kevinAlbs/libmongocrypt --branch fix-leak
```
kevinAlbs, line 231: TODO: revert this change and the change in compile-windows.sh once https://github.com/mongodb/libmongocrypt/pull/357 is merged.

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -612,6 +600,18 @@ mongoc_cluster_run_command_monitored (mongoc_cluster_t *cluster,
       mongoc_apm_command_failed_cleanup (&failed_event);
    }
 
+   if (retval && _mongoc_cse_is_enabled (cluster->client)) {
```
jmikola, line 603: If I understand correctly, the two changes here are:

 * Moving `_mongoc_cse_auto_decrypt` until _after_ all APM callbacks have been dispatched
 * Only decrypting if the original command was successful

> Logic try to decrypt an error reply was considered. I do not think an error reply can contain ciphertext to decrypt.

Just throwing ideas here, but a possible edge case might be if an error message included some document data (e.g. duplicate value for unique index, JSON schema violation); however, I don't think that's important to handle.

src/libmongoc/tests/test-mongoc-client-side-encryption.c
```diff
@@ -4181,6 +4181,387 @@ test_unique_index_on_keyaltnames (void *unused)
...
+                          "socket error");
+   ASSERT (!got);
+   mongoc_cursor_destroy (cursor);
+   decryption_events_fixture_destroy (def);
+}
+
+/* test_decryption_events_decrypt_error is a regression test for CDRIVER-4401.
+ * Decrypt a reply with a malformed ciphertext.
+ * Expect an error returned and a CommandSucceeded event to be emitted with
+ * ciphertext. */
```
jmikola, line 4477: CommandSucceeded is expected because the command itself succeeds, and only decryption fails, correct?

jmikola (approved):

eramongodb (approved):

src/libmongoc/src/mongoc/mongoc-cluster.c
kevinAlbs: >If I understand correctly, the two changes here are:
>
> * Moving _mongoc_cse_auto_decrypt until after all APM callbacks have been dispatched
> * Only decrypting if the original command was successful

Correct.

> Just throwing ideas here, but a possible edge case might be if an error message included some document data (e.g. duplicate value for unique index, JSON schema violation); however, I don't think that's important to handle.

Good idea. A duplicate value for a unique index could include ciphertext in the response. Here is an example with the shell:

```js
db.encryptedColl.createIndex({encrypted: 1}, {unique: true})
db.runCommand({insert: "encryptedColl", documents: [{encrypted: "ciphertext"}]})
got = db.runCommand({insert: "encryptedColl", documents: [{encrypted: "ciphertext"}]})
printjson (got)
/* Prints:
{
    "n" : 0,
    "electionId" : ObjectId("7fffffff0000000000000001"),
    "opTime" : {
            "ts" : Timestamp(1654288999, 3),
            "t" : NumberLong(1)
    },
    "writeErrors" : [
            {
                    "index" : 0,
                    "code" : 11000,
                    "errmsg" : "E11000 duplicate key error collection: test.encryptedColl index: encrypted_1 dup key: { encrypted: \"ciphertext\" }",
                    "keyPattern" : {
                            "encrypted" : 1
                    },
                    "keyValue" : {
                            "encrypted" : "ciphertext"
                    }
            }
    ],
    "ok" : 1,
    "$clusterTime" : {
            "clusterTime" : Timestamp(1654289001, 1),
            "signature" : {
                    "hash" : BinData(0,"AAAAAAAAAAAAAAAAAAAAAAAAAAA="),
                    "keyId" : NumberLong(0)
            }
    },
    "operationTime" : Timestamp(1654288999, 3)
}
*/
```

However, trying to decrypt an error reply may also error. Rather than overwrite the original command error, I think it is preferable to return the command error immediately.

src/libmongoc/tests/test-mongoc-client-side-encryption.c
kevinAlbs: Correct.

PR #1028, opened 2022-06-06
Set MONGO_USE_CCACHE to OFF by default

eramongodb:
### Description

```
ccache: error: /home/ubuntu/.ccache/ccache.conf: No such file or directory
```

Various Evergreen tasks are failing spuriously due to the above ccache error frequently enough that I would like to suggest the option be disabled by default until the issue is properly tested and addressed.

### DIFF
--- a/build/cmake/CCache.cmake
+++ b/build/cmake/CCache.cmake
@@ -12,7 +12,7 @@
 find_program (CCACHE_EXECUTABLE ccache)
 if (CCACHE_EXECUTABLE)
     message (STATUS "Found ccache: ${CCACHE_EXECUTABLE}")
-    option (MONGO_USE_CCACHE "Use CCache when compiling" ON)
+    option (MONGO_USE_CCACHE "Use CCache when compiling" OFF)
 endif ()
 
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
I think this will help with test stability.

build/cmake/CCache.cmake
```diff
@@ -12,7 +12,7 @@
 find_program (CCACHE_EXECUTABLE ccache)
 if (CCACHE_EXECUTABLE)
     message (STATUS "Found ccache: ${CCACHE_EXECUTABLE}")
-    option (MONGO_USE_CCACHE "Use CCache when compiling" ON)
+    option (MONGO_USE_CCACHE "Use CCache when compiling" OFF)
```
vector-of-bool, line 15: I would prefer that this be specified by a build script or evergreen option, as this pessimises the overwhelmingly common case where CCache will "just work."

eramongodb (commented):
Superceded by https://github.com/mongodb/mongo-c-driver/pull/1029.

PR #1031, opened 2022-06-07
Do not elide find_package() if MONGO_USE_CCACHE is set

eramongodb:
### Description

Fix overzealous elison of call to ~`find_package()`~ `find_program()` introduced in https://github.com/mongodb/mongo-c-driver/pull/1029 when `MONGO_USE_CCACHE` is set. `CCACHE_EXECUTABLE` still needs to be found and set when `MONGO_USE_CCACHE` is manually set by the user on initial configuration, not just post-configuration.

### DIFF
--- a/build/cmake/CCache.cmake
+++ b/build/cmake/CCache.cmake
@@ -8,41 +8,40 @@
     ON or OFF.
 ]]
 
-# Find and enable ccache for compiling if not already found.
-if (NOT DEFINED MONGO_USE_CCACHE)
-    find_program (CCACHE_EXECUTABLE ccache)
-    if (CCACHE_EXECUTABLE)
-        message (STATUS "Found ccache: ${CCACHE_EXECUTABLE}")
+find_program (CCACHE_EXECUTABLE ccache)
 
-        execute_process (
-            COMMAND ${CCACHE_EXECUTABLE} --version | perl -ne "print $1 if /^ccache version (.+)$/"
-            OUTPUT_VARIABLE CCACHE_VERSION
-            OUTPUT_STRIP_TRAILING_WHITESPACE
-        )
+# Enable ccache for compiling if not already configured.
+if (CCACHE_EXECUTABLE AND NOT DEFINED MONGO_USE_CCACHE)
+    message (STATUS "Found ccache: ${CCACHE_EXECUTABLE}")
 
-        # Assume `ccache --version` mentions a simple version string, e.g. "1.2.3".
-        # Permit patch number to be omitted, e.g. "1.2".
-        set (SIMPLE_SEMVER_REGEX "([0-9]+)\.([0-9]+)(\.([0-9]+))?")
-        string (REGEX MATCH "${SIMPLE_SEMVER_REGEX}" CCACHE_VERSION ${CCACHE_VERSION})
+    execute_process (
+        COMMAND ${CCACHE_EXECUTABLE} --version | perl -ne "print $1 if /^ccache version (.+)$/"
+        OUTPUT_VARIABLE CCACHE_VERSION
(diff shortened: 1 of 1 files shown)


build/cmake/CCache.cmake
```diff
@@ -8,41 +8,40 @@
...
+    # Avoid spurious "ccache.conf: No such file or directory" errors due to ccache being invoked in parallel, which was patched in ccache version 3.4.3.
+    if (${CCACHE_VERSION} VERSION_LESS 3.4.3)
+        message (STATUS "Detected ccache version ${CCACHE_VERSION} is less than 3.4.3, which may lead to spurious failures when run in parallel. See https://github.com/ccache/ccache/issues/260 for more information.")
+        message (STATUS "Compiling with CCache disabled. Enable by setting MONGO_USE_CCACHE to ON")
+        option (MONGO_USE_CCACHE "Use CCache when compiling" OFF)
+    else ()
+        message (STATUS "Compiling with CCache enabled. Disable by setting MONGO_USE_CCACHE to OFF")
+        option (MONGO_USE_CCACHE "Use CCache when compiling" ON)
+    endif ()
+endif (CCACHE_EXECUTABLE AND NOT DEFINED MONGO_USE_CCACHE)
```
vector-of-bool, line 44: Is there a reason you're repeating the condition in the `endif()` call?

build/cmake/CCache.cmake
```diff
@@ -8,41 +8,40 @@
...
+
+    if (CCACHE_VERSION)
+        message (STATUS "Detected ccache version: ${CCACHE_VERSION}")
+    else ()
+        message (WARNING "Could not obtain ccache version from `ccache --version`. Defaulting to 0.1.0.")
+        set (CCACHE_VERSION 0.1.0)
+    endif ()
+
+    # Avoid spurious "ccache.conf: No such file or directory" errors due to ccache being invoked in parallel, which was patched in ccache version 3.4.3.
+    if (${CCACHE_VERSION} VERSION_LESS 3.4.3)
```
vector-of-bool, line 36:
```suggestion
    if (CCACHE_VERSION VERSION_LESS "3.4.3")
```

The `if()`, `elseif()`, and `while()` commands process their arguments differently depending on whether they are quoted:

- If an operand is unquoted, it will be treated as the name of a variable and dereferenced.
- If an operand is quoted, it will be treated exactly as evaluated.

This is one of the biggest quirks of CMake's syntax, and only really happens in conditionals, but it's easy to remember once you've seen it a few times. (Refer: https://cmake.org/cmake/help/latest/command/if.html#variable-expansion)

vector-of-bool (commented):
I discovered one other quirk of note. Also, I'm guessing you probably meant "`find_program()`" in the commit/PR message 🙂.

build/cmake/CCache.cmake
eramongodb: Just to not lose track of the corresponding `if ()` given the size of the block, a la `#endif // COND` or `} // namespace <name>`. I can remove if you think it is undesirable/unnecessary.

build/cmake/CCache.cmake
eramongodb: Thank you. Looks like I need to review my understanding of CMake intricacies.

build/cmake/CCache.cmake
vector-of-bool: It only tends to be a code smell as it is a very very old pattern that is disrecommended anymore in favor of just using  prose-like comments for the purpose, but I don't feel strongly about it here.

vector-of-bool (approved):

eramongodb:
> I'm guessing you probably meant "find_program()" in the commit/PR message

Fixed it in the commit message in the description, but not in the commit message header itself. 🤦

PR #1036, opened 2022-06-14
[CDRIVER-4402] Document Queryable Encryption as an experimental API

vector-of-bool:
[CDRIVER-4402]

The ticket requests updating the documentation on `indexKeyId`, but that parameter is not exposed in the public API (yet).


kevinAlbs (approved):
LGTM with a punctuation suggestion.

> The ticket requests updating the documentation on indexKeyId, but that parameter is not exposed in the public API (yet).

indexKeyId is TBD in CDRIVER-4405.

I like the `versionadded` directives.

PR #1038, opened 2022-06-15
CDRIVER-4379 add example use of Queryable Encryption

kevinAlbs:
# Summary

- Add example use of Queryable Encryption.

# Background & Motivation

This example was requested to assist the documentation team. It is not intended to be in the C driver public documentation.


eramongodb (approved):
Minor feedback and some questions; otherwise, LGTM.

PR #1043, opened 2022-06-22
[CDRIVER-4382]: Implement on-demand credentials for AWS in KMS

vector-of-bool:
[DRIVERS ticket](https://jira.mongodb.org/browse/DRIVERS-2280)

This does not (yet) contain test cases. I'm unsure on how we want to do tests since this is very environment-sensitive behavior. It may be easier to implement this functionality in terms of the on-demand credentials callback, since the callback behavior can be validated without depending on the environment. ~~I have a separate set of changes that implement the credentials callback, but I'll put that in a separate PR, and we can rebase if desired.~~ (edit: done).


src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -575,6 +580,133 @@ _state_need_kms (_state_machine_t *state_machine, bson_error_t *error)
...
+   /// subdocument that is not "aws". Therefore, this 'return true' is only
+   /// reachable if 'aws' is present and is the only empty subdocument.
+   BSON_ASSERT (found_empty_aws_doc);
+   return true;
+}
+
+static bool
+_state_need_kms_credentials (_state_machine_t *sm, bson_error_t *error)
+{
+   if (_validate_kms_on_demand_aws_only (&sm->crypt->kms_providers, error)) {
```
kevinAlbs, line 653:
```suggestion
   if (!_validate_kms_on_demand_aws_only (&sm->crypt->kms_providers, error)) {
```

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -1019,6 +1163,7 @@ _mongoc_crypt_new (const bson_t *kms_providers,
 
    if (!success) {
       _mongoc_crypt_destroy (crypt);
+      bson_destroy (&crypt->kms_providers);
```
kevinAlbs, line 1166:
```suggestion
```

This is destroyed in `_mongo_crypt_destroy`. This may be the cause of double frees in ASAN tasks.

kevinAlbs:
> @kevinAlbs I added logic to check that `aws` is the only kmsProvider that is unspecified. Is this necessary? i.e. should we rely on libmongocrypt to detect that we are failing to fill out other providers?

Good question. IMO it should be libmongocrypt's responsibility to error if requested credentials are not satisfied. Here is a test validating this behavior: https://github.com/mongodb/libmongocrypt/pull/374.

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -576,37 +581,158 @@ _state_need_kms (_state_machine_t *state_machine, bson_error_t *error)
...
+      const bool okay1 = sm->crypt->creds_cb.fn (
+         sm->crypt->creds_cb.userdata, &empty, &creds, error);
+      if (!okay1) {
+         if (!error->code) {
+            bson_set_error (error,
+                            MONGOC_ERROR_CLIENT_SIDE_ENCRYPTION,
+                            MONGOC_ERROR_CLIENT_INVALID_ENCRYPTION_ARG,
+                            "Unknown error from user-provided callback for "
+                            "on-demand KMS credentials");
+         }
```
eramongodb, line 696:
```suggestion
         if (!error->code) {
            bson_set_error (
               error,
               MONGOC_ERROR_CLIENT_SIDE_ENCRYPTION,
               MONGOC_ERROR_CLIENT_INVALID_ENCRYPTION_ARG,
               "user-provided callback for on-demand credentials failed");
         }
```
Suggest avoiding stating "unknown error"; user-provided callback may report additional details as desired.

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -576,37 +581,158 @@ _state_need_kms (_state_machine_t *state_machine, bson_error_t *error)
...
    bson_t creds = BSON_INITIALIZER;
-   BSON_ASSERT (sm->crypt->creds_cb.fn);
    const bson_t empty = BSON_INITIALIZER;
+   bool okay = false;
+
+   if (sm->crypt->creds_cb.fn) {
+      // We have a user-provided credentials callback. Try it.
+      const bool okay1 = sm->crypt->creds_cb.fn (
+         sm->crypt->creds_cb.userdata, &empty, &creds, error);
+      if (!okay1) {
```
eramongodb, line 689:
```suggestion
      if (!sm->crypt->creds_cb.fn (
             sm->crypt->creds_cb.userdata, &empty, &creds, error)) {
```

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -576,37 +581,158 @@ _state_need_kms (_state_machine_t *state_machine, bson_error_t *error)
...
+   // Attempt to obtain AWS credentials from the environment.
+   _mongoc_aws_credentials_t got_creds;
+   if (!_mongoc_aws_credentials_obtain (NULL, &got_creds, error)) {
+      // Failed to get credentials. This isn't necessarily an error.
+      return error->code == 0;
+   }
+
+   // We have the credentials. Give them back to libmongocrypt
+
+   bson_t aws = {0};
```
eramongodb, line 651:
```suggestion
   bson_t aws = BSON_INITIALIZER;
```

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -576,37 +581,158 @@ _state_need_kms (_state_machine_t *state_machine, bson_error_t *error)
...
+   // Begin the new "aws" subdoc
+   if (!BSON_APPEND_DOCUMENT_BEGIN (out, "aws", &aws)) {
```
eramongodb, line 653: Suggest converting `if (expr) { return false; }` into `BSON_ASSERT (expr);` in this function given failure is not recorded as an error.

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -576,37 +581,158 @@ _state_need_kms (_state_machine_t *state_machine, bson_error_t *error)
...
+ * @return false If there was an error while updating the BSON data or obtaining
+ * credentials.
+ */
+static bool
+_try_add_aws_from_env (bson_t *out, bson_error_t *error)
+{
+   // Attempt to obtain AWS credentials from the environment.
+   _mongoc_aws_credentials_t got_creds;
+   if (!_mongoc_aws_credentials_obtain (NULL, &got_creds, error)) {
+      // Failed to get credentials. This isn't necessarily an error.
```
eramongodb, line 645: This comment contradicts [documentation](https://github.com/mongodb/mongo-c-driver/blob/8761de488ef9655144a5a4811a98d755f5c1e9a7/src/libmongoc/src/mongoc/mongoc-cluster-aws.c#L493) for `_mongoc_aws_credentias_obtain ()`. Does the latter need to be updated?

vector-of-bool: Upon a closer reading, the docs on `obtain()` are correct, and that does indicate a failure.

src/libmongoc/src/mongoc/mongoc-crypt.c
vector-of-bool: This always boils down to whether one would consider allocation failure to be recoverable, as that is the only condition in which these statements will fail.

eramongodb: Given we do not have a `MONGOC_ERROR_OUT_OF_MEMORY` or equivalent error code to report this class of failure, I am in favor of converting these to assertions. If not converted to assertions, `error` should at least be set prior to `goto fail;` under the `// Error while trying to add AWS credentials` condition.

kevinAlbs (approved):
Nice, LGTM.

eramongodb (approved):

PR #1049, opened 2022-06-24
CDRIVER-4417 add explicit encryption case 6 for default contention

kevinAlbs:


kevinAlbs:
No longer applicable.

PR #1048, opened 2022-06-24
CDRIVER-4397 Reapply updates to FLE 2 find payloads

kevinAlbs:
Resyncs FLE 2 tests from https://github.com/mongodb/specifications/pull/1261.

PR #1051, opened 2022-06-24
[CDRIVER-4276] On-demand Credentials Callback

vector-of-bool:
This implements CDRIVER-4276 [edit: See below], which allows the user to set a callback function to fill in `kmsProviders` lazily when they are needed for encryption.

This contains logic to handle the `NEEDS_CREDENTIALS` state, which will also be useful as a basis for #1043

The test case added asserts that the callback is called as appropriately, but does not (yet) fill in the `kmsProviders` sufficient to let libmongocrypt succeed.

---

Edit: This _does not_ implement all of CDRIVER-4276, which concerns itself with AWS auth on client connection. This only implements on-demand credentials for client-side encryption.


src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -575,6 +576,35 @@ _state_need_kms (_state_machine_t *state_machine, bson_error_t *error)
 #undef BUFFER_SIZE
 }
 
+static bool
+_state_need_kms_credentials (_state_machine_t *sm, bson_error_t *error)
+{
+   bson_t creds = BSON_INITIALIZER;
+   BSON_ASSERT (sm->crypt->creds_cb.fn);
```
eramongodb, line 583: Should this assertion be converted into a "nice" BSON error instead? Given the possibility (is there?) that the `NEED_KMS_CREDENTIALS` state is reached without a callback being set (the default). If we want to enforce this assertion, I suggest adding a not-null assertion to `mongoc_auto_encryption_opts_set_kms_credential_provider_callback`.

vector-of-bool: In its current state, the code only gets here if the callback is non-null. This state needs to be explicitly enabled when the context was created.

eramongodb (approved):

src/libmongoc/src/mongoc/mongoc-client-side-encryption.h
```diff
@@ -33,10 +33,16 @@ struct _mongoc_cursor_t;
...
 
+#define MONGOC_ENCRYPT_QUERY_TYPE_EQUALITY "equality"
+
+
 BSON_BEGIN_DECLS
 
 typedef struct _mongoc_auto_encryption_opts_t mongoc_auto_encryption_opts_t;
 
+typedef bool (*mongoc_kms_credentials_provider_callback_fn) (
+   void *userdata, bson_t *out, bson_error_t *error);
```
kevinAlbs, line 44:
```suggestion
   void *userdata, bson_t *out, void* reserved, bson_error_t *error);
```

Consider adding an argument for future extension. For example: libmongocrypt does not distinguish which KMS provider credentials are being requested. That was discussed and [determined out of scope](https://github.com/kevinAlbs/mongo-go-driver/pull/1#issuecomment-1058214761). If that is added in the future, it may be convenient to pass another handle to the callback.

src/libmongoc/tests/test-mongoc-client-side-encryption.c
```diff
@@ -4757,6 +4757,120 @@ test_qe_docs_example (void *unused)
...
+   struct kms_callback_data *ctx = userdata;
+   ctx->value = 42;
+   if (ctx->set_error) {
+      bson_set_error (error,
+                      MONGOC_ERROR_CLIENT_SIDE_ENCRYPTION,
+                      1729,
+                      "%s",
+                      ctx->set_error);
+      return false;
+   }
```
kevinAlbs, line 4777:
```suggestion
   }
   uint8_t data[96] = {0};
   BCON_APPEND (out,
                "local",
                "{",
                "key",
                BCON_BIN (BSON_SUBTYPE_BINARY, data, sizeof (data)),
                "}");
```

Consider adding successful case with the 96 byte "key".

kevinAlbs (commented):
Please add a docs pages for the new `mongoc_client_encryption_opts_set_kms_credential_provider_callback` and `mongoc_auto_encryption_opts_set_kms_credential_provider_callback` functions.

src/libmongoc/src/mongoc/mongoc-client-side-encryption.h
vector-of-bool: Should I give it a stronger type than `void*`? i.e. `const bson_t* params`

kevinAlbs: `const bson_t* params` SGTM.

I thought it may not be ABI / API breaking if a `void*` argument is later changed to a non-void pointer. [Further experimentation](https://godbolt.org/z/dWW3j7a6x) suggest that is incorrect.

src/libmongoc/doc/mongoc_client_encryption_opts_set_kms_credential_provider_callback.rst
```diff
@@ -0,0 +1,60 @@
...
+.. code-block:: c
+
+  void
+  mongoc_client_encryption_opts_set_kms_credential_provider_callback (
+    mongoc_client_encryption_opts_t *opts,
+    mongoc_kms_credentials_provider_callback_fn fn,
+    void *userdata);
+
+Set the user-provided callback to provide KMS credentials on-demand when they
+are needed for explicit encryption.
```
kevinAlbs, line 20:
```suggestion
are needed.
```

Credentials are also needed for creating a datakey and rewrapping datakeys.

kevinAlbs (approved):

kevinAlbs:
The failing ASAN tasks may be resolved with https://github.com/mongodb/mongo-c-driver/pull/1048

PR #1059, opened 2022-07-06
Inject our own project as the mongoc source for building libmongocrypt

vector-of-bool:
libmongocrypt needs a copy of mongo-c-driver to build a libbson for itself. We are already such a mongo-c-driver, so send that to the libmongocrypt sub-build.


rcsanchez97 (approved):
After merge, please cherry-pick the commit onto the `r1.22` branch and push it there as well.

PR #1063, opened 2022-07-08
Reducing Warnings - Address division by zero warnings by scan-build

eramongodb:
## Description

Resolves the following warnings that are causing failure of `debug-compile-scan-build` tasks:

```
.../src/zlib-1.2.12/gzread.c:438:43: warning: Division by zero
    return len ? gz_read(state, buf, len) / size : 0;
                 ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
```
```
.../src/zlib-1.2.12/gzwrite.c:309:44: warning: Division by zero
    return len ? gz_write(state, buf, len) / size : 0;
                 ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
```

### DIFF
--- a/.evergreen/compile-unix.sh
+++ b/.evergreen/compile-unix.sh
@@ -259,6 +259,11 @@ if [ "$ANALYZE" = "ON" ]; then
    else
       SCAN_BUILD_COMMAND="scan-build-3.9"
    fi
+
+   # Do not include bundled zlib in scan-build analysis.
+   # scan-build `--exclude`` flag is not available on all Evergreen variants.
+   CONFIGURE_FLAGS="$CONFIGURE_FLAGS -DENABLE_ZLIB=OFF"
+
    $SCAN_BUILD_COMMAND $CMAKE $CONFIGURE_FLAGS .
 
    # Put clang static analyzer results in scan/ and fail build if warnings found.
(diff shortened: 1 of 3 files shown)


kevinAlbs (approved):
LGTM.

Consider filing a PR to https://github.com/madler/zlib. If and when we vendor a newer version of zlib, we may accidentally overwrite this fix.

eramongodb:
An issue was already filed https://github.com/madler/zlib/issues/214 but rejected on grounds of the static analyzer warning being a false positive (which on closer reading appears to still hold true), so it is unlikely they will accept a PR to address this issue. It may be preferable to exclude the bundled zlib library from scan-build analysis, deferring static analysis of zlib code to zlib maintainers instead.

kevinAlbs (approved):
> An issue was already filed as [madler/zlib#214](https://github.com/madler/zlib/issues/214) but rejected on grounds of the static analyzer warning being a false positive (which on closer reading appears to still hold true), so it is unlikely they will accept a PR to address this issue. It may be preferable to exclude the bundled zlib library from scan-build analysis, deferring static analysis of zlib code to zlib maintainers instead.

Agreed. I think it is preferable to exclude zlib from scan-build.

PR #1065, opened 2022-07-11
Reducing Warnings - Unused Parameters

eramongodb:
## Description

This PR addresses all `-Wunused-parameter` warnings emitted by GCC and Clang given the environment and configuration described below. As of https://github.com/mongodb/mongo-c-driver/commit/3ed55ed9b01a22e8208f9f382c9a976645bdbe4a, there are 583 unique `-Wunused-parameter` warnings given `ENABLE_CLIENT_SIDE_ENCRYPTION=OFF` and 556 unique `-Wunused-parameter` warnings given `ENABLE_CLIENT_SIDE_ENCRYPTION=ON` on both GCC and Clang. "Unique" meaning a unique combination of warning type, file, line, and column; duplicates do not contribute to warning count.

The new `BSON_UNUSED` macro uses the well-known cast-to-void pattern to silence a `-Wunused-parameter` warning.

All functions that emitted an `-Wunused-parameter` warning were updated to include a `BSON_UNUSED (param);` statement accordingly.

A few instances of using `BSON_ASSERT` in `test-mongoc-dns.c` to silence `-Wunused-parameter` warnings were instead inadvertently generating `comparison of unsigned expression >= 0 is always true [-Wtype-limit]` warnings instead. These were replaced with `BSON_UNUSED` accordingly.

Several files under the `examples` directory were updated to avoid declaring unused `argc` and `argv` parameters in `main ()`.

All updated files were run through clang-format, addressing several `-Wnewline-eof` warnings.

## Configuration and Environment

Not all combinations of build configuration or target platforms have been tested, which may affect the set of warnings generated. The numbers mentioned in this PR description are given the following environment and configuration:

### Environment

- OS: Ubuntu 20.04
- GCC Version: 9.4.0
- Clang Version: 10.0.0
- zlib Version (system): 1.2.11
- zstd Version (system): 1.4.4
- No Snappy Compression
- No ICU Library
- OpenSSL Version (system): 1.1.1f
- Cyrus SASL Version (system): 2.1.27

### CMake Configuration

- CMAKE_C_FLAGS: -std=c99 -Wall -Wextra -pedantic -Wconversion -Wsign-conversion
- ENABLE_AUTOMATIC_INIT_AND_CLEANUP: OFF
- ENABLE_EXTRA_ALIGNMENT: OFF
- ENABLE_CLIENT_SIDE_ENCRYPTION: OFF / ON

All other CMake configuration variables were set to their default value.


src/libbson/src/bson/bson-json.c
```diff
@@ -1370,42 +1369,26 @@ _bson_json_read_map_key (bson_json_reader_t *reader, /* IN */
...
+                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             val))
+         {
+            bson->bson_type = BSON_TYPE_TIMESTAMP;
+            bson->read_state = BSON_JSON_IN_BSON_TYPE_TIMESTAMP_STARTMAP;
+         }
+      else if (!strcmp ("$regularExpression", (const char *) val)) {
```
vector-of-bool, line 1391: Probably not intended.

src/libmongoc/src/mongoc/mongoc-rpc.c
```diff
@@ -635,42 +635,46 @@ _mongoc_rpc_gather (mongoc_rpc_t *rpc, mongoc_array_t *array)
...
-   case MONGOC_OPCODE_KILL_CURSORS:
-      _mongoc_rpc_swab_to_le_kill_cursors (&rpc->kill_cursors);
-      break;
-   case MONGOC_OPCODE_COMPRESSED:
-      _mongoc_rpc_swab_to_le_compressed (&rpc->compressed);
-      break;
-   default:
-      MONGOC_WARNING ("Unknown rpc type: 0x%08x", opcode);
-      break;
+#if BSON_BYTE_ORDER != BSON_LITTLE_ENDIAN
```
vector-of-bool, line 640: Aside: I'm strongly in favor of using language-level checks rather than conditional compilation if at all possible. That would remove the need to annotate "this variable might be unused depending on the preprocessor state." It also ensures we have valid code in the conditional block even on systems where the block would otherwise be unevaluated.

vector-of-bool (approved):
This will be especially helpful. I've reviewed all the files (there are many) and nothing is suspicious.

Do we want to include the "unused" warnings-as-error as part of CI/dev checks?

eramongodb:
> Do we want to include the "unused" warnings-as-error as part of CI/dev checks?

I am hoping to eventually reach a state where compiling with `-Wall -Wextra -pedantic -Wconversion -Wsign-conversion` does not generate any warnings. As for whether we want to enforce this in CI via `-Werror`, I am undecided. 🤔

I am willing to revisit this question after other warnings currently being generated have also been addressed.

kevinAlbs (approved):
Nicely done. LGTM.

PR #1066, opened 2022-07-12
Improve clang-format friendliness of HANDLE_OPTION macro

eramongodb:
## Description

Motivated by the number of times this file has been attempted to be clang-formatted only for the `HANDLE_OPTION` blocks to be rendered utterly unreadable. Hopefully this is preferable to disabling clang-format for such a large block of code. Given how strange the `HANDLE_OPTION` macro is to begin with, I'm willing to accept the strangeness of the `_selection_statement` pattern. This code could use a refactor down the road to avoid these unusual patterns.

### DIFF
--- a/src/libbson/src/bson/bson-json.c
+++ b/src/libbson/src/bson/bson-json.c
@@ -370,8 +370,9 @@ _noop (void)
                                  read_state_names[bson->read_state]);          \
       return;                                                                  \
    }
-#define HANDLE_OPTION(_key, _type, _state)                                     \
-   (len == strlen (_key) && strncmp ((const char *) val, (_key), len) == 0)    \
+#define HANDLE_OPTION(_selection_statement, _key, _type, _state)               \
+   _selection_statement (len == strlen (_key) &&                               \
+                         strncmp ((const char *) val, (_key), len) == 0)       \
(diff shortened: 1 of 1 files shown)


src/libbson/src/bson/bson-json.c
```diff
@@ -1368,98 +1369,98 @@ _bson_json_read_map_key (bson_json_reader_t *reader, /* IN */
       _bson_json_save_map_key (bson, val, len);
    }
 
-   /* clang-format off */
    if (bson->read_state == BSON_JSON_IN_BSON_TYPE) {
-      if
```
vector-of-bool, line 1373: If you lead with an `if (false) {}`, then every subsequent HANDLE_OPTION can just be `else if`.

kevinAlbs (approved):

src/libbson/src/bson/bson-json.c
eramongodb: I would strongly prefer to avoid hiding control flow statements within what is already an esoteric macro. The introduction of a dead code block feels in essence no different in spirit to an `#if 0`, which I would also like to avoid.

vector-of-bool (approved):

PR #1067, opened 2022-07-12
Reducing Warnings - Compile with C99 on Evergreen by default

eramongodb:
## Description

This PR is a followup to https://github.com/mongodb/mongo-c-driver/pull/920 and is related to CDRIVER-4136.

This PR ensures that all builds of the C Driver on Evergreen are able to be compiled with `-std=c99` (or equivalent) by default to ensure strict conformance with the C99 standard across all platforms and configurations. This addresses all C89/C90 compatibility related warnings emitted due to use of C99 features.

In the spirit of minimizing platform-specific configuration flags, the C standard is specified via CMake's `CMAKE_C_STANDARD` and `CMAKE_C_STANDARD_REQUIRED` on both Linux and Windows. `CMAKE_C_EXTENSIONS` is set to `OFF` to ensure strict C99 conformance and that any extensions being depended on are explicitly opted into.

Setting `CMAKE_C_EXTENSIONS` to `OFF` exposed a configuration bug on RHEL 6.2 and RHEL 7.0, which provides glibc versions 2.12 and 2.17 respectively. https://github.com/mongodb/mongo-c-driver/pull/920 proposed removing `_BSD_SOURCE` in favor of `_DEFAULT_SOURCE` due to the former being deprecated; however, `_DEFAULT_SOURCE` [is only supported given glibc version 2.19 and newer](https://man7.org/linux/man-pages/man7/feature_test_macros.7.html). To support glibc version 2.19 and earlier while also supporting strict C99 conformance, this PR partially reverts the changes in https://github.com/mongodb/mongo-c-driver/pull/920.

Therefore, the full set of compile definitions to explicitly opt into POSIX features and non-standard extensions currently being used by the C driver include:

- `_XOPEN_SOURCE=700`: required to enable use of `strnlen` (>=700) in addition to other features in the X/Open namespace (i.e. `pthread_rwlock_*` (>=500)) [as required by the POSIX specification](https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/).
- `_DEFAULT_SOURCE`: required to enable use of `getpagesize`, `h_errno`, and other BSD-derived definitions.
- `_BSD_SOURCE`: required for [compatibility with glibc 2.19 and earlier](https://man7.org/linux/man-pages/man7/feature_test_macros.7.html).
- `_DARWIN_C_SOURCE`: required to enable use of `_SC_NPROCESSORS_ONLN` and other [non-POSIX extensions on MacOS](https://opensource.apple.com/source/Libc/Libc-1439.40.11/gen/compat.5.auto.html) that are otherwise disabled by `_POSIX_C_SOURCE` (set to `200112L` by `_XOPEN_SOURCE=700`).

`_BSD_SOURCE` could be removed once all platforms with glibc version 2.19 or earlier are no longer supported.

The `debug-compile-c99` task is removed due to redundancy.


.evergreen/compile-unix.sh
```diff
@@ -28,6 +29,7 @@ set -o errexit  # Exit the script with error if any of the commands fail
 #       ZSTD                    Build against system zstd.
 
 # Options for this script.
+C_STD_VERSION=${CSTD_VERSION:-99}
```
vector-of-bool, line 32: +1 for a dedicated env var

vector-of-bool (approved):

kevinAlbs (approved):
LGTM. The comments in CMakeLists.txt look helpful for future readers.

kevinAlbs:
If you have not already, I suggest running a full patch build to check compile tasks on all platforms.

eramongodb:
@kevinAlbs Latest patch builds are noisy due to unrelated server compat issues, but compile tasks were verified here: https://spruce.mongodb.com/version/62cc9a6557e85a7ecf6b16f2/tasks

.evergreen/compile-unix.sh
jmikola, line 32: @eramongodb: I think this was a typo, but it appears to have been fixed in a subsequent commit: https://github.com/mongodb/mongo-c-driver/commit/f4a83c084088e3878097b17881458548bc621ea2#diff-5621bd7804f10985a5f14a01ee66b14439494aa0a6cdfa6c6debd151e7c90d7aR10

Likewise for `compile-windows.sh`.

eramongodb: Correct. This mistake was in part what motivated the addition of the `check_var_opt`/`check_var_req` functions in the referenced commit.

PR #1068, opened 2022-07-13
Update generated files for Evergreen config and future functions

eramongodb:
## Description

~It seems like something has changed with how the Evergreen config file is generated (namely the formatting). This PR commits these updates to the `config.yml` file.~ Patch applied, prior "block style" formatting is now preserved.

This PR also updates `generate-future-functions.py` to account for the new (as of https://github.com/mongodb/mongo-c-driver/pull/904) `must_use_primary` parameter for `mongoc_topology_select`. Call sites for `future_topology_select` were updated with an additional `NULL` argument accordingly.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -13,7 +13,8 @@ exec_timeout_secs: 2400
 functions:
   fetch source:
   - command: git.get_project
-    params: {directory: mongoc}
+    params:
+      directory: mongoc
   - command: shell.exec
     params:
       working_dir: mongoc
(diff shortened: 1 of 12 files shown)


eramongodb:
I am not opposed to reverting the cause of the `config.yml` formatting changes; I prefer the prior syntax to the currently generated syntax.

eramongodb:
Still not clear what prompted the formatting change (why was it being serialized in block style prior?), but identified a correction/workaround [per PyYAML documentation](https://pyyaml.org/wiki/PyYAMLDocumentation):

> If you want collections to be always serialized in the block style, set the parameter `default_flow_style` of `dump()` to `False`.

kevinAlbs (approved):

vector-of-bool (approved):

PR #1071, opened 2022-07-20
Fix setting of mongo_crypt_v1 shared library path

eramongodb:
## Description

The `cryptSharedLibPath` auto encryption option was not being correctly included in the call to `mongoc_auto_encryption_opts_set_extra`.


kevinAlbs (approved):
Nice catch. LGTM.

PR #1080, opened 2022-07-29
CDRIVER-4413, CDRIVER-4399

rcsanchez97:
Included changes:

* CDRIVER-4399 update to Fedora 37 environment for RPM build
* sync RPM spec file for 1.22.0 release
* CDRIVER-4413 adjust configuration for building docs with Sphinx 5.0+

Evergreen patch build: https://spruce.mongodb.com/version/62e438853627e033781fccc3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/libmongoc/doc/conf.py
+++ b/src/libmongoc/doc/conf.py
@@ -29,6 +29,8 @@ language = 'en'
 exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
 master_doc = 'index'
 
+intersphinx_disabled_reftypes = []
+
 # don't fetch libbson's inventory from mongoc.org during build - Debian and
 # Fedora package builds must work offline - maintain a recent copy here
 intersphinx_mapping = {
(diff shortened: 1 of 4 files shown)


src/libmongoc/doc/conf.py
```diff
@@ -29,6 +29,8 @@
 exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
 master_doc = 'index'
 
+intersphinx_disabled_reftypes = []
```
kevinAlbs, line 32: Am I correct that the error is due to intersphinx including `std:doc` as a default in Sphinx 5.0: https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#confval-intersphinx_disabled_reftypes

If so, I suggest noting this with a comment.

```suggestion
# Set an empty list of disabled reftypes.
# Sphinx 5.0 disables "std:doc" by default.
# Many documentation references use :doc:
intersphinx_disabled_reftypes = []
```

kevinAlbs (approved):
LGTM with suggested comment.

PR #1084, opened 2022-08-01
Fix _mongoc_get_cpu_count on Android

ii14:
get_nprocs is not available on Android below API level 23.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-counters-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-counters-private.h
@@ -48,8 +48,14 @@ _mongoc_counters_cleanup (void);
 static BSON_INLINE unsigned
 _mongoc_get_cpu_count (void)
 {
-#if defined(__linux__)
+#if defined(__linux__) && !defined(__ANDROID__)
    return get_nprocs ();
+#elif defined(__ANDROID__) && defined(_SC_NPROCESSORS_CONF)
+   long count = sysconf (_SC_NPROCESSORS_CONF);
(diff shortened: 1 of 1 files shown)


src/libmongoc/src/mongoc/mongoc-counters-private.h
```diff
@@ -48,8 +48,14 @@ _mongoc_counters_cleanup (void);
...
 {
-#if defined(__linux__)
+#if defined(__linux__) && !defined(__ANDROID__)
    return get_nprocs ();
+#elif defined(__ANDROID__) && defined(_SC_NPROCESSORS_CONF)
+   long count = sysconf (_SC_NPROCESSORS_CONF);
+   if (count < 1) {
+      return 1;
+   }
+   return count;
```
vector-of-bool, line 58: It looks like `sysconf()` is generally more portable on Linux systems, so it may be sufficient to replace the call to `get_nprocs()` entirely:

```suggestion
#if defined(__linux__) && defined(_SC_NPROCESSORS_CONF)
   long count = sysconf (_SC_NPROCESSORS_CONF);
   if (count < 1) {
      return 1;
   }
   return count;
```

vector-of-bool (commented):
Thank you for your submission, and thank you for your patience.

I've left an inline comment on possible replacing `get_nprocs()` entirely, which would be preferable to adding another branch, and would be more portable on Linux systems that don't use glibc.

vector-of-bool (approved):
LGTM. Thank you!

PR #1089, opened 2022-08-04
[CDRIVER-4436] Add do-not-retry-read-in-transaction tests

vector-of-bool:
Refer: CDRIVER-4436


vector-of-bool:
Assuming the tests are correct, this appears to "just work" with the code as-is.

kevinAlbs (approved):

PR #1090, opened 2022-08-04
[CDRIVER-4448] Fix interaction of bson_visit with bson_validate

vector-of-bool:
Refer to CDRIVER-4448 for details


src/libbson/src/bson/bson.c
```diff
@@ -3584,7 +3584,19 @@ _bson_iter_validate_document (const bson_iter_t *iter,
       state->phase = BSON_VALIDATE_PHASE_LF_REF_KEY;
    }
 
-   (void) bson_iter_visit_all (&child, &bson_validate_funcs, state);
+   bson_iter_visit_all_v2 (
```
eramongodb, line 3587: Should the result of this call be part of the condition below? e.g.
```c
   if (!bson_iter_visit_all_v2 (
          &child, &bson_validate_funcs, BSON_ITER_VISIT_NOFLAGS, state) &&
       child.err_off > 0 && state->err_offset < 0) {
```

src/libbson/src/bson/bson.c
```diff
@@ -3584,7 +3584,19 @@ _bson_iter_validate_document (const bson_iter_t *iter,
...
+      &child, &bson_validate_funcs, BSON_ITER_VISIT_NOFLAGS, state);
+
+   if (child.err_off > 0 && state->err_offset < 0) {
+      // Iteration on a direct element of 'child' failed
+      state->err_offset = child.err_off;
+      bson_set_error (&state->error,
+                      BSON_ERROR_INVALID,
+                      BSON_VALIDATE_NONE,
+                      "Iteration failed on element at offset %d",
+                      (int) (iter->off + child.err_off));
```
eramongodb, line 3597:
```suggestion
                      "Iteration failed on element at offset %" PRIu32,
                      iter->off + child.err_off);
```

src/libbson/src/bson/bson.c
```diff
@@ -3584,7 +3584,19 @@ _bson_iter_validate_document (const bson_iter_t *iter,
       state->phase = BSON_VALIDATE_PHASE_LF_REF_KEY;
    }
 
-   (void) bson_iter_visit_all (&child, &bson_validate_funcs, state);
+   bson_iter_visit_all_v2 (
+      &child, &bson_validate_funcs, BSON_ITER_VISIT_NOFLAGS, state);
+
+   if (child.err_off > 0 && state->err_offset < 0) {
+      // Iteration on a direct element of 'child' failed
+      state->err_offset = child.err_off;
```
eramongodb, line 3592: There is currently no test coverage of this branch.

src/libbson/doc/bson_iter_visit_all_v2.rst
```diff
@@ -0,0 +1,118 @@
...
+
+  Validate all keys and values. Permits a zero ``0x00`` UTF-8 code unit in
+  strings.
+
+  This flag has the same behavior as :symbol:`bson_iter_visit_all`.
+
+.. note::
+
+  If a requested element fails validation, the ``bson_iter_visit_all_v2`` call
+  will return and indicate the position of the erring element via ``iter``.
```
eramongodb, line 110:
```suggestion
  will return and indicate the position of the erring element via ``iter->err_off``.
```

src/libbson/src/bson/bson.c
```diff
@@ -3584,7 +3584,19 @@ _bson_iter_validate_document (const bson_iter_t *iter,
       state->phase = BSON_VALIDATE_PHASE_LF_REF_KEY;
    }
 
-   (void) bson_iter_visit_all (&child, &bson_validate_funcs, state);
+   bson_iter_visit_all_v2 (
+      &child, &bson_validate_funcs, BSON_ITER_VISIT_NOFLAGS, state);
+
+   if (child.err_off > 0 && state->err_offset < 0) {
```
eramongodb, line 3590:
```suggestion
   if (child.err_off > 0 && state->err_offset <= 0) {
```
Should it be `state->err_offset <= 0` for consistency with the condition used below (line 3604)? Or should the condition below be changed to `state->err_offset < 0`?

eramongodb (changes requested):

src/libbson/src/bson/bson.c
vector-of-bool: By careful reading of the source, the return value from this function is nearly meaningless, because it can return `true` or `false` both for success and for failure. It's just quite unfortunate that we're trying to cram too much information into a single `bool`.

src/libbson/src/bson/bson.c
vector-of-bool: Actually, I believe the addition of `BSON_ITER_VISIT_NOFLAGS` makes this branch unreachable, as `child.err_off` is only assigned when `visit_all` fails to validate an element, and we're now disabling all of those checks. Now I'm getting more suspicious. I'll add more tests, just to make sure.

src/libbson/src/bson/bson.c
```diff
@@ -3584,7 +3584,15 @@ _bson_iter_validate_document (const bson_iter_t *iter,
       state->phase = BSON_VALIDATE_PHASE_LF_REF_KEY;
    }
 
-   (void) bson_iter_visit_all (&child, &bson_validate_funcs, state);
+   bson_iter_visit_all_v2 (
+      &child, &bson_validate_funcs, BSON_ITER_VISIT_VALIDATE_KEYS, state);
```
kevinAlbs, line 3588: If `state.flags` has `BSON_VALIDATE_UTF8`, should `BSON_ITER_VISIT_VALIDATE_STRINGS` be passed instead?

Same question for `BSON_ITER_VISIT_ALLOW_NUL_IN_UTF8`.

PR #1091, opened 2022-08-08
CDRIVER-4451 Avoid unintentional conflict with macro-types in bson-cmp.h macros

eramongodb:
## Description

This PR resolves CDRIVER-4451.

Some types such as `ssize_t` may be defined as macros (rather than typedefs) for existing fundamental types on certain platforms. Expansion of said macro may unintentionally lead to conflicts when defining type-specific inline functions in `bson-cmp.h`:

```c
// Defines `bson_in_range_long_signed` and `bson_in_range_long_unsigned`.
BSON_IN_RANGE_SET_SIGNED (long, LONG_MIN, LONG_MAX)

// `ssize_t` may be a macro that expands to `long` during `BSON_CONCAT3`.
// Accidentally redefines `bson_in_range_long_signed` and `bson_in_range_long_unsigned`.
BSON_IN_RANGE_SET_SIGNED (ssize_t, SSIZE_MIN, SSIZE_MAX)
```

The definitions have been adjusted to avoid unintentionally expanding type macros by moving the `_` suffix from `bson_in_range_` into a prefix on the type name, e.g. `_ssize_t`, to avoid unintentionally using and expanding type macros during function name generation.

```c
// `_ssize` does not accidentally use or expand the `ssize_t` macro.
// Defines `bson_in_range_ssize_t_signed` and `bson_in_range_ssize_t_unsigned` as intended.
BSON_IN_RANGE_SET_SIGNED (_ssize_t, SSIZE_MIN, SSIZE_MAX)
```

~~[As noted below](https://github.com/mongodb/mongo-c-driver/pull/1091#issuecomment-1209440295), the `bson_in_range_signed (Type, value)` and `bson_in_range_unsigned (Type, value)` macros are removed to prevent the same issue on the user ("caller") side. Usage now requires explicitly naming the function with the appropriate type, e.g. `bson_in_range_ssize_signed (value)`.~~ Update: a more direct and simple workaround for potential undesirable type macro expansion is applied so that an API break is not required to fix user code.

### DIFF
--- a/src/libbson/src/bson/bson-cmp.h
+++ b/src/libbson/src/bson/bson-cmp.h
@@ -163,21 +163,32 @@ BSON_IN_RANGE_SET_UNSIGNED (_unsigned_int, UINT_MAX)
 BSON_IN_RANGE_SET_UNSIGNED (_unsigned_long, ULONG_MAX)
 BSON_IN_RANGE_SET_UNSIGNED (_unsigned_long_long, ULLONG_MAX)
 
-BSON_IN_RANGE_SET_SIGNED (_int8, INT8_MIN, INT8_MAX)
-BSON_IN_RANGE_SET_SIGNED (_int16, INT16_MIN, INT16_MAX)
-BSON_IN_RANGE_SET_SIGNED (_int32, INT32_MIN, INT32_MAX)
-BSON_IN_RANGE_SET_SIGNED (_int64, INT64_MIN, INT64_MAX)
+BSON_IN_RANGE_SET_SIGNED (_int8_t, INT8_MIN, INT8_MAX)
+BSON_IN_RANGE_SET_SIGNED (_int16_t, INT16_MIN, INT16_MAX)
+BSON_IN_RANGE_SET_SIGNED (_int32_t, INT32_MIN, INT32_MAX)
+BSON_IN_RANGE_SET_SIGNED (_int64_t, INT64_MIN, INT64_MAX)
 
-BSON_IN_RANGE_SET_UNSIGNED (_uint8, UINT8_MAX)
-BSON_IN_RANGE_SET_UNSIGNED (_uint16, UINT16_MAX)
-BSON_IN_RANGE_SET_UNSIGNED (_uint32, UINT32_MAX)
-BSON_IN_RANGE_SET_UNSIGNED (_uint64, UINT64_MAX)
+BSON_IN_RANGE_SET_UNSIGNED (_uint8_t, UINT8_MAX)
+BSON_IN_RANGE_SET_UNSIGNED (_uint16_t, UINT16_MAX)
+BSON_IN_RANGE_SET_UNSIGNED (_uint32_t, UINT32_MAX)
+BSON_IN_RANGE_SET_UNSIGNED (_uint64_t, UINT64_MAX)
 
-BSON_IN_RANGE_SET_SIGNED (_ssize, SSIZE_MIN, SSIZE_MAX)
-BSON_IN_RANGE_SET_UNSIGNED (_size, SIZE_MAX)
+BSON_IN_RANGE_SET_SIGNED (_ssize_t, SSIZE_MIN, SSIZE_MAX)
+BSON_IN_RANGE_SET_UNSIGNED (_size_t, SIZE_MAX)
 
 #undef BSON_IN_RANGE_SET_SIGNED
 #undef BSON_IN_RANGE_SET_UNSIGNED
+
+
+/* Return true if the value with *signed* type is in the representable range of
+ * Type and false otherwise. */
+#define bson_in_range_signed(Type, value) \
+   BSON_CONCAT3 (bson_in_range_, Type, _signed) (value)
+
+/* Return true if the value with *unsigned* type is in the representable range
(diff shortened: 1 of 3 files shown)


jmikola (approved):
User in https://github.com/mongodb/mongo-php-driver/issues/1342#issuecomment-1208879777 confirmed that this works.

eramongodb:
As the solution is verified to work, this followup commit extends the solution to remove the `bson_in_range_signed (Type, value)` and `bson_in_range_unsigned (Type, value)` macros in favor of explicit calls to `bson_in_range_Type_signed` and `bson_in_range_Type_unsigned` to avoid the type macro expansion issue on the user side of the `bson_in_range_*` utilities. For example, should `ssize_t` be defined as `#define ssize_t long long`, this will not correctly expand into a valid function call:
```c
// Expands to `bson_in_range_long long_signed (0)`
bson_in_range_signed (ssize_t, 0); // error
```
```
error: use of undeclared identifier 'bson_in_range_long'
   bson_in_range_signed (ssize_t, 0);
   ^
```

jmikola (approved):

src/libbson/src/bson/bson-cmp.h
```diff
@@ -118,77 +118,68 @@ BSON_CMP_SET (greater_equal,
...
+BSON_IN_RANGE_SET_UNSIGNED (_uint32, UINT32_MAX)
+BSON_IN_RANGE_SET_UNSIGNED (_uint64, UINT64_MAX)
 
-BSON_IN_RANGE_SET_SIGNED (ssize_t, SSIZE_MIN, SSIZE_MAX)
-BSON_IN_RANGE_SET_UNSIGNED (size_t, SIZE_MAX)
+BSON_IN_RANGE_SET_SIGNED (_ssize, SSIZE_MIN, SSIZE_MAX)
+BSON_IN_RANGE_SET_UNSIGNED (_size, SIZE_MAX)
 
 #undef BSON_IN_RANGE_SET_SIGNED
 #undef BSON_IN_RANGE_SET_UNSIGNED
```
kevinAlbs, line 182: Can these macros be added back to avoid breaking API.

Consider prepending an `_`.

```suggestion
#define bson_in_range_signed(Type, value) \
   BSON_CONCAT3 (bson_in_range, _##Type, _signed) (value)
```

kevinAlbs (changes requested):
I prefer avoiding a breaking API. There is no precedent in the C driver for breaking API or ABI.

Aside: I am not sure why the [abi-compliance-check task passed](https://spruce.mongodb.com/task/mongo_c_driver_releng_abi_compliance_check_patch_d831109c831beb870c24c658eb519443b82a6e8b_62f26bec562343194bc4db8e_22_08_09_14_15_09/files?execution=0).

src/libbson/src/bson/bson-cmp.h
eramongodb: I am surprised this prevents the undesirable type macro expansion behavior. I had incorrectly assumed `Type`  would still be expanded before the `_` prefix is concatenated. I will gladly apply this solution over removal of these macros.

jmikola: Would adding an underscore prefix _and_ suffix to the type and removing the prefix on `_signed` and `_unsigned` be preferable, or is there no risk that we'll run into a case where `_size_t` is a definition?

Just asking because I think I have seen leading or trailing underscores appear in types before (usually within internal headers).

eramongodb: Actually, the `_##Type` solution should be applicable to the `BSON_IN_RANGE_SET_(UN)SIGNED` macros as well, which would obviate the need for the underscore handling as an expansion workaround, which further reduces the scope of changes necessary.

kevinAlbs (approved):

eramongodb:
Validity of the `_##Type` workaround is confirmed by [section 6.10.3.3 The ## operator](https://port70.net/~nsz/c/c99/n1256.html#6.10.3.3) of the C99 standard:

> For both object-like and function-like macro invocations, before the replacement list is reexamined for more macro names to replace, each instance of a `##` preprocessing token in the replacement list (not from an argument) is deleted and the preceding preprocessing token is concatenated with the following preprocessing token. Placemarker preprocessing tokens are handled specially: concatenation of two placemarkers results in a single placemarker preprocessing token, and concatenation of a placemarker with a non-placemarker preprocessing token results in the non-placemarker preprocessing token. If the result is not a valid preprocessing token, the behavior is undefined. The resulting token is available for further macro replacement. The order of evaluation of `##` operators is unspecified.

And in [section 6.10.3.1 Rescanning and further replacement](https://port70.net/~nsz/c/c99/n1256.html#6.10.3.4):

> After all parameters in the replacement list have been substituted and `#` and `##` processing has taken place, all placemarker preprocessing tokens are removed. Then, the resulting preprocessing token sequence is rescanned, along with all subsequent preprocessing tokens of the source file, for more macro names to replace.

Validated by this example:

```c
#define INNER(a, b) a##b
#define OUTER(a, b) INNER(a, b)

#define A abc
#define B def

int INNER(A, B); // Expands to `AB`.
int OUTER(A, B); // Expands to `abcdef`.
```

PR #1092, opened 2022-08-08
Ensure mongocryptdBypassSpawn is set according to environment

eramongodb:
## Description

This PR aims to resolve the `/client_side_encryption/single_and_pool_mismatches` test failure on Evergreen due to:
```
mongocryptd error: No suitable servers found: `serverSelectionTimeoutMS` expired: [connection refused calling hello on 'localhost:27020']
```
The unconditional bypassing of mongocryptd spawn was setting up incorrect expectations of a mongocryptd instance being present. This PR ensures the field is only set if `MONGOC_TEST_MONGOCRYPTD_BYPASS_SPAWN` is set accordingly.


vector-of-bool (approved):

kevinAlbs (approved):
Nice catch. LGTM.

PR #1094, opened 2022-08-15
CDRIVER-4503 A BSON Handling EDSL

vector-of-bool:
This changeset includes a semi-experimental EDSL for working with `bson_t`. This is (currently) a private EDSL. A few locations in the C codebase have been updated to use the EDSL as a "proof of concept," but many more locations could be changed, if desired.

For more information, refer to the `bson-dsl.md` file added by this PR.

A lot of the code quality may reflect that it was prepared quickly as a skunkworks project (thus this is a _private_ API). Many improvements are surely possible. Additional thorough test cases are prudent.


kevinAlbs (commented):
The thorough documentation is much appreciated.

I reviewed the README, tests, and changes to existing source files. In the interest of unblocking CDRIVER-4428, I skimmed bson-dsl.h.

I am not convinced that building BSON is much more readable than the `BCON_*` API. I think the biggest benefit of the DSL is improved readability and error handling consistency of parsing BSON.

src/libbson/src/bson/bson-dsl.h
```diff
@@ -0,0 +1,1320 @@
...
+#define bsonParse(Document, ...)                              \
+   _bsonDSL_begin ("bsonParse(%s)", _bsonDSL_str (Document)); \
+   _bsonDSL_disableWarnings ();                               \
+   bool _bvHalt = false;                                      \
+   bsonParseError = NULL;                                     \
+   const bool _bvContinue = false;                            \
+   const bool _bvBreak = false;                               \
+   (void) _bvContinue;                                        \
+   (void) _bvBreak;                                           \
+   (void) _bvHalt;                                            \
```
eramongodb, line 57:
```suggestion
   BSON_MAYBE_UNUSED const bool _bvContinue = false;          \
   BSON_MAYBE_UNUSED const bool _bvBreak = false;             \
   BSON_MAYBE_UNUSED bool _bvHalt = false;                    \
   bsonParseError = NULL;                                     \
```
Suggest use of `BSON_MAYBE_UNUSED` or `BSON_UNUSED` instead of explicit cast-to-void. Suggestion applies to other instances of cast-to-void in this file.

src/libbson/src/bson/bson-dsl.h
```diff
@@ -0,0 +1,1320 @@
...
+#define bsonVisitIter (bsonVisitContext.iter)
+#define bsonParseIter (bsonVisitContext.iter)
+#else
+/// The context of the current build operation
+extern struct _bsonBuildContext_t bsonBuildContext;
+/// The context of the current parse/visit operation
+extern struct _bsonVisitContext_t bsonVisitContext;
+/// An iterator pointing to the current BSON value being visited/parsed.
+extern bson_iter_t bsonVisitIter, bsonParseIter;
+#endif
```
eramongodb, line 86: What is the purpose of this block (why `__INTELLISENSE__`)?

eramongodb (commented):
Initial set of suggestions and questions. Further review comments pending.

src/libbson/src/bson/bson-dsl.h
```diff
@@ -0,0 +1,1320 @@
...
+   /// BSON_DSL_DEBUG within the scope of a DSL command to selectively debug
+   /// only the commands within that scope.
+   BSON_DSL_DEBUG = 0
+};
+
+#ifdef _WIN32
+#define _bson_thread_local_comdat __declspec(thread) __declspec(selectany)
+#else
+#define _bson_thread_local_comdat __attribute__ ((weak)) __thread
+#endif
```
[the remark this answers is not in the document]
vector-of-bool: The use of globals here is to allow transparent passage across function call boundaries. Whether that is necessary is debatable. It certainly simplifies error handling with the `errno`-style.

src/libbson/src/bson/bson-dsl.h
vector-of-bool: It's to "lie" to the IDE, only to hide the macro expansion-on-hover, but that's getting really pedantic. Probably not worth the added complexity.

src/libbson/src/bson/bson-dsl.h
vector-of-bool: I was unaware we had that macro.

src/libbson/src/bson/bson-dsl.md
```diff
@@ -0,0 +1,974 @@
...
+}
+```
+
+
+## `bsonBuild`
+
+> `bsonBuild(BSON, DocOperation... ops)`
+
+Calls `bson_init(&BSON)`, followed by `bsonBuildAppend(BSON, ops...)`. If the
+build operation results in an error, calls `bson_destroy(&BSON)`.
```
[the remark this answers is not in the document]
vector-of-bool: Aside: In an earlier revision, I attempted to build the document into a temporary, then move it into the destination after it succeeded, otherwise it would just destroy the temporary and leave the destination unchanged. This had the added benefit that the destination document could still be used during construction. This proved infeasible as there was no API to "move" the ownership of the document into another `bson_t` (The intuitive API `bson_steal` marks the destination as read-only for some reason?)

src/libmongoc/src/mongoc/mongoc-aggregate.c
```diff
@@ -92,53 +93,81 @@ _make_agg_cmd (const char *ns,
...
+          if (dot,
+              /* Note: we're not validating that the collection name's length is
+                 one or more characters, as functions such as
+                 mongoc_client_get_collection also do not validate. */
+              // If 'ns' contains a dot, insert the string after the dot:
+              then (cstr (dot + 1)),
+              // Otherwise just an integer 1:
+              else(int32 (1)))));
+   if ((error_hint = "append-aggregate", error = bsonBuildError)) {
+      goto fail;
```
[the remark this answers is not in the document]
vector-of-bool: If `bsonBuild` does not destroy its destination (as per other comment), the "always initialized comment" becomes correct. Should the caller be fixed instead?

src/libbson/src/bson/bson-dsl.md
```diff
@@ -0,0 +1,974 @@
...
+For more information, refer to [`ArrayOperation`](#ArrayOperation)
+
+
+## Syntax
+
+### *DocOperation*
+
+The *DocOperations* are used to generate elements of a `bson_t` document or
+subdocument thereof. The top-level `bsonBuildAppend()` accepts a list of
+*DocOperations*, as well as the `doc()` *ValueOperation*.
```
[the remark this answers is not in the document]
vector-of-bool: I worded this sentence very confusingly. I meant that `doc()` accepts the `DocOperations`. Fixed.

src/libbson/src/bson/bson-dsl.md
```diff
@@ -0,0 +1,993 @@
...
+Evaluate the given C expression, which must return a boolean-testable value. The
+code can refer to the element under consideration via `bsonVisitIter`. The code
+can test other predicates by using `bsonPredicate(P)`, which evaluates to a
+boolean value.
+
+
+# Error Handling
+
+The DSL uses two global (thread-local) names to perform error handling:
+`bsonBuildError` and `bsonParseError`.
```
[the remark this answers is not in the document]
vector-of-bool: I worried about message truncation, but I suppose it is better to be able to format than to generate lengthy messages.

src/libmongoc/tests/unified/entity-map.c
```diff
@@ -269,71 +271,118 @@ command_succeeded (const mongoc_apm_command_succeeded_t *succeeded)
...
+            })),
+      // Which events should be available as entities:
+      find (key ("storeEventsAsEntities"),
+            if (not(type (array)),
+                then (error ("'storeEventsAsEntities' must be an array"))),
+            visitEach (if (not(type (utf8)),
+                           then (error ("Every 'storeEventsAsEntities' element "
+                                        "must be a string")))),
+            /* TODO: CDRIVER-3867 Comprehensive Atlas Testing */
+            error ("'storeEventsAsEntities' is not yet supported")));
```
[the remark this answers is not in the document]
vector-of-bool: I was unable to formulate a way for `parse` to do this cleanly without requiring arbitrary allocations. `parse` doesn't keep track of which elements it has inspected. Because a document can have arbitrarily many elements, one would need a multiset to keep track of them. The `find(not(...))` pattern will work, even if it is less pretty.

Something similar could be done by using `case()` with the visit API instead:

```c
bsonVisitEach(doc,
  case(when(key("foo"), ...),
       when(key("bar"), ...),
       when(key("baz"), ...),
       else(error("Unrecognized key"))));
```

But: This form doesn't support a "required elements" semantics, which would have to be implemented manually.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -1736,71 +1738,62 @@ bool
...
+            error, "%s: %s [readPreference=%s]", prefix, bsonParseError, pref);
+      } else {
+         MONGOC_URI_ERROR (error, "%s: %s", prefix, bsonParseError);
       }
+      return false;
+   }
+
+   if (mode != 0) {
+      mongoc_read_prefs_set_mode (uri->read_prefs, mode);
    }
```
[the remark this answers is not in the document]
vector-of-bool: In the prior code: If "readPreference" was not found, we do not assign a read preference. This conditional just implements that logic.

src/libmongoc/tests/unified/entity-map.c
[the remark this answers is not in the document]
vector-of-bool: I was able to implement this in an unlikely-to-allocate manner without any key tracking required. `parse` now keeps an array of bits, with each bit corresponding to an element. When `find` or `require` visits an element, it sets a one bit in the array for the element. The `visitOthers` op iterates over the document again and applies a visitor to every element that has a zero bit. This will only allocate for very large documents that require many bits. (I also simplified the implementation of `find` and `require` significantly.)

src/libbson/src/bson/bson-dsl.md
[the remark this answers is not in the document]
kevinAlbs: > The intuitive API bson_steal marks the destination as read-only for some reason?

That is a rough edge. `bson_steal` sets the destination `bson_t` flag `BSON_FLAG_STATIC`. `BSON_FLAG_STATIC` means the `bson_t` struct is not freed on `bson_destroy`. It is independent of the underlying data.

The caveat is `bson_steal` cannot be used on a `bson_t` that has been heap allocated. E.g.

```
bson_t *b = bson_new ();
bson_steal (b, src); // leaks the bson_t struct
```

src/libmongoc/src/mongoc/mongoc-aggregate.c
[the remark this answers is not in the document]
kevinAlbs: Yes. I think the caller should be fixed instead.

src/libmongoc/src/mongoc/mongoc-uri.c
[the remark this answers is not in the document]
kevinAlbs: I misread.

kevinAlbs (approved):

src/libbson/src/bson/bson-dsl.h
```diff
@@ -0,0 +1,1336 @@
...
+      } else {
+         return bson_strcasecmp (string, s) == 0;
+      }
+   }
+   return false;
+}
+
+static BSON_INLINE bool
+_bson_dsl_key_is_anyof (const char *key,
+                        const int keylen,
```
eramongodb, line 1026:
```suggestion
                        const size_t keylen,
```
`keylen` should have type `size_t` or `uint32_t` (as returned by `bson_iter_key_len`).

src/libbson/src/bson/bson-dsl.h
```diff
@@ -0,0 +1,1336 @@
...
+
+/// A pointer to the current thread's bsonBuild context
+_bson_thread_local_comdat struct _bsonBuildContext_t
+   *_bsonBuildContextThreadLocalPtr = NULL;
+
+struct _bsonVisitContext_t {
+   const bson_t *doc;
+   bson_iter_t iter;
+   const struct _bsonVisitContext_t *parent;
+   int index;
```
eramongodb, line 973:
```suggestion
   size_t index;
```
`_bsonVisitContext_t::index` should have type `size_t`.

src/libbson/src/bson/bson-dsl.h
```diff
@@ -0,0 +1,1336 @@
...
+      struct _bsonVisitContext_t _bpCtx = {                                    \
+         .doc = &(Doc),                                                        \
+         .parent = &bsonVisitContext,                                          \
+      };                                                                       \
+      _bsonVisitContextThreadLocalPtr = &_bpCtx;                               \
+      /* Keep track of which elements have been visited based on their index*/ \
+      uint64_t _bpVisitBits_static[4] = {0};                                   \
+      BSON_MAYBE_UNUSED uint64_t *_bpVisitBits = _bpVisitBits_static;          \
+      BSON_MAYBE_UNUSED int _bpNumVisitBitInts =                               \
+         sizeof _bpVisitBits_static / sizeof (uint64_t);                       \
```
eramongodb, line 663:
```suggestion
      BSON_MAYBE_UNUSED size_t _bpNumVisitBitInts =                            \
         sizeof (_bpVisitBits_static) / sizeof (*_bpVisitBits_static);         \
```
`_bpNumVisitBitInts` should have type `size_t`.

src/libbson/src/bson/bson-dsl.h
```diff
@@ -0,0 +1,1336 @@
...
+   do {                                      \
+      if (!_bvHalt && !bsonParseError) {     \
+         _bsonParseOperation_##P;            \
+      }                                      \
+   } while (0);
+
+#define _bsonParseMarkVisited(Index)                                   \
+   if (1) {                                                            \
+      const int nth_int = Index / 64;                                  \
+      const int nth_bit = Index % 64;                                  \
```
eramongodb, line 688:
```suggestion
      const size_t nth_int = Index / 64u;                              \
      const size_t nth_bit = Index % 64u;                              \
```
Type consistency with `_bsonVisitContext_t::index`.

src/libbson/src/bson/bson-dsl.h
```diff
@@ -0,0 +1,1336 @@
...
+      while (nth_int >= _bpNumVisitBitInts) {                          \
+         /* Say that five times, fast: */                              \
+         int new_num_visit_bit_ints = _bpNumVisitBitInts * 2;          \
```
eramongodb, line 691:
```suggestion
         size_t new_num_visit_bit_ints = _bpNumVisitBitInts * 2u;      \
```
Type consistency with `_bsonVisitContext_t::index`.

src/libbson/src/bson/bson-dsl.h
```diff
@@ -0,0 +1,1336 @@
...
+                 _bpVisitBits,                                         \
+                 sizeof (uint64_t) * _bpNumVisitBitInts);              \
+         if (_bpVisitBits != _bpVisitBits_static) {                    \
+            bson_free (_bpVisitBits);                                  \
+         }                                                             \
+         _bpVisitBits = new_visit_bit_ints;                            \
+         _bpNumVisitBitInts = new_num_visit_bit_ints;                  \
+      }                                                                \
+                                                                       \
+      _bpVisitBits[nth_int] |= (1 << nth_bit);                         \
```
eramongodb, line 704:
```suggestion
      _bpVisitBits[nth_int] |= (UINT64_C (1) << nth_bit);              \
```
Type consistency with `_bpVisitbits[nth_int]`.

src/libbson/src/bson/bson-dsl.h
```diff
@@ -0,0 +1,1336 @@
...
+   } else                                                              \
+      ((void) 0)
+
+#define _bsonParseDidVisitNth(Index) \
+   _bsonParseDidVisitNth_1 (Index / 64, Index % 64)
+#define _bsonParseDidVisitNth_1(NthInt, NthBit) \
+   (NthInt < _bpNumVisitBitInts && (_bpVisitBits[NthInt] & (1 << NthBit)))
```
eramongodb, line 711:
```suggestion
#define _bsonParseDidVisitNth(Index) \
   _bsonParseDidVisitNth_1 (Index / 64u, Index % 64u)
#define _bsonParseDidVisitNth_1(NthInt, NthBit) \
   (NthInt < _bpNumVisitBitInts &&              \
    (_bpVisitBits[NthInt] & (UINT64_C (1) << NthBit)))
```
Type consistency with `_bsonVisitContext_t::index` and `_bpVisitbits[NthBit]`.

src/libbson/src/bson/bson-dsl.h
```diff
@@ -0,0 +1,1336 @@
...
+
+/**
+ * Finally, the Map() macro that allows us to do the magic, which we've been
+ * building up to all along.
+ *
+ * The dance with mapMacro_first, mapMacro_final, and _bsonDSL_nothing
+ * conditional on argument count is to prevent warnings from pre-C99 about
+ * passing no arguments to the '...' parameters. Yet again, if we had C99 and
+ * __VA_OPT__ this would be simpler.
+ */
```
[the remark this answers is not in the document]
vector-of-bool: MSVC didn't get a non-broken C99 preprocessor until very recently. We can't use it until we drop older MSVC versions.

src/libbson/src/bson/bson-dsl.h
```diff
@@ -0,0 +1,1336 @@
...
+#define _bson_thread_local_comdat __declspec(thread) __declspec(selectany)
+#else
+#define _bson_thread_local_comdat __attribute__ ((weak)) __thread
+#endif
+
+#ifdef __GNUC__
+// GCC has a bug handling pragma statements that disable warnings within complex
+// nested macro expansions. If we're GCC, just disable -Wshadow outright:
+BSON_IF_GNU_LIKE (_Pragma ("GCC diagnostic ignored \"-Wshadow\""))
+#endif
```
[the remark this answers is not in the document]
vector-of-bool: I suppose no code in this PR exhibits it, but nested visit/parse/build operations will shadow the local variables of the outer operations.

src/libbson/src/bson/bson-dsl.h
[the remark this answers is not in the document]
eramongodb: 😢

src/libbson/src/bson/bson-dsl.h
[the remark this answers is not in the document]
eramongodb: I'm not sure this is the case. I am not able to coerce `-Wshadow` warnings given:
```c
bsonParse (*TMP_BSON_FROM_JSON ({}),
           do(bsonParse (*TMP_BSON_FROM_JSON ({}), do());));
```
or:
```c
bsonParse (*TMP_BSON_FROM_JSON ({}),
           find (key ("foo"),
                 do(bsonParse (*TMP_BSON_FROM_JSON ({}),
                               find (key ("foo"), do())))));
```
Although the latter did trigger a `-Woverlength-strings` warning. 😏

src/libmongoc/tests/unified/entity-map.c
```diff
@@ -269,71 +271,125 @@ command_succeeded (const mongoc_apm_command_succeeded_t *succeeded)
...
+            visitEach (if (not(type (utf8)),
+                           then (error ("Every 'storeEventsAsEntities' element "
+                                        "must be a string")))),
+            /* TODO: CDRIVER-3867 Comprehensive Atlas Testing */
+            error ("'storeEventsAsEntities' is not yet supported")),
+      visitOthers (dupPath (errpath),
+                   errorf (err,
+                           "At [%s]: Unknown key '%s' given in entity options",
+                           errpath,
+                           bson_iter_key (&bsonVisitIter))));
```
eramongodb, line 389:
```suggestion
      visitOthers (dupPath (errpath),
                   errorf (err,
                           "At [%s]: Unknown key '%s' given in entity options",
                           errpath,
                           bson_iter_key (&bsonVisitIter)),
                   do(bson_free (errpath))));
```
This suggestion is also a question: given `dupPath` allocates and responsibility for freeing is on caller, is the addition of this `do(bson_free (errpath))` appropriate? (Besides the fact that this test unconditionally aborts if this path is taken.)

src/libbson/src/bson/bson-dsl.h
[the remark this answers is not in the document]
vector-of-bool: I've definitely seen it in my system. Further investigation may be warranted.

src/libbson/src/bson/bson-dsl.md
```diff
@@ -0,0 +1,1029 @@
...
+
+#### `truthy`
+
+Matches is the element is "truthy": A `true` boolean, a non-zero numeric type,
+or other non-null value.
+
+
+#### `falsey`
+
+Matches any element that is not `truthy`
```
eramongodb, line 685:
```suggestion
#### `is_true`

Matches if the element is `true` according to `bson_iter_as_bool`.


#### `is_false`

Matches if the element is `false` according to `bson_iter_as_bool`.
```

Suggest renaming `truthy`/`falsey` to `is_true` and `is_false` and adjusting the description to be defined in terms of `bson_iter_as_bool`.

src/libmongoc/src/mongoc/mongoc-cursor.c
```diff
@@ -790,16 +790,14 @@ _mongoc_cursor_monitor_succeeded (mongoc_cursor_t *cursor,
...
-                      first_batch ? 10 : 9,
-                      &docs_array);
-   bson_append_document_end (&reply, &reply_cursor);
+   bsonBuildDecl (reply,
+                  kv ("ok", int32 (1)),
+                  kv ("cursor",
+                      doc (kv ("id", int64 (mongoc_cursor_get_id (cursor))),
+                           kv ("ns", utf8_w_len (cursor->ns, cursor->nslen)),
+                           kv (first_batch ? "firstBatch" : "nextBatch",
+                               bsonArray (docs_array)))));
```
eramongodb, line 799:
```suggestion
   bsonBuildDecl (
      reply,
      kvl ("ok", 2, int32 (1)),
      kvl ("cursor",
           6,
           doc (kvl ("id", 2, int64 (mongoc_cursor_get_id (cursor))),
                kvl ("ns", 2, utf8_w_len (cursor->ns, cursor->nslen)),
                kvl (first_batch ? "firstBatch" : "nextBatch",
                     first_batch ? 10 : 9,
                     bsonArray (docs_array)))));
```
Suggest preserving lengths to avoid calls to `strlen`.

src/libmongoc/src/mongoc/mongoc-cursor.c
```diff
@@ -845,8 +842,7 @@ _mongoc_cursor_monitor_failed (mongoc_cursor_t *cursor,
    /* we sent OP_QUERY/OP_GETMORE, fake a reply to find/getMore command:
     * {ok: 0}
     */
-   bson_init (&reply);
-   bson_append_int32 (&reply, "ok", 2, 0);
+   bsonBuildDecl (reply, kv ("ok", int32 (0)));
```
eramongodb, line 845:
```suggestion
   bsonBuildDecl (reply, kvl ("ok", 2, int32 (0)));
```
Suggest preserving length to avoid call to `strlen`.

src/libmongoc/src/mongoc/mongoc-collection.c
```diff
@@ -750,27 +752,22 @@ mongoc_collection_count_with_opts (
...
+   bsonBuildAppend (
+      cmd,
+      kv ("count",
+          utf8_w_len (collection->collection, collection->collectionlen)),
+      kv ("query",
+          if (query,                // If we have a query,
+              then (bson (*query)), // Copy it
+              else(doc ()))),       // Otherwise, add an empty doc
+      if (limit, then (kv ("limit", int64 (limit)))),
+      if (skip, then (kv ("skip", int64 (skip)))));
```
eramongodb, line 769:
```suggestion
   bsonBuildAppend (
      cmd,
      kvl ("count",
           5,
           utf8_w_len (collection->collection, collection->collectionlen)),
      kvl ("query",
           5,
           if (query,                // If we have a query,
               then (bson (*query)), // Copy it
               else(doc ()))),       // Otherwise, add an empty doc
      if (limit, then (kvl ("limit", 5, int64 (limit)))),
      if (skip, then (kvl ("skip", 4, int64 (skip)))));
```
Suggest preserving lengths to avoid calls to `strlen`.

src/libmongoc/src/mongoc/mongoc-aggregate.c
```diff
@@ -92,53 +93,81 @@ _make_agg_cmd (const char *ns,
...
    }
 
-   has_write_key = _has_write_key (&has_write_key_iter);
-   bson_append_document_begin (command, "cursor", 6, &child);
-   /* Ignore batchSize=0 for aggregates with $out or $merge */
-   if (opts->batchSize_is_set && !(has_write_key && opts->batchSize == 0)) {
-      BSON_APPEND_INT32 (&child, "batchSize", opts->batchSize);
+   bsonBuildAppend (
+      *command,
+      kv ("cursor",
```
eramongodb, line 150:
```suggestion
      kvl ("cursor",
           6,
```
Suggest preserving length to avoid call to `strlen`.

src/libmongoc/tests/unified/entity-map.c
vector-of-bool: It is safe to add the free(). I only omitted it because the program definitely terminates on this condition.

src/libmongoc/src/mongoc/mongoc-cursor.c
vector-of-bool: The calls to strlen() will be (_should be_) constant-folded (by any competent optimizer) because this macro-expands to a call to strlen() of a string literal.

src/libbson/src/bson/bson-dsl.h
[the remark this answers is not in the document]
vector-of-bool: Solved: The shadow warnings appear on Clang (at least, the version I am using), but not on GCC.

src/libmongoc/tests/unified/entity-map.c
```diff
@@ -269,71 +271,125 @@ command_succeeded (const mongoc_apm_command_succeeded_t *succeeded)
...
+               if (not(type (bool)),
+                   then (error ("serverApi.deprecationErrors must be a bool"))),
+               do(mongoc_server_api_deprecation_errors (api, bsonAs (bool)))))),
+      // Toggle observation of sensitive commands
+      find (key ("observeSensitiveCommands"),
+            if (not(type (bool)),
+                then (error ("'observeSensitiveCommands' must be a bool"))),
+            do({
+               bool *p = entity->observe_sensitive_commands =
+                  BSON_ALIGNED_ALLOC0 (bool);
```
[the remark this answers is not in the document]
vector-of-bool: Because it accepts the type parameter `bool`. No other reason.

src/libbson/src/bson/bson-dsl.md
vector-of-bool: Renamed with `isTrue` and `isFalse` (to match the casing of everything else)

src/libbson/src/bson/bson-dsl.h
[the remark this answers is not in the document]
eramongodb: I still do not observe it in my tests, but I do not think it is serious enough to be a blocker.

src/libbson/src/bson/bson-dsl.h
```diff
@@ -0,0 +1,1378 @@
...
+
+#define _bsonDSL_disableWarnings()                                        \
+   if (1) {                                                               \
+      BSON_IF_GNU_LIKE (_Pragma ("GCC diagnostic push");)                 \
+      BSON_IF_GNU_LIKE (_Pragma ("GCC diagnostic ignored \"-Wshadow\"");) \
+   } else                                                                 \
+      ((void) 0)
+
+#define _bsonDSL_restoreWarnings() \
+   BSON_IF_GNU_LIKE (_Pragma ("GCC diagnostic pop");)
```
eramongodb, line 44:
```suggestion
#define _bsonDSL_restoreWarnings()                       \
   if (1) {                                              \
      BSON_IF_GNU_LIKE (_Pragma ("GCC diagnostic pop");) \
   } else                                                \
      ((void) 0)
```

src/libmongoc/tests/unified/entity-map.c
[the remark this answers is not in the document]
eramongodb: I think I would prefer avoiding use of aligned alloc except when necessary (asking for more than what is required), but not strongly against its usage.

src/libmongoc/src/mongoc/mongoc-cursor.c
eramongodb: Resolving in response to https://github.com/mongodb/mongo-c-driver/pull/1094#discussion_r1004799519.

src/libmongoc/src/mongoc/mongoc-collection.c
eramongodb: Resolving in response to https://github.com/mongodb/mongo-c-driver/pull/1094#discussion_r1004799519.

src/libmongoc/src/mongoc/mongoc-aggregate.c
eramongodb: Resolving in response to https://github.com/mongodb/mongo-c-driver/pull/1094#discussion_r1004799519.

eramongodb (approved):

PR #1097, opened 2022-08-26
[CDRIVER-4454] Automatic Azure KMS Credentials

vector-of-bool:
Refer: CDRIVER-4454

This implements automatic KMS credentials for Azure, in the same manner as are available for AWS.

To optimize for test-ability in the absence of an Azure environment, the following changes are made:

1. The private HTTP API was refactored to allow inspection of the HTTP request string that _would be_ sent. This way we can build a function that constructs an HTTP request, and then check it against a known-good request message, no HTTP server needed. This alleviates the necessity to have a "mock server" for the API, as we can assume that the proper server, when presented with such a known-good request, will respond as we expect it to. This assumes that the internal HTTP code sends/receives HTTP correctly, but that proposition can be validated and tested without implementing a fake Azure IMDS API server.
2. The Azure code is implemented as 1) a request-generator that fills in an HTTP request object, and 2) A JSON-response parser that inspects the HTTP JSON body and decomposes it according to the Azure schema. Because (2) does not perform actual I/O, we can simply feed it known-good and known-bad JSON strings and validate that it behaves appropriately, and therefore assume that it will behave correctly later on when fed the actual HTTP responses from an IMDS server.


vector-of-bool:
The HTTP retry-with-backoff logic has been moved to a separate function that doesn't concern itself with the request content. This allows easier informal verification of both the retry logic as well as code that invokes that that function.

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -628,6 +630,32 @@ _needs_on_demand_aws_kms (bson_t const *kms_providers, bson_error_t *error)
...
+ * @return false Otherwise. Check error->code for failure.
+ */
+static bool
+_check_azure_kms_auto (const bson_t *kmsprov, bson_error_t *error)
+{
+   bson_iter_t iter;
+   const bool has_azure = bson_iter_init_find (&iter, kmsprov, "azure");
+   if (!has_azure) {
+      return false;
+   }
```
eramongodb, line 649:
```suggestion
   if (!bson_iter_init_find (&iter, kmsprov, "azure")) {
      return false;
   }
```

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -628,6 +630,32 @@ _needs_on_demand_aws_kms (bson_t const *kms_providers, bson_error_t *error)
...
 }
 
+/**
+ * @brief Check whether the given kmsProviders object requests automatic Azure
+ * credentials, and what the managedIdentity parameters might be
+ *
+ * @param kmsprov The input kmsProviders that may have an "azure" property
+ * @param error An output error
+ * @return true If success AND `kmsprov` requests automatic Azure credentials
+ * @return false Otherwise. Check error->code for failure.
```
eramongodb, line 640:
```suggestion
 * @return true on success and `kmsprov` requests automatic Azure credentials.
 * False otherwise. Check error->code for failure.
```

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -666,6 +694,72 @@ _try_add_aws_from_env (bson_t *out, bson_error_t *error)
...
 }
 
+/**
+ * @brief Attempt to load an Azure access token from the environment and append
+ * them to the kmsProviders
+ *
+ * @param out A kmsProviders object to update
+ * @param error An error-out parameter
+ * @return true If there was no error and we loaded credentials
+ * @return false If there was an error obtaining or appending credentials
```
eramongodb, line 704:
```suggestion
 * @return true if there was no error and we loaded credentials. False if there
 * was an error obtaining or appending credentials.
```

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -666,6 +694,72 @@ _try_add_aws_from_env (bson_t *out, bson_error_t *error)
...
+
+   okay = true;
+
+bson_build_failed:
+   mcd_azure_access_token_destroy (&tok);
+token_parse_failed:
+   bson_destroy (&new_azure_creds);
+req_failed:
+   _mongoc_http_response_cleanup (&resp);
+uri_failed:
```
eramongodb, line 758:
```suggestion
```
No reference to `uri_failed` label.

src/libmongoc/src/mongoc/mcd-azure.h
```diff
@@ -0,0 +1,104 @@
...
+   MCD_AZURE_HTTP_NO_FLAGS = 0,
+   // Retry a request if we receive an HTTP 404 response.
+   MCD_AZURE_RETRY_ON_404 = 1 << 0
+};
+
+/**
+ * @brief Send an Azure HTTP request with automatic retries.
+ *
+ * @param req The request to send.
+ * @param resp The output parameter for the response
```
eramongodb, line 85: Does `resp` need to be destroyed even on failure? If so, `_mongoc_http_response_cleanup` should document this requirement.

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -696,6 +790,27 @@ _state_need_kms_credentials (_state_machine_t *sm, bson_error_t *error)
    const bool callback_provided_aws =
       bson_iter_init_find (&iter, &creds, "aws");
 
+   // Whether the callback requested auto-Azure credentials
```
eramongodb, line 793: May want to shift this below the `if (... && _needs_on_demand_aws_kms (...))` block to avoid breaking up the AWS-related code + add block scopes per provider.

src/libmongoc/src/mongoc/mcd-azure.h
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * @param out The token to initialize. Should be uninitialized.
+ * @param json The JSON string body
+ * @param len The length of 'body'
+ * @param error An output parameter for errors
+ * @return true If 'out' was successfully initialized to a token.
+ * @return false Otherwise
+ *
+ * @note The 'out' token must later be given to @ref
+ * mcd_azure_acecss_token_destroy
```
eramongodb, line 31:
```suggestion
 * mcd_azure_access_token_destroy
```

src/libmongoc/src/mongoc/mcd-azure.c
```diff
@@ -0,0 +1,167 @@
...
+
+#define AZURE_API_VERSION "2018-02-01"
+
+static const char *const DEFAULT_METADATA_PATH =
+   "/metadata/identity/oauth2/"
+   "token?api-version=" AZURE_API_VERSION
+   "&resource=https%3A%2F%2Fvault.azure.net";
+
+void
+mcd_azure_imds_request_init (mcd_azure_imds_request *req)
```
eramongodb, line 13: Missing null check for `req` parameter.

src/libmongoc/src/mongoc/mcd-azure.c
```diff
@@ -0,0 +1,167 @@
...
+   // 'Metadata: true' is required
+   req->req.extra_headers = "Metadata: true\r\n"
+                            "Accept: application/json\r\n";
+   // The default path is suitable. In the future, we may want to add query
+   // parameters to disambiguate a managed identity.
+   req->req.path = bson_strdup (DEFAULT_METADATA_PATH);
+}
+
+void
+mcd_azure_imds_request_destroy (mcd_azure_imds_request *req)
```
eramongodb, line 31: Missing null check for `req` parameter.

src/libmongoc/src/mongoc/mcd-azure.c
```diff
@@ -0,0 +1,167 @@
...
+{
+   bson_free ((void *) req->req.path);
+   *req = (mcd_azure_imds_request){0};
+}
+
+bool
+mcd_azure_access_token_try_init_from_json_str (mcd_azure_access_token *out,
+                                               const char *json,
```
eramongodb, line 39: Missing null checks for `out` and `json` parameters.

eramongodb (changes requested):

src/libmongoc/src/mongoc/mcd-azure.c
```diff
@@ -0,0 +1,98 @@
...
+   const char *const resource = !found ? NULL : bson_iter_utf8 (&iter, NULL);
+   // token_type
+   found = bson_iter_init_find (&iter, &bson, "token_type");
+   const char *const token_type = !found ? NULL : bson_iter_utf8 (&iter, NULL);
+
+   if (!(access_token && resource && token_type)) {
+      bson_set_error (
+         error,
+         MONGOC_ERROR_PROTOCOL_ERROR,
+         64,
```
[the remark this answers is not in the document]
kevinAlbs: I think the error codes in each domain can have a different meaning. Two domains can have the same error code. 

Consider adding an error code to `mongoc_error_code_t`, like `MONGOC_ERROR_PROTOCOL_HTTP` to signal an error in an HTTP response.

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -696,6 +790,27 @@ _state_need_kms_credentials (_state_machine_t *sm, bson_error_t *error)
...
+   }
+   // Whether the original kmsProviders requested auto-Azure credentials
+   const bool orig_wants_auto_azure =
+      !cb_wants_auto_azure &&
+      _check_azure_kms_auto (&sm->crypt->kms_providers, error);
+   if (error->code) {
+      // _check_azure_kms_auto failed
+      goto fail;
+   }
+   const bool wants_auto_azure = cb_wants_auto_azure || orig_wants_auto_azure;
```
kevinAlbs, line 807: Checking `cb_wants_auto_azure` differs from the `aws` logic below. If the callback provides `"aws": {}` that will not attempt to add AWS credentials from the environment.

I do not see a clear advantage to either approach. Consider making the logic consistent, or adding a comment to note the inconsistency.

vector-of-bool: The callback function in this situation is unfortunately under-specified. I think it will be better to go with the AWS definition for now.

src/libmongoc/src/mongoc/mongoc-http.c
```diff
@@ -44,8 +44,51 @@ _mongoc_http_response_cleanup (mongoc_http_response_t *response)
...
+   }
+
+   // Final terminator
+   bson_string_append (string, "\r\n");
+   return string;
+}
+
 bool
-_mongoc_http_send (mongoc_http_request_t *req,
+_mongoc_http_send (mongoc_http_request_t const *req,
```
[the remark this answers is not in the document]
vector-of-bool: "East const" versus "west const" is a decision I make depending on whether I realize a parameter should be const before or after I have already typed the base type name.

src/libmongoc/tests/test-mcd-azure-imds.c
```diff
@@ -0,0 +1,68 @@
+#include <mongoc/mcd-azure.h>
+
+#include "TestSuite.h"
+
+#define STRING(...) #__VA_ARGS__
```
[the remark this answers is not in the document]
vector-of-bool: This macro will of course break the moment you feed in any invalid or un-balanced preprocessor tokens 🙃. It just happens that JSON is coincidentally is a valid balanced pp-token sequence.

src/libmongoc/tests/test-libmongoc.c
```diff
@@ -2923,131 +2679,140 @@ main (int argc, char *argv[])
...
-   test_value_install (&suite);
-   test_writer_install (&suite);
-   test_b64_install (&suite);
-   test_bson_cmp_install (&suite);
+#define TEST_INSTALL(FuncName)                 \
+   if (1) {                                    \
+      extern void FuncName (TestSuite *suite); \
+      FuncName (&suite);                       \
+   } else                                      \
+      ((void) 0)
```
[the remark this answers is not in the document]
vector-of-bool: You can blame this one on Jens Gustedt for converting me [he One True Way™️](https://gustedt.wordpress.com/2011/02/02/handling-control-flow-inside-macros). It doesn't matter here, but I've made a habit of preferring this shape.

src/libmongoc/src/mongoc/mcd-azure.c
```diff
@@ -0,0 +1,167 @@
...
+   bson_free (c->access_token);
+   bson_free (c->resource);
+   bson_free (c->token_type);
+}
+
+bool
+mcd_azure_send_request_with_retries (const mongoc_http_request_t *req,
+                                     mongoc_http_response_t *resp,
+                                     enum mcd_azure_http_flags flags,
+                                     bson_error_t *error)
```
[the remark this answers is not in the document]
vector-of-bool: The retry logic is unfortunately ambiguously scoped. One _could_ say "send the request, retrying on these conditions, use exponential +2s backoff <=30s", and just assume that the correctness of that logic is outside of the scope of the document (and thus should be tested separately without being "part of" the Azure KMS logic).

We can't even be sure how common the retry would be needed. I only added it at the recommendation of the API docs.

It could also be specified as a "hint" recommendation, i.e. "drivers MAY add additional retry logic for the HTTP request."

src/libmongoc/src/mongoc/mcd-azure.h
```diff
@@ -0,0 +1,104 @@
...
+ * @param c The access token to destroy
+ */
+void
+mcd_azure_access_token_destroy (mcd_azure_access_token *token);
+
+/**
+ * @brief An Azure IMDS HTTP request
+ *
+ */
+typedef struct mcd_azure_imds_request {
```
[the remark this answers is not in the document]
vector-of-bool: Mostly as a strongly-type communication channel between the Azure wrapper APIs, to ensure additional invariants hold on this request versus a bare `mongoc_http_request_t`, and to discourage callers from manipulating it directly. There _was_ more to it in a previous iteration, but the scope of the changeset has been scaled back.

src/libmongoc/src/mongoc/mongoc-crypt.c
vector-of-bool: This documentation shape was stamped out by Doxygen generator I use, but it looks to be goofing here. It _should_ be using the `@retval` command, which consumes the following token as the return value, and the remaining line as a description corresponding to that value.

src/libmongoc/src/mongoc/mcd-azure.c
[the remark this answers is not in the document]
kevinAlbs: > The retry logic is unfortunately ambiguously scoped. One _could_ say "send the request, retrying on these conditions, use exponential +2s backoff <=30s", and just assume that the correctness of that logic is outside of the scope of the document (and thus should be tested separately without being "part of" the Azure KMS logic).

IMO specification changes should include guidance for testing. Otherwise it adds significant work to downstream teams to determine how to test.

> We can't even be sure how common the retry would be needed. I only added it at the recommendation of the API docs.

Good point. If there are no specification or prose tests, I say remove the retry logic. I expect adding any specification tests may be significant work (e.g. add a mock server). And at this point we do not know if there is any benefit.

> It could also be specified as a "hint" recommendation, i.e. "drivers MAY add additional retry logic for the HTTP request."

I think the general preference in the specifications is consistency between driver implementations (e.g. the [Strive to be idiomatic, but favor consistency](https://github.com/mongodb/specifications/tree/51741620a6af40337b8ec9dcb6fe44f8586f07f6#strive-to-be-idiomatic-but-favor-consistency) and [When in doubt, use "MUST" not "SHOULD](https://github.com/mongodb/specifications/tree/51741620a6af40337b8ec9dcb6fe44f8586f07f6#when-in-doubt-use-must-not-should-in-specs) mantras.

Rather than making it optional, I suggest pick one way (either with retry logic and tests, or no retry logic) and document the Rationale.

kevinAlbs (approved):
Nice, LGTM!

src/libmongoc/src/mongoc/CMakeLists.txt
```diff
@@ -74,6 +74,7 @@ extra_dist_generated (
 )
 
 set (src_libmongoc_src_mongoc_DIST_noinst_hs
+   mcd-azure.h
```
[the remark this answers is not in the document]
vector-of-bool: I would prefer the latter, if only because my tools better understand that the `.c` and `.h` files are related 😃 so I can automatically jump between them with a hotkey.

In fullness, I'd prefer private headers to be namespaced with a subdirectory, but that may require additional CMake caches.

vector-of-bool:
The three added commits can be viewed in isolation in commit order to build off of each other.

src/libmongoc/src/mongoc/mcd-azure.c
```diff
@@ -0,0 +1,116 @@
...
+      // Set the output, duplicate each string
+      *out = (mcd_azure_access_token){
+         .access_token = bson_strdup (access_token),
+         .resource = bson_strdup (resource),
+         .token_type = bson_strdup (token_type),
+      };
+      // "expires_in" encodes the number of seconds since the issue time for
+      // which the token will be valid. strtoll() will saturate on range errors
+      // and return zero on parse errors.
+      long long s = strtoll (expires_in_str, NULL, 0);
```
eramongodb, line 97: No parse failure handling (return `false`)? Immediate expiration on parse failure (expires in 0 seconds) does not seem ideal.

src/libmongoc/src/mongoc/mcd-azure.c
```diff
@@ -71,8 +71,12 @@ mcd_azure_access_token_try_init_from_json_str (mcd_azure_access_token *out,
    // token_type
    found = bson_iter_init_find (&iter, &bson, "token_type");
    const char *const token_type = !found ? NULL : bson_iter_utf8 (&iter, NULL);
+   // expires_on
```
kevinAlbs, line 74: "expires_in"

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -666,6 +702,111 @@ _try_add_aws_from_env (bson_t *out, bson_error_t *error)
...
+         return false;
+      }
+   }
+
+   // Build the new KMS credentials
+   bson_t new_azure_creds = BSON_INITIALIZER;
+   const bool okay =
+      (!BSON_APPEND_UTF8 (
+          &new_azure_creds, "accessToken", crypt->azure_token.access_token) ||
+       !BSON_APPEND_DOCUMENT (out, "azure", &new_azure_creds));
```
eramongodb, line 798:
```suggestion
   const bool okay = BSON_APPEND_UTF8 (&new_azure_creds,
                                       "accessToken",
                                       crypt->azure_token.access_token) &&
                     BSON_APPEND_DOCUMENT (out, "azure", &new_azure_creds);
```

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -696,26 +703,24 @@ _try_add_aws_from_env (bson_t *out, bson_error_t *error)
...
- * @retval true If there was no error and we loaded credentials
- * @retval false If there was an error obtaining or appending credentials
+ * @param out The token to populate. Must later be destroyed by the caller.
+ * @param error An output parameter to capture any errors
+ * @retval true Upon successfully obtaining and parsing a token
+ * @retval false If any error occurs.
  */
 static bool
-_try_add_azure_from_env (bson_t *out, bson_error_t *error)
+_request_new_azure_mid_token (mcd_azure_access_token *out, bson_error_t *error)
```
kevinAlbs, line 714: What does "mid" refer to?

src/libmongoc/src/mongoc/mcd-time.h
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/**
+ * @brief Compare two time points to create an ordering.
+ *
+ * A time point "in the past" is "less than" a time point "in the future".
+ *
+ * @retval -1 If 'left' is before 'right'
+ * @retval +1 If 'right' is before 'left'
+ * @retval  0 If 'left' and 'right' are equivalent
```
eramongodb, line 119: Why guarantee exact values of `-1` and `1` instead of defining as `< 0` and `> 0`?

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -727,36 +732,78 @@ _try_add_azure_from_env (bson_t *out, bson_error_t *error)
...
+   if (crypt->azure_token.access_token) {
+      // The access-token is non-null, so we may have one cached.
+      mcd_time_point one_min_from_now = mcd_later (mcd_now (), mcd_minutes (1));
+      mcd_time_point expires_at = mcd_later (crypt->azure_token_issued_at,
+                                             crypt->azure_token.expires_in);
+      if (mcd_time_compare (expires_at, one_min_from_now) >= 0) {
+         // The token is still valid for another minute
+      } else {
+         // The token is about to expire. Destroy it. We will then ask IMDS for
+         // a new oen.
```
kevinAlbs, line 773: "one"

kevinAlbs (approved):

src/libmongoc/src/mongoc/mongoc-crypt.c
vector-of-bool: "managed identity" in this case. I can change that if it is unclear.

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -727,36 +732,78 @@ _try_add_azure_from_env (bson_t *out, bson_error_t *error)
...
+
+   if (crypt->azure_token.access_token == NULL) {
+      // There is no access token in our cache.
+      // Save the current time point as the "issue time" of the token, even
+      // though it will take some time for the HTTP request to hit the metadata
+      // server. This time is only used to track token expiry. IMDS gives us a
+      // number of seconds that the token will be valid relative to its issue
+      // time. Avoid reliance on system clocks by comparing the issue time to an
+      // abstract monotonic "now"
+      crypt->azure_token_issued_at = mcd_now ();
```
eramongodb, line 786: Can call to `mcd_now()` be moved into `_request_new_azure_mid_token`?

src/libmongoc/src/mongoc/mongoc-crypt.c
kevinAlbs: Since this is the only reference to "mid", I suggest renaming to `_request_new_azure_access_token` or defining "mid" in a comment.

src/libmongoc/src/mongoc/mcd-time.h
```diff
@@ -0,0 +1,138 @@
...
+ * Intuition: If "then" is "in the future" relative to "from", you will
+ * receive a positive duration, indicating an amount of time to wait beginning
+ * at 'from' to reach 'then'. If "then" is actually *before* "from", you will
+ * receive a paradoxical *negative* duration, indicating the amount of time
+ * needed to time-travel backwards to reach "then."
+ */
+inline static mcd_duration
+mcd_time_difference (mcd_time_point then, mcd_time_point from)
+{
+   int64_t diff = then._rep - from._rep;
```
eramongodb, line 107: Potential undefined behavior due to signed integer overflow. Suggest asserting against overflow conditions or conducting the subtraction safely.

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -628,14 +637,41 @@ _needs_on_demand_aws_kms (bson_t const *kms_providers, bson_error_t *error)
...
+ *
+ * @param kmsprov The input kmsProviders that may have an "azure" property
+ * @param error An output error
+ * @retval true If success AND `kmsprov` requests automatic Azure credentials
+ * @retval false Otherwise. Check error->code for failure.
+ */
+static bool
+_check_azure_kms_auto (const bson_t *kmsprov, bson_error_t *error)
+{
+   bson_set_error (error, 0, 0, " ");
```
eramongodb, line 652:
```suggestion
   *error = (bson_error_t){0};
```

src/libmongoc/src/mongoc/mcd-time.h
```diff
@@ -0,0 +1,138 @@
...
+   // Sixty seconds per minute:
+   return mcd_seconds (m * 60);
+}
+
+/// Obtain the time point relative to a base time as if by waiting for `delta`
+/// amount of time (which may be negatve)
+inline static mcd_time_point
+mcd_later (mcd_time_point from, mcd_duration delta)
+{
+   from._rep += delta._rep;
```
eramongodb, line 85: Potential undefined behavior due to signed integer overflow. Suggest asserting against overflow conditions or conducting the addition safely.

eramongodb (changes requested):

src/libmongoc/src/mongoc/mcd-time.h
```diff
@@ -0,0 +1,138 @@
...
+ * duration).
+ *
+ * Intuition: If "then" is "in the future" relative to "from", you will
+ * receive a positive duration, indicating an amount of time to wait beginning
+ * at 'from' to reach 'then'. If "then" is actually *before* "from", you will
+ * receive a paradoxical *negative* duration, indicating the amount of time
+ * needed to time-travel backwards to reach "then."
+ */
+inline static mcd_duration
+mcd_time_difference (mcd_time_point then, mcd_time_point from)
```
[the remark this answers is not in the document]
eramongodb: Disregard; may be more confusing given `difference` is behaving as `then - from`.

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -41,6 +44,12 @@ struct __mongoc_crypt_t {
    /// credentials.
    bson_t kms_providers;
    mc_kms_credentials_callback creds_cb;
+
+   /// The most recently auto-acquired Azure token, on null if it was destroyed
+   /// or not yet acquired.
+   mcd_azure_access_token azure_token;
+   /// The time point at which the `azure_token` was acquired.
+   mcd_time_point azure_token_issued_at;
```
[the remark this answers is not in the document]
vector-of-bool: It could, but you can't determine when a token was issued based only on the token json data (except for the issue time according to the server, which is a Unix timestamp, and would therefore be subject to local clock offsets.)

src/libmongoc/src/mongoc/mcd-azure.c
vector-of-bool: I considered it, but if the parse fails and returns zero, we will still use the token exactly once, since the expiry check occurs earlier in the process. It's unfortunately not trivial to determine whether a return of zero is because of an error, or because the number is actually zero.

src/libmongoc/src/mongoc/mcd-time.h
vector-of-bool: No specific reason, but it would be safe to weaken.

src/libmongoc/src/mongoc/mongoc-crypt.c
vector-of-bool: bson_set_error handles null pointers safely.

src/libmongoc/src/mongoc/mcd-azure.c
eramongodb: I think it is worth checking over deferring the instant-expiration check to the next use of the credentials, which I suspect will often occur very soon:
```c
char *str_end;
long long s = strtoll (expires_in_str, &str_end, 0);
if (
   // Should not be an empty string.
   expires_in_str == str_end ||
   // Should have consumed entire string.
   *str_end != '\0') {
   // Parse Failure.
}
```

src/libmongoc/src/mongoc/mongoc-crypt.c
eramongodb: Point. Should `" "` be `""`?
```suggestion
   bson_set_error (error, 0, 0, "");
```

vector-of-bool: You'd think so, but GCC warns that empty format strings are a non-standard C extension (??)

eramongodb: I suppose `bson_set_error (error, 0, 0, "%s", "");` would be a way to satisfy GCC, although I am not very happy about it.

Avoiding the string literal was what motivated my initial suggestion; a simple null check may suffice:
```c
   if (error) {
      *error = (bson_error_t){0};
   }
```
I'd prefer not use `bson_set_error` to "reset" an error object. I'll leave it to you to decide which is preferable.

src/libmongoc/src/mongoc/mcd-time.h
```diff
@@ -7,6 +7,62 @@
 
 #include <bson/bson.h>
 
+/// Return 'true' iff (left * right) would overflow with int64
```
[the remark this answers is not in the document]
eramongodb: Recommend adding test cases to ensure overflow checks behave as expected given their complexity.
e.g. `_mcd_i64_mul_would_overflow (INT64_MIN / 2 + 1, -2) -> true` is incorrect.

eramongodb (changes requested):

src/libmongoc/src/mongoc/mcd-time.h
```diff
@@ -7,6 +7,62 @@
...
+static inline bool
+_mcd_i64_mul_would_overflow (int64_t left, int64_t right)
```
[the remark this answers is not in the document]
vector-of-bool: Now I'm baffled where I got this bad definition from, because I am certain that I saw it somewhere...

src/libmongoc/tests/test-mcd-azure-imds.c
```diff
@@ -0,0 +1,69 @@
+#include <mongoc/mcd-azure.h>
+
+#include "TestSuite.h"
+
+#define RAW_STRING(...) #__VA_ARGS__
+
+void
```
eramongodb, line 7:
```suggestion
static void
```

src/libmongoc/tests/test-mcd-azure-imds.c
```diff
@@ -0,0 +1,69 @@
...
+      }),
+      -1,
+      &error));
+   ASSERT_ERROR_CONTAINS (error, 0, 0, "");
+   ASSERT_CMPSTR (token.access_token, "meow");
+
+   mcd_azure_access_token_destroy (&token);
+}
+
+void
```
eramongodb, line 42:
```suggestion
static void
```

src/libmongoc/src/mongoc/mcd-integer.h
```diff
@@ -0,0 +1,411 @@
...
+               if (left < min_fac) {
+                  /**
+                  Given: left         < min_fac
+                    and:        right < 0
+                   then: left * right > min_fac * right
+                    b/c: [Mult by negative flips ordering]
+                   with: MIN          = min_fac * right
+                   THEN: left * right > MIN
+                  */
+                  return false;
```
eramongodb, line 183: Strongly recommend adding a comprehensive test suite (e.g. `test-mcd-integer.c`) to ensure overflow checks behave as expected given their complexity. e.g. `_mcd_i64_mul_would_overflow (-1, INT64_MIN) -> false` is incorrect.

vector-of-bool: Fool me twice,

src/libmongoc/src/mongoc/mcd-integer.h
```diff
@@ -0,0 +1,380 @@
...
+#define MCD_INTEGER_H_INCLUDED
+
+#include <mongoc-prelude.h>
+
+#include <stdint.h>
+#include <stdbool.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
```
eramongodb, line 11: This does not seem particularly necessary. Remove?

eramongodb (approved):
New files should probably include copyright notices.
One minor question; otherwise LGTM.

src/libmongoc/src/mongoc/mcd-integer.h
vector-of-bool: Practically unecessary because everything is `static` and therefore bounded to the TU. This just becomes a habit when creating C headers.

PR #1098, opened 2022-08-26
CDRIVER-4361 Update zSeries RHEL 8.3 tasks and server versions

eramongodb:
## Description

This PR resolves CDRIVER-4361.

Server versions 4.0, 4.2, and 4.4 are not available for zSeries RHEL 8.3, so they have been removed. Server version 6.0 has been added instead.

The `download-mongodb.sh` file was synced with [it's latest state](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/8159c42f559900ca28381b284590756ce3a672c6/.evergreen/download-mongodb.sh) to resolve System Failures on Evergreen due to 5.0.8 no longer being available.

Changes verified by [this patch](https://spruce.mongodb.com/version/6308f5b4850e615766de8ca9).

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -24681,10 +24681,8 @@ buildvariants:
   - .authentication-tests .openssl
   - .latest .openssl !.nosasl .server
   - .latest .nossl
+  - .6.0 .openssl !.nosasl .server
   - .5.0 .openssl !.nosasl .server
-  - .4.4 .openssl !.nosasl .server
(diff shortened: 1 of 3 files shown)


rcsanchez97 (changes requested):
If you change `.evergreen/config.yml` manually rather than `build/evergreen_config_lib/variants.py` (in this case) and regenerate the Evergreen config, then the next time the config is regenerated this change will be lost.

rcsanchez97 (approved):

PR #1103, opened 2022-09-09
CDRIVER-4454 Some HTTP Fixes

vector-of-bool:
While preparing additional tests for CDRIVER-4454, a few issues were found in the internal HTTP API.


kevinAlbs (approved):
Nice, I did not realize `mongoc_http_response_t::status` was unused.

PR #1104, opened 2022-09-09
[CDRIVER-4454] Azure auto-KMS testing

vector-of-bool:
Refer: CDRIVER-4454

This changeset adds a simple IMDS-mimicking HTTP server based on Bottle. Further test cases and behaviors can be added to this server in the future. This also includes addition tweaks to the HTTP API and a timer API. (This changeset builds upon #1103) 

Individual commits can be viewed in order.


src/libmongoc/src/mongoc/mongoc-http.c
```diff
@@ -115,7 +119,11 @@ _mongoc_http_send (const mongoc_http_request_t *req,
       goto fail;
    }
 
-   stream = mongoc_client_connect_tcp (timeout_ms, &host_list, error);
+   stream = mongoc_client_connect_tcp (
+      // +1 to prevent passing zero as a timeout
+      mcd_get_milliseconds (mcd_timer_remaining (timer)) + 1,
+      &host_list,
+      error);
```
kevinAlbs, line 126: Good catch. If `_mongoc_http_send` requires a non-zero timeout, I suggest checking the timeout before each blocking call:

```c
#define CHECK_TIMEOUT(msg)                                        \
   if (mcd_get_milliseconds (mcd_timer_remaining (timer)) == 0) { \
      bson_set_error (error,                                      \
                      MONGOC_ERROR_STREAM,                        \
                      MONGOC_ERROR_STREAM_SOCKET,                 \
                      "Timeout before %s: %s",                    \
                      msg,                                        \
                      req->host);                                 \
      goto fail;                                                  \
   }

   CHECK_TIMEOUT ("connect")
```

`mongoc_stream_tls_handshake_block` interprets 0 as "no timeout".

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -715,38 +715,11 @@ _try_add_aws_from_env (bson_t *out, bson_error_t *error)
...
-   }
-
-fail:
-   _mongoc_http_response_cleanup (&resp);
-   mcd_azure_imds_request_destroy (&req);
-   return okay;
+   return mcd_azure_access_token_from_imds (out,
+                                            NULL, // Use the default host
+                                            0,    //  Default port as well
+                                            NULL, // No extra headres
```
kevinAlbs, line 721:
```suggestion
                                            NULL, // No extra headers
```

kevinAlbs (commented):
Nice, LGTM with an update to the timeout handling and documentation for the error constants.

I suggest moving mock server to [drivers-evergreen-tools ](https://github.com/mongodb-labs/drivers-evergreen-tools/tree/master/.evergreen/csfle) to make it accessible to other drivers when adding specification tests.

kevinAlbs (approved):

build/bottle.py
```diff
@@ -0,0 +1,3806 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Bottle is a fast and simple micro-framework for small web applications. It
+offers request dispatching (Routes) with url parameter support, templates,
+a built-in HTTP Server and adapters for many third party WSGI/HTTP-server and
+template engines - all in a single file and with no dependencies other than the
+Python Standard Library.
+
+Homepage and documentation: http://bottlepy.org/
```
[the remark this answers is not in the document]
vector-of-bool: I chose Bottle here because it is a single-file library that does not required additional dependencies. Flask etc will require a form of external dependency management. I'm fine doing external dev/CI Python dependencies, but I would like for there to be a more reliable development/CI workflow then is currently present.

src/libmongoc/src/mongoc/mongoc-http.c
vector-of-bool: Looking into this, there's a several goofs in timing that may be a problem, e.g. an unguarded multiply of `timeout_ms * 1000` that can overflow. Requiring the timeout to not be zero is strange behavior.

I also realize now that asserting on overflow in the duration code is probably bad (for example, one might want `mcd_seconds(INT64_MAX)` to become a maximum duration, rather than terminate the program, and using assertions implies that the caller is supposed to know what the time encoding is).

vector-of-bool:
@kevinAlbs I did some changes to the time arithmetic to be saturating rather than asserting (See [this commit](https://github.com/mongodb/mongo-c-driver/pull/1104/commits/02e73918a83a4bc11621f5433a5ba12762147bf5)). It shouldn't have any effect on the behavior of the in most cases, but I feel better about it. Could you take a quick look at that change?

src/libmongoc/src/mongoc/mcd-time.h
```diff
@@ -62,56 +74,117 @@ mcd_now (void)
...
-   BSON_ASSERT (!_mcd_i64_mul_would_overflow (m, 60));
+   if (_mcd_i64_mul_would_overflow (m, 60)) {
+      return m < 0 ? MCD_DURATION_MIN : MCD_DURATION_MAX;
+   }
    return mcd_seconds (m * 60);
 }
 
-/// Convert an abstract duration to a number of milliseconds
+/**
+ * @brief Obtain the cound of full milliseconds encoded in the given duration
```
kevinAlbs, line 152: "count"

src/libmongoc/src/mongoc/mcd-time.h
```diff
@@ -62,56 +74,117 @@ mcd_now (void)
...
 
-/// Create a duration from a number of microseconds
+/**
+ * @brief Create a duration from a number of microseconds.
+ *
+ * @param s A number of microseconds
+ * @return mcd_duration A duration corresponding to 's' microseconds.
+ *
+ * @note Saturates to the min/max duration if the duration is too great in
+ * magnitude.
```
kevinAlbs, line 84: Consider removing the note. It is not necessary if there is no possibility of overflow.

src/libmongoc/src/mongoc/mcd-time.h
```diff
@@ -132,10 +205,19 @@ mcd_later (mcd_time_point from, mcd_duration delta)
 static inline mcd_duration
 mcd_time_difference (mcd_time_point then, mcd_time_point from)
 {
-   BSON_ASSERT (!_mcd_i64_sub_would_overflow (then._rep, from._rep));
-   int64_t diff = then._rep - from._rep;
-   // Our time_point encodes the time using a microsecond counter.
-   return mcd_microseconds (diff);
+   if (_mcd_i64_sub_would_overflow (then._rep, from._rep)) {
+      if ((then._rep < 0) == (from._rep < 0)) {
```
kevinAlbs, line 209: Should this be `if (from._rep < 0) {`?

Positive overflow could only occur if subtracting a negative.

vector-of-bool: You are correct. I was thinking multiplication.

src/libmongoc/src/mongoc/mcd-time.h
vector-of-bool: For the current impl it won't overflow, but I want to avoid making the current impl (as amicrosecond counter) part of the API guarantees.

kevinAlbs (approved):

PR #1109, opened 2022-09-27
CDRIVER-3882 Support AWS temporary credentials

eramongodb:
## Description

This PR address CDRIVER-3882.

No changes were necessary to the C Driver interfaces to support AWS temp creds, as they are specified via BSON fields. Support for the `aws.sessionToken` field already added by https://github.com/mongodb/mongo-c-driver/pull/1043. This PR only required updating the legacy test runner to support the new corresponding spec tests.

Documentation was updated to describe the new `MONGOC_TEST_AWS_TEMP_*` environment variables by which AWS temp creds are communicated to the legacy test runner. The Evergreen config file and test scripts were updated to set these new environment variables accordingly via the [Drivers Evergreen Tools set-temp-creds.sh script](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/fd450866cbff621b3753ba6101b41454df612781/.evergreen/csfle/set-temp-creds.sh).

As a drive-by fix, some noisy error messages generated by `add-build-dirs-to-paths.sh` were addressed.

## Implementation Difficulties

During implementation of these changes, various unexpected issues were discovered or encountered when attempting to verify changes using Evergreen. The effort invested in dealing with these issues far outweighed the time it took to actually implement the required features relevant to CDRIVER-3882 itself. Some of these issues include:

- the need to explicitly set the `$PYTHON` env var when invoking the `set-temp-creds.sh` script to avoid the script from defaulting to an incorrect python binary.
- deactivation of the virtual environment to avoid potential double-activation issues.
- avoiding racing on the creation of the virtual environment in parallel tasks.
- avoid racing on the activation of the virtual environment in parallel tasks (see https://github.com/mongodb-labs/drivers-evergreen-tools/pull/234).

Additional issues encountered during patch builds that are deliberately not addressed in this PR due to being seemingly unrelated or demanding more investigation effort than is relevant to CDRIVER-3882 (to be/should be addressed separately):

- `AssertionError: Filename ... does not start with any of these prefixes: [...]` failure on zSeries (see BUILD-15997).
- Miscellaneous failures of `debug-*` tasks on some variants (not caused by changes in this PR).
- Miscellaneous failures of `test-*` tasks on some variants due to `/http` test (not caused by changes in this PR).

An attempt to avoid parallel invocation of `activate_venv.sh` by moving mock KMS server management into `run-tests.sh` was obstructed by task timeout issues on Windows variants. This issue may be related to [EVG-7818](https://jira.mongodb.org/browse/EVG-7818), but the attempt was aborted due to consuming too much time spent on debugging the problem; may revisit in a separate PR.

Changes verified by [this patch](https://spruce.mongodb.com/version/63335ee756234317acad9e25/tasks). Unrelated nature of failures verified by [this patch](https://spruce.mongodb.com/version/633366be9ccd4e283c62c1cd/tasks) (no changes relative to target branch).

### DIFF
--- a/build/evergreen_config_lib/functions.py
+++ b/build/evergreen_config_lib/functions.py
@@ -587,7 +587,7 @@ all_functions = OD([
         fi
         ''', test=False)
     )),
-    ('prepare csfle venv', Function(
+    ('run kms servers', Function(
         shell_exec(r'''
         echo "Preparing CSFLE venv environment..."
         cd ./drivers-evergreen-tools/.evergreen/csfle
@@ -599,9 +599,7 @@ all_functions = OD([
         . ./activate_venv.sh
         deactivate
         echo "Preparing CSFLE venv environment... done."
-        ''', test=False)
-    )),
-    ('run kms servers', Function(
+        ''', test=False),
         shell_exec(r'''
         echo "Starting mock KMS servers..."
         cd ./drivers-evergreen-tools/.evergreen/csfle
--- a/build/evergreen_config_lib/functions.py
+++ b/build/evergreen_config_lib/functions.py
@@ -587,6 +587,20 @@ all_functions = OD([
         fi
         ''', test=False)
     )),
+    ('prepare csfle venv', Function(
+        shell_exec(r'''
+        echo "Preparing CSFLE venv environment..."
+        cd ./drivers-evergreen-tools/.evergreen/csfle
+        # This function ensures future invocations of activate_venv.sh conducted in
+        # parallel do not race to setup a venv environment; it has already been prepared.
+        # This primarily addresses the situation where the "run tests" and "run kms servers"
+        # functions invoke 'activate_venv.sh' simultaneously.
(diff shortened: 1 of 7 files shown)


vector-of-bool (approved):
LGTM

build/evergreen_config_lib/functions.py
```diff
@@ -562,6 +587,20 @@
         fi
         ''', test=False)
     )),
+    ('prepare csfle venv', Function(
+        shell_exec(r'''
```
kevinAlbs, line 591: `prepare csfle venv` is always called before `run kms servers`.

Instead of a new function, consider putting the `shell_exec` as the first command of the `Function` in `run kms servers`.

kevinAlbs (approved):

kevinAlbs:
The efforts made to improve the implementation difficulties (filing BUILD-15997, improving race behavior of dos2unix, reopening EVG-7818) are appreciated.

PR #1114, opened 2022-10-05
build/mongodl.py: Fall back to /usr/lib/os-release

mrzelak:
According to `man os-release', file /etc/os-release isn't mandatory to be present in the system. Fallback to /usr/lib/os-release in absence of /etc/os-release.


vector-of-bool:
Thank you for this fix! I didn't know about `/usr/lib/os-release`. I made a small tweak to the lookup pattern, but it otherwise looks okay to merge.

vector-of-bool (approved):

PR #1115, opened 2022-10-05
build/mongodl.py: Port the build system to Mint

mrzelak:
In Linux Mint in /etc/os-release there is ID=linuxmint instead of 'mint' present in DISTRO_ID_MAP. Add a mapping to handle it.

Add 'linuxmint' mapping in DISTRO_VERSION_MAP corresponding to official Mint docs: https://linuxmint.com/download_all.php


vector-of-bool (approved):

vector-of-bool:
Thank you for this contribution! I will let CI run (just to be sure), then it will probably be fine to merge.

PR #1116, opened 2022-10-05
build/mongodl.py: Add quirk for Ubuntu 22.04

mrzelak:
Currently there is no downloadable tarballs for mongo_shared for Ubuntu 22.04. Compilation terminates when this phase is reached. Define a mapping in DISTRO_ID_TO_TARGET from '22.*' to 'ubuntu2004' in order to finish build on latest Ubuntu version.

### DIFF
--- a/build/mongodl.py
+++ b/build/mongodl.py
@@ -50,7 +50,8 @@ DISTRO_VERSION_MAP = {
 
 DISTRO_ID_TO_TARGET = {
     'ubuntu': {
-        '22.*': 'ubuntu2204',
+        #quirk
+        '22.*': 'ubuntu2004',
         '20.*': 'ubuntu2004',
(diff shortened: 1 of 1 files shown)


vector-of-bool:
Sorry for the delay on this one. We've been debating on the best action here, since this fix might soon become obsolete with an official Ubuntu 22.04 build. Until then, this fix can be used locally, or a build can be done that disables the invocation of this script. This script is only needed for testing purposes, so building with tests disabled will prevent this script from executing.

kevinAlbs:
Thank you for the contribution.

mongodl.py is no longer required to build tests as of CDRIVER-4427. That will be released in 1.23.2.
In the meantime, if it is not necessary to build tests, configure with cmake using `-DENABLE_TESTS=OFF`.

PR #1117, opened 2022-10-05
build/mongodl.py: Update version or leave it

mrzelak:
When version of OS is not provided in DISTRO_VERSION_MAP but name of OS is, KeyValue RuntimeError is raised because of accessing nonexistant key. Check whether mapped version is provided; otherwise leave it.


kevinAlbs:
Thank you for the PR. Closing PR. mongodl.py is no longer used in the build of the C driver.

PR #1119, opened 2022-10-12
CDRIVER-4427 Test with consistent versions of crypt_shared and server.

kevinAlbs:
# Summary

- Test with consistent versions of crypt_shared and server.
- Remove `MONGOC_TEST_USE_CRYPT_SHARED` cmake option.

# Background & Motivation

Using download-mongodb.sh to download crypt_shared is intended for consistency with drivers-evergreen-tools. Eventually drivers-evergreen-tools may be cloned to resolve CDRIVER-4029.

The `MONGOC_TEST_USE_CRYPT_SHARED` option was removed as it was no longer needed. 

Full patch build: https://spruce.mongodb.com/version/6346f78d9ccd4e5b4bb1054b/


.evergreen/integration-tests.sh
```diff
@@ -183,3 +183,21 @@ echo $MONGO_SHELL_CONNECTION_FLAGS
...
+  echo "Downloading crypt_shared package from $MONGO_CRYPT_SHARED_DOWNLOAD_URL"
+  download_and_extract_crypt_shared "$MONGO_CRYPT_SHARED_DOWNLOAD_URL" "$EXTRACT" CRYPT_SHARED_LIB_PATH
+  echo "CRYPT_SHARED_LIB_PATH:" $CRYPT_SHARED_LIB_PATH
+  if [ -z $CRYPT_SHARED_LIB_PATH ]; then
+    echo "CRYPT_SHARED_LIB_PATH must be assigned, but wasn't" 1>&2 # write to stderr"
+    exit 1
+  fi
+cat <<EOT >> mo-expansion.yml
+CRYPT_SHARED_LIB_PATH: "$CRYPT_SHARED_LIB_PATH"
+EOT
```
eramongodb, line 201:
```suggestion
if [ -z "$MONGO_CRYPT_SHARED_DOWNLOAD_URL" ]; then
  echo "There is no crypt_shared library for distro='$DISTRO' and version='$MONGODB_VERSION'".
else
  echo "Downloading crypt_shared package from $MONGO_CRYPT_SHARED_DOWNLOAD_URL"
  download_and_extract_crypt_shared "$MONGO_CRYPT_SHARED_DOWNLOAD_URL" "$EXTRACT" "CRYPT_SHARED_LIB_PATH"
  echo "CRYPT_SHARED_LIB_PATH: $CRYPT_SHARED_LIB_PATH"
  if [ -z "$CRYPT_SHARED_LIB_PATH" ]; then
    echo "CRYPT_SHARED_LIB_PATH must be assigned, but wasn't" 1>&2 # write to stderr"
    exit 1
  fi
cat >>mo-expansion.yml <<EOT 
CRYPT_SHARED_LIB_PATH: "$CRYPT_SHARED_LIB_PATH"
EOT
```
Minor improvements to scripting.

kevinAlbs: Thank you for the suggested change. That appears to pass ShellCheck. Updated.

.evergreen/integration-tests.sh
```diff
@@ -183,3 +183,21 @@ echo $MONGO_SHELL_CONNECTION_FLAGS
 
 `pwd`/mongodb/bin/mongo $MONGO_SHELL_CONNECTION_FLAGS --eval 'printjson(db.serverBuildInfo())' admin
 `pwd`/mongodb/bin/mongo $MONGO_SHELL_CONNECTION_FLAGS --eval 'printjson(db.isMaster())' admin
+
+touch mo-expansion.yml
```
eramongodb, line 187: Recommend documenting purpose of this `touch` command. Is `mo-expansion.yml` expected to already exist (hence append `>>mo-expansion.yml` below)? If not, it seems more appropriate to remove/replace any existing `mo-expansion.yml` file to avoid the possibility of unexpected additional expansions.

eramongodb (approved):

.evergreen/integration-tests.sh
kevinAlbs: `mo-expansion.yml` is expected to not exist before the `touch`. The `touch` was added to limit changing the snippet copied from [run-orchestration.sh from drivers-evergreen-tools](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/651efa00f0118fff65e075706a80f4fafd7b0b04/.evergreen/run-orchestration.sh#L105-L119).

Since `integration-tests.sh` already diverges from run-orchestration, updated to remove the `touch mo-expansion.yml` and change `>>mo-expansion.yml` to `>mo-expansion.yml`.

.evergreen/integration-tests.sh
```diff
@@ -183,3 +183,21 @@ echo $MONGO_SHELL_CONNECTION_FLAGS
...
+  echo "Downloading crypt_shared package from $MONGO_CRYPT_SHARED_DOWNLOAD_URL"
+  download_and_extract_crypt_shared "$MONGO_CRYPT_SHARED_DOWNLOAD_URL" "$EXTRACT" "CRYPT_SHARED_LIB_PATH"
+  echo "CRYPT_SHARED_LIB_PATH: $CRYPT_SHARED_LIB_PATH"
+  if [ -z "$CRYPT_SHARED_LIB_PATH" ]; then
+    echo "CRYPT_SHARED_LIB_PATH must be assigned, but wasn't" 1>&2 # write to stderr"
+    exit 1
+  fi
+cat >>mo-expansion.yml <<EOT 
+CRYPT_SHARED_LIB_PATH: "$CRYPT_SHARED_LIB_PATH"
+EOT
```
vector-of-bool, line 201: In the above, if the path contains a backslash `\`, it will generate escape sequences in the YAML. Using `jq`:

```sh
jq -n '{CRYPT_SHARED_LIB_PATH: $p}' --arg p "$CRYPT_SHARED_LIB_PATH" > mo-expansions.yml
```

Which will handle string quoting, etc. This is just me being somewhat paranoid, though.

.evergreen/run-tests.sh
```diff
@@ -112,7 +112,7 @@ if [ "$CLIENT_SIDE_ENCRYPTION" = "on" ]; then
    echo "Waiting for mock KMS servers to start... done."
    if ! test -d /cygdrive/c; then
       # We have trouble with this test on Windows. only set cryptSharedLibPath on other platforms
-      export MONGOC_TEST_CRYPT_SHARED_LIB_PATH="$(find . -wholename '*src/libmongoc/mongo_crypt_v1.*' -and -regex '.*\(.dll\|.dylib\|.so\)' | head -n1)"
+      export MONGOC_TEST_CRYPT_SHARED_LIB_PATH=$CRYPT_SHARED_LIB_PATH
```
vector-of-bool, line 115: Quote this expansion

vector-of-bool (approved):
Only minor comments, LGTM

.evergreen/run-tests.sh
kevinAlbs: Done.

.evergreen/integration-tests.sh
kevinAlbs: `jq` appears not to be available on the `windows-64-vs2017-large` host. The output path from download-mongodb.sh is run through `cygpath -m`. A backslash may not be a concern.

PR #1120, opened 2022-10-13
CDRIVER-4484 Address -Wformat and -Wstrict-prototypes warnings

eramongodb:
## Description

This PR resolves CDRIVER-4484 by addressing `-Wformat=2` and `-Wstrict-prototypes` warnings.

Warnings were detected using GCC 12.1.0 and Clang 10.0.0 on Ubuntu 20.04 and GCC 12.1.0 (mingw-w64) on Windows with `ENABLE_CLIENT_SIDE_ENCRYPTION=(OFF|ON)` and `ENABLE_TRACING=(OFF|ON)`.

### DIFF
--- a/src/kms-message/src/kms_crypto.h
+++ b/src/kms-message/src/kms_crypto.h
@@ -35,10 +35,10 @@ typedef struct {
 } _kms_crypto_t;
 
 int
-kms_crypto_init ();
+kms_crypto_init (void);
 
 void
-kms_crypto_cleanup ();
+kms_crypto_cleanup (void);
 
 bool
 kms_sha256 (void *ctx, const char *input, size_t len, unsigned char *hash_out);
--- a/src/kms-message/src/kms_crypto_libcrypto.c
+++ b/src/kms-message/src/kms_crypto_libcrypto.c
@@ -39,13 +39,13 @@ EVP_MD_CTX_free (EVP_MD_CTX *ctx)
 #endif
 
 int
-kms_crypto_init ()
+kms_crypto_init (void)
 {
    return 0;
 }
 
 void
-kms_crypto_cleanup ()
+kms_crypto_cleanup (void)
 {
 }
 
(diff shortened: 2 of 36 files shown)


kevinAlbs (approved):

vector-of-bool (approved):
LGTM, but can we add these warnings to somewhere that they will appear again? i.e. maintainer-flags or MongoC-Warnings.cmake?

eramongodb:
> can we add these warnings to somewhere that they will appear again?

`-Wformat=1` is already enabled implicity via `-Wall` according to [GCC Options documentation](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html), which is already enabled when building on Linux and MacOS distros ([ENABLE_MAINTAINER_FLAGS=ON](https://github.com/mongodb/mongo-c-driver/blob/c9296d582d263a28297bb8314a2926308e18a62a/.evergreen/compile-unix.sh#L94) -> [maintainer-flags.txt](https://github.com/mongodb/mongo-c-driver/blob/c9296d582d263a28297bb8314a2926308e18a62a/build/maintainer-flags.txt#L1)). I do not think any of the instances that were reported required `-Wformat=2`.

Most of these warnings have escaped noticed due to requiring:

1. `ENABLE_TRACING=ON`, which is only done for the [compile-tracing](https://github.com/mongodb/mongo-c-driver/blob/c9296d582d263a28297bb8314a2926308e18a62a/build/evergreen_config_lib/tasks.py#L268) task run on [Ubuntu 18.04](https://github.com/mongodb/mongo-c-driver/blob/c9296d582d263a28297bb8314a2926308e18a62a/build/evergreen_config_lib/variants.py#L50)), and
2. compiling Windows code paths with GCC (i.e. MinGW), as MSVC does not have `-Wformat`.

This may be an argument for adding the `compile-tracing` task to more variants such as Windows. I will consider doing so when working on CDRIVER-3620.

As for `-Wstrict-prototypes`, it seems to be a hit-or-miss when testing with Clang. Testing with GCC uncovered a few more instances. Still, I think it is safe to add to the set of maintainer flags. Filed CDRIVER-4504 as followup work.

jeroen:
Which release would this be included in? I noticed it hasn't been included in 1.23.1 and 1.22.2

PR #1121, opened 2022-10-14
CDRIVER-4502: Handle int64 connectionId values in hello responses

alcaeus:
CDRIVER-4502

This updates the server description and command monitoring struct to store the connection ID as `int64_t` to reflect the type returned by the server. `mongoc_server_description_handle_hello` was updated to accept any integer for the `connectionId` field in the response and will read the field as int64.

To preserve backward compatibility, `mongoc_apm_command_started_get_server_connection_id` (and the equivalent methods for other command monitoring events) still return the value as `int32_t`. If the connection ID returned from the server exceeds the 32-bit range, `-1` is returned.


kevinAlbs (approved):
Thank you for the fix.

PR #1123, opened 2022-10-18
CDRIVER-4454 fix error return and HTTP version check

kevinAlbs:
# Summary

- Allow HTTP/1.1 protocol in response
- Fix error return check in `_try_add_azure_from_env`

# Background & Motivation

Discovered during integration testing for DRIVERS-2411. WIP integration tests pass with these fixes.

HTTP requests from libmongoc use [HTTP/1.0](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/mongoc-http.c#L68). The Azure IMDS server responds with HTTP/1.1. https://www.rfc-editor.org/rfc/rfc2145#section-2.3 does not prohibit a server response with a minor version greater than the one received.

> An HTTP server SHOULD send a response version equal to the highest
   version for which the server is at least conditionally compliant, and
   whose major version is less than or equal to the one received in the
   request.  An HTTP server MUST NOT send a version for which it is not
   at least conditionally compliant.  A server MAY send a 505 (HTTP
   Version Not Supported) response if cannot send a response using the
   major version used in the client's request.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-crypt.c
+++ b/src/libmongoc/src/mongoc/mongoc-crypt.c
@@ -760,7 +760,7 @@ _try_add_azure_from_env (_mongoc_crypt_t *crypt,
       // abstract monotonic "now"
       crypt->azure_token_issued_at = mcd_now ();
       // Get the token:
-      if (_request_new_azure_token (&crypt->azure_token, error)) {
+      if (!_request_new_azure_token (&crypt->azure_token, error)) {
          return false;
       }
    }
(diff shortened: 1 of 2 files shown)


vector-of-bool (approved):
LGTM, but: Do we know for certain that the server will respond with a plain Transfer Encoding in this case? (i.e. it won't unexpectedly send a chunked response)

kevinAlbs:
> LGTM, but: Do we know for certain that the server will respond with a plain Transfer Encoding in this case? (i.e. it won't unexpectedly send a chunked response)

Good question. I am not sure. libmongocrypt added chunked support due to GCP responding with it: MONGOCRYPT-282.

PR #1132, opened 2022-10-22
CDRIVER-4507 Changed 8 bit lock word gEmulAtomicLock to 32bit to solve WIN32 build failure and for optimized performance.

gopukrishnasIBM:
I got the following error message while building on WIN32:
```
BLD_SHARED.out:5565:libbson.a(bson-atomic.obj) : error LNK2019: unresolved external symbol __InterlockedCompareExchange8 referenced in function __bson_emul_atomic_int64_fetch_add.
BLD_SHARED.out:5566:libbson.a(bson-atomic.obj) : error LNK2019: unresolved external symbol __InterlockedExchange8 referenced in function __bson_emul_atomic_int64_fetch_add.
BLD_SHARED.out:5567:db2engn.dll : fatal error LNK1120: 2 unresolved externals
BLD_SHARED.out:6462:gnumake-380[1]: Target `engn_DLLS' not remade because of errors.
```

In bson-atomic.h file, we use _emul functions to emulate 64-bit atomic operations when we're not on a 64-bit platform.
```
#if (defined(_MSC_VER) && !defined(_M_IX86)) || (defined(__LP64__) && __LP64__)
...
#else
...
static BSON_INLINE int64_t
bson_atomic_int64_fetch_add (int64_t volatile *val,
                             int64_t v,
                             enum bson_memory_order order)
{
   return _bson_emul_atomic_int64_fetch_add (val, v, order);
}
...
```

The implementation of `_bson_emul_atomic_int64_fetch_add` is in `src/bson-atomic.c`:
```
int64_t
_bson_emul_atomic_int64_fetch_add (volatile int64_t *p,
                                   int64_t n,
                                   enum bson_memory_order _unused)
{
   int64_t ret;
   _lock_emul_atomic ();
   ret = *p;
   *p += n;
   _unlock_emul_atomic ();
   return ret;
}
```

We emulate the atomic operation with a standalone lock and normal arithmetic.
The standalone lock uses an 8-bit lock word:
```
static int8_t gEmulAtomicLock = 0;

static void
_lock_emul_atomic ()
{
   int i;
   if (bson_atomic_int8_compare_exchange_weak (
          &gEmulAtomicLock, 0, 1, bson_memory_order_acquire) == 0) {
      /* Successfully took the spinlock */
      return;
   }
   /* Failed. Try taking ten more times, then begin sleeping. */
   for (i = 0; i < 10; ++i) {
      if (bson_atomic_int8_compare_exchange_weak (
             &gEmulAtomicLock, 0, 1, bson_memory_order_acquire) == 0) {
         /* Succeeded in taking the lock */
         return;
      }
   }
   /* Still don't have the lock. Spin and yield */
   while (bson_atomic_int8_compare_exchange_weak (
             &gEmulAtomicLock, 0, 1, bson_memory_order_acquire) != 0) {
      bson_thrd_yield ();
   }
}

static void
_unlock_emul_atomic ()
{
   int64_t rv = bson_atomic_int8_exchange (
      &gEmulAtomicLock, 0, bson_memory_order_release);
   BSON_ASSERT (rv == 1 && "Released atomic lock while not holding it");
}
```
Which requires an implementation of bson_atomic_int8_compare_exchange_weak().

From `bson-atomic.h`, this is instantiated via macros:
```
#define DEF_ATOMIC_CMPEXCH_WEAK(                                              \
   VCSuffix1, VCSuffix2, GNU_MemOrder, Ptr, ExpectActualVar, NewValue)        \
   do {                                                                       \
      BSON_IF_MSVC (ExpectActualVar = BSON_CONCAT3 (                          \
                       _InterlockedCompareExchange, VCSuffix1, VCSuffix2) (   \
                       Ptr, NewValue, ExpectActualVar);)                      \
...
   } while (0)
   ```
   and
   ```
   #define DECL_ATOMIC_INTEGRAL(NamePart, Type, VCIntrinSuffix)                  \
...
   static BSON_INLINE Type bson_atomic_##NamePart##_compare_exchange_weak (   \
      Type volatile *a,                                                       \
      Type expect,                                                            \
      Type new_value,                                                         \
      enum bson_memory_order ord)                                             \
   {                                                                          \
      Type actual = expect;                                                   \
...
         DEF_ATOMIC_CMPEXCH_WEAK (VCIntrinSuffix,                             \
                                  MSVC_MEMORDER_SUFFIX (_acq),                \
                                  __ATOMIC_ACQUIRE,                           \
                                  a,                                          \
                                  actual,                                     \
                                  new_value);                                 \
...
      }                                                                       \
      return actual;                                                          \
   }
   ```
   Which ultimately gets instantiated with this:
   ```
   DECL_ATOMIC_STDINT (int8, 8) -> DECL_ATOMIC_INTEGRAL (int8, int8_t, 8)
   ```
   So pulling this all together:
   ```
   NamePart = int8
Type = int8_t
VCIntrinSuffix = 8
   ```
   Which gives:
   ```
   bson_atomic_int8_compare_exchange_weak(...)
{
   int8_t actual = new_value;
   actual = _InterlockedCompareExchange8(a, new_value, actual);
}
```
So this explains how the lock code path drives the call to `_InterlockedComparedExchange8()`.
In a similar manner, the unlock code path drives the call to `_InterlockedExchange8()`.

```
```
The root issue here is that we're using an 8-bit lock in our emulation of 64-bit atomics and 32bit intel16 compiler (which am using) doesn't have 8bit atomic intrinsics. Instead of an 8-bit lock, we can use a 32-bit lock in the emulation, which will actually perform better.

### DIFF
--- a/src/libbson/src/bson/bson-atomic.c
+++ b/src/libbson/src/bson/bson-atomic.c
@@ -51,27 +51,27 @@ bson_memory_barrier (void)
  * Some platforms do not support compiler intrinsics for atomic operations.
  * We emulate that here using a spin lock and regular arithmetic operations
  */
-static int8_t gEmulAtomicLock = 0;
+static int32_t gEmulAtomicLock = 0;
 
 static void
 _lock_emul_atomic ()
 {
    int i;
-   if (bson_atomic_int8_compare_exchange_weak (
+   if (bson_atomic_int32_compare_exchange_weak (
           &gEmulAtomicLock, 0, 1, bson_memory_order_acquire) == 0) {
       /* Successfully took the spinlock */
       return;
    }
    /* Failed. Try taking ten more times, then begin sleeping. */
    for (i = 0; i < 10; ++i) {
-      if (bson_atomic_int8_compare_exchange_weak (
+      if (bson_atomic_int32_compare_exchange_weak (
              &gEmulAtomicLock, 0, 1, bson_memory_order_acquire) == 0) {
          /* Succeeded in taking the lock */
          return;
       }
    }
    /* Still don't have the lock. Spin and yield */
-   while (bson_atomic_int8_compare_exchange_weak (
+   while (bson_atomic_int32_compare_exchange_weak (
              &gEmulAtomicLock, 0, 1, bson_memory_order_acquire) != 0) {
       bson_thrd_yield ();
    }
@@ -80,7 +80,7 @@ _lock_emul_atomic ()
 static void
 _unlock_emul_atomic ()
 {
-   int64_t rv = bson_atomic_int8_exchange (
+   int64_t rv = bson_atomic_int32_exchange (
       &gEmulAtomicLock, 0, bson_memory_order_release);
    BSON_ASSERT (rv == 1 && "Released atomic lock while not holding it");
 }


vector-of-bool:
Thank you for your contribution!

Unfortunately, it seems that this can infinitely recurse if the int32 locks are emulated, since it will try to perform an atomic op while taking the lock, and that atomic op might itself try to take the lock recursively. In the linked build results, this fails on the s390x platform, which looks to be an odd-one-out in this case. For 32-bit Windows, it may be better to provide a special case here, or it may be better to replace the custom spinlock with a regular platform mutex.

vector-of-bool:
Due to lack of activity, this PR will be closed.

The CDRIVER-4507 ticket will remain open, and this PR can be re-opened if necessary.

PR #1136, opened 2022-10-25
CDRIVER-4505 Fix build on MacOS due to undeclared getpagesize

kkloberdanz:
CDRIVER-4505

It appears that getpagesize has been deprecated on newer MacOS versions. We can see that POSIX version 7 (link below) has removed getpagesize altogether from the standard, so even including unistd.h does not solve this problem on newer MacOS versions.

Please see here at section B.3.2 System Interfaces Removed in the Previous Version:
https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xsh_chap03.html

Relevant stackoverflow:
https://stackoverflow.com/a/37193970

Jira:
https://jira.mongodb.org/browse/CDRIVER-4505

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-counters.c
+++ b/src/libmongoc/src/mongoc/mongoc-counters.c
@@ -20,11 +20,11 @@
 #include <fcntl.h>
 #include <stdlib.h>
 #include <string.h>
-#include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


galon1 (approved):
LGTM!

PR #1138, opened 2022-10-31
CDRIVER-4136 Use C99 as default C standard without requiring explicit CMake config variables

eramongodb:
## Description

This PR is related to CDRIVER-4136 and is a followup to https://github.com/mongodb/mongo-c-driver/pull/1067.

Explicit CMake config variables `CMAKE_C_STANDARD`, `CMAKE_C_STANDARD_REQUIRED`, and `CMAKE_C_EXTENSIONS` that were required by https://github.com/mongodb/mongo-c-driver/pull/1067 are now set by default during CMake configuration unless otherwise directed (e.g. passing `-DCMAKE_C_STANDARD=11` during CMake configuration). This means the C Driver will now _implicitly default to_ conforming to a minimum C standard (currently C99).

Note: The variables are deliberately set as normal variables, not cache variables, to avoid influencing the state of parent CMake projects that may `add_subdirectory()` the C Driver, notably libmongocrypt.

Changes verified by [this patch](https://spruce.mongodb.com/version/635ff3903066151bd7e2edc6).


vector-of-bool (approved):
LGTM. Huzzah!

kevinAlbs (approved):

PR #1124, opened 2022-10-19
CDRIVER-4454 add integration tests for Azure KMS

kevinAlbs:
# Summary

- Implements integration tests proposed in https://github.com/mongodb/specifications/pull/1325

Here is an Evergreen patch build with the new tasks: https://spruce.mongodb.com/version/63500fcbd1fe071dfc2682da/

# Notes

The additions to the Python evergreen generation scripts are minimal. IMO the Python scripts are not significantly more readable than the Evergreen config YML. For this PR, I found it easier to start with Evergreen YML, then translate to the Python scripts.

PR #1140, opened 2022-11-02
CDRIVER-4435 Add support for GCP service accounts in GCP KMS 

galon1:
## Summary:
1. upgrade libmongocrypt dependency to 1.6.0 
2. If the KMS provider is set up with an empty `gcp: {}`, create a `gcp_service_account_token` by sending a HTTP request to the GCP metadata server. Most of these changes are modeled after how azure tokens were added.  
3. Testing file for `service-gcp.c` which uses a mock GCP server to test the code handles different HTTP responses. This test uses the `fake_azure.py` mock server.
4. Integration test for GCP. Adds a new task group in `.evergreen/config.yml` to spin up a remote GCE instance using the scripts in `drivers-evergreen-tools` ([link to github](https://github.com/mongodb-labs/drivers-evergreen-tools/tree/master/.evergreen/csfle/gcpkms)). Here is a patch [build](https://spruce.mongodb.com/version/6362cf9ae3c331270cf4c53c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) with the new GCP KMS tasks.

## Notes:
the files listed below were copied from this [PR ](https://github.com/mongodb/mongo-c-driver/pull/1124#event-7643575802) since they will be added late, but are needed for integration testing.
1. [build/evergreen_config_generator/taskgroups.py](https://github.com/mongodb/mongo-c-driver/compare/master...galon1:mongo-c-driver:CDRIVER-4435?expand=1#diff-974bdd422c8cf802e9d17ce4ab23440a2bcc19601adf9560e4f0024f8bd45c25)
2. [build/evergreen_config_lib/taskgroups.py](https://github.com/mongodb/mongo-c-driver/compare/master...galon1:mongo-c-driver:CDRIVER-4435?expand=1#diff-261f6a621f7139535e9f71f484eed10e0a7906818c8f30337a43c674460c6034)

Most recent patch build with the new tasks can be found here: https://spruce.mongodb.com/version/6363fa95d1fe076083e2114f/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


build/evergreen_config_lib/testgcpkms.py
```diff
@@ -0,0 +1,122 @@
...
+            export GCPKMS_GCLOUD=${GCPKMS_GCLOUD}
+            export GCPKMS_PROJECT=${GCPKMS_PROJECT}
+            export GCPKMS_ZONE=${GCPKMS_ZONE}
+            export GCPKMS_INSTANCENAME=${GCPKMS_INSTANCENAME}
+            DRIVERS_TOOLS=$(pwd)/drivers-evergreen-tools
+            mkdir testgcpkms
+            cp ./mongoc/src/libmongoc/test-gcpkms ./mongoc/install/lib/libmongocrypt.* testgcpkms
+            tar czf testgcpkms.tgz testgcpkms/*
+            GCPKMS_SRC="testgcpkms.tgz" GCPKMS_DST=$GCPKMS_INSTANCENAME: $DRIVERS_TOOLS/.evergreen/csfle/gcpkms/copy-file.sh
+            echo "Copying files ... end"
```
vector-of-bool, line 45: Trailing whitespace on this line (35) is causing the YAML output string to be garbled in `config.yml`.

build/evergreen_config_lib/testgcpkms.py
```diff
@@ -0,0 +1,122 @@
...
+            GCPKMS_CMD="LD_LIBRARY_PATH=./testgcpkms MONGODB_URI='mongodb://localhost:27017' ./testgcpkms/test-gcpkms" $DRIVERS_TOOLS/.evergreen/csfle/gcpkms/run-command.sh
+            ''')]
+
+    failtask = NamedTask(task_name="testgcpkms-fail-task")
+    failtask.commands = [
+        func("fetch source"),
+        shell_exec (r'''
+            pushd mongoc
+            ./.evergreen/compile-test-gcpkms.sh
+            popd ''', test=False),
```
vector-of-bool, line 65: Trailing whitespace after `popd` is garbling the generated string in `config.yml`

build/fake_azure.py
```diff
@@ -83,6 +83,35 @@ def wrapped():
 def test_params() -> 'dict[str, str]':
     return parse_qs(request.headers.get('X-MongoDB-HTTP-TestParams', ''))
 
+@imds.get('/computeMetadata/v1/instance/service-accounts/default/token')
+@handle_asserts
+def get_gcp_token():
```
vector-of-bool, line 88: Aside: If we're using this file for multiple mock services, it may be useful to rename it and/or split it into multiple Bottle applications.

src/libmongoc/src/mongoc/mongoc-error.h
```diff
@@ -139,6 +140,10 @@ typedef enum {
    // Azure errors
    MONGOC_ERROR_AZURE_HTTP,
    MONGOC_ERROR_AZURE_BAD_JSON,
+
+   // GCP errors
+   MONGOC_ERROR_GCP_HTTP,
+   MONGOC_ERROR_GCP_BAD_JSON
```
vector-of-bool, line 146: Maybe these should be merged since errors are already distinguished by the error domain.

src/libmongoc/tests/test-service-gcp.c
```diff
@@ -0,0 +1,136 @@
...
+      "Connection: close\r\n"
+      "Metadata-Flavor: Google\r\n"
+      "\r\n");
+   bson_string_free (req_str, true);
+}
+
+static const char *
+_get_test_host (void)
+{
+   return getenv ("MCD_TEST_AZURE_IMDS_HOST");
```
vector-of-bool, line 78: This env var (and the test fixture) should be renamed since it is used for both Azure and GCP testing.

vector-of-bool (changes requested):
Some changes requested, some about consolidating with Azure KMS work.

build/fake_azure.py
galon1: I'm thinking of just renaming it and not splitting it into multiple Bottle applications, because the mock services at this point are so similar and use identical helper functions.

src/libmongoc/src/mongoc/mongoc-error.h
```diff
@@ -136,9 +137,10 @@ typedef enum {
 
    MONGOC_ERROR_CLIENT_INVALID_LOAD_BALANCER,
 
-   // Azure errors
-   MONGOC_ERROR_AZURE_HTTP,
-   MONGOC_ERROR_AZURE_BAD_JSON,
+   /* An error related to either GCP metadata or Azure IMDS server */
+   MONGOC_ERROR_KMS_SERVER_HTTP,
+   MONGOC_ERROR_KMS_SERVER_BAD_JSON,
```
vector-of-bool, line 142: @kevinAlbs Do we want to keep the old names for these constants present? Have they been released for long enough to be relied on?

vector-of-bool (approved):
LGTM, pending one comment.

src/libmongoc/src/mongoc/mongoc-error.h
kevinAlbs: No need. The constants were introduced in CDRIVER-4454 and have not been released yet.

kevinAlbs (approved):
Nicely done. Before merging this PR, please run an Evergreen patch build with the two new tasks. The two new tasks are not run by default.
That can be done with:
```
evergreen patch --project=mongo-c-driver --description="Test GCP $(git log --oneline HEAD~1..HEAD)" \
    --yes --finalize \
    -v testgcpkms-variant -t testgcpkms-task \
    -v testgcpkms-variant -t testgcpkms-fail-task
```

build/evergreen_config_lib/testgcpkms.py
```diff
@@ -0,0 +1,120 @@
...
+            LD_LIBRARY_PATH=$(pwd)/install MONGODB_URI='mongodb://localhost:27017' EXPECT_ERROR='Failed to connect to: metadata.google.internal' ./mongoc/src/libmongoc/test-gcpkms''')]
+    
+    return [passtask, failtask]
+
+def _create_variant():
+        return Variant(
+        name="testgcpkms-variant",
+        display_name="GCP KMS",
+        # GCP Virtual Machine created is Debian 10.
+        run_on="debian10-small", tasks=[
```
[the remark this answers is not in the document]
galon1: I made it debian 10 because I couldn't download libmongocrypt using debian 11 on my local computer since the package wasn't available yet. Is it fine though for the GCP virtual machine to be on debian 11?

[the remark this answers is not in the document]
kevinAlbs: The GCP virtual machine is already using Debian 11. Suggested change is to have the OS in Evergreen match the GCP OS. libmongocrypt is built from source in Evergreen, so I expect it should be OK.

.evergreen/mongo-c-driver.spec
```diff
@@ -44,7 +44,7 @@ BuildRequires: mongodb-server
 BuildRequires: openssl
 %endif
 %if %{with libmongocrypt}
```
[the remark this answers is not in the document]
rcsanchez97: @galon1 I concur with reverting the changes to this file.  It is only updated when the downstream maintainer publishes a new version.  If you really need to modify the spec file (i.e., to keep the `rpm-package-build` task green), then you can make the changes you need and regenerate `.evergreen/spec.patch`.  Don't forget to first apply `spec.patch` to preserve the existing changes, and when you regenerate make sure that it will apply with a patch strip level of 0.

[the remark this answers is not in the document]
galon1: Just to confirm if we want to upgrade the dependency on libmongocrypt to a newer version like 1.6.0, this file doesn't need to be changed. This file is only changed when a external third party publishes a new version and we want to stay up to date with their version?

rcsanchez97:
Essentially, that is the case.  However, the Linux distro packaging
tasks are a little more complicated than that.  In this particular
instance, you don't need to do anything.  The mongo-c-driver spec calls
for `(mongocrypt) >= 1.5.2`.  However, the current target environment of
the `.evergreen/build_snapshot_rpm.sh` script is Fedora 37.  The current
version of
[libmongocrypt](https://src.fedoraproject.org/rpms/libmongocrypt/) in
Fedora 37 is 1.6.0.  So, your changes will "just work" without any
modifications to the way the RPM builds.

.evergreen/mongo-c-driver.spec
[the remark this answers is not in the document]
rcsanchez97: Essentially, that is the case.  However, the Linux distro packaging tasks are a little more complicated than that.  In this particular instance, you don't need to do anything.  The mongo-c-driver spec calls for `(mongocrypt) >= 1.5.2`.  However, the current target environment of the `.evergreen/build_snapshot_rpm.sh` script is Fedora 37.  The current
version of [libmongocrypt](https://src.fedoraproject.org/rpms/libmongocrypt/) in Fedora 37 is 1.6.0.  So, your changes will "just work" without any modifications to the way the RPM builds.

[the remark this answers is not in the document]
galon1: That makes sense. Thank you! I reverted my changes

PR #1143, opened 2022-11-09
Add test_example_59

kkloberdanz:
CDRIVER-4295

related: DRIVERS-2181

Implement a snapshot query example for the manual.


src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -2474,6 +2474,182 @@ test_example_56 (mongoc_database_t *db)
...
+                        "}",
+                        "}",
+                        "{",
+                        "$count",
+                        BCON_UTF8 ("adoptableCount"),
+                        "}",
+                        "]");
+
+   cursor = mongoc_collection_aggregate (
+      collection, MONGOC_QUERY_NONE, pipeline, NULL, NULL);
```
kevinAlbs, line 2563: To apply a `mongoc_client_session_t` to an operation, it needs to be appended to the options with `mongoc_client_session_append`.

```suggestion
   bson_t opts;

   if (!mongoc_client_session_append (cs, &opts, &error)) {
      MONGOC_ERROR ("could not apply session options");
      goto cleanup;
   }
   cursor = mongoc_collection_aggregate (
      collection, MONGOC_QUERY_NONE, pipeline, &opts, NULL);
   bson_destroy (&opts);
```

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -2474,6 +2474,182 @@ test_example_56 (mongoc_database_t *db)
...
+      value = bson_iter_value (&iter);
+      if (value->value_type == BSON_TYPE_INT32) {
+         *accumulator += value->value.v_int32;
+      } else if (value->value_type == BSON_TYPE_INT32) {
+         *accumulator += value->value.v_int64;
+      } else {
+         MONGOC_ERROR ("%s", "'adoptableCount' must be an integer");
+         rc = false;
+         goto cleanup;
+      }
```
kevinAlbs, line 2589: Suggest using `bson_iter_as_int64` to handle multiple possible numeric types.
```suggestion
      *accumulator += bson_iter_as_int64 (&iter);
```

kevinAlbs:
Looking good. Include CDRIVER-4295 in the commit message before merging. That will result in a comment with the commit being added to the Jira ticket.

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -2474,6 +2474,182 @@ test_example_56 (mongoc_database_t *db)
...
+   cs = mongoc_client_start_session (client, session_opts, &error);
+   mongoc_session_opts_destroy (session_opts);
+   if (!cs) {
+      MONGOC_ERROR ("Could not start session: %s", error.message);
+      goto cleanup;
+   }
+
+   accumulate_adoptable_count (cats_collection, &adoptable_pets_count);
+   accumulate_adoptable_count (dogs_collection, &adoptable_pets_count);
+
```
galon1, line 2638: It might be helpful to add a comment here describing that we are using an aggregation pipeline to return the count, since other drivers have been displaying the aggregation pipeline command used. Since this code example is already longer than other drivers, I think a comment is ok.

galon1 (approved):
LGTM!

src/libmongoc/tests/test-mongoc-sample-commands.c
kkloberdanz: I'm glad you brought this up, because that's something I'm curious about. It would be a fair amount of duplicate code and a bit lengthy to inline both calls to `accumulate_adoptable_count` in this function, but I'm wondering if I should do one of the following:

1. Include the entire function `accumulate_adoptable_count` in the example, i.e., place the start example comment just above the declaration of `accumulate_adoptable_count`, and end example comment stays where it is within `test_example_59`.
2. Place the text of `accumulate_adoptable_count` in a comment in `test_example_59`.
3. Explain in a comment what is happening in the function `accumulate_adoptable_count` purely in English, and omit code in the comment.
4. Inline all the code from `accumulate_adoptable_count` within `test_example_59`.

I'm leaning towards 1., but I'm not sure how that would look when rendered. I'd like to know what everyone thinks about this, since I'm not sure which way would be the cleanest for the purpose of producing an example.

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -2474,6 +2474,182 @@ test_example_56 (mongoc_database_t *db)
...
+   /* End Example 59 */
+
+   /* Start Example 59 Post */
+cleanup:
+   mongoc_collection_destroy (dogs_collection);
+   mongoc_collection_destroy (cats_collection);
+   mongoc_client_session_destroy (cs);
+   mongoc_client_destroy (client);
+   /* End Example 59 Post */
+}
```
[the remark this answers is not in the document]
kkloberdanz: I'm not quite sure how that works. I with the `Post` section, I was following the style of other examples in this file that place the cleanup code within a `Post` section. From other C examples that I saw in the documentation, it seems that cleanup code is being excluded, but I'm not sure if that is truly what we want.

For example, here's a function from the examples that leaves out the cleanup code:
https://www.mongodb.com/docs/v4.4/core/read-isolation-consistency-recency/

src/libmongoc/tests/test-mongoc-sample-commands.c
kkloberdanz: P.S. I'm planning on reaching out to Jeffrey Allen, the reporter for the original ticket, regarding this before merging.

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -2474,6 +2474,182 @@ test_example_56 (mongoc_database_t *db)
...
+ */
+static void
+test_example_59 (mongoc_database_t *db)
+{
+   /* Start Example 59 */
+   mongoc_client_t *client = NULL;
+   mongoc_client_session_t *cs = NULL;
+   mongoc_collection_t *cats_collection = NULL;
+   mongoc_collection_t *dogs_collection = NULL;
+   long long adoptable_pets_count = 0;
```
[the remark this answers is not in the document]
kkloberdanz: One note on this, some platforms define int64_t to be `long`, while others define it to be `long long`, so portable print specifiers get a little hairy. There is a portable macro `PRId64` that solves this, but the code is just a little ugly.

```C
   printf ("there are %" PRId64 " adoptable pets\n", adoptable_pets_count);
```
This macro would fix the following warning:
```
[2022/11/14 19:18:51.748] /data/mci/53a76e388c58b50897191ae23d61ac25/mongoc/src/libmongoc/tests/test-mongoc-sample-commands.c:2639:26: error: format '%lld' expects argument of type 'long long int', but argument 2 has type 'int64_t {aka long int}' [-Werror=format=]
[2022/11/14 19:18:51.748]     printf ("there are %lld adoptable pets\n", adoptable_pets_count);
[2022/11/14 19:18:51.748]                        ~~~^
[2022/11/14 19:18:51.748]                        %ld
```

Leaving the variable as a `long long` would also be portable since a `long long` is defined to be at least 64 bits. Which way would you prefer?

src/libmongoc/tests/test-mongoc-sample-commands.c
kkloberdanz: I got some feed back from Jeffery Allen. It sounds like he's in favor of option 3.

(quote from Jeff)
>Hey Kyle, thanks for reaching out. So my recollection with that ticket is that the main goal is to emphasize the snapshot functionality in the example. Therefore, I'm inclined to say that we can exclude the agg pipeline from the example, and leave a comment explaining the aggregation. That way, users aren't too overwhelmed by a large example and they'll still be able to see the snapshot behavior in action

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -2474,6 +2474,210 @@ test_example_56 (mongoc_database_t *db)
...
+    * Remember in order to apply the client session to
+    * this operation, you must append the client session to the options passed
+    * to `mongoc_collection_aggregate, i.e.,
+    *
+    * mongoc_client_session_append (cs, &opts, &error);
+    * cursor = mongoc_collection_aggregate (
+    *    collection, MONGOC_QUERY_NONE, pipeline, &opts, NULL);
+    */
+   accumulate_adoptable_count (cs, cats_collection, &adoptable_pets_count);
+   accumulate_adoptable_count (cs, dogs_collection, &adoptable_pets_count);
```
kkloberdanz, line 2658: Jeff has checked off on using this comment as a part of the documentation.

src/libmongoc/tests/test-mongoc-sample-commands.c
galon1: I agree option 3 would be the clearest, but I'm also not sure about how strict the formatting constraints are. Maybe the docs team could help make the choice.

galon1: In [this page](https://www.mongodb.com/docs/v6.0/core/transactions/) the C example is much longer than others. So it could potentially be ok.

src/libmongoc/tests/test-mongoc-sample-commands.c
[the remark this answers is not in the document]
galon1: I think if space is not an issue then it would be helpful to have it. [This page](https://www.mongodb.com/docs/v6.0/core/transactions/) for example does include it.

src/libmongoc/tests/test-mongoc-sample-commands.c
galon1: I just saw the addition of the comment! Ignore my comment above.

kkloberdanz: That's interesting. I do like when an example is explicit, since this example includes 2 functions, then I suppose no reason why I can't either. I'll make that change for now, and request another review from both you and Kevin to see what you think.

kkloberdanz:
FYI, Gil brought up that another C driver example is lengthy and has multiple functions, so it seems that if I were to include multiple functions as a part of the example, that would fit with how things are currently shown. I pushed that in order for us to compare and decide which way we would prefer.

For reference, the example that Gil found: https://www.mongodb.com/docs/v6.0/core/transactions/

galon1 (commented):
I think the example looks good! I just have one question about using python in this example.

src/libmongoc/tests/test-mongoc-sample-commands.c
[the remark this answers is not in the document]
kevinAlbs: We use `int64_t` in the codebase. I think it is safe to assume users have stdint.h. I prefer `int64_t` since it matches `bson_iter_as_int64` return.

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -2474,6 +2474,210 @@ test_example_56 (mongoc_database_t *db)
...
+      goto cleanup;
+   }
+
+   pipeline = BCON_NEW ("pipeline",
+                        "[",
+                        "{",
+                        "$match",
+                        "{",
+                        BCON_UTF8 ("adoptable"),
+                        BCON_BOOL ("true"),
```
kevinAlbs, line 2560:
```suggestion
                        BCON_BOOL (true),
```

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -2474,6 +2474,210 @@ test_example_56 (mongoc_database_t *db)
...
+      MONGOC_ERROR ("could not apply session options: %s", error.message);
+      goto cleanup;
+   }
+
+   pipeline = BCON_NEW ("pipeline",
+                        "[",
+                        "{",
+                        "$match",
+                        "{",
+                        BCON_UTF8 ("adoptable"),
```
kevinAlbs, line 2559:
```suggestion
                       "adoptable",
```

`BCON_UTF8` is for string values (not keys).

src/libmongoc/tests/test-mongoc-sample-commands.c
kevinAlbs: Defer to Jeff's judgement since this example is requested by the documentation team.

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -2474,6 +2474,210 @@ test_example_56 (mongoc_database_t *db)
...
+                            int64_t *accumulator /* OUT */
+)
+{
+   bson_t *pipeline = NULL;
+   mongoc_cursor_t *cursor = NULL;
+   bool rc = false;
+   const bson_t *doc = NULL;
+   bson_error_t error;
+   bson_iter_t iter;
+   bson_t opts;
```
kevinAlbs, line 2546:
```suggestion
   bson_t opts = BSON_INITIALIZER;
```

`mongoc_client_session_append` expects the output BSON to be initialized.

kevinAlbs (approved):
LGTM with small fixes.

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -2474,6 +2474,210 @@ test_example_56 (mongoc_database_t *db)
...
+
+   /*
+    * Perform the following aggregation pipeline, and accumulate the count in
+    * `adoptable_pets_count`.
+    *
+    *  adoptablePetsCount = db.cats.aggregate(
+    *      [ { "$match": { "adoptable": true } },
+    *        { "$count": "adoptableCatsCount" } ], session=s
+    *  ).next()["adoptableCatsCount"]
+    *
```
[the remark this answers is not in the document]
kkloberdanz: I was thinking that the Python looked enough like regular MQL mixed with pseudo code that one would be able to more clearly see what the aggregation pipeline is doing. Do you think it needs to be called out that this is from the Python example?

PR #1144, opened 2022-11-10
CDRIVER-4425 Provide access to raw result if server error occurs

galon1:
## Summary
These changes provide access to the raw result document from the server if an error has occurred. For read operations the cursor will already return the original result from the server, as the result is stored in `cursor->error_doc`. For write operations a new field in `write_result_t` was required (`raw_error_response`), which is only populated if a server error occurs. New unified tests were added according to this specification (original [new tests](https://github.com/mongodb/specifications/commit/081578a9d951a81e70e99e165a6c254be88a8959), but the latest edit is [here](https://github.com/mongodb/specifications/commit/c04f2ec111ec0e5c9cfd46e37a91ea2f3c63f0f5)).

## Notes
1. The raw responses from the server for write operations are wrapped in an `errorReplies` field. This is to avoid test failures that parse out expected fields in the response, since write operations process write results. For example, the index for bulkwrite operations are increased by 1 when the raw server result is processed. However, since the goal of this ticket was to return the raw result from the server with no processing (in case of future changes), I left the raw response in from the server, and I left the processed result, since users are expecting the processed result.
2. Tests did not include a `name` field when running `createIndex`, but a name field is required for `createIndexes` which is how the test runner executes `createIndex` commands. So if no `name` field was set, I just created a `name` field so the command can be executed.


src/libmongoc/src/mongoc/mongoc-error-private.h
```diff
@@ -85,4 +85,7 @@ _mongoc_error_is_state_change (bson_error_t *error);
 bool
 _mongoc_error_is_network (const bson_error_t *error);
 
+bool
+_mongoc_error_is_server (bson_error_t *error);
```
kkloberdanz, line 89: Since `error` is not being modified, and is therefore not an out parameter, should this instead be `const bson_error_t *error`?

src/libmongoc/tests/unified/result.c
```diff
@@ -565,6 +567,43 @@ result_check (result_t *result,
             goto done;
          }
       }
+
+      if (error_response) {
```
kkloberdanz, line 571: This isn't something that should block merging, but it looks like this function has grown to be pretty long. I wonder if we should start thinking of breaking it up into smaller pieces?

kkloberdanz (approved):
Looks good! Just a couple of minor questions, neither of which should block merging. Although, I'm still getting up to speed on the codebase, so someone else might want to double check. Good job! 👍

src/libmongoc/src/mongoc/mongoc-write-command.c
```diff
@@ -1498,6 +1509,13 @@ _mongoc_write_result_complete (
       }
    }
 
+   /* If there is a raw error response then we know a server error has occurred.
+    * We should add the raw result to the reply. */
+   if (bson && !bson_empty (&result->raw_error_response)) {
+      bson_append_document (
+         bson, "serverResponse", -1, &result->raw_error_response);
```
kevinAlbs, line 1516:
```suggestion
         bson, "errorReplies", -1, &result->raw_error_response);
```

Suggest naming "error" since this is only set when a server error occurs. And suggest using plural (along with other suggestion to store an array of server replies).

src/libmongoc/src/mongoc/mongoc-write-command.c
```diff
@@ -1324,6 +1326,14 @@ _mongoc_write_result_merge (mongoc_write_result_t *result,   /* IN */
       result->n_writeConcernErrors++;
    }
 
+   /* We have an error from the server and should append the entire reply to the
+    * result. */
+   if (result->failed || result->n_writeConcernErrors ||
+       _mongoc_error_is_server (&result->error)) {
+      bson_destroy (&result->raw_error_response);
+      bson_copy_to (reply, &result->raw_error_response);
```
kevinAlbs, line 1334: Instead of storing only the last server error reply, I suggest storing an array of all error server replies. Having all server error replies may be useful to diagnose the cause of errors in a bulk write operation.

Using an array more closely matches the returned `writeConcernErrors` field. `writeConcernErrors` is an array of all `writeConcernError` found from server error replies.

The specification tests do not test this. I suggest adding a test that receives multiple server error replies to test this behavior. Here is an example: https://github.com/kevinAlbs/mongo-c-driver/commit/50d4e9e446caaa2f6b60a8d15973bae1489a2051

src/libmongoc/tests/unified/runner.c
```diff
@@ -596,7 +596,7 @@ get_topology_type (mongoc_client_t *client)
 static void
 check_schema_version (test_file_t *test_file)
 {
-   const char *supported_version_strs[] = {"1.8"};
+   const char *supported_version_strs[] = {"1.8", "1.12"};
```
kevinAlbs, line 599:
```suggestion
   const char *supported_version_strs[] = {"1.8", "1.12" /* 1.12 is partially supported to run tests with `errorResponse` */};
```

The Unified Test Format [Changelog](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#changelog) suggests there are operations in older versions not yet supported (e.g. the `thread` entities added in 1.10).

kevinAlbs (changes requested):
Great work. I suggest storing an array of error replies to give callers access to all server errors on a bulk write.

src/libmongoc/src/mongoc/mongoc-write-command.c
```diff
@@ -1355,6 +1357,22 @@ _mongoc_write_result_merge (mongoc_write_result_t *result,   /* IN */
       result->n_writeConcernErrors++;
    }
 
+   /* If a server error ocurred, then append the raw response to the
+    * error_replies array. */
+   if (result->failed || result->n_writeConcernErrors ||
+       _mongoc_error_is_server (&result->error)) {
```
kevinAlbs, line 1363: `result->n_writeConcernErrors` is a running count in `mongoc_write_result_t`. The `mongoc_write_result_t` is the result of merging possibly many server replies. If `n_writeConcernErrors > 0`, that does not imply the current reply is an error.

Suggest adding `bool` variables to track existence of "writeErrors" and "writeConcernError" in the if blocks above, and using `_mongoc_cmd_check_ok` to check for the `'ok':0` case:

```suggestion
   /* If a server reply contains 'writeErrors', 'writeConcernError' or 'ok':0,
    * then append the raw response to the error_replies array. */
   if (has_writeErrors || has_writeConcernError ||
       !_mongoc_cmd_check_ok (
          reply, MONGOC_ERROR_API_VERSION_2, NULL /* error */)) {
```

src/libmongoc/tests/test-mongoc-bulk.c
```diff
@@ -4933,6 +4933,67 @@ test_bulk_let_multi (void)
...
+
+   ASSERT_MATCH (&reply,
+                 "{'nInserted': 2,"
+                 " 'nMatched':  0,"
+                 " 'nModified': 0,"
+                 " 'nRemoved':  1,"
+                 " 'nUpserted': 0,"
+                 " 'errorReplies': [{'n': 2, 'writeErrors': [{ 'index' : 1, "
+                 "'code': {'$exists': true}}]}, "
+                 "{'n' : 1}, {'n' : 0, 'writeErrors': [{ 'index' : 0, 'code': "
```
kevinAlbs, line 4983:
```suggestion
                 "{'n' : 0, 'writeErrors': [{ 'index' : 0, 'code': "
```

The first operation and third operation (both `inserts`) are expected to result in error. The second `delete` is expected to succeed. See other comment with a suggested fix in `_mongoc_write_result_merge`.

src/libmongoc/src/mongoc/mongoc-write-command.c
```diff
@@ -1355,6 +1357,22 @@ _mongoc_write_result_merge (mongoc_write_result_t *result,   /* IN */
       result->n_writeConcernErrors++;
    }
 
+   /* If a server error ocurred, then append the raw response to the
+    * error_replies array. */
```
jmikola, line 1361: Offhand, would the first `{ ok: 0 }` command response (post-retry) completely interrupt a bulk write operation regardless of the `ordered` option (note: `ordered: false` is essentially continue-on-error).

Our specs are ambiguous about this (see: [DRIVERS-2090](https://jira.mongodb.org/browse/DRIVERS-2090), [DRIVERS-2159](https://jira.mongodb.org/browse/DRIVERS-2159)).

I noted in the unified test runner code below, the assertion is only done on `errorReplies.0`, which assumes that three will only ever be on error reply. That's fine, as I intentionally wrote the tests with the above ambiguity in mind (see: [mongodb/specifications@33a03f2](https://github.com/mongodb/specifications/blob/33a03f2eae1f479165fbde97993afb18241204af/source/crud/tests/unified/bulkWrite-errorResponse.yml#L19).

But here, I'm more curious about whether multiple `errorReplies` are even possible in libmongoc. If not, are we just using an `errorReplies` array on the result for flexibility?

src/libmongoc/tests/unified/result.c
```diff
@@ -565,6 +567,43 @@ result_check (result_t *result,
...
+         }
+
+         /* Write operations have the raw response wrapped in a serverRespons
+          * document. Read operations return the raw response as the reply. */
+         bson_iter_t iter;
+         bson_t doc_to_match;
+         bson_init (&doc_to_match);
+
+         if (bson_iter_init_find (&iter, result->reply, "errorReplies")) {
+            bson_lookup_doc (result->reply, "errorReplies.0", &doc_to_match);
```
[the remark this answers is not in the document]
galon1: Both `bson_lookup_doc` and `bson_look_up_doc_null_ok` will abort the test if the key is absent because they both use the helper function `bson_lookup` that will abort the test. However, I can still use `bson_lookup_doc_null_ok` to conditionally check if `errorReplies.0` was empty or not.

src/libmongoc/src/mongoc/mongoc-write-command-private.h
```diff
@@ -88,6 +88,9 @@ typedef struct {
     * primary, this contains the server id of the newly selected primary. Only
     * applies to OP_MSG. Is left at 0 if no retry occurs. */
    uint32_t retry_server_id;
+   /* store the raw server response if an error occured */
+   uint32_t n_errorReplies;
+   bson_t raw_error_replies;
```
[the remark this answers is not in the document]
galon1: Yes, that change has been made.

src/libmongoc/tests/unified/result.c
```diff
@@ -565,6 +567,43 @@ result_check (result_t *result,
...
       }
+
+      if (error_response) {
+         if (!result->reply) {
+            test_set_error (
+               error, "%s", "expected error with a reply, but reply unset");
+            goto done;
+         }
+
+         /* Write operations have the raw response wrapped in a serverRespons
```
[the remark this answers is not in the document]
galon1: My mistake. `errorReplies` used to named `serverResponse`, and I forgot to update this comment. It should be clearer now

src/libmongoc/src/mongoc/mongoc-write-command.c
galon1: It is possible for libmongoc to return multiple `errorReplies`. Here is a [test](https://github.com/mongodb/mongo-c-driver/pull/1144/commits/f9ece40cb73db9921873e762e641be6d3e628e70) displaying this behavior. Since `ordered: false` continues on error when the `failPoint` is hit twice, two `errorReplies` are returned by libmongoc, as well as other errors that occurred if they did (in the test a `writeError` also was raised).

src/libmongoc/tests/test-mongoc-bulk.c
galon1: I updated this test based on the discussion above of only appending an error response to `errorReplies` if a command error has occurred or if `_mongoc_cmd_check_ok ` returns false.

src/libmongoc/tests/unified/result.c
```diff
@@ -565,6 +567,51 @@ result_check (result_t *result,
...
+            test_set_error (
+               error, "%s", "expected error with a reply, but reply unset");
+            goto done;
+         }
+
+         /* Write operations have the raw response wrapped in a errorReplies
+          * array. Read operations return the raw response as the reply. */
+         bson_iter_t iter;
+         bson_t doc_to_match;
+         // bson_init (&doc_to_match);
```
kevinAlbs, line 582:
```suggestion

```

src/libmongoc/tests/unified/operation.c
```diff
@@ -1378,35 +1378,49 @@ operation_create_index (test_t *test,
...
    char *name = NULL;
    bson_t *keys = NULL;
-   bson_t *create_indexes = NULL;
+   bool *unique = NULL;
+   bson_t *create_indexes = bson_new ();
    bson_t op_reply = BSON_INITIALIZER;
    bson_error_t op_error = {0};
    bson_t *opts = bson_new ();
+   bson_t *arguments = bson_new ();
+   bson_t *array = bson_new ();
```
kevinAlbs, line 1388:
```suggestion
   bson_t arguments;
   bson_t array;
```

A `bson_t` may be heap allocated with `bson_new()` or stack allocated. Use stack allocated `bson_t` for a child `bson_t`.

If the `bson_t` used as a child in `BSON_APPEND_DOCUMENT_BEGIN` is heap allocated, it is leaked. This is a quirk of libbson.

src/libmongoc/src/mongoc/mongoc-write-command.c
kevinAlbs: Good catch.

kevinAlbs (approved):
LGTM with leak fixed. Great work.

src/libmongoc/tests/unified/result.c
```diff
@@ -565,6 +567,50 @@ result_check (result_t *result,
...
+         }
+
+         /* Write operations have the raw response wrapped in a errorReplies
+          * array. Read operations return the raw response as the reply. */
+         bson_iter_t iter;
+         bson_t doc_to_match;
+
+         if (bson_iter_init_find (&iter, result->reply, "errorReplies")) {
+            bson_lookup_doc_null_ok (
+               result->reply, "errorReplies.0", &doc_to_match);
```
jmikola, line 585: Continuing the conversation in https://github.com/mongodb/mongo-c-driver/pull/1144#discussion_r1021030892

Sorry, but I misread `bson_lookup_doc_null_ok` and thought it gracefully allowed for missing values. I don't think there's any reason to use that, as it could still abort if the array is empty.

I could have sworn there was a function that allowed full field paths and gracefully handled missing keys. Perhaps I was thinking of `bson_iter_find_descendant`, which `bson_lookup` uses internally before it aborts on a failed resolution.

Perhaps it'd be best to use `assert_match_bson()` to check that "errorReplies.0" is a document. That function will emit a test error on failure, unlike `ASSERT_MATCH()`, which aborts. I believe you can use the `$$type`, although I don't know offhand if the function supports field paths (e.g. "errorReplies.0"), or if you might need to do something like:

```
{ "errorReplies": { "0": { "$$type": "object" } } }
```

/cc @kevinAlbs

src/libmongoc/tests/unified/result.c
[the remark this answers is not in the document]
jmikola: Continuing this conversation in the current diff: https://github.com/mongodb/mongo-c-driver/pull/1144/files#r1022215846

src/libmongoc/src/mongoc/mongoc-write-command.c
jmikola: Thanks for indulging my question. Looks like an oversight in the test runner syntax I introduced in [DRIVERS-2385](https://jira.mongodb.org/browse/DRIVERS-2385). I've made a note about this in [DRIVERS-1547](https://jira.mongodb.org/browse/DRIVERS-1547?focusedCommentId=4981812&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4981812) as that's the most likely ticket we'd use to add spec test for multiple command failures in a bulk write.

src/libmongoc/tests/unified/result.c
galon1: I just pushed changes that check if `errorReplies.0` is a document. I ended up using `BSON_ITER_HOLDS_DOCUMENT` and `BSON_ITER_HOLDS_ARRAY` to check that `errorReplies` was the expected type. I thought to use these because that is how the reply is also checked right above my code (line 496 where `error_labels_contain` is checked).

src/libmongoc/tests/unified/result.c
```diff
@@ -565,6 +567,59 @@ result_check (result_t *result,
...
+         if (!result->reply) {
+            test_set_error (
+               error, "%s", "expected error with a reply, but reply unset");
+            goto done;
+         }
+
+         /* Write operations have the raw response wrapped in a errorReplies
+          * array. Read operations return the raw response as the reply. */
+         bson_iter_t iter;
+         bson_iter_t child;
```
jmikola, line 581: Nit: you can potentially move this within the `if (bson_iter_init_find...` condition scope below since `child` isn't referenced outside of that.

jmikola (approved):
Optional suggestion, but LGTM regardless.

PR #1147, opened 2022-11-14
CDRIVER-4424: Additional error handling tests for CEC

vector-of-bool:
This adds more to the work for CDRIVER-4424, and adds two very basic error-handling tests. This already caught one minor issue in the implementation (a missing check for encryptedFields resulted in a regular collection being created without issue).


kevinAlbs (approved):
Nice, LGTM

PR #1148, opened 2022-11-15
CDRIVER-4480 Do not connect to mongocryptd if shared library is loaded

galon1:
The C driver already does not connect to mongocrypted if the shared library is loaded. So the only changes was to add the new prose test described [here](https://github.com/mongodb/specifications/commit/09f1d51d549c141c310055e6d6802bc9e297258d).


src/libmongoc/tests/test-mongoc-client-side-encryption.c
```diff
@@ -5679,6 +5679,140 @@ test_create_encrypted_collection_bad_keyId (void *unused)
...
+   BSON_ASSERT (r == 0);
+   _mongoc_usleep (1000); // wait to see if received connection
+   mongoc_socket_t *ret = mongoc_socket_accept (socket, 100);
+   if (ret) {
+      // not null received a connection and test should fail
+      bson_mutex_lock (&args->mutex);
+      args->failed = true;
+      mongoc_cond_signal (&args->cond);
+      bson_mutex_unlock (&args->mutex);
+   }
```
kevinAlbs, line 5733:
```suggestion
   if (ret) {
      // not null received a connection and test should fail
      args->failed = true;
   }
   // signal that test is complete.
   bson_mutex_lock (&args->mutex);
   mongoc_cond_signal (&args->cond);
   bson_mutex_unlock (&args->mutex);
```

Signal as soon as the listener thread stops listening. This results in the test speeding up.

src/libmongoc/tests/test-mongoc-client-side-encryption.c
```diff
@@ -5679,6 +5679,140 @@ test_create_encrypted_collection_bad_keyId (void *unused)
...
+   db = mongoc_client_get_database (client_encrypted, "db");
+   coll = mongoc_database_get_collection (db, "coll");
+   ret = mongoc_collection_insert_one (coll,
+                                       tmp_bson ("{'unencrypted': 'test'}"),
+                                       NULL /* opts */,
+                                       NULL /* reply */,
+                                       &error);
+   ASSERT_OR_PRINT (ret, error);
+   // checking for signal on thread
+   ret = mongoc_cond_timedwait (&args->cond, &args->mutex, 5000);
```
kevinAlbs, line 5801:
```suggestion
   ret = mongoc_cond_timedwait (&args->cond, &args->mutex, 5000);
   /* ret non-zero indicates an error (a timeout) */
   BSON_ASSERT (!ret);
```

vector-of-bool (approved):
LGTM

PR #1151, opened 2022-11-21
rename example to 'Snapshot Query Example 1'

kkloberdanz:


kevinAlbs (approved):
LGTM. Please include CDRIVER-4295 in the commit message when merging. That will result in a comment being added to the ticket with the commit.

PR #1156, opened 2022-12-02
Kyle/snapshot example 2

kkloberdanz:


src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -2676,6 +2675,157 @@ test_example_59 (mongoc_database_t *db)
...
+   mongoc_session_opt_t *session_opts;
+   bson_t *pipeline = NULL;
+   bson_t opts = BSON_INITIALIZER;
+   mongoc_cursor_t *cursor = NULL;
+   const bson_t *doc = NULL;
+   bool ok = true;
+   bson_iter_t iter;
+   int64_t total_sales = 0;
+
+   client = test_framework_new_default_client ();
```
kevinAlbs, line 2732: I suggest constructing `client` outside of the example comments. `test_framework_new_default_client` is not a public API function.

I missed this before. Please also update test_example_59.

kevinAlbs (approved):
LGTM with one suggested change.

PR #1157, opened 2022-12-05
Fix incorrect assumption of relative CMAKE_INSTALL_LIBDIR

knedlsepp:
While working on packaging of libbson and libmongoc in NixOS/nixpkgs, I noticed an incorrect assumption in the pkg-config files.

From the CMake docs at
https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

    > CMAKE_INSTALL_<dir>
    >
    > Destination for files of a given type. This value may be passed to
    > the DESTINATION options of install() commands for the corresponding
    > file type. It should typically be a path relative to the
    > installation prefix so that it can be converted to an absolute path
    > in a relocatable way (see CMAKE_INSTALL_FULL_<dir>). However, an
    > absolute path is also allowed.

See also:
https://github.com/jtojnar/cmake-snips#assuming-cmake_install_dir-is-relative-path

### DIFF
--- a/src/libbson/CMakeLists.txt
+++ b/src/libbson/CMakeLists.txt
@@ -394,7 +394,7 @@ endforeach ()
 
 set (VERSION "${BSON_VERSION}")
 set (prefix "${CMAKE_INSTALL_PREFIX}")
-set (libdir "\${prefix}/${CMAKE_INSTALL_LIBDIR}")
+set (libdir "${CMAKE_INSTALL_FULL_LIBDIR}")
 configure_file (
    ${CMAKE_CURRENT_SOURCE_DIR}/src/libbson-1.0.pc.in
    ${CMAKE_CURRENT_BINARY_DIR}/src/libbson-1.0.pc
--- a/src/libmongoc/CMakeLists.txt
+++ b/src/libmongoc/CMakeLists.txt
@@ -1173,7 +1173,7 @@ endif ()
 # Define pkg-config files
 set (VERSION "${MONGOC_VERSION}")
 set (prefix "${CMAKE_INSTALL_PREFIX}")
-set (libdir "\${prefix}/${CMAKE_INSTALL_LIBDIR}")
+set (libdir "${CMAKE_INSTALL_FULL_LIBDIR}")
 
 foreach (
       FLAG


rcsanchez97 (changes requested):
@knedlsepp thanks for your contribution.  I like the idea of making the directory handling more robust.  However, I think that your approach is sub-optimal.  Here is an example a comparison of a generated `.pc` files before and after your suggested change:

```
diff -uNr old/libmongoc-ssl-1.0.pc new/libmongoc-ssl-1.0.pc
--- old/libmongoc-ssl-1.0.pc	2022-11-09 09:08:48.000000000 +0000
+++ new/libmongoc-ssl-1.0.pc	2022-12-05 22:23:20.000000000 +0000
@@ -1,6 +1,6 @@
 prefix=/usr
 exec_prefix=${prefix}
-libdir=${prefix}/lib/x86_64-linux-gnu
+libdir=/usr/lib/x86_64-linux-gnu
 includedir=${exec_prefix}/include
 
 Name: libmongoc-1.0
```

To me, dropping the use of `${prefix}` in the `libdir` line is undesirable.  If I look at the resource you linked as a reference, the section following the section you linked to is titled "Concatenating paths when building pkg-config files".  In that section, it says: "This means you cannot just use the `CMAKE_INSTALL_FULL_<dir>` variables and need to construct the paths yourself."  However, your approach was to use `CMAKE_INSTALL_FULL_LIBDIR`.

If you could consider implementing something like what the author of that module recommends (e.g., like his `JoinPaths` CMake module, keeping in mind that our build must work with CMake back to version 3.1) then that would be more in line with the best practice he advocates and would allow to maintain usage of the `${prefix}` variable in the `.pc` files.

Another thing to note is that the author's `JoinPaths` CMake module does include a note in the header comment that says "Windows not supported".  Our builds also need to be able to run on Windows and to produce working pkg-config scripts for Windows.  So some investigation may be required to ensure either that this works for Windows builds or that we introduce some sort of switch or flag to make this optional.

knedlsepp:
Thank you for the review and pointing out the issues. I'll work in the improvements as soon as possible. :-)

thessem:
I am interested in the outcome of this PR. The author of this PR is probably not going to address it, as the PR is being manually applied over in the [nix repository](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/libraries/mongoc/default.nix#L53) and so the problem is solved (in a slightly brittle manner)

knedlsepp:
> I am interested in the outcome of this PR. The author of this PR is probably not going to address it, as the PR is being manually applied over in the [nix repository](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/libraries/mongoc/default.nix#L53) and so the problem is solved (in a slightly brittle manner)

Yes, I admittedly had trouble addressing the valid concerns raised with this PR. (Turns out CMake is hard to get right. 😕) And yes, the pressure for a proper upstream solution was reduced by the downstream workaround. I guess it's better to close it for now as long as there is no good solution.

PR #1158, opened 2022-12-07
CDRIVER-4275 and CDRIVER-4275

kevinAlbs:
# Summary
- Fix load balancer tasks.
- Remove `mongoc_global_mock_service_id`.
- Update load balancer DNS tests to https://github.com/mongodb/specifications/commit/5d1b42a05017ba27ffdbe6fd068a45e93e5b68c5

Resolves CDRIVER-4275 and CDRIVER-4387.

# Background & Motivation

`integration-tests.sh` and orchestration configs have been updated for load balancer support. The orchestration configs were added in https://github.com/mongodb-labs/drivers-evergreen-tools/pull/183 and later updated in https://github.com/mongodb-labs/drivers-evergreen-tools/pull/186.

The C driver maintains a modified copy of the drivers-evergreen-tools script [`run-orchestration.sh`](https://github.com/mongodb-labs/drivers-evergreen-tools) as `integration-tests.sh`. `integration-tests.sh` may be removed in CDRIVER-4029.

The JSON tests in `src/libmongoc/tests/json/initial_seedlist_dns_discovery/load-balanced` were updated to account for https://github.com/mongodb/specifications/pull/1148.

Here is a full patch build: https://spruce.mongodb.com/version/6390ffd0850e613823e658de

### DIFF
--- a/.evergreen/integration-tests.sh
+++ b/.evergreen/integration-tests.sh
@@ -61,6 +61,10 @@ if [ -z "$ORCHESTRATION_FILE" ]; then
    if [ "$SSL" != "nossl" ]; then
       ORCHESTRATION_FILE="${ORCHESTRATION_FILE}-ssl"
    fi
+
+   if [ -n "$LOAD_BALANCER" ]; then
+      ORCHESTRATION_FILE="${ORCHESTRATION_FILE}-load-balancer"
+   fi
 fi
 
 # Set up mongo orchestration home.
--- a/.evergreen/integration-tests.sh
+++ b/.evergreen/integration-tests.sh
@@ -62,7 +62,7 @@ if [ -z "$ORCHESTRATION_FILE" ]; then
       ORCHESTRATION_FILE="${ORCHESTRATION_FILE}-ssl"
    fi
 
-   if [ -n "$LOAD_BALANCER" ]; then
+   if [ "$LOAD_BALANCER" = "on" ]; then
       ORCHESTRATION_FILE="${ORCHESTRATION_FILE}-load-balancer"
    fi
 fi
--- a/.evergreen/integration-tests.sh
+++ b/.evergreen/integration-tests.sh
@@ -14,11 +14,14 @@
 # OCSP: off, on
 # REQUIRE_API_VERSION: set to a non-empty string to set the requireApiVersion parameter
 #   This is currently only supported for standalone servers
+# LOAD_BALANCER: off, on
 #
 # This script may be run locally.
 #
 
 set -o errexit  # Exit the script with error if any of the commands fail
+
+: "${LOAD_BALANCER:=off}"
 
 DIR=$(dirname $0)
 # Functions to fetch MongoDB binaries
--- a/.evergreen/integration-tests.sh
+++ b/.evergreen/integration-tests.sh
@@ -1,4 +1,4 @@
-#! /bin/sh
+#! /bin/bash
 # Start up mongo-orchestration (a server to spawn mongodb clusters) and set up a cluster.
 #
 # Specify the following environment variables:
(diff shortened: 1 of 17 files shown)


.evergreen/integration-tests.sh
```diff
@@ -61,6 +61,10 @@ if [ -z "$ORCHESTRATION_FILE" ]; then
    if [ "$SSL" != "nossl" ]; then
       ORCHESTRATION_FILE="${ORCHESTRATION_FILE}-ssl"
    fi
+
+   if [ -n "$LOAD_BALANCER" ]; then
```
rcsanchez97, line 65: I'm not super excited about using `-n` to test a variable that is set with the value `true` in `config.yml`.  This seems to create the possibility that someone could think that setting `LOAD_BALANCER=false` would disable load balancer tests, but that would clearly not happen.  I can't decide on a clearly better approach, but what about testing for `= "on"` as is done for `IPV4_ONLY`?  The the `on/off` semantics seem a bit more natural.

rcsanchez97 (approved):
LGTM.

I'll leave it up to you as to whether you want to go to the trouble of changing the semantics of the `LOAD_BALANCER` variable, but it's not worth holding up the PR so I'm going ahead and approving.

.evergreen/integration-tests.sh
eramongodb: Agreed; suggest using same pattern as `$IPV4_ONLY` check above:
```bash
if [ "$LOAD_BALANCER" = "on" ]; then
```
Also suggest adding `LOAD_BALANCER=${LOAD_BALANCER:-off}` above to document script parameters and default behavior.

Note: the `VAR=${VAR:-value}` pattern used above can be rewritten as:
```bash
: "${VAR:=value}"
```

kevinAlbs: > This seems to create the possibility that someone could think that setting LOAD_BALANCER=false would disable load balancer tests, but that would clearly not happen.

Good point. Updated to check for the value "on".

> Note: if this script is unconditionally executed in Bash, the VAR=${VAR:-value} pattern used above can be rewritten

Thank you for the suggestion. Updated to explicitly run with bash and declare the `LOAD_BALANCER` variable.

.evergreen/integration-tests.sh
```diff
@@ -61,6 +64,10 @@ if [ -z "$ORCHESTRATION_FILE" ]; then
    if [ "$SSL" != "nossl" ]; then
       ORCHESTRATION_FILE="${ORCHESTRATION_FILE}-ssl"
    fi
+
+   if [ "$LOAD_BALANCER" = "on" ]; then
```
eramongodb, line 68: Confirming, given these conditionals are all setting the same `ORCHESTRATION_FILE` variable, it is expected that only _one_ of these sequence of conditionals are true in a given invocation of `integration-tests.sh`, correct?

eramongodb (approved):

.evergreen/integration-tests.sh
kevinAlbs: It is expected that more than one condition may be true. E.g. both `AUTH` and `SSL` conditions may be true to make `ORCHESTRATION_FILE`: `auth-ssl`.

FWIW this script is a modified copy of [run-orchestration.sh](https://github.com/mongodb-labs/drivers-evergreen-tools) from drivers-evergreen-tools. I prefer not diverging behavior too much to make a transition to drivers-evergreen-tools easier.

eramongodb: Ah, I overlooked that these were appending to the `ORCHESTRATION_FILE` variable rather than overwriting it. Thank you and disregard.

PR #1161, opened 2022-12-13
Use httpbin.org instead of example.com for /http tests

eramongodb:
Resolves CDRIVER-4538. Verified by [this patch](https://spruce.mongodb.com/version/6398bf553066157ccb88b3e0/tasks).

The `/http` test has been split into `/http/get` and `/http/post` and uses `httpbin.org` instead of `example.org`.

As a drive-by fix, this PR also resolves CDRIVER-4537 by ensure calls to `abort()` in the test suite are preceeded by both `fflush (stdout)` and `fflush (stderr)` to improve the quality of test output on failures.

### DIFF
--- a/src/libmongoc/tests/test-mongoc-http.c
+++ b/src/libmongoc/tests/test-mongoc-http.c
@@ -20,12 +20,12 @@
 #include "mongoc/mongoc-http-private.h"
 
 void
-test_mongoc_http (void *unused)
+test_mongoc_http_get (void *unused)
 {
    mongoc_http_request_t req;
    mongoc_http_response_t res;
    bool r;
-   bson_error_t error;
+   bson_error_t error = {0};
 
    BSON_UNUSED (unused);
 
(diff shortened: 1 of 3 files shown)


src/libmongoc/tests/test-mongoc-http.c
```diff
@@ -34,24 +34,48 @@ test_mongoc_http (void *unused)
...
    r = _mongoc_http_send (&req, 10000, false, NULL, &res, &error);
-   ASSERT_CMPINT (res.status, ==, 200);
+   ASSERT_WITH_MSG (res.status == 200,
+                    "unexpected status code %d\n"
+                    "RESPONSE BODY BEGIN\n"
+                    "%s"
+                    "RESPONSE BODY END\n",
+                    res.status,
+                    res.body_len > 0 ? res.body : "");
    ASSERT_OR_PRINT (r, error);
```
kevinAlbs, line 48:
```suggestion
   ASSERT_OR_PRINT (r, error);
   ASSERT_WITH_MSG (res.status == 200,
                    "unexpected status code %d\n"
                    "RESPONSE BODY BEGIN\n"
                    "%s"
                    "RESPONSE BODY END\n",
                    res.status,
                    res.body_len > 0 ? res.body : "");
```

Suggest checking for an error first. An error in socket connection may result in `res.status` being `0`. The `error` may have more useful information.

src/libmongoc/tests/test-mongoc-http.c
```diff
@@ -34,24 +34,48 @@ test_mongoc_http (void *unused)
...
```
kevinAlbs, line 79:
```suggestion
   ASSERT_OR_PRINT (r, error);
   ASSERT_WITH_MSG (res.status == 200,
                    "unexpected status code %d\n"
                    "RESPONSE BODY BEGIN\n"
                    "%s"
                    "RESPONSE BODY END\n",
                    res.status,
                    res.body_len > 0 ? res.body : "");
```

kevinAlbs (commented):
Looks good with an assertion swap.

If httpbin.org is not reliable, another alternative may be needed. Perhaps a local HTTP server in Python?

kevinAlbs (approved):

PR #1171, opened 2022-12-28
Update http to https for mongoc.org links

bisht2050:


kevinAlbs (approved):
Thank you for the fix. The documentation pages should get updated after the next release.

PR #1152, opened 2022-11-28
CDRIVER-4394 Add explicit encryption support and tests for range indexes 

galon1:
# Summary

Adds support for explicit encryption with the Range Index.

- Add `mongoc_client_encryption_encrypt_range_opts` and pass to libmongocrypt.
- Add `mongoc_client_encryption_encrypt_expression`.
- Add prose tests.
- Add `ASSERT_BSONVALUE_EQ`.
- Bump test dependency of libmongocrypt to 1.7.0-alpha1.

# Background & Motivation

See https://github.com/mongodb/specifications/commit/4de4e3c1897ca07efa09d456943badbe002cd756 for a description of the API change and prose tests.

Range Index requires MongoDB server 6.2+.

The API is marked experimental due to [this comment](https://jira.mongodb.org/browse/DRIVERS-2286?focusedCommentId=4975162&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4975162).

PR #1175, opened 2023-01-10
CDRIVER-4394 Fix libmongocrypt branch in compile-test-gcpkms.sh

eramongodb:
Followup to https://github.com/mongodb/mongo-c-driver/pull/1152. The git clone command for libmongocrypt in `compile-test-gcpkms.sh` appears to have been missed.

As a drive-by improvement, ensured all libmongocrypt git clone commands use `--depth 1` for consistency.


kevinAlbs (approved):
Thank you for the fix.

PR #1184, opened 2023-01-19
fix unused function warning

kevinAlbs:
# Summary

- Fix an unused function warning when building on macOS

# Background & Motivation

Resolves the following warning when building on macOS with secure transport:

```
/Users/kevin.albertson/review/mongo-c-driver-1152/src/libmongoc/tests/test-mongoc-stream-tls.c:281:1: warning: unused function 'test_mongoc_tls_common_name' [-Wunused-function]
    test_mongoc_tls_common_name (void)
    ^
    1 warning generated.
```

The caller of `test_mongoc_tls_common_name` function [is not defined](https://github.com/mongodb/mongo-c-driver/blob/9e18dd1b08591242a4da33be1f55f3a33c629d07/src/libmongoc/tests/test-mongoc-stream-tls.c#L422-L425) with secure transport.

### DIFF
--- a/src/libmongoc/tests/test-mongoc-stream-tls.c
+++ b/src/libmongoc/tests/test-mongoc-stream-tls.c
@@ -276,7 +276,7 @@ test_mongoc_tls_expired (void)
    ASSERT_CMPINT (sr.result, ==, SSL_TEST_SUCCESS);
 }
 
-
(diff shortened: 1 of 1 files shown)


kevinAlbs:
Resolved by https://github.com/mongodb/mongo-c-driver/pull/1205/files#diff-c2d61efcede75585f1d947157163a445210d6dcfa56df9fd13c5eea11c349c62

PR #1185, opened 2023-01-20
Enable parallel compilation with MinGW and fix gmtime_s detection

eramongodb:
Verified by [this patch](https://spruce.mongodb.com/version/63cab952d1fe07422f8ed3dc/tasks).

Increases the speed of MinGW compilation on Windows by passing the `-j` flag to ~CMake~ MinGW Make. No limit on concurrent processes is set in order to maximize use of the resources available on Evergreen.

As a drive-by fix, updated feature detection for MSVC-style `gmtime_s` in kms_request.c to be consistent with the logic used in [bson-iso8601.c](https://github.com/eramongodb/mongo-c-driver/blob/80333a1f92889a10fe4ee1b77285050357613757/src/libbson/src/bson/bson-iso8601.c#L310), as it may otherwise fail when building with MinGW depending on the build configurations used.


kevinAlbs (approved):

eramongodb:
Fixed PR description to indicate the `-j` flag is being passed to MinGW Make, not CMake. System-available CMake versions on Windows distros are currently 3.11, which do not support `-j` yet (introduced in 3.12).

As an additional drive-by fix, added default `CMAKE_CXX_STANDARD` to CMake configs when C++ is enabled to address [MinGW failure to compile `cpp_check.cpp` presumably due to pre-C++11 defaults](https://parsley.mongodb.com/evergreen/mongo_c_driver_mingw_debug_compile_nosasl_nossl_patch_80333a1f92889a10fe4ee1b77285050357613757_63cab952d1fe07422f8ed3dc_23_01_20_15_54_59/0/task?bookmarks=0,6554&selectedLine=834). Because `test-libmongoc` happens to be generated prior to build failure, and the BAT file does not have error handling to exit on failure, the task happened to continue successfully. This will be addressed in a separate PR.

Verified by [this patch](https://spruce.mongodb.com/task/mongo_c_driver_mingw_debug_compile_nosasl_nossl_patch_80333a1f92889a10fe4ee1b77285050357613757_63cac7d4d1fe07422f8f029b_23_01_20_16_56_53/logs?execution=0).

kevinAlbs (approved):

PR #1190, opened 2023-01-25
CDRIVER-3620 Reorganize Evergreen scripts and files under .evergreen

eramongodb:
This PR is related to CDRIVER-3620. Verified by [this patch](https://spruce.mongodb.com/version/63d1820d850e617adf059f41/tasks).

To setup for the introduction of a new set of config generator scripts that use [shrub.py](https://github.com/evergreen-ci/shrub.py), the files under `.evergreen` have been reorganized into a `scripts` and `etc` directory to distinguish scripts executed (`scripts`) or used (`etc`) by Evergreen, clearing up the directory for `*.yml` files only. The `ocsp` directory has been left alone for symmetry with the `.evergreen/oscp` directory in Drivers Evergreen Tools. Other `*.yml` files such as `.codecov.yml` and `benchmark.yml` have also been left alone to avoid requiring changes to Evergreen Project settings.

```
.evergreen/                  # All files related to Evergreen.
├── .codecov.yml
├── benchmark.yml
├── config.yml               # Includes legacy-config.yml.
├── etc/                     # Misc. non-executed files used by Evergreen.
├── legacy-config.yml        # Generated by legacy_config_generator.
├── legacy_config_generator/ # Current config generator scripts.
├── ocsp/                    # Symmetry with DET .evergreen/ocsp.
└── scripts/                 # Scripts executed by Evergreen tasks.
```

Current config generation scripts have been moved from `build` into `.evergreen/legacy_config_generator`. The legacy config generator scripts now generate a `legacy-config.yml` file that is included by `config.yml` rather than generating `config.yml` directly. This is to accomodate the upcoming introduction of more `*.yml` files that will be generated by shrub.py and included by `config.yml`. This will facilitate the incremental conversion of Evergreen config content from the legacy config generator into the shrub.py config generator.

The grouping of to-be-executed Evergreen scripts permits the slight simplification of the `windows fix` and `make files executable` Evergreen functions to apply `tr -d '\r'` and `chmod +x` to files under `.evergreen/scripts` instead of filtering for `*.sh` files under `.evergreen`.

Any and all changes to the current Evergreen configuration are (intended) to accomodate the relocation of files only.

PR #1191, opened 2023-01-26
CDRIVER-3620 Relocate legacy-config.yml into .evergreen/generated_configs

eramongodb:
Minor followup to https://github.com/mongodb/mongo-c-driver/pull/1190.

After some reconsideration, I think it is preferable that _all_ generated `*.yml` files, including `legacy-config.yml`, be grouped under the same `generated_configs` directory. This means any `*.yml` file _not_ under `generated_configs` will be safe to directly modify (only `config.yml`, `.codecov.yml`, and `benchmark.yml`), and conversely, `*.yml` files under `generated_configs` will not require `THIS FILE IS GENERATED DO NOT MODIFY`-esque warnings, as their location makes it apparent.


kevinAlbs (approved):
Seems reasonable. That may help prevent accidental manual modification of generated configs.

PR #1196, opened 2023-01-30
CDRIVER-4565 Use clock_gettime_nsec_np instead of mach_absolute_time

eramongodb:
Resolves(?) CDRIVER-4565. Verified by [this patch](https://spruce.mongodb.com/version/63d803720ae60614180f6406/tasks).

Per [Apple documentation](https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time) for `mach_absolute_time`:

> Prefer to use the equivalent `clock_gettime_nsec_np(CLOCK_UPTIME_RAW)` in nanoseconds.

Simplifies the implementation of `bson_get_monotonic_time` on MacOS and hopefully addresses the occasional `t2 >= t1` assertion violation observed on MacOS by delegating the required ~intermediate floating-point arithmetic~ unit conversion to the implementation of `clock_gettime_nsec_np(CLOCK_UPTIME_RAW)`.

~To fully avoid the intermediate floating-point arithmetic,~ We may want to consider using `CLOCK_MONOTONIC_RAW` instead of `CLOCK_UPTIME_RAW`. Per [MAN pages](https://www.unix.com/man-page/mojave/3/CLOCK_GETTIME/):

> CLOCK_MONOTONIC_RAW: clock that increments monotonically, tracking the time since an arbitrary point like CLOCK_MONOTONIC.  However, this clock is unaffected by frequency or time adjustments.  It should not be compared to other system time sources.

> CLOCK_UPTIME_RAW: clock that increments monotonically, in the same manner as CLOCK_MONOTONIC_RAW, but that does not increment while the system is asleep.  The returned value is identical to the result of mach_absolute_time() after the appropriate mach_timebase conversion is applied.

Using `CLOCK_MONOTONIC_RAW` would be equivalent to using [mach_continuous_time](https://developer.apple.com/documentation/kernel/1646199-mach_continuous_time) instead of `mach_absolute_time`. It is unclear to me why `mach_absolute_time` is currently being used instead of `mach_continuous_time`.

### DIFF
--- a/src/libbson/src/bson/bson-clock.c
+++ b/src/libbson/src/bson/bson-clock.c
@@ -126,18 +126,8 @@ bson_get_monotonic_time (void)
    clock_gettime (CLOCK_MONOTONIC, &ts);
    return (((int64_t) ts.tv_sec * 1000000) + (ts.tv_nsec / 1000));
 #elif defined(__APPLE__)
-   static mach_timebase_info_data_t info = {0};
-   static double ratio = 0.0;
-
-   if (!info.denom) {
-      /* the value from mach_absolute_time () * info.numer / info.denom
-       * is in nano seconds. So we have to divid by 1000.0 to get micro
-       * seconds*/
-      mach_timebase_info (&info);
-      ratio = (double) info.numer / (double) info.denom / 1000.0;
-   }
-
-   return mach_absolute_time () * ratio;
+   const uint64_t nsec = clock_gettime_nsec_np (CLOCK_UPTIME_RAW);
+   return (int64_t) (nsec / 1000u);
 #elif defined(_WIN32)
    /* Despite it's name, this is in milliseconds! */
    int64_t ticks = GetTickCount64 ();


src/libbson/src/bson/bson-clock.c
```diff
@@ -126,18 +126,8 @@ bson_get_monotonic_time (void)
...
-
-   if (!info.denom) {
-      /* the value from mach_absolute_time () * info.numer / info.denom
-       * is in nano seconds. So we have to divid by 1000.0 to get micro
-       * seconds*/
-      mach_timebase_info (&info);
-      ratio = (double) info.numer / (double) info.denom / 1000.0;
-   }
-
-   return mach_absolute_time () * ratio;
```
kevinAlbs, line 140: >  by delegating the required intermediate floating-point arithmetic to the implementation of clock_gettime_nsec_np(CLOCK_UPTIME_RAW).

Does this suggest multiplying by `ratio` could be the cause of a decrease in return values? If so, that surprises me.

kevinAlbs:
> It is unclear to me why `mach_absolute_time` is currently being used instead of `mach_continuous_time`.

I am not sure either.

Apple libc is open source [here](https://opensource.apple.com/releases/). The libc linked to macOS 13.2 appear to [do no floating point arithmetic](https://github.com/apple-oss-distributions/Libc/blob/Libc-1534.81.1/gen/clock_gettime.c#L110-L129):

```c
    switch(clock_id){
    case CLOCK_MONOTONIC_RAW:
        mach_time = mach_continuous_time();
        break;
    case CLOCK_MONOTONIC_RAW_APPROX:
        mach_time = mach_continuous_approximate_time();
        break;
    case CLOCK_UPTIME_RAW:
        mach_time = mach_absolute_time();
        break;
    case CLOCK_UPTIME_RAW_APPROX:
        mach_time = mach_approximate_time();
        break;
    case CLOCK_THREAD_CPUTIME_ID:
        mach_time = __thread_selfusage();
        break;
    default:
        errno = EINVAL;
        return 0;
    }

    return (mach_time * tb_info.numer) / tb_info.denom;
```

kevinAlbs (approved):
This seems like a positive change regardless. I am not sure if this will resolve the test failure.

src/libbson/src/bson/bson-clock.c
eramongodb: I suspect there may be some floating-point shenanigans due to the potentially large magnitudes of values being converted to and computed in the floating-point domain (coarseness? rounding?) before being converted back into the integral domain.

eramongodb (commented):
> The libc linked to macOS 13.2 appear to do no floating point arithmetic

Whoops, I misunderstood the fields of `mach_timebase_info_data_t` to be `double` due to the current implementation. But that is even better.

PR #1178, opened 2023-01-12
Assert bounds of tmp->tm_mon to address scan-build warnings

eramongodb:
Observed a [surprise task failure](https://parsley.mongodb.com/evergreen/mongo_c_driver_darwin_debug_compile_scan_build_patch_21275453cd46253b9d252fdfc319c1880b3dca3a_63c04dcfd6d80a4b2c695915_23_01_12_18_13_36/0/task?bookmarks=0,994&selectedLine=331) during a patch build. I do not expect `tmp->tm_mon < MONSPERYEAR` to ever be violated given the context, but to satisfy scan-build, I elected to add the condition to the loop anyways. Verified by [this patch](https://spruce.mongodb.com/version/63c051db3066157a49372e66/tasks).

PR #1201, opened 2023-02-03
CDRIVER-3620 Add SASL, CSE, and Sanitizers matrices

eramongodb:
## Description

This PR is part of CDRIVER-3620.

Please refer to [this patch](https://spruce.mongodb.com/version/63dd69832fbabe50d8864a33/tasks) to verify the proposed Evergreen matrix (at the time of PR creation). New patches will be posted as changes are made to the matrices as part of this PR review.

The objective of this PR is to evaluate the set of compile and test tasks currently captured by the new config generator under `components.sasl`, `components.cse`, `components.sanitizers`, and `components.std`. At the time of PR creation, the matrices of test tasks are _the same as_ the current set of tasks and mapping to distros in order to facilitate their review (status quo) and subsequent modifications (desired matrices).

Note: with the exception of the `macos-1100-arm64` distro, compile tasks that do not have a dependent test task have been removed in an effort to avoid redundancies. Compile coverage can be added back in via modifications to appropriate `COMPILE_MATRIX` tables as part of this PR.

## Matrix Design

There are several notable aspects and/or goals to the proposed matrices.

### One Tag for All Tasks per Variant

The current Evergreen config (mostly) defines build variants that maps to a given distro (i.e. ubuntu1804-large). A common compile task (i.e. debug-compile-sasl-openssl) is defined and included in multiple variants, and a common test task (i.e. test-latest-server-auth-sasl-openssl) is _separately_ defined and _separately_ included in multiple variants accordingly. This makes it difficult to track what test task is being run on which variant, especially given the liberal use of generic tags to specify included tasks (e.g. what tasks are included by `.latest .nossl`?).

Furthermore, it is difficult to verify that a required compile tasks is being included in the same variant, as Evergreen does not enforce a dependent task is included in the same variant. It is also difficult to confirm whether a given set of tasks is being included in _any_ variant, as Evergreen does not enforce a defined task is included in at least one variant (e.g. none of the `test-.*-sharded-.*-cse` tasks defined in the current Evergreen configuration are included in any variant: there is no test coverage of CSE with sharded clusters!).

The new matrix components under sasl, cse, sanitizers, and c11 define a _single_ build variant for each logical matrix (i.e. by SSL library) and includes _all_ tasks in the given variant using a _single_ tag unique to each matrix (the `TAG` variable defined at the top of each matrix component; typically the same name as the variant). For example, in `components.sasl.openssl`:

```python
SSL = 'openssl'
TAG = f'sasl-matrix-{SSL}'

# ...

def variants():
    expansions = {
        'DEBUG': 'ON'
    }

    return [
        BuildVariant(
            name=TAG,
            display_name=TAG,
            tasks=[EvgTaskRef(name=f'.{TAG}')], # <-- !!
            expansions=expansions,
        ),
    ]
```

The new matrix components also ensures that _every_ task has a unique name across _all_ variants to ensure any dependent task required by the variant (i.e. `sasl-auto-openssl-ubuntu1804-gcc-test-latest-server-auth` depends on `sasl-auto-openssl-ubuntu1804-gcc-compile`) will trigger Evergreen validation failure if missing.

### Distro-Specific Tasks

Instead of defining the distro to be executed on in the build variant definition (requiring N build variants for N distros as in the current Evergreen configuration), the distro is defined at the _task_ level. Not only does this allow for better logical grouping of features being tested by each task (similar tasks are in the same variant instead of split across multiple variants), but the task names are also refactored to better express logical groupings. For example, in the `sasl-matrix-openssl` variant, tasks that test `SASL=AUTO` + `SSL=OPENSSL` on Ubuntu 18.04 are lexicographically grouped as follows:

```
sasl-auto-openssl-ubuntu1804-gcc-compile
sasl-auto-openssl-ubuntu1804-gcc-test-4.0-replica-auth
sasl-auto-openssl-ubuntu1804-gcc-test-4.0-replica-noauth
sasl-auto-openssl-ubuntu1804-gcc-test-4.0-server-auth
sasl-auto-openssl-ubuntu1804-gcc-test-4.0-server-noauth
...
```

An effort is made to use higher-level categories (i.e. `cse` -> `sasl` -> ssl`) as prefixes and lower-level categories (i.e. server version -> topology -> authentication) to faciliate nice lexicographical sorting and grouping of task names.

### Sanitizers

The `sanitizers-matrix-asan` variant enables UBSAN alongside ASAN:

```yaml
expansions = {
    'ASAN': 'on',
    # ...
    'EXTRA_CONFIGURE_FLAGS': '-DENABLE_EXTRA_ALIGNMENT=OFF',
    'SANITIZE': 'address,undefined',
}
```

This is a discrepancy from the current Evergreen configuration. This makes the `debug-compile-ubsan` task redundant. The `debug-compile-ubsan-with-extra-alignment` task has been dropped due lack of runtime testing after https://github.com/mongodb/mongo-c-driver/commit/f4a83c084088e3878097b17881458548bc621ea2 and false-positive misaligned address failures being observed during execution of the extended test suite.

The `DEBUG: ON` expansion has also been dropped when running compile tasks in sanitizers matrices. This means CMake is configured with a build type of `RelWithDebInfo` instead of `Debug` to offset the performance penalty of compiling with ASAN+UBSAN or TSAN.

### Expansions

Expansions (that are treated as environment variables when a subprocess command is executed with `add_expansions_to_env`) that are common to all tasks in a variant are defined by the variant. Expansions that are specific to a given task (i.e. `CC` for compiler selection) are defined by the task. An effort is made to define environment variables required by a given task in the Evergreen configuration as expansions via `update_expansions()` or `env=` to reduce complexity of the executed script itself.

A notable exception are expansions that are expected to be prepended/appended to. For example, `EXTRA_CONFIGURE_FLAGS` in `components.sanitizers.asan_cse` is defined both by the variant:

```python
'EXTRA_CONFIGURE_FLAGS': '-DENABLE_EXTRA_ALIGNMENT=OFF'
```

and by the function:

```python
bash_exec(
    # ...
    script='EXTRA_CONFIGURE_FLAGS="-DENABLE_PIC=ON -DENABLE_CLIENT_SIDE_ENCRYPTION=ON ${EXTRA_CONFIGURE_FLAGS}" .evergreen/scripts/compile.sh',
    # ...
```

`env=` does not play well with `add_expansions_to_env` when a given expansion is already defined: the existing expansion takes precedence over the definition provided by `env=`. The variables is therefore defined/expanded in the script instead.

## Evergreen Project Settings

Once this PR is merged, the Evergreen validator will emit the following warning:

```
WARNING: Github PR alias matching variant regexp 'asan-ubuntu$' has no matching variants
WARNING: Patch alias matching variant regexp 'asan-ubuntu$' has no matching variants
```

The Evergreen project settings will have to be updated to match against the new `sanitizers-matrix-asan` variant instead.

## Followup Work

Unfortunately not many current by-distro variants could be removed yet due to tasks still assigned to them. In particular, relocating `release-compile`, IPv4/IPv6, and authentication tasks should allow for several by-distro variants to be removed. However, this work is deferred to a later date due to priorities and scheduling.


.evergreen/generated_configs/functions.yml
```diff
@@ -70,6 +70,90 @@ functions:
...
+        env:
+          COMPILE_LIBMONGOCRYPT: "ON"
+        args:
+          - -c
+          - EXTRA_CONFIGURE_FLAGS="-DENABLE_PIC=ON -DENABLE_CLIENT_SIDE_ENCRYPTION=ON ${EXTRA_CONFIGURE_FLAGS}" .evergreen/scripts/compile.sh
+  cse-sasl-off-openssl-compile:
+    - command: expansions.update
+      params:
+        updates:
+          - { key: SSL, value: OPENSSL }
```
rcsanchez97, line 145: I'm not sure what's going on here, but [this task](https://spruce.mongodb.com/task/mongo_c_driver_sanitizers_matrix_asan_asan_cse_sasl_off_openssl_ubuntu1804_clang_compile_patch_cb616d9b0b95ac5d2a0a464c9ec5befcff3ccd93_63dd69832fbabe50d8864a33_23_02_03_20_07_33/logs?execution=0) still seems to be compiling with SASL, rather than without it:

```
[2023/02/03 20:09:30.334] -- Searching for sasl/sasl.h
[2023/02/03 20:09:30.334] --   Found in /usr/include
[2023/02/03 20:09:30.334] -- Searching for libsasl2
[2023/02/03 20:09:30.336] --   Found /usr/lib/x86_64-linux-gnu/libsasl2.so
[2023/02/03 20:09:30.336] -- Looking for sasl_client_done
[2023/02/03 20:09:30.527] -- Looking for sasl_client_done - found
[2023/02/03 20:09:30.527] -- Check size of socklen_t
[2023/02/03 20:09:30.527] -- Check size of socklen_t - done
[2023/02/03 20:09:30.611] -- Looking for res_nsearch
[2023/02/03 20:09:30.611] -- Looking for res_nsearch - found
[2023/02/03 20:09:30.611] -- Looking for res_ndestroy
[2023/02/03 20:09:30.743] -- Looking for res_ndestroy - not found
[2023/02/03 20:09:30.743] -- Looking for res_nclose
[2023/02/03 20:09:30.743] -- Looking for res_nclose - found
[2023/02/03 20:09:30.856] -- Looking for sched_getcpu
[2023/02/03 20:09:30.856] -- Looking for sched_getcpu - not found
[2023/02/03 20:09:30.856] -- Detected parameters: accept (int, struct sockaddr *, socklen_t *)
[2023/02/03 20:09:30.857] -- Searching for compression library header snappy-c.h
[2023/02/03 20:09:30.857] --   Found in /usr/include
[2023/02/03 20:09:30.857] -- Searching for libsnappy
[2023/02/03 20:09:30.865] --   Found /usr/lib/x86_64-linux-gnu/libsnappy.so
[2023/02/03 20:09:30.865] -- Performing Test MONGOC_HAVE_SS_FAMILY
[2023/02/03 20:09:30.941] -- Performing Test MONGOC_HAVE_SS_FAMILY - Success
[2023/02/03 20:09:30.941] -- Compiling against OpenSSL
[2023/02/03 20:09:30.941] -- Compiling against Cyrus SASL
```

I am not positive that this output corresponds to this particular task definition, but the task name is `*-sasl-off-*`, which implies that it should be compiling without SASL.

eramongodb: This output is coming from libmongocrypt compiling libmongoc (for tests) alongside libbson (what it wants). The output is [also present](https://parsley.mongodb.com/evergreen/mongo_c_driver_asan_ubuntu_debug_compile_asan_openssl_cse_8b5a14e5a35fb7b365a1a53f24c933f8c497f337_23_02_03_22_40_36/0/task?bookmarks=0,2604&shareLine=457) in the [current corresponding task](https://spruce.mongodb.com/task/mongo_c_driver_asan_ubuntu_debug_compile_asan_openssl_cse_8b5a14e5a35fb7b365a1a53f24c933f8c497f337_23_02_03_22_40_36/logs?execution=0). This suggests we can add some environment variables to pass to libmongocrypt to avoid redundant compilation.

rcsanchez97: Well, that's odd and unexpected.  It might be good to consider that sort of improvement in a follow-up PR.

rcsanchez97 (approved):
LGTM

.evergreen/config_generator/components/sasl/winssl.py
```diff
@@ -0,0 +1,100 @@
...
+    ('windows-64-vs2015', 'vs2015x86', None, 'off',   ['noauth', 'auth'], ['server'], [       '4.0', '4.2',                       ]),
+    ('windows-64-vs2017', 'vs2017x86', None, 'off',   ['noauth', 'auth'], ['server'], [                                   'latest']),
+
+    ('windows-64-vs2013', 'vs2013x64', None, 'cyrus', ['noauth', 'auth'], ['server'], [       '4.0', '4.2',                       ]),
+    ('windows-64-vs2015', 'vs2015x64', None, 'cyrus', ['noauth', 'auth'], ['server'], ['3.6', '4.0', '4.2',                       ]),
+    ('windows-64-vs2017', 'vs2017x64', None, 'cyrus', ['noauth', 'auth'], ['server'], [                     '4.4', '5.0', 'latest']),
+
+    ('windows-64-vs2017', 'mingw',     None, 'sspi',  [          'auth'], ['server'], [                                   'latest']),
+    ('windows-64-vs2017', 'vs2017x64', None, 'sspi',  [          'auth'], ['server'], [                                   'latest']),
+    ('windows-64-vs2017', 'vs2017x86', None, 'sspi',  [          'auth'], ['server'], [                                   'latest']),
```
eramongodb, line 40: @rcsanchez97 noted that, as currently proposed, SSPI is only being tested with WinSSL. The legacy config [used to have](https://spruce.mongodb.com/version/mongo_c_driver_8b5a14e5a35fb7b365a1a53f24c933f8c497f337/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&taskName=sspi&variant=windows) compile coverage of SSPI with NoSSL and OpenSSL on some Windows distros, but they had no corresponding test tasks and were thus (currently) removed.

We could either reintroduce `sasl-sspi-(nossl|openssl|openssl-static)-compile` compile coverage, or opt to remove the `SaslSspi(NoSSL|OpenSSL|StaticOpenSSL)Compile` classes from the other SSL matrices (currently defined but not being used).

@kevinAlbs Thoughts on compile/test coverage of SSPI with SSL configs other than WinSSL?

.evergreen/config_generator/components/std/std_c11.py
```diff
@@ -0,0 +1,52 @@
...
+# fmt: off
+COMPILE_MATRIX = [
+    ('archlinux',  'clang', None,   ['auto']),
+    ('debian81',   'clang', None,   ['auto']),
+    ('debian92',   'clang', None,   ['auto']),
+    ('ubuntu1604', 'clang', 'i686', ['auto']),
+    ('ubuntu1604', 'clang', None,   ['auto']),
+    ('ubuntu1804', 'clang', 'i686', ['auto']),
+    ('ubuntu1804', 'gcc',   None,   ['auto']),
+]
```
rcsanchez97, line 22: Given that even recent compiler versions have made substantial improvements in c11 support, I think it would make sense to add ubuntu2004 and debian10/debian11 to the matrix.

```suggestion
    ('debian10',   'clang', None,   ['auto']),
    ('debian10',   'gcc',   None,   ['auto']),
    ('debian11',   'clang', None,   ['auto']),
    ('debian11',   'gcc',   None,   ['auto']),
    ('ubuntu2004', 'clang', None,   ['auto']),
    ('ubuntu2004', 'gcc',   None,   ['auto']),
]
```

Also, the gcc in both Debian 10 and 11 supports c17.  Do we need to think about adding a configuration to test with `-std=c17`?

rcsanchez97 (approved):
One additional comment/suggestion, but otherwise still LGTM

.evergreen/config_generator/components/sasl/winssl.py
kevinAlbs: I expect compiling with SSPI without an SSL to be a very rare use case. SSPI is required for enterprise authentication. I expect typical users of authentication are also using SSL.

I am less sure about testing with `openssl` and `openssl-static`. But I prefer leaving them out unless there is a known need or use case motivating the additional tasks.

kevinAlbs (changes requested):
Fantastic work. The concise matrix form makes it significantly easier to audit the tasks being run.

Most suggestions are based on these general opinions:
- It is OK to have compile tasks that do not have dependent test tasks. That may validate some of the many ways to compile the C driver without paying the cost of running the full test suite.
- Running all tests with multiple Linux distros on the same architecture does not add much value.
- Running all tests with multiple Visual Studio versions does not add much value.
- Testing with `nossl` and `noauth` provides little value. I expect the vast majority of users are using TLS and auth. TLS and auth are required in MongoDB Atlas.
- Testing with `sasl=off` provides no value. Disabling sasl does not appear to provide any additional test coverage. Checks for `sasl=off` in [test-libmongoc.c](https://github.com/mongodb/mongo-c-driver/blob/8b5a14e5a35fb7b365a1a53f24c933f8c497f337/src/libmongoc/tests/test-libmongoc.c#L2776-L2778) and [run-auth-test.sh](https://github.com/mongodb/mongo-c-driver/blob/8b5a14e5a35fb7b365a1a53f24c933f8c497f337/.evergreen/scripts/run-auth-tests.sh#L142-L164) remove tests when `sasl=off`
- Tests with `5.0` and `latest` should include `6.0`. Not including `6.0` may have been an oversight of CDRIVER-4370.
- Testing `replica` adds value. A significant amount of tests requiring an oplog are skipped on the `server` topology (e.g. Transactions and Queryable Encryption). Testing on all topologies is suggested in the drafted [CI Matrix Policy](https://docs.google.com/document/d/1KLB9tWEBai3mvgAP19zI65MtmZbHxw0DNIvSdIKJbTs/edit). I expect adding tests for `sharded` may involve more investigation.

eramongodb:
Updates to matrices have been applied. Verified by [this patch](https://spruce.mongodb.com/version/63e54ee0a4cf477477551ed5/tasks).

Notes:

- [cse-sasl-cyrus-darwinssl-macos-1014-clang-test-latest-replica-auth](https://spruce.mongodb.com/task/mongo_c_driver_cse_matrix_darwinssl_cse_sasl_cyrus_darwinssl_macos_1014_clang_test_latest_replica_auth_patch_8b5a14e5a35fb7b365a1a53f24c933f8c497f337_63e54ee0a4cf477477551ed5_23_02_09_19_52_01/logs?execution=0) is failing due to: 

```
error initializing cipher: Unimplemented (-4305). CTR mode is only supported on macOS 10.15+
```

- [cse-sasl-cyrus-darwinssl-macos-1014-clang-test-6.0-replica-auth](https://spruce.mongodb.com/task/mongo_c_driver_cse_matrix_darwinssl_cse_sasl_cyrus_darwinssl_macos_1014_clang_test_6.0_replica_auth_patch_8b5a14e5a35fb7b365a1a53f24c933f8c497f337_63e54ee0a4cf477477551ed5_23_02_09_19_52_01/logs?execution=0) is failing due to:

```
Error in "encryptedFields is preferred over jsonSchema" test : expected 1 keys, not 0
Expected:
{ "0" : { "_id" : 1, "encryptedIndexed" : "123" } }
Actual:
{ }
```

Possibly due to the following warning generated by the mock KMIP server(?):

```
SAWarning: TypeDecorator EnumType() will not produce a cache key because the ``cache_ok`` attribute is not set to True.  This can have significant performance implications including some performance degradations in comparison to prior SQLAlchemy versions.  Set this attribute to True if this type object's state is safe to use in a cache key, or False to disable this warning. (Background on this error at: https://sqlalche.me/e/14/cprf)
```

- Moved Static OpenSSL matrix into its own `openssl_static_compile` component accompanied by a simplified `scripts/compile-openssl-static.sh` script.
- Moved C Standard compatibility matrix into its own `c_std_compile` component accompanied by a simplified `scripts/compile-std.sh` script.

The use of separate, simpler scripts for separate components is intended to better separate concerns.

- As a drive-by fix, ensured the `MARCH` variable is actually being used again after mistakenly removing the addition of the `"-m32" "-march=i386"` compile flags when `${MARCH} == i686`. I am considering redefining the `MARCH` flag into a more understandable `USE_32_BIT_ARCH` option instead.

- As a drive-by fix, silenced a benign "file exists" warning generated during `upload-test-results`.

.evergreen/config_generator/components/cse/winssl.py
```diff
@@ -0,0 +1,76 @@
...
+    ('windows-64-vs2015', 'vs2015x64', None, ['cyrus']),
+    ('windows-64-vs2017', 'vs2017x64', None, ['cyrus']),
+]
+
+# TODO (CDRIVER-3789): test cse with the 'sharded' topology.
+TEST_MATRIX = [
+    ('windows-64-vs2017', 'vs2017x64', None, 'cyrus', ['auth'], ['server'], ['4.2', '4.4', '5.0']),
+
+    # Test 6.0+ with a replica set since Queryable Encryption does not support the 'server' topology.
+    ('windows-64-vs2017', 'vs2017x64', None, 'cyrus', ['auth'], ['server', 'replica' ], ['6.0', 'latest']),
```
eramongodb, line 26: The [combination of CSE + 6.0 and latest tests](https://spruce.mongodb.com/version/63e5517461837d4373055658/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&taskName=cse-.%2A-vs2017-x64-test) seems to exceed the current exec_timeout_secs of 1 hour. Should we bump exec_timeout_secs for Windows test tasks up to 2 hours? Or can add `--no-fork` when executing `test-libmongoc` to speed up execution? (Or another workaround?)

.evergreen/config_generator/components/cse/openssl.py
```diff
@@ -0,0 +1,101 @@
...
+    ('ubuntu1604',        'clang',     None, ['cyrus']),
+    ('ubuntu1804-arm64',  'gcc',       None, ['cyrus']),
+    ('ubuntu1804',        'gcc',       None, ['cyrus']),
+    ('ubuntu2004',        'gcc',       None, ['cyrus']),
+    ('windows-64-vs2017', 'vs2017x64', None, ['cyrus']),
+]
+
+# TODO (CDRIVER-3789): test cse with the 'sharded' topology.
+TEST_MATRIX = [
+    ('rhel83-zseries',    'gcc',       None, 'cyrus', ['auth'], ['server'], ['4.2', '4.4', '5.0']),
```
eramongodb, line 33: It seems like [4.2](https://spruce.mongodb.com/task/mongo_c_driver_cse_matrix_openssl_cse_sasl_cyrus_openssl_rhel83_zseries_gcc_test_4.2_server_auth_patch_8b5a14e5a35fb7b365a1a53f24c933f8c497f337_63e5517461837d4373055658_23_02_09_20_03_02/logs?execution=0) and [4.4](https://spruce.mongodb.com/task/mongo_c_driver_cse_matrix_openssl_cse_sasl_cyrus_openssl_rhel83_zseries_gcc_test_4.4_server_auth_patch_8b5a14e5a35fb7b365a1a53f24c933f8c497f337_63e5517461837d4373055658_23_02_09_20_03_02/logs?execution=0) are not available on `rhel83-zseries`. Shall we remove them from the matrix or apply an alternative solution?

rcsanchez97: I think that we just remove them.  The server only supports 5.0 and newer on RHEL8 on zSeries.  To get 4.4/4.2 we have to go back to RHEL7, but my understanding is that those hosts are deprecated.  That said, using RHEL7 to get 4.4/4.2 test coverage seems like a reasonable thing and I don't think that it would violate any deprecation principle (being that 4.4/4.2 are legacy versions at this point).

.evergreen/config_generator/components/funcs/upload_test_results.py
```diff
@@ -10,7 +10,7 @@ class UploadTestResults(Function):
         # Ensure attach_results does not fail even if no tests results exist.
         bash_exec(
             script='''\
-                mkdir mongoc
+                mkdir -p mongoc
```
rcsanchez97, line 13: Neat.  TIL that one of the uses of the `-p` to `mkdir` is "no error if existing".

rcsanchez97 (approved):

.evergreen/config_generator/components/cse/openssl.py
eramongodb: Given 4.2 and 4.4 tests are being covered by other distros below, and I do not expect there to be significant differences in code behavior across platforms, I think I will opt to remove them rather than adding deprecated distros.

.evergreen/config_generator/components/cse/winssl.py
eramongodb: Will be addressed by restricting executed tests by CSE tasks to `/client_side_encryption/*`.

eramongodb:
Latest changes verified by [this patch](https://spruce.mongodb.com/version/63e6a4b132f4174313baa9be/tasks).

Notes:

- Changed all non-NoSSL tests to use Cyrus instead in accordance with the advice: "Testing with sasl=off provides no value". SaslOff tests are now only executed by the NoSSL matrix.
- Client Side Encryption tasks now only execut `/client_side_encryption/*` tests, reducing the runtime range from [13 to +60 minutes](https://spruce.mongodb.com/version/63e5517461837d4373055658/task-duration?duration=DESC&page=0&taskName=cse-.%2A-test) down to [2 to 19 minutes](https://spruce.mongodb.com/version/63e6a4b132f4174313baa9be/task-duration?duration=DESC&page=0&taskName=cse-.%2A-test). 🎉 
- New CSE test matrices appear to have resolved CDRIVER-4347, so the CSE tests have been re-enabled.
- A drive-by fix of the random [scan-build failure](https://parsley.mongodb.com/evergreen/mongo_c_driver_darwin_debug_compile_scan_build_patch_8b5a14e5a35fb7b365a1a53f24c933f8c497f337_63e6a4b132f4174313baa9be_23_02_10_20_10_28/0/task?bookmarks=0,1025&shareLine=372) is included (https://github.com/mongodb/mongo-c-driver/pull/1178 was not enough?).

.evergreen/config_generator/components/cse/darwinssl.py
```diff
@@ -0,0 +1,75 @@
...
+COMPILE_MATRIX = [
+    ('macos-1014', 'clang', None, ['cyrus']),
+]
+
+# TODO (CDRIVER-3789): test cse with the 'sharded' topology.
+TEST_MATRIX = [
+    ('macos-1014', 'clang', None, 'cyrus', ['auth'], ['server', 'replica' ], ['4.2', '4.4', '5.0']),
+
+    # Test 6.0+ with a replica set since Queryable Encryption does not support the 'server' topology.
+    ('macos-1014', 'clang', None, 'cyrus', ['auth'], ['server', 'replica' ], ['6.0', 'latest']),
```
eramongodb, line 25: The replica + 6.0 test is [currently failing](https://spruce.mongodb.com/task/mongo_c_driver_cse_matrix_darwinssl_cse_sasl_cyrus_darwinssl_macos_1014_clang_test_6.0_replica_auth_patch_8b5a14e5a35fb7b365a1a53f24c933f8c497f337_63e6a4b132f4174313baa9be_23_02_10_20_10_28/logs?execution=0) due to the [following error](https://parsley.mongodb.com/evergreen/mongo_c_driver_cse_matrix_darwinssl_cse_sasl_cyrus_darwinssl_macos_1014_clang_test_6.0_replica_auth_patch_8b5a14e5a35fb7b365a1a53f24c933f8c497f337_63e6a4b132f4174313baa9be_23_02_10_20_10_28/0/task?bookmarks=0,806&shareLine=765) during `/client_side_encryption/legacy/fle2-EncryptedFields-vs-jsonSchema`:

```
test error in: ./mongoc/src/libmongoc/tests/json-test-operations.c 691:check_result()
Error in "encryptedFields is preferred over jsonSchema" test : expected 1 keys, not 0
Expected:
{ "0" : { "_id" : 1, "encryptedIndexed" : "123" } }
Actual:
{ }
```

The replica + latest test is [also failing](https://spruce.mongodb.com/task/mongo_c_driver_cse_matrix_darwinssl_cse_sasl_cyrus_darwinssl_macos_1014_clang_test_latest_replica_auth_patch_8b5a14e5a35fb7b365a1a53f24c933f8c497f337_63e6a4b132f4174313baa9be_23_02_10_20_10_28/logs?execution=0) due to the [following error](https://parsley.mongodb.com/evergreen/mongo_c_driver_cse_matrix_darwinssl_cse_sasl_cyrus_darwinssl_macos_1014_clang_test_latest_replica_auth_patch_8b5a14e5a35fb7b365a1a53f24c933f8c497f337_63e6a4b132f4174313baa9be_23_02_10_20_10_28/0/task?bookmarks=0,730&shareLine=689) during `/client_side_encryption/legacy/fle2-Range-DecimalPrecision-FindOneAndUpdate`:

```
error initializing cipher: Unimplemented (-4305). CTR mode is only supported on macOS 10.15+
```

It seems like this issue is exclusive to either DarwinSSL or the macos-1014 distro, as [no other task](https://spruce.mongodb.com/version/63e6a4b132f4174313baa9be/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&taskName=cse-.%2A-test-%286%5C.0%7Clatest%29-replica-auth) with the same test parameters are failing.

We could attempt to address these issues in this PR, or defer them to followup work (new JIRA tickets).

eramongodb:
SASL was defaulting to "SSPI" in `compile-windows.sh` if not defined. Added explicit `sasl='OFF'` for all `SaslOff` compile functions. Verified by [this patch](https://spruce.mongodb.com/version/63e6bcbdc9ec44147e627dfe/tasks).

.evergreen/config_generator/components/cse/darwinssl.py
kevinAlbs: > The replica + 6.0 test is [currently failing](https://spruce.mongodb.com/task/mongo_c_driver_cse_matrix_darwinssl_cse_sasl_cyrus_darwinssl_macos_1014_clang_test_6.0_replica_auth_patch_8b5a14e5a35fb7b365a1a53f24c933f8c497f337_63e6a4b132f4174313baa9be_23_02_10_20_10_28/logs?execution=0) due to the [following error](https://parsley.mongodb.com/evergreen/mongo_c_driver_cse_matrix_darwinssl_cse_sasl_cyrus_darwinssl_macos_1014_clang_test_6.0_replica_auth_patch_8b5a14e5a35fb7b365a1a53f24c933f8c497f337_63e6a4b132f4174313baa9be_23_02_10_20_10_28/0/task?bookmarks=0,806&shareLine=765) during `/client_side_encryption/legacy/fle2-EncryptedFields-vs-jsonSchema`:
> 
> ```
> test error in: ./mongoc/src/libmongoc/tests/json-test-operations.c 691:check_result()
> Error in "encryptedFields is preferred over jsonSchema" test : expected 1 keys, not 0
> Expected:
> { "0" : { "_id" : 1, "encryptedIndexed" : "123" } }
> Actual:
> { }
> ```

Additional logging [shows the "insertOne" operation fails with this message](https://spruce.mongodb.com/task/mongo_c_driver_cse_matrix_darwinssl_cse_sasl_cyrus_darwinssl_macos_1014_clang_test_6.0_replica_auth_patch_8b5a14e5a35fb7b365a1a53f24c933f8c497f337_63ea7a2c3066153b5a6afa86_23_02_13_17_58_06/logs?execution=0):

> CTR mode is only supported on macOS 10.15+

The test runner ignores the error from the "insertOne" operation. The "insertOne" operation [does not include a result assertion](https://github.com/mongodb/specifications/blob/31bbf60c4f3008d77fc5ca5287b165e5ea97a4a6/source/client-side-encryption/tests/legacy/fle2-EncryptedFields-vs-jsonSchema.yml#L21-L23). Operations without result assertions are [not checked for errors in the test runner](https://github.com/mongodb/mongo-c-driver/blob/6a7692a8357df279dfbfd9c89ed8724ec0aacd9f/src/libmongoc/tests/json-test-operations.c#L633-L637).

I suggest running the `cse` tests with macOS 10.15. MONGOCRYPT-385 introduced the Queryable Encryption feature that is unavailable on macOS 10.14.

Continue running the non-`cse` tests with macOS 10.14, since it is listed under [Supported Platforms](https://mongoc.org/libmongoc/current/installing.html).

.evergreen/config_generator/components/sasl/openssl.py
```diff
@@ -0,0 +1,119 @@
...
+    ('ubuntu1604-arm64',  'gcc',        None, 'auto', ['noauth', 'auth'], ['server',          ], [       '4.0',                                     ]),
+    ('ubuntu1604',        'clang',      None, 'auto', ['noauth', 'auth'], ['server',          ], ['3.6', '4.0', '4.2', '4.4',                       ]),
+    ('ubuntu1804-arm64',  'gcc',        None, 'auto', ['noauth', 'auth'], ['server',          ], [              '4.2', '4.4', '5.0',        'latest']),
+    ('ubuntu1804',        'gcc',        None, 'auto', ['noauth', 'auth'], ['server',          ], [              '4.2', '4.4', '5.0',                ]),
+    ('ubuntu1804',        'gcc',        None, 'auto', ['noauth', 'auth'], ['server', 'replica'], [       '4.0',                             'latest']),
+    ('ubuntu2004',        'gcc',        None, 'auto', ['noauth', 'auth'], ['server',          ], [                                          'latest']),
+    ('windows-64-vs2017', 'vs2017x64',  None, 'auto', ['noauth', 'auth'], ['server',          ], [                                          'latest']),
+]
```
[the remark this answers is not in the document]
eramongodb: Applied comment with clarification in https://github.com/mongodb/mongo-c-driver/commit/f509b966ac43ea7f61d6648123d07ce049ce2980#diff-f7a5f6d2c55a36dbce42ddaf946571d4f9c57b817dbc6666ced47bf8029ea4daR45.

eramongodb:
Latest changes (prior to merge with upstream/master) verified by [this patch](https://spruce.mongodb.com/version/63eab7593e8e863afae925c0/tasks).

.evergreen/config_generator/components/cse/darwinssl.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+SSL = 'darwinssl'
+TAG = f'cse-matrix-{SSL}'
+
+
+# pylint: disable=line-too-long
+# fmt: off
+COMPILE_MATRIX = [
+    ('macos-1015', 'clang', None, ['cyrus']),
```
eramongodb, line 17: It seems like [this task](https://spruce.mongodb.com/task/mongo_c_driver_cse_matrix_darwinssl_cse_sasl_cyrus_darwinssl_macos_1015_clang_compile_patch_04d975b2e140e9a40859f4a44e533959bdaffa37_63eab7593e8e863afae925c0_23_02_13_22_19_07/logs?execution=0) may be blocked by macos-1015 [not yet(?) providing](https://parsley.mongodb.com/evergreen/mongo_c_driver_cse_matrix_darwinssl_cse_sasl_cyrus_darwinssl_macos_1015_clang_compile_patch_04d975b2e140e9a40859f4a44e533959bdaffa37_63eab7593e8e863afae925c0_23_02_13_22_19_07/0/task?bookmarks=0,881&shareLine=557) the CoreFoundation library (aka Secure Transport) required by both libmongocrypt and libmongoc when Client Side Encryption is enabled.

.evergreen/config_generator/components/sasl/nossl.py
```diff
@@ -0,0 +1,118 @@
...
+from config_generator.etc.sasl.test import generate_test_tasks
+
+
+SSL = 'nossl'
+TAG = f'sasl-matrix-{SSL}'
+
+
+# pylint: disable=line-too-long
+# fmt: off
+COMPILE_MATRIX = [
```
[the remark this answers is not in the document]
eramongodb: Applied, but also kept ubuntu1604 as it is required by the 3.6 test task.

.evergreen/config_generator/components/cse/darwinssl.py
kevinAlbs: I forgot I found the same problem in libmongocrypt before and filed MONGOCRYPT-443. I suggest commenting out cse tasks on macOS until MONGOCRYPT-443 is resolved.

kevinAlbs (approved):

.evergreen/config_generator/components/cse/darwinssl.py
eramongodb: Rather than commenting out the tasks, I added `disable: true` to the tasks instead and renamed `EvgTaskWithRunOn` to simply `Task` accordingly. Per [docs](https://github.com/evergreen-ci/evergreen/wiki/Project-Configuration-Files#limiting-when-a-task-will-run):

> This behaves similarly to commenting out the task but will not trigger any validation errors.

PR #1198, opened 2023-01-31
CDRIVER-4394 add decimal128 Range Index tests

kevinAlbs:
# Summary
- add decimal128 explicit encryption tests
- sync fle2-Range* specification tests
- compile libmongocrypt with compile.sh script instead of invoking cmake directly
- remove `*-cse` tasks from clang 3.4 Ubuntu 14.04 variant.

# Background & Motivation

## Use of compile.sh

The C driver currently builds libmongocrypt by invoking `cmake` directly in `compile-unix.sh` and `compile-windows.sh`. https://github.com/mongodb/libmongocrypt/pull/522#discussion_r1072337705 notes the required version of cmake to build libmongocrypt is cmake will soon be 3.12.0. The C driver builds with older versions of cmake in some distros. libmongocrypt's `ensure-cmake.sh` script ensures the presence of a CMake executable matching a specific version. Therefore, this PR proposes building libmongocrypt using the `compile.sh` script. Using `compile.sh` is [currently done in the Go driver](https://github.com/mongodb/mongo-go-driver/blob/60cfd91eb66c2646de5034327c50208d13c05db3/.evergreen/config.yml#L123).

## Removal of `*-cse` tasks from clang 3.4 Ubuntu 14.04

Removal of `*-cse` tasks from clang 3.4 Ubuntu 14.04 variant is motivated by an observed compiler error for the libmongocrypt C++ tests (described https://github.com/mongodb/libmongocrypt/pull/522#discussion_r1072436349). libmongocrypt does not build on Ubuntu 14.04. Ubuntu 14.04 does not support server versions >= 4.2. Server version 4.2 is the minimum required for CSFLE support.

PR #1204, opened 2023-02-13
Silence xtrace and debug output when compiling libmongocrypt

eramongodb:
A small followup to https://github.com/mongodb/mongo-c-driver/pull/1198 and https://github.com/mongodb/mongo-c-driver/pull/1202 that reduces the verbosity of output generated by compiling libmongocrypt. Verified by [this patch](https://spruce.mongodb.com/version/63ea5dc232f4174313be458d/tasks).

Invoking `exec 3>/dev/null` and setting `BASH_XTRACEFD="3"` [redirects xtrace output to /dev/null](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-BASH_005fXTRACEFD). This workaround can be removed once the libmongocrypt commit being built contains https://github.com/mongodb/libmongocrypt/pull/571. Note, [macos-1014 still emits xtrace output](https://parsley.mongodb.com/evergreen/mongo_c_driver_darwin_debug_compile_sasl_darwinssl_cse_patch_04d975b2e140e9a40859f4a44e533959bdaffa37_63ea5dc232f4174313be458d_23_02_13_15_56_53/0/task?bookmarks=0,2400&shareLine=73) due to running Bash version 3.2; `BASH_XTRACEFD` [requires 4.1 or newer](https://mywiki.wooledge.org/BashFAQ/061).

Setting `DEBUG="0"` silences debug output generated by [calls to debug](https://github.com/mongodb/libmongocrypt/blob/361b208d1e9b90071feb8b842481338189330e0d/.evergreen/init.sh#L55). This variable needs to be set explicitly since `DEBUG` is also being used by [compile-unix.sh](https://github.com/mongodb/mongo-c-driver/blob/04d975b2e140e9a40859f4a44e533959bdaffa37/.evergreen/scripts/compile-unix.sh#L79) to toggle the CMake build configuration.

### DIFF
--- a/.evergreen/scripts/compile-test-azurekms.sh
+++ b/.evergreen/scripts/compile-test-azurekms.sh
@@ -17,7 +17,10 @@ git clone --depth=1 https://github.com/mongodb/libmongocrypt --branch 1.7.0
     git checkout -q 4c4aa8bf # Allows -DENABLE_MONGOC=OFF.
     popd # libmongocrypt
 }
-MONGOCRYPT_INSTALL_PREFIX=${INSTALL_DIR} \
+exec 3>/dev/null
+BASH_XTRACEFD="3" \
+    DEBUG="0" \
+    MONGOCRYPT_INSTALL_PREFIX=${INSTALL_DIR} \
     DEFAULT_BUILD_ONLY=true \
     LIBMONGOCRYPT_EXTRA_CMAKE_FLAGS="-DMONGOCRYPT_MONGOC_DIR=$ROOT -DBUILD_TESTING=OFF -DENABLE_ONLINE_TESTS=OFF -DENABLE_MONGOC=OFF" \
     ./libmongocrypt/.evergreen/compile.sh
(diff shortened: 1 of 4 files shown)


.evergreen/scripts/compile-test-azurekms.sh
```diff
@@ -17,7 +17,10 @@ git clone --depth=1 https://github.com/mongodb/libmongocrypt --branch 1.7.0
     git checkout -q 4c4aa8bf # Allows -DENABLE_MONGOC=OFF.
```
kevinAlbs, line 17: Suggest updating the commit after https://github.com/mongodb/libmongocrypt/pull/571 is merged.

kevinAlbs (commented):
The quieter build is appreciated.

vector-of-bool (approved):

.evergreen/scripts/compile-test-azurekms.sh
eramongodb: Commit updated as suggested; piping of xtrace is no longer necessary and thus removed.

PR #1207, opened 2023-02-15
CDRIVER-4439 add AWS credential cache

kevinAlbs:
# Summary
- Add global cache for AWS credentials.
- Add a new target, `test-awsauth`, to test new auth in AWS environments.

Verified with [this patch build](https://spruce.mongodb.com/version/63ed00e2a4cf4704208111ef) of AWS tasks.

# Background & Motivation

DRIVERS-2333 describes the required behavior.

The `test-awsauth` target replaces `mongoc-ping` in AWS tests. `test-awsauth` includes caching tests specific to AWS authentication. `test-awsauth` requires access to API in private headers to modify the cache.

## Cache locking
The cache is locked when making requests for credentials that may be cached.
This is consistent with the AWS SDK behavior [locking around fetching](https://github.com/aws/aws-sdk-cpp/blob/91eb614069aab1b3da3bf724c95a0500aac26705/src/aws-cpp-sdk-core/source/auth/AWSCredentialsProvider.cpp#L432-L447).
I assume this is intended to prevent duplicate requests.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-cluster-aws-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-cluster-aws-private.h
@@ -29,7 +29,7 @@ _mongoc_cluster_auth_node_aws (mongoc_cluster_t *cluster,
                                bson_error_t *error);
 
 /* The following are declared in the private header for testing. It is only used
- * in test-mongoc-aws.c and mongoc-cluster.aws.c */
+ * in test-mongoc-aws.c, mongoc-cluster-aws.c, and test-awsauth.c */
 typedef struct {
    char *access_key_id;
    char *secret_access_key;
--- a/src/libmongoc/src/mongoc/mongoc-cluster-aws-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-cluster-aws-private.h
@@ -21,6 +21,7 @@
 
 #include "bson/bson.h"
 #include "mongoc/mongoc-cluster-private.h"
+#include "common-thread-private.h" // bson_mutex_t
 
 bool
 _mongoc_cluster_auth_node_aws (mongoc_cluster_t *cluster,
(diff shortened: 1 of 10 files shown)


src/libmongoc/src/mongoc/mongoc-cluster-aws-private.h
```diff
@@ -29,18 +30,90 @@ _mongoc_cluster_auth_node_aws (mongoc_cluster_t *cluster,
...
+_mongoc_aws_credentials_cache_cleanup (void);
+
+
 bool
 _mongoc_aws_credentials_obtain (mongoc_uri_t *uri,
                                 _mongoc_aws_credentials_t *creds,
                                 bson_error_t *error);
 
+void
+_mongoc_aws_credentials_copy_to (const _mongoc_aws_credentials_t *src,
```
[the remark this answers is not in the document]
kevinAlbs: I plan to keep as is. The `src, dst` order is consistent with all other C driver functions named `*_copy_to`.

src/libmongoc/src/mongoc/mongoc-cluster-aws-private.h
```diff
@@ -29,18 +30,90 @@ _mongoc_cluster_auth_node_aws (mongoc_cluster_t *cluster,
...
- * in test-mongoc-aws.c and mongoc-cluster.aws.c */
+ * in test-mongoc-aws.c, mongoc-cluster-aws.c, and test-awsauth.c */
 typedef struct {
    char *access_key_id;
    char *secret_access_key;
    char *session_token;
+   // expiration is the time in milliseconds since the Epoch when these
+   // credentials expire. If expiration is 0, the credentials do not have a
+   // known expiration.
+   uint64_t expiration_ms;
```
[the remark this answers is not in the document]
kevinAlbs: I think so. Updated to use `mcd_timer`.

src/libmongoc/src/mongoc/mongoc-cluster-aws.c
```diff
@@ -967,6 +1079,9 @@ _mongoc_cluster_auth_node_aws (mongoc_cluster_t *cluster,
 
    ret = true;
 fail:
+   if (!ret) {
+      _mongoc_aws_credentials_cache_clear ();
```
eramongodb, line 1083: Can/should clearing of the credentials cache be moved into `_mongoc_aws_credentials_obtain` with `_mongoc_aws_credentials_cache_clear_nolock`? It is odd that the cache is cleared in a separate lock from the operations that determine the "should clear cache" condition.

src/libmongoc/src/mongoc/mongoc-init.c
```diff
@@ -144,9 +146,11 @@ static BSON_ONCE_FUN (_mongoc_do_init)
 #endif
 
 #if defined(MONGOC_ENABLE_OCSP_OPENSSL)
-  _mongoc_ocsp_cache_init ();
+   _mongoc_ocsp_cache_init ();
 #endif
 
+   _mongoc_aws_credentials_cache_init ();
```
eramongodb, line 152: Can/should this be guarded by a macro checking whether auth is necessary (i.e. `MONGOC_ENABLE_SSL`)?

src/libmongoc/tests/test-awsauth.c
```diff
@@ -0,0 +1,515 @@
...
+   }
+   if (a->expiration.set != b->expiration.set) {
+      return false;
+   }
+   if (a->expiration.set) {
+      if (mcd_time_compare (a->expiration.value.expire_at,
+                            b->expiration.value.expire_at) != 0) {
+         return false;
+      }
+   }
```
eramongodb, line 117: Confirming, is the expiration date supposed to be considered part of credentials equality?

src/libmongoc/tests/test-awsauth.c
```diff
@@ -0,0 +1,515 @@
...
+test_cache (const mongoc_uri_t *uri)
+{
+   bson_error_t error;
+   _mongoc_aws_credentials_t creds;
+
+   if (!caching_expected (uri)) {
+      printf ("Caching credentials is not expected. Skipping tests expecting "
+              "credential caching.\n");
+      return;
+   }
```
eramongodb, line 219: Confirming, do we want to skip tests if expected operations are not working properly (easy to miss) instead of failing loudly? Question also applies to other skip conditions in this file.

src/libmongoc/src/mongoc/mongoc-cluster-aws-private.h
```diff
@@ -29,18 +31,92 @@ _mongoc_cluster_auth_node_aws (mongoc_cluster_t *cluster,
...
 typedef struct {
    char *access_key_id;
    char *secret_access_key;
    char *session_token;
+   // expiration_timer is the time when these credentials expire.
+   // If expiration_timer is 0, the credentials do not have a known expiration.
+   struct {
+      mcd_timer value;
+      bool set;
+   } expiration;
```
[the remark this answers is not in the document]
vector-of-bool: It looks like the comment does not match the code, which uses an optional-idiom for the timer?

src/libmongoc/src/mongoc/mongoc-cluster-aws.c
```diff
@@ -350,6 +405,17 @@ _obtain_creds_from_ecs (_mongoc_aws_credentials_t *creds, bson_error_t *error)
       ecs_session_token = bson_iter_utf8 (&iter, NULL);
    }
 
+   mcd_timer expiration_timer;
+   bool expiration_set;
```
vector-of-bool, line 409: These can be dropped and the values written directly into `creds->expiration`, IIUC.

src/libmongoc/tests/test-awsauth.c
kevinAlbs: I think "yes" for the `caching_expected` check. The [test specification](https://github.com/mongodb/specifications/blob/c06650d86f7e47ea30cb2d992942bcec6ef155f9/source/auth/tests/mongodb-aws.rst#7cached-credentials) suggests:

> To determine whether to run the cache tests, the driver can check for the absence of the AWS_ACCESS_KEY_ID environment variable and of credentials in the URI.

The `can_setenv` check appears unnecessary on the tested distro. `can_setenv` has been removed.

vector-of-bool (approved):

kevinAlbs (commented):
`*-auth` tasks with latest are failing with:

```
[2023/02/22 21:47:29.636] /data/mci/a00f187bc87fef01fb2eecba06c8c43c/mongoc/src/libmongoc/tests/test-mongoc-scram.c:295 _check_error(): [Unable to use SCRAM-SHA-256 based authentication for user without any SCRAM-SHA-256 credentials registered] does not contain [Authentication failed]
```
This is failing on the waterfall, and appears unrelated to this commit. The AWS tasks still pass, proceeding with merge.

PR #1208, opened 2023-02-16
CDRIVER-4580 Move compile-test coverage from macos-1014 to macos-1100 

eramongodb:
This PR resolves CDRIVER-4580 and CDRIVER-3483. Verified by [this patch](https://spruce.mongodb.com/version/63ee7e83e3c331348025f299/tasks).

Compile tasks were added/left-in for macos-1014 and macos-1015 in the `sasl.darwinssl.COMPILE_MATRIX` as the C Driver still documents their support. However, macos-1015 compile is disabled due to MONGOCRYPT-443 (assumed to be related due to absence of `<Security/Security.h>` required by `mongoc-stream-tls-secure-transport-private.h`). These compile tasks may be removed if preferable.

These changes only apply to components defined by the new config generator. Most tasks defined by the legacy config generator still use macos-1014 (i.e. several `link-with-*` tasks and some tasks in the OCSP and Darwin variants). These may also be updated to use macos-1100 if preferable.

`ENABLE_ZSTD=OFF` workarounds related to CDRIVER-3483 were removed, as the issues regarding inconsistent availability of ZSTD libraries was specific to macos-1014 and do not seem to be present on macos-1100. See related discussion in BUILD-16814 which motivated this PR.

`add_expansions_to_env: true` was removed from `run-mock-kms-servers` as some CFLAGS being defined for libmongoc compilation were interfering with compilation of some Python packages during package installation. The function did not actually require any expansions to behave correctly.

The scan_build matrix in https://github.com/mongodb/mongo-c-driver/pull/1205 will also need to be updated to use macos-1100 instead of macos-1014 once merged.


eramongodb:
Latest changes verified by [this patch](https://spruce.mongodb.com/version/63eea7f0e3c331437080e103/tasks).

kevinAlbs (approved):
Nice, LGTM. I suggest leaving CDRIVER-4580 in the Backlog status with the fixVersion 1.24.0. I intended that ticket include the communication to drop macOS 10.14 in the 1.24.0 release.

PR #1213, opened 2023-03-09
Disable /speculative_auth tests when SASL is disabled

eramongodb:
My understanding is that the speculative auth tests require SASL for correct behavior, but are not being disabled when SASL is disabled. This PR adds a `MONGOC_ENABLE_SASL` check in addition the existing `MONGOC_ENABLE_CRYPTO` check to avoid running SASL tests when crypto is enabled but SASL is disabled (not a normal/expected configuration, but nevertheless).

### DIFF
--- a/src/libmongoc/tests/test-mongoc-speculative-auth.c
+++ b/src/libmongoc/tests/test-mongoc-speculative-auth.c
@@ -434,7 +434,7 @@ test_mongoc_speculative_auth_request_scram_pool (void)
 void
 test_speculative_auth_install (TestSuite *suite)
 {
-#ifdef MONGOC_ENABLE_CRYPTO
+#if defined(MONGOC_ENABLE_CRYPTO) && defined(MONGOC_ENABLE_SASL)
    TestSuite_AddMockServerTest (suite,
                                 "/speculative_auth/request_none",
                                 test_mongoc_speculative_auth_request_none);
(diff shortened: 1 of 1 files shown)


kevinAlbs (commented):
I think this check is not necessary.

The C driver SASL library dependency (indicated by `MONGOC_ENABLE_SASL`) is only required to implement the `authMechanism`'s `GSSAPI` and `PLAIN`. [Authentication](http://mongoc.org/libmongoc/current/authentication.html) notes:

> Note The MongoDB C Driver must be compiled with SASL support in order to use SASL PLAIN authentication.

> Note On UNIX-like environments, Kerberos support requires compiling the driver against cyrus-sasl.
> On Windows, Kerberos support requires compiling the driver against Windows Native SSPI or cyrus-sasl. The default configuration of the driver will use Windows Native SSPI.

Speculative authentication only applies to the `authMechanism`'s `MONGODB-X509`, `SCRAM-SHA-1` and `SCRAM-SHA-256`. DRIVERS-918 notes:

> Speculative authentication is ignored for server versions less than or equal to 4.2 and is supported for X.509, SCRAM-SHA-1 and SCRAM-SHA-256.

eramongodb:
Observed issue is better addressed by https://github.com/mongodb/mongo-c-driver/pull/1214.

PR #1216, opened 2023-03-13
CDRIVER-4590 don't call ERR_load_BIO_strings for OpenSSL 3.0+

jmikola:
https://jira.mongodb.org/browse/CDRIVER-4590

OpenSSL 3.0 loads error strings automatically and deprecated this function.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-openssl.c
+++ b/src/libmongoc/src/mongoc/mongoc-openssl.c
@@ -77,6 +77,7 @@ _mongoc_openssl_init (void)
    SSL_library_init ();
    SSL_load_error_strings ();
 #if OPENSSL_VERSION_NUMBER < 0x30000000L
+   // See: https://www.openssl.org/docs/man3.0/man7/migration_guide.html#Deprecated-function-mappings
    ERR_load_BIO_strings ();
 #endif
    OpenSSL_add_all_algorithms ();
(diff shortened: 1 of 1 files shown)


src/libmongoc/src/mongoc/mongoc-openssl.c
```diff
@@ -76,7 +76,9 @@ _mongoc_openssl_init (void)
 
    SSL_library_init ();
    SSL_load_error_strings ();
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
```
eramongodb, line 79:
```suggestion
#if OPENSSL_VERSION_NUMBER < 0x30000000L
   // See "deprecated function mappings" in:
   // https://www.openssl.org/docs/man3.0/man7/migration_guide.html
```
Suggest including a reference to corresponding docs.

eramongodb (approved):
Minor feedback; otherwise LGTM.

src/libmongoc/src/mongoc/mongoc-openssl.c
jmikola: Are `//` comments kosher in libmongoc? AFAIK the project uses `/* ... */` exclusively.

eramongodb: Use of `//` had been restricted due to C89/C90 compatibility due to being a C99 feature. Given progress on CDRIVER-4136 we have begun freely using this comment style throughout the repo.

I personally prefer documentation to be as local as possible, but I would not be opposed to pushing it into the commit message instead (as rationale is still captured). I will leave that up to you.

jmikola: Added a one-line comment with a link to the docs.

PR #1218, opened 2023-03-13
[CDRIVER-4568] Limit options available for createEncryptedCollection

vector-of-bool:
Refer: CDRIVER-4568

This changeset changes the parameters available in `create-encrypted-collection` to specifying only the `masterKey` to be used for datakey creation. The prior version allowed all datakey options to be passed through, but most of them were conflicting or nonsensical in this context. Refer to the linked DRIVERS ticket.

The required changes also re-specified the test cases to test both `local` as well as `aws` KMS providers. The test cases in question have been updated to run over both configurations.


src/libmongoc/tests/test-mongoc-client-side-encryption.c
```diff
@@ -6245,12 +6244,29 @@ test_auto_datakeys (void *unused)
 }
 
 static void
-test_create_encrypted_collection_simple (void *unused)
+_do_cec_test (void (*test) (const char *kmsProvider))
+{
+   // Run the test using the "local" key:
```
eramongodb, line 6249: A bit unclear what this comment is documenting. Is it documenting that local should be tested before AWS?

src/libmongoc/tests/test-mongoc-client-side-encryption.c
```diff
@@ -6245,12 +6244,29 @@ test_auto_datakeys (void *unused)
...
+   test ("local");
+   test ("aws");
+}
+
+// Declare a createEncryptedCollection test case (See usage below)
+#define CEC_TEST(name, ...)               \
```
eramongodb, line 6255: I don't have an alternative solution to suggest, but nevertheless wanted to document that this macro pattern gives me mixed feelings. I think I would prefer not using the macro despite the greater verbosity without it, but not strongly enough to suggest that this should be replaced. 😐

eramongodb (approved):
Minor feedback; otherwise LGTM.

src/libmongoc/doc/mongoc_client_encryption_create_encrypted_collection.rst
```diff
@@ -16,7 +16,7 @@ Synopsis
         const bson_t *in_options,
         bson_t *out_options,
         const char *kms_provider,
-        const mongoc_client_encryption_datakey_opts_t *dk_opts,
+        const bson_t *opt_masterKey,
```
kevinAlbs, line 19: I suggest adding a way to extend options. When this API becomes stable, it may be helpful to have a way to add supported options. Consider adding a `mongoc_client_encryption_cec_opts_t` type to support only this option:

```rst
:man_page: mongoc_client_encryption_cec_opts_t

mongoc_client_encryption_cec_opts_t
===================================

Synopsis
--------

.. code-block:: c

  typedef struct _mongoc_client_encryption_cec_opts_t mongoc_client_encryption_cec_opts_t;

Used to set options for :symbol:`mongoc_client_encryption_create_encrypted_collection()`.

.. only:: html

  Functions
  ---------

  .. toctree::
    :titlesonly:
    :maxdepth: 1

    mongoc_client_encryption_cec_opts_new
    mongoc_client_encryption_cec_opts_destroy
    mongoc_client_encryption_cec_opts_set_masterkey

.. seealso::

  | :symbol:`mongoc_client_encryption_create_encrypted_collection()`
```

kevinAlbs: Fair point. The option structs may not add much benefit. Adding options to the structs still requires an API addition of the function call. The alternative API addition of a `_v2` / `_v3` may be preferable.

PR #1220, opened 2023-03-14
CDRIVER-2964 Add test-atlas-executor for Atlas Planned Maintenance Tests

eramongodb:
## Description

This PR introduces the `test-atlas-executor` binary to be used by the [Drivers Atlas Testing](https://github.com/mongodb-labs/drivers-atlas-testing) integration scripts for the C Driver and implements the `loop` unified test runner operation.

Addresses CDRIVER-2964 and is related to https://github.com/mongodb-labs/drivers-atlas-testing/pull/164.

Verified by [this patch](https://spruce.mongodb.com/version/640f8e347742aee833e8e489/tasks).

## Unified Test Runner

The [Atlas Planned Maintenance Test Scenario Format](https://mongodb-labs.github.io/drivers-atlas-testing/spec-test-format.html) reuses the [Unified Test Format](https://github.com/mongodb/specifications/blob/ed45dc95ca174a5832d653adec5a842184b7a82f/source/unified-test-format/unified-test-format.rst) specification but in an environment incompatible with many assumptions made by the C Driver. This necessated several refactors of the unified test runner implementation to support both normal test execution (via the existing `test-libmongoc` binary) and Atlas Testing (via the new `test-atlas-executor` binary).

### test-libmongoc-lib

First and foremost, the `main()` function for `test-libmongoc` was separated from the `test-libmongoc` library into a new `test-libmongoc-main.c`. This permits defining a separate `main()` for the new `test-atlas-executor` executable while reusing the same unified test runner implementation. Both executables now link to a common `test-libmongoc-lib` static library (whereas prior the sources `${test-libmongoc-sources}` were passed directly to compilation of the old `test-libmongoc` target). Two new functions `test_libmongoc_init` and `test_libmongoc_destroy` were introduced to maintain correct initialization behavior of `test-libmongoc-lib` globals.

### test-atlas-executor

The new `test-atlas-executor` implements a unique, barebones `TestSuite` environment designed specifically to work according to the behavior of the [Atlas Testing workload executor](https://mongodb-labs.github.io/drivers-atlas-testing/spec-workload-executor.html).

The workload executor script passes the `connection-string` via the `MONGOC_TEST_URI` environment variable (reusing existing test API) and passes the `workload-spec` as `argv[1]`. The provided `workload-spec` is converted to a BSON object as-if it were a unified spec test file via `workload_spec_to_bson`, then directly executed via `TestSuite_Run_Atlas` as the sole test to be executed by the test suite.

The only other notable feature of `test-atlas-executor` is the `SIGINT` handler. This is explained in the section below concerning the `loop` operation.

### New Unified Test Runner Entities

Two new entity types were required to support the `store.*AsEntity` parameters of the new `loop` operation: a BSON array entity to store events/failures/errors and a "counter" entity to keep track of iterations/successes.

The BSON array entity was implemented as a `mongoc_array_t` of `bson_t*` objects rather than a `bson_t` storing a BSON array. This is to simplify append and iteration operations on the array of BSON objects, which is cumbersome when done via a `bson_t` object. The BSON object containing a BSON array as required by the spec is created on-demand in `test_generate_atlas_results`. The element type is `bson_t *` rather than `bson_t` to avoid potential memory management complications caused by by-value copying `bson_t` objects into the array.

The "counter" entity was implemented as a simple `size_t`.

### usecs_since_epoch

A new `usecs_since_epoch` utility function was added, supplementing `_mongoc_get_real_time_ms`, as millisecond precision was observed to be a bit too low for the floating-point timestamps in the elements of the events/failures/errors arrays, which occasionally led to multiple events with identical timestamps.

### Entity Map

The entity map's handling of event handlers required refactoring to support the specification of _either_ `observeEvents` or `storeEventsAsEntities` or both.

New arrays `observe_events` and `store_events` were added to `entity_t` that indicate which event types were requested to be "observed" or "stored". These arrays are populated during parsing and creation of a "client" entity. When any (supported) event type is specified, the `set_command_callback` function is invoked, registering the appropriate callback function via `mongoc_apm_set_.*_cb`. When invoked, the registered callback function invokes both `observe_event` and `store_event_to_entities`, each of which inspect the `observe_events` and `store_events` arrays respectively to determine whether or not to record the given event.

These changes were also accompanied by an update of command type string comparison to be case-insensitive to support both `commandStartedEvent` and `CommandStartedEvent` syntax (was not a problem prior, but the casing is inconsistent for Atlas Testing workload specs).

### Miscellaneous Behavior Exceptions

The unified test runner library makes assumptions about the environment that do not apply when executed by the Atlas Testing workload executor. These assumptions (and corresponding workarounds) include:

- If defined, `MONGOC_TEST_URI` always starts with `mongodb://`.
- If provided, username and password is only ever specified via `MONGOC_TEST_USER` and `MONGOC_TEST_PASSWORD`.
- If there is a live server, its URI always satisfies a certain format (see `call_hello_with_host_and_port`).

These assumptions are handled (avoided) by querying the presence of a new `MONGOC_TEST_ATLAS` environment variable. Grepping for `MONGOC_TEST_ATLAS` should reveal where prior assumptions are not compatible with the Atlas Testing environment with corresponding comments documentation rationale, including the conditional invocation of `test_generate_atlas_results` prior to returning from `test_run` (before destruction of recorded events).

### serverSelectionTryOnce=false

This URI option was being set in `test_framework_get_uri_str_no_auth` (typically invoked as `test_framework_get_uri`) for convenience. When run as `test-atlas-executor`, this option is required to ensure correct behavior during maintenance scenarios (keep looking for the server instead of failing on first attempt).

## `loop` Operation

All the changes to the unified test runner above were to permit the implementation of the new `loop` operation. Due to the complexity of the `loop` operation's expected behavior, most behavior and rationale is documented with inline comments.

The only major point to note is that the `loop` operation is implemented assuming it is _only_ executed when run by `test-atlas-executor`, not `test-libmongoc`. This assumption is primarily motivated by the `operation_loop_terminated` global variable that is set by the `SIGINT` signal handler registered by `test-atlas-executor`. This signal handler is required to "gracefully" cease execution of further loop iterations as may be requested by the workload executor. Similar handling could be added to `test-libmongoc` but was omitted on grounds of unnecessary complexity (not an expected use case).

### DIFF
--- a/src/libmongoc/CMakeLists.txt
+++ b/src/libmongoc/CMakeLists.txt
@@ -952,7 +952,6 @@ target_sources (test-libmongoc-lib PRIVATE
    ${PROJECT_SOURCE_DIR}/tests/mock_server/sync-queue.c
    ${PROJECT_SOURCE_DIR}/tests/test-conveniences.c
    ${PROJECT_SOURCE_DIR}/tests/test-happy-eyeballs.c
-   ${PROJECT_SOURCE_DIR}/tests/test-libmongoc.c
    ${PROJECT_SOURCE_DIR}/tests/test-mongoc-aggregate.c
    ${PROJECT_SOURCE_DIR}/tests/test-mongoc-array.c
    ${PROJECT_SOURCE_DIR}/tests/test-mongoc-async.c
@@ -1073,7 +1072,7 @@ function (mongoc_add_test test)
 endfunction ()
 
 mongoc_add_test (test-libmongoc ${PROJECT_SOURCE_DIR}/tests/test-libmongoc-main.c)
-mongoc_add_test (test-atlas-executor ${PROJECT_SOURCE_DIR}/tests/test-atlas-executor)
+mongoc_add_test (test-atlas-executor ${PROJECT_SOURCE_DIR}/tests/test-atlas-executor.c)
 mongoc_add_test (test-mongoc-gssapi ${PROJECT_SOURCE_DIR}/tests/test-mongoc-gssapi.c)
 mongoc_add_test (test-mongoc-cache ${PROJECT_SOURCE_DIR}/tests/test-mongoc-cache.c)
 mongoc_add_test (test-azurekms ${PROJECT_SOURCE_DIR}/tests/test-azurekms.c)
--- a/src/libmongoc/CMakeLists.txt
+++ b/src/libmongoc/CMakeLists.txt
@@ -1046,6 +1046,7 @@ function (mongoc_add_test test use_shared)
 endfunction ()
 
 mongoc_add_test (test-libmongoc FALSE ${PROJECT_SOURCE_DIR}/tests/test-libmongoc-main.c)
+mongoc_add_test (test-atlas-executor FALSE ${PROJECT_SOURCE_DIR}/tests/test-atlas-executor)
 mongoc_add_test (test-mongoc-gssapi FALSE ${PROJECT_SOURCE_DIR}/tests/test-mongoc-gssapi.c)
 mongoc_add_test (test-mongoc-cache FALSE ${PROJECT_SOURCE_DIR}/tests/test-mongoc-cache.c)
 mongoc_add_test (test-azurekms FALSE ${PROJECT_SOURCE_DIR}/tests/test-azurekms.c)
--- a/src/libmongoc/CMakeLists.txt
+++ b/src/libmongoc/CMakeLists.txt
@@ -1035,6 +1035,10 @@ if (ENABLE_TESTS)
             $<$<C_COMPILER_ID:MSVC>:DbgHelp.dll>
             mongoc_static
       )
+      # We have tests that test our deprecated api.
+      target_compile_options (test-libmongoc-lib
+         PRIVATE $<$<NOT:$<C_COMPILER_ID:MSVC>>:-Wno-deprecated-declarations>
+      )
       target_include_directories (${test}
          PRIVATE
             ${PROJECT_SOURCE_DIR}/tests
--- a/src/libmongoc/CMakeLists.txt
+++ b/src/libmongoc/CMakeLists.txt
@@ -1036,7 +1036,7 @@ if (ENABLE_TESTS)
             mongoc_static
       )
       # We have tests that test our deprecated api.
-      target_compile_options (test-libmongoc-lib
+      target_compile_options (${test}
          PRIVATE $<$<NOT:$<C_COMPILER_ID:MSVC>>:-Wno-deprecated-declarations>
       )
       target_include_directories (${test}
--- a/src/libmongoc/CMakeLists.txt
+++ b/src/libmongoc/CMakeLists.txt
@@ -759,20 +759,9 @@ if (MONGOC_ENABLE_MONGODB_AWS_AUTH)
    endif()
 endif ()
 
-set (MONGOC_INTERNAL_INCLUDE_DIRS
-   $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/src>
-   $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/src/mongoc>
-   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>
-   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/mongoc>
-   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/../../src/common>
-   $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/../../src>
-   $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/../../src/common>
-)
-
 add_library (mongoc_shared SHARED ${SOURCES} ${HEADERS} ${HEADERS_FORWARDING})
 set_target_properties (mongoc_shared PROPERTIES CMAKE_CXX_VISIBILITY_PRESET hidden)
 target_link_libraries (mongoc_shared PRIVATE ${LIBRARIES} PUBLIC ${BSON_LIBRARIES})
-target_include_directories (mongoc_shared BEFORE PRIVATE ${MONGOC_INTERNAL_INCLUDE_DIRS})
 target_include_directories (mongoc_shared PRIVATE ${ZLIB_INCLUDE_DIRS})
 target_include_directories (mongoc_shared PRIVATE ${LIBMONGOCRYPT_INCLUDE_DIRECTORIES})
 if (MONGOC_ENABLE_MONGODB_AWS_AUTH)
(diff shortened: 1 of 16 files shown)


src/libmongoc/tests/unified/entity-map.c
```diff
@@ -184,28 +206,61 @@ should_ignore_event (entity_t *client_entity, event_t *event)
...
+         bson_strdup (funcs.get_database_name (apm_command));
    }
 
-   event->server_connection_id = started->server_connection_id;
+   if (funcs.get_service_id (apm_command)) {
+      bson_oid_copy (funcs.get_service_id (apm_command), &event->service_id);
+   }
+
+   BSON_ASSERT (funcs.get_server_connection_id);
+   event->server_connection_id = funcs.get_server_connection_id (apm_command);
```
kevinAlbs, line 263: I am not yet sure why. When `mongoc_apm_command_started_get_server_connection_id` returns `MONGOC_NO_SERVER_CONNECTION_ID` (-1) it is stored as 4294967295. 4294967295 is 2**32 - 1, so I suspected something is casting to a `uint32_t`.

This appears to fail [tests with servers <= 4.0](https://spruce.mongodb.com/task/mongo_c_driver_sanitizers_matrix_asan_asan_sasl_cyrus_openssl_ubuntu1804_clang_test_4.0_sharded_auth_patch_4b1c7751a1b00c3b79f044fe023075ec4a833903_6410a7323066156d84d2c918_23_03_14_16_56_20/logs?execution=0) expecting no server connection ID to be set: `expected no server connectionId, but got 4294967295`

kevinAlbs (commented):
Great work. The separated commits are much appreciated.

Can you investigate the `expected no server connectionId, but got 4294967295` test failure?

src/libmongoc/tests/unified/entity-map.c
```diff
@@ -1761,6 +2003,28 @@ entity_map_add_bson (entity_map_t *em,
...
 
+bool
+entity_map_add_bson_array (entity_map_t *em,
+                           const char *id,
+                           bson_error_t *error)
+{
+   // Note: the specification states we should be storing a BSON object of array
+   // type, but we use an array of BSON objects instead to make append and
+   // iteration easier.
+   mongoc_array_t *array = BSON_ALIGNED_ALLOC (mongoc_array_t);
```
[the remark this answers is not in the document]
eramongodb: This was a leftover from prior iterations that had attempted to use `bson_t` objects as elements instead of `bson_t*`. Alignment is not necessary for `mongoc_array_t`. Will simplify.

src/libmongoc/tests/unified/entity-map.c
eramongodb: Identified cause of failure to be undefined behavior due to incorrect `int64_t` return type for `get_server_connection_id_func_t` where assigned function pointers actually return `int32_t` instead:

```diff
- typedef int64_t (*get_server_connection_id_func_t) (const void *);
+ typedef int32_t (*get_server_connection_id_func_t) (const void *);
```

eramongodb:
Latest changes are verified by [this patch](https://spruce.mongodb.com/version/641b404dc9ec44165cde6d0d/tasks).

Compilation failures were addressed by converting `test-libmongoc-lib` into an `OBJECT` library (rather than a `STATIC` library) + ensuring proper compile and link properties were being set for both `test-libmongoc-lib` and individual test executables (some of which may not link against `test-libmongoc-lib`). The `test-libmongoc-sources` variable was re-introduced to minimize diffs against the target branch.

Additional changes include the removal of references to the obsolete `${MONGOC_INTERNAL_INCLUDE_DIRS}` variable (never set) and grouping of `ENABLE_TESTING` and `ENABLE_EXAMPLE` routines for improved readability/locality.

src/libmongoc/CMakeLists.txt
```diff
@@ -1058,75 +1009,133 @@ if (MONGOC_ENABLE_SASL_CYRUS)
...
 
-mongoc_add_test (test-libmongoc FALSE ${test-libmongoc-sources})
-mongoc_add_test (test-mongoc-gssapi FALSE ${PROJECT_SOURCE_DIR}/tests/test-mongoc-gssapi.c)
-mongoc_add_test (test-mongoc-cache FALSE ${PROJECT_SOURCE_DIR}/tests/test-mongoc-cache.c)
-mongoc_add_test (test-azurekms FALSE ${PROJECT_SOURCE_DIR}/tests/test-azurekms.c)
-mongoc_add_test (test-gcpkms FALSE ${PROJECT_SOURCE_DIR}/tests/test-gcpkms.c)
-mongoc_add_test (test-awsauth FALSE ${PROJECT_SOURCE_DIR}/tests/test-awsauth.c)
-
 if (ENABLE_TESTS)
+   add_library (test-libmongoc-lib STATIC ${test-libmongoc-sources})
```
kevinAlbs, line 1013:
```suggestion
   add_library (test-libmongoc-lib OBJECT ${test-libmongoc-sources})
```

src/libmongoc/tests/unified/util.c
```diff
@@ -185,9 +185,29 @@ is_unsupported_event_type (const char *event_type)
...
+
+   const int64_t secs = (int64_t) tv.tv_sec;
+   const int64_t usecs = (int64_t) tv.tv_usec;
+
+   const int64_t factor = 1000000;
+
+   BSON_ASSERT (INT64_MAX / factor >= secs);
+   BSON_ASSERT (INT64_MAX - (factor * secs) >= usecs);
+
+   return tv.tv_sec * factor + usecs;
```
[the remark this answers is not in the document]
kevinAlbs: Still not addressed.

kevinAlbs (approved):

src/libmongoc/tests/unified/util.c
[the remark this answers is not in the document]
eramongodb: Sorry, got lost in commit rebase routines. Will address.

src/libmongoc/tests/unified/operation.c
```diff
@@ -3202,18 +3203,279 @@ operation_assert_session_unpinned (test_t *test,
...
+   BSON_APPEND_DOUBLE (doc, "time", secs);
+
+   _mongoc_array_append_val (array, doc); // Transfer ownership.
+
+   return true;
+}
+
+// Expected to be set by a SIGINT (on Linux and OS X) or CTRL_BREAK_EVENT (on
+// Windows) handler when test suite is being executed by test-atlas-executor.
+bool operation_loop_terminated = false;
```
vector-of-bool, line 3321: This should probably be of type `sig_atomic_t`

vector-of-bool (approved):
One minor comment, but LGTM

src/libmongoc/tests/unified/operation.c
kkloberdanz: Should probably be `volatile` as well. I'll find the reference for that in the Linux Programming Interface.

kkloberdanz: IMG_1414

kkloberdanz (approved):

PR #1227, opened 2023-03-30
Don't guess a target's filepath for the install rule

vector-of-bool:
This fixes installation when the target file's destination is not at the guessed location (e.g. multiconf installs)


rcsanchez97 (approved):
LGTM

PR #1230, opened 2023-04-04
Update NEWS for 1.23.3

vector-of-bool:
It looks like the 1.22 and 1.23 NEWS changes are trampling each other. Will get that fixed.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,9 @@
+libmongoc 1.24.0 (Unreleased)
+=============================
+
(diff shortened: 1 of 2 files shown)


NEWS
```diff
@@ -1,33 +1,30 @@
-libmongoc 1.24.0 (Unreleased)
```
kevinAlbs, line 1: The 1.24.0 release notes were intended to act as a reminder for the 1.24.0 release. I suggest keeping the 1.24.0 notes.

kevinAlbs (approved):

PR #1237, opened 2023-04-14
CDRIVER-4134 Use simple HTTP server for /http tests

eramongodb:
## Description

Followup to https://github.com/mongodb/mongo-c-driver/pull/1179. Partially resolves CDRIVER-4134. Verified by [this patch](https://spruce.mongodb.com/version/6439af08850e6141fdb1a681).

Due to [continued failure](https://spruce.mongodb.com/version/mongo_c_driver_abb0bb7da8230d342dac5061623c1e4936ed46d9/tasks) of the `/http` tests, this PR replaces use of `httpbin.org` with a simple Python HTTP server we run instead.

~Note: some tasks still fail on the rhel70 and windows-64-vs2017 distros, but deliberately deferring them to a followup PR to unblock _most_ of the other tasks that currently fail that this PR addresses.~ Identified the issue being due to the absence of a `python3` binary when `simple_http_server.py` is directly invoked.

## Simple HTTP Server

Uses the Python `http` package. It is a simple HTTP server that runs on port 8000 and sends a response to GET and POST requests. The port number can be changed to something else if 8000 (a very common default) conflicts with another process.

## config-generator Dependency

As a drive-by fix, observed a missing package error for `packaging` when running the config generator with Python 3.10, so added it to the requirements list.

## EVG Config

Added the `run_simple_http_server` component to the config generator.

The `run-simple-http-server` EVG function is now invoked before all calls to the "run-tests" EVG function, excluding the DNS, CSE, and loadbalancer tasks which limit the tests executed (thus excluding the `/http` tests).

The simple HTTP server runs in the background in a manner similar to what is already being done for mock KMS servers. This is to avoid difficult "clean up background processes" logic in the `run-tests.sh` script which has been observed to lead to hanging tasks that timeout.

The `wait_for_kms_server` Bash function in `run-tests.sh` was refactored into a more general `wait_for_server` to accomodate both the simple HTTP server and the mock KMS servers.

### DIFF
--- a/.evergreen/config_generator/components/funcs/fetch_det.py
+++ b/.evergreen/config_generator/components/funcs/fetch_det.py
@@ -1,4 +1,5 @@
 from shrub.v3.evg_command import EvgCommandType
+from shrub.v3.evg_command import expansions_update
 
 from config_generator.etc.function import Function
 from config_generator.etc.utils import bash_exec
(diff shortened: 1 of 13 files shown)


eramongodb:
Resolved the task failures on `rhel70` and `windows-64-vs2017` distros by using `find_python3` via a new `PYTHON3_BINARY` EVG expansion defined by `fetch-det`. Verified by [this patch](https://spruce.mongodb.com/version/6439bea52fbabe0436e65650).

kevinAlbs (approved):
Thank you for the fix. The addition of `PYTHON3_BINARY` seems useful.

PR #1238, opened 2023-04-15
A new and improved lldb_bson.py

vector-of-bool:
This PR rewrites the lldb_bson.py Python module for use in the LLDB debugger. The prior version supported simple inspection of BSON data as its equivalent JSON encoding, but this was often cumbersome and difficult to read.

This new module allows traversing and watching document elements (with arbitrary nesting), as well as introspection on element values. In VSCode, this also enables jump-to-memory on individual elements.

The new `lldb_bson.py` is large, and shares very little with the original (it also no longer requires the Python driver to be installed, and only uses the standard library). I've done my best to annotate and comment it as well as possible, for future non-Pythonic maintainers.

## Viewing document elements in the VSCode variables tab

Viewing document elements in a heirarchy

## Viewing document content from the LLDB CLI

Viewing document content in the LLDB CLI

## Introspection of a BSON binary and a BSON datetime

Decomposition of a BSON binary and a BSON datetime

## Decoding a Decimal128 value

Inspecting a Decimal128 value

## Adding a watch expression in VSCode with CodeLLDB

A watch expression using CodeLLDB's Python evaluator


vector-of-bool:
Added review requests, but mostly curious to see if this "(only) works on my machine" and what may need to be tweaked to work in different environments/LLDBs.

vector-of-bool:
I've tweaked the summary strings to show the type info in a similar manner to what is displayed on the Mongo CLI (e.g. `NumberInt`, `NumberLong`)

@kkloberdanz you should be able to import it with: `command script import lldb_bson.py`

kkloberdanz (approved):

lldb_bson.py
```diff
@@ -1,236 +1,1449 @@
...
+        err = SBError()
+        buf = proc.ReadMemory(addr, size, err)
+        if err.fail:
+            raise LookupError(err.description)
+        self._segments[addr] = buf
+        return addr, buf
 
-    debugger.HandleCommand(
-        'command script add --help \"%s\"'
-        ' -f lldb_bson.bson_as_json_command bson' %
```
kevinAlbs, line 233: Update the [LLDB documentation](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/doc/debugging.rst#lldb), since it may no longer be correct.

kevinAlbs (approved):
Fantastic. This seems very helpful for inspecting BSON while debugging.

I did not review the implementation but checked the output by debugging the `/bson/iter/init_from_data_at_offset` test, which includes a BSON document with all types.

Here is the launch.json I used:
```json
{
    "name": "test-libmongoc",
    "type": "cppdbg",
    "request": "launch",
    "program": "${workspaceFolder}/cmake-build/src/libmongoc/test-libmongoc",
    "args": [
        "--no-fork",
        "--match",
        "/bson/iter/init_from_data_at_offset"
    ],
    "stopAtEntry": false,
    "cwd": "${workspaceFolder}",
    "setupCommands": [
        {
            "text": "command script import ${workspaceFolder}/lldb_bson.py"
        }
    ],
    "environment": [],
    "externalConsole": false,
    "MIMode": "lldb"
}
```

LGTM with documentation updated.

eramongodb (approved):

PR #1240, opened 2023-04-18
CDRIVER-4598 Support optional checksum in OP_MSG

eramongodb:
Resolves CDRIVER-4598; see ticket for details. Verified by [this patch](https://spruce.mongodb.com/version/643ea3d757e85a236fa3d8f6).

Introduced a dedicated `CHECKSUM_FIELD` X macro that unconditional excludes/ignores the checksum value in `_mongoc_rpc_gather_msg`. It still sets the checksum field in `_mongoc_rpc_scatter_msg`, even if it is ultimately never used, which allows tests to verify the field was parsed correctly.

Added regression test `/Rpc/msg/checksum` that validates correct scatter/gather behavior when the optional checksum may be present. The new test coverage exposed an additional bug in `_mongoc_rpc_reply_get_first_msg` where it [unconditionally returns `true`](https://github.com/mongodb/mongo-c-driver/blob/7ae8fd3c33a9a7a21c1057091378bf3bd5f7d96e/src/libmongoc/src/mongoc/mongoc-rpc.c#L1028) despite potential BSON document initialization failure. For consistency with `_mongoc_rpc_reply_get_first`, the function was updated to ensure `_mongoc_rpc_get_first_document` correctly returns `false` on `OP_MSG` document parse failure.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-rpc-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-rpc-private.h
@@ -74,6 +74,7 @@ typedef struct _mongoc_rpc_section_t {
    const uint8_t *_name;        \
    int32_t _name##_len;
 #define BSON_OPTIONAL(_check, _code) _code
+#define CHECKSUM_FIELD(_name) uint32_t _name;
 
 
 #pragma pack(1)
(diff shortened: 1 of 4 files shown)


kevinAlbs (approved):
Nicely done. LGTM

PR #1242, opened 2023-04-20
CDRIVER-4121 Ensure OP_MSG is used for authentication

eramongodb:
## Description

Resolves CDRIVER-4121. Verified by [this patch](https://spruce.mongodb.com/version/6441a6e93e8e8676d012a6c3).

## OP_MSG for Authentication

Noting that `WIRE_VERSION_MIN` is currently set to 6 for the C Driver (meaning _all_ currently supported MongoDB server versions support OP_MSG, thus we should _never_ be using OP_QUERY for authentication), I believe there are _four_ possible code paths by which the OP_MSG protocol may used for authentication handshake requests (and where OP_QUERY should _not_ be used):

- `_mongoc_cluster_auth_node` via `_cluster_fetch_stream_single`
- `_mongoc_cluster_auth_node` via `_cluster_fetch_stream_pooled`
- `_mongoc_cluster_finish_speculative_auth` via `_cluster_fetch_stream_single`
- `_mongoc_cluster_finish_speculative_auth` via `_cluster_fetch_stream_pooled`

Which code path is selected thus depends on two factors:

1) is speculative authentication enabled or disabled?
2) is the client being used single or from a pool?

In all four cases, it seems that the C Driver is correctly selecting the OP_MSG protocol. This selection occurs in [`mongoc_cluster_run_command_private`](https://github.com/mongodb/mongo-c-driver/blob/46a010f28d3052b3721e316b0bdecce47de519f1/src/libmongoc/src/mongoc/mongoc-cluster.c#L687), which is invoked by all four code paths listed above via `_mongoc_cluster_run_scram_command`, which in turn is invoked either via `_mongoc_cluster_auth_scram_start` or `_mongoc_cluster_auth_scram_continue` depending on the stage and method of authentication.

## /counters/auth_with_op_msg

CDRIVER-4121 discusses the viability of implementing a test to validate this behavior. This PR implements such a test by relying on the "shared memory (performance) counters" (aka SHM counters) provided by [mongoc-counters-private.h](https://github.com/mongodb/mongo-c-driver/blob/46a010f28d3052b3721e316b0bdecce47de519f1/src/libmongoc/src/mongoc/mongoc-counters-private.h) to determine the exact number of OP_QUERY and OP_MSG requests sent by the C Driver when the `ENABLE_SHM_COUNTERS` CMake configuration variable is set to `AUTO` ([default](https://github.com/mongodb/mongo-c-driver/blob/46a010f28d3052b3721e316b0bdecce47de519f1/CMakeLists.txt#L27)) or `ON`.

This test assumes the number of OP_QUERY and OP_MSG requests sent during an authentication handshake (including the initial handshake connection) deterministic and consistent for each of the four possible code paths described above for a given authentication mechanism. SCRAM-SHA-1 was chosen due to availability on all server versions currently supported and the mechanism requiring multiple authentication steps to guarantee at least one OP_MSG request is required regardless of pooling or speculative authentication. The expected request count for each set of conditions is thoroughly documented in the test definition.

Note, this test is only run if the server is a replica set and authentication is enabled; otherwise it is skipped. It also assumes _exactly_ three members in the replica set (corresponding assertion included).

I considered augmenting tests in test-mongoc-scram.c and test-mongoc-speculative-auth.c with SHM counter assertions, but decided against it due to the complexity of deriving expected request counts based on varying authentication mechanisms and success/failure behavior. Decided to add a specialized test to test-mongoc-counters.c instead given its use of SHM counters for consistency (all SHM counter tests are defined in test-mongoc-counters.c).

## mongoc_counter_##ident##_count()

As a drive-by improvement (motivated by prior attempts to define tests in test-mongoc-scram.c and test-mongoc-speculative-auth.c as mentioned above), this series of helper functions were moved from test-mongoc-counters.c into mongoc-counters-private.h due to its general applicability. Note, it is currently still only being used by test-mongoc-counters.c.

### DIFF
--- a/src/libmongoc/tests/test-mongoc-counters.c
+++ b/src/libmongoc/tests/test-mongoc-counters.c
@@ -25,42 +25,32 @@
 /* test statistics counters excluding OP_INSERT, OP_UPDATE, and OP_DELETE since
  * those were superseded by write commands in 2.6. */
 #ifdef MONGOC_ENABLE_SHM_COUNTERS
-/* define a count function for each counter. */
-#define COUNTER(id, category, name, description)                               \
-   int32_t prev_##id;                                                          \
-   int32_t count_##id (void)                                                   \
-   {                                                                           \
-      int32_t _sum = 0;                                                        \
-      uint32_t _i;                                                             \
-      for (_i = 0; _i < _mongoc_get_cpu_count (); _i++) {                      \
-         const int64_t *counter =                                              \
-            &BSON_CONCAT (__mongoc_counter_, id)                               \
-                .cpus[_i]                                                      \
-                .slots[BSON_CONCAT (COUNTER_, id) % SLOTS_PER_CACHELINE];      \
-         _sum += bson_atomic_int64_fetch (counter, bson_memory_order_seq_cst); \
-      }                                                                        \
-      return _sum;                                                             \
-   }
+
+/* define prev_* counters for testing convenience. */
+#define COUNTER(ident, Category, Name, Description) static int32_t prev_##ident;
 #include "mongoc/mongoc-counters.defs"
 #undef COUNTER
 
-#define RESET(id)               \
-   bson_atomic_int32_exchange ( \
-      &prev_##id, count_##id (), bson_memory_order_seq_cst)
+/* helper to reset a prev_* counter */
+#define RESET(ident)                                              \
+   bson_atomic_int32_exchange (&prev_##ident,                     \
+                               mongoc_counter_##ident##_count (), \
+                               bson_memory_order_seq_cst)
 
-#define DIFF_AND_RESET(id, cmp, expected)     \
-   do {                                       \
-      int32_t old_count = prev_##id;          \
-      int32_t new_count = count_##id ();      \
-      int32_t _diff = new_count - old_count;  \
-      ASSERT_CMPINT32 (_diff, cmp, expected); \
-      RESET (id);                             \
+/* helper to compare and reset a prev_* counter. */
+#define DIFF_AND_RESET(ident, cmp, expected)                 \
+   do {                                                      \
+      int32_t old_count = prev_##ident;                      \
+      int32_t new_count = mongoc_counter_##ident##_count (); \
+      int32_t _diff = new_count - old_count;                 \
+      ASSERT_CMPINT32 (_diff, cmp, expected);                \
+      RESET (ident);                                         \
    } while (0)
 
 static void
 reset_all_counters (void)
 {
-#define COUNTER(id, category, name, description) RESET (id);
+#define COUNTER(ident, Category, Name, Description) RESET (ident);
 #include "mongoc/mongoc-counters.defs"
 #undef COUNTER
 }
--- a/src/libmongoc/tests/test-mongoc-counters.c
+++ b/src/libmongoc/tests/test-mongoc-counters.c
@@ -471,6 +471,189 @@ test_counters_streams_timeout (void)
    mongoc_client_destroy (client);
    mock_server_destroy (server);
 }
+
+static void
+test_counters_auth_with_op_msg (bool pooled)
+{
+   // This test is sensitive to the number of members in the replica set. Assert
+   // expectations to guard against the possibility of expanding the test suite
+   // to run against replica sets with a varying number of members.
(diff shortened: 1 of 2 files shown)


src/libmongoc/tests/test-mongoc-counters.c
```diff
@@ -481,6 +471,189 @@ test_counters_streams_timeout (void)
...
+            sent_queries);
+      }
+   } else {
+      // OP_QUERY requests consists of:
+      //  - initial connection handshake (x3)
+      ASSERT_WITH_MSG (
+         sent_queries == 3,
+         "expected exactly three OP_QUERY requests, but observed %" PRId32
+         " requests",
+         sent_queries);
```
kevinAlbs, line 594: This assertion fails on my machine with `expected exactly three OP_QUERY requests, but observed 6 requests`.

This led to bug report in the C driver: CDRIVER-4622
OP_QUERY egress appears to be overcounted when the hostname resolves to multiple DNS records. On my machine, "localhost" resolves to both IPv4 `127.0.0.1` and IPv6 `::1`.

I suggest adding a function to count the DNS records of "localhost" (modified from `test_framework_skip_if_no_dual_ip_hostname`):
```c
// get_dns_record_count returns the number of DNS records for "localhost".
// "localhost" may resolve to both the IPv4 `127.0.0.1` and IPv6 `::1`.
int
get_dns_record_count (void)
{
   struct addrinfo hints = {0}, *res = NULL, *iter;
   int res_count = 0;
   char *host = "localhost";

   hints.ai_family = AF_UNSPEC;
   hints.ai_socktype = SOCK_STREAM;
   hints.ai_flags = 0;
   hints.ai_protocol = 0;

   BSON_ASSERT (getaddrinfo (host, "27017", &hints, &res) != -1);

   iter = res;

   while (iter) {
      res_count++;
      iter = iter->ai_next;
   }

   freeaddrinfo (res);

   return res_count;
}
```
And conditionally skipping this test:
```c
   if (get_dns_record_count () > 1) {
      MONGOC_DEBUG (
         "Skipping test due to overcounting bug described in CDRIVER-4622");
      return;
   }
```

src/libmongoc/tests/test-mongoc-counters.c
```diff
@@ -481,6 +471,189 @@ test_counters_streams_timeout (void)
...
+         //  - polling hello by RTT monitor thread (x3)
+         //  - initial connection handshake by new cluster node
+         ASSERT_WITH_MSG (
+            sent_queries == 10,
+            "expected exactly ten OP_QUERY requests, but observed %" PRId32
+            " requests",
+            sent_queries);
+      } else {
+         // OP_QUERY requests consists of:
+         //  - initial connection handshake by server monitor (x3)
```
kevinAlbs, line 579: The server monitor threads are started on the first call to `mongoc_client_pool_pop`. The OP_QUERY messages sent by the server monitoring threads may not be guaranteed to have completed.
I suggest adding an `ASSERT_SOON` to assert this condition happens within some time interval. This could be similar to [OBSERVE_SOON](https://github.com/mongodb/mongo-c-driver/blob/46a010f28d3052b3721e316b0bdecce47de519f1/src/libmongoc/tests/test-mongoc-background-monitoring.c#L286-L301).

src/libmongoc/tests/test-mongoc-counters.c
```diff
@@ -471,6 +471,189 @@ test_counters_streams_timeout (void)
...
+      mongoc_client_pool_set_error_api (pool, MONGOC_ERROR_API_VERSION_2);
+      client = mongoc_client_pool_pop (pool);
+   } else {
+      // Note: no server API version ensures OP_QUERY for initial handshake.
+      client = test_framework_client_new_from_uri (uri, NULL);
+      test_framework_set_ssl_opts (client);
+      mongoc_client_set_error_api (client, MONGOC_ERROR_API_VERSION_2);
+   }
+
+   mongoc_counter_auth_success_reset ();
```
kevinAlbs, line 533: The server monitor threads are started on the first call to `mongoc_client_pool_pop`.
Suggest moving the calls to `reset` above `mongoc_client_pool_pop` to prevent a possible race between the monitoring thread sending an OP_QUERY and reseting the counter.

eramongodb:
Closing this PR due to work on CDRIVER-4193 and CDRIVER-4622 motivating an alternative test suite implementation.

PR #1243, opened 2023-04-21
Fix build for FreeBSD

kkloberdanz:
CDRIVER-4465

- Some standard library features need __BSD_VISIBLE set to 1 to be enabled on FreeBSD
- Cmake does not properly detect res_search on FreeBSD. Since res_search has existed since 4.3 BSD (which is the predecessor of FreeBSD), it is safe to assume that this function will exist in libc on any version of FreeBSD.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -268,8 +268,14 @@ endif ()
 # Both _BSD_SOURCE and _DEFAULT_SOURCE are defined for backwards-compatibility with glibc 2.19 and earlier.
 # _BSD_SOURCE and _DEFAULT_SOURCE are required by `getpagesize`, `h_errno`, etc.
 # _XOPEN_SOURCE=700 is required by `strnlen`, etc.
-add_definitions (-D_XOPEN_SOURCE=700 -D_BSD_SOURCE -D_DEFAULT_SOURCE -D__BSD_VISIBLE=1)
-list (APPEND CMAKE_REQUIRED_DEFINITIONS -D_XOPEN_SOURCE=700 -D_BSD_SOURCE -D_DEFAULT_SOURCE -D__BSD_VISIBLE=1)
+add_definitions (-D_XOPEN_SOURCE=700 -D_BSD_SOURCE -D_DEFAULT_SOURCE)
+list (APPEND CMAKE_REQUIRED_DEFINITIONS -D_XOPEN_SOURCE=700 -D_BSD_SOURCE -D_DEFAULT_SOURCE)
+
+# Enable non-standard features on FreeBSD with __BSD_VISIBLE=1
+if(CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
+   add_definitions (-D__BSD_VISIBLE=1)
+   list (APPEND CMAKE_REQUIRED_DEFINITIONS -D__BSD_VISIBLE=1)
+endif ()
 
 # https://opensource.apple.com/source/Libc/Libc-1439.40.11/gen/compat.5.auto.html
 # Non-POSIX extensions are required by `_SC_NPROCESSORS_ONLN`.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -268,8 +268,8 @@ endif ()
 # Both _BSD_SOURCE and _DEFAULT_SOURCE are defined for backwards-compatibility with glibc 2.19 and earlier.
 # _BSD_SOURCE and _DEFAULT_SOURCE are required by `getpagesize`, `h_errno`, etc.
 # _XOPEN_SOURCE=700 is required by `strnlen`, etc.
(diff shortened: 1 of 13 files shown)


kkloberdanz:
You can use the following code to test that connecting via SRV works. One might want to connect to an instance on Atlas since the server does not support FreeBSD at this time.

```C
#include <mongoc/mongoc.h>

int main (int argc, char **argv)
{
  bson_error_t error = {0};
  mongoc_server_api_t *api = NULL;
  mongoc_database_t *database = NULL;
  bson_t *command = NULL, reply;

  // Initialize the MongoDB C Driver.
  mongoc_init ();

  client = mongoc_client_new("<INSERT_YOUR_CONNECTION_STRING_HERE>");

  // Set the version of the Stable API on the client.
  api = mongoc_server_api_new (MONGOC_SERVER_API_V1);
  if (!mongoc_client_set_server_api (client, api, &error))
  {
    // Error condition.
    printf("Error: %s\n", error.message);
    return 0;
  }

  // Get a handle on the "admin" database.
  database = mongoc_client_get_database (client, "admin");

  // Ping the database.
  command = BCON_NEW("ping", BCON_INT32(1));
  if (mongoc_database_command_simple(database, command, NULL, &reply, &error))
  {
    printf("Pinged your deployment. You successfully connected to MongoDB!\n");
  }
  else
  {
    // Error condition.
    printf("Error: %s\n", error.message);
    return 0;
  }


  // Perform Cleanup.
  bson_destroy (&reply);
  bson_destroy (command);
  mongoc_database_destroy (database);
  mongoc_cleanup ();

  return 0;
}
```

src/libbson/examples/bson-streaming-reader.c
```diff
@@ -14,6 +14,12 @@
  * limitations under the License.
  */
 
+#ifdef __FreeBSD__
+#ifndef __BSD_VISIBLE
+#define __BSD_VISIBLE 1
```
kevinAlbs, line 19: Suggest defining `__BSD_VISIBLE` in cmake following the [pattern here](https://github.com/mongodb/mongo-c-driver/blob/95d7a187e3b410793916d854806a81d6bff9a8a4/CMakeLists.txt#L265-L279).

That may remove the need for defining `__BSD_VISIBLE` in individual files. Since FreeBSD is not tested in Evergreen, I expect future code may also need `__BSD_VISIBLE` defined.

src/libbson/tests/test-json.c
```diff
@@ -2114,7 +2120,7 @@ test_bson_json_double (void)
    ASSERT_CMPDOUBLE (bson_iter_double (&iter), ==, 0.0);
 
 /* check that "x" is -0.0. signbit not available on Solaris or VS 2010 */
```
kevinAlbs, line 2122:
```suggestion
/* check that "x" is -0.0. signbit not available on Solaris, FreeBSD, or VS 2010 */
```

CMakeLists.txt
```diff
@@ -268,8 +268,8 @@ endif ()
 # Both _BSD_SOURCE and _DEFAULT_SOURCE are defined for backwards-compatibility with glibc 2.19 and earlier.
 # _BSD_SOURCE and _DEFAULT_SOURCE are required by `getpagesize`, `h_errno`, etc.
 # _XOPEN_SOURCE=700 is required by `strnlen`, etc.
-add_definitions (-D_XOPEN_SOURCE=700 -D_BSD_SOURCE -D_DEFAULT_SOURCE)
-list (APPEND CMAKE_REQUIRED_DEFINITIONS -D_XOPEN_SOURCE=700 -D_BSD_SOURCE -D_DEFAULT_SOURCE)
+add_definitions (-D_XOPEN_SOURCE=700 -D_BSD_SOURCE -D_DEFAULT_SOURCE -D__BSD_VISIBLE=1)
+list (APPEND CMAKE_REQUIRED_DEFINITIONS -D_XOPEN_SOURCE=700 -D_BSD_SOURCE -D_DEFAULT_SOURCE -D__BSD_VISIBLE=1)
```
kevinAlbs, line 272:
```suggestion
add_definitions (-D_XOPEN_SOURCE=700 -D_BSD_SOURCE -D_DEFAULT_SOURCE)
list (APPEND CMAKE_REQUIRED_DEFINITIONS -D_XOPEN_SOURCE=700 -D_BSD_SOURCE -D_DEFAULT_SOURCE)

# Enable non-standard features on FreeBSD with __BSD_VISIBLE=1
if(CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
   add_definitions (-D__BSD_VISIBLE=1)
   list (APPEND CMAKE_REQUIRED_DEFINITIONS -D__BSD_VISIBLE=1)
endif ()
```

Suggest limiting the `__BSD_VISIBLE=1` to FreeBSD.

kevinAlbs (approved):
LGTM with one comment addressed.

vector-of-bool (approved):
LGTM

PR #1246, opened 2023-04-25
CDRIVER-4557 add int64 `connection_id` functions

kevinAlbs:
# Summary

- Add `connection_id` functions returning int64:
  - `mongoc_apm_command_started_get_server_connection_id_int64`
  - `mongoc_apm_command_succeeded_get_server_connection_id_int64`
  - `mongoc_apm_command_failed_get_server_connection_id_int64`
- Deprecate `connection_id` functions returning int32

# Background & Motivation

CDRIVER-4593 and CDRIVER-4557 applied fixes to parsing `connectionId` to handle `int32`, `int64`, and `double`. No further fix was needed.

This PR intends to satisfy the requirement of DRIVERS-2503:

> If the client and server connectionId fields are part of the driver's public API, you may have to add new int64 connectionId fields and deprecate the existing int32 fields.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-apm.h
+++ b/src/libmongoc/src/mongoc/mongoc-apm.h
@@ -105,6 +105,9 @@ mongoc_apm_command_started_get_service_id (
 MONGOC_EXPORT (int32_t)
 mongoc_apm_command_started_get_server_connection_id (
    const mongoc_apm_command_started_t *event);
+MONGOC_EXPORT (int64_t)
+mongoc_apm_command_started_get_server_connection_id_int64 (
+   const mongoc_apm_command_started_t *event);
 MONGOC_EXPORT (void *)
 mongoc_apm_command_started_get_context (
    const mongoc_apm_command_started_t *event);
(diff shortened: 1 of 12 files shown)


src/libmongoc/src/mongoc/mongoc-apm.h
```diff
@@ -104,6 +104,11 @@ mongoc_apm_command_started_get_service_id (
    const mongoc_apm_command_started_t *event);
 MONGOC_EXPORT (int32_t)
 mongoc_apm_command_started_get_server_connection_id (
+   const mongoc_apm_command_started_t *event)
+   BSON_GNUC_DEPRECATED_FOR (
```
eramongodb, line 108: Can we add a `MONGOC_WARNING` to the deprecated `mongoc_apm_command_(started|succeeded|failed)_get_server_connection_id` functions just prior to returning `MONGOC_NO_SERVER_CONNECTION_ID;` as a runtime indicator for when the deprecated function is returning `-1` due to out-of-range rather than an actual "no serverConnectionId" error?

eramongodb (approved):

src/libmongoc/src/mongoc/mongoc-apm.h
kevinAlbs: That seems helpful. Done.

PR #1248, opened 2023-04-27
CDRIVER-4629 Remove code requiring maxWireVersion less than WIRE_VERSION_OP_MSG

eramongodb:
## Description

Resolves CDRIVER-4629. Verified by [this patch](https://spruce.mongodb.com/version/644a90ae7742aef2bc3a2ce6).

Removes code that is conditioned on the reported `maxWireVersion` being less than `WIRE_VERSION_OP_MSG (6)`, which is also the current value of `WIRE_VERSION_MIN (6)`, meaning we do not support conducting operations on such servers anymore since CDRIVER-4098 (https://github.com/mongodb/mongo-c-driver/pull/926, released in 1.21.0). In short, this PR removes obsolete code that is no longer being executed.

## Obsolete WIRE_VERSION_* Macros

This PR's primary set of changes involve removal of code referencing `WIRE_VERSION_*` macros with values less than or equal to `WIRE_VERSION_MIN`.

For (almost) all such code, `wire_version > WIRE_VERSION_OBSOLETE` is assumed to be always true. Similarly, `wire_version <= WIRE_VERSION_OBSOLETE` is assumed to be always false. Using these assumptions, code transformations are applied to remove subsequently redundant or obsolete code. Changes include:

- removal of `if`/`else` branches.
- removal of `wire_version` parameters no longer needed by helper functions.
- removal of the `mongoc-write-command-legacy` component.

The only notable exception to the patterns above is in `mongoc_cluster_run_command_private`:

```c
if (mongoc_cluster_uses_server_api (cluster) ||
      server_stream->sd->max_wire_version >= WIRE_VERSION_OP_MSG) {
   retval = mongoc_cluster_run_opmsg (cluster, cmd, reply, error);
} else {
   retval =
      mongoc_cluster_run_command_opquery (cluster, cmd, -1, reply, error);
}
```

`server_stream->sd->max_wire_version >= WIRE_VERSION_OP_MSG` _cannot_ be assumed to be always true, as `mongoc_cluster_run_command_private` is sometimes used to run the initial connection handshake with a stream whose `maxWireVersion` has not yet been set (defaults to 0). The comparison was updated to use `WIRE_VERSION_MIN` instead to account for this behavior. All other comparisons with `server_stream->sd->max_wire_version` used a post-initial-connection-handshake stream that has an updated `maxWireVersion` (verified by [this patch](https://spruce.mongodb.com/version/644a94c2d1fe07ebf9750420) which temporarily inserts assertions in place of all the modified "always true/false" code paths).

## Test Suite

Some tests had to be updated to account for the changes described above.

### /max_staleness

CDRIVER-4152 and CDRIVER-4204 synced new and modified test files into src/libmongoc/tests/json/max_staleness, but did not remove deleted test files. Some of these files were triggering test failures after updates due to assumptions that [no longer apply](https://github.com/mongodb/specifications/commit/e2914e7c88ad8578b78f26706bfa8e0737d0fcc0).

### /retryable_writes/failover

The mock server "not primary" reply was missing the `RetryableWriteError` error label required by retryWrites since server 4.4.

### /inheritance/aggregate

A workaround concerning readPreference with secondaries to account for CDRIVER-4224 is no longer applicable on server 5.0+ thanks to CDRIVER-3893. Bumped the mock server version to `WIRE_VERSION_MAX` for consistency and removed the workaround.

### test_framework_skip_if_.*_6

References to `WIRE_VERSION_CHECKS (6)` functions were removed from the test suite as the corresponding conditions are now always true (3.6+) or always false. Use of `test_framework_skip_if_not_rs_version_6` were replaced with `test_framework_skip_if_not_replset` to preserve the replica set condition.


vector-of-bool (approved):
So much code deleted ❤️.

Looks _great_ to me!

src/libmongoc/tests/test-mongoc-retryable-writes.c
```diff
@@ -813,14 +818,12 @@ retryable_writes_original_error_general_command (void *ctx)
...
-      JSON_DIR,
-      "retryable_writes/legacy",
-      test_retryable_writes_cb,
-      test_framework_skip_if_max_wire_version_less_than_6,
-      test_framework_skip_if_no_crypto,
-      test_framework_skip_if_slow);
+   install_json_test_suite_with_check (suite,
+                                       JSON_DIR,
+                                       "retryable_writes/legacy",
+                                       test_retryable_writes_cb,
```
kevinAlbs, line 824:
```suggestion
                                       test_retryable_writes_cb,
                                       TestSuite_CheckLive,
```

To skip test when environment variable `MONGOC_TEST_SKIP_LIVE=ON` is set.

kevinAlbs (approved):
Fantastic, LGTM

PR #1249, opened 2023-04-27
CDRIVER-4556 Remove use of `collStats`

kevinAlbs:
# Summary

- Remove use of `collStats`

# Background & Motivation

From DRIVERS-2232:
> For drivers that have helpers for collStats or currentOp, the helpers should be adjusted to use the $collStats / $currentOp aggregation pipeline stages.  Alternatively, drivers can deprecate these helpers for future removal.  This is left to the discretion of drivers.

`mongoc_collection_stats` is already deprecated. Examples using `mongoc_collection_stats` were updated to use aggregation with the `$collStats` pipeline stage. There is no helper for `currentOp`.

---
Example code was moved from .rst files into the `examples` directory and included with the `literalinclude` directive. This is intended to have the examples compile in Evergreen.

---
The `runCommand Example 2` example, which used `collStats`, was removed. This example was added for the documentation team in DRIVERS-448. [DOCSP-21320](https://jira.mongodb.org/browse/DOCSP-21320) requests removing `runCommand Example 2` from documentation.

### DIFF
--- a/src/libmongoc/doc/mongoc_collection_command_simple.rst
+++ b/src/libmongoc/doc/mongoc_collection_command_simple.rst
@@ -50,33 +50,7 @@ Example
 
 The following is an example of executing the collection stats command.
 
-.. code-block:: c
-
-  #include <bson/bson.h>
-  #include <mongoc/mongoc.h>
-  #include <stdio.h>
-
-  static void
-  print_collection_stats (mongoc_collection_t *collection)
-  {
-     bson_error_t error;
-     const char *name;
-     bson_t *cmd;
-     bson_t reply;
-
-     name = mongoc_collection_get_name (collection);
-     cmd = BCON_NEW ("collStats", BCON_UTF8 (name));
-
-     if (mongoc_collection_command_simple (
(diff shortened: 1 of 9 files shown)


src/libmongoc/doc/mongoc_collection_command_simple.rst
```diff
@@ -48,35 +48,9 @@ This function does not check the server response for a write concern error or wr
...
-        fprintf (stderr, "%s\n", error.message);
-     }
-
-     bson_destroy (&reply);
-     bson_destroy (cmd);
-  }
+The following is an example of executing the ``ping`` command.
 
+.. literalinclude:: ../examples/example-collection-command.c
+   :caption: Example use of mongoc_collection_command
```
eramongodb, line 54:
```suggestion
```
Caption is redundant given the "The following is an example..." note above. Suggest picking one or the other.

eramongodb (approved):

vector-of-bool (approved):
LGTM

I was unaware of the `start-after` and `end-before` options. Very useful!

PR #1254, opened 2023-05-02
CDRIVER-4633 Fix compiler warning

kkloberdanz:
GCC 13 warns that `pool` may be used uninitialized. This is a false positive because pool will only be used in the case that it is initialized, but by initializing it to NULL we can avoid this warning.


kevinAlbs (approved):
LGTM.

After merging, cherry-pick the commit to the r1.23 branch so this fix can be included in the next 1.23.x patch release.

PR #1255, opened 2023-05-02
CDRIVER-4483 add docs page for "In-Use Encryption"

kevinAlbs:
# Summary

- Add page for "In-Use Encryption" linking to sub pages "Queryable Encryption" and "Client-Side Field Level Encryption".
- Update links that apply to both QE and CSFLE to point to IUE.

# Background & Motivation

DRIVERS-2454 requests:
> Both features CSFLE and QE are nested under "In Use Encryption" rather than "Client Side Encryption".  Please verify and rename any documents to follow this convention.

Links to specifications were replaced with links to the MongoDB Manual where appropriate. The MongoDB Manual is intended for users. Specifications are intended for driver developers.

To reduce scope of this PR, Evergreen tasks (e.g. `*-cse-*` tasks) and private code symbols (e.g. `_mongoc_cse_auto_encrypt`) were not renamed.


src/libmongoc/doc/mongoc_auto_encryption_opts_destroy.rst
```diff
@@ -22,5 +22,5 @@ Parameters
 
   | :symbol:`mongoc_auto_encryption_opts_new()`
 
-  | The guide for :doc:`Using Client-Side Field Level Encryption <using_client_side_encryption>`
+  | :doc:`In-Use Encryption <in-use-encryption>`
```
vector-of-bool, line 25: The explicit link title is the same as the target page title, so it is redundant on each occurrence, and can be ":doc:`in-use-encryption`".

vector-of-bool (approved):
LGTM, but maybe one tweak for reducing redundancy.

kkloberdanz (approved):
LGTM

src/libmongoc/doc/mongoc_auto_encryption_opts_destroy.rst
kevinAlbs: Nice. I was not aware of that. Updated.

PR #1259, opened 2023-05-04
CDRIVER-4498 return error if `masterKey` is set, but `provider` is not set

kevinAlbs:
# Summary
- return error if `masterKey` is set, but `provider` is not set for `rewrapManyDataKey`
- add prose test specified in DRIVERS-2441

# Background & Motivation

The specified API for `rewrapManyDataKey` does not permit calling with a set `masterKey` without a `provider`.

```typescript
    class ClientEncryption {
        rewrapManyDataKey(filter: Document, opts: RewrapManyDataKeyOpts | null): RewrapManyDataKeyResult;
    }
    class RewrapManyDataKeyOpts {
        provider: String
        masterKey: Optional<Document>
    }
```

Some driver implementations do not represent `RewrapManyDataKeyOpts` as a separate type. Instead, the `provider` and `masterKey` are both optional arguments to `RewrapManyDataKey`.

The C driver API permits setting `masterKey` without `provider`:
```c
MONGOC_EXPORT (bool)
mongoc_client_encryption_rewrap_many_datakey (
   mongoc_client_encryption_t *client_encryption,
   const bson_t *filter,
   const char *provider,
   const bson_t *master_key,
   mongoc_client_encryption_rewrap_many_datakey_result_t *result,
   bson_error_t *error);
```

The current behavior of `mongoc_client_encryption_rewrap_many_datakey` silently ignores the `master_key` option if `provider` is NULL.

This may result in unexpected behavior. A user may be attempting to rewrap keys with a new `master_key` and mistakenly passed a NULL `provider`. A NULL `provider` results in rewrapping with the same `master_key`.

The Java driver had similar behavior, which was fixed in [JAVA-4717](https://jira.mongodb.org/browse/JAVA-4717).

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-crypt.c
+++ b/src/libmongoc/src/mongoc/mongoc-crypt.c
@@ -2001,6 +2001,12 @@ _mongoc_crypt_rewrap_many_datakey (_mongoc_crypt_t *crypt,
    mongocrypt_binary_t *filter_bin = NULL;
    bool ret = false;
 
+   if (master_key) {
+      // Caller must verify that `provider` is not-NULL when `master_key` is
+      // not-NULL.
+      BSON_ASSERT_PARAM (provider);
+   }
+
    bson_init (doc_out);
    state_machine = _state_machine_new (crypt);
    state_machine->keyvault_coll = keyvault_coll;
--- a/src/libmongoc/src/mongoc/mongoc-crypt.c
+++ b/src/libmongoc/src/mongoc/mongoc-crypt.c
@@ -2001,11 +2001,8 @@ _mongoc_crypt_rewrap_many_datakey (_mongoc_crypt_t *crypt,
    mongocrypt_binary_t *filter_bin = NULL;
    bool ret = false;
 
-   if (master_key) {
-      // Caller must verify that `provider` is not-NULL when `master_key` is
-      // not-NULL.
-      BSON_ASSERT_PARAM (provider);
-   }
+   // Caller must ensure `provider` is provided alongside `master_key`.
+   BSON_ASSERT (!master_key || provider);
 
    bson_init (doc_out);
    state_machine = _state_machine_new (crypt);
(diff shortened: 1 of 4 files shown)


src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -2001,6 +2001,12 @@ _mongoc_crypt_rewrap_many_datakey (_mongoc_crypt_t *crypt,
    mongocrypt_binary_t *filter_bin = NULL;
    bool ret = false;
 
+   if (master_key) {
+      // Caller must verify that `provider` is not-NULL when `master_key` is
+      // not-NULL.
+      BSON_ASSERT_PARAM (provider);
+   }
+
```
eramongodb, line 2009:
```suggestion
   // Caller must ensure `provider` is provided alongside `master_key`.
   BSON_ASSERT (!master_key || provider);

```
Small simplification.

src/libmongoc/tests/test-mongoc-client-side-encryption.c
```diff
@@ -5717,6 +5717,81 @@ test_rewrap_with_separate_client_encryption (void *unused)
...
+   ASSERT_OR_PRINT (clientEncryption, error);
+
+   // 2. Call ``clientEncryption.rewrapManyDataKey`` with an empty ``filter``
+   // and these options: (see below).
+   {
+      bool ok = mongoc_client_encryption_rewrap_many_datakey (
+         clientEncryption,
+         NULL /* filter */,
+         NULL /* kms_provider */,
+         tmp_bson ("{'foo': 'bar'}"),
```
eramongodb, line 5776:
```suggestion
         tmp_bson ("{}"),
```

src/libmongoc/tests/test-mongoc-client-side-encryption.c
```diff
@@ -5717,6 +5717,81 @@ test_rewrap_with_separate_client_encryption (void *unused)
...
+         NULL /* result */,
+         &error);
+      // Assert an error is returned from the driver suggesting that the
+      // ``provider`` option is required.
+      ASSERT (!ok && "expected error, but got success");
```
eramongodb, line 5781:
```suggestion
      ASSERT_WITH_MSG (!ok, "expected error, but got success");
```

eramongodb (approved):

PR #1260, opened 2023-05-04
Reduce internal component dependencies for mongoc-rpc

eramongodb:
Motivated by CDRIVER-4625. Moves the `_mongoc_rpc_prep_command`, `_mongoc_rpc_compress`, and `_mongoc_rpc_decompress_if_necessary` functions into the components where they are actually required. This significantly reduces the internal component dependencies required by the mongoc-rpc component.


eramongodb:
Reconsidered splitting up the compress and decompress functions and decided to group them in mongoc-cluster instead.

PR #1271, opened 2023-05-16
CDRIVER-2982 test and document duplicate key handling

kevinAlbs:
# Summary

- test bson_t behavior with duplicate keys
- note `bson_iter_find_*` advances to the first key
- note undefined server behavior with duplicate keys

# Background & Motivation

It is possible to constructing `bson_t` with duplicate keys by manually appending, reading BSON data, or parsing JSON. The [BSON specification](https://bsonspec.org) does not require keys be unique. The motivation for testing behavior of `bson_t` with duplicate keys was to ensure this behavior is not unintentionally changed in the future.

DRIVERS-612 notes:
> The documentation MUST mention that the server's behavior related to duplicate key names is undefined

### DIFF
--- a/src/libbson/doc/bson_iter_find.rst
+++ b/src/libbson/doc/bson_iter_find.rst
@@ -20,7 +20,7 @@ Parameters
 Description
 -----------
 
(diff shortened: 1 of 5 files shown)


vector-of-bool (approved):
LGTM

PR #1275, opened 2023-05-18
CDRIVER-4638 Remove "Public Technical Preview" from Queryable Encryption Equality API

kevinAlbs:
# Summary

- update libmongocrypt test dependency to 1.8.0
- remove "Public Technical Preview" from QE API for equality
- revise Queryable Encryption page to document backwards break

# Background & Motivation

Changes are requested in the "Downstream Changes Summary" of DRIVERS-2614


vector-of-bool (approved):
LGTM

PR #1281, opened 2023-05-25
CDRIVER-4603 use setuptools rather than distutils (deprecated) for version calc

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/646e3c8932f417f4c6dc9c06/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


rcsanchez97:
> LGTM. Consider adding a self-test script. Here is one proposal: https://gist.github.com/kevinAlbs/7105dce823f0732a1330cd4e17fc9213

I'll adapt this for each of the projects that uses this script and then I'll submit separate PRs to add this script (along with an invocation of it someplace that makes sense in each project).

PR #1291, opened 2023-05-31
Do not treat a WSL Linux environment as Windows

eramongodb:
When in WSL, the environment is (for all intents and purposes) equivalent to being on the corresponding Linux OS. It is incorrect to treat the OS as Windows in such an environment.

### DIFF
--- a/tools/platform.sh
+++ b/tools/platform.sh
@@ -24,7 +24,7 @@ case "$OSTYPE" in
     linux-*)
         if have-command cmd.exe; then
             _is_wsl=true
-            _is_windows=true
+            _is_unix_like=true
             _os_family=windows
         else
             _is_linux=true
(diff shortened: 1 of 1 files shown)


vector-of-bool (approved):

tools/platform.sh
```diff
@@ -24,7 +24,7 @@ case "$OSTYPE" in
     linux-*)
         if have-command cmd.exe; then
             _is_wsl=true
-            _is_windows=true
+            _is_unix_like=true
```
kevinAlbs, line 27: Update comment at the top of the file `(i.e. $IS_WSL and $IS_WINDOWS)` to `(i.e. $IS_WSL and $IS_UNIX_LIKE`).

kevinAlbs (approved):
LGTM with a comment fix.

PR #1294, opened 2023-06-02
Add C11 and C17 conformance coverage with VS 2017

eramongodb:
## Description

This PR resolves CDRIVER-4549 by adding a `vs2017x64` task to the `std-matrix` EVG variant and resolves CDRIVER-4334 by removing use of the rhel62 variant. Verified by [this patch](https://spruce.mongodb.com/version/647a65e80305b95dc57a294f).

## windows-64-vs2017 -> windows-vsCurrent

As recommended by the distro guidelines, this PR moves all tasks that require the VS 2017 compiler to the "windows-vsCurrent" distro. Tasks that ran on "windows-64-vs2017-test" in the legacy config were moved to "windows-vsCurrent-large".

The task naming convention uses the Windows Server version for the specified distro rather than vsCurrent to avoid redundancy/confusion about the VS version being used. e.g. the "cse-sasl-cyrus-openssl-vs2017-x64-compile" which ran on windows-64-vs2017 is now named "cse-sasl-cyrus-openssl-windows-2019-vs2017-x64-compile" and runs on "windows-vsCurrent".

This migration required updating the "debug-compile-aws" task to use "find-cmake-latest.sh", as the CMake binary selected by "find-cmake.sh" is broken on the "windows-vsCurrent" distro (cannot find MSBuild.exe during CMake configuration).

## VS 2022 Compatibility

Although the C Driver has no issues with VS 2022 compatibility, libmongocrypt needed a patch (https://github.com/mongodb/mongo-c-driver/commit/106768d19b961d476dc56a1687450da954493cc6) in CSE tasks to address an internal compiler error. Applied a cherry-pick of the required patch to minimize changes relative to the 1.8.0 release, which may be removed once the commit is included in the latest minimum libmongocrypt version (bump from 1.8.0).

## RHEL 6.2 Removal

As a drive-by improvement, removed references to the "rhel62" distro as well, which also obviated some test skips.

## C11 and C17 Coverage on Windows

The primary goal of this PR. Given we now use "find-latest-cmake.sh" with CMake version 3.25.2, the `CMAKE_C_STANDARD` workaround is no longer necessary and was thus removed.

Use of `/std:cXY` was complicated by [Windows 10 SDK](https://developercommunity.visualstudio.com/t/stdc17-generates-warning-compiling-windowsh/1249671#T-N1257345) header files not being standard-conforming. This was worked-around by specifying a patched version of the Windows 10 SDK via [CMAKE_SYSTEM_VERSION](CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION). Alas not all warnings are addressed, so `/wd5105` was also added to compiler flags.

## Miscellaneous

- Updated calls to "compile-libmongocrypt.sh" to include stdout output on failure. This was motivated by the ICE error addressed by https://github.com/mongodb/libmongocrypt/commit/fcf2b5b568145df6490cf23d6e0e9c368ff5aa29 not being emitted as stderr output.
- Replaced all calls to GNU Make with `cmake --build` for platform-independent command consistency as well as to address the occasional absence of the "all" target when using the VS 2017 generator (when/why?).


kevinAlbs (approved):
Move to recommended `windows-vsCurrent` distro, and associated CI improvements are much appreciated. TIL about CMAKE_SYSTEM_VERSION.

PR #1300, opened 2023-06-07
CDRIVER-4316 Allow both int32 and int64 for batchSize in cursor opts

adriandole:
## Description

The server allows batchSize to be either `int32` or `int64`. This should also be allowed in the cursor.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-cursor.c
+++ b/src/libmongoc/src/mongoc/mongoc-cursor.c
@@ -45,6 +45,26 @@ _translate_query_opt (const char *query_field,
 
 
 bool
+_mongoc_cursor_set_opt_int32 (mongoc_cursor_t *cursor,
+                              const char *option,
+                              int32_t value)
+{
+   bson_iter_t iter;
+
+   if (bson_iter_init_find (&iter, &cursor->opts, option)) {
+      if (!BSON_ITER_HOLDS_INT32 (&iter)) {
+         return false;
+      }
+
+      bson_iter_overwrite_int32 (&iter, value);
+      return true;
+   }
+
+   return BSON_APPEND_INT32 (&cursor->opts, option, value);
+}
+
+
+bool
 _mongoc_cursor_set_opt_int64 (mongoc_cursor_t *cursor,
                               const char *option,
                               int64_t value)
@@ -1449,9 +1469,16 @@ void
 mongoc_cursor_set_batch_size (mongoc_cursor_t *cursor, uint32_t batch_size)
 {
    BSON_ASSERT (cursor);
+   bson_iter_t iter;
 
-   _mongoc_cursor_set_opt_int64 (
-      cursor, MONGOC_CURSOR_BATCH_SIZE, (int64_t) batch_size);
+   if (bson_iter_init_find (&iter, &cursor->opts, MONGOC_CURSOR_BATCH_SIZE) &&
+       BSON_ITER_HOLDS_INT32 (&iter)) {
+      _mongoc_cursor_set_opt_int32 (
+         cursor, MONGOC_CURSOR_BATCH_SIZE, (int32_t) batch_size);
+   } else {
+      _mongoc_cursor_set_opt_int64 (
+         cursor, MONGOC_CURSOR_BATCH_SIZE, (int64_t) batch_size);
+   }
 }
 
 
--- a/src/libmongoc/tests/test-mongoc-cursor.c
+++ b/src/libmongoc/tests/test-mongoc-cursor.c
@@ -2350,6 +2350,117 @@ test_find_error_is_alive (void)
    mongoc_client_destroy (client);
 }
 
+typedef struct _started_event_t {
+   char *command_name;
+   bson_t *command;
+   struct _started_event_t *next;
+} started_event_t;
+
+static void
+command_started (const mongoc_apm_command_started_t *event)
+{
+   mongoc_array_t *events =
+      (mongoc_array_t *) mongoc_apm_command_started_get_context (event);
+   started_event_t *started_event = bson_malloc0 (sizeof (started_event_t));
+
+   started_event->command =
+      bson_copy (mongoc_apm_command_started_get_command (event));
+   started_event->command_name =
+      bson_strdup (mongoc_apm_command_started_get_command_name (event));
+   _mongoc_array_append_val (events, started_event);
+
+   MONGOC_DEBUG ("Command %s started",
+                 mongoc_apm_command_started_get_command_name (event));
+}
+
+static void
+clear_started_events (mongoc_array_t *events)
+{
+   size_t i;
+   for (i = 0; i < events->len; i++) {
+      started_event_t *started_event =
+         _mongoc_array_index (events, started_event_t *, i);
+      bson_destroy (started_event->command);
+      bson_free (started_event->command_name);
+      bson_free (started_event);
+   }
+   _mongoc_array_clear (events);
+}
+
+/* Test that mongoc_cursor_set_batch_size overrides a previously set int32
+ * batchSize. */
+void
+test_cursor_batchsize_override_int32 (void)
+{
+   mongoc_client_t *client;
+   mongoc_apm_callbacks_t *cbs;
+   mongoc_collection_t *coll;
+   bson_error_t error;
+   mongoc_array_t started_events;
+
+   client = test_framework_new_default_client ();
+   cbs = mongoc_apm_callbacks_new ();
+   _mongoc_array_init (&started_events, sizeof (started_event_t *));
+   mongoc_apm_set_command_started_cb (cbs, command_started);
+   mongoc_client_set_apm_callbacks (client, cbs, &started_events);
+   coll = mongoc_client_get_collection (client, "db", "coll");
+
+   /* Drop and insert two documents into the collection */
+   {
+      bson_t *to_insert = BCON_NEW ("x", "y");
+
+      ASSERT_OR_PRINT (mongoc_collection_drop (coll, &error), error);
+      ASSERT_OR_PRINT (
+         mongoc_collection_insert_one (
+            coll, to_insert, NULL /* opts */, NULL /* reply */, &error),
+         error);
+      ASSERT_OR_PRINT (
+         mongoc_collection_insert_one (
+            coll, to_insert, NULL /* opts */, NULL /* reply */, &error),
+         error);
+      bson_destroy (to_insert);
+   }
+
+   clear_started_events (&started_events);
+
+   /* Create a cursor and iterate once. */
+   {
+      const bson_t *got;
+      bson_t *filter = bson_new ();
+      bson_t *findopts = BCON_NEW ("batchSize", BCON_INT32 (1));
+      mongoc_cursor_t *cursor = mongoc_collection_find_with_opts (
+         coll, filter, findopts, NULL /* read_prefs */);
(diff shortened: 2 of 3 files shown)


src/libmongoc/src/mongoc/mongoc-cursor.c
```diff
@@ -1449,9 +1469,16 @@ void
 mongoc_cursor_set_batch_size (mongoc_cursor_t *cursor, uint32_t batch_size)
 {
    BSON_ASSERT (cursor);
+   bson_iter_t iter;
 
-   _mongoc_cursor_set_opt_int64 (
-      cursor, MONGOC_CURSOR_BATCH_SIZE, (int64_t) batch_size);
+   if (bson_iter_init_find (&iter, &cursor->opts, MONGOC_CURSOR_BATCH_SIZE) &&
```
eramongodb, line 1474: The error handling of `mongoc_cursor_set_batch_size` appears to be underspecified. I would like to set `cursor->error` if `_mongoc_cursor_set_opt_*()` returns `false` (currently ignored), but given that would be an API break (a previously never-erroring function now potentially setting an error), perhaps may be worth adding a `MONGOC_WARNING` on failure to set the option instead.

eramongodb (approved):
Some minor suggestions and comments to consider, but otherwise, LGTM.

src/libmongoc/tests/test-mongoc-cursor.c
```diff
@@ -2350,6 +2350,119 @@ test_find_error_is_alive (void)
...
+   _mongoc_array_init (&started_events, sizeof (started_event_t *));
+   mongoc_apm_set_command_started_cb (cbs, command_started);
+   mongoc_client_set_apm_callbacks (client, cbs, &started_events);
+   coll = mongoc_client_get_collection (client, "db", "coll");
+
+   /* Drop and insert two documents into the collection */
+   {
+      bson_t *to_insert = BCON_NEW ("x", "y");
+
+      ASSERT_OR_PRINT (mongoc_collection_drop (coll, &error), error);
```
kevinAlbs, line 2411:
```suggestion
      // Ignore "ns not found" error on drop.
      mongoc_collection_drop (coll, NULL);
```

Pre 7.0 MongoDB Servers may return "ns not found" error when attempting to drop a collection that does not exist. MongoDB 7.0 changed this behavior as part of DRIVERS-2548.

Suggest removing the assertion.

src/libmongoc/tests/test-mongoc-cursor.c
```diff
@@ -2350,6 +2350,119 @@ test_find_error_is_alive (void)
    mongoc_client_destroy (client);
 }
 
+typedef struct _started_event_t {
+   char *command_name;
+   bson_t *command;
+   struct _started_event_t *next;
```
kevinAlbs, line 2356:
```suggestion
```

Unused.

src/libmongoc/src/mongoc/mongoc-cursor.c
```diff
@@ -1449,9 +1469,16 @@ void
...
+       BSON_ITER_HOLDS_INT32 (&iter)) {
+      _mongoc_cursor_set_opt_int32 (
+         cursor, MONGOC_CURSOR_BATCH_SIZE, (int32_t) batch_size);
```
kevinAlbs, line 1477:
```suggestion
      BSON_ASSERT (bson_in_range_int32_t_unsigned (batch_size));
      bson_iter_overwrite_int32 (&iter, (int32_t) batch_size);
```

Suggest asserting uint32 is in range of int32 before the cast to avoid implementation-defined behavior of unsigned => signed cast.

Suggest removing the `_mongoc_cursor_set_opt_int32` helper, as it is only needed by this caller.

src/libmongoc/tests/test-mongoc-cursor.c
```diff
@@ -2436,4 +2549,7 @@ test_cursor_install (TestSuite *suite)
       suite, "/Cursor/error_document/command", test_error_document_command);
    TestSuite_AddLive (
       suite, "/Cursor/find_error/is_alive", test_find_error_is_alive);
+   TestSuite_AddLive (suite,
+                      "/Cursor/overwrite_int64_batch_size",
```
kevinAlbs, line 2553:
```suggestion
                      "/Cursor/batchsize_override_int32",
```

Suggest matching name of test function.

src/libmongoc/src/mongoc/mongoc-cursor.c
```diff
@@ -1449,9 +1469,16 @@ void
...
+   } else {
+      _mongoc_cursor_set_opt_int64 (
+         cursor, MONGOC_CURSOR_BATCH_SIZE, (int64_t) batch_size);
+   }
```
kevinAlbs, line 1481: I had not considered other numeric types when filing the ticket. The server code appears to specify the `batchSize` option in the IDL as a `safeInt64`: https://github.com/10gen/mongo/blob/3aa51a4623b5869c6f57745ad1371aacf02c8d34/src/mongo/db/query/find_command.idl#L124-L129

A `safeInt64` appears to be representable as `int32`, `int64`, `double`, or `decimal128`: https://github.com/10gen/mongo/blob/3aa51a4623b5869c6f57745ad1371aacf02c8d34/src/mongo/db/basic_types.idl#L66-L74:

```
    safeInt64:
        bson_serialization_type:
                                - long
                                - int
                                - decimal
                                - double
        description: "Accepts any numerical type within int64 range"
        cpp_type: std::int64_t
        deserializer: "mongo::BSONElement::safeNumberLong"
```

I suggest also handling `double` and `decimal` similarly.

src/libmongoc/src/mongoc/mongoc-cursor.c
```diff
@@ -1449,9 +1469,16 @@ void
 mongoc_cursor_set_batch_size (mongoc_cursor_t *cursor, uint32_t batch_size)
 {
    BSON_ASSERT (cursor);
+   bson_iter_t iter;
 
-   _mongoc_cursor_set_opt_int64 (
-      cursor, MONGOC_CURSOR_BATCH_SIZE, (int64_t) batch_size);
+   if (bson_iter_init_find (&iter, &cursor->opts, MONGOC_CURSOR_BATCH_SIZE) &&
+       BSON_ITER_HOLDS_INT32 (&iter)) {
+      BSON_ASSERT (bson_in_range_int32_t_unsigned (batch_size));
```
eramongodb, line 1476: @kevinAlbs Are we sure about this assertion? This is effectively adding a new _precondition_ to this function that the input value is conditionally representable by whatever type `batchSize` happens to be set with. I think it may be more appropriate for this to be a `MONGOC_WARNING` instead.

kevinAlbs: Good catch. I agree using `MONGOC_WARNING` seems more appropriate. I think a caller could expect `mongoc_cursor_set_batch_size (UINT32_MAX)` to succeed if the stored `batchSize` is represented with an int64.

Suggest using a warning and returning early if not in range:
```c
      if (!bson_in_range_int32_t_unsigned (batch_size)) {
         MONGOC_WARNING ("unable to overwrite stored int32 batchSize with "
                         "out-of-range value %" PRIu32,
                         batch_size);
         return;
      }
```

src/libmongoc/src/mongoc/mongoc-cursor.c
eramongodb: Related to / superceded by https://github.com/mongodb/mongo-c-driver/pull/1300/files#r1224707455.

> I agree using `MONGOC_WARNING` seems more appropriate.

src/libmongoc/src/mongoc/mongoc-cursor.c
```diff
@@ -1449,9 +1449,25 @@ void
...
+      bson_iter_overwrite_int32 (&iter, (int32_t) batch_size);
+   } else if (BSON_ITER_HOLDS_DOUBLE (&iter)) {
+      bson_iter_overwrite_double (&iter, (double) batch_size);
+   } else if (BSON_ITER_HOLDS_DECIMAL128 (&iter)) {
+      bson_decimal128_t val;
+      val.high = 0x3040000000000000;
+      val.low = (uint64_t) batch_size;
+      bson_iter_overwrite_decimal128 (&iter, &val);
+   }
```
eramongodb, line 1470: Add appropriate `MONGOC_WARNING` logs on failure. Recommend using `capture_logs()` + `ASSERT_CAPTURED_LOG()` in corresponding failure tests.

src/libmongoc/src/mongoc/mongoc-cursor.c
eramongodb: Bump: replace assertion with `MONGOC_WARNING` to avoid narrowing the contract of this function.

eramongodb (changes requested):

src/libmongoc/tests/test-mongoc-cursor.c
```diff
@@ -2350,6 +2350,165 @@ test_find_error_is_alive (void)
...
+/* Test that mongoc_cursor_set_batch_size overrides a previously set int32
+ * batchSize. */
+void
+test_cursor_batchsize_override_int32 (void)
+{
+   bson_t *findopts = BCON_NEW ("batchSize", BCON_INT32 (1));
+   test_cursor_batchsize_override (findopts, numeric_iter_eq);
+}
+
+/* Test that mongoc_cursor_set_batch_size overrides a previously set in64
```
kevinAlbs, line 2484:
```suggestion
/* Test that mongoc_cursor_set_batch_size overrides a previously set int64
```

kevinAlbs (approved):
LGTM with pending comments addressed.

src/libmongoc/tests/test-mongoc-cursor.c
```diff
@@ -2350,6 +2350,203 @@ test_find_error_is_alive (void)
...
+   test_cursor_batchsize_override (findopts, decimal128_iter_eq);
+}
+
+/* Test that attempting to overwrite an int32 batchSize with an out-of-range
+ * value raises a warning */
+void
+test_cursor_batchsize_override_range_warning (void)
+{
+   mongoc_client_t *client;
+   mongoc_apm_callbacks_t *cbs;
```
eramongodb, line 2519:
```suggestion
```
Unused variable warning.

src/libmongoc/tests/test-mongoc-cursor.c
```diff
@@ -2350,6 +2350,203 @@ test_find_error_is_alive (void)
...
+   mongoc_collection_t *coll;
+   bson_t *findopts = BCON_NEW ("batchSize", BCON_INT32 (1.0));
+   bson_error_t error;
```
eramongodb, line 2522:
```suggestion
```
Unused variable warning.

src/libmongoc/tests/test-mongoc-cursor.c
```diff
@@ -2350,6 +2350,203 @@ test_find_error_is_alive (void)
...
+
+   client = test_framework_new_default_client ();
+   coll = mongoc_client_get_collection (client, "db", "coll");
+
+   /* Create a cursor and attempt to override outside int32 range. */
+   {
+      const bson_t *got;
```
eramongodb, line 2529:
```suggestion
```
Unused variable warning.

eramongodb (approved):
Some unused variables and a missing warning; otherwise, LGTM.

PR #1302, opened 2023-06-08
CDRIVER-4255 Add asserts for functions with client and pool parameters

joshbsiegel:
- Looked up every instance of a function that has a `mongoc_client_t` or `mongoc_client_pool_t` parameter and added an assert to ensure that the respective function would not segfault when one of the parameters is `NULL`
- Decided to use `BSON_ASSERT_PARAM` so that the error message that will print out when the assert fails is specific to issue
- Used plain `ASSERT` in the test files
- In functions where one of the parameters I was looking for was marked as `BSON_UNUSED`, I did not add an assert as it seemed counterintuitive and there are some test cases where a client or pool is passed through to the function as NULL
- In functions where it was clear that the client or pool was not being dereferenced (such as one line functions that just call another functions), I did not add an assert

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
@@ -88,8 +88,6 @@ void
 mongoc_auto_encryption_opts_set_keyvault_client (
    mongoc_auto_encryption_opts_t *opts, mongoc_client_t *client)
 {
-   BSON_ASSERT_PARAM (client);
-
    if (!opts) {
       return;
    }
@@ -101,8 +99,6 @@ void
 mongoc_auto_encryption_opts_set_keyvault_client_pool (
    mongoc_auto_encryption_opts_t *opts, mongoc_client_pool_t *pool)
 {
-   BSON_ASSERT_PARAM (pool);
-
    if (!opts) {
       return;
    }
@@ -270,8 +266,6 @@ void
 mongoc_client_encryption_opts_set_keyvault_client (
    mongoc_client_encryption_opts_t *opts, mongoc_client_t *keyvault_client)
 {
-   BSON_ASSERT_PARAM (keyvault_client);
-
    if (!opts) {
       return;
    }
--- a/src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
@@ -88,6 +88,8 @@ void
 mongoc_auto_encryption_opts_set_keyvault_client (
    mongoc_auto_encryption_opts_t *opts, mongoc_client_t *client)
 {
+   BSON_ASSERT_PARAM (client);
+
    if (!opts) {
       return;
    }
@@ -99,6 +101,8 @@ void
 mongoc_auto_encryption_opts_set_keyvault_client_pool (
    mongoc_auto_encryption_opts_t *opts, mongoc_client_pool_t *pool)
 {
+   BSON_ASSERT_PARAM (pool);
+
    if (!opts) {
       return;
    }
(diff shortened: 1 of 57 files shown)


src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
```diff
@@ -88,6 +88,8 @@ void
 mongoc_auto_encryption_opts_set_keyvault_client (
    mongoc_auto_encryption_opts_t *opts, mongoc_client_t *client)
 {
+   BSON_ASSERT_PARAM (client);
```
kevinAlbs, line 91:
```suggestion
```

Suggest removing this assertion. A caller may be relying on passing `NULL` to clear the option.

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
```diff
@@ -99,6 +101,8 @@ void
 mongoc_auto_encryption_opts_set_keyvault_client_pool (
    mongoc_auto_encryption_opts_t *opts, mongoc_client_pool_t *pool)
 {
+   BSON_ASSERT_PARAM (pool);
```
kevinAlbs, line 104:
```suggestion
```

Suggest removing this assertion. A caller may be relying on passing `NULL` to clear the option.

src/libmongoc/src/mongoc/mongoc-client-side-encryption.c
```diff
@@ -266,6 +270,8 @@ void
 mongoc_client_encryption_opts_set_keyvault_client (
    mongoc_client_encryption_opts_t *opts, mongoc_client_t *keyvault_client)
 {
+   BSON_ASSERT_PARAM (keyvault_client);
+
```
kevinAlbs, line 274:
```suggestion
```

Suggest removing this assertion. A caller may be relying on passing `NULL` to clear the option.

src/libmongoc/src/mongoc/mongoc-opts.c
```diff
@@ -23,6 +23,8 @@ _mongoc_insert_one_opts_parse (
 {
    bson_iter_t iter;
 
+   BSON_ASSERT_PARAM (client);
+
```
kevinAlbs, line 27: mongoc-opts.c is generated by `build/generate-opts.py`. Suggest updating [mongoc-opts.c.template](https://github.com/mongodb/mongo-c-driver/blob/572fd74d70b2294deaff7e760d2c0175099cd331/build/opts_templates/mongoc-opts.c.template#L20) and regenerating mongoc-opts.c with `python ./build/generate-opts.py`.

src/libmongoc/tests/json-test-monitoring.c
```diff
@@ -262,6 +262,7 @@ set_apm_callbacks (json_test_ctx_t *ctx, mongoc_client_t *client)
 {
    mongoc_apm_callbacks_t *callbacks;
 
+   ASSERT (client);
```
kevinAlbs, line 265: `client` is unused in this function. Suggest removing the `mongoc_client_t *client` argument.

src/libmongoc/src/mongoc/mongoc-opts.c
joshbsiegel: What should I be updating in the template? I searched for `ASSERT`s but didn't find any.

kevinAlbs: Add `BSON_ASSERT_PARAM (client);` to the template that generates the function:

```
bool
_{{ struct_name }}_parse (
   mongoc_client_t *client,
   const bson_t *opts,
   {{ struct_type }} *{{ struct_name }},
   bson_error_t *error)
{
```

And regenerate mongoc-opts.c by running `python ./build/generate-opts.py`.

kevinAlbs (approved):

PR #1306, opened 2023-06-13
Document dropping support of Ubuntu 14.04 and Debian 8.1

kevinAlbs:
Resolves CDRIVER-4542 and CDRIVER-4541.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,9 +1,14 @@
 libmongoc 1.24.0 (Unreleased)
 =============================
 
+Platform Support:
+
   * Support for macOS 10.14 will be dropped in an upcoming release.
   * Support for Ubuntu 14.04 will be dropped in an upcoming release.
   * Support for Debian 8.1 will be dropped in an upcoming release.
+
+Build Configuration:
+
   * The `ENABLE_BSON` configure-time option has been removed. Build scripts that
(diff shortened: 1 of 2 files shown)


NEWS
```diff
@@ -1,3 +1,14 @@
+libmongoc 1.24.0 (Unreleased)
+=============================
+
+  * Support for macOS 10.14 will be dropped in an upcoming release.
+  * Support for Ubuntu 14.04 will be dropped in an upcoming release.
+  * Support for Debian 8.1 will be dropped in an upcoming release.
+  * The `ENABLE_BSON` configure-time option has been removed. Build scripts that
```
eramongodb, line 7: Given updates to supported platforms may become a regular occurrence + we do not(?) consider CMake configuration breaking changes as deserving a major version bump, suggest splitting the first three items from the fourth item into dedicated sections:

```
Platform Support:

  * Support for MacOS 10.14 ...
  * ...

Build Configuration:

  * The `ENABLE_BSON` CMake option is no longer supported. ...
```

eramongodb (approved):

NEWS
kevinAlbs: I like that. Sections have been separated.

> we do not(?) consider CMake configuration breaking changes as deserving a major version bump

I think CMake configuration breaking changes do not require a major version bump.

PR #1307, opened 2023-06-13
CDRIVER-4617 Refactor libmongoc to use mcd-rpc

eramongodb:
## Description

Followup to https://github.com/mongodb/mongo-c-driver/pull/1296. This PR replaces `mongoc_rpc_t` with `mcd_rpc_message` in libmongoc. This PR deliberately leaves the (now unused) `mongoc_rpc_t` utilities to facilitate comparison of equivalent utilities and avoid messy diffs. The `mongoc_rpc_t` utilities will be removed in a followup commit.

Verified by ~[this patch](https://spruce.mongodb.com/version/6488d60432f417d1d7205efa)~ [this patch](https://spruce.mongodb.com/version/6488e608562343fac1777ac5) (includes the commit addressing distinct pointer type warnings on Windows).

## RPC Utility Functions

There are four RPC message utility functions in libmongoc that require new equivalents:

- `_mongoc_rpc_op_(egress|ingress)_inc` -> `mcd_rpc_message_(egress|ingress)`
- `_mongoc_rpc_(compress|decompress)` -> `mcd_rpc_message_(compress|decompress)`
- `_mongoc_rpc_decompress_if_necessary` -> `mcd_rpc_message_decompress_if_necessary`
- `_mongoc_rpc_reply_get_first_msg` -> `mcd_rpc_message_get_body`
- `_mongoc_rpc_check_ok` -> `mcd_rpc_message_check_ok`

The new ingress and egress functions are fairly straightforward. Compared to the old `mongoc_rpc_t` equivalents which conflated the counters with conversion from data / to iovecs (partially addressed for the egress case by https://github.com/mongodb/mongo-c-driver/pull/1256), the `mcd_rpc_message` functions completely separate conversion (with `from_data` and `to_iovecs`) from counter increment (with `egress` and `ingress`).

The new compression functions (`compress`, `decompress`, and `decompress_if_necessary`) remove the dependency on `mongoc_cluster_t` by requesting the compressor level (derived from the cluster's URI) explicitly as a parameter instead. They also do not handle ingress/egress counter increments (as currently being done by the `_mongoc_rpc_scatter` function) to improve separation of concerns.

The `get_first_document` function was renamed to `get_body` to better reflect it's intended use (obtaining the Kind 0 section BSON document of an OP_MSG or the first document in an OP_REPLY). The function now properly handles the possibility of the Kind 0 section not being the first section in the OP_MSG, although this freedom does not appear to have been exercised yet by the server.

The `check_ok` function is a simple translation and does not require further elaboration.

## mongoc_cluster_t::request_id

A drive-by fix to make the type of the request ID consistent with its specification and use.

## Mock Server

The new `mcd_rpc_message` utilities already exposed inconsistent opcode handling of RPC messages that were addressed in https://github.com/mongodb/mongo-c-driver/pull/1280 (caught by `mcd_rpc_op_query_get_flags` in `assert_request_matches_flags`). This PR contains the changes that led to their discovery. Most of the changes involve improvements to type consistency, greater detail on message parsing failure, and refactors to simplify code (notably, removal of the redundant `rpc` parameter for the `request_from_*` functions).

## Server Monitor Thread

A fairly straightforward one-to-one translation from `mongoc_rpc_t` to `mcd_rpc_message` equivalents. Nearby code was updated to improve readability.

## mongoc-async-cmd

Also fairly straightforward one-to-one translation, but removed the redundant `mongoc_async_cmd::array` data member in favor of local variables to store the iovec structures used for async send operations. The existing `mongoc_async_cmd::iovecs` data member is used directly instead (now made an owning pointer).

## mongoc_cluster_run_command_opquery

Refactored into `_mongoc_cluster_run_command_opquery_send` and `_mongoc_cluster_run_command_opquery_recv`. The old implementation used a truly bizarre algorithm when receiving the reply that assumes the message is an OP_REPLY, receiving the incoming bytes as if it is an OP_REPLY, then later realizing it may _actually_ be an OP_COMPRESSED message instead, then reading the remaining bytes of the OP_COMPRESSED message. The implementation was changed to use the more consistent pattern of reading `messageLength` only first, before consuming all the remaining bytes of the incoming message in one go, _then_ validating the opcode of the entire message, also aided by the much nicer-to-use `_mongoc_buffer_append_from_stream` interface.

## mongoc_cluster_run_opmsg

Also refactored into separate `_mongoc_cluster_run_opmsg_send` and `_mongoc_cluster_run_opmsg_recv` functions. It's mostly straightforward translation with the exception of more thorough error-handling checks.

I remain fairly confused by the inconsistent use of `RUN_CMD_ERR_DECORATE`, `_handle_network_error`, `server_stream->stream = NULL`, and `network_error_reply (reply, cmd)` (including their ordering) in the original code. I would appreciate a closer look into their use to confirm they are being used correctly in the new functions.

## Cursor Functions

The only major change of note that isn't straightforward one-to-one translation is the use of local `iovecs` instead of the `mongoc_cluster_t::iov` array object, a small refactor of some blocks of code into `_mongoc_cursor_op_*_send` functions, and improved type consistency.


src/libmongoc/src/mongoc/mongoc-async-cmd.c
```diff
@@ -134,42 +134,47 @@ mongoc_async_cmd_run (mongoc_async_cmd_t *acmd)
...
    }
 
+   mcd_rpc_message_set_length (acmd->rpc, message_length);
+
    /* This will always be hello, which are not allowed to be compressed */
-   _mongoc_rpc_gather (&acmd->rpc, &acmd->array);
-   acmd->iovec = (mongoc_iovec_t *) acmd->array.data;
-   acmd->niovec = acmd->array.len;
-   _mongoc_rpc_swab_to_le (&acmd->rpc);
+   acmd->iovec = mcd_rpc_message_to_iovecs (acmd->rpc, &acmd->niovec);
```
vector-of-bool, line 175: Aside: I seem to recall some confusion expressed about the meaning of "gather" in this context. For reference, the word refers to "scatter/gather I/O", which is just a different name for vectored I/O.

src/libmongoc/src/mongoc/mongoc-rpc.c
```diff
@@ -959,6 +959,58 @@ _mongoc_rpc_reply_get_first (mongoc_rpc_reply_t *reply, bson_t *bson)
...
+            return bson_init_static (reply, body, (size_t) body_len);
+         }
+
+         case 1: // Document Sequence.
+            continue;
+
+         default:
+            // Validated by `mcd_rpc_message_from_data`.
+            BSON_UNREACHABLE ("invalid OP_MSG section kind");
+         }
```
vector-of-bool, line 988: Is it the case that a hostile/broken peer could generate invalid messages and terminate the client application?

vector-of-bool (approved):
LGTM, and the new form is often significantly more readable than the prior version.

My only concern is about validation of received messages. I'm unsure how much of a concern it is, or if there are unseen checks that guard against firing assertions.

src/libmongoc/src/mongoc/mongoc-async-cmd.c
eramongodb: Personally, I am constantly confused by the perspective of the terminology. I am tempted to interpret the operations as "gathering" data into an RPC object and "scattering" data into iovec data structures (where the perspective is relative to the RPC object). This happens be the exact opposite of how libmongoc currently defines them: `mongoc_rpc_gather` "gathers" data from an RPC object into an array of iovecs and (this is truly the most confusing part) `mongoc_rpc_scatter` "scatters" data from a buffer into an RPC object (???). It doesn't help that "scatter/gather I/O" doesn't seem to rigorously define these terms anyways (their concern is with the vectored read/write interfaces, not with how those vectors are created), so `mcd-rpc` deliberately does away with that terminology in favor of a much more concrete vocabulary (from data, to iovecs).

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -3532,213 +3496,292 @@ network_error_reply (bson_t *reply, mongoc_cmd_t *cmd)
...
+   const int32_t message_length = _int32_from_le (buffer.data);
+
+   if (message_length < message_header_length ||
+       message_length > server_stream->sd->max_msg_size) {
+      RUN_CMD_ERR (MONGOC_ERROR_PROTOCOL,
+                   MONGOC_ERROR_PROTOCOL_INVALID_REPLY,
+                   "message length %" PRId32
+                   " is not within valid range of 16-%" PRId32 " bytes",
+                   message_length,
+                   server_stream->sd->max_msg_size);
```
kevinAlbs, line 3644:
```suggestion
      RUN_CMD_ERR (MONGOC_ERROR_PROTOCOL,
                   MONGOC_ERROR_PROTOCOL_INVALID_REPLY,
                   "message length %" PRId32
                   " is not within valid range of %" PRId32 "-%" PRId32
                   " bytes",
                   message_header_length,
                   message_length,
                   server_stream->sd->max_msg_size);
```

Use `message_header_length` constant.

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -3355,139 +3346,112 @@ mongoc_cluster_legacy_rpc_sendv_to_server (
...
+   if (message_length < message_header_length ||
+       message_length > max_msg_size) {
       bson_set_error (error,
                       MONGOC_ERROR_PROTOCOL,
                       MONGOC_ERROR_PROTOCOL_INVALID_REPLY,
-                      "Corrupt or malicious reply received.");
+                      "message length %" PRId32
+                      " is not within valid range of 16-%" PRId32 " bytes",
+                      message_length,
+                      server_stream->sd->max_msg_size);
```
kevinAlbs, line 3401: Use `message_header_length`: 

```c
      bson_set_error (error,
                      MONGOC_ERROR_PROTOCOL,
                      MONGOC_ERROR_PROTOCOL_INVALID_REPLY,
                      "message length %" PRId32
                      " is not within valid range of %" PRId32 " -%" PRId32
                      " bytes",
                      message_header_length,
                      message_length,
                      server_stream->sd->max_msg_size);
```

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -3532,213 +3496,292 @@ network_error_reply (bson_t *reply, mongoc_cmd_t *cmd)
...
+                                              : MONGOC_OP_MSG_FLAG_MORE_TO_COME;
+
+   {
+      int32_t message_length = 0;
+
+      message_length += mcd_rpc_header_set_message_length (rpc, 0);
+      message_length +=
+         mcd_rpc_header_set_request_id (rpc, ++cluster->request_id);
+      message_length += mcd_rpc_header_set_response_to (rpc, 0);
+      message_length += mcd_rpc_header_set_op_code (rpc, MONGOC_OPCODE_MSG);
```
kevinAlbs, line 3523:
```suggestion
      message_length += mcd_rpc_header_set_op_code (rpc, MONGOC_OP_CODE_MSG);
```

kevinAlbs (approved):
Nicely done. LGTM with minor comments addressed.

kevinAlbs:
> I remain fairly confused by the inconsistent use of `RUN_CMD_ERR_DECORATE`,
`_handle_network_error`, `server_stream->stream = NULL`, and
`network_error_reply (reply, cmd)` (including their ordering) in the original
code. I would appreciate a closer look into their use to confirm they are being
used correctly in the new functions.

It looks correct to me.

For some background on `server_stream->stream = NULL`: I think `server_stream->stream = NULL;` is to prevent reusing an invalidated `mongoc_server_stream_t` in functions that use the same `mongoc_server_stream_t` for multiple commands.
`mongoc_cluster_stream_valid` is used to determine if a `mongoc_server_stream_t` is invalid.

A network error may free the `mongoc_server_stream_t::mongoc_stream_t`:
- For single threaded via: `_handle_network_error` => `mongoc_cluster_disconnect_node` => `mongoc_topology_scanner_node_disconnect`.
- For pooled via: `_handle_network_error` => `mongoc_cluster_disconnect_node` => `_mongoc_cluster_node_destroy`.

Adding `server_stream->stream = NULL` to error handling in `_mongoc_cluster_run_command_opquery` and `_mongoc_cluster_run_command_opquery_recv` may be a reasonable precaution. But I expect it would not change any observable behavior. OP_QUERY is only used for exhaust cursors and the handshake. Neither use the same stream for multiple commands.

eramongodb:
Latest changes verified by [this patch](https://spruce.mongodb.com/version/648c99a8c9ec443ed71a6df4).

PR #1310, opened 2023-06-15
CDRIVER-4533 Document behavior of bson_new_from_json with multiple JSONs

adriandole:


src/libbson/doc/bson_new_from_json.rst
```diff
@@ -21,8 +21,10 @@ Parameters
 Description
 -----------
 
-The ``bson_new_from_json()`` function allocates and initialize a new :symbol:`bson_t` by parsing the JSON found in ``data``. Only a single JSON object may exist in ``data`` or an error will be set and NULL returned.
-
+The ``bson_new_from_json()`` function allocates and initialize a new :symbol:`bson_t` by parsing the JSON found in ``data``.
+If there are multiple comma-separated JSONs in ``data``, the keys from all JSONs are merged in the returned BSON.
+If the first extra character after the last JSON is ``{``, all following characters are ignored and no error is set.
+Otherwise, an error will be set and NULL returned. 
 Errors
```
kevinAlbs, line 28: Suggest adding tests for this behavior. That will prevent code changes from breaking this now documented behavior.

src/libbson/doc/bson_new_from_json.rst
```diff
@@ -21,8 +21,10 @@ Parameters
 Description
 -----------
 
-The ``bson_new_from_json()`` function allocates and initialize a new :symbol:`bson_t` by parsing the JSON found in ``data``. Only a single JSON object may exist in ``data`` or an error will be set and NULL returned.
-
+The ``bson_new_from_json()`` function allocates and initialize a new :symbol:`bson_t` by parsing the JSON found in ``data``.
+If there are multiple comma-separated JSONs in ``data``, the keys from all JSONs are merged in the returned BSON.
+If the first extra character after the last JSON is ``{``, all following characters are ignored and no error is set.
```
eramongodb, line 26:
```suggestion
If the first character encountered after the last valid JSON object is ``{``, all following characters are ignored and no error is set.
```
Are we sure we do not want to treat this condition as an error instead?

eramongodb: Formerly documented behavior permitted "only a single JSON object", meaning multi-document behavior should have technically been treated as an error (even if it actually ended up merging multiple documents, as it is now specified). I think we may be justified in taking this opportunity to explicitly declare invalid input to be treated as an error. @kevinAlbs thoughts?

kevinAlbs: I think these behaviors of `bson_new_from_json` and `bson_init_from_json` are surprising:

1. `{"a": 1} {"b": 2}` is parsed the same as `{"a": 1}`.
2. `{"a": 1},{"b": 2}` is parsed the same as `{"a": 1, "b": 2}`.

`bson_new_from_json` and `bson_init_from_json` document:

> Only a single JSON object may exist in ``data`` or an error will be set and NULL returned.

That suggests (1) is expected to error and this is a bug in the behavior. I suggest:
- Change the behavior of (1) to be an error for `bson_new_from_json` and `bson_init_from_json`.
- Document in `bson_new_from_json` and `bson_init_from_json` that reading multiple documents should use `bson_reader_t`. If users are mistakenly relying on this behavior, this may help to discover the recommended solution. Here is an example using `bson_reader_t`:

```c
{
    const char *input = BSON_STR ({"a" : 1} {"b" : 2});
    bson_json_reader_t *reader =
        bson_json_data_reader_new (false /* allow multiple */, 0);
    bson_json_data_reader_ingest (
        reader, (const uint8_t *) input, strlen (input));
    // Check first document.
    {
        bson_t got = BSON_INITIALIZER;
        ASSERT_CMPINT (bson_json_reader_read (reader, &got, &error), ==, 1);
        ASSERT (bson_equal (&got, tmp_bson ("{'a': 1}")));
    }
    // Check second document.
    {
        bson_t got = BSON_INITIALIZER;
        ASSERT_CMPINT (bson_json_reader_read (reader, &got, &error), ==, 1);
        ASSERT (bson_equal (&got, tmp_bson ("{'b': 2}")));
    }
    // Check exhausted.
    {
        bson_t got = BSON_INITIALIZER;
        ASSERT_CMPINT (bson_json_reader_read (reader, &got, &error), ==, 0);
    }
    bson_json_reader_destroy (reader);
}
```

I am less sure about (2). If (2) is changed to error, should it also error in `bson_json_reader_read`? If so, that may not be as easy for users to work around. 

My inclination is to document (2) parsing behavior as deprecated, and file a CDRIVER ticket with fixVersion 2.0.0 to consider removing this behavior in a the next major release.

@jmikola was CDRIVER-4533 motivated by seeing strings of multiple documents passed to `bson_new_from_json` or `bson_init_from_json`? I want to know if there are known use cases for this parsing behavior.

src/libbson/doc/bson_new_from_json.rst
```diff
@@ -21,8 +21,10 @@ Parameters
 Description
 -----------
 
-The ``bson_new_from_json()`` function allocates and initialize a new :symbol:`bson_t` by parsing the JSON found in ``data``. Only a single JSON object may exist in ``data`` or an error will be set and NULL returned.
-
+The ``bson_new_from_json()`` function allocates and initialize a new :symbol:`bson_t` by parsing the JSON found in ``data``.
```
kevinAlbs, line 24:
```suggestion
The ``bson_new_from_json()`` function allocates and initializes a new :symbol:`bson_t` by parsing the JSON found in ``data``.
```

src/libbson/tests/test-json.c
```diff
@@ -2445,6 +2447,60 @@ test_bson_json_array_subdoc (void)
...
+   bson_init (&compare);
+   bson_append_int32 (&compare, "a", 1, 12);
+   bson_append_int32 (&compare, "b", 1, 34);
+   bson_append_int32 (&compare, "c", 1, 56);
+   bson_append_int32 (&compare, "d", 1, 78);
+
+   bson_eq_bson (b, &compare);
+   bson_destroy (b);
+   bson_destroy (&compare);
+}
```
kevinAlbs, line 2470: Suggested test simplification:

```c
static void
test_bson_json_merge_multiple (void)
{
   bson_error_t error;
   bson_t *b = bson_new_from_json (
      (const uint8_t *) BSON_STR ({"a" : 1}, {"b" : 2}), -1, &error);
   ASSERT_OR_PRINT (b, error);

   bson_t *compare = bson_new_from_json (
      (const uint8_t *) BSON_STR ({"a" : 1, "b" : 2}), -1, &error);

   bson_eq_bson (b, compare);
   bson_destroy (b);
   bson_destroy (compare);
}
```

Using `BSON_STR` for the string literals representing valid JSON escapes the double quotes, and may improve syntax highlighting.

src/libbson/tests/test-json.c
```diff
@@ -2445,6 +2447,60 @@ test_bson_json_array_subdoc (void)
...
+      bson_t compare;
+      bson_init (&compare);
+      bson_append_int32 (&compare, "a", 1, 12);
+      bson_append_int32 (&compare, "b", 1, 34);
+
+      bson_eq_bson (b, &compare);
+      bson_destroy (b);
+      bson_destroy (&compare);
+   }
+}
```
kevinAlbs, line 2502: Suggested test simplification:

```c
static void
test_bson_json_extra_chars (void)
{
   bson_error_t error;

   {
      bson_t *b =
         bson_new_from_json ((const uint8_t *) "{\"a\": 1}abc", -1, &error);
      ASSERT (b == NULL);
      ASSERT_ERROR_CONTAINS (error,
                             BSON_ERROR_JSON,
                             BSON_JSON_ERROR_READ_CORRUPT_JS,
                             "Got parse error at \"a\"");
   }

   {
      bson_t *b = bson_new_from_json (
         (const uint8_t *) "{\"a\" : 1}{abc,[],{},123", -1, &error);
      ASSERT_OR_PRINT (b, error);
      bson_t *compare =
         bson_new_from_json ((const uint8_t *) "{\"a\" : 1}", -1, &error);

      bson_eq_bson (b, compare);
      bson_destroy (b);
      bson_destroy (compare);
   }
}
```

src/libbson/doc/bson_new_from_json.rst
jmikola: I don't recall [CDRIVER-4533](https://jira.mongodb.org/browse/CDRIVER-4533) being motivated by any runtime tests. It was just something I noticed while reading the code.

It's still not clear to me where the parsing behavior in (2) is coming from. Is this entirely caused by jsonsl, or something in libbson?

Given that the functions were already documented as only accepting a single JSON object, I don't see any reason this can't be considered a longstanding bug fix.

src/libbson/doc/bson_init_from_json.rst
```diff
@@ -27,6 +27,11 @@ Description
 
 The ``bson_init_from_json()`` function will initialize a new :symbol:`bson_t` by parsing the JSON found in ``data``. Only a single JSON object may exist in ``data`` or an error will be set and false returned.
 
+Deprecated behavior: If there are multiple comma-separated JSONs in ``data``, the keys from all JSONs are merged in the returned BSON.
+For example, ``{"a": 1},{"b": 2}`` is parsed as ``{"a": 1, "b": 2}``. If the first character encountered after the last valid
+JSON object is ``{``, all following characters are ignored and no error is set.
+Otherwise, an error will be set and NULL returned. 
+
 ``data`` should be in `MongoDB Extended JSON <https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/>`_ format.
```
kevinAlbs, line 35:
```suggestion
``data`` should be in `MongoDB Extended JSON <https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/>`_ format.


Deprecated behavior
-------------------

If there are multiple comma-separated JSONs in ``data``, the keys from all JSONs are merged in the returned BSON.
For example, ``{"a": 1},{"b": 2}`` is parsed as ``{"a": 1, "b": 2}``.

If the first character encountered after the last valid
JSON object is ``{``, all following characters are ignored and no error is set.
Otherwise, an error will be set and NULL returned.

This deprecated behavior is subject to change in a future release.
```

Suggest splitting into a separate section and nothing this may change in a future release.

src/libbson/tests/test-json.c
```diff
@@ -3639,4 +3689,4 @@ test_json_install (TestSuite *suite)
    TestSuite_Add (suite,
                   "/bson/as_json_with_opts/all_types",
                   test_bson_as_json_with_opts_all_types);
-}
+}
```
eramongodb, line 3692: Missing EOF newline.

eramongodb (approved):

kevinAlbs (approved):
LGTM with suggested documentation change.

PR #1312, opened 2023-06-16
CDRIVER-4617 Remove legacy mongoc-rpc entities

eramongodb:
Followup to https://github.com/mongodb/mongo-c-driver/pull/1307. Removes old `mongoc_rpc_t` code that is now no longer being used (entirely replaced by `mcd_rpc_message` code).


kevinAlbs (approved):
Satisfying. LGTM

vector-of-bool (approved):
Muy bien

PR #1314, opened 2023-06-20
CDRIVER-4665 do not assert `client` in option parsing

kevinAlbs:
# Summary
- do not assert `client` is not NULL in option parsing.

# Background & Motivation

Follow up to https://github.com/mongodb/mongo-c-driver/pull/1302. `client` may expectedly be NULL when parsing options. See the included regression test for an example.

### DIFF
--- a/build/opts_templates/mongoc-opts.c.template
+++ b/build/opts_templates/mongoc-opts.c.template
@@ -19,7 +19,7 @@ _{{ struct_name }}_parse (
 {
    bson_iter_t iter;
 
-   BSON_ASSERT_PARAM (client);
+   BSON_ASSERT (client || true); // client may be NULL.
 
 {% for path, opt_name, info in paths(description) %}
 {% if info['type'] == 'bool' %}
--- a/src/libmongoc/tests/test-mongoc-bulk.c
+++ b/src/libmongoc/tests/test-mongoc-bulk.c
@@ -5011,6 +5011,35 @@ test_bulk_write_multiple_errors (void *unused)
    mongoc_client_destroy (client);
 }
 
+// test_bulk_write_set_client_after_operation is a regression test for
+// CDRIVER-4665.
+static void
+test_bulk_write_set_client_after_operation (void)
+{
+   mongoc_client_t *client;
+   mongoc_bulk_operation_t *bulk;
+   bson_t reply;
+   bson_error_t error;
+   bool ok;
+
+   client = test_framework_new_default_client ();
+   BSON_ASSERT (client);
+   mongoc_client_set_appname (client, "test_bulk_write_client_after_operation");
+   bulk = mongoc_bulk_operation_new (false /* ordered */);
+   ok = mongoc_bulk_operation_insert_with_opts (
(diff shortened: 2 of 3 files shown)


build/opts_templates/mongoc-opts.c.template
```diff
@@ -19,7 +19,7 @@ _{{ struct_name }}_parse (
 {
    bson_iter_t iter;
 
-   BSON_ASSERT_PARAM (client);
+   BSON_ASSERT (client || true); // client may be NULL.
```
jmikola, line 22: Is there any value to retaining this line given `|| true`? It might make more sense to replace this with a comment explaining why you don't assert `client` at all, and how it's not really necessary since the option validation itself doesn't work with a client and any called functions that do will perform a real assertion.

Up to you.

src/libmongoc/tests/test-mongoc-bulk.c
```diff
@@ -5011,6 +5011,35 @@ test_bulk_write_multiple_errors (void *unused)
    mongoc_client_destroy (client);
 }
 
+// test_bulk_write_set_client_after_operation is a regression test for
+// CDRIVER-4665.
+static void
+test_bulk_write_set_client_after_operation (void)
```
jmikola, line 5017: Thanks for adding a regression test. I was a bit pressed for time when I created [CDRIVER-4665]https://jira.mongodb.org/browse/CDRIVER-4665) earlier an intentionally omitted that.

build/opts_templates/mongoc-opts.c.template
kevinAlbs: I have seen the `|| true` pattern recently introduced as a way to document "this option has been considered, and is explicitly OK to be NULL". [Example](https://github.com/mongodb/mongo-c-driver/blob/6f58b9d7f32d5ea83b32c162a59d1f97e958657b/src/libmongoc/src/mongoc/mongoc-cluster.c#L2410-L2420).

src/libmongoc/tests/test-mongoc-bulk.c
```diff
@@ -5011,6 +5011,35 @@ test_bulk_write_multiple_errors (void *unused)
...
+{
+   mongoc_client_t *client;
+   mongoc_bulk_operation_t *bulk;
+   bson_t reply;
+   bson_error_t error;
+   bool ok;
+
+   client = test_framework_new_default_client ();
```
jmikola, line 5025: Does the regression test actually need a `client`, or is this just to demonstrate that you can ultimately execute the bulk operation.

The most minimal test case would be to create a bulk via `mongoc_bulk_operation_new()` (never calling `mongoc_bulk_operation_set_client()`) and then add one operation of each type w/o any assertion errors. I agree the additional operations are redundant given how the opts functions are generated.

jmikola (approved):
A few comments but I'll leave you to take or discard the suggestions as you wish.

I can confirm the regenerated `mongoc-opts.c` file resolves our failing tests in PHPC.

PR #1315, opened 2023-06-21
CDRIVER-4185 log and do not update state when creating thread fails

kevinAlbs:
# Summary

- Log and do not update state when creating thread fails.
- Add `errno_out` arg to `mcommon_thread_create`.

# Background & Motivation

CDRIVER-4185 reports an access violation attempting to join a thread after creating failed.

https://man7.org/linux/man-pages/man3/pthread_create.3.html notes:

> On success, pthread_create() returns 0; on error, it returns an error number, and the contents of *thread are undefined.

Using `BSON_ASSERT` to abort on failure to create a thread was considered. However, IMO the results of a failed thread create do not seem necessary to abort the process. I expect the following:
- Failure to create a server monitor thread would result in the server not being selectable.
- Failure to create a Round-Trip Time monitor thread would result in a stale estimate of the Round Trip Time and impact [the following](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#rtt):
> The RTT is used to support localThresholdMS from the Server Selection spec and timeoutMS from the Client Side Operations Timeout Spec.
- Failure to create the SRV polling thread would result in stale SRV records.

### DIFF
--- a/src/common/common-thread-private.h
+++ b/src/common/common-thread-private.h
@@ -119,10 +119,14 @@ typedef struct {
  * libbson and libmongoc statically. */
 int
 mcommon_thread_join (bson_thread_t thread);
+// mcommon_thread_create returns 0 on success. Returns non-zero on error.
+// `errno_out` is set to an error code on error. Callers may use
+// `bson_strerror_r` to get an error message from `errno_out`.
 int
 mcommon_thread_create (bson_thread_t *thread,
                        BSON_THREAD_FUN_TYPE (func),
-                       void *arg);
+                       void *arg,
+                       int *errno_out);
 
 #if defined(MONGOC_ENABLE_DEBUG_ASSERTIONS) && defined(BSON_OS_UNIX)
 #define mcommon_mutex_is_locked COMMON_NAME (mutex_is_locked)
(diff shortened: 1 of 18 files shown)


src/common/common-thread-private.h
```diff
@@ -119,10 +119,14 @@ typedef struct {
...
 mcommon_thread_join (bson_thread_t thread);
+// mcommon_thread_create returns 0 on success. Returns non-zero on error.
+// `errno_out` is set to an error code on error. Callers may use
+// `bson_strerror_r` to get an error message from `errno_out`.
 int
 mcommon_thread_create (bson_thread_t *thread,
                        BSON_THREAD_FUN_TYPE (func),
-                       void *arg);
+                       void *arg,
+                       int *errno_out);
```
vector-of-bool, line 129: Could we simple return the error number as the return value of the function, rather than using the out-param?

vector-of-bool (approved):
LGTM, although I think that returning the error code may be simpler.

src/common/common-thread-private.h
kevinAlbs: I like that idea. That simplifies the changes. Updated to return an error code rather than use an out-param.

PR #1317, opened 2023-06-21
CDRIVER-4119 Replace ICU SASLprep with internal implementation

joshbsiegel:
# Summary
This PR will replace the ICU library that was used for SASLprep with an internal implementation of SASLprep. Once merged, a separate PR will be made to remove the `ENABLE_ICU` option and related code.

# Background
[SASLprep](https://datatracker.ietf.org/doc/html/rfc4013) defines the steps required to prepare usernames and passwords for comparison and authentication. This is necessary as the vast amount of UTF-8 [characters](https://www.fileformat.info/info/charset/UTF-8/list.htm) that a user can input for a password presents the need for some standardization. For the C driver, we use it to prepare passwords when authenticating with SCRAM-SHA-256. SASLprep is simply a profile built upon [stringprep](https://datatracker.ietf.org/doc/html/rfc3454#section-2) that explains specifically what parts of the stringprep document to abide by.  From a high level, here is what is required by SASLprep:

1. Mapping: There are some characters that are similar to a space character and should be mapped to that. There are also characters that should be mapped to nothing (deletion).
2. Normalization: SASLprep requires [Unicode normalization form KC](https://unicode.org/reports/tr15/). There are many changes that this normalization will make. For example, a character such as "Ⅳ" (one character) will be modified to "IV" (two characters).
3. Prohibited characters: Some characters are fully prohibited.
4. Bidirectional characters: While most characters are left to right, there are some characters that are right to left. These characters are allowed but must be checked for some requirements (see code comments).

Previously SASLprep was done with the ICU library and the goal of this ticket is to get rid of that dependency.

# [utf8proc](https://juliastrings.github.io/utf8proc/)
The normalization described above is the most involved step and we opted to use a library to handle it. utf8proc is bundled as a part of this ticket and handles this.

# Preliminary Work
All of the references to characters in the stringprep and SASLprep RFC documents (such as what characters are mapped to what, prohibited characters, etc.) are displayed using their Unicode [codepoints](https://en.wikipedia.org/wiki/Code_point). This is *not* the same as a character's byte encoding in UTF-8 (see [here](https://www.fileformat.info/info/charset/UTF-8/list.htm)). Because of this, all of the lookup tables that are required use Unicode codepoints, which is different than the `const char *in_utf8` parameter in the existing `_mongoc_sasl_prep_impl` function. As such, it was necessary to convert the UTF-8 characters to their Unicode codepoints before doing any other work.

UTF-8 characters can be 1-4 bytes, so that is why `_mongoc_utf8_string_length` and `_mongoc_utf8_char_length` were added instead of using `strlen` (which will return the number of bytes in the string).  The length of each UTF8 character can be determined by the first byte. `_mongoc_utf8_to_unicode` handles the conversion. 

# Lookup Tables
The mapping, prohibited characters, and bidirectional steps all use lookup tables to easily find whether a codepoint belongs in a particular category. The tables work by using ranges, where the `2n` element is the lower range and the `2n + 1` element is the upper range (both inclusive). `_mongoc_is_code_in_table` will search through the table passed through and return whether the codepoint exists in the table. There might be a potential improvement here as I believe a binary search could be implemented for a faster lookup.

# Misc
All of the tables and much of the instructions on how to perform SASLprep actually come from the stringprep RFC [document](https://datatracker.ietf.org/doc/html/rfc3454#section-2). The SASLprep document just specifies what parts of stringprep to follow (which is most of it).

# What's New
- `_mongoc_sasl_prep_impl` was almost completely scrapped for the SASLprep steps listed above
- `_mongoc_utf8_char_length`
- `_mongoc_utf8_string_length`
- `_mongoc_utf8_is_valid`
- `_mongoc_char_between_chars`
- `_mongoc_is_code_in_table`
- `_mongoc_utf8_to_unicode`
- `_mongoc_unicode_to_utf8`
- `_mongoc_unicode_codepoint_length`
- Tests added to `test_mongoc_scram_sasl_prep`
- `test_mongoc_utf8_char_length`
- `test_mongoc_utf8_string_length`
- `test_mongoc_utf8_to_unicode`


kevinAlbs (commented):
I suggest splitting the bundling of `utf8proc` into a separate PR. That may help keep the scope of the review smaller and to get passing on CI.

I have an open question about the bundling of `utf8proc` and I want input from others on the team. I can post this open question on the separate PR:
- If `utf8proc` is bundled, is an option needed to use a system install of `utf8proc`? 
  - Would Debian and other package managers require having an option to use a system install of `utf8proc` if available?
  - IMO `utf8proc` is a small library. I prefer not having another option between using a system install or the bundled if it is not needed.

src/libmongoc/src/mongoc/mongoc-scram-private.h
```diff
@@ -123,6 +123,346 @@ _mongoc_sasl_prep_required (const char *str);
 char *
 _mongoc_sasl_prep (const char *in_utf8, int in_utf8_len, bson_error_t *err);
 
+/* returns how many bytes a UTF-8 character is. */
+int
```
eramongodb, line 127: Use `size_t` to represent lengths, not `int`, for consistency with `strlen()`, `sizeof()`, etc.

src/libmongoc/src/mongoc/mongoc-scram-private.h
```diff
@@ -123,6 +123,346 @@ _mongoc_sasl_prep_required (const char *str);
...
+_mongoc_utf8_string_length (const char *s);
+
+/* returns whether a UTF-8 character is valid or not. */
+bool
+_mongoc_utf8_is_valid (const char *c, int length);
+
+/* returns whether a character is between two limits (inclusive). */
+bool
+_mongoc_char_between_chars (const char c, const char lower, const char upper);
```
eramongodb, line 140: Use `uint8_t` to represent code units, even if the UTF-8 string is represented as `char*`.

src/libmongoc/src/mongoc/mongoc-scram-private.h
```diff
@@ -123,6 +123,346 @@ _mongoc_sasl_prep_required (const char *str);
...
+/* returns whether a codepoint exists in the specified table. The table format
+ * is that the 2*n element is the lower bound and the 2*n + 1 is the upper bound
+ * (both inclusive). */
+bool
+_mongoc_is_code_in_table (unsigned int code,
+                          const unsigned int *table,
+                          int size);
+
+/* converts a UTF-8 character to unicode code point. */
+unsigned int
```
eramongodb, line 151: Use `uint32_t` to represent code points.

src/libmongoc/src/mongoc/mongoc-scram-private.h
```diff
@@ -123,6 +123,346 @@ _mongoc_sasl_prep_required (const char *str);
...
+_mongoc_utf8_to_unicode (const char *c, int length);
```
eramongodb, line 152: Suggest renaming function to `_mongoc_utf8_get_first_codepoint` and updating the description accordingly.

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1052,81 +1052,160 @@ _mongoc_sasl_prep_impl (const char *name,
...
-      out_utf8, out_utf8_len + 1, NULL, out_utf16, out_utf16_len, &error_code);
-   if (error_code) {
-      bson_free (out_utf8);
-      bson_free (out_utf16);
-      SASL_PREP_ERR_RETURN ("could not convert %s back to UTF-8");
+   /* convert to unicode. */
+   utf8_codepoints = bson_malloc (sizeof (unsigned int) *
+                                  (num_chars + 1)); /* add one for null byte. */
+   c = in_utf8;
+   for (i = 0; i < num_chars; ++i) {
```
eramongodb, line 1066:
```suggestion
   for (size_t i = 0; i < num_chars; ++i) {
```
C99 permits mixing declarations and code. Recommend reducing scope of local variables.

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1150,4 +1229,176 @@ _mongoc_sasl_prep (const char *in_utf8, int in_utf8_len, bson_error_t *err)
    return bson_strdup (in_utf8);
 #endif
 }
+
+int
+_mongoc_utf8_char_length (const char *c)
+{
```
eramongodb, line 1235:
```suggestion
{
   BSON_ASSERT_PARAM (c);

```
Assert parameters that must not be NULL.

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1150,4 +1229,176 @@ _mongoc_sasl_prep (const char *in_utf8, int in_utf8_len, bson_error_t *err)
...
+   int length;
+
+   // UTF-8 characters are either 1, 2, 3, or 4 bytes and the character length
+   // can be determined by the first byte
+   if ((*c & 0x80) == 0)
```
eramongodb, line 1240:
```suggestion
   const uint8_t unit = (uint8_t) *c;

   // UTF-8 characters are either 1, 2, 3, or 4 bytes and the character length
   // can be determined by the first byte
   if ((unit & UINT8_C (0x80)) == 0) {
      return 1u;
   }
```
Use `uint8_t` to represent code units. Cast `char -> uint8_t` before operating on code units, and use `UINT8_C ()` to ensure correct and consistent suffixes are used for integer literals.

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1150,4 +1229,176 @@ _mongoc_sasl_prep (const char *in_utf8, int in_utf8_len, bson_error_t *err)
...
+   else if ((*c & 0xe0) == 0xc0)
+      length = 2;
+   else if ((*c & 0xf0) == 0xe0)
+      length = 3;
+   else if ((*c & 0xf8) == 0xf0)
+      length = 4;
+   else
+      length = 1;
+
+   return length;
```
eramongodb, line 1251: Suggest using early returns instead of assignment:
```c
if (...) {
   return 1u;
}

if (...) {
   return 2u;
}

...

return 1u;
```

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1150,4 +1229,176 @@ _mongoc_sasl_prep (const char *in_utf8, int in_utf8_len, bson_error_t *err)
...
+int
+_mongoc_utf8_string_length (const char *s)
+{
+   const char *c = s;
+
+   int str_length = 0;
+   int utf8_char_length;
+
+   while (*c) {
+      utf8_char_length = _mongoc_utf8_char_length (c);
```
eramongodb, line 1263:
```suggestion
   while (*c) {
      const size_t utf8_char_length = _mongoc_utf8_char_length (c);
```
C99 permits mixing declarations and code. Recommend reducing scope of local variables.

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1150,4 +1229,176 @@ _mongoc_sasl_prep (const char *in_utf8, int in_utf8_len, bson_error_t *err)
...
+      c += utf8_char_length;
+   }
+
+   return str_length;
+}
+
+
+bool
+_mongoc_utf8_is_valid (const char *c, int length)
+{
```
eramongodb, line 1278:
```suggestion
{
   BSON_ASSERT_PARAM (c);

```
Suggest asserting `BSON_ASSERT_PARAM (c);`. It does not make sense to request the length of a not-string (+ `strlen()` consistency).

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1150,4 +1229,176 @@ _mongoc_sasl_prep (const char *in_utf8, int in_utf8_len, bson_error_t *err)
...
+              _mongoc_char_between_chars (c[2], 0x80, 0xBF) &&
+              _mongoc_char_between_chars (c[3], 0x80, 0xBF));
+   default:
+      return true;
+   }
+}
+
+
+bool
+_mongoc_char_between_chars (const char c, const char lower, const char upper)
```
eramongodb, line 1322: Suggest renaming `char` to `code_unit`.

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1150,4 +1229,176 @@ _mongoc_sasl_prep (const char *in_utf8, int in_utf8_len, bson_error_t *err)
...
+{
+   return (c >= lower && c <= upper);
+}
+
+bool
+_mongoc_is_code_in_table (unsigned int code,
```
eramongodb, line 1328: Suggest being specific and state `code_point` instead of simply `code` (ambiguous).

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1150,4 +1229,176 @@ _mongoc_sasl_prep (const char *in_utf8, int in_utf8_len, bson_error_t *err)
...
+      return 0;
+   }
+}
+
+int
+_mongoc_unicode_to_utf8 (unsigned int c, char *out)
+{
+   if (c <= 0x7F) {
+      // Plain ASCII
+      out[0] = (char) c;
```
eramongodb, line 1365: Rather than cast to `char`, suggest casting `char *out -> uint8_t *ptr` and directly assign `uint8_t` code units to `*ptr` instead:
```c
uint8_t *ptr = (uint8_t *) out;

if (c <= UINT32_C (0x7F)) {
   ptr[0] = (uint8_t) c;
}
```

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1150,4 +1229,176 @@ _mongoc_sasl_prep (const char *in_utf8, int in_utf8_len, bson_error_t *err)
...
+      return 1;
+   } else if (c <= 0x07FF) {
+      // 2-byte unicode
+      out[0] = (char) (((c >> 6) & 0x1F) | 0xC0);
```
eramongodb, line 1369: Recommend enabling `-Wconversion` and `-Wsign-conversion` warnings to validate type correctness in bitwise operations on integers with size less than `int`. e.g. this line of code (assuming `uint8_t` consistency as suggested above) may need to look like this:
```c
if (c <= 0x07FFu) {
   ptr[0] = (uint8_t) (((c >> 6) & UINT8_C (0x1F)) | UINT8_C (0xC0));
}
```

src/libmongoc/tests/test-mongoc-scram.c
```diff
@@ -669,6 +685,49 @@ test_mongoc_saslprep_auth_no_icu (void *ctx)
...
 
+static void
+test_mongoc_utf8_char_length (void)
+{
+   const char *c0 = ",";
+   const char *c1 = "ɶ";
+   const char *c2 = "ྡྷ";
+   const char *c3 = "🌂";
+
+   ASSERT_CMPINT (_mongoc_utf8_char_length (c0), ==, 1);
```
eramongodb, line 696:
```suggestion
   ASSERT_CMPSIZE_T (_mongoc_utf8_char_length (c0), ==, 1u);
```
Use `size_t` to represent lengths.

src/libmongoc/tests/test-mongoc-scram.c
```diff
@@ -669,6 +685,49 @@ test_mongoc_saslprep_auth_no_icu (void *ctx)
...
+
+static void
+test_mongoc_utf8_string_length (void)
+{
+   const char *s0 = ",ase";
+   const char *s1 = "ɸɴ";
+   const char *s2 = "ྡྷ🌂e4🌕";
+   const char *s3 = "no special characters";
+
+   ASSERT_CMPINT (_mongoc_utf8_string_length (s0), ==, 4);
```
eramongodb, line 710:
```suggestion
   ASSERT_CMPINT (_mongoc_utf8_string_length (",ase"), ==, 4);
```
Local variables may not be necessary for simple, single-use cases like this.

eramongodb (changes requested):

src/libmongoc/src/mongoc/mongoc-scram.c
joshbsiegel: I removed all of the variable declarations from the top of the function and only kept the variables that I determined weren't "local" to a part of the function. I recognize how that is subjective though so let me know if I should remove more.

src/libmongoc/src/mongoc/mongoc-scram-private.h
joshbsiegel: I ended up changing every instance of `char` (UTF-8 input or output code unit) into `uint8_t`, every instance of `unsigned int` (Unicode code point) into `uint32_t`, and every instance of a measure of `int` or another measure of length into `size_t`. This is true for everything besides the functions that already existed prior to the changes in this PR (e.g. `_mongoc_sasl_prep` and `_mongoc_sasl_prep_impl`), as I didn't want to change the interfaces that were already established.

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1035,12 +1036,12 @@ _mongoc_sasl_prep_impl (const char *name,
                         int in_utf8_len,
```
eramongodb, line 1036: This parameter is now unused: should `_mongoc_utf8_string_length` account for the possibility of `in_utf8_len < strlen(in_utf8)`?

src/libmongoc/src/mongoc/mongoc-scram-private.h
```diff
@@ -123,6 +123,348 @@ _mongoc_sasl_prep_required (const char *str);
...
+size_t
+_mongoc_utf8_string_length (const uint8_t *s);
+
+/* returns whether a UTF-8 character is valid or not. */
+bool
+_mongoc_utf8_is_valid (const uint8_t *c, size_t length);
+
+/* returns whether a character is between two limits (inclusive). */
+bool
+_mongoc_code_unit_between_code_unit (const uint8_t c,
```
eramongodb, line 140:
```suggestion
_mongoc_utf8_code_unit_in_range (const uint8_t c,
```
Suggest renaming for consistency with other UTF-8 functions.

src/libmongoc/src/mongoc/mongoc-scram-private.h
```diff
@@ -123,6 +123,348 @@ _mongoc_sasl_prep_required (const char *str);
...
+                                     const uint8_t lower,
+                                     const uint8_t upper);
+
+/* returns whether a codepoint exists in the specified table. The table format
+ * is that the 2*n element is the lower bound and the 2*n + 1 is the upper bound
+ * (both inclusive). */
+bool
+_mongoc_is_code_point_in_table (uint32_t code,
```
eramongodb, line 148:
```suggestion
_mongoc_utf8_code_point_is_in_table (uint32_t code,
```
Suggest renaming for consistency with other UTF-8 functions.

src/libmongoc/src/mongoc/mongoc-scram-private.h
```diff
@@ -123,6 +123,348 @@ _mongoc_sasl_prep_required (const char *str);
...
+                                const uint32_t *table,
+                                size_t size);
+
+/* returns the first Unicode codepoint of a UTF-8 character. */
+uint32_t
+_mongoc_utf8_get_first_code_point (const uint8_t *c, size_t length);
+
+/* returns how many bytes a unicode codepoint is. */
+size_t
+_mongoc_unicode_codepoint_length (uint32_t c);
```
eramongodb, line 158:
```suggestion
_mongoc_utf8_code_point_length (uint32_t c);
```
Suggest renaming for consistency with other UTF-8 functions.

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1052,82 +1053,165 @@ _mongoc_sasl_prep_impl (const char *name,
...
+   for (size_t i = 0; i < num_chars; ++i) {
+      utf8_pre_norm_len +=
+         _mongoc_unicode_codepoint_length (utf8_codepoints[i]);
+   }
+   uint8_t *utf8_pre_norm =
+      (uint8_t *) bson_malloc (sizeof (uint8_t) * (utf8_pre_norm_len + 1));
+
+   uint8_t *loc = utf8_pre_norm;
+   for (size_t i = 0; i < num_chars; ++i) {
+      utf8_char_length = _mongoc_unicode_to_utf8 (utf8_codepoints[i], loc);
```
eramongodb, line 1126: Missing validity check.

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1052,82 +1053,165 @@ _mongoc_sasl_prep_impl (const char *name,
...
+   // this is an optional step for stringprep, but Unicode normalization with
+   // form KC is required for SASLPrep.
+
+   // in order to do this, we must first convert back to UTF-8.
+
+   // preflight for length
+   size_t utf8_pre_norm_len = 0;
+   for (size_t i = 0; i < num_chars; ++i) {
+      utf8_pre_norm_len +=
+         _mongoc_unicode_codepoint_length (utf8_codepoints[i]);
```
eramongodb, line 1119: Missing validity check.

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1052,82 +1053,165 @@ _mongoc_sasl_prep_impl (const char *name,
...
-   (void) u_strToUTF8 (
-      out_utf8, out_utf8_len + 1, NULL, out_utf16, out_utf16_len, &error_code);
-   if (error_code) {
+   /* convert to unicode. */
+   utf8_codepoints = bson_malloc (sizeof (uint32_t) *
+                                  (num_chars + 1)); /* add one for null byte. */
+   const uint8_t *c = (uint8_t *) in_utf8;
+
+   for (size_t i = 0; i < num_chars; ++i) {
+      utf8_char_length = _mongoc_utf8_char_length (c);
```
eramongodb, line 1069:
```suggestion
      const size_t utf8_char_length = _mongoc_utf8_char_length (c);
```
The `utf8_char_length` variable can be declared in each scope it is locally used in.

+ missing validity check.

joshbsiegel: `_mongoc_utf8_char_length` doesn't return -1 on error, so I did not add a validity check.

src/libmongoc/src/mongoc/mongoc-scram.c
joshbsiegel: I'm a little unsure on what you mean by this. In all cases of `_mongoc_sasl_prep_impl` being called, `in_utf8_len` is passed through as `strlen(in_utf8)`, meaning that `in_utf8_len == strlen(in_utf8)`.

eramongodb: Good point. I suggest removing the redundant/unused parameter from `_mongoc_sasl_prep` and `_mongoc_sasl_prep_impl` to address the unused parameter warning(s).

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1150,4 +1241,222 @@ _mongoc_sasl_prep (const char *in_utf8, int in_utf8_len, bson_error_t *err)
...
+}
+
+ssize_t
+_mongoc_utf8_string_length (const char *s)
+{
+   BSON_ASSERT_PARAM (s);
+
+   const uint8_t *c = (uint8_t *) s;
+
+   size_t str_length = 0;
```
eramongodb, line 1272:
```suggestion
   ssize_t str_length = 0;
```
Consistent signedness with return value.

src/libmongoc/tests/test-mongoc-scram.c
```diff
@@ -130,6 +145,35 @@ test_mongoc_scram_sasl_prep (void)
...
+   ASSERT_CMPINT (_mongoc_utf8_char_length (","), ==, 1u);
+   ASSERT_CMPINT (_mongoc_utf8_char_length ("ɶ"), ==, 2u);
+   ASSERT_CMPINT (_mongoc_utf8_char_length ("ྡྷ"), ==, 3u);
+   ASSERT_CMPINT (_mongoc_utf8_char_length ("🌂"), ==, 4u);
+}
+
+static void
+test_mongoc_utf8_string_length (void)
+{
+   ASSERT_CMPINT (_mongoc_utf8_string_length (",ase"), ==, 4u);
```
eramongodb, line 161:
```suggestion
   ASSERT_CMPSIZE_T (_mongoc_utf8_string_length (",ase"), ==, 4u);
```
Use `ASSERT_CMPSIZE_T` to compare `size_t`.

src/libmongoc/tests/test-mongoc-scram.c
```diff
@@ -130,6 +145,35 @@ test_mongoc_scram_sasl_prep (void)
...
+   ASSERT_CMPINT (_mongoc_utf8_string_length ("ɸɴ"), ==, 2u);
+   ASSERT_CMPINT (_mongoc_utf8_string_length ("ྡྷ🌂e4🌕"), ==, 5u);
+   ASSERT_CMPINT (
+      _mongoc_utf8_string_length ("no special characters"), ==, 21u);
+}
+
+static void
+test_mongoc_utf8_to_unicode (void)
+{
+   ASSERT_CMPINT (_mongoc_utf8_get_first_code_point (",", 1), ==, 0x002C);
```
eramongodb, line 171:
```suggestion
   ASSERT_CMPUINT32 (_mongoc_utf8_get_first_code_point (",", 1), ==, 0x002C);
```
Use `ASSERT_CMPUINT32` to compare `uint32_t`.

src/libmongoc/src/mongoc/mongoc-scram-private.h
```diff
@@ -121,7 +121,352 @@ _mongoc_sasl_prep_required (const char *str);
...
  * `name` should be "username" or "password". */
 char *
-_mongoc_sasl_prep (const char *in_utf8, int in_utf8_len, bson_error_t *err);
+_mongoc_sasl_prep (const char *in_utf8, bson_error_t *err);
+
+/* returns how many bytes a UTF-8 character is. */
+size_t
+_mongoc_utf8_char_length (const char *c);
+
+/* returns how many characters are in a UTF-8 string. Returns -1 on error. */
```
eramongodb, line 130:
```suggestion
/* Returns the byte length of the UTF-8 string. Returns -1 if `s` is not a valid
 * UTF-8 string. */
```

src/libmongoc/src/mongoc/mongoc-scram-private.h
```diff
@@ -121,7 +121,352 @@ _mongoc_sasl_prep_required (const char *str);
  * null on error and sets err.
  * `name` should be "username" or "password". */
 char *
-_mongoc_sasl_prep (const char *in_utf8, int in_utf8_len, bson_error_t *err);
+_mongoc_sasl_prep (const char *in_utf8, bson_error_t *err);
+
+/* returns how many bytes a UTF-8 character is. */
```
eramongodb, line 126:
```suggestion
/* Returns the byte length of the first UTF-8 code point in `s`. */
```

src/libmongoc/src/mongoc/mongoc-scram-private.h
```diff
@@ -121,7 +121,352 @@ _mongoc_sasl_prep_required (const char *str);
...
+size_t
+_mongoc_utf8_char_length (const char *c);
+
+/* returns how many characters are in a UTF-8 string. Returns -1 on error. */
+ssize_t
+_mongoc_utf8_string_length (const char *s);
+
+/* returns whether a UTF-8 character is valid or not. */
```
eramongodb, line 134:
```suggestion
/* Returns true if the first UTF-8 code point in `s` is valid. */
```

src/libmongoc/src/mongoc/mongoc-scram-private.h
```diff
@@ -121,7 +121,352 @@ _mongoc_sasl_prep_required (const char *str);
...
+
+/* returns whether a codepoint exists in the specified table. The table format
+ * is that the 2*n element is the lower bound and the 2*n + 1 is the upper bound
+ * (both inclusive). */
+bool
+_mongoc_utf8_code_point_is_in_table (uint32_t code,
+                                     const uint32_t *table,
+                                     size_t size);
+
+/* returns the first Unicode codepoint of a UTF-8 character. */
```
eramongodb, line 152:
```suggestion
/* Returns the first Unicode code point in `c`. Returns 0 if length is 0.
 * `c` must be a valid UTF-8 string. */
```

src/libmongoc/src/mongoc/mongoc-scram-private.h
```diff
@@ -121,7 +121,352 @@ _mongoc_sasl_prep_required (const char *str);
...
+uint32_t
+_mongoc_utf8_get_first_code_point (const char *c, size_t length);
+
+/* returns how many bytes a unicode codepoint is. Returns -1 on invalid code
+ * point. */
```
eramongodb, line 157:
```suggestion
/* Returns the byte length of the UTF-8 code point. Returns -1 if `c` is not a
 * valid UTF-8 code point. */
```

eramongodb (approved):

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1052,93 +1051,185 @@ _mongoc_sasl_prep_impl (const char *name,
...
 
 char *
-_mongoc_sasl_prep (const char *in_utf8, int in_utf8_len, bson_error_t *err)
+_mongoc_sasl_prep (const char *in_utf8, bson_error_t *err)
 {
-   BSON_UNUSED (in_utf8_len);
-
 #ifdef MONGOC_ENABLE_ICU
-   return _mongoc_sasl_prep_impl ("password", in_utf8, in_utf8_len, err);
+   return _mongoc_sasl_prep_impl ("password", in_utf8, err);
```
kevinAlbs, line 1232:
```suggestion
   if (_mongoc_sasl_prep_required (in_utf8)) {
      return _mongoc_sasl_prep_impl ("password", in_utf8, err);
   }
   return bson_strdup (in_utf8);
```

Suggest only running `_mongoc_sasl_prep_impl` if a password includes non-ASCII characters. I expect the majority of passwords are ASCII only characters and do not need to undergo SASLPrep.

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1052,93 +1051,185 @@ _mongoc_sasl_prep_impl (const char *name,
...
+                     sizeof (uint32_t))) {
+         // effectively skip over the character because we don't increment curr.
+      } else
+         utf8_codepoints[curr++] = utf8_codepoints[i];
+   }
+   utf8_codepoints[curr] = '\0';
+   num_chars = curr;
+
+
+   // b. Normalize - normalize the result of step 1 using Unicode
```
kevinAlbs, line 1105:
```suggestion
   // b. Normalize - normalize the result of step `a` using Unicode
```

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1052,93 +1051,185 @@ _mongoc_sasl_prep_impl (const char *name,
...
+
+
+   for (size_t i = 0; i < num_chars; ++i) {
+      if (_mongoc_utf8_code_point_is_in_table (utf8_codepoints[i],
+                                               LCat_bidi_ranges,
+                                               sizeof (LCat_bidi_ranges) /
+                                                  sizeof (uint32_t))) {
+         contains_LCat = true;
+         if (contains_RandALCar)
+            break;
```
kevinAlbs, line 1192:
```suggestion
```

Suggest removing unnecessary early break for consistency with assignment of `contains_RandALCar = true;`

kevinAlbs (approved):
Great work. LGTM with minor comments addressed.

PR #1319, opened 2023-06-22
CDRIVER-4668 Add utf8proc library and update CMake configuration

joshbsiegel:
# Summary
This ticket is necessary before https://github.com/mongodb/mongo-c-driver/pull/1317 can be reviewed. The [utf8proc](https://juliastrings.github.io/utf8proc/) library is used for Unicode Normalization Form KC which is required by [SASLprep](https://datatracker.ietf.org/doc/html/rfc4013).

# What's New
- Added utf8proc library

Note: Apologies for all of the commits going back and forth on changes. There were multiple times where I thought something was working locally and it turns out it wasn't.


src/libmongoc/CMakeLists.txt
```diff
@@ -108,6 +108,33 @@ else ()
...
+   message (STATUS "Enabling utf8proc compression (bundled)")
+   add_library (utf8proc_obj OBJECT "${UTF8PROC_SOURCES}")
+   set_property (TARGET utf8proc_obj PROPERTY POSITION_INDEPENDENT_CODE TRUE)
+   # This tells the bundled utf8proc where to find its generated headers
+   target_include_directories (utf8proc_obj
+      BEFORE PUBLIC
+      "${SOURCE_DIR}/src/utf8proc-2.8.0"
+      "${CMAKE_BINARY_DIR}/src/utf8proc-2.8.0"
+   )
+   set (SOURCES ${SOURCES} $<TARGET_OBJECTS:utf8proc_obj>)
```
kevinAlbs, line 129:
```suggestion
   set (SOURCES ${SOURCES} $<TARGET_OBJECTS:utf8proc_obj>)
   target_compile_definitions (utf8proc_obj PUBLIC UTF8PROC_STATIC)
```

Windows tasks are failing with [this log](https://parsley.mongodb.com/evergreen/mongo_c_driver_windows_2017_32_debug_compile_nosasl_nossl_patch_7d5beeae25e3c821b6e01e204ba6d323184c8b7d_6494b5b93627e04d43aaaff4_23_06_22_20_57_31/0/task?bookmarks=0,220&shareLine=140):

```
C:\data\mci\f6caccb9ade092c41e4af9bff293259e\mongoc\src\utf8proc-2.8.0\utf8proc.c(56): error C2491: 'utf8proc_utf8class': definition of dllimport data not allowed
```

I expect the dllimport attribute needs to be removed. dllimport is used by consumers to declare that symbols are defined in another DLL. The utf8proc source files are compiled directly in the C driver (not as a separate DLL). 

https://github.com/JuliaStrings/utf8proc/blob/1cb28a66ca79a0845e99433fd1056257456cef8b/utf8proc.h#L123-L125 suggests defining `UTF8PROC_STATIC` will remove the dllimport attribute.

If this does not resolve the Windows tasks, try spawning a Windows host in Evergreen and building. That may help to more quickly diagnose the issue.

src/CMakeLists.txt
```diff
@@ -18,6 +18,11 @@ set (src_zlib_DIST
    src/zlib-1.2.13/zlib.h
    src/zlib-1.2.13/zutil.h
 )
+
+set (src_utf8proc_DIST
+   src/utf8proc-2.8.0/utf8proc.h
+   src/utf8proc-2.8.0/utf8proc.c
```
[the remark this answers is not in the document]
joshbsiegel: By this do you mean delete all of the other files in the directory?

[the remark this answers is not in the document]
kevinAlbs: Yes.

kevinAlbs, line 24:
```suggestion
   src/utf8proc-2.8.0/utf8proc.c
   src/utf8proc-2.8.0/utf8proc_data.c
```

This may resolve the error in the [make_release_archive](https://parsley.mongodb.com/evergreen/mongo_c_driver_releng_make_release_archive_patch_7d5beeae25e3c821b6e01e204ba6d323184c8b7d_64998425306615e8c37fede6_23_06_26_12_27_19/0/task?bookmarks=0,531&shareLine=385):

```
/Users/kevin.albertson/review/mongo-c-driver-1319/cmake-build/mongo-c-driver-1.24.0-pre/src/utf8proc-2.8.0/utf8proc.c:53:10: fatal error: 'utf8proc_data.c' file not found
#include "utf8proc_data.c"
         ^~~~~~~~~~~~~~~~~
1 error generated.
```

The utf8proc_data.c file needs to be included in the distribution tarball.

This error can be reproduced locally by running the `distcheck` target:
```
cmake --build cmake-build --target distcheck -- -j16
```
The `distcheck` target builds the distribution tarball, extracts it, then rebuilds the driver.

CMakeLists.txt
```diff
@@ -85,6 +85,9 @@ mongo_setting(ENABLE_ZSTD "Enable zstd compression support"
 mongo_setting(ENABLE_ZLIB "Enable zlib compression support"
               OPTIONS BUNDLED SYSTEM OFF
               DEFAULT VALUE BUNDLED)
+mongo_setting(ENABLE_UTF8PROC "Enable building with utf8proc. Needed for SCRAM-SHA-256 authentication with non-ASCII passwords"
+              OPTIONS BUNDLED SYSTEM
+              DEFAULT VALUE BUNDLED)
```
kevinAlbs, line 90: These may be questions for @rcsanchez97 and @vector-of-bool.

This PR bundles `utf8proc` to help remove the dependency on ICU.

- If `utf8proc` is bundled, is an option needed to use a system install of `utf8proc`?
- Would Debian packaging require an option to use a system install of `utf8proc`?

IMO `utf8proc` is a small library. I prefer not having another option between using a system install or the bundled if it is not needed. `utf8proc` appears to be packaged on many systems (https://repology.org/project/utf8proc/versions).

Using a `SYSTEM` install of IntelDFP in libmongocrypt was added (required?) for Debian packaging in MONGOCRYPT-517. If supporting a system install of `utf8proc` is necessary, more work may be needed in this PR to support the `SYSTEM` value.

src/libmongoc/CMakeLists.txt
```diff
@@ -107,6 +107,15 @@ else ()
    message (STATUS "Disabling zlib compression")
 endif ()
 
+set (UTF8PROC_INCLUDE_DIRS "")
```
kevinAlbs, line 110:
```suggestion
```

`UTF8PROC_INCLUDE_DIRS` is unused.

src/libmongoc/CMakeLists.txt
```diff
@@ -705,11 +714,11 @@ endif ()
 
 set (LIBRARIES
    ${SASL_LIBRARIES} ${SSL_LIBRARIES} ${SHM_LIBRARIES} ${RESOLV_LIBRARIES}
-   ${SNAPPY_LIBRARIES} ${ZLIB_LIBRARIES} ${MONGOC_ZSTD_LIBRARIES} Threads::Threads ${ICU_LIBRARIES} ${LIBMONGOCRYPT_LIBRARY}
+   ${SNAPPY_LIBRARIES} ${ZLIB_LIBRARIES} ${UTF8PROC_LIBRARIES} ${MONGOC_ZSTD_LIBRARIES} Threads::Threads ${ICU_LIBRARIES} ${LIBMONGOCRYPT_LIBRARY}
```
kevinAlbs, line 717:
```suggestion
   ${SNAPPY_LIBRARIES} ${ZLIB_LIBRARIES} ${MONGOC_ZSTD_LIBRARIES} Threads::Threads ${ICU_LIBRARIES} ${LIBMONGOCRYPT_LIBRARY}
```

Only bundling `utf8proc` is currently supported in this PR (not using a system install of `utf8proc`).
`${UTF8PROC_LIBRARIES} ` and `${UTF8PROC_INCLUDE_DIRS}` are always empty. Suggest removing all references to both `${UTF8PROC_LIBRARIES} ` and `${UTF8PROC_INCLUDE_DIRS}` in this file.

CMakeLists.txt
rcsanchez97: So, the short answer is "yes, Debian (and Fedora) will require the ability to build against a system-provided utf8proc."

Additionally, from a Debian packaging perspective, we will want to filter out the embedded code (as we do for zlib, which can be seen in `debian/gdb.conf`). In essence, embedding code copies is very strongly discouraged in Debian (Fedora discourages it as well, but they aren't quite as stringent about it). We would have to have a very good reason for the archive maintainers to OK the introduction of an embedded code copy. We probably don't have a very good reason here, since the utf8proc package is available at the current latest upstream version.

The Debian-specific parts of this could be a little tricky. So, my recommendation is that Josh handle implementing the ability to specify a system install of utf8proc to the CMake configuration, and then I can handle updating the Debian packaging parts in a follow-up PR.

vector-of-bool: I would recommend changing the option to be a boolean `USE_BUNDLED_UTF8PROC` (default `TRUE`), and mark it as `ADVANCED`. Because utf8proc does not (appear to) ship a CMake config-module, we can add our own find-module `Findutf8proc.cmake` that imports the external library. We will also require modification to the `mongoc-1.0-config.cmake` to also handle the case where mongoc was built with `USE_BUNDLED_UTF8PROC`.

build/cmake/Findutf8proc.cmake
rcsanchez97, line 1: In keeping with the CMake naming conventions, this should probably be named `FindUtf8proc.cmake` (with a capital 'U').

CMakeLists.txt
```diff
@@ -85,6 +85,10 @@ mongo_setting(ENABLE_ZSTD "Enable zstd compression support"
 mongo_setting(ENABLE_ZLIB "Enable zlib compression support"
               OPTIONS BUNDLED SYSTEM OFF
               DEFAULT VALUE BUNDLED)
+mongo_setting(USE_BUNDLED_UTF8PROC "Enable building with utf8proc. Needed for SCRAM-SHA-256 authentication with non-ASCII passwords"
+              OPTIONS TRUE FALSE
+              DEFAULT VALUE TRUE
+              ADVANCED)
```
joshbsiegel, line 91: To test the system install, I ran `sudo apt-get install libutf8proc-dev` (`brew install utf8proc` on Mac), manually set `USE_BUNDLED_UTF8PROC` to `FALSE`, and included the library in a random file to make sure that CMake was able to configure and build correctly.

rcsanchez97 (approved):
LGTM

build/cmake/FindUtf8Proc.cmake
```diff
@@ -0,0 +1,22 @@
...
+ message (STATUS "Enabling utf8proc")
+  find_package(PkgConfig)
+  pkg_check_modules(PC_UTF8PROC REQUIRED libutf8proc)
+  add_library(utf8proc STATIC IMPORTED)
+  set_target_properties(utf8proc 
+                        PROPERTIES
+                        COMPILE_OPTIONS ${PC_UTF8PROC_STATIC_CFLAGS} 
+                        LINK_LIBRARIES ${PC_UTF8PROC_LINK_LIBRARIES} 
+                        INCLUDE_DIRECTORIES ${PC_UTF8PROC_STATIC_LIBRARY_DIRS} 
+                        IMPORTED_LOCATION ${PC_UTF8PROC_LINK_LIBRARIES})
```
joshbsiegel, line 17: `IMPORTED_LOCATION` feels redundant but I got an error when I didn't include it.

build/cmake/FindUtf8Proc.cmake
```diff
@@ -2,21 +2,18 @@ if(USE_BUNDLED_UTF8PROC STREQUAL "TRUE")
...
   find_package(PkgConfig)
   pkg_check_modules(PC_UTF8PROC REQUIRED libutf8proc)
   add_library(utf8proc STATIC IMPORTED)
-  set_target_properties(utf8proc 
-                        PROPERTIES
-                        COMPILE_OPTIONS ${PC_UTF8PROC_STATIC_CFLAGS} 
-                        LINK_LIBRARIES ${PC_UTF8PROC_LINK_LIBRARIES} 
-                        INCLUDE_DIRECTORIES ${PC_UTF8PROC_STATIC_LIBRARY_DIRS} 
-                        IMPORTED_LOCATION ${PC_UTF8PROC_LINK_LIBRARIES})
+  set_property(TARGET utf8proc PROPERTY IMPORTED_LOCATION "${PC_UTF8PROC_LINK_LIBRARIES}")
```
eramongodb, line 11: On further inspection, this is incorrectly linking against the _shared_ utf8proc library, and `${PC_UTF8PROC_STATIC_LINK_LIBRARIES}` does not seem to be set when testing locally (even though `${PC_UTF8PROC_STATIC_LIBRARY_DIRS}` is set...?).

I think `find_library()` may be required to set `IMPORTED_LIBRARY` correctly:

```cmake
pkg_check_modules(PC_UTF8PROC REQUIRED libutf8proc)

find_library(
   UTF8PROC_STATIC_LIBRARY libutf8proc.a
   PATHS "${PC_UTF8PROC_STATIC_LIBRARY_DIRS}"
)

add_library(utf8proc STATIC IMPORTED)

set_property(TARGET utf8proc PROPERTY IMPORTED_LOCATION "${UTF8PROC_STATIC_LIBRARY}")
```

I think we can get away with just `libutf8proc.a` (Unix-like), but if we also want to support using system utf8proc on Windows, I think we need to add `libutf8proc.lib` to the list.

CC @vector-of-bool, mind double-checking this logic?

vector-of-bool: That logic should work, although I would recommend `HINTS` instead of `PATHS`.

---

However, this case is mainly to satisfy requirements of system distributors, and it is likely that they would want to use the dynamic library.

I recommend using the `IMPORTED_TARGET` argument to `pkg_check_modules()` instead. It will handle generating the imported target and setting all of these properties. You can then define an `ALIAS` library to the pkg-config library, as an additional level of indirection may be required so that the rest of the build is agnostic to `USE_BUNFLED_UTF8PROC`.

This is an unfortunately subtle and non-obvious first experience with CMake 🙂.

src/libmongoc/CMakeLists.txt
```diff
@@ -708,6 +707,15 @@ set (STATIC_LIBRARIES
    ${SNAPPY_LIBRARIES} ${ZLIB_LIBRARIES} ${MONGOC_ZSTD_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT} ${ICU_LIBRARIES} ${LIBMONGOCRYPT_LIBRARY}
 )
 
+# utf8proc configuration
+find_package(Utf8Proc REQUIRED)
+if ("${USE_BUNDLED_UTF8PROC}" STREQUAL "TRUE")
+   list(APPEND SOURCES $<TARGET_OBJECTS:utf8proc_obj>)
```
vector-of-bool, line 713: Propagating object files into the mongoc static libraries may wreak havoc if users wish to use a different utf8proc in their application. This could be mitigated with linker tricks, but I think a cleaner CMake-based approach may be possible and simpler to maintain. I would not consider this a blocker for now, and can be postponed.

src/libmongoc/CMakeLists.txt
```diff
@@ -708,6 +707,15 @@ set (STATIC_LIBRARIES
...
+else()
+   list(APPEND LIBRARIES utf8proc)
+   list(APPEND STATIC_LIBRARIES utf8proc)
```
vector-of-bool, line 716: Use the _mongo-dependencies target (defined just below here) to propagate these linker requirements

vector-of-bool (changes requested):
Some comments, and recommended changes.

Unfortunately, a lot of the existing CMake code of bygone years sets a bad precedent and may be misleading on the ideal way to do things.

build/cmake/FindUtf8Proc.cmake
joshbsiegel: The way that I understood this was that @vector-of-bool's suggestion would resolve @eramongodb's concerns, so I didn't add any `find_library` calls. Let me know if that is a misunderstanding.

build/cmake/FindUtf8Proc.cmake
```diff
@@ -0,0 +1,11 @@
+if(USE_BUNDLED_UTF8PROC)
+  message (STATUS "Enabling utf8proc (bundled)")
+  add_library (utf8proc_obj OBJECT "${UTF8PROC_SOURCES}")
+  set_property (TARGET utf8proc_obj PROPERTY POSITION_INDEPENDENT_CODE TRUE)
+  target_compile_definitions (utf8proc_obj PUBLIC UTF8PROC_STATIC)
+else ()
+  message (STATUS "Enabling utf8proc")
```
kevinAlbs, line 7:
```suggestion
  message (STATUS "Searching for utf8proc on system")
```

Suggest updating message since utf8proc may fail to enable if it is not found.

kevinAlbs (approved):
LGTM

src/libmongoc/CMakeLists.txt
```diff
@@ -707,20 +707,19 @@ set (STATIC_LIBRARIES
...
 
+add_library(_mongoc-dependencies INTERFACE)
+add_library(mongo::detail::c_dependencies ALIAS _mongoc-dependencies)
+install(TARGETS _mongoc-dependencies EXPORT mongoc-targets)
+set_property(TARGET _mongoc-dependencies PROPERTY EXPORT_NAME detail::c_dependencies)
+
 # utf8proc configuration
 find_package(Utf8Proc REQUIRED)
-if ("${USE_BUNDLED_UTF8PROC}" STREQUAL "TRUE")
+if (${USE_BUNDLED_UTF8PROC})
```
vector-of-bool, line 717: Best to just use the variable name without a dereference:

```suggestion
if (USE_BUNDLED_UTF8PROC)
```

The syntax of `if()` (and `while()`) has some special non-obvious rules ([Refer](https://cmake.org/cmake/help/latest/command/if.html))

src/libmongoc/CMakeLists.txt
```diff
@@ -713,6 +712,14 @@ add_library(mongo::detail::c_dependencies ALIAS _mongoc-dependencies)
 install(TARGETS _mongoc-dependencies EXPORT mongoc-targets)
 set_property(TARGET _mongoc-dependencies PROPERTY EXPORT_NAME detail::c_dependencies)
 
+# utf8proc configuration
+find_package(Utf8Proc REQUIRED)
+if (${USE_BUNDLED_UTF8PROC})
+   list(APPEND SOURCES $<TARGET_OBJECTS:utf8proc_obj>)
+else()
+   target_link_libraries(_mongoc-dependencies INTERFACE utf8proc)
```
vector-of-bool, line 720: If `USE_BUNDLED_UTF8PROC` is false, then the installed `_mongoc-dependencies` will contain a link reference to `PkgConfig::PC_UTF8PROC`, but that target name will not be defined for importers by default.

The following (or something close to it) will need to be added in `mongoc-config.cmake`:

```cmake
set(_mongoc_built_with_bundled_utf8proc "@USE_BUNDLED_UTF8PROC@")
if(NOT _mongoc_built_with_bundled_utf8proc AND NOT TARGET PkgConfig::PC_UTF8PROC)
  # libmongoc was compiled against an external utf8proc and links against a 
  # FindPkgConfig-genreated IMPORTED target. Find that package and generate that 
  # imported target here:
  find_dependency(PkgConfig)
  pkg_config_check_modules(PC_UTF8PROC REQUIRED utf8proc IMPORTED_TARGET GLOBAL)
endif()
```

vector-of-bool (changes requested):
One final addition is recommended, but looks good so far.

src/libmongoc/CMakeLists.txt
joshbsiegel: I added an `ALIAS` call beneath this to make sure that the generated target can also be recognized by `utf8proc`

vector-of-bool: ALIAS targets are not installed and only create names in the build tree. When installed, all occurrences of the alias will be replaced with the aliased target name. The ALIAS target in mongoc-config.cmake is thus unnecessary.

src/libmongoc/src/mongoc-config.cmake
```diff
@@ -1,3 +1,12 @@
...
 include("${CMAKE_CURRENT_LIST_DIR}/mongoc-targets.cmake")
+
+set(_mongoc_built_with_bundled_utf8proc "@USE_BUNDLED_UTF8PROC@")
+if(NOT _mongoc_built_with_bundled_utf8proc AND NOT TARGET PkgConfig::PC_UTF8PROC)
+  # libmongoc was compiled against an external utf8proc and links against a 
+  # FindPkgConfig-generated IMPORTED target. Find that package and generate that 
+  # imported target here:
+  find_dependency(PkgConfig)
+  pkg_config_check_modules(PC_UTF8PROC REQUIRED utf8proc IMPORTED_TARGET GLOBAL)
+endif()
```
eramongodb, line 12: Missing EOF newline.

build/cmake/FindUtf8Proc.cmake
```diff
@@ -0,0 +1,11 @@
+if(USE_BUNDLED_UTF8PROC)
+  message (STATUS "Enabling utf8proc (bundled)")
+  add_library (utf8proc_obj OBJECT "${UTF8PROC_SOURCES}")
+  set_property (TARGET utf8proc_obj PROPERTY POSITION_INDEPENDENT_CODE TRUE)
+  target_compile_definitions (utf8proc_obj PUBLIC UTF8PROC_STATIC)
+else ()
+  message (STATUS "Searching for utf8proc on system")
+  find_package(PkgConfig)
+  pkg_check_modules(PC_UTF8PROC REQUIRED libutf8proc IMPORTED_TARGET)
+  add_library(utf8proc ALIAS PkgConfig::PC_UTF8PROC)
```
eramongodb, line 10: As written, this still links against the shared library (`PkgConfig::PC_UTF8PROC` -> `libutf8proc.so`) and propogates this requirement to downstream users. `find_library()` may still be required if we want to ensure linkage with the static library `libutf8proc.a` rather than the shared library `libutf8proc.so`. Support for static libraries via `IMPORTED_TARGET` [seems to be in a troublesome state](https://gitlab.kitware.com/cmake/cmake/-/issues/21714).

Note, this issue/complexity is predecated on the static library requirement. Confirming, is static linking actually an essential requirement for `utf8proc` library integration? If we do _not_ require static linking of `utf8proc`, the `IMPORTED_TARGET` solution should suffice as written.

vector-of-bool: Since this codepath is most intended for use by distribution package maintainers, and that they will likely wish to use the dynamic library almost exclusively, I would consider that to be acceptable.

eramongodb (approved):

vector-of-bool:
Failing tasks are related to #1327, which was merged. If you merge `master` into this branch, it should go green.

vector-of-bool (approved):

PR #1325, opened 2023-06-28
remove unnecessary cmake version checks

kevinAlbs:
# Summary
Remove unnecessary cmake version checks.

# Background & Motivation
As of 3165a356473a1d50d1142d7071bd46bb5bb81c2a the minimum required cmake version has increased to 3.15. This PR removes cmake version checks and documentation that assumes older cmake versions could be used.

### DIFF
--- a/build/cmake/make_dist/MakeDistCheck.cmake
+++ b/build/cmake/make_dist/MakeDistCheck.cmake
@@ -17,9 +17,7 @@ function (RUN_DIST_CHECK PACKAGE_PREFIX EXT)
    endif ()
 
    set (MY_CMAKE_COMMAND "")
-   if (${CMAKE_VERSION} VERSION_GREATER 3.1)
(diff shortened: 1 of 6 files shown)


build/cmake/make_dist/MakeDistCheck.cmake
```diff
@@ -17,9 +17,7 @@ function (RUN_DIST_CHECK PACKAGE_PREFIX EXT)
    endif ()
 
    set (MY_CMAKE_COMMAND "")
-   if (${CMAKE_VERSION} VERSION_GREATER 3.1)
-      set (MY_CMAKE_COMMAND ${CMAKE_COMMAND} -E env)
-   endif ()
+   set (MY_CMAKE_COMMAND ${CMAKE_COMMAND} -E env)
```
vector-of-bool, line 20: Line 19 can be deleted.

vector-of-bool (approved):

PR #1326, opened 2023-06-28
Cleanup pkg-config

vector-of-bool:
On my yak-shaving journey to remove the source-dist, I ended up needing to rewrite the generation of pkg-config files (again) (long story). Fortunately, this is much cleaner and more robust than my prior silliness in #1301. The following now Just Works™:

1. If you build with instrumentation (coverage, sanitizers), the installed pkg-config files will now include the necessary link flags as usage requirements. This came up while testing the `link-with` scripts, since my local system would enable sanitizers, but the generated example executable did not have the appropriate link flags when it consumed libmongoc/libbson via pkg-config. This will now work out-of-the-box.
2. The `prefix` value in the installed `.pc` values is now bound at install-time, and not at configure-time. This allows using `cmake --install --prefix=<some-dir>` to do an install into an arbitrary `<some-dir>` which may not be the `CMAKE_INSTALL_PREFIX`. This is a somewhat-obscure feature, but I've found it useful and it was easy enough to make it work.
3. Multi-config generation is now supported. Now you can `cmake --install --config=<config>` to install the named configuration, and the appropriate `.pc` file will be installed. (But multi-config `.pc` install requires each configuration to receive a different `.pc` filename, which is really weird for pkg-config and is not used here.)


build/cmake/GeneratePkgConfig.cmake
```diff
@@ -0,0 +1,298 @@
...
+
+INSTALL [DESTINATION <dir>] [RENAME <filename>]
+    - Generate a rule to install the generated pkg-config file. This is better
+      than using a `file(INSTALL)` on the generated file directly, since it
+      ensures that the installed .pc file will have the correct install prefix
+      value. The following additional arguments are also accepted:
+
+      DESTINATION <dir>
+        - If provided, specify the *directory* (relative to the install-prefix)
+          in which the generated file will be installed. If unspecfied, the default
```
rcsanchez97, line 97:
```suggestion
          in which the generated file will be installed. If unspecified, the default
```

rcsanchez97 (approved):
LGTM, modulo the one typo I found.

build/cmake/ResSearch.cmake
```diff
@@ -1,6 +1,9 @@
 include(CheckSymbolExists)
 include(CMakePushCheckState)
 
+# Required so that some libresolv symbols will be visible:
+set(CMAKE_C_EXTENSIONS ON)
```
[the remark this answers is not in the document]
eramongodb: Can this be scoped or moved into a target property? I have reservations about enabling C extensions for potentially all targets that follow `include(ResSearch)`. If this must be set unconditionally, suggest moving up into the root CMakeLists.txt file alongside other project-wide variables.

src/libbson/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.1)
+cmake_minimum_required (VERSION 3.15)
```
eramongodb, line 1: This is a significant change. I suggest at least creating a CXX ticket to track/document this new requirement.

Side question: is there a compelling reason we should continue to support a per-library CMake Project configuration instead of using only the root CMake project for all targets defined in the CXX Driver? I don't think we intend to support `add_subdirectory()` of a CXX Driver library (i.e. bsoncxx) _without_ going through the root directory, do we?

build/cmake/GeneratePkgConfig.cmake
```diff
@@ -0,0 +1,298 @@
...
+          when built/installed using a multi-config generator!)
+
+CONDITION <cond>
+    - The file will only be generated/installed if the condition `<cond>`
+      results in the string "1" after evaluating generator expressions.
+
+All named parameters accept generator expressions.
+
+]==]
+function(mongo_generate_pkg_config target)
```
eramongodb, line 113: It baffles me that CMake doesn't provide any utilities to support smart target-based pkg-config file generation. 😞

build/cmake/GeneratePkgConfig.cmake
```diff
@@ -0,0 +1,298 @@
...
+    string(APPEND content "$<${has_requires}:${requires_joiner}$<JOIN:${gx_requires},${requires_joiner}>>\n")
+    string(APPEND content "\n")
+    # Add "Libs:"
+    set(libs)
+    # Link options:
+    set(gx_libs
+        "-L\${libdir}"
+        "-l$<TARGET_PROPERTY:OUTPUT_NAME>"
+        $<GENEX_EVAL:$<TARGET_PROPERTY:pkg_config_LIBS>>
+        $<TARGET_PROPERTY:INTERFACE_LINK_OPTIONS>
```
eramongodb, line 268: > If you build with instrumentation (coverage, sanitizers), the installed pkg-config files will now include the necessary link flags as usage requirements.

I do not think this currently works as intended. Building with `-DMONGO_SANITIZE=address,undefined` does not inject `-fsanitize=address,undefined` into the the LFLAGS list as it should. Other dependent libraries (i.e. resolv) are also not present. Consider using `$<TARGET_PROPERTY:INTERFACE_LINK_LIBRARIES>` with appropriate parsing instead?

vector-of-bool: Oh oop. This is both right and wrong. The sanitize flags are not propagated because they are not added as interface properties on the targets that need them, but I have an upcoming set of changes that will make that work correctly, and I forgot that I hadn't yet PR'd the other changes.

src/libbson/CMakeLists.txt
vector-of-bool: We're already requiring this at the top level, so this line could likely actually just be removed. It's actually probably bad that we had `3.1` there, since it would rewind a bunch of CMake policies and probably had some unintentional effects. (Such as the problem in ResSearch).

build/cmake/ResSearch.cmake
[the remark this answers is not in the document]
vector-of-bool: I also did not like this change, and it isn't obvious from the documentation that this variable effects `try_compile`. I believe the reason we get away with it is that we enable bits and pieces by using platform-specific feature toggle macros. I will narrow it down and find what macros actually need to be defined to enable libresolv to compile properly.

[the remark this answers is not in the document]
vector-of-bool: Looks to actually be `_BSD_SOURCE`, which we already set, and I'm dropping it on the floor by using `cmake_push_check_state(RESET)`. Oops.

src/libbson/CMakeLists.txt
kevinAlbs: > I don't think we intend to support `add_subdirectory()` of a CXX Driver library (i.e. bsoncxx) _without_ going through the root directory, do we?

AFAIK there is no expectation to support `add_subdirectory()` without going through the root directory.

I expect C++ driver can be included with `add_subdirectory` of the root CMakeLists.txt, and CXX-2104 was filed to document.

This PR is for the C driver (not C++).

eramongodb: Had my repos mixed up. Sorry for the confusion.

src/libmongoc/CMakeLists.txt
```diff
@@ -807,6 +807,7 @@ set_target_properties (mongoc_shared PROPERTIES
    SOVERSION 0
    pkg_config_REQUIRES "libbson-1.0"
    )
+mongo_generate_pkg_config(mongoc_shared INSTALL RENAME libmongoc-${MONGOC_API_VERSION}.pc)
```
kevinAlbs, line 810: Consider using `FILENAME` for consistency with other `mongo_generate_pkg_config` calls.

Is there a use case for setting different values for `RENAME` and `FILENAME`? Consider removing `RENAME`.

kevinAlbs (approved):
GeneratePkgConfig.cmake seems generally useful.

src/libmongoc/CMakeLists.txt
vector-of-bool: `RENAME` comes from `install(FILES ...)`. The split in this case helps with multi-config generators. We want to allow multi-conf generation to work in the build, but we only need to install one version at a time.

eramongodb (approved):

PR #1332, opened 2023-06-30
Define a platform-support CMake target

vector-of-bool:
This changeset introduces a target `mongo::detail::c_platform`, which exposes usage requirements common to the platform on which we are being built. This consolidates some duplicated build setting handling between libbson and libmongoc, and introduces a central location for such settings.

This also replaces usage of `add_compile_options` for directory-level build settings. Such settings are now propagated properly on install results (including the generated pkg-config files).


joshbsiegel (approved):
LGTM

build/cmake/MongoPlatform.cmake
```diff
@@ -0,0 +1,72 @@
...
+            target_link_libraries(_mongo-platform INTERFACE
+                $<IF:$<TARGET_EXISTS:${item}>,${item},NO_SUCH_TARGET::${item}>)
+        endif()
+    endforeach()
+endfunction()
+
+# Enable multi-threading:
+set(THREADS_PREFER_PTHREAD_FLAG TRUE)
+find_package(Threads REQUIRED)
+mongo_platform_use_target(Threads::Threads)
```
eramongodb, line 72: Seems a little out-of-place. Suggest moving into root `CMakeLists.txt` given it applies to both libbson and libmongoc.

build/cmake/MongoPlatform.cmake
```diff
@@ -0,0 +1,72 @@
...
+install this target in both bson-targets and mongoc-targets export sets would
+lead to duplicate definitions of mongo::detail::c_platform for downstream
+users).
+
+]]
+
+add_library(_mongo-platform INTERFACE)
+add_library(mongo::detail::c_platform ALIAS _mongo-platform)
+set_property(TARGET _mongo-platform PROPERTY EXPORT_NAME detail::c_platform)
+install(TARGETS _mongo-platform EXPORT bson-targets)
```
eramongodb, line 22:
```suggestion
install(TARGETS _mongo-platform EXPORT bson-targets)
install(TARGETS _mongo-platform EXPORT mongoc-targets)
```
Do we want to include the imported target in the mongoc-targets file as well? Or are we fine with obtaining it transitively via libbson?

eramongodb (approved):
Minor feedback; otherwise, LGTM. 👏 for more target-based configuration.

build/cmake/MongoPlatform.cmake
eramongodb: Seems a little out-of-place. Consider moving the `install()` command into `src/libbson/CMakeLists.txt`?

vector-of-bool: It has to be exported only once, otherwise it will be defined twice and goofs will ensue. I chose the bson-targets export set only because it is pulled by both bson and mongoc. An "ideal" would be to define a third export set, but I think that has diminishing value when it can just ride along with the bson targets. I've also tucked it into the "detail" namespace so that we "reserve the right" to move/break/change/delete it in the future.

kevinAlbs (approved):

rcsanchez97 (approved):

PR #1333, opened 2023-06-30
[CDRIVER-4640] Remove the Release Archive

vector-of-bool:
Refer: CDRIVER-4640

This changeset removes the release archive.

# Background

The release archive was implemented as a source distribution method that contained only the relevant files to build the libraries and tests, as well as containing the built documentation.

# Why Remove It?

It is currently unclear if any users are particularly dependent upon this exact release format, and if there is significant benefit over a regular snapshot of the repository at the release tag. Maintaining the source distribution required manually listing the constituent files, which led to a lot of redundancy and trivial CMakeLists.txt that only contained these file lists. Forgetting to add/remove items from the listings would result in CI failures upon push.

Notably, the enumerated file listing restricted the inclusion of any generated build results that were not strictly known ahead-of-time. For documentation, this was a problem as Sphinx would change its output with different theme updates. This led to us vendoring in a Sphinx theme that is still used several years later, and it's age is showing 😦.

It was decided that the additional complexity and build constraints are not worth it going forward, so this changeset removes the release archive tasks.

# What's Here?

- Most of this changeset is deletion of the file listings that had to be maintained.
- Several CI tasks consumed the output source archive as their primary input. These tasks had to be modified to consume the repository directly.
  - These changes are found within the `link-sample-...` scripts, which now copy the source tree into their scratch space rather than expand an archive.
  - Because the dependency upon the `make-release-archive` task is gone, the `link-...` test tasks now queue much faster and will not bottleneck.
  - The `**Release Archive Creator` variant has been removed and replaced with a `Smoke Tests` variant, since that is what remains within it.
  - A new task `make-docs` is defined, which builds and uploads the generated documentation pages. (This was previously done as part of `make-release-archive`)
- The `release-compile` tasks common across most variants no longer uses the release archive. The `RELEASE` environment variable is now used simply to disable debug assertions. We may be safe to remove this task as low-value?


CMakeLists.txt
```diff
@@ -548,63 +554,7 @@ if (ENABLE_MONGOC)
    set (PACKAGE_PREFIX "mongo-c-driver-${MONGOC_DIST_VERSION}")
    set (DIST_FILE "${PACKAGE_PREFIX}.tar.gz")
```
kevinAlbs, line 555:
```suggestion
```

I suggest grepping this file for `dist`. Several comments and lines look no longer needed.

CMakeLists.txt
```diff
@@ -548,63 +554,7 @@ if (ENABLE_MONGOC)
...
-         -D BUILD_SOURCE_DIR=${BUILD_SOURCE_DIR}
-         -D SHELL=${SHELL}
-         "-Ddist_generated=\"${dist_generated}\""
-         -P ${PROJECT_SOURCE_DIR}/build/cmake/make_dist/MakeDist.cmake
-
-      DEPENDS
-         ${ALL_DIST} ${dist_generated_depends}
-   )
-
    if (NOT USE_SYSTEM_LIBBSON AND ENABLE_MAN_PAGES STREQUAL ON AND ENABLE_HTML_DOCS STREQUAL ON)
```
kevinAlbs, line 557: Remove unneeded `if` and `else`.

kevinAlbs (approved):
LGTM with additional comments addressed. The build simplification and removal of the CI bottleneck is much appreciated.

---

If this change is merged, I expect the C driver 1.25.0 will not release create/sign/upload a tarball. I suggest adding a note to the NEWS file for a new "Unreleased" section noting the removal of the tarball:

```
libmongoc 1.25.0 (Unreleased)
============================

* The release archive tarball is no longer created. Please instead use the snapshot tarball attached in GitHub for this release tag.
```
---

Remove the `## Creating and checking a distribution tarball` section from CONTRIBUTING.md.

---

> The RELEASE environment variable is now used simply to disable debug assertions. We may be safe to remove this task as low-value?

I incorrectly assumed those tasks were building with a release config (e.g. `-DCMAKE_BUILD_TYPE=RelWithDebInfo`). Though maybe that does not provide much value either. I support removing the `compile-release` tasks.

.evergreen/config_generator/components/make_docs.py
```diff
@@ -0,0 +1,140 @@
...
+    commands = [
+        bash_exec(
+            command_type=EvgCommandType.TEST,
+            working_dir="mongoc",
+            include_expansions_in_env=["distro_id"],
+            script="""\
+                set -o errexit
+                bash tools/poetry.sh install --with=docs
+                bash tools/poetry.sh run \
+                    bash .evergreen/scripts/build-docs.sh
```
eramongodb, line 21:
```suggestion
                bash tools/poetry.sh run bash .evergreen/scripts/build-docs.sh
```
Avoids messy formatting in the generated config file.

.evergreen/legacy_config_generator/evergreen_config_lib/tasks.py
```diff
@@ -171,7 +171,7 @@ def __init__(
 
         super().__init__(
             task_name=task_name,
-            depends_on=[OD([("name", "make-release-archive"), ("variant", "releng")])],
+            depends_on=[],
```
eramongodb, line 174:
```suggestion
```
Now unused; simply remove?

.evergreen/legacy_config_generator/evergreen_config_lib/variants.py
```diff
@@ -76,6 +76,7 @@ def days(n: int) -> int:
             # Disable the MongoDB legacy shell download, which is not available in 5.0 for u22
             "SKIP_LEGACY_SHELL": "1"
         },
+        tags=["merge-gate"],
```
eramongodb, line 79:
```suggestion
        tags=["pr-merge-gate"],
```
To make it clearer what "merge" is referring to.

.evergreen/scripts/install-uninstall-check.sh
```diff
@@ -10,19 +10,19 @@ DIR=$(dirname $0)
...
 
-if command -v gtar 2>/dev/null; then
-  TAR=gtar
-else
-  TAR=tar
-fi
+SCRATCH_DIR=$(pwd)/.scratch
+rm -rf "$SCRATCH_DIR"
+mkdir -p "$SCRATCH_DIR"
+cp -vr -- "$SRCROOT"/* "$SCRATCH_DIR"
```
eramongodb, line 18: Consider using `cmake -S . -B "$BUILD_DIR"` and `cmake --build "$BUILD_DIR"` instead of copying the source root directory and running CMake commands with relative paths from inside `$BUILD_DIR`.

eramongodb (approved):

rcsanchez97 (approved):
The changes look good to me.

I don't think that this has been brought up yet, but remember that this change will also break `release.py` in the tools repo. That script uses `make dist` for generating the release artifacts. It will require updating, and that probably also means that the document which describes the release steps will require updating.

PR #1339, opened 2023-07-12
(Debian packaging) use system uft8proc, ensure bundled source is not …

rcsanchez97:
…in .orig.tar.gz

Follow up for 47dc666478cfdde9f38ccefab033c6e0865e3fbf

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+mongo-c-driver (1.24.2-2) UNRELEASED; urgency=medium
+
+  * use system uft8proc, ensure bundled source is not in .orig.tar.gz
+
+ -- Roberto C. Sanchez <roberto@connexer.com>  Mon, 10 Jul 2023 14:04:12 -0400
+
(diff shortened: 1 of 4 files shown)


debian/changelog
```diff
@@ -1,3 +1,9 @@
+mongo-c-driver (1.24.2-2) UNRELEASED; urgency=medium
+
+  * use system uft8proc, ensure bundled source is not in .orig.tar.gz
```
kevinAlbs, line 3:
```suggestion
  * use system utf8proc, ensure bundled source is not in .orig.tar.gz
```

Including utf8proc is on the master branch and is planned in 1.25.0. Including utf8proc has not been released in 1.24.2. Should this PR be updated to remove references to utf8proc?

kevinAlbs: Disregard, after discussion I now understand this is not suggesting utf8proc changes are includes in 1.24.2.

kevinAlbs (approved):
LGTM with typo fix.

PR #1348, opened 2023-07-18
CDRIVER-4674 Remove unnecessary code from bson-clock.c

joshbsiegel:
# Summary
This PR will remove an `elif` block from `bson_get_monotonic_time` that checks if the driver is running on macOS as it serves no purpose anymore.

# Background
https://jira.mongodb.org/browse/CDRIVER-4565 fixed an assertion failure that occasionally occurred on macOS 10.14. This involved using `clock_gettime_nsec_np` and `CLOCK_UPTIME_RAW` to get the monotonic time on macOS machines. However, because `clock_gettime_nsec_np` and `CLOCK_UPTIME_RAW` were introduced in macOS 10.12 and `clock_gettime` and `CLOCK_MONOTONIC` were introduced in OS X 10.11 (one of these must not be defined to hit the `elif` block in question), this function won't compile on OS X 10.10 or earlier. 
As of now, all supported versions of macOS have `clock_gettime` and `CLOCK_MONOTONIC`, so this `elif` statement is no longer necessary. This PR will also ensure that the driver will compile on older versions of OS X, even if they are unable to use `bson_get_monotonic_time` sucessfully.

# What's New
- Removed unnecessary code from `bson_get_monotonic_time`
- Removed unnecessary headers from `bson-clock.c`


src/libbson/src/bson/bson-clock.c
```diff
@@ -16,9 +16,6 @@
 
 
 #ifdef __APPLE__
-#include <mach/clock.h>
-#include <mach/mach.h>
-#include <mach/mach_time.h>
 #include <sys/time.h>
```
jmikola, line 19: Is this include still needed? It dates back to https://github.com/mongodb/libbson/commit/c6e981ef08f60bb505578fadfe96d328e4b2ea7a, although it's not clear if it was specific to the use of `mach_` functions since that commit also included it for the `defined(HAVE_CLOCK_GETTIME)` case.

src/libbson/src/bson/bson-clock.c
```diff
@@ -16,10 +16,6 @@
 
 
 #ifdef __APPLE__
```
jmikola, line 18: This entire `#ifdef` can be removed if it no longer has a body.

joshbsiegel: I thought I already did this and I have no idea why that commit wasn't showing before. I just made a new commit ("remove empty line") and now the "remove #ifdef" commit is showing :man-shrugging:

jmikola (approved):
@kevinAlbs should probably still sign off on this.

kevinAlbs (approved):
> even if they are unable to use `bson_get_monotonic_time` sucessfully.

I expect `bson_get_monotonic_time` will continue to "succeed" and fall back to the `else` case of using `bson_gettimeofday`, and show the "Monotonic clock is not yet supported on your platform" warning on compile.
I think that is OK. Support of macOS 10.14 and 10.15 was decidedly dropped in 1.24.0 as part of CDRIVER-4580.

PR #1349, opened 2023-07-19
CDRIVER-4662 Check for Decimal128 exponent overflow

joshbsiegel:
# Summary
This PR will check that the exponent on a Decimal128 does not overflow when converting from JSON to BSON.

# Background
It was brought up [here](https://github.com/mongodb/specifications/pull/1432#pullrequestreview-1482601195) that the C driver does not throw an error when a user tries to convert a Decimal128 value with an exponent that doesn't fit in a 32 bit integer. 

# What's new
- Modified `bson_decimal128_from_string` to read in the exponent as an `int64_t` and checking if that number would fit in a 32 bit integer, returning `false` on failure
- Checking the return status of the `bson_decimal128_from_string` in `_bson_json_read_string` and setting error on failure
- Spec tests added from https://github.com/mongodb/specifications/pull/1432
- Test cases for both `bson_new_from_json` and `bson_decimal128_from_string`


src/libbson/src/bson/bson-json.c
```diff
@@ -1124,9 +1124,9 @@ _bson_json_read_string (bson_json_reader_t *reader, /* IN */
...
       case BSON_JSON_LF_DECIMAL128: {
          bson_decimal128_t decimal128;
-         bson_decimal128_from_string (val_w_null, &decimal128);
 
-         if (bson->read_state == BSON_JSON_IN_BSON_TYPE) {
+         if (bson_decimal128_from_string (val_w_null, &decimal128) &&
+             bson->read_state == BSON_JSON_IN_BSON_TYPE) {
             bson->bson_type_data.v_decimal128.value = decimal128;
          } else {
             goto BAD_PARSE;
```
joshbsiegel, line 1132: On failure, `BAD_PARSE` will set the error message as 
```
Invalid input string "0E+xxx", looking for DECIMAL128
```
I think this message is fine as is, but potentially could be more helpful to the user with some changes. I believe it would involve adding some return information from `bson_decimal128_from_string` as to why the parsing fails (because it may not have to do with an overflowing exponent). Open to thoughts.

src/libbson/tests/test-json.c
```diff
@@ -3555,6 +3555,86 @@ test_bson_as_json_with_opts_all_types (void)
...
+         "{\"d\" : {\"$numberDecimal\" : \"-0E-2147483649\"}}";
+
+      BSON_ASSERT (!bson_new_from_json (
+         (const uint8_t *) degenerate_extjson, -1, &error));
+      ASSERT_ERROR_CONTAINS (
+         error,
+         BSON_ERROR_JSON,
+         BSON_JSON_ERROR_READ_INVALID_PARAM,
+         "Invalid input string \"-0E-2147483649\", looking for DECIMAL128");
+   }
```
kevinAlbs, line 3635: Suggest testing what happens when parsing an exponent outside of range of INT64 min and max.

src/libbson/src/bson/bson-decimal128.c
```diff
@@ -581,10 +581,12 @@ bson_decimal128_from_string_w_len (const char *string,     /* IN */
 #else
 #define SSCANF sscanf
 #endif
-      int read_exponent = SSCANF (++str_read, "%d%n", &exponent, &nread);
```
kevinAlbs, line 584: What was the cause of the issue? Was `SSCANF` not returning an error when parsing integer literals outside of the range of INT32?

That may help to determine if a problem would occur with an exponent outside of the range of INT64.

src/libbson/src/bson/bson-decimal128.c
```diff
@@ -581,10 +581,12 @@ bson_decimal128_from_string_w_len (const char *string,     /* IN */
...
+      int read_exponent =
+         SSCANF (++str_read, "%" PRId64 "%n", &exponent, &nread);
```
kevinAlbs, line 585: Please investigate the [test failures on Windows](https://parsley.mongodb.com/evergreen/mongo_c_driver_windows_2017_test_latest_server_ipv4_client_ipv4_noauth_nosasl_nossl_patch_797b371ce9c23a4a557cbd4ac5ac20ddd762c280_64b8476a3066154b1c1e58a9_23_07_19_20_28_28/0/task?bookmarks=0,6124&shareLine=930).

The definition of `SSCANF` differs on Windows and may be related to the error.

src/libbson/src/bson/bson-decimal128.c
joshbsiegel: SSCANF was parsing integers without error, but would overflow as such:
```
+2147483648 -> -2147483648 // min 32 bit integer
+2147483649 -> -2147483647
+2147483650 -> -2147483646

-2147483649 -> 2147483647 // max 32 bit integer
-2147483650 -> 2147483646
-2147483651 -> 2147483645
```
This behavior is demonstrated in this branch https://github.com/mongodb/mongo-c-driver/compare/master...joshbsiegel:mongo-c-driver:cdriver-4662-demo?expand=1 by running `./cmake-build/src/libmongoc/test-libmongoc -d --no-fork -l "/bson/test_overflowing_exponent"`

src/libbson/src/bson/bson-decimal128.c
joshbsiegel: It ended up being an issue with not converting the INT32 max and min values to `int64_t` :expressionless:

vector-of-bool (approved):
LGTM

src/libbson/src/bson/bson-decimal128.c
```diff
@@ -581,14 +581,18 @@ bson_decimal128_from_string_w_len (const char *string,     /* IN */
...
 #endif
-      int read_exponent = SSCANF (++str_read, "%d%n", &exponent, &nread);
+      int64_t temp_exponent = 0;
+      int read_exponent =
+         SSCANF (++str_read, "%" SCNd64 "%n", &temp_exponent, &nread);
       str_read += nread;
 
-      if (!read_exponent || nread == 0) {
+      if (!read_exponent || nread == 0 || temp_exponent > (int64_t) INT32_MAX ||
+          temp_exponent < (int64_t) INT32_MIN) {
```
kevinAlbs, line 590:
```suggestion
      if (!read_exponent || nread == 0 || !bson_in_range_int32_t_signed (temp_exponent)) {
```

src/libbson/src/bson/bson-decimal128.c
```diff
@@ -501,7 +501,7 @@ bson_decimal128_from_string_w_len (const char *string,     /* IN */
    size_t first_digit = 0;           /* The index of the first non-zero digit */
    size_t last_digit = 0;            /* The index of the last digit */
 
-   int32_t exponent = 0;
+   int64_t exponent = 0;
```
kevinAlbs, line 504:
```suggestion
   int32_t exponent = 0;
```

Change can be reverted since `exponent` is assigned after casting to `int32_t`.

src/libbson/src/bson/bson-json.c
kevinAlbs: I think it may help to determine the cause of the error with a new function: `bson_decimal128_from_string_w_error`. But I think that is separate from this bug fix and could be considered in a future change.

kevinAlbs (approved):
LGTM with minor changes.

This is a bug fix and I think suitable for a patch release. After merging, please cherry-pick the commit to the `r1.24` branch with:

```
git checkout --track upstream/r1.24
git cherry-pick <commit from master>
git push upstream r1.24
```

PR #1351, opened 2023-07-20
CDRIVER-2928 assert returned values of `bson_init_static`

kevinAlbs:
# Summary
- assert returned values of `bson_init_static`

# Background & Motivation
Addresses Coverity issues with CIDs: 134030, 134029, 120108, and 120107.

The test macro `ASSERT` was used in test code. `BSON_ASSERT` was used in driver code.

`bson_init_static` returns false on failure. Failure may occur if the input BSON data is invalid. Callers ignoring the returned value have been updated to assert to result in an early error on invalid BSON data input.

### DIFF
--- a/src/libbson/src/bson/bson-dsl.h
+++ b/src/libbson/src/bson/bson-dsl.h
@@ -613,7 +613,7 @@ BSON_IF_GNU_LIKE (_Pragma ("GCC diagnostic ignored \"-Wshadow\""))
          break;                                                                \
       }                                                                        \
       bson_t inner;                                                            \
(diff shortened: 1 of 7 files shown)


joshbsiegel (commented):
Should we also be asserting the 3 calls to `bson_init_static` that I see in `bson-dsl.h` for completeness?

kevinAlbs:
> Should we also be asserting the 3 calls to `bson_init_static` that I see in `bson-dsl.h` for completeness?

Good catch. Done.

joshbsiegel (approved):
LGTM. Auto-formatting is fun.

PR #1354, opened 2023-07-21
Update Sphinx, Theming, and Faster Builds

vector-of-bool:
This changeset does a few things:

1. Sphinx is now updated to 7.0.1
2. The vendored theme has been removed, and documentation is now generated with [Furo](https://pradyunsg.me/furo/)
3. Generated documentation uses the `dirhtml` Sphinx builder. This generates "prettier" URLs by placing every page within its own directory with a single `index.html` inside. The result is that URLs no longer have a `.html` suffix.
4. To keep old links working, the old `.html` files are generated and just redirect to the new URLs.
5. The "man pages hack" that we were using was breaking parallel builds in Sphinx. This hack has been replaced with a less-janky hack that works with parallel builds.

Other than the above, the documentation content has remained unchanged. Future changes will include documentation content updates. The rendered results of these changes are available in the `make-docs` tasks in EVG.


src/libbson/doc/conf.py
```diff
@@ -5,49 +5,43 @@
...
-}
+html_sidebars = {}
 
 html_use_index = False
 
 
 def add_canonical_link(app, pagename, templatename, context, doctree):
-    link = ('<link rel="canonical"'
-            ' href="https://www.mongoc.org/libbson/current/%s.html"/>' % pagename)
+    link = '<link rel="canonical"' ' href="https://www.mongoc.org/libbson/current/%s.html"/>' % pagename
```
eramongodb, line 40:
```suggestion
    link = '<link rel="canonical" href="https://www.mongoc.org/libbson/current/%s.html"/>' % pagename
```
Simplification?

build/sphinx/mongoc_common.py
```diff
@@ -1,12 +1,20 @@
...
+from sphinx.application import Sphinx
+from sphinx.application import logger as sphinx_log
+from sphinx.builders.dirhtml import DirectoryHTMLBuilder
+from sphinx.config import Config
+from sphinx.project import Project
+from sphinx.util.docutils import SphinxDirective
 
-needs_sphinx = '1.6'
-author = 'MongoDB, Inc'
+needs_sphinx = "1.6"
```
eramongodb, line 16:
```suggestion
needs_sphinx = "7.0"
```
Update Sphinx version (as done for Poetry)?

Makefile
eramongodb, line 1: Is this file intentionally included in this PR?

vector-of-bool: Yes, this is intentional

build/sphinx/mongoc_common.py
vector-of-bool: Updated with 6.0, as that is the version available in Fedora Rawhide and used for the RPM build (for now)

Makefile
eramongodb: Where/how is this intended to be used? Unless I missed something, I don't think it's currently being referenced by the other changes in the PR. If it's just to define convenience targets to ensure proper execution of Poetry commands, I strongly recommend avoiding introducing a top-level Makefile into the repo and instead define it within the CMake configs or as `build/` scripts instead.

build/sphinx/mongoc_common.py
vector-of-bool: Updated again to 5.0, since we're building on Fedora 38 in CI (not Rawhide).

Makefile
vector-of-bool: Deleted for now. May bring back someday in a different form.

pyproject.toml
```diff
@@ -17,9 +17,8 @@ python = "^3.8"
 optional = true
 
 [tool.poetry.group.docs.dependencies]
-sphinx = "1.8"
-jinja2 = "<3"
-markupsafe = "<2.1"
+sphinx = "^7.0.1"
```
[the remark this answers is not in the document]
vector-of-bool: I don't think so, in this case. The only reason `needs_sphinx` is lower is to satisfy packager requirements, which only really build the `man` pages. For generating the HTML pages, newer Sphinx versions would be preferable. It's possible that the man pages will be dropped soon, which would remove the packager requirement entirely.

kevinAlbs (approved):
LGTM. I like the new documentation theme. And the parallel build is much appreciated.

PR #1355, opened 2023-07-24
CDRIVER-4493 Override host platform, os, flags, compiler info for testing

adriandole:
Tests in `test-mongo-handshake.c` expect that the handshake doc is not truncated, causing failures with some test hosts with a long platform string. This PR overrides host info for testing.

### DIFF
--- a/src/libmongoc/tests/test-mongoc-handshake.c
+++ b/src/libmongoc/tests/test-mongoc-handshake.c
@@ -220,12 +220,14 @@ _handshake_check_os (bson_t *doc)
    ASSERT (val);
    ASSERT (strlen (val) > 0);
 
+   /* Check os version valid */
    ASSERT (bson_iter_find (&inner_iter, "version"));
    ASSERT (BSON_ITER_HOLDS_UTF8 (&inner_iter));
    val = bson_iter_utf8 (&inner_iter, NULL);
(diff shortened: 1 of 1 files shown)


src/libmongoc/tests/test-mongoc-handshake.c
```diff
@@ -220,18 +220,20 @@ _handshake_check_os (bson_t *doc)
...
-   ASSERT (bson_iter_find (&inner_iter, "version"));
-   ASSERT (BSON_ITER_HOLDS_UTF8 (&inner_iter));
-   val = bson_iter_utf8 (&inner_iter, NULL);
-   _check_os_version_valid (val);
+   /* Check os version is valid if present (may be truncated) */
+   if (bson_iter_find (&inner_iter, "version")) {
+      ASSERT (BSON_ITER_HOLDS_UTF8 (&inner_iter));
+      val = bson_iter_utf8 (&inner_iter, NULL);
+      _check_os_version_valid (val);
+   }
```
kevinAlbs, line 228: Since `_override_host_platform_os` is used to avoid truncation, can `version` or `architecture` be assumed to exist? If so, suggest reverting the added `if` checks to improve test coverage.

kevinAlbs (approved):
LGTM with question and possible suggestion.

PR #1356, opened 2023-07-24
CDRIVER-4678 Fix `$code` and `$dbPointer` array parsing

joshbsiegel:
# Summary
This PR will fix an error when parsing an array with `$code` or `$dbPointer` where the implicit keys (0, 1, 2, etc.) would not get correctly set on the values when converting to BSON.

# Background
Previously, when parsing JSON such as
```
 [{"$code" : "A"}]
```
the BSON outputted would not correctly attach a key of `0` to the value of `{"$code" : "A"}`. The same issue was caused when attempting to parse JSON with `$dbpointer`.

# Fix
The `$code` and `$dbPointer` handlers in `bson-json.c` were the only two types that called `_bson_json_buf_set` to save the value of the parent key (e.g. `{"key": {"$code" : "A"}}`, where we want to save `"key"` somewhere). These function calls were made by passing through `bson->key_buf.buf` as `const void *from`. However, `bson->key_buf.buf` wasn't set to 0, 1, 2, etc., `bson->key` was. By passing through `bson->key` instead and still using `bson->key_buf.len` as the length, the implicit keys will get set correctly. 
Although mixing `key` with `key_buf` seems unorthodox, this is a practice done elsewhere in the codebase: https://github.com/mongodb/mongo-c-driver/blob/ac436db8c1c0ee45a663a469e0783d50713e16e3/src/libbson/src/bson/bson-json.c#L1290

# What's New
- Fixed calls to `_bson_json_buf_set` by passing through `bson->key`
- Added tests for `$code` and `$dbPointer` arrays with 1 and 2 elements


kevinAlbs (approved):
LGTM with comment removed.

> Although mixing `key` with `key_buf` seems unorthodox

I think it is expected to use `key` to refer to the key string. Consider this call:
```c
      bson->key_buf.len = bson_uint32_to_string (
         STACK_I, &bson->key, (char *) bson->key_buf.buf, 12);
```

`bson_uint32_to_string` documents that the buffer may or may not be used. The out pointer is always set to the resulting string:

> ```
>  size_t
>   bson_uint32_to_string (uint32_t value,
>                          const char **strptr,
>                          char *str,
>                          size_t size);
> ```
> 
> 
> Converts ``value`` to a string.
> 
> If ``value`` is from 0 to 999, it will use a constant string in the data section of the library.
> 
> If not, a string will be formatted using ``str`` and ``snprintf()``.
> 
> ``strptr`` will always be set. It will either point to ``str`` or a constant string. Use this as your key.
>

PR #1360, opened 2023-07-25
assert return of `mcd_rpc_message_to_iovecs`

kevinAlbs:
# Summary
- assert return value of `mcd_rpc_message_to_iovecs`.

# Background & Motivation

Addresses Coverity issue with CID 138246: "Dereference null return value". I do not expect these calls to `mcd_rpc_message_to_iovecs` to possibly fail given the current implementation. The added `BSON_ASSERT` is consistent with other calls. If future changes to `mcd_rpc_message_to_iovecs` result in a condition where `mcd_rpc_message_to_iovecs` may return NULL, this may prevent a NULL dereference.


eramongodb (approved):
Looks like I had missed a few spots. Thank you for the fixes. LGTM!

PR #1363, opened 2023-07-27
Address C4267 size_t conversion warnings

eramongodb:
Followup to https://github.com/mongodb/mongo-c-driver/pull/1344.


kevinAlbs (approved):
Thank you for the fix.

PR #1366, opened 2023-07-31
CDRIVER-4612 document management of search indexes

kevinAlbs:
# Summary

- Document management of Atlas Search Indexes.

# Background & Motivation

DRIVERS-2637 adds new API to manage Atlas Search Indexes. This PR documents how to run the management commands with existing C driver API.

## Rejected idea: Adding API

An API to model the specified [Standard API for Search Indexes](https://github.com/mongodb/specifications/blob/c773534925f83ecc5728971442baa63d69ecf0ce/source/index-management/index-management.rst#standard-api-for-search-indexes) was considered but rejected.

The Search Indexes helpers do not have special option handling. IMO adding the API would not provide value to wrapping drivers. This PR proposes not adding the API, and instead documents how to manage Search Indexes. If the specified behavior of Search Indexes helpers changes in the future, adding helpers can be reconsidered.

## Rejected idea: Adding end-to-end tests to Evergreen

End-to-end (E2E) tests in DRIVERS-2630 were decidedly not added. The proposed C driver changes do not add new API. The E2E tests in Evergreen may not add helpful test coverage. E2E tests require creating an Atlas cluster with [setup-atlas-cluster.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/tree/d0ede2a2a5ac8090343ac5c04ab616b376658bdc/.evergreen/atlas). A local run took 9m28s. I do not think it is worth the effort or maintenance burden.

`example-manage-search-indexes` was tested manually with an Atlas cluster created with [setup-atlas-cluster.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/tree/d0ede2a2a5ac8090343ac5c04ab616b376658bdc/.evergreen/atlas):
```
% ./cmake-build/src/libmongoc/example-manage-search-indexes $ATLAS_SEARCH_INDEXES_URI
Created index: "test-index"
Listing indexes:
  { "id" : "64c3c96874355a6d0050331f", "name" : "test-index", "status" : "PENDING", "queryable" : false, "latestDefinition" : { "mappings" : { "dynamic" : false } } }
Updated index: "test-index"
Dropped index: "test-index"
```

The JSON specification tests are added. The JSON specification tests may be a simpler and lower maintenance test of the documented approach for sending management commands: use of `mongoc_collection_command_simple`. The JSON specification tests can be removed if desired.


src/libmongoc/examples/example-manage-search-indexes.c
```diff
@@ -0,0 +1,201 @@
...
+               } ]
+            }),
+            collname);
+         ASSERT (bson_init_from_json (&cmd, cmd_str, -1, &error));
+         bson_free (cmd_str);
+      }
+      if (!mongoc_collection_command_simple (
+             coll, &cmd, NULL /* read_prefs */, NULL /* reply */, &error)) {
+         bson_destroy (&cmd);
+         HANDLE_ERROR ("Failed to run createSearchIndexes: %s", error.message);
```
eramongodb, line 104: If this example is run against a non-Atlas server, this example fails on this command attempt and error with a "no such cmd" message. Can we consider adding an earlier "example must be run against Atlas" check and error to make the required conditions a bit more clear?

eramongodb (approved):

joshbsiegel (approved):
LGTM

src/libmongoc/examples/example-manage-search-indexes.c
kevinAlbs: I like that idea. ee7806230ad4b84a73e5b75ff256785fe02d262c adds an earlier check that `$listSearchIndexes` is supported, and prints an error suggesting the example may not be using Atlas if an error occurred.

PR #1370, opened 2023-08-03
CDRIVER-4679 Revert BSON_HAVE_XSI_STRERROR_R and use strerror_l instead

eramongodb:
This PR reverts https://github.com/mongodb/mongo-c-driver/pull/1350 and re-addresses CDRIVER-4679 using a more nuanced solution. Given the information that [strerror_r is scheduled to be marked as obsolete](https://www.austingroupbugs.net/view.php?id=655) in the upcoming POSIX Issue, this PR favors using `strerror_l` instead of `strerror_r`, which is not susceptible to the GNU extension and musl library compatibility problems documented in https://github.com/mongodb/mongo-c-driver/pull/1350. Per the provided link:

> The strerror_r function is problematic because there are historically at least two versions, and at least one major implementation, the GNU C Library, seems committed to providing a version incompatible with the version specified in the standard. While an attempt is made in glibc to provide a conforming version when the proper feature test macros are defined, the workaround is header-based and is in fact non-conforming since it does not support applications that omit inclusion of string.h and declare the function manually.
> 
> The strerror_l function is mandatory beginning with issue 7, provides a much easier-to-use interface to the same functionality, and does not have incompatible historical versions of the interface that can interfere with application compatibility.

Unfortunately, the Apple platform does not seem to yet provide `strerror_l`. This PR proposes treating the Apple platform specially, as is already being done for the Windows platform, by unconditionally using `strerror_r` (as `strerror_s` on Windows). Furthermore, given the possibility of the C Driver being compiled with GCC on Apple, it is still susceptible to the `strerror_r` problems that are documented in https://github.com/mongodb/mongo-c-driver/pull/1350. This PR proposes a compromise _on the Apple platform only_ that deliberately ignores checking the return value for errors by depending on the quality of implementation to avoid undefined behavior, per the [POSIX Specification](https://pubs.opengroup.org/onlinepubs/9699919799/functions/strerror.html):

> If an invalid error number is supplied as the value of errnum, applications should be prepared to handle any of the following: [...] Since applications frequently use the return value of strerror() as an argument to functions like `fprintf()` (without checking the return value) and since applications have no way to parse an error message string to determine whether errnum represents a valid error number, implementations are encouraged to implement [the behavior where _errno_ is set to [EINVAL] and the return value points to a string like `"unknown error"` or `"error number xxx"` (where _xxx_ is the value of _errnum_)]. Similarly, implementations are encouraged to have `strerror_r()` return [EINVAL] and put a string like `"unknown error"` or `"error number xxx"` in the buffer pointed to by _strerrbuf_ when the value of _errnum_ is not a valid error number.

We will assume similar QoI for potential [ERANGE] errors.

Although we do not expect anyone to be compiling the C Driver _without_ `_XOPEN_SOURCE=700` on non-Windows or non-Apple platforms (per default), this PR leaves the `_GNU_SOURCE` condition in out of abundance of caution.

### DIFF
--- a/src/libbson/src/bson/bson-error.c
+++ b/src/libbson/src/bson/bson-error.c
@@ -125,6 +125,19 @@ bson_strerror_r (int err_code,                    /* IN */
    // string even when `strerror_r` fails.
    (void) strerror_r (err_code, buf, buflen);
 #elif defined(_XOPEN_SOURCE) && _XOPEN_SOURCE >= 700
+   // The behavior (of `strerror_l`) is undefined if the locale argument to
+   // `strerror_l()` is the special locale object LC_GLOBAL_LOCALE or is not a
+   // valid locale object handle.
+   locale_t locale = uselocale ((locale_t) 0);
+   // No need to test for error (it can only be [EINVAL]).
+   if (locale == LC_GLOBAL_LOCALE) {
+      // Only use our own locale if a thread-local locale was not already set.
+      // This is just to satisfy `strerror_l`. We do NOT want to unconditionally
+      // set a thread-local locale.
+      locale = newlocale (LC_MESSAGES_MASK, "C", (locale_t) 0);
+   }
+   BSON_ASSERT (locale != LC_GLOBAL_LOCALE);
+
    // Avoid `strerror_r` compatibility headaches with GNU extensions and the
    // musl library by using `strerror_l` instead. Furthermore, `strerror_r` is
    // scheduled to be marked as obsolete in favor of `strerror_l` in the
@@ -134,10 +147,18 @@ bson_strerror_r (int err_code,                    /* IN */
    // POSIX Spec: since strerror_l() is required to return a string for some
    // errors, an application wishing to check for all error situations should
    // set errno to 0, then call strerror_l(), then check errno.
-   errno = 0;
-   ret = strerror_l (err_code, uselocale ((locale_t) 0));
-   if (errno != 0) {
-      ret = NULL;
+   if (locale != (locale_t) 0) {
+      errno = 0;
+      ret = strerror_l (err_code, locale);
+
+      if (errno != 0) {
+         ret = NULL;
+      }
+
+      freelocale (locale);
+   } else {
+      // Could not obtain a valid `locale_t` object to satisfy `strerror_l`.
+      // Fallback to `bson_strncpy` below.
    }
 #elif defined(_GNU_SOURCE)
    // Unlikely, but continue supporting use of GNU extension in cases where the
--- a/src/libbson/src/bson/bson-error.c
+++ b/src/libbson/src/bson/bson-error.c
@@ -121,8 +121,10 @@ bson_strerror_r (int err_code,                    /* IN */
    // the XSI-compliant `strerror_r`, but only when compiling with Apple Clang.
    // GNU extensions may still be a problem if we are being compiled with GCC on
    // Apple. Avoid the compatibility headaches with GNU extensions and the musl
-   // library by assuming QoI will not cause UB when reading the error message
-   // string even when `strerror_r` fails.
+   // library by assuming the implementation will not cause UB when reading the
+   // error message string even when `strerror_r` fails, as encouraged (but not
+   // required) by the POSIX spec (see:
+   // https://pubs.opengroup.org/onlinepubs/9699919799/functions/strerror.html#tag_16_574_08).
    (void) strerror_r (err_code, buf, buflen);
 #elif defined(_XOPEN_SOURCE) && _XOPEN_SOURCE >= 700
    // The behavior (of `strerror_l`) is undefined if the locale argument to
--- a/src/libbson/src/bson/bson-error.c
+++ b/src/libbson/src/bson/bson-error.c
@@ -109,14 +109,12 @@ bson_strerror_r (int err_code,  /* IN */
    if (strerror_s (buf, buflen, err_code) != 0) {
       ret = buf;
    }
-#elif defined(BSON_HAVE_XSI_STRERROR_R)
+#elif defined(__GNUC__) && defined(_GNU_SOURCE)
+   ret = strerror_r (err_code, buf, buflen);
+#else /* XSI strerror_r */
    if (strerror_r (err_code, buf, buflen) == 0) {
       ret = buf;
    }
-#elif defined(_GNU_SOURCE)
-   ret = strerror_r (err_code, buf, buflen);
-#else
-   #error "Unable to find a supported strerror_r candidate"
 #endif
 
    if (!ret) {
--- a/src/libbson/src/bson/bson-error.c
+++ b/src/libbson/src/bson/bson-error.c
@@ -25,6 +25,11 @@
 #include "bson-string.h"
 #include "bson-types.h"
 
+// See `bson_strerror_r()` definition below.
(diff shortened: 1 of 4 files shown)


src/libbson/src/bson/bson-error.c
```diff
@@ -98,25 +103,47 @@ bson_set_error (bson_error_t *error, /* OUT */
...
-   if (strerror_r (err_code, buf, buflen) == 0) {
-      ret = buf;
+#elif defined(__APPLE__)
+   // Apple does not provide `strerror_l`, but it does unconditionally provide
+   // the XSI-compliant `strerror_r`, but only when compiling with Apple Clang.
+   // GNU extensions may still be a problem if we are being compiled with GCC on
+   // Apple. Avoid the compatibility headaches with GNU extensions and the musl
+   // library by assuming QoI will not cause UB when reading the error message
+   // string even when `strerror_r` fails.
+   (void) strerror_r (err_code, buf, buflen);
```
jmikola, line 126: Should we consider using this code path for the final `#else` case instead of raising a build error of "Unable to find a supported strerror_r candidate"?

I'll note that PHP core does that (see: https://github.com/php/php-src/pull/11624#discussion_r1272237069). IMO it's not the ideal approach (I disagree with PHP doing it for _all_ cases, since they never bother to detect the XSI variant), but it generally works. And given that errno translation probably isn't a critical component of libbson, I think a user would be more inclined to receive "Unknown error" at runtime than a build error.

Looking at uses of `bson_strerror_r()` throughout libbson and libmongoc, I see that the `err_code` param is only _sometimes_ included in the logged error message. That seems like something that could be addressed independently. Rather than hunt through all of those occurrences, a compromise might be to incorporate it into the returned string for `!ret` at the end of this function (e.g. `"Unknown error (errno: %d)"`).

src/libbson/src/bson/bson-error.c
```diff
@@ -98,25 +103,47 @@ bson_set_error (bson_error_t *error, /* OUT */
...
+   // musl library by using `strerror_l` instead. Furthermore, `strerror_r` is
+   // scheduled to be marked as obsolete in favor of `strerror_l` in the
+   // upcoming POSIX Issue 8 (see:
+   // https://www.austingroupbugs.net/view.php?id=655).
+   //
+   // POSIX Spec: since strerror_l() is required to return a string for some
+   // errors, an application wishing to check for all error situations should
+   // set errno to 0, then call strerror_l(), then check errno.
+   errno = 0;
+   ret = strerror_l (err_code, uselocale ((locale_t) 0));
```
jmikola, line 138: [`strerror(3)`](https://man7.org/linux/man-pages/man3/strerror.3.html) states:

> The behavior of strerror_l() is undefined if locale is the special locale object LC_GLOBAL_LOCALE or is not a valid locale object handle.

Quoting [`uselocale(3)`](https://man7.org/linux/man-pages/man3/uselocale.3.html):

> On success, uselocale() returns the locale handle that was set by the previous call to uselocale() in this thread, or LC_GLOBAL_LOCALE if there was no such previous call. On error, it returns (locale_t) 0, and sets [errno](https://man7.org/linux/man-pages/man3/errno.3.html) to indicate the error.

I don't see any other calls to `uselocale()` throughout libbson and libmongoc. Do we have to worry about `LC_GLOBAL_LOCALE` being returned? If the intention is to disregard a locale altogether, then would it make more sense to just pass `(locale_t) 0` directly to `strerror_l` instead?

I don't have much experience with `uselocale()`, but a related discussion that came up while searching for this `LC_GLOBAL_LOCALE` edge case was: https://austin-group-l.opengroup.narkive.com/MFGftIBP/use-of-lc-global-locale-with-l-functions

eramongodb: Good call. Implemented the missing locale handling to ensure we always pass a valid locale.

src/libbson/src/bson/bson-error.c
jmikola: > For example, bson_json_reader_new_from_file's errmsg_buf buffer is [uninitialized](https://github.com/mongodb/mongo-c-driver/blob/14edfd2c605be54e11ec133e04ca0e320dec3ec8/src/libbson/src/bson/bson-reader.c#L784). If strerror_r fails internally and does not write a failure message to the buffer, the successful result of bson_strerror_r would point to the uninitialized buffer as the error message string, and any subsequent reads of that error message string would be undefined behavior.

Ah, very good point. I was confused by jumping back and forth between libbson and PHP, which allocates the buffer in https://github.com/php/php-src/blob/3d832da8f53d2c2f93d4ffbd1c807d9a5d66536b/Zend/zend.c#L1631-L1641 (where the worst case scenario there is just an empty string).

SGTM.

src/libbson/src/bson/bson-error.c
```diff
@@ -134,10 +147,18 @@ bson_strerror_r (int err_code,                    /* IN */
...
    // set errno to 0, then call strerror_l(), then check errno.
-   errno = 0;
-   ret = strerror_l (err_code, uselocale ((locale_t) 0));
-   if (errno != 0) {
-      ret = NULL;
+   if (locale != (locale_t) 0) {
+      errno = 0;
+      ret = strerror_l (err_code, locale);
+
+      if (errno != 0) {
```
jmikola, line 154: I had a fleeting thought of "what if we wanted to translate `errno` here? We can't possibly use `strerror_l()`!"

:laugh-cry:

jmikola (approved):
This seems OK to me, so over to @kevinAlbs.

src/libbson/src/bson/bson-error.c
eramongodb: > where the worst case scenario there is just an empty string

FYI the linked function is stil susceptible to the same potential undefined behavior as described for `bson_json_reader_new_from_file`: the `buf` array is allocated but its elements are left uninitialized. Subsequent read of the `buf` elements in `zend_error_noreturn` may be undefined behavior if `strerror_r` failed and did not write a message to `buf`. The risk could be reduced by initializing the elements of the `buf` array, but it would still be a risk nevertheless (reliance on QoI to always write something to `buf` even on failure).

src/libbson/src/bson/bson-error.c
```diff
@@ -98,25 +103,69 @@ bson_set_error (bson_error_t *error, /* OUT */
...
    }
-#elif defined(BSON_HAVE_XSI_STRERROR_R)
-   if (strerror_r (err_code, buf, buflen) == 0) {
-      ret = buf;
+#elif defined(__APPLE__)
+   // Apple does not provide `strerror_l`, but it does unconditionally provide
+   // the XSI-compliant `strerror_r`, but only when compiling with Apple Clang.
+   // GNU extensions may still be a problem if we are being compiled with GCC on
+   // Apple. Avoid the compatibility headaches with GNU extensions and the musl
+   // library by assuming QoI will not cause UB when reading the error message
```
kevinAlbs, line 124:
```suggestion
   // library by assuming implementation will not cause UB when reading the error message
```

I was unaware of QoI abbreviation. Suggestion can be ignored if this is common parlance.

kevinAlbs (approved):

src/libbson/src/bson/bson-error.c
jmikola: Ah, right you are again. The `buf` is declared within the function and unallocated. I was thinking of a static/global decl being allocated with zeroes.

PR #1371, opened 2023-08-03
[CDRIVER-3034] New Installation/Getting-Started Documentation Pages

vector-of-bool:
This is the first large changeset that introduces substantial new documentation updates, especially regarding tutorials and how-tos.

This adopts guidelines from [The Documentation System](https://documentation.divio.com/) (strongly recommended reading), and begins to subdivide our documentation into its prescribed four distinct categories:

1. Getting Started & Tutorials: Placed within `/learn/`
2. Cookbook & How-Tos: Placed within `/howto/`
3. Reference Material: Placed within `/ref/` (The existing API pages are not modified, but the platform support reference has been moved here).
4. Explanation & Background: Not (yet) present in our documentation.

**To view the built documentation**, explore the "libmongoc docs" file result attached to the "make-docs" task on the latest EVG build for this PR.

# Tutorials

The tutorials have been subdivided further, and currently only contains a `/get/` grouping regarding how to obtain the libraries. There are currently three tutorials:

1. Obtaining the libraries by building from source: `/learn/get/from-source`
2. Obtaining the libraries by installing on Linux and macOS (via Homebrew): `/learn/get/installing`
3. Generating the documentation pages from source: `/learn/get/docs`

# How-Tos

How to articles discuss particular common problems/tasks and their solutions. For now, we only have one:

1. Installing libbson/libmongoc from Source: `/howto/source-install`
   - This is distinct from `/learn/get/from-source`, as it addresses how to select a destination and method of installing the binaries that have been built. This decision is too nuanced to be addressed in a document that walks through the very basics, so it was placed here where it can be expounded upon without muddying the waters.

# Reference Pages

The majority of our documentation is API reference pages, and clutter the top-level directory. These pages have not been moved or modified, but a new `/ref/` directory is created that may later host our API pages. For now, it only contains one page:

1. `mongo-c-driver` Platform Support: `/ref/platforms`
   - This was previously placed within the initial tutorial, which is an unintuitive location for users that would be searching for it. Due to its nature as just a table of values, it is naturally a reference page. The page has been reformatted and structured to clarify our platform support statements.

# Changed Pages

The top-level `/installing` page has been modified with most content removed, but has been left in place to allow existing links to work. Existing sections covered by this page now refer readers to the appropriate new page.

# Feedback & Request for Comments

Of course, this is a much less structured aspect of the project, so bike-shedding opportunities abound :). I'll welcome bike-shedding on things like URL slugs, since those will be set in stone forever once published (we cannot move page URLs without breaking external links), so they need to be as close to "perfect" as possible. Additionally, suggesting page titles and for adding/removing page interlinking is welcome, but these can always be modified after-the-fact.

**Note:** ~~I have noticed that Sphinx's `dirhtml` builder doesn't play well with EVG uploads to Amazon, as the file server does not resolve to the `index.html` page when one browses to the URL of a directory. I'll investigate the best way to fix this. Until then, you can manually append `/index.html` to URLs that return 403 errors.~~ Fixed.


src/libmongoc/doc/conf.py
```diff
@@ -40,6 +44,10 @@
 # Fedora package builds must work offline - maintain a recent copy here
 intersphinx_mapping = {
     "bson": ("https://www.mongoc.org/libbson/current", "libbson-objects.inv"),
+    # NOTE: Pinned to CMake version hear, but is safe to update as new CMake is released.
+    # This pin version ensures that documentation labels remain stable between builds.
+    # When updating, fix any moved/renamed references as applicable.
+    "cmake": ("https://cmake.org/cmake/help/v3.27", None),
```
eramongodb, line 50:
```suggestion
    "cmake": ("https://cmake.org/cmake/help/latest/", None),
```
Given the mongoc docs track latest, wouldn't it be more consistent (and convenient) for the CMake docs to also track latest?

src/libmongoc/doc/howto/index.rst
```diff
@@ -0,0 +1,15 @@
+##################
+Cookbook & How-Tos
```
eramongodb, line 2: The use of the term "Cookbook" is somewhat confusing. Suggest removing it in favor of just "How-Tos".

src/libmongoc/doc/learn/get/from-source.rst
```diff
@@ -0,0 +1,376 @@
...
+
+.. note::
+
+  In the following examples, `$SOURCE_DIR` refers to the directory in which you
+  will be placing the |mongo-c-driver| source code, and `$VERSION` refers to the
+  version of the C driver that you are. The current version written for this
+  documentation is |version.pre|.
+
+It is **highly recommended** that new users use a stable released version of the
+driver, rather than building from a development branch.
```
eramongodb, line 22: Suggest using an "important" admonition here and moving it up above the note for priority.

Similarly, recommend use of the "important" admonition for other "highly recommended" remarks.

src/libmongoc/doc/learn/get/from-source.rst
```diff
@@ -0,0 +1,376 @@
...
+Both |libmongoc| and |libbson| projects use CMake__ for build configuration.
+
+__ https://cmake.org
+
+.. note::
+
+  It is *highly recommended* -- but not *required* -- that you download the
+  latest stable CMake available for your platform.
+
+.. dropdown:: Getting the Latest CMake …
```
eramongodb, line 131: Other dropdowns have "Aside:" in their title. Do we want one here? (Admittedly I'm of the opinion they're superfluous given the purpose is denoted by the admonition type and would rather they be omitted.)

src/libmongoc/doc/learn/get/from-source.rst
```diff
@@ -0,0 +1,376 @@
...
+  $ cmake --build $BUILD --config RelWithDebInfo --parallel
+
+If configured properly and all dependencies are satisfied, then the above
+command should proceed to compile and link the configured components. If the
+above command fails, then there is likely an error with your environment, or you
+are using an unsupported/untested platform. Refer to the build tool output for
+more information.
+
+.. dropdown:: Aside: The ``--config`` option
+  :class-container: admonition hint
```
eramongodb, line 269: Either "note" or "tip" seem more appropriate for these dropdowns than "hint". Consider changing the hint admonitions to either "note" (for extra details) or "tip" (for non-important suggestions).

src/libmongoc/doc/learn/index.rst
```diff
@@ -0,0 +1,11 @@
+###########################
+Getting Started & Tutorials
```
eramongodb, line 2:
```suggestion
Tutorials
```
Suggest keeping titles simple and consistent with TDS terminology ("Tutorials", "How-To Guides", "Explanation", "Reference").

src/libmongoc/doc/visual-studio-guide.rst
```diff
@@ -21,7 +23,9 @@ Right-click on your console application in the Solution Explorer and select "Pro
 .. image::
   static/msvc-add-include-directories.png
 
-(If you chose a different ``CMAKE_INSTALL_PREFIX`` :ref:`when you ran CMake <build-on-windows>`, your include paths will be different.)
+(If you chose a different ``$PREFIX``
+:doc:`when you installed mongo-c-driver </howto/source-install>`, your include
```
eramongodb, line 27: Can/should we link to anchors (`#choosing-a-prefix`) if we can?

src/libmongoc/doc/static/styles.css
```diff
@@ -0,0 +1,110 @@
...
+
+details.sd-dropdown>.sd-card-header {
+  cursor: pointer;
+}
+
+details.sd-dropdown>.sd-card-body.sd-summary-content {
+  cursor: initial;
+}
+
+/* Fix: Furo doesn't intend footnote elements other than 'p' */
```
eramongodb, line 100:
```suggestion
/* Fix: Furo doesn't support footnote elements other than 'p' */
```

eramongodb (changes requested):
The `src/libmongoc/doc/tutorial.rst` page needs the `.. contents::` removed to avoid an `ERROR: Adding a table of contents in Furo-based documentation is unnecessary, ...` message. No other files seem to have this issue.

build/cmake/SphinxBuild.cmake
```diff
@@ -24,10 +24,21 @@ function (sphinx_build_html target_name doc_dir)
...
+   set(is_evg_docs_build "$ENV{EVG_DOCS_BUILD}")
+   if(NOT is_evg_docs_build)
+      # Every .rst builds two corresponding .html files:
+      list (TRANSFORM doc_rsts
+            REPLACE "^(.+)\\.rst$" "html/\\1.html;html/\\1/index.html"
+            OUTPUT_VARIABLE doc_htmls)
+      set (builder dirhtml)
+   else()
+      # We use the regular html builder here. This is only used to support file browsing
+      # were "autoindex" semantics are not supplied by the HTTP server
```
kkloberdanz, line 36:
```suggestion
      # where "autoindex" semantics are not supplied by the HTTP server
```

src/libmongoc/doc/howto/source-install.rst
```diff
@@ -0,0 +1,225 @@
...
+  .. tab-item:: Windows
+
+    On Windows, there exists a dedicated directory for user-local files in
+    ``~\AppData\Local``. To reference it, expand the :batch:`%LocalAppData%`
+    environment variable. (**Do not** use the :batch:`%AppData%` environment
+    variable.)
+
+    Despite this directory existing, it has no prescribed structure that suites
+    our purposes. As with macOS, the choice of user-local installation prefix is
+    arbitrary. This guide *strongly disrecommends* creating additional files and
```
kkloberdanz, line 148:
```suggestion
    arbitrary. This guide *strongly discourages* creating additional files and
```

src/libmongoc/doc/howto/source-install.rst
```diff
@@ -0,0 +1,225 @@
...
+    of manually installed components.
+
+
+.. _source-install.system-prefix:
+
+Selecting a System-Wide Installation Prefix
+===========================================
+
+If you wish to install the |mongo-c-driver| libraries in a directory that is
+visible to all users, there are a few very standard options.
```
kkloberdanz, line 161:
```suggestion
visible to all users, there are a few standard options.
```

src/libmongoc/doc/howto/source-install.rst
```diff
@@ -0,0 +1,225 @@
...
+      unmanaged software installation.
+
+    Alternatively, consider installing to a distinct directory that can be
+    easily removed or relocated, such as ``/opt/mongo-c-driver/``. This will be
+    easily identifiable and not interact with other software on the system
+    without explicitly opting-in.
+
+  .. tab-item:: Windows
+
+    It is **discrecommended** to manually install software system-wide on
```
kkloberdanz, line 184:
```suggestion
    It is **discouraged** to manually install software system-wide on
```

src/libmongoc/doc/howto/source-install.rst
```diff
@@ -0,0 +1,225 @@
...
+    :batch:`%ProgramData%\\mongo-c-driver`)
+
+
+Installing with CMake
+*********************
+
+After you have successfully configured and built the libraries and have selected
+a suitable `$PREFIX`, you can install the built results. Let the name `$BUILD`
+refer to the directory where you executed the build (this is the directory that
+contains ``CMakeCache.txt``, amongst many other files).
```
kkloberdanz, line 200:
```suggestion
contains ``CMakeCache.txt``, among many other files).
```

src/libmongoc/doc/howto/source-install.rst
```diff
@@ -0,0 +1,225 @@
...
+:option:`cmake --install subcommand <cmake.--install>`::
+
+  $ cmake --install "$BUILD" --prefix "$PREFIX"
+
+.. important::
+
+  If you configured the libraries while using a *multi-config generator* (e.g
+  Visual Studio, Xcode), then you will also need to pass the
+  :option:`--config <cmake--install.--config>` command-line option, and must
+  pass the same value for the build configuration which you wish to install.
```
kkloberdanz, line 213:
```suggestion
  pass the same value for the build configuration that you wish to install.
```
"That" is used to indicate a specific object, item, person, condition, etc., while "which" is used to add information to objects, items, people, situations

src/libmongoc/doc/learn/get/from-source.rst
```diff
@@ -0,0 +1,376 @@
...
+
+  .. tab-item:: Using ``wget`` + ``tar``
+
+    ::
+
+      ## Download using wget:
+      $ wget "https://github.com/mongodb/mongo-c-driver/archive/refs/tags/$VERSION.tar.gz" \
+          --output-document="mongo-c-driver-$VERSION.tar.gz"
+      ## Extract using tar:
+      $ tar -x -f "mongo-c-driver-$VERSION.tar.gz"
```
kkloberdanz, line 85:
```suggestion
      $ tar xf "mongo-c-driver-$VERSION.tar.gz"
```

src/libmongoc/doc/learn/get/from-source.rst
```diff
@@ -0,0 +1,376 @@
...
+
+  .. tab-item:: Using ``curl`` + ``tar``
+
+    ::
+
+      ## Using curl:
+      $ curl "https://github.com/mongodb/mongo-c-driver/archive/refs/tags/$VERSION.tar.gz" \
+          --output="mongo-c-driver-$VERSION.tar.gz"
+      ## Extract using tar:
+      $ tar -x -f "mongo-c-driver-$VERSION.tar.gz"
```
kkloberdanz, line 95:
```suggestion
      $ tar xf "mongo-c-driver-$VERSION.tar.gz"
```

kkloberdanz (changes requested):

src/libmongoc/doc/conf.py
vector-of-bool: I would like to use `/latest/`, but my concern is about reference stability. The intersphinx extension downloads an `objects.inv` from the given base URL when Sphinx executes. If we point to a specific version, it is guaranteed that the returned inventory will be stable, while `/latest/` might spuriously break references we have in the docs. I'd like to keep it up-to-date regularly, though, hence the comment.

src/libmongoc/doc/howto/index.rst
vector-of-bool: This one I must have subconsciously borrowed from TDS "recipes" analogy, but I agree that it is a strange term. Will change.

src/libmongoc/doc/learn/get/from-source.rst
vector-of-bool: I was searching for a way to visually distinguish dropdown-admonitions from regular admonitions, but I think I can do better by adding an inset using a CSS margin.

src/libmongoc/doc/learn/get/from-source.rst
vector-of-bool: I chose "hint" purely for the iconography that Furo inserts (a circled question mark "?"), to indicate "Here's the answer to a question that you might have."

src/libmongoc/doc/static/styles.css
vector-of-bool: Should be "indent" :)

src/libmongoc/doc/visual-studio-guide.rst
vector-of-bool: In this case, probably yes.

src/libmongoc/doc/learn/get/from-source.rst
```diff
@@ -0,0 +1,376 @@
...
+Downloading Using Git
+=====================
+
+Using Git, the C driver repository can be cloned from the GitHub URL
+``https://github.com/mongodb/mongo-c-driver.git``. Git tags for released
+versions are named after the version for which they correspond (e.g.
+"|version.pre|"). To clone the repository using the command line, the following
+command may be used::
+
+  $ git clone https://github.com/mongodb/mongo-c-driver.git --branch="$VERSION" "$SOURCE_DIR"
```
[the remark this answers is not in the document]
vector-of-bool: This refers to a tag in this case, but I'll note that `--branch` can also point to tags.

src/libmongoc/doc/learn/get/docs.rst
```diff
@@ -0,0 +1,58 @@
...
+tools for building this documentation. The `poetry install` command only needs
+to be run when the `pyprojct.toml` file is changed.
+
+
+Running Sphinx
+**************
+
+Poetry can be used to execute the `sphinx-build` command::
+
+  $ poetry run sphinx-build -b dirhtml "./src/libmongoc/doc" "./_build/docs/html"
```
eramongodb, line 41: Suggest adding a tip to include the `-j` flag in the `sphinx-build` command for build parallelism.

src/libmongoc/doc/howto/source-install.rst
```diff
@@ -0,0 +1,240 @@
...
+
+    Using an install `$PREFIX` of ``/usr/local/`` is the primary recommendation
+    for all Unix platforms, but this may vary on some obscure systems.
+
+    .. important::
+
+      **DO NOT** use ``/usr/`` nor ``/`` (the root directory) as a prefix: These
+      directories are designed to be carefully managed by the system. The
+      ``/usr/local`` directory is intentionally reserved for the purpose of
+      unmanaged software installation.
```
eramongodb, line 185: I think this paragraph + the "strongly discouraged" paragraph below for Windows deserve to be marked as "Warning" admonitions.

eramongodb (approved):
Some minor suggestions; otherwise, LGTM.

kkloberdanz (approved):

src/libmongoc/doc/learn/get/from-source.rst
```diff
@@ -0,0 +1,386 @@
...
+
+.. _learn.get.install:
+
+Installing the Built Results
+****************************
+
+Let `$PREFIX` be the path `$SOURCE/_install`. We can use CMake to install the
+built results::
+
+  $ cmake --install $BUILD --prefix "$PREFIX" --config RelWithDebInfo
```
kevinAlbs, line 296: Can/should `cmake --build $BUILD --target install` be used instead of `cmake --install`? That seems to bypass the need to build separately.

src/libmongoc/doc/learn/get/from-source.rst
```diff
@@ -0,0 +1,386 @@
...
+
+With the source directory for |mongo-c-driver| at `$SOURCE` and build directory
+`$BUILD`, the following command can be executed from a command-line to configure
+the project with both |libbson| and |libmongoc|::
+
+  $ cmake -S $SOURCE -B $BUILD \
+    -D ENABLE_EXTRA_ALIGNMENT=FALSE \
+    -D ENABLE_AUTOMATIC_INIT_AND_CLEANUP=FALSE \
+    -D CMAKE_BUILD_TYPE=RelWithDebInfo \
+    -D ENABLE_MONGOC=FALSE
```
kevinAlbs, line 226:
```suggestion
    -D ENABLE_MONGOC=OFF
```

In 1.24.3, this results in an error:
```
CMake Error at CMakeLists.txt:162 (message):
  ENABLE_MONGOC option must be ON or OFF
```
I expect this will not be an error in 1.25.0. Building older versions may encounter this error. Suggestion applies to other references of `ENABLE_MONGOC`.

src/libmongoc/doc/learn/get/from-source.rst
```diff
@@ -0,0 +1,386 @@
...
+
+.. tab-set::
+
+  .. tab-item:: Using ``wget`` + ``tar``
+
+    ::
+
+      ## Download using wget:
+      $ wget "https://github.com/mongodb/mongo-c-driver/archive/refs/tags/$VERSION.tar.gz" \
+          --output-document="mongo-c-driver-$VERSION.tar.gz"
```
kevinAlbs, line 93: The source archive from GitHub does not include the `VERSION_CURRENT` file. 

Running the following:
```sh
VERSION=1.24.3                      
wget "https://github.com/mongodb/mongo-c-driver/archive/refs/tags/$VERSION.tar.gz" \
    --output-document="mongo-c-driver-$VERSION.tar.gz"
tar xf mongo-c-driver-$VERSION.tar.gz
cd mongo-c-driver-$VERSION
cmake -S=. -B=cmake-build \
    -D ENABLE_EXTRA_ALIGNMENT=FALSE \
    -D ENABLE_AUTOMATIC_INIT_AND_CLEANUP=FALSE \
    -D CMAKE_BUILD_TYPE=RelWithDebInfo \
    -D ENABLE_MONGOC=OFF
```

Results in this output from CMake:
```
-- calculated BUILD_VERSION 0.0.0
```

This may result in consumers getting an error when attempting to use the build with version 0.0.0.
Consider adding `-D BUILD_VERSION=$VERSION` to the CMake step.

src/libmongoc/doc/learn/get/from-source.rst
kevinAlbs, line 1: IMO the tutorial includes a lot of helpful information that may be better placed in the "How-to" guide.
Consider making the tutorial an opinionated recipe to follow.

Here is a sketch of an idea:

On Linux or macOS, the following may be used to install libbson version 1.24.3 from source into ``$HOME/.local``:

```bash
VERSION=1.24.3
PREFIX=$HOME/.local
git clone git@github.com:mongodb/mongo-c-driver.git --branch $VERSION
cd mongo-c-driver
cmake -S=. -B=cmake-build \
    -D ENABLE_EXTRA_ALIGNMENT=FALSE \
    -D ENABLE_AUTOMATIC_INIT_AND_CLEANUP=FALSE \
    -D CMAKE_BUILD_TYPE=RelWithDebInfo \
    -D ENABLE_MONGOC=OFF
cmake --build cmake-build --target install --config RelWithDebInfo
```

kevinAlbs (commented):
Exciting. I suggest simplifying the tutorial as much as possible.

src/libmongoc/doc/learn/get/from-source.rst
vector-of-bool: The `--install` CMake subcommand was added to replace invoking an "install" target, since the `install` target is not defined on all generators and `--install` gives more fine-grained controls.

src/libmongoc/doc/learn/get/from-source.rst
vector-of-bool: There's a conflict between "too much information" and "too little information." I decided to hide some "extra" information within dropdowns so that they aren't directly in the user's line-of-sight unless they actively choose to see them. I do intend to write a longer how-to for additional configuration options, but there's not enough material here to yet fill such a page. I think some existing additional information can be hidden behind dropdowns, as well as some information hidden behind tabs (e.g. `git` versus downloading an archive).

PR #1372, opened 2023-08-03
Fix detection of ASN1_STRING_get0_data

eramongodb:
Fixes compilation failures on EVG due to unexpected use of the deprecated `ASN1_STRING_data` function triggering warnings. Verified by [this patch](https://spruce.mongodb.com/version/64cc1917c9ec4458a5c8fc16).

The CMake variable `HAVE_ASN1_STRING_GET0_DATA` was being incorrectly set using `check_library_exists`, whose purpose is to check if a _library_ containing the specified _function_ (ABI) exists. This was being done even before the actual SSL library to be linked with was determined (setting of `SSL_LIBRARIES`). Potential problems with checking for the presence of a function are also noted in documentation for [CheckFunctionExists](https://cmake.org/cmake/help/latest/module/CheckFunctionExists.html).

This PR fixes `ASN1_STRING_get0_data` detection by using the correct `check_symbol_exists` macro _after_ the library to link with has been identified (`check_symbol_exists` also checks if linking is successful given `CMAKE_REQUIRED_LIBRARIES`).

### DIFF
--- a/src/libmongoc/CMakeLists.txt
+++ b/src/libmongoc/CMakeLists.txt
@@ -4,8 +4,12 @@ project (libmongoc C)
 
 include (MongoC-Warnings)
 
-include (InstallRequiredSystemLibraries)
+include (CheckSchedGetCPU)
 include (CheckStructHasMember)
+include (CheckSymbolExists)
+include (CheckTypeSize)
+include (CMakePushCheckState)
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
Fix to the `openssl-static-*` tasks is much appreciated. LGTM!

PR #1374, opened 2023-08-08
CDRIVER-4682 log informational message for nongenuine hosts

jmikola:
https://jira.mongodb.org/browse/CDRIVER-4682

The log messages are specified in [DRIVERS-2583](https://jira.mongodb.org/browse/DRIVERS-2583).

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-topology.c
+++ b/src/libmongoc/src/mongoc/mongoc-topology.c
@@ -312,22 +312,38 @@ _mongoc_apply_srv_max_hosts (const mongoc_host_list_t *hl,
    return hl_array;
 }
 
-static bool
-_is_nongenuine_host (const char *host)
+static void
+_detect_nongenuine_hosts (const mongoc_host_list_t *hosts)
 {
-   char *host_lowercase = bson_strdup (host);
+   const mongoc_host_list_t *iter;
 
-   mongoc_lowercase (host, host_lowercase);
+   LL_FOREACH (hosts, iter)
+   {
+      char *host_lowercase = bson_strdup (iter->host);
 
-   if (mongoc_ends_with (host_lowercase, ".cosmos.azure.com") ||
-       mongoc_ends_with (host_lowercase, ".docdb.amazonaws.com") ||
-       mongoc_ends_with (host_lowercase, ".docdb-elastic.amazonaws.com")) {
+      mongoc_lowercase (iter->host, host_lowercase);
+
+      if (mongoc_ends_with (host_lowercase, ".cosmos.azure.com")) {
+         MONGOC_INFO (
+            "You appear to be connected to a CosmosDB cluster. For more "
+            "information regarding feature compatibility and support please "
+            "visit https://www.mongodb.com/supportability/cosmosdb");
+         bson_free (host_lowercase);
+         return;
+      }
+
+      if (mongoc_ends_with (host_lowercase, ".docdb.amazonaws.com") ||
+          mongoc_ends_with (host_lowercase, ".docdb-elastic.amazonaws.com")) {
+         MONGOC_INFO (
+            "You appear to be connected to a DoumentDB cluster. For more "
+            "information regarding feature compatibility and support please "
+            "visit https://www.mongodb.com/supportability/documentdb");
+         bson_free (host_lowercase);
+         return;
+      }
+
       bson_free (host_lowercase);
-      return true;
    }
-
-   bson_free (host_lowercase);
-   return false;
 }
 
 /*
@@ -356,7 +372,6 @@ mongoc_topology_new (const mongoc_uri_t *uri, bool single_threaded)
    mongoc_rr_data_t rr_data;
    bool has_directconnection;
    bool directconnection;
-   bool found_nongenuine_host = false;
 
    BSON_ASSERT (uri);
 
@@ -621,13 +636,9 @@ mongoc_topology_new (const mongoc_uri_t *uri, bool single_threaded)
 
       mongoc_topology_description_add_server (td, elem->host_and_port, &id);
       mongoc_topology_scanner_add (topology->scanner, elem, id, false);
-
-      if (!found_nongenuine_host &&
-          (found_nongenuine_host = _is_nongenuine_host (elem->host))) {
-         /* TODO: the final copy for this log message is TBD */
-         MONGOC_INFO ("Nongenuine host detected");
-      }
    }
+
+   _detect_nongenuine_hosts (*hl_array);
 
    bson_free ((void *) hl_array);
 
--- a/src/libmongoc/src/mongoc/mongoc-topology.c
+++ b/src/libmongoc/src/mongoc/mongoc-topology.c
@@ -312,6 +312,24 @@ _mongoc_apply_srv_max_hosts (const mongoc_host_list_t *hl,
    return hl_array;
 }
 
+static bool
+_is_nongenuine_host (const char *host)
+{
+   char *host_lowercase = bson_strdup (host);
+
+   mongoc_lowercase (host, host_lowercase);
+
+   if (mongoc_ends_with (host_lowercase, ".cosmos.azure.com") ||
+       mongoc_ends_with (host_lowercase, ".docdb.amazonaws.com") ||
+       mongoc_ends_with (host_lowercase, ".docdb-elastic.amazonaws.com")) {
+      bson_free (host_lowercase);
+      return true;
+   }
+
+   bson_free (host_lowercase);
+   return false;
+}
+
 /*
  *-------------------------------------------------------------------------
  *
@@ -338,6 +356,7 @@ mongoc_topology_new (const mongoc_uri_t *uri, bool single_threaded)
    mongoc_rr_data_t rr_data;
    bool has_directconnection;
    bool directconnection;
+   bool found_nongenuine_host = false;
 
    BSON_ASSERT (uri);
 
@@ -602,6 +621,11 @@ mongoc_topology_new (const mongoc_uri_t *uri, bool single_threaded)
 
       mongoc_topology_description_add_server (td, elem->host_and_port, &id);
       mongoc_topology_scanner_add (topology->scanner, elem, id, false);
+
+      if (!found_nongenuine_host &&
+          (found_nongenuine_host = _is_nongenuine_host (elem->host))) {
+         MONGOC_INFO ("Nongenuine host detected");
+      }
    }
 
    bson_free ((void *) hl_array);
--- a/src/libmongoc/src/mongoc/mongoc-topology.c
+++ b/src/libmongoc/src/mongoc/mongoc-topology.c
@@ -335,7 +335,7 @@ _detect_nongenuine_hosts (const mongoc_host_list_t *hosts)
       if (mongoc_ends_with (host_lowercase, ".docdb.amazonaws.com") ||
           mongoc_ends_with (host_lowercase, ".docdb-elastic.amazonaws.com")) {
          MONGOC_INFO (
-            "You appear to be connected to a DoumentDB cluster. For more "
+            "You appear to be connected to a DocumentDB cluster. For more "
             "information regarding feature compatibility and support please "
             "visit https://www.mongodb.com/supportability/documentdb");
          bson_free (host_lowercase);
--- a/src/libmongoc/src/mongoc/mongoc-topology.c
+++ b/src/libmongoc/src/mongoc/mongoc-topology.c
@@ -624,6 +624,7 @@ mongoc_topology_new (const mongoc_uri_t *uri, bool single_threaded)
 
       if (!found_nongenuine_host &&
           (found_nongenuine_host = _is_nongenuine_host (elem->host))) {
+         /* TODO: the final copy for this log message is TBD */
          MONGOC_INFO ("Nongenuine host detected");
       }
    }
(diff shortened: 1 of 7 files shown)


src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -1526,6 +1526,23 @@ mongoc_uri_finalize_directconnection (mongoc_uri_t *uri, bson_error_t *error)
    return true;
 }
 
+static void
+mongoc_uri_check_for_genuine_hosts (const char *hosts)
+{
+   char *hosts_lowercase = lowercase_str_new (hosts);
+
+   if (strstr (hosts_lowercase, ".cosmos.azure.com")) {
+      MONGOC_INFO ("Azure Cosmos DB is not genuine");
```
jmikola, line 1535: Note: these messages are not finalized. My understanding is that we're waiting on URLs to include here. I believe @alexbevi is our point of contact on that.

eramongodb, line 1535:
```suggestion
      // CDRIVER-4682: MONGOC_INFO content is TBD.
```
Suggest refraining from emitting a message yet; instead, leave placeholders for now until DRIVERS-2583 has been finalized.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -1526,6 +1526,23 @@ mongoc_uri_finalize_directconnection (mongoc_uri_t *uri, bson_error_t *error)
    return true;
 }
 
+static void
+mongoc_uri_check_for_genuine_hosts (const char *hosts)
+{
+   char *hosts_lowercase = lowercase_str_new (hosts);
```
jmikola, line 1532: In https://github.com/mongodb-js/mongodb-build-info/commit/52f4b0d73eaeeecd85598c07e89033fb3f5b1f16, the shell uses an anchored regex with case-insensitive match on _each_ host. I opted to keep things simple by searching the entire host string, but kept the case-insensitive behavior.

I briefly looked into [`strcasestr`](https://linux.die.net/man/3/strcasestr) but found not prior art in libmongoc for that. It looks like it may be non-portable, and I've had my fill of that after the `strerror_r()` debacle.

There a notably no tests for a case-insensitive match, but I'm happy to add some if requested. The PR currently uses the same fixtures from the shell (where case-insensitive matching also seems to be untested).

src/libmongoc/src/mongoc/mongoc-uri.c
alexbevi: Correct. We have WEBSITE-15470 tracking that for now.

src/libmongoc/tests/test-mongoc-uri.c
```diff
@@ -2740,6 +2740,72 @@ test_casing_options (void)
...
 }
 
+
+static void
+test_check_for_genuine_hosts (void)
+{
+   mongoc_uri_t *uri;
+   int i;
+
+   const char *cosmos_hosts[] = {
```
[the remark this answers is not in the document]
alexbevi: DocumentDB only shows [`mongodb://` scheme connections](https://docs.aws.amazon.com/documentdb/latest/developerguide/connect_programmatically.html), and StrongDM (I don't know the product) documents [lack of SRV support as a DocumentDB limitation](https://www.strongdm.com/docs/admin/resources/datasources/amazon-documentdb-replica/#limitations)

src/libmongoc/src/mongoc/mongoc-uri.c
jmikola: I wasn't planning to merge this until we had final copy. Replacing `MONGOC_INFO` with a comment would render the tests irrelevant.

src/libmongoc/tests/test-libmongoc.c
```diff
@@ -206,7 +206,7 @@ log_handler (mongoc_log_level_t log_level,
 
    suite = (TestSuite *) user_data;
 
-   if (log_level < MONGOC_LOG_LEVEL_INFO) {
+   if (log_level <= MONGOC_LOG_LEVEL_INFO) {
```
jmikola, line 209: I'm not sure why `INFO` messages were previously ignored, but this was necessary for the tests. Had a hell of a time figuring out why `ASSERT_CAPTURED_LOG` was failing before I found this.

Let me know if it should be called out in the commit message.

src/libmongoc/src/mongoc/mongoc-uri.c
jmikola: Thanks for pointing out `ends_with`. I missed that previously.

It wasn't clear from [DRIVERS-2583](https://jira.mongodb.org/browse/DRIVERS-2583) whether we need to log messages for _each_ host or just once. I'm inclined to say that once is enough to avoid log pollution -- and also because it seems unlikely you'd ever have a mixed cluster (i.e. all hosts will either be genuine or not). And a single string match across the entire host/port component of the URI seemed OK to limit overhead and ensure at most one log message. As for matching within the string vs. using `ends_with`, I also think it's unlikely that one of these substrings would appear in a genuine host (e.g. `mydb.cosmos.azure.com.mydomain.com`) so I wasn't worried about that.

Furthermore, I think it's reasonable to just check the initial SRV host string, knowing that a Cosmos connection string is going to resolve to more Cosmos hosts. If someone is going out of their way to configure SRV with their own domain and point it to a Cosmos or DocumentDB, I'd argue they're probably not the intended audience for this log message.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -1551,6 +1568,8 @@ mongoc_uri_parse_before_slash (mongoc_uri_t *uri,
       hosts = before_slash;
    }
 
+   mongoc_uri_check_for_genuine_hosts (hosts);
```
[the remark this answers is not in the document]
jmikola: Related to my response in https://github.com/mongodb/mongo-c-driver/pull/1374#discussion_r1291326865, [DRIVERS-2583](https://jira.mongodb.org/browse/DRIVERS-2583) is unclear about all of the edge cases that must be considered.

@alexbevi: Can you provide some guidance here?

 * Is it sufficient to just match against the initial connection string or should we defer this logic until _after_ all hosts have been parsed (and SRV  initially resolved)?
 * If we defer and check hosts individually, do you agree it only makes sense to log a single message to avoid polluting the logs?
 * Furthermore, can we avoid subsequent checks after SRV polling?

src/libmongoc/src/mongoc/mongoc-uri.c
jmikola: Resolving this, as I'm now using `mongoc_ends_with()` to do a rear-anchored match.

src/libmongoc/src/mongoc/mongoc-uri.c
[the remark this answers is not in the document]
jmikola: Pushed changes. Detection is now done in `mongoc_topology_new()` and will take place _after_ the initial SRV lookup. It's not feasible to mock initial SRV responses in libmongoc, so that is intentionally left untested.

I also switched to using a rear-anchored host name match (consistent with mongosh) and added test coverage for that.

And we only log a single message when the first nongenuine host is detected. That behavior is unchanged from the first draft of this PR.

src/libmongoc/tests/test-mongoc-uri.c
```diff
@@ -2740,6 +2740,72 @@ test_casing_options (void)
...
+      "mongodb://"
+      "x:y@docdb-2001-01-01-01-01-01.cluster-abc.eu-central-1.docdb.amazonaws."
+      "com:27017/"
+      "?replicaSet=rs0&readPreference=secondaryPreferred&retryWrites=false",
+      "mongodb://"
+      "x:y@elastic-docdb-123456789.eu-central-1.docdb-elastic.amazonaws.com:"
+      "27017",
+   };
+
+   const char *genuine_hosts[] = {
```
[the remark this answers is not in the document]
jmikola: I revised these, as the auth and query string components really have no impact on the test. The new connection strings now fit on a single line after running clang-format.

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -602,6 +621,12 @@ mongoc_topology_new (const mongoc_uri_t *uri, bool single_threaded)
 
       mongoc_topology_description_add_server (td, elem->host_and_port, &id);
       mongoc_topology_scanner_add (topology->scanner, elem, id, false);
+
+      if (!found_nongenuine_host &&
+          (found_nongenuine_host = _is_nongenuine_host (elem->host))) {
+         /* TODO: the final copy for this log message is TBD */
+         MONGOC_INFO ("Nongenuine host detected");
```
jmikola, line 628: @alexbevi: I just noticed that [WEBSITE-15470](https://jira.mongodb.org/browse/WEBSITE-15470) mentions separate pages for Cosmos and DocumentDB. Do I assume correctly that we'll need to log distinct messages for each of those platforms?

That was present in my original draft of this PR but I ended up removing it when moving the check to `mongoc_topology_new()` (for no particular reason other than it made the implementation simpler). I can restore separate log messages if needed, though.

alexbevi: @jmikola that is correct (distinct URLs will be shown based on host detection)

alexbevi: * CosmosDB: https://www.mongodb.com/supportability/cosmosdb
* DocumentDB: https://www.mongodb.com/supportability/documentdb

jmikola: These URLs are now incorporated into the PR. I opted to omit the trailing period from the log message to avoid users copy-pasting the URL with a period and having it 404.

src/libmongoc/src/mongoc/mongoc-uri.c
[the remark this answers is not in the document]
jmikola: Revised the PR to log distinct messages for Cosmos and DocumentDB.

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -604,6 +638,8 @@ mongoc_topology_new (const mongoc_uri_t *uri, bool single_threaded)
       mongoc_topology_scanner_add (topology->scanner, elem, id, false);
    }
 
+   _detect_nongenuine_hosts (*hl_array);
```
jmikola, line 641: I realize this is an additional iteration (after `srvMaxHosts` is applied when applicable), but the resulting code is much cleaner than the previous commit where I incorporated the logic to log at-most-once during the preceding loop.

That seems preferable given that `mongoc_topology_new` is already quite complex and the only overhead here is following a handful of `mongoc_host_list_t` pointers.

src/libmongoc/tests/test-mongoc-topology.c
```diff
@@ -2675,6 +2675,85 @@ test_failure_to_setup_after_retry (void)
...
+   for (i = 0; i < sizeof (docdb_uris) / sizeof (const char *); i++) {
+      capture_logs (true);
+      uri = mongoc_uri_new (docdb_uris[i]);
+      ASSERT (uri);
+      topology = mongoc_topology_new (uri, true);
+      ASSERT (topology);
+      ASSERT_CAPTURED_LOG (
+         "nongenuine host should log",
+         MONGOC_LOG_LEVEL_INFO,
+         "You appear to be connected to a DoumentDB cluster. For more "
```
eramongodb, line 2738:
```suggestion
         "You appear to be connected to a DocumentDB cluster. For more "
```

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -312,6 +312,40 @@ _mongoc_apply_srv_max_hosts (const mongoc_host_list_t *hl,
...
+            "information regarding feature compatibility and support please "
+            "visit https://www.mongodb.com/supportability/cosmosdb");
+         bson_free (host_lowercase);
+         return;
+      }
+
+      if (mongoc_ends_with (host_lowercase, ".docdb.amazonaws.com") ||
+          mongoc_ends_with (host_lowercase, ".docdb-elastic.amazonaws.com")) {
+         MONGOC_INFO (
+            "You appear to be connected to a DoumentDB cluster. For more "
```
eramongodb, line 338:
```suggestion
            "You appear to be connected to a DocumentDB cluster. For more "
```

src/libmongoc/src/mongoc/mongoc-util.c
```diff
@@ -546,6 +546,39 @@ bson_copy_to_including_noinit (const bson_t *src,
...
+   s2 = suffix + suffix_len;
+
+   /* until either pointer reaches start of its string, compare the pointers */
+   for (; s1 >= str && s2 >= suffix; s1--, s2--) {
+      if (*s1 != *s2) {
+         return false;
+      }
+   }
+
+   return true;
```
eramongodb, line 580:
```suggestion
   BSON_ASSERT_PARAM (str);
   BSON_ASSERT_PARAM (suffix);

   const size_t str_len = strlen (str);
   const size_t suffix_len = strlen (suffix);

   if (str_len < suffix_len) {
      return false;
   }

   return strcmp (str + (str_len - suffix_len), suffix) == 0;
```
Simplification + precondition assertions.

eramongodb (approved):
Some minor suggestions; otherwise, LGTM.

PR #1375, opened 2023-08-10
WIP: CDRIVER-3514 Sharing SCRAM cache among all clients in pool

joshbsiegel:
# Summary
This PR will modify `mongoc_client_t` architecture so that the SCRAM cache will be shared among all of the clients in the pool. Unfortunately, my internship ended before I was able to finish this ticket, so the details below are designed to get the next assignee up to speed.

# Background
The PERF team pointed out a potential bottleneck with SCRAM authentication among clients in the C driver: https://jira.mongodb.org/browse/PERF-4166. We verified this by creating a simple program and generating our own FlameGraph: https://jira.mongodb.org/browse/CDRIVER-3514. 

# Proposed Solution
Currently, the SCRAM cache is created and managed by `mongoc_cluster_t` and every `mongoc_client_t` has a `mongoc_cluster_t`. We believe the best way to improve SCRAM authentication time is to share this SCRAM cache among all `mongoc_client_t`s through `mongoc_topology_t`. All `mongoc_client_t`s have a `mongoc_topology_t` attached to them, and if the client is in a `mongoc_pool_t`, then all of the clients in the pool have the same topology. Moving the SCRAM cache to `mongoc_topology_t` will provide all clients in the pool access. Here are the steps of the new proposed architecture:
1. A `mongoc_client_t` exists or gets popped from a pool. It wants initialize its SCRAM cache.
2. The client will check the `mongoc_topology_t` attached to it to see if it already has a copy of the SCRAM cache.
3. If it does not, it will generate a SCRAM cache on the topology. If it does, do nothing.
4. The client will copy the topology's "main" copy of the SCRAM cache so that it can have a local copy to read from without worrying about thread safety.

# Current Changes
I added a new SCRAM cache `mongoc_scram_cache_v2_t` for a quick POC. As mentioned above, this struct will be used in two places, `mongoc_client_t` and `mongoc_topology_t`. Due to multiple threads potentially accessing the topology SCRAM cache at once, I also added a `bson_shared_mutex_t` (RW lock) on `mongoc_topology_t`. 

There are two new functions in `mongoc_client_t`, `_mongoc_scram_cache_destroy_v2` and `_mongoc_scram_get_cache_v2`. The latter is thread-safe as it will potentially modify the SCRAM cache in topology. Please note that at the moment we are generating a fake cache.

I added two tests as a POC that the cache will be shared correctly. The first will setup 2 clients on the same pool, assert that they have the same topology, and assert that they have the same "iteration" of cache (should be iteration 0 because it's only generated once). The second test will setup a pool and threads to pop from the pool, get the cache, and then push back onto the pool. Based on print statements I added, it's evident that the cache is only getting generated once.

# What's Next
-  Start generating a real cache in `_mongoc_scram_get_cache_v2`. Admittedly, I did not have time to do research on how I think this can be done. Right now, all of the SCRAM authentication is done in `mongoc_cluster_t` so I would suggest starting there and most likely moving things over to the topology file. There is also some SCRAM cache activity happening in `mongoc-scram` and I think there's the potential for consolidation. I would think this is going to be the most time consuming part of this ticket.
- Test the changes to see if SCRAM authentication is any faster. The information on how to generate a FlameGraph is in the JIRA ticket.
- If this solution proves to work, phase out the old SCRAM cache architecture.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-client-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-client-private.h
@@ -241,19 +241,6 @@ mongoc_client_uses_server_api (const mongoc_client_t *client);
 bool
 mongoc_client_uses_loadbalanced (const mongoc_client_t *client);
 
-// The workflow is designed to be as follows:
-// 1. Client has it's own copy of the scram cache. It wants to get the "main"
-// copy from topology.
-// 2. The client calls get_cache_v2
-// 3. If the cache isn't set on the topology, the client will set the cache on
-// its topology and then copy it over to its own copy of the scram cache.
-// Ideally this will only happen once, although with multithreading we cannot be
-// certain of this. This can be changes with more locking in get_cache_v2, but
-// we opted to try to do as minimal of locking as possible. As long as it's not
-// generating the cache everytime, we should theoretically save time.
-// 4. If the cache is set on the topology, simply copy it over to the client's
-// copy of the scram cache.
-
 void
 _mongoc_scram_cache_destroy_v2 (mongoc_scram_cache_v2_t *cache);
 
--- a/src/libmongoc/src/mongoc/mongoc-client-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-client-private.h
@@ -118,6 +118,8 @@ struct _mongoc_client_t {
    unsigned int csid_rand_seed;
 
    uint32_t generation;
+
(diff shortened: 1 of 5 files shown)


Neustradamus:
@joshbsiegel: Nice :)

Linked to:
- https://github.com/scram-sasl/info/issues/1

kevinAlbs:
Will be picked up again as part of CDRIVER-3514

PR #1380, opened 2023-08-14
Refactor uninstall-script generation

vector-of-bool:
This aside change refactors and changes a few minor issues with the uninstall script:

1. The script did not handle a late-defined install-prefix, and entered an infinite loop in that case
2. The script handled `DESTDIR` strangely (?)
3. There were two different generators: for Windows and for Unix
4. Using .bat scripting *at all* is just fraught with peril and should be avoided, if at all possible

This change also removes the top-level `generate_uninstall/` directory, as well as the two generator scripts, instead replacing it with a single `GenerateUninstalle.cmaker` module (which *should* also be portable between projects).


build/cmake/GeneratePkgConfig.cmake
rcsanchez97, line 1: I am assuming that the DEBUG messages were meant to be left in. I don't have a problem with them (I rather like the extra output), but I wanted to mention it in the event you intended to remove them and forgot.

build/cmake/GenerateUninstaller.cmake
```diff
@@ -0,0 +1,271 @@
...
+        FILES "${_uninstaller_script}"
+        DESTINATION "${UNINSTALL_PROG_DIR}"
+        PERMISSIONS
+            OWNER_READ OWNER_WRITE OWNER_EXECUTE
+            GROUP_READ GROUP_EXECUTE
+            WORLD_READ WORLD_EXECUTE
+        )
+
+    # If applicable, generate an "uninstall" target to run the uninstaller:
+    if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR OR PROJECT_IS_TOP_LEVEL)
```
rcsanchez97, line 46: Is this how a collision is prevented between the uninstall generated in this project and, for instance, the uninstall generated in the C++ driver when it includes the C driver directly?

build/cmake/GenerateUninstaller.cmake
```diff
@@ -0,0 +1,271 @@
...
+    file(RELATIVE_PATH relpath "${install_prefix}" "${installed}")
+    # Add a removal:
+    add_rmfile("${relpath}")
+    # Climb the path and collect directories:
+    while("1")
+        get_filename_component(installed "${installed}" DIRECTORY)
+        file(TO_NATIVE_PATH "${installed}" installed)
+        get_filename_component(parent "${installed}" DIRECTORY)
+        file(TO_NATIVE_PATH "${parent}" parent)
+        # Don't account for the prefix or directly children of the prefix:
```
rcsanchez97, line 253:
```suggestion
        # Don't account for the prefix or direct children of the prefix:
```

build/cmake/GenerateUninstaller.cmake
```diff
@@ -0,0 +1,271 @@
...
+            break()
+        endif()
+        # Keep track of this directory for later:
+        list(APPEND dirs_to_remove "${installed}")
+    endwhile()
+endforeach()
+
+# Now generate commands to remove (empty) directories:
+list(REMOVE_DUPLICATES dirs_to_remove)
+# Order them in descending so that we remove subdirectories before their parents:
```
rcsanchez97, line 264:
```suggestion
# Order them by depth so that we remove subdirectories before their parents:
```

rcsanchez97 (approved):
I'm a little sad that you didn't like my generator script written in batch :disappointed: 

Seriously, though, this is much nicer. Aside from a few minor tweaks/questions, it looks good to me.

kkloberdanz (approved):

build/cmake/GeneratePkgConfig.cmake
vector-of-bool: Yes, they were useful while debugging and don't get in the way, so I just left them there.

build/cmake/GenerateUninstaller.cmake
vector-of-bool: Yes. This is mostly the same as what was done previously, but with the addition of the newer CMake feature `PROJECT_IS_TOP_LEVEL`, which was added to address this exact use case (on older CMake, that variable check will be a no-op).

PR #1383, opened 2023-08-16
Speedup of calc_release_version.py

vector-of-bool:
This changeset adds some minor stylistic changes and a significant speedup for calc_release_version. Refer to 18b2dded7ade19b5b50ac8e46f99aef93dfe346d for details.


rcsanchez97 (commented):
As I mentioned in a comment to #1382 , the algorithm in this script (which is shared across the C driver, C++ driver, and libmongocrypt) still needs to function on Python 2.6. While I like the changes, I am not sure that these will work on older Python implementations. Do you know if that presents a problem for what you want to do?

PR #1381, opened 2023-08-15
CDRIVER-4691 sync KMS sources with libmongocrypt

rcsanchez97:
This PR also adds an Evergreen task (in the smoke test build variant) to check for divergences of the KMS sources.

Here is a green patch build (i.e., the sources are sync), which represents what will be merged by this PR: https://spruce.mongodb.com/version/64dbbea8d6d80a63b8e50287/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Here is a red patch build (i.e., with a forced divergence to demonstrate a failure of the task): https://spruce.mongodb.com/version/64dbbee8d1fe078cc794d18b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

In the failing case, the task outputs a diff before bailing out, like this:

```
[2023/08/15 18:09:39.261] diff -uNr ../libmongocrypt/kms-message/CMakeLists.txt src/kms-message/CMakeLists.txt
[2023/08/15 18:09:39.261] --- ../libmongocrypt/kms-message/CMakeLists.txt	2023-08-15 18:09:39.208186064 +0000
[2023/08/15 18:09:39.261] +++ src/kms-message/CMakeLists.txt	2023-08-15 18:09:38.404189072 +0000
[2023/08/15 18:09:39.261] @@ -1,4 +1,5 @@
[2023/08/15 18:09:39.261]  cmake_minimum_required (VERSION 3.5)
[2023/08/15 18:09:39.261] +# This commend should trigger a failure in the kms-divergence-check task.
[2023/08/15 18:09:39.261]  project (kms_message
[2023/08/15 18:09:39.261]     VERSION 0.0.1
[2023/08/15 18:09:39.261]     LANGUAGES C
```

@kevinAlbs also, should the release docs have a step to update the `--branch master` option in `.evergreen/config_generator/components/kms_divergence_check.py` and regenerate the Evergreen config? For instance, I am assuming that when C driver 1.25.0 is released, we will want to "lock" the KMS source sync check on the libmongocrypt 1.9 branch. Otherwise, stable C driver releases will be pointing at the `master` branch of libmongocrypt and we would have to introduce changes from libmongocrypt `master` to the C driver release branch in order to keep the task green.

PR #1385, opened 2023-08-18
Fix ValueError error during config generation

eramongodb:
Followup to https://github.com/mongodb/mongo-c-driver/pull/1381. Addresses this very unhelpful error during EVG config generation:

```
  File "<string>", line 1, in <module>
  File ".evergreen/config_generator/generate.py", line 34, in main
    m.generate()
  File ".evergreen/config_generator/generators/functions.py", line 13, in generate
    functions.update(component.functions())
ValueError: dictionary update sequence element #0 has length 1; 2 is required
```

### DIFF
--- a/.evergreen/config_generator/components/kms_divergence_check.py
+++ b/.evergreen/config_generator/components/kms_divergence_check.py
@@ -22,7 +22,7 @@ class KmsDivergenceCheck(Function):
 
 
 def functions():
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
Thank you for the fix. LGTM.

Aside: I do not see this error when running `./tools/poetry.sh run mc-evg-generate`. Did this error occur running the script a different way?

PR #1387, opened 2023-08-24
CDRIVER-4673 fix build of libmongoc using system libbson

kevinAlbs:
# Summary

- Move `bson-dsl.h` to `src/common`.
- Do not define `mongo::detail::c_platform` if using system libbson.
- Add regression test to ensure libmongoc can build with system install of libbson.

# Background & Motivation

This PR resolves two observed issues when building libmongoc using a system install of libbson.

On the `r1.24` branch, a compile error referenced in CDRIVER-4673 occurs: `fatal error: 'bson/bson-dsl.h' file not found`. `bson-dsl.h` is not installed with libbson, but `bson-dsl.h` is needed to compile libmongoc. 16d623786b5c5d5f731ab264fb8117e27312025e moves `bson-dsl.h` to `src/common`. This is applied to r1.24 in https://github.com/mongodb/mongo-c-driver/pull/1388

On `master`, there is an additional error reported in cmake. Running `./.evergreen/scripts/install-libmongoc-after-libbson.sh` results in:

```
-- Searching for libbson CMake packages
CMake Error at install-dir/lib/cmake/bson-1.0/bson-targets.cmake:42 (message):
  Some (but not all) targets in this export set were already defined.

  Targets Defined: mongo::detail::c_platform

  Targets not yet defined: mongo::bson_shared, mongo::bson_static

  install-dir/lib/cmake/bson-1.0/bson-1.0-config.cmake:4 (include)
  CMakeLists.txt:329 (find_package)
```
88830176250cefab22be7fc60fe4281ebb3982c9 will not define `c_platform` if using a system libbson install to fix this error.


vector-of-bool (approved):
LGTM!

PR #1401, opened 2023-09-13
CDRIVER-4682 Detect non-MongoDB host names before SRV lookup.

kevinAlbs:
# Summary

- Detect non-MongoDB host names before SRV lookup.

# Background & Motivation

CDRIVER-4682 requires the driver to detect connection to a non-MongoDB server and log a message.

CDRIVER-4682 was initially implemented to detect after SRV lookup. Requirements were later changed to detect before SRV lookup:

> If SRV is used, this check should be done on the SRV's `{hostname}.{domainname}` prior to SRV lookup (ex: xyz.mongodb.net).

As a drive-by improvement, `test_detect_nongenuine_hosts` is updated to print the input URI string on test failure.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-topology.c
+++ b/src/libmongoc/src/mongoc/mongoc-topology.c
@@ -387,6 +387,8 @@ mongoc_topology_new (const mongoc_uri_t *uri, bool single_threaded)
 
    BSON_ASSERT (uri);
 
+   _detect_nongenuine_hosts (uri);
+
 #ifndef MONGOC_ENABLE_CRYPTO
    if (mongoc_uri_get_option_as_bool (
           uri, MONGOC_URI_RETRYWRITES, MONGOC_DEFAULT_RETRYWRITES)) {
@@ -649,8 +651,6 @@ mongoc_topology_new (const mongoc_uri_t *uri, bool single_threaded)
       mongoc_topology_description_add_server (td, elem->host_and_port, &id);
       mongoc_topology_scanner_add (topology->scanner, elem, id, false);
    }
-
-   _detect_nongenuine_hosts (uri);
 
    bson_free ((void *) hl_array);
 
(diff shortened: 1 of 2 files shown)


src/libmongoc/tests/test-mongoc-topology.c
```diff
@@ -2704,6 +2708,13 @@ test_detect_nongenuine_hosts (void)
       "mongodb://a.mongo.cosmos.azure.com.tld:19555/",
       "mongodb://a.docdb.amazonaws.com.tld:27017/",
       "mongodb://a.docdb-elastic.amazonaws.com.tld:27017/",
+      /* Test genuine SRV URIs */
+      "mongodb+srv://a.example.com/",
+      "mongodb+srv://a.mongodb.net/",
+      /* Host names do not end with expected suffix */
```
jmikola, line 2714:
```suggestion
      /* SRV host names do not end with expected suffix */
```

jmikola (approved):

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -312,37 +312,55 @@ _mongoc_apply_srv_max_hosts (const mongoc_host_list_t *hl,
...
-static void
-_detect_nongenuine_hosts (const mongoc_host_list_t *hosts)
+// `_detect_nongenuine_host` logs if the host string suggests use of CosmosDB or
+// DocumentDB. See DRIVERS-2583 for behavior requirements. Returns true if a
+// CosmosDB or DocumentDB host is detected.
+static bool
+_detect_nongenuine_host (const char *host)
 {
-   const mongoc_host_list_t *iter;
+   char *host_lowercase = bson_strdup (host);
```
eramongodb, line 321:
```suggestion
   char *const host_lowercase = bson_strdup (host);
```

eramongodb (approved):

PR #1402, opened 2023-09-14
Fix testing instructions

kkloberdanz:
Previously, following the testing instructions in CONTRIBUTING.md results in the following error:
```
"error":"NonExistentPath: Data directory /data/db not found. Create the missing directory or specify another path using (1) the --dbpath command line option, or (2) by adding the 'storage.dbPath' option in the configuration file."
```
This adds the dbpath to the command line so that way one can copy and paste the instructions in order to run tests.

Also, prefer to use the newer mongosh over mongo.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -135,13 +135,14 @@ To run the entire test suite, including authentication and support for the
 enabled:
 
 ```
-$ mongod --auth --setParameter enableTestCommands=1
+$ mkdir db
+$ mongod --auth --setParameter enableTestCommands=1 --dbpath db/
 ```
 
 In another terminal, use the `mongo` shell to create a user:
 
 ```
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -135,13 +135,14 @@ To run the entire test suite, including authentication and support for the
 enabled:
 
 ```
-$ mongod --auth --setParameter enableTestCommands=1
+$ mkdir db
+$ mongod --auth --setParameter enableTestCommands=1 --dbpath db/
 ```
 
 In another terminal, use the `mongo` shell to create a user:
```
kevinAlbs, line 142:
```suggestion
In another terminal, use the `mongosh` shell to create a user:
```

kevinAlbs (approved):
Thank you. LGTM with a rename.

PR #1405, opened 2023-09-14
Find-SASL cleanup

vector-of-bool:
This is the first of several smaller changes that will go together into a full refactoring of the libmongoc build. This piece targets the SASL support library, with a few aside fixes.

## Changed

- The `FindSASL2.cmake` module, despite having the `Find` name prefix, did not behave as a proper find-module. This refactoring changes the module to behave as a find-module and generates an `IMPORTED` library target against which libmongoc can link. This also resulted in changes to the installed mongoc targets to support the downstream users to link against a libsasl2. Whereas the prior install hard-coded the path to sasl2 within the -targets.cmake file, this change adds a level of indirection `_mongoc::detail::sasl` that is resolved at import-time to the SASL implementation that is available and that matches what the targets were built against.

- The code that decided which SASL to use was tangled, but the actual decision process is not complex. This refactors that process to a more straightforward implementation.

- Additional Earthly targets and tasks have been defined to test the SASL on/off configurations.

- Aside: Use the FeatureSummary module from CMake to print a definitive summary of enabled/disabled features and found/missing packages at the end of the configure process.

- Aside: A `test-cxx-driver` target in Earthly, which will perform a simple build of the C++ driver using the built result of the C driver. This does not have any EVG tasks, it's just for local testing (for now).


build/cmake/FindSASL2.cmake
```diff
@@ -1,54 +1,74 @@
...
+# Now ask pkg-config to find libsasl2 for us
+message(DEBUG "Using pkg-config to search for module “${_modspec}”")
+pkg_check_modules(libsasl2 ${_required} ${_quiet} IMPORTED_TARGET "${_modspec}" GLOBAL)
+# Note: pkg_check_modules() does support static libs, but we don't use them here,
+# and static libsasl2 is uncommon.
+if(NOT libsasl2_FOUND)
+    # The pkg_check_modules() will have issued the error for us if we were REQUIRED.
+    set(SASL2_NOT_FOUND_MESSAGE [[pkg-config failed to find a libsasl2 (Cyrus) dynamic library package]])
+    return()
+endif()
```
kevinAlbs, line 64: Before these changes: libsasl2 is located on macOS. Here is the output of cmake:

```
-- Searching for sasl/sasl.h
--   Found in /Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk/usr/include
-- Searching for libsasl2
--   Found /Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk/usr/lib/libsasl2.tbd
```

After these changes: libsasl2 is not found.

The install of libsasl2 on macOS does not appear to include pkgconfig `.pc` files:

``` 
% find /Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk/ -name "*sasl*.pc"
(no results)
```

The `cyrus-sasl` Homebrew package includes pkg-config files, but will not install to default locations. `brew install cyrus-sasl` outputs these messages:

```
cyrus-sasl is keg-only, which means it was not symlinked into /opt/homebrew,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.
```

And:

```
For pkg-config to find cyrus-sasl you may need to set:
  export PKG_CONFIG_PATH="/opt/homebrew/opt/cyrus-sasl/lib/pkgconfig"
```

Can the prior behavior (search for header + library) be added as a fallback to prevent breaking macOS configurations?

kkloberdanz (approved):
Looks good! Minor comment.

build/cmake/FindSASL2.cmake
vector-of-bool: I think I will revert to the prior behavior, as the pkg-config impl is fragile in several ways.

build/cmake/GeneratePkgConfig.cmake
```diff
@@ -282,6 +282,10 @@ function(_generate_pkg_config_content out)
         $<GENEX_EVAL:$<TARGET_PROPERTY:pkg_config_LIBS>>
         $<REMOVE_DUPLICATES:$<TARGET_PROPERTY:INTERFACE_LINK_OPTIONS>>
         )
+
+    # XXX: Could we define a genex that can transform the INTERFACE_LINK_LIBRARIES to a list of
```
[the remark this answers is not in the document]
vector-of-bool: That's a callout to future readers that may know better than I do, since the current way its handled is sub-optimal but not "broken" to warrant a "TODO" or "FIXME".

build/cmake/FindSASL2.cmake
kevinAlbs: Confirmed building with b2c72b662cfeed8677d72d1b9a9d1013a21b8641 is able to locate libsasl2 on my macOS.

Earthfile
```diff
@@ -71,23 +75,36 @@ BUILD_AND_INSTALL:
...
+
+alpine-build-env-base:
+    ARG --required version
+    FROM +alpine-base --version=$version
+    RUN apk add openssl-dev cyrus-sasl-dev snappy-dev ccache
 
+alpine-test-env-base:
+    ARG --required version
+    FROM +alpine-base --version=$version
+    RUN apk add libsasl snappy pkgconfig
```
kevinAlbs, line 88:
```suggestion
    RUN apk add libsasl snappy
```

Remove package config dependency? I assume this is no longer needed due to 0c7694670c339aaf82ddaeee999299daf420ccd4. Comment applies to other package config dependencies in this file: `pkgconf` and `pkg-config`

.evergreen/config_generator/components/earthly.py
```diff
@@ -47,9 +49,17 @@ def __init__(self, *, suffix: str, target: str) -> None:
     "amazon2",
 ]
 
+# Other optoins: SSPI (Windows only), AUTO (not reliably test-able without more environments)
```
kevinAlbs, line 52:
```suggestion
# Other options: SSPI (Windows only), AUTO (not reliably test-able without more environments)
```

kevinAlbs (approved):
Changes LGTM, though one task failure may need to be investigated:

The [MinGW task](https://spruce.mongodb.com/task/mongo_c_driver_smoke_link_with_cmake_mingw_patch_fe68d3b0c878c7bb1ba7083d713daa392835e441_6504c6b51e2d173d1b0b419a_23_09_15_21_03_52/logs?execution=0) appears to fail to locate libsasl2:

```
[2023/09/15 21:08:31.592] CMake Error at C:/Users/mci-exec/AppData/Local/mongo-c-driver/tmp.pTUuQRMADo/cmake-3.25.2-windows-x86_64/share/cmake-3.25/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
[2023/09/15 21:08:31.593]   Could NOT find SASL2 (missing: SASL2_VERSION SASL2_INCLUDE_DIR
[2023/09/15 21:08:31.593]   SASL2_LIBRARY) (Required is at least version "2.0")
```

PR #1406, opened 2023-09-15
Share SCRAM cache via a global lookup table

kkloberdanz:
CDRIVER-3514

# Overview

This fixes an issue that Genny encountered where performance was degraded due to clients not sharing a SCRAM cache within a pool.

This change makes the scram cache global so that way all clients share the SCRAM cache.

# Measurements

I measured the time it took to execute the program below and recorded the runtime in milliseconds. I first did a "warmup" run of 100 iterations to literally warm up my CPU so that there would be less variability in boost clock speed, thus ensuring the system would be at a steady state by the time I start collecting data points. The data from these first 100 iterations was thrown away. I then immediately ran 1000 iterations of this program and collected the timing measurements to produce the median results below.

On master (8a0e9c28b), the median runtime is 1299.5 ms.
On this branch (aa0133e9d), the median runtime is 186 ms.

This represents a 7 times speedup by using the new global scram cache.

The following source code was used for collecting measurements. The program below is a slight modification to the test program that Josh wrote when he initially started this project.

```C
#include <stdio.h>
#include <mongoc/mongoc.h>
#include <pthread.h>

#define NUM_THREADS 5000

static void *threadScramAuth(void *data) {
   mongoc_client_pool_t *pool = data;
   mongoc_client_t *client;
   bson_t ping = BSON_INITIALIZER;
   bson_error_t error;
   bool r;
   int rc;

   BSON_APPEND_INT32 (&ping, "ping", 1);

   client = mongoc_client_pool_pop (pool);
   r = mongoc_client_command_simple (
         client, "admin", &ping, NULL, NULL, &error);
   if (!r) {
      fprintf(stderr, "FAILED TO PING SERVER!\n");
      abort();
   }
   mongoc_client_pool_push(pool, client);

   bson_destroy (&ping);

   return NULL;
}

int main() {
   mongoc_database_t *database = NULL;
   mongoc_client_pool_t *pool = NULL;
   bson_error_t error;
   const char *uri_string = "mongodb://127.0.0.1/";
   char * scram_uri;
   mongoc_uri_t *uri = NULL;
   bson_t roles;
   void *ret;
   pthread_t threads[NUM_THREADS];
   int rc = 0;
   bool success;
   mongoc_init ();

   bson_init (&roles);

   uri = mongoc_uri_new_with_error (uri_string, &error);
   if (!uri) {
      fprintf (stderr,
               "failed to parse URI: %s\n"
               "error message:       %s\n",
               uri_string,
               error.message);
      return EXIT_FAILURE;
   }

   client = mongoc_client_new_from_uri (uri);
   if (!client) {
      return EXIT_FAILURE;
   }

   mongoc_uri_destroy(uri);

   success = mongoc_client_set_error_api (client, 2);
   if (!success) {
      perror("mongoc_client_set_error_api");
      exit(EXIT_FAILURE);
   }

   database = mongoc_client_get_database (client, "test");
   if (!database) {
      perror("mongoc_client_get_database");
      exit(EXIT_FAILURE);
   }

   BCON_APPEND (&roles, "0", "{", "role", "root", "db", "admin", "}");
   mongoc_database_add_user (database, "user,=", "pass", &roles, NULL, &error);
   mongoc_database_destroy (database);

   scram_uri = bson_strdup_printf("mongodb://user,=:pass@127.0.0.1/test?appname=scram-example&maxPoolSize=%d&authMechanism=SCRAM-SHA-1", NUM_THREADS);
   if (!scram_uri) {
      perror("bson_strdup_printf");
      exit(EXIT_FAILURE);
   }

   uri = mongoc_uri_new_with_error (scram_uri, &error);

   bson_free(scram_uri);
   if (!uri) {
      fprintf (stderr,
               "failed to parse URI\n"
               "error message: %s\n",
               error.message);
      return EXIT_FAILURE;
   }

   pool = mongoc_client_pool_new (uri);
   if (!pool) {
      printf("pool failure\n");
   }
   mongoc_client_pool_set_error_api (pool, 2);

   printf("starting with %d threads \n", NUM_THREADS);

   int64_t before = bson_get_monotonic_time();
   for (int i = 0; i < NUM_THREADS; ++i) {
      rc = pthread_create(&threads[i], NULL, threadScramAuth, pool);
      if (rc) {
         perror("pthread_create");
         exit(1);
      }
   }

   for (int i = 0; i < NUM_THREADS; ++i) {
      rc = pthread_join (threads[i], &ret);
      if (rc) {
         perror("pthread_join");
         exit(1);
      }
   }

   int64_t after = bson_get_monotonic_time();

   printf("total time taken: %"PRId64 " ms\n", (after - before)/1000);

   bson_destroy (&roles);
   mongoc_uri_destroy(uri);
   mongoc_client_pool_destroy(pool);
   mongoc_cleanup ();
   return 0;
}
```

I used the following script to run the above program 1000 times per branch

```sh
#!/bin/bash

set -e

for i in {1..100}
do
    ./cmake-build/main.out &> /dev/null
done

for i in {1..1000}
do
    ./cmake-build/main.out | grep 'total time taken' | awk '{ print $4 }'
done
```

### DIFF
--- a/.tsan-suppressions
+++ b/.tsan-suppressions
@@ -1,1 +1,2 @@
 race:mongoc_counter*
+race:_mongoc_handshake_freeze
--- a/src/libmongoc/src/mongoc/mongoc-scram.c
+++ b/src/libmongoc/src/mongoc/mongoc-scram.c
@@ -375,9 +375,7 @@ again:
          !memcmp (cache_entry->salted_password,
                   scram->salted_password,
                   sizeof (cache_entry->salted_password)) &&
-         !memcmp (cache_entry->decoded_salt,
-                  scram->decoded_salt,
-                  sizeof (cache_entry->salted_password));
+         );
 
       if (already_exists) {
          /* cache entry already populated between read and write lock
--- a/src/libmongoc/src/mongoc/mongoc-scram.c
+++ b/src/libmongoc/src/mongoc/mongoc-scram.c
@@ -104,6 +104,7 @@ _mongoc_scram_cache_clear (void)
 static BSON_ONCE_FUN (_mongoc_scram_cache_init)
 {
    bson_shared_mutex_init (&g_scram_cache_rwlock);
+   bson_mutex_init (&clear_cache_lock);
    _mongoc_scram_cache_clear ();
 
    BSON_ONCE_RETURN;
--- a/src/libmongoc/src/mongoc/mongoc-scram.c
+++ b/src/libmongoc/src/mongoc/mongoc-scram.c
@@ -86,7 +86,7 @@ static BSON_ONCE_FUN (_block_until_first_cache_entry_populated)
    BSON_ONCE_RETURN;
 }
 
-static BSON_ONCE_FUN (_unblock_when_first_cache_entry_populated)
+static BSON_ONCE_FUN (_unblock_after_first_cache_entry_populated)
 {
    bson_mutex_unlock (&first_cache_entry_lock);
 
@@ -1222,7 +1222,7 @@ CLEANUP:
    bson_free (val_v);
 
    bson_once (&unblock_after_first_entry_once_control,
-              _unblock_when_first_cache_entry_populated);
+              _unblock_after_first_cache_entry_populated);
    return rval;
 }
 
--- a/src/libmongoc/src/mongoc/mongoc-scram.c
+++ b/src/libmongoc/src/mongoc/mongoc-scram.c
@@ -80,31 +80,8 @@ _mongoc_utf8_code_point_to_str (uint32_t c, char *out);
 
 static bson_shared_mutex_t g_scram_cache_rwlock;
 
-/*
- * On the first cache miss, let one thread fill the first cache entry before
- * letting other threads proceed. This stops the inrush of many threads trying
- * to get scram credentials on an empty scram cache.
- */
-static bson_once_t block_until_first_entry_once_control = BSON_ONCE_INIT;
-static bson_once_t unblock_after_first_entry_once_control = BSON_ONCE_INIT;
-static bson_mutex_t first_cache_entry_lock;
-
 static bson_once_t init_cache_once_control = BSON_ONCE_INIT;
 static bson_mutex_t clear_cache_lock;
-
-static BSON_ONCE_FUN (_block_until_first_cache_entry_populated)
-{
-   bson_mutex_lock (&first_cache_entry_lock);
-
-   BSON_ONCE_RETURN;
-}
-
-static BSON_ONCE_FUN (_unblock_after_first_cache_entry_populated)
-{
-   bson_mutex_unlock (&first_cache_entry_lock);
-
-   BSON_ONCE_RETURN;
-}
 
 /*
  * Cache lookups are a linear search through this table. This table is a
@@ -127,8 +104,6 @@ _mongoc_scram_cache_clear (void)
 static BSON_ONCE_FUN (_mongoc_scram_cache_init)
 {
    bson_shared_mutex_init (&g_scram_cache_rwlock);
-   bson_mutex_init (&first_cache_entry_lock);
-   bson_mutex_init (&clear_cache_lock);
    _mongoc_scram_cache_clear ();
 
    BSON_ONCE_RETURN;
@@ -952,8 +927,6 @@ _mongoc_scram_step2 (mongoc_scram_t *scram,
        * Block other threads here to until SCRAM step 3 finishes so that the
        * cache will get populated for the others threads.
        */
-      bson_once (&block_until_first_entry_once_control,
-                 _block_until_first_cache_entry_populated);
    }
 
    if (!*scram->salted_password) {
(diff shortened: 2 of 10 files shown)


src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -675,24 +675,6 @@ test_mongoc_client_authenticate_cached (bool pooled)
       }
    }
 
-   /* screw up the cache */
```
kkloberdanz, line 678: With this change, the cache can no longer be tampered with as such. Tampering with the local cache gets fixed during the next call to `_mongoc_cluster_auth_scram_continue`, specifically at this line:
```C
   cluster->scram_cache = _mongoc_scram_get_cache (scram);
```

.tsan-suppressions
```diff
@@ -1 +1,2 @@
-race:mongoc_counter*
\ No newline at end of file
+race:mongoc_counter*
+race:_mongoc_handshake_freeze
```
kkloberdanz, line 2: Suppress the pre-existing race condition which is detailed in this ticket:
https://jira.mongodb.org/browse/CDRIVER-3685

src/libmongoc/tests/test-libmongoc.c
```diff
@@ -1295,16 +1295,17 @@ test_framework_replset_name (void)
...
    call_hello (&reply);
    if (!bson_iter_init_find (&iter, &reply, "setName")) {
-      return NULL;
+      goto cleanup;
    }
 
    replset_name = bson_strdup (bson_iter_utf8 (&iter, NULL));
-   bson_destroy (&reply);
 
+cleanup:
```
kkloberdanz, line 1307: This function had a pre-existing memory leak which is uncovered by the new unit test.

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -997,6 +1221,8 @@ _mongoc_scram_step3 (mongoc_scram_t *scram,
    bson_free (val_e);
    bson_free (val_v);
 
+   bson_once (&unblock_after_first_entry_once_control,
+              _unblock_after_first_cache_entry_populated);
```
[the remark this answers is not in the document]
kkloberdanz: This logic indeed seems to no longer have a noticeable affect, so I will remove it.

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -127,8 +104,6 @@ _mongoc_scram_cache_clear (void)
 static BSON_ONCE_FUN (_mongoc_scram_cache_init)
 {
    bson_shared_mutex_init (&g_scram_cache_rwlock);
-   bson_mutex_init (&first_cache_entry_lock);
-   bson_mutex_init (&clear_cache_lock);
```
kevinAlbs: Unintentionally removed?

kevinAlbs (approved):
LGTM.

Running the repro and [printing cache hits](https://github.com/kevinAlbs/mongo-c-driver/commit/d9e2487819247ae1c8b67bbfeb47e370e35f43a5) resulted in ~91 of 100 threads hitting the cache.

Aside: Attempting to run the repro with 5000 threads on my machine results in many failures to connect:

```
2023/10/09 10:38:17.0179: [7585806]:  WARNING:      cluster: Failed connection to 127.0.0.1:27017 (Failed to connect to target host: 127.0.0.1:27017 : Connection reset by peer)
2023/10/09 10:38:17.0179: [7585994]:  WARNING:       stream: Failed to buffer 4 bytes
```

Do you recall any needed configuration on mongod or macOS needed to run the test?

Reducing the thread count from 5000 to 100 resulted in no errors.

kkloberdanz:
> LGTM.
> 
> Running the repro and [printing cache hits](https://github.com/kevinAlbs/mongo-c-driver/commit/d9e2487819247ae1c8b67bbfeb47e370e35f43a5) resulted in ~91 of 100 threads hitting the cache.
> 
> Aside: Attempting to run the repro with 5000 threads on my machine results in many failures to connect:
> 
> ```
> 2023/10/09 10:38:17.0179: [7585806]:  WARNING:      cluster: Failed connection to 127.0.0.1:27017 (Failed to connect to target host: 127.0.0.1:27017 : Connection reset by peer)
> 2023/10/09 10:38:17.0179: [7585994]:  WARNING:       stream: Failed to buffer 4 bytes
> ```
> 
> Do you recall any needed configuration on mongod or macOS needed to run the test?
> 
> Reducing the thread count from 5000 to 100 resulted in no errors.

It's hard to say how to get this working on MacOS. The error that you're getting appears to be a result of mongod being overwhelmed by all of the connections. Anecdotally when I was running Nginx on a Mac a while back, I would see this exact same issue on a lot lower load than running Nginx on either Linux or FreeBSD (doing an `ab` test). I believe the MacOS networking stack is not very optimized for high load networking, where Linux and FreeBSD have pretty highly optimized networking stacks, so for pathological cases like this -- running an insane number of threads on a single host -- MacOS, might not be the ideal test bench.

I ran all of my tests on an 8 core AMD Ryzen 7 2700X using Pop!_OS 20.04 (Linux).

PR #1408, opened 2023-09-20
CDRIVER-4730 TLS Backend Cleanup

vector-of-bool:
This PR is a continuation of work from #1405, and addresses the `ENABLE_SSL` setting and related logic.

## Changed

- The `FindLibreSSL.cmake` module has been added, and comes from the LibreSSL project itself, plus a few bugfixes. Newer LibreSSL ships CMake config-file packages, so maybe we should rely on those instead? It depends on whether our LibreSSL userbase is all on the newer version (very unlikely, as that's a new feature).
- The Earthfile has been expanded to include parameters for the `tls` build option, which controls which packages are installed and sets the appropriate CMake build settings.
- Earthfile parameters have been renamed and reformatted for just being "prettier" in general.
- More EVG tasks were added for the additional configurations. The `earthly.py` file has been expanded to express these additional complexities, since it now needs to exclude tasks on certain platforms that do not (yet) have a LibreSSL.

## Task Goofiness

I wasn't particularly happy with the way EVG tasks are named, as they're quite hard on the eyes. Evergreen does not allow spaces in task names, forward-slash breaks things, and commas are "not supported", so I broke out the unicode to create new exciting names, such as:

> check : sasl:Cyrus • tls:LibreSSL • test_mongocxx_ref:master

which uses U+00a0 "No-break space" and U+2022 "bullet". I'm not married to this trick, so feel free to veto this aside change.


CMakeLists.txt
```diff
@@ -623,4 +618,4 @@ if (CMAKE_GENERATOR STREQUAL "Ninja Multi-Config" AND PROJECT_IS_TOP_LEVEL)
    set (CMAKE_CROSS_CONFIGS "all")
 endif ()
 
-feature_summary(WHAT ENABLED_FEATURES DISABLED_FEATURES)
+feature_summary(WHAT ALL)
```
kevinAlbs, line 621:
```suggestion
feature_summary(WHAT ENABLED_FEATURES DISABLED_FEATURES)
```

Revert?

Testing configuring with:
```
cmake \
    -G "Ninja" \
    -DENABLE_SSL=DARWIN \
    -S$(pwd) -B$(pwd)/cmake-build
```

Before these changes shows:

```
-- The following features have been enabled:

 * TLS, for secure network communication (SecureTransport)
 * Cryptography, cryptographic primitives (SecureTransport)
 * AWS Authentication, authenticate with MongoDB servers using credentials from AWS instance metadata
 * SASL Authentication, authenticate with MongoDB servers using SASL: “Simple Authentication and Security Layer” (Cyrus)

-- Configuring done (0.8s)
```

After these changes:
```
-- The following features have been enabled:

 * TLS, for secure network communication (SecureTransport)
 * Cryptography, cryptographic primitives (SecureTransport)
 * AWS Authentication, authenticate with MongoDB servers using credentials from AWS instance metadata
 * SASL Authentication, authenticate with MongoDB servers using SASL: “Simple Authentication and Security Layer” (Cyrus)

-- The following OPTIONAL packages have been found:

 * Python
 * Python3
 * PkgConfig
 * mongocrypt

-- The following REQUIRED packages have been found:

 * SASL2 (required version >= 2.0), Cyrus implementation of SASL
   Provides libmongoc SASL authentication support
 * Threads
 * Sphinx

-- The following REQUIRED packages have not been found:

 * Utf8Proc

-- Configuring done (0.7s)
```

The listing of `Utf8Proc` as "REQUIRED" and "not found" seems unexpected. That may be surfacing a separate issue. Until that is addressed, I suggest excluding from the feature summary.

tools/str.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/sh
+
+# This is a tiny shell utility for string manipulation.
+# Allow us to use "local"
```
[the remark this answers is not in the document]
vector-of-bool: It refers to the shellcheck directive on the next line, which would otherwise warn about the `local` command.

src/libmongoc/CMakeLists.txt
```diff
@@ -147,99 +135,180 @@ if (NOT ENABLE_ZSTD STREQUAL OFF)
...
 
-   if (ENABLE_SSL STREQUAL WINDOWS OR (WIN32 AND ENABLE_SSL STREQUAL "AUTO"))
-      if (WIN32)
-         set (SECURE_CHANNEL 1)
-      else ()
-         message (FATAL_ERROR "ENABLE_SSL=WINDOWS only supported on Windows")
-      endif ()
-   endif ()
+# Per-backend link libs/options:
+set(SecureTransport/LINK_OPTIONS "SHELL:-framework CoreFoundation" "SHELL:-framework Security")
```
kevinAlbs, line 190: Remove `SHELL:`?

On local install (with macOS), this produced a package config file with `SHELL:` in `Libs`:

```
# pkg-config .pc file generated by CMake 3.27.4 for libmongoc-1.11.1. DO NOT EDIT!
prefix=/Users/kevin.albertson/review/mongo-c-driver-1408/.install
exec_prefix=${prefix}
libdir=${exec_prefix}/lib

Name: libmongoc
Description: The MongoDB C Database Driver library
Version: 1.11.1
Requires: libbson-static-1.0

Libs: -L${libdir} -lmongoc-static-1.0 -lresolv -lsasl2 -framework Corefoundation -framework Security -L/opt/homebrew/Cellar/zstd/1.5.5/lib -lzstd -L/Users/kevin.albertson/code/c-bootstrap/install/libmongocrypt-1.8.0-alpha1/lib -lmongocrypt -fsanitize=address,leak SHELL:-framework CoreFoundation SHELL:-framework Security
Cflags: -fPIC -fsanitize=address,leak -DMONGOC_STATIC -DBSON_STATIC -I${prefix}/include/libmongoc-1.0
```

Attempting to use resulted in an error:
```
% export PKG_CONFIG_PATH=/Users/kevin.albertson/review/mongo-c-driver-1408/.install/lib/pkgconfig
% /opt/homebrew/opt/llvm/bin/clang -o test.c $(pkg-config --libs --cflags libmongoc-static-1.0)
clang-16: error: no such file or directory: 'SHELL:-framework'
clang-16: error: no such file or directory: 'CoreFoundation'
clang-16: error: no such file or directory: 'SHELL:-framework'
clang-16: error: no such file or directory: 'Security'
```

kevinAlbs:
> which uses U+00a0 "No-break space" and U+2022 "bullet". I'm not married to this trick, so feel free to veto this aside change.

I prefer avoiding the "No-break space". I often use the `evergreen` CLI to run individual tasks (with the `-v` and `-t` arguments). I expect the "No-break space" may be a surprise when copy-pasting a task name. I suggest removing the spaces: `check:sasl:Cyrus•tls:LibreSSL•test_mongocxx_ref:master` or replacing with a non-space Unicode character.

.evergreen/config_generator/components/earthly.py
```diff
@@ -1,41 +1,171 @@
...
+TLSOption = Literal["LibreSSL", "OpenSSL", "off"]
+"Options for the TLS backend configuration parameter (AKA 'ENABLE_SSL')"
+CxxVersion = Literal["master", "r3.8.0"]
+"C++ driver refs that are under CI test"
+
+# A Unicode non-breaking space character
+_NBSP = "\N{No-break space}"
+
+
+class Configuration(NamedTuple):
```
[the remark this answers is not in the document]
vector-of-bool: Created: CDRIVER-4731

Currently, none of the tasks are actual duplicates, since they target a disjoint set of build environments.

CMakeLists.txt
vector-of-bool: I forgot about the misleading `Utf8Proc` report, which is why I didn't use `ALL` in the prior PR. I can see the cause and opened a ticket for this one: CDRIVER-4732

src/libmongoc/CMakeLists.txt
vector-of-bool: Reverted

.evergreen/config_generator/components/earthly.py
```diff
@@ -1,41 +1,171 @@
...
+        fields: Iterable[tuple[str, type]] = get_type_hints(Configuration).items()
+        # Generate lists of pairs of parameter names their options:
+        all_pairs: Iterable[Iterable[tuple[str, str]]] = (
+            # Generate a (key, opt) pair for each option in parameter 'key'
+            [(key, opt) for opt in get_args(typ)]
+            # Over each parameter and type thereof:
+            for key, typ in fields
+        )
+        # Now generate the cross product of all alternative for all options:
+        matrix: Iterable[dict[str, Any]] = map(dict, itertools.product(*all_pairs))
```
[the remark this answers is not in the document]
vector-of-bool: My concern isn't about the runtime test coverage, but that the rest of the project continues to build and install successfully when it is turned off (i.e. does code exist that assumes it is enabled?) At the moment, the tasks only execute the build without running the test suite, so they are very fast. It will be worthwhile to narrow down our task matrix if/when it becomes prohibitively expensive.

.evergreen/config_generator/components/earthly.py
```diff
@@ -1,41 +1,159 @@
...
+            yield Configuration(**items)
+
+    @property
+    def suffix(self) -> str:
+        return f"{_BULLET}".join(f"{k}={v}" for k, v in self._asdict().items())
+
+
+def task_filter(env: EnvKey, conf: Configuration) -> bool:
+    """
+    Control which tasks are actually defined by matching on the
```
kevinAlbs, line 81: Unfinished comment?

kevinAlbs (approved):

PR #1409, opened 2023-09-20
CDRIVER-4676 Switch benchmarks to rhel90-dbx-perf-large

adriandole:
# Background
DRIVERS-2666 requires:
* Benchmarks be performed on `rhel90-dbx-perf-large`
* A pinned server version, currently 6.0.6.
* The `perf.send` backend to report results.

C driver benchmarking hasn't worked since October 2022.

# Changes
- Compilation and benchmarking switched from `centos6` to `rhel90-dbx-perf`.
- `find-cmake.sh` removed from the compilation step since it tries to download an old incompatible cmake version. The cmake provided by the distro is fine.
- MongoDB 4.4 and 5.0.2 replaced with 6.0.6.
- Ubuntu 18.04 testing, which was motivated by inconsistent benchmarks on `centos6-perf` (CDRIVER-4105), has been removed since `rhel90-dbx-perf` should be more stable.

# Validation
[Passing patch.](https://spruce.mongodb.com/version/650a2d6f1e2d17fc0d68a306/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/.evergreen/benchmark.yml
+++ b/.evergreen/benchmark.yml
@@ -18,10 +18,8 @@ variables:
 
   ## Common download urls (merge in as hashes)
   mongo_download_url_prefixes:
-    mongo_v44: &mongo_v44
-      mongo_url: "http://downloads.10gen.com/linux/mongodb-linux-x86_64-enterprise-rhel62-v4.4-latest.tgz"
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
LGTM with a minor suggestion and consideration for moving the compile script.

PR #1411, opened 2023-09-25
CDRIVER-4733 Replace version macros expanded in configure-generated files

vector-of-bool:


kevinAlbs (approved):
LGTM.

Filed CDRIVER-4733 to track the change in the C driver project. Since it may impact downstream users, having the fixVersion on the ticket may make it easier to identify which release introduced the change.

Include `CDRIVER-4733` in the commit message to have the git hook add a comment with the commit to the ticket.

PR #1412, opened 2023-09-25
assert that pthread_* functions return a successful return code

kkloberdanz:


src/common/common-thread-private.h
```diff
@@ -38,7 +38,10 @@ BSON_BEGIN_DECLS
 #define BSON_ONCE_FUN(n) void n (void)
 #define BSON_ONCE_RETURN return
 #define BSON_ONCE_INIT PTHREAD_ONCE_INIT
-#define bson_once pthread_once
+#define bson_once(o, c)                           \
+   do {                                           \
+      BSON_ASSERT (pthread_once ((o), (c)) == 0); \
+   } while (0)
```
vector-of-bool, line 44: Suggest that this (and others) be replaced with a `static inline` function of the same name. e.g.

```suggestion
static inline void bson_once(bson_once_t* once, void (*cb)(void)) {
  int rc = pthread_once(once, db);
  BSON_ASSERT(rc == 0);
}
```

But: I'd really like to completely rewrite this file to be less macro-ful, so it may be preferable to just defer such work to a larger task.

vector-of-bool (approved):

kevinAlbs (approved):

PR #1419, opened 2023-09-29
CDRIVER-4740 Improve performance of bson_utf8_escape_for_json

alcaeus:
While implementing the BSON performance benchmarks in the PHP driver, we noticed a performance problem when converting BSON to JSON. In our tests, we found that encoding a `bson_t*` to JSON using `bson_as_json` was significantly slower than converting the BSON to PHP objects using a `bson_visitor_t`, then running PHP's own JSON encoder on it. Using `bson_as_json` was slower by about a factor of 5. A bit of trial and error brought me to `bson_utf8_escape_for_json`: since the performance payload didn't need it, I removed the calls for testing purposes and the performance of `bson_as_json` improved dramatically to the point where it was not significantly slower than JSON encoding using PHP's logic.

This pull request is the result of the analysis @jmikola and I did, comparing libbson's escape logic with PHP's.

The first commit introduces a performance test for `bson_utf8_escape_for_json`. This test escapes the string `"my\0key"` 1 million times. The baseline time for this was 0.359726 according to test output.

Comparing libbson's implementation with PHP's [`php_json_escape_string`](https://github.com/php/php-src/blob/7200bc23a097d6f55b2810c81a76928ae723eee5/ext/json/json_encoder.c#L318), I found several fixes which I applied commit by commit:

The first fix regards allocation of a new `bson_string_t` for the escaped value. Previously, an empty `bson_string_t` was created, which then grows by factors of 2 as we append escaped characters. Since we know that escaping a string of length `n` produces a string with a length greater or equal than `n`, I added a `bson_string_alloc` similar to `bson_string_new`, but allocating a number of bytes while setting the length to 0. This reduces the number of reallocations while encoding and resulted in a benchmark time of 0.327364 seconds, yielding a ~9% improvement.

The second commit was heavily inspired by PHP's logic in `php_json_escape_string` and changes the way many "straightforward" characters are handled. To determine whether characters need escaping, it leverages a bit mask. Note that our bit mask looks slightly different from PHP's, as PHP offers to escape significantly more characters depending on options.
Characters that don't need escaping are no longer copied to the escaped string one by one, but rather in a single chunk once the first special character is encountered. This reduces the number of calls to `memcpy`. To help with this, I extracted `bson_string_append_ex` which appends a given number of characters from the source string to the destination. This is necessary as we don't want to append the entire unescaped string, but only a certain number of characters.
Any non-ASCII characters (`c >= 0x80`) are appended using the previous logic of using `bson_string_append_unichar`.
At this point in the loop, the only characters left to handle are those that need escaping (`"`, `\`, `\b`, `\f`, `\n`, `\r`, `\t`) and all non-printable ASCII characters, which are added either escaped, or as code points (`\uxxxx`).

This change resulted in a significant performance improvement, bringing the benchmark time down to 0.207105 seconds, a 42% improvement compared to the baseline.

The last optimisation is around adding code points: I noticed that `bson_string_append_printf` was rather slow, especially when the format is rather predictable. A new `bson_string_append_codepoint` method optimises for this path, bypassing `printf` for small enough values.

This last optimisation further reduces the benchmark time to 0.122769 seconds, a whopping 65% faster than the original.

Here's a summary of the benchmark results:

Test | Time | Improvement
-- | -- | --
Baseline | 0,359726 |  
`bson_string_alloc` | 0,327364 | 8,996 %
Bit mask optimisation | 0,207105 | 42,427 %
Code point optimisation | 0,122769 | 65,872 %

For now, I created this PR as a draft as I'm sure there are a number of potential improvements. For one, all of the new methods are part of the public API, which might not be desired. I'm also sure there are a number of tests that can be added to cover all potential "unhappy paths".

### DIFF
--- a/src/libbson/src/bson/bson-string.c
+++ b/src/libbson/src/bson/bson-string.c
@@ -77,10 +77,8 @@ bson_string_new (const char *str) /* IN */
    if (str) {
       memcpy (ret->str, str, ret->len);
    }
-   ret->str[ret->len] = '\0';
 
    ret->str[ret->len] = '\0';
-
    return ret;
 }
 
@@ -127,11 +125,7 @@ bson_string_alloc (uint8_t len)
    BSON_ASSERT (ret->alloc >= 1);
 
    ret->str = bson_malloc (ret->alloc);
-
    ret->str[ret->len] = '\0';
-
-   ret->str[ret->len] = '\0';
-
    return ret;
 }
 
@@ -302,7 +296,7 @@ bson_string_append_codepoint (bson_string_t *string,  /* IN */
 
    BSON_ASSERT (string);
 
-   if (unichar > 0xFFFF) {
+   if (unichar > 0xffff) {
       bson_string_append_printf (string, "\\u%04x", unichar);
       return;
    }
--- a/src/libbson/src/bson/bson-string.c
+++ b/src/libbson/src/bson/bson-string.c
@@ -160,6 +160,45 @@ bson_string_free (bson_string_t *string, /* IN */
 /*
  *--------------------------------------------------------------------------
  *
+ * bson_string_append_ex --
+ *
+ *       Append the UTF-8 string @str of given length @len to @string.
+ *
+ * Returns:
+ *       None.
+ *
+ * Side effects:
+ *       None.
+ *
+ *--------------------------------------------------------------------------
+ */
+
+void
+bson_string_append_ex (bson_string_t *string, /* IN */
+                       const char *str,       /* IN */
+                       uint32_t len)          /* IN */
+{
+   BSON_ASSERT (string);
+   BSON_ASSERT (str);
+
+   if ((string->alloc - string->len - 1) < len) {
+      string->alloc += len;
+      if (!bson_is_power_of_two (string->alloc)) {
+         string->alloc =
+            (uint32_t) bson_next_power_of_two ((size_t) string->alloc);
+      }
+      string->str = bson_realloc (string->str, string->alloc);
+   }
+
+   memcpy (string->str + string->len, str, len);
+   string->len += len;
+   string->str[string->len] = '\0';
+}
+
+
+/*
+ *--------------------------------------------------------------------------
+ *
  * bson_string_append --
  *
  *       Append the UTF-8 string @str to @string.
@@ -177,25 +216,7 @@ void
 bson_string_append (bson_string_t *string, /* IN */
                     const char *str)       /* IN */
 {
-   uint32_t len;
-
-   BSON_ASSERT (string);
-   BSON_ASSERT (str);
-
-   len = (uint32_t) strlen (str);
-
-   if ((string->alloc - string->len - 1) < len) {
-      string->alloc += len;
-      if (!bson_is_power_of_two (string->alloc)) {
-         string->alloc =
-            (uint32_t) bson_next_power_of_two ((size_t) string->alloc);
-      }
-      string->str = bson_realloc (string->str, string->alloc);
-   }
-
-   memcpy (string->str + string->len, str, len);
-   string->len += len;
-   string->str[string->len] = '\0';
+   bson_string_append_ex (string, str, (uint32_t) strlen (str));
 }
 
 
--- a/src/libbson/src/bson/bson-string.c
+++ b/src/libbson/src/bson/bson-string.c
@@ -84,6 +84,57 @@ bson_string_new (const char *str) /* IN */
    return ret;
 }
 
+/*
+ *--------------------------------------------------------------------------
+ *
+ * bson_string_alloc --
+ *
+ *       Create a new bson_string_t and allocate memory for it.
+ *
+ *       bson_string_t is a power-of-2 allocation growing string. Every
+ *       time data is appended the next power of two size is chosen for
+ *       the allocation. Pretty standard stuff.
+ *
+ *       It is UTF-8 aware through the use of bson_string_append_unichar().
+ *       The proper UTF-8 character sequence will be used.
+ *
+ * Parameters:
+ *       @len: Length of the string to allocate
+ *
+ * Returns:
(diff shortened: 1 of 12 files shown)


src/libbson/src/bson/bson-string.c
```diff
@@ -84,6 +84,57 @@ bson_string_new (const char *str) /* IN */
...
+
+   BSON_ASSERT (ret->alloc >= 1);
+
+   ret->str = bson_malloc (ret->alloc);
+
+   ret->str[ret->len] = '\0';
+
+   ret->str[ret->len] = '\0';
+
+   return ret;
```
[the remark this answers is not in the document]
alcaeus: Oh wow, and I blindly copied it over. Removed here and in `bson_string_new`.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -233,6 +233,9 @@ bson_utf8_validate (const char *utf8, /* IN */
 }
 
 
+#define BSON_BIT_TEST(bits, bit) \
+   (((bits)[(bit) / (sizeof((bits)[0])*8)] >> ((bit) & (sizeof((bits)[0])*8-1))) & 1)
+
```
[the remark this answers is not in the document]
alcaeus: Wow, thank you for rewriting it. It took me a while to understand how this worked, but I didn't feel brave enough to rewrite it. Unfortunately, PHP's source code has a strange affection to macros, this being no exception.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -266,54 +288,106 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
...
+   if (utf8_len == 0) {
+      return bson_strdup ("");
+   }
+
+   str = bson_string_alloc (utf8_len);
+
+   pos = 0;
+
+   do {
+      c = (unsigned char) utf8[pos];
```
kevinAlbs, line 309: Is the input to `bson_utf8_escape_for_json` expected to be valid UTF-8?

This test results in an error reported by ASAN:

```c
str = bson_utf8_escape_for_json ("\xc6", 1);
ASSERT_MEMCMP (str, "\\u00c6", 2);
bson_free (str);
```

Error:
```
==25574==ERROR: AddressSanitizer: global-buffer-overflow on address 0x000100f51942 at pc 0x000100ebc978 bp 0x00016f97a2b0 sp 0x00016f97a2a8
READ of size 1 at 0x000100f51942 thread T0
    #0 0x100ebc974 in bson_utf8_escape_for_json bson-utf8.c:309
    #1 0x10080b540 in test_bson_utf8_escape_for_json test-utf8.c:87
    #2 0x100d32d54 in TestSuite_AddHelper TestSuite.c:261
    #3 0x100d383d0 in TestSuite_RunTest TestSuite.c:648
    #4 0x100d35b9c in TestSuite_RunAll TestSuite.c:1094
    #5 0x100d34a48 in TestSuite_Run TestSuite.c:1139
    #6 0x100486da0 in main test-libmongoc-main.c:157
    #7 0x101d85088 in start+0x204 (dyld:arm64+0x5088) (BuildId: 6c9a46d971373785afc97feeceb8cadb32000000200000000100000000070c00)
    #8 0x12597ffffffffffc  (<unknown module>)
```

If this `utf8` is required to be valid UTF-8, I suggest updating the documentation to note.

kevinAlbs:
I recommend the use of a fuzz tester on `bson_utf8_escape_for_json`. Here is [a branch](https://github.com/mongodb/mongo-c-driver/compare/master...kevinAlbs:mongo-c-driver:bson-performance-improvements.fuzzed?expand=1) with a fuzz test in this commit: https://github.com/mongodb/mongo-c-driver/commit/50f36d703a90cd2a05a4a66782570cd7832da923

Testing shows a string that returns true for `bson_utf8_validate` but results in a NULL return from `bson_utf8_escape_for_json`:

```c
// Testing \x0a\x0a\xc0\x80 results in NULL return.
// `bson_utf8_validate` reports as valid. This may be an existing bug?
BSON_ASSERT (bson_utf8_validate ("\x0a\x0a\xc0\x80", 4, true /* allow_null */));
str = bson_utf8_escape_for_json ("\x0a\x0a\xc0\x80", 4);
BSON_ASSERT (str); // Fails.
bson_free (str);
```

src/libbson/src/bson/bson-utf8.c
alcaeus: > Is the input to `bson_utf8_escape_for_json` expected to be valid UTF-8?

I'd say "yes". In the context of JSON encoding, for which this function is written, the `bson_iter_visit_all` function always calls `bson_utf8_validate` before invoking the respective visitor function, which is where `bson_utf8_escape_for_json` is called. With that in mind, I don't think we need to handle invalid UTF-8, but I defer to your judgment on this.

kevinAlbs: Prior to this PR, the behavior of encoding `\xc6` appears to produce incorrect results:

```c
   // Test invalid UTF-8.
   // Hex C6 represents the bits 11000110.
   // The leading bits 110 signals a sequence of two bytes.
   // But only one byte is in the input.
   str = bson_utf8_escape_for_json ("\xc6", 1);
   BSON_ASSERT (!str); // Expected NULL returned?
   // Prior to PR, results in encoded value C680.
   // With PR, results in global-buffer-overflow.
```

I think it is reasonable to assume `bson_utf8_escape_for_json` has undefined behavior for invalid UTF-8 input. I suggest updating the documentation in `bson_utf8_escape_for_json.rst` to warn:

```rst
.. warning::
  
  If ``utf8`` represents invalid UTF-8 data the behavior of :symbol:`bson_utf8_escape_for_json` is undefined.
  UTF-8 may be validated with :symbol:`bson_utf8_validate`.
```

alcaeus:
Continued in #1732.

PR #1423, opened 2023-10-02
CDRIVER-3052 Replace most uses of mongoc_collection_update

adriandole:
`mongoc_collection_update` is deprecated. The sample commands are part of the documentation and should not showcase use of a deprecated function. I also replaced `mongoc_collection_update` in a few other tests where the replacement function is obvious. `gridfs` also uses this function, but the tests failed so I skipped it.

### DIFF
--- a/src/libmongoc/tests/test-mongoc-sample-commands.c
+++ b/src/libmongoc/tests/test-mongoc-sample-commands.c
@@ -2600,6 +2600,10 @@ cleanup:
 static void
 test_example_59 (mongoc_database_t *db)
 {
+   if (!test_framework_skip_if_no_txns ()) {
+      return;
+   }
+
    mongoc_client_t *client = NULL;
    client = test_framework_new_default_client ();
 
@@ -2713,6 +2717,10 @@ cleanup:
 static void
 test_example_60 (mongoc_database_t *db)
 {
+   if (!test_framework_skip_if_no_txns ()) {
+      return;
+   }
+
    mongoc_client_t *client = NULL;
    client = test_framework_new_default_client ();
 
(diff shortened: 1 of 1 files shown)


src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -4333,11 +4331,13 @@ test_sample_commands (void)
    test_sample_command (test_example_53, 53, db, collection, false);
    test_sample_command (test_example_54, 54, db, collection, true);
    test_sample_command (test_example_55, 55, db, collection, false);
+   test_sample_command (test_example_56, 56, db, collection, true);
    test_sample_command (test_example_57, 57, db, collection, false);
    test_sample_command (test_example_58, 58, db, collection, false);
-   test_sample_command (test_example_56, 56, db, collection, true);
-   test_sample_command (test_example_59, 59, db, collection, true);
-   test_sample_command (test_example_60, 60, db, collection, true);
+   if (test_framework_skip_if_not_replset ()) {
```
eramongodb, line 4337: It might be more consistent to move this check into the sample function(s) as done [here](https://github.com/mongodb/mongo-c-driver/blob/0e54f99dada9db1196f82a7ef868f58ca8b0f022/src/libmongoc/tests/test-mongoc-sample-commands.c#L3022) for `test_sample_causal_consistency()`. Using `test_framework_skip_if_no_txns()` instead of `test_framework_skip_if_not_replset()` might be preferable for similar reasons.

src/libmongoc/tests/test-mongoc-sample-commands.c
```diff
@@ -4333,11 +4331,13 @@ test_sample_commands (void)
    test_sample_command (test_example_53, 53, db, collection, false);
    test_sample_command (test_example_54, 54, db, collection, true);
    test_sample_command (test_example_55, 55, db, collection, false);
+   test_sample_command (test_example_56, 56, db, collection, true);
```
eramongodb, line 4334: The `/Samples` tests are currently [being skipped](https://github.com/mongodb/mongo-c-driver/blob/0e54f99dada9db1196f82a7ef868f58ca8b0f022/.evergreen/etc/skip-tests.txt#L39). Manually unskipping and running the test suggests there is a test order dependency here: may need to move example 56 back below example 58.

Aside: `test_sample_txn_commands ()` also fails locally with a "shouldn't have logged" error which I believe is due to https://github.com/mongodb/mongo-c-driver/commit/36c1f141b55272f8d09a8492ef74fd9d701d16f6#diff-b71f20ac0b55a16f43592d6bc46791721a0a075febda9e508f3c66d555fae9e9, which I think can be addressed by replacing `ASSERT_NO_CAPTURED_LOGS()` with `ASSERT()` + `has_captured_log()` instead (for both warning and error levels?).

kkloberdanz (approved):

PR #1424, opened 2023-10-02
[CDRIVER-4651] Documentation on vcpkg installation + more

vector-of-bool:
This PR includes updates to documentation for CDRIVER-4651 (and soon CDRIVER-4649). It also re-enables the documentation index and adds markup for miscellaneous index entries.

There is a simple example with a Makefile and Earthly target that builds and tests a small application against the vcpkg packages, but it is not a CI task at the moment.

Refer to the commit messages for details.


src/libmongoc/examples/cmake/vcpkg/app.c
```diff
@@ -0,0 +1,9 @@
+#include <mongoc/mongoc.h>
+
+#include <stdio.h>
+
+int
+main ()
```
kkloberdanz, line 6: A subtle difference between C and C++, it is advised by the ISO C99 standard to use the signature `int main(void)` for a `main` function that takes no arguments. See section `5.1.2.2.1 Program startup` in the link below:
```suggestion
main (void)
```

https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf

src/libmongoc/doc/learn/get/installing.rst
```diff
@@ -30,8 +32,110 @@ management repositories of `common Linux distributions <linux_>`_ and
...
+
+.. index::
+  ! pair: installation; vcpkg
+
+Installing using vcpkg
+======================
+
+.. note::
+  This page will not detail how to get started using vcpkg_. For that, refer to
+  `"Get started with vcpkg" (external)`__
```
eramongodb, line 66: Is this `(external)` necessary? Suggest removing if not (there are already plenty of links to external sites without this qualifier).

src/libmongoc/doc/learn/get/installing.rst
```diff
@@ -30,8 +32,110 @@ management repositories of `common Linux distributions <linux_>`_ and
...
+        ]
+      }
+
+    When you build a CMake project with vcpkg integration and have a
+    ``vcpkg.json`` manifest file, vcpkg will automatically install the project's
+    dependencies before proceeding with the configuration phase, so no
+    additional manual work is required.
+
+
+  .. tab-item:: vcpkg Classic Mode
```
eramongodb, line 97: Do we need to bother documenting and testing Classic Mode if it is not recommended either by our documentation or the vcpkg documentation?

src/libmongoc/doc/learn/get/installing.rst
```diff
@@ -48,7 +152,19 @@ The following Linux distributions provide |libbson| and |libmongoc| packages:
 - `Debian <debian_>`_ and Debian-based distributions, including
   `Ubuntu <debian_>`_ and Ubuntu derivatives, via APT.
 
+.. seealso::
+
+  For a complete list of package names, see: :doc:`/ref/packages`.
```
eramongodb, line 157:
```suggestion
  For a list of available packages, see: :doc:`/ref/packages`.
```
Avoid implying it is a complete list, consistent with the "Important" admonition on the page stating "the information found here may be incomplete or out-of-date".

src/libmongoc/examples/cmake/vcpkg/.gitignore
```diff
@@ -0,0 +1 @@
+_build/
```
eramongodb, line 1:
```suggestion
_build/

```
Missing EOF newline.

The root `.gitignore` file already contains `_build/`; does that not apply here?

src/libmongoc/examples/cmake/vcpkg/Makefile
```diff
@@ -0,0 +1,27 @@
+.SILENT:
+
+THIS_FILE := $(realpath $(lastword $(MAKEFILE_LIST)))
+THIS_DIR := $(shell dirname $(THIS_FILE))
+
+VCPKG_ROOT ?= /opt/vcpkg
+
+test-classic:
```
eramongodb, line 8:
```suggestion
.PHONY: test-classic
test-classic:
```

src/libmongoc/examples/cmake/vcpkg/Makefile
```diff
@@ -0,0 +1,27 @@
...
+	vcpkg install mongo-c-driver
+	rm -rf -- $(THIS_DIR)/_build
+	cmake \
+		-S . \
+		-B _build \
+		-D CMAKE_PROJECT_INCLUDE=$(VCPKG_ROOT)/scripts/buildsystems/vcpkg.cmake
+	cmake --build _build --config=Debug
+	cmake --build _build --config=Debug --target test
+
+test-manifest-mode:
```
eramongodb, line 18:
```suggestion
.PHONY: test-manifest-mode
test-manifest-mode:
```

src/libmongoc/doc/index.rst
```diff
@@ -26,3 +26,5 @@ It depends on :doc:`libbson <bson:index>` to generate and parse BSON documents,
    guides
    api
    application-performance-monitoring
+
+   genindex
```
eramongodb, line 30: Nice! But also exposes an out-of-place inclusion of the `mongoc_kms_credentials_provider_callback_fn` code block from the `mongoc_client_encryption_opts_set_kms_credential_provider_callback` page in the index. Recommend removing this entry from the index if able.

src/libmongoc/doc/ref/packages.rst
```diff
@@ -0,0 +1,200 @@
...
+      the ``epel-release`` package first)
+
+  - - :index:`APK <!pair: APK; package names>`
+    - :index:`Alpine Linux <!pair: Alpine Linux; package names>`
+    - ``libbson``, ``libbson-dev``, ``libbson-static``
+    - ``libmongoc``, ``libmongoc-dev``, ``libmongoc-static``
+    - .. empty cell
+
+  - - :index:`pacman <pair: !package names; pacman>`
+    - :index:`Arch Linux <pair: !package names; Arch Linux>`
```
eramongodb, line 67:
```suggestion
  - - :index:`pacman <!pair: package names; pacman>`
    - :index:`Arch Linux <!pair: package names; Arch Linux>`
```
Are the exclamation marks here misplaced? (Inconsistent with other instances of adding entires to the index.)

eramongodb (changes requested):

kkloberdanz (changes requested):

src/libmongoc/doc/ref/packages.rst
```diff
@@ -0,0 +1,200 @@
...
+
+  - - :index:`APT <!pair: APT; package names>` (``apt``/``apt-get``)
+    - Debian-based Linux distributions (:index:`Debian`, :index:`Ubuntu`, Linux
+      Mint, etc.)
+    - ``libbson-1.0-0``, ``libbson-dev``,
+      ``libbson-doc``
+    - ``libmongoc-1.0-0``, ``libmongoc-dev``,
+      ``libmongoc-doc``
+    - The ``libmongoclient`` packages are for the C++ database driver and will
+      transitively depend on ``libmongoc``.
```
kevinAlbs, line 50:
```suggestion
    - .. empty cell
```

libmongoclient appears to refer to the legacy C++ driver and is not maintained by us.

[mongo-cxx-driver](https://packages.debian.org/source/experimental/mongo-cxx-driver) is in experimental. Suggest removing mention of the C++ driver.

Earthfile
```diff
@@ -245,6 +245,44 @@ multibuild:
...
+        make test-manifest-mode
+
+vcpkg-base:
+    FROM alpine:3.18
+    RUN apk add cmake curl gcc g++ musl-dev ninja-is-really-ninja zip unzip tar \
+                build-base git pkgconf perl bash linux-headers
+    ENV VCPKG_ROOT=/opt/vcpkg-git
+    GIT CLONE --branch=2023.04.15 https://github.com/microsoft/vcpkg $VCPKG_ROOT
+    RUN sh $VCPKG_ROOT/bootstrap-vcpkg.sh -disableMetrics -musl && \
+        install -spD -m 755 $VCPKG_ROOT/vcpkg /usr/local/bin/
```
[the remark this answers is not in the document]
vector-of-bool: My guess is that `bootstrap-vcpkg.sh` is downloading an x86_64 pre-built binary which doesn't run properly on ARM64. [It looks like there is an insufficient check here](https://github.com/microsoft/vcpkg/blob/2db8a9cfb01514e7390327bb2d323d247eddab9f/scripts/bootstrap.sh#L182-L185), which doesn't validate that the libmuslc platform is *also* and x64 platform before doing the download. It looks like it should be fixed by removing the `-musl` argument to the bootstrap script. Can you confirm?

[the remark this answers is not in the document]
kevinAlbs: That sounds plausible. Though applying the change results in the same error:

```
+vcpkg-base *failed* | --> RUN sh $VCPKG_ROOT/bootstrap-vcpkg.sh -disableMetrics && install -spD -m 755 $VCPKG_ROOT/vcpkg /usr/local/bin/
         +vcpkg-base *failed* | Downloading vcpkg-muslc...
         +vcpkg-base *failed* | strip: Unable to recognise the format of the input file `/usr/local/bin/vcpkg'
         +vcpkg-base *failed* | install: strip: No such file or directory
         +vcpkg-base *failed* | ERROR Earthfile line 280:4
         +vcpkg-base *failed* |       The command
         +vcpkg-base *failed* |           RUN sh $VCPKG_ROOT/bootstrap-vcpkg.sh -disableMetrics && install -spD -m 755 $VCPKG_ROOT/vcpkg /usr/local/bin/
         +vcpkg-base *failed* |       did not complete successfully. Exit code 1
```

As long as this runs on x86_64, I do not consider this to block the PR. I can investigate further as needed.

src/libmongoc/doc/index.rst
vector-of-bool: Ideally, everything would be documented with directives that generate appropriate index entries. Since that lone `typedef` is the only one (yet) using a directive, it generates an index entry. Maybe in the future that can be changed. Until then, the `:noindexentry:` option will suffice.

src/libmongoc/doc/learn/get/installing.rst
vector-of-bool: Until its outright deprecated, I think it is worthwhile to keep this. It isn't much to maintain here (only four paragraphs), especially as the feature is probably on the way out and won't be seeing significant modifications

src/libmongoc/doc/learn/get/installing.rst
vector-of-bool: I consider it (generally) good practice to annotate links to external domains, but it would be annoying to have to add/update/maintain those on every links, so adding ` (external)` everyone isn't going to work.

So I look into it, and I discover that Sphinx marks every external-linking-`<a>` with an `external` class. Did a quick test, and it seems simple enough to stylize and attach a psuedo-element to all external links at once. :)

I like this solution, but feel free to veto.

src/libmongoc/doc/ref/packages.rst
vector-of-bool: Yes, these are misplaced. They are supposed to appear before the "`pair:`"

src/libmongoc/examples/cmake/vcpkg/.gitignore
vector-of-bool: Yeah, this file can be removed.

Earthfile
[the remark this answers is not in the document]
vector-of-bool: I believe I found the fix. This was fixed in bootstrap-vcpkg.sh only two days after the release tag that I originally had pinned. I bumped to a newer version that contains the fix.

src/libmongoc/doc/ref/packages.rst
```diff
@@ -0,0 +1,200 @@
...
+    - See: :ref:`ref.vcpkg.features`
+
+
+.. - This prevents the quote-pair rewriting if we use had written two U+0027's directly:
+.. |ditto| unicode:: U+0027 U+0027
+
+.. index:: !Conan; Settings and Features
+.. _ref.conan.settings:
+
+Conan Settings and Features
```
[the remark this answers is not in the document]
vector-of-bool: As far as I can find, those parametres are not described or documented outside of the source itself. Since they correspond to our own build settings, I doubt they will be changing out of lock-step with out own releases. With all that considered, and that we are going to support these tools as installation routes, I decided to document them here. Thoughts?

Either way, the links to the recipes will be useful to include.

src/libmongoc/doc/learn/get/installing.rst
eramongodb: SGTM, as long as it is consistent across the docs. 👍

src/libmongoc/doc/ref/packages.rst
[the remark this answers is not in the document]
kevinAlbs: Considering the options are unlikely to change, documenting here seems OK, and may help readers. Resolving.

Earthfile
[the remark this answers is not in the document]
kevinAlbs: Updating resulted in a different error:

```
 +test-vcpkg-classic | --> RUN vcpkg install mongo-c-driver
 +test-vcpkg-classic | Environment variable VCPKG_FORCE_SYSTEM_BINARIES must be set on arm, s390x, ppc64le and riscv platforms.
 +test-vcpkg-classic | ERROR Earthfile line 254:4
 +test-vcpkg-classic |       The command
 +test-vcpkg-classic |           RUN vcpkg install mongo-c-driver
 +test-vcpkg-classic |       did not complete successfully. Exit code 1

============================= ❌ FAILURE [Build 🔧] ==============================
```

This was resolved applying this change:

```diff
diff --git a/Earthfile b/Earthfile
index 5a4af34f2..2cc34031b 100644
--- a/Earthfile
+++ b/Earthfile
@@ -251,6 +251,7 @@ multibuild:
 #   mongo-c-driver repository.
 test-vcpkg-classic:
     FROM +vcpkg-base
+    ENV VCPKG_FORCE_SYSTEM_BINARIES=1
     RUN vcpkg install mongo-c-driver
     RUN rm -rf _build && \
         make test-classic
```

kevinAlbs (approved):
LGTM with on more possible tweak to support arm in Earthly script.

src/libmongoc/doc/index.rst
```diff
@@ -26,3 +26,9 @@ It depends on :doc:`libbson <bson:index>` to generate and parse BSON documents,
    guides
    api
    application-performance-monitoring
+
+
+Indices and Tables
```
eramongodb, line 31:
```suggestion
Index
```
Should this just be "Index"? Not clear what "Tables" is referring to.

image

src/libmongoc/doc/ref/packages.rst
```diff
@@ -0,0 +1,213 @@
...
+  - - :index:`pacman <!pair: package names; pacman>`
+    - :index:`Arch Linux <!pair: package names; Arch Linux>`
+    - ``mongo-c-driver``
+    - ``mongo-c-driver``
+    - A single package provides both runtime and development support for both
+      |libbson| and |libmongoc|
+
+  - - :index:`Homebrew <!pair: Homebrew; package names>`
+    - :index:`macOS <!pair: macOS; package names>`
+    - |ditto|
```
eramongodb, line 74: Should these `|ditto|`s be changed to `.. empty cell`? Or do they not mean the same thing?

If the latter, rather than `‘’` as the replacement text, suggest "_Same as cell above_" instead for better clarity.

src/libmongoc/doc/static/styles.css
```diff
@@ -107,4 +107,14 @@ aside.footnote> :not(:first-child) {
 /* Pygments does not set an .nv color for body[data-theme="light"], which makes them a bit indistinguishable. */
 body[data-theme="light"] .highlight .nv {
   color: #0004ff;
+}
+
+a.reference.external {
+  text-decoration: underline;
+}
+
```
eramongodb, line 115:
```suggestion
```

Suggest removing the underline since we are already using `⇗` to mark external links. It makes some text ugly and difficult to read.

image

eramongodb (approved):

src/libmongoc/doc/index.rst
vector-of-bool: "Indices and Tables" is the default title that Sphinx generates for this section i.e. when you run `sphinx-quickstart`. I'm actually not sure what "tables" refers to either. Default projects also include a ```:ref:`modinex` ```, but we don't have that, so the title might be superfluous entirely.

kkloberdanz (approved):
Looks good!

PR #1426, opened 2023-10-03
CDRIVER-4342 unskip `/Topology/multiple_selection_errors`

kevinAlbs:
# Summary

- unskip `/Topology/multiple_selection_errors`
- use `.invalid` domain in test to get reliable error

# Background & Motivation

This PR proposes use of the TLD `.invalid` intended to reliably return a DNS error. [RFC 2606](https://www.rfc-editor.org/rfc/rfc2606.html#section-2) notes:

> ".invalid" is intended for use in online construction of domain names that are sure to be invalid and which it is obvious at a glance are invalid.

Similar use of the `.invalid` TLD was similarly done in https://github.com/mongodb/specifications/pull/1136

### DIFF
--- a/src/libmongoc/tests/test-mongoc-topology.c
+++ b/src/libmongoc/tests/test-mongoc-topology.c
@@ -1071,9 +1071,8 @@ test_select_after_try_once (void)
 static void
 test_multiple_selection_errors (void)
 {
-   const char *uri =
-      "mongodb://doesntexist.invalid,example.invalid/?replicaSet=rs"
-      "&connectTimeoutMS=100";
+   const char *const uri = "mongodb://doesntexist.invalid,example.invalid/"
+                           "?replicaSet=rs&connectTimeoutMS=100";
    mongoc_client_t *client;
    bson_t reply;
    bson_error_t error;
--- a/src/libmongoc/tests/test-mongoc-topology.c
+++ b/src/libmongoc/tests/test-mongoc-topology.c
@@ -1069,7 +1069,7 @@ test_select_after_try_once (void)
 
 
 static void
-test_multiple_selection_errors (void *context)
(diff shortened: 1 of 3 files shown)


src/libmongoc/tests/test-mongoc-topology.c
```diff
@@ -1069,16 +1069,15 @@ test_select_after_try_once (void)
...
 
 static void
-test_multiple_selection_errors (void *context)
+test_multiple_selection_errors (void)
 {
-   const char *uri = "mongodb://doesntexist,example.com:2/?replicaSet=rs"
-                     "&connectTimeoutMS=100";
+   const char *uri =
+      "mongodb://doesntexist.invalid,example.invalid/?replicaSet=rs"
+      "&connectTimeoutMS=100";
```
eramongodb, line 1076:
```suggestion
   const char *const uri = "mongodb://doesntexist.invalid,example.invalid/"
                           "?replicaSet=rs&connectTimeoutMS=100";
```
Slightly nicer formatting + const.

eramongodb (commented):
Should the `mongodb://example-localhost:27017/` URIs used by `server_selection_error_dns()` tests also be updated to use `.invalid`?

kevinAlbs:
> Should the `mongodb://example-localhost:27017/` URIs used by `server_selection_error_dns()` tests also be updated to use `.invalid`?

Good catch.

`/server_selection/errors/dns/multi/success/*` tests do DNS lookup on `example-localhost`. `.invalid` has been appended to get a consistent error.

`/server_selection/errors/dns/multi/fail/*` and `/server_selection/errors/dns/direct*` tests generate an error before DNS lookup. Though no DNS lookup is expected, the hosts have been updated for consistency.

PR #1428, opened 2023-10-03
CDRIVER-2838 Document ObjectID unsafe functions

adriandole:


src/libbson/doc/bson_oid_compare_unsafe.rst
```diff
@@ -0,0 +1,9 @@
+:man_page: bson_oid_compare_unsafe
+
+bson_oid_compare_unsafe()
+=========================
+Description
+-----------
+
+Identical to :symbol:`bson_oid_compare()`, but does not check that its arguments are non-null.
```
eramongodb, line 8:
```diff
- Identical to :symbol:`bson_oid_compare()`, but does not check that its arguments are non-null.
+ Identical to :symbol:`bson_oid_compare()`, but performs no integrity checking.
```
For consistency with (some) `bson_iter_*_unsafe()` [comment documentation](https://github.com/mongodb/mongo-c-driver/blob/2baa45ca752ffadd57ed8ea167fbe633db7e044a/src/libbson/src/bson/bson-iter.h#L146) which expands the scope to all preconditions, not just not-null preconditions.

Suggestion applies to other instances in this PR.

src/libbson/doc/bson_oid_compare_unsafe.rst
```diff
@@ -0,0 +1,9 @@
+:man_page: bson_oid_compare_unsafe
+
+bson_oid_compare_unsafe()
+=========================
```
eramongodb, line 4:
```suggestion
=========================

Synopsis
--------

.. code-block:: c

  static inline int
  bson_oid_compare_unsafe (const bson_oid_t *oid1, const bson_oid_t *oid2);
```
Documentation pages should still include declaration of the function.

That being said, this would be the first instance of a `static` and/or `BSON_INLINE` function being included in our public documentation (although we have plenty of as-yet-unused Doxygen comment documentation), so there is no precedence here.

To that end, imo it would be more consistent for the documentation being added in this PR to be implemented as comment documentation instead. @kevinAlbs thoughts?

kevinAlbs: > To that end, imo it would be more consistent for the documentation being added in this PR to be implemented as Doxygen comment documentation instead. @kevinAlbs thoughts?

No strong preference. If a tiebreaker opinion is needed: I have a slight preference to document in RST to include the symbols on https://mongoc.org/. Though use of Doxygen to generate documentation is planned in the near future with CDRIVER-4574.

eramongodb: Then I think I am fine with reducing this comment suggestion to including the declaration of functions in the RST doc pages (with macros like `BSON_INLINE` substituted with their intended expansions, e.g. as done for `BSONCXX_EXPORT(ret-type)` in existing docs).

src/libbson/doc/bson_oid_compare_unsafe.rst
```diff
@@ -0,0 +1,17 @@
...
+
+Synopsis
+--------
+
+.. code-block:: c
+
+  static inline int
+  bson_oid_compare_unsafe (const bson_oid_t *oid1, const bson_oid_t *oid2);
+Description
```
eramongodb, line 13:
```diff
    bson_oid_compare_unsafe (const bson_oid_t *oid1, const bson_oid_t *oid2);
+ 
  Description
```

Add newline after code block and before Description to address the following Doxygen warning:

```
WARNING: Explicit markup ends without a blank line; unexpected unindent.
```

Also applies to other instances in this PR.

src/libbson/doc/bson_oid_compare_unsafe.rst
```diff
@@ -0,0 +1,17 @@
...
+-----------
+
+Identical to :symbol:`bson_oid_compare()`, but performs no integrity checking.
+
```
eramongodb, line 17: Remove extra trailing newline at end of file. Also applies to other instances in this PR.

eramongodb (approved):
Minor suggestions remaining; otherwise, LGTM.

PR #1430, opened 2023-10-04
Ensure Python binary used by Poetry is consistent post-install

eramongodb:
A minor improvement to the `poetry.sh` script that works around unintuitive Poetry Python binary selection behavior post-install. The venv created by Poetry doesn't always use the correct Python binary post-install, even if a specific Python binary was used to run the install script. See the linked PR in the comment for further details.

The separate call to `with-lock` makes me frown, but I could not find a straightforward solution to combine both the Poetry `install` and `run` commands, as `with-lock` [does not support](https://github.com/mongodb/mongo-c-driver/blob/23e375f44249907a44ab043957b722f5d997fd23/tools/with_lock.sh#L6) using shell functions as the executed command, and Poetry itself doesn't support specifying the Python binary in the install command (user requests for `--python <path/to/python>` seem to have been rejected; see linked PR). I don't think this will cause issues in practice (it's not _that_ likely that we run into a race condition here), so proposed as-is, but if there's a more elegant solution here that I missed, I'd be happy to apply it.

### DIFF
--- a/tools/poetry.sh
+++ b/tools/poetry.sh
@@ -70,7 +70,7 @@ ensure-poetry() {
     # See: https://github.com/python-poetry/poetry/issues/522
     with-lock "$POETRY_HOME/.install.lock" \
         env POETRY_HOME="$POETRY_HOME" \
-        "$POETRY_EXE" env use "$POETRY_PYTHON_BINARY" \
+        "$POETRY_EXE" env use --quiet -- "$POETRY_PYTHON_BINARY" \
     || (
         fail "Poetry failed to set Python binary to $POETRY_PYTHON_BINARY"
     )
--- a/tools/poetry.sh
+++ b/tools/poetry.sh
@@ -28,11 +28,14 @@
 #       executed).
 # * WANT_POETRY_VERSION (overridable) (default 1.5.1)
 #     • The version of Poetry that will be installed by run-poetry when executed.
+# * POETRY_PYTHON_VERSION (overridable) (default to result of find-python)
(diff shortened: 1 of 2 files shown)


kevinAlbs (approved):

tools/poetry.sh
```diff
@@ -45,13 +45,22 @@ install-poetry() {
...
         cat -- poetry-installer*.log && fail "Poetry installation failed"
     )
+    # Extra step must be taken to ensure Poetry's virtual environment continues to use the correct Python binary.
+    # See: https://github.com/python-poetry/poetry/issues/522#issuecomment-492369712
+    with-lock "$POETRY_HOME/.install.lock" \
+        env POETRY_HOME="$poetry_home" \
+        "$POETRY_EXE" env use "$python_binary" \
+    || (
+        fail "Poetry installation failed (env use <path/to/python>)"
+    )
```
vector-of-bool, line 63: Having a separate `with-lock` is safe, as the `env use` subcommand applies to only the current project (from `pwd` or from `-C`), so concurrent invocations in separate project directories will not contend. It also runs very quickly as all it does is splat some files down.

This also means, however, that this step needs to run any time that Poetry will create a new virtualenv, and not just once when it is installed. Because it's fast, idempotent, and cheap, it would suffice to run `poetry env use --quiet <python>` before every Poetry command.

vector-of-bool (changes requested):
This is almost correct, but will probably break subtly in the future. See comment for details.

vector-of-bool (approved):

PR #1438, opened 2023-10-10
CDRIVER-4346 unskip `/BulkOperation/split`

kevinAlbs:
# Summary

- Unskip `/BulkOperation/split`.
- Use "read-your-own-writes" semantics in test to prevent undercounting.

# Background & Motivation

Three repeated runs of the `/BulkOperation/split` test on all variants and tasks with names matching `".*test.*` did not result in the test failing: https://spruce.mongodb.com/version/651eaee656234357657a0c22

Changes in this PR intend to address a possible cause of the original reported error. The original error reported an undercount after inserting: `Assert Failure: count of split_1512376901_50824 is 97759, not 100010`.

This PR proposes using "read-your-own-writes" semantics to read the count. This is done by using a session with causal consistency, and performing operations with majority read and write concerns described in [Client Sessions and Causal Consistency Guarantees](https://www.mongodb.com/docs/manual/core/read-isolation-consistency-recency/#client-sessions-and-causal-consistency-guarantees).

Updated test was tested in this patch build: https://spruce.mongodb.com/version/65255627850e61e7577a509b

### DIFF
--- a/src/libmongoc/tests/test-mongoc-bulk.c
+++ b/src/libmongoc/tests/test-mongoc-bulk.c
@@ -3498,12 +3498,21 @@ test_bulk_split (void)
    int n_docs;
    int i;
    uint32_t r;
+   mongoc_client_session_t *session;
 
    /* ensure we need two batches */
    n_docs = (int) test_framework_max_write_batch_size () + 10;
 
    client = test_framework_new_default_client ();
    BSON_ASSERT (client);
+
+   // Start a session.
+   {
+      mongoc_session_opt_t *sopts = mongoc_session_opts_new ();
+      mongoc_session_opts_set_causal_consistency (sopts, true);
+      session = mongoc_client_start_session (client, sopts, NULL);
+      mongoc_session_opts_destroy (sopts);
+   }
 
    collection = get_test_collection (client, "split");
    BSON_ASSERT (collection);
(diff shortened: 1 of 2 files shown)


src/libmongoc/tests/test-mongoc-bulk.c
```diff
@@ -3548,7 +3575,20 @@ test_bulk_split (void)
    BSON_ASSERT (!r);
 
    /* all 100,010 docs were inserted, either by the first or second bulk op */
-   ASSERT_COUNT (n_docs, collection);
+   {
+      bson_t count_opts = BSON_INITIALIZER;
+      ASSERT_OR_PRINT (
+         mongoc_client_session_append (session, &count_opts, &error), error);
+      int64_t count = mongoc_collection_count_documents (collection,
```
eramongodb, line 3582: Confirming, is this the only test using `ASSERT_COUNT` that must be updated to ensure causal consistency because it is the only(?) test that whose assertion depends on multiple bulk write operations?

kevinAlbs: > Confirming, is this the only test using `ASSERT_COUNT` that must be updated to ensure causal consistency because it is the only(?) test that whose assertion depends on multiple bulk write operations?

I expect many tests implicity assume "read-your-own-writes" semantics without guaranteeing the behavior (e.g. any test that inserts and expects to find the document). `/BulkOperation/split` may be more likely to result in a test failure due to the large write (100,000 documents) taking longer to replicate. 

I expect the cause of the original error is the count operation runs on a different server which has not fully replicated the inserts. Alternatively: this PR can unskip the test without changes and wait for a possible future failure. Inspecting the server logs on a task failure may help to confirm this diagnosis.

eramongodb: I am fine with proposed changes as-is.

eramongodb (approved):

PR #1442, opened 2023-10-12
CDRIVER-4736 add an api to set custom sleep function 

itrofimow:


kevinAlbs:
Closing due to deciding not to include this change in public API.

src/libmongoc/src/mongoc/mongoc-util.h
```diff
@@ -0,0 +1,34 @@
...
+ * @user_data: User data provided to mongoc_usleep_set_impl().
+ */
+typedef void (*mongoc_usleep_func_t) (int64_t usec, void *user_data);
+
+/**
+ * mongoc_usleep_set_impl:
+ * @usleep_func: A function to perform microsecond sleep.
+ *
+ * Sets the function to be called to perform sleep.
+ * Not thread-safe.
```
kevinAlbs, line 24:
```suggestion
 * Not thread-safe.
 * Providing a `usleep_func` that does not sleep (e.g. coroutine suspension) is
 * not supported. Doing so is at the user's own risk.
```

At present, suspending threads before blocking work is not a supported use-case of libmongoc. This note is intended to advise users of the risk.

kevinAlbs (changes requested):

src/libmongoc/CMakeLists.txt
```diff
@@ -705,6 +705,7 @@ set (HEADERS
    ${PROJECT_SOURCE_DIR}/src/mongoc/mongoc-stream-socket.h
    ${PROJECT_SOURCE_DIR}/src/mongoc/mongoc-topology-description.h
    ${PROJECT_SOURCE_DIR}/src/mongoc/mongoc-uri.h
+   ${PROJECT_SOURCE_DIR}/src/mongoc/mongoc-util.h
```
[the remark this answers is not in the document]
itrofimow: done

src/libmongoc/tests/test-mongoc-usleep.c
```diff
@@ -16,8 +17,28 @@ test_mongoc_usleep_basic (void)
...
+static void
+test_mongoc_usleep_custom (void)
+{
+   static const int64_t expected = 42;
+   int64_t last_sleep_dur = -1;
+
+   mongoc_usleep_set_impl(custom_usleep_impl, &last_sleep_dur);
+
+   _mongoc_usleep (expected);
+   ASSERT_CMPINT64 (last_sleep_dur, =, expected);
```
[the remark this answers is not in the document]
itrofimow: done

src/libmongoc/src/mongoc/mongoc-util.h
```diff
@@ -0,0 +1,34 @@
...
+ */
+MONGOC_EXPORT (void)
+mongoc_usleep_set_impl (mongoc_usleep_func_t usleep_func, void *user_data);
+
```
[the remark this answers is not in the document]
itrofimow: done

src/libmongoc/src/mongoc/mongoc-util.h
itrofimow: done

itrofimow:
Hi @kevinAlbs! I've addressed your comments, could you please have a look?

kevinAlbs (approved):

src/libmongoc/src/mongoc/mongoc-sleep.h
vector-of-bool, line 1: IIUC, the only place that needs to do a sleep is in `mongoc_topology_select_server_id` (other instances are for testing purposes?). Would it be feasible to attach the sleep callback to the `mongoc_topology_t` rather than having a global?

kevinAlbs: I think so. And I prefer setting the custom sleep on `mongoc_client_t` to avoid a global override.
Testing is a bit trickier. Here is a [branch](https://github.com/mongodb/mongo-c-driver/compare/master...kevinAlbs:mongo-c-driver:libmongoc_custom_sleep?expand=1) with commits scoping the custom sleep to `mongoc_client_t`.

@itrofimow does setting the custom sleep on `mongoc_client_t` satisfy the use case? If so, I suggest applying https://github.com/mongodb/mongo-c-driver/commit/37152b2db770c670286cfe93721a23cea1c69e5c and https://github.com/mongodb/mongo-c-driver/commit/3fd794f4c5ca1e22c94555f1188cba42e2342c67 to scope the sleep to a `mongoc_client_t`.

itrofimow: It does.

I don't have a strong opinion on this, but i'm a bit worried that scoping the sleep to a client creates potentials risks for my use-case and ads a maintenance burden for libmongoc developers:

consider following usage
https://github.com/mongodb/mongo-c-driver/blob/3b68632c10634013ce690028ab07467f70525680/src/libmongoc/src/mongoc/mongoc-async.c#L130-L134
if the logic here changes (or another place needs to sleep) i find it likely for people to reach to `_mongoc_usleep` instead of the sleep callback stored somewhere potentially very high up the call chain.

In my opinion this is a library-wide change and it shouldn't necessary be scoped to the client, since it doesn't have a direct logical relation to the client itself.

What do you think? @kevinAlbs @vector-of-bool

vector-of-bool: Since `_mongoc_usleep` is an internal function, I think it safe to delete/rename it to avoid future temptation to use it. I think attaching the custom sleep to the client will be better for maintenance in the long run.

itrofimow: I've cherry-picked https://github.com/mongodb/mongo-c-driver/commit/37152b2db770c670286cfe93721a23cea1c69e5c and https://github.com/mongodb/mongo-c-driver/commit/3fd794f4c5ca1e22c94555f1188cba42e2342c67 to scope custom usleep implementation to `mongoc_client_t`.

vector-of-bool (approved):
Thank you for your patience with coordinating these changes. LGTM!

PR #1445, opened 2023-10-16
CDRIVER-3514 Fix Windows shared mutex unlocking

kkloberdanz:
Unlike a POSIX read/write lock, in which both a read lock and a write lock are unlocked by the same function, Windows requires that a read lock is unlocked by a different function than a write lock.


kevinAlbs (approved):
Good catch. LGTM

PR #1460, opened 2023-10-31
CDRIVER-4759 Define `Version` in `calc_release_version.py`.

kevinAlbs:
# Summary

- Define `Version` in `calc_release_version.py`.

Changes were tested with this patch build: https://spruce.mongodb.com/version/6541536f7742aeebaeef9c1b/

# Background & Motivation

Intended to resolve recently observed task failures. [Example](https://spruce.mongodb.com/task/mongo_c_driver_sasl_matrix_darwinssl_sasl_cyrus_darwinssl_macos_1100_arm64_clang_compile_25f6bdf938050bf3506c75b7fc8eb7f20b3b3d52_23_10_31_15_18_42/logs?execution=0):

```
  File "/System/Volumes/Data/data/mci/1caa136a8a866fee5ec81f49bed8c650/mongoc/build/cmake/../calc_release_version.py", line 52, in <module>
    from packaging.version import Version as Version
ModuleNotFoundError: No module named 'packaging'
During handling of the above exception, another exception occurred:
  File "/System/Volumes/Data/data/mci/1caa136a8a866fee5ec81f49bed8c650/mongoc/build/cmake/../calc_release_version.py", line 57, in <module>
    from pkg_resources.extern.packaging.version import Version  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'pkg_resources'
During handling of the above exception, another exception occurred:
  File "/System/Volumes/Data/data/mci/1caa136a8a866fee5ec81f49bed8c650/mongoc/build/cmake/../calc_release_version.py", line 61, in <module>
    from distutils.version import LooseVersion as Version
ModuleNotFoundError: No module named 'distutils'
CMake Error at build/cmake/BuildVersion.cmake:43 (message):
  Computing the build version failed! [1]:
  build/cmake/BuildVersion.cmake:50 (compute_build_version)
  CMakeLists.txt:5 (include)
```

This is assumed to be caused by a change in Python 3.12.0. Python 3.12.0 [release notes](https://docs.python.org/3/whatsnew/3.12.html) include:

> This means that `distutils`, `setuptools`, `pkg_resources`, and `easy_install` will no longer available by default

This PR adds a simple `Version` class into the `calc_release_version.py` to remove the dependency.

I prefer to merge this before the 1.25.0 release to resolve possible downstream build issues with Python 3.12.0.


rcsanchez97 (approved):
That is strange that it would break entirely. I guess this is what makes progress somewhat painful.

LGTM.

PR #1466, opened 2023-11-03
CDRIVER-6357 Use git-subrepo for the tools/ directory

vector-of-bool:
These changes are just an "maybe something interesting to pursue", and not a finalized proposal. This should help us consolidate scripts (CMake, shell, Python, etc.) between libmongocrypt, mongocxx, and mongoc without manually copy-pasting changes.

See `README.md` for details

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -1,0 +1,90 @@
+# `mongo-c-tools`
+
+This directory contains shared scripts and tools for native C and C++
+development.
+
+
+## As a Subrepo
+
+If this directory contains a `.gitrepo` file, then this directory is being
+tracked as a *subrepo* of some parent repository. If this file does not exist
+but a `.git/` directory *does* exist, then you are looking at the tools
+repository itself.
+
+[`git subrepo`](https://github.com/ingydotnet/git-subrepo) is a set of scripts
+and conventions for pushing/pulling/syncing changes in a subdirectory between
+different repositories with full history tracking. *Using* the contents of this
+directory does not require the `git subrepo` tool, but pushing and pulling
+changes in this subrepo directory will require `git subrepo` (or knowing
+the appropriate commands to execute for the same effect).
+
+If you are simply using these scripts, you do not have to deal with the subrepo
+commands. If you wish to modify/sync these scripts, read on below.
+
+
+### Preparing
+
+`git-subrepo` tracks a subrepo directory using a remote+branch pair as the
+"source of truth" for the subrepo contents. The remote+branch pair is fixed in
+the `.gitrepo` file. In this subrepo directory, the `remote` is not set to a
+URL, but to a name `tools-origin`. In order to push/pull, you will need to
+ensure that you have a `tools-origin` set to the proper URL. Unless you "know
+what you are doing", you will want to use the upstream `mongo-c-driver`
+repository as the remote:
+
+```sh
+$ git remote add tools-origin "git@github.com:mongodb/mongo-c-driver.git"
+```
+
+The content of this subrepo is contained by the disjoint branch `mongo-c-tools`,
+which contains only the history of this directory as the root.
+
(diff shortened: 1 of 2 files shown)


tools/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Preparing
+
+`git-subrepo` tracks a subrepo directory using a remote+branch pair as the
+"source of truth" for the subrepo contents. The remote+branch pair is fixed in
+the `.gitrepo` file. In this subrepo directory, the `remote` is not set to a
+URL, but to a name `tools-origin`. In order to push/pull, you will need to
+ensure that you have a `tools-origin` set to the proper URL. Unless you "know
+what you are doing", you will want to use the upstream `mongo-c-driver`
+repository as the remote:
```
eramongodb, line 33: For rationale documentation purposes, why `mongodb/mongo-c-driver` as the remote rather than a separate `mongodb(-labs)/mongo-c-driver-tools` repo (a la `mongodb-labs/drivers-evergreen-tools`)? I find the use of the "disjoint branch" to be strange. It is effectively already behaving as a separate git repo that is using an [arbitrary commit](https://github.com/mongodb/mongo-c-driver/commit/9aa60260db9d9c45a0ae3c5bc65e2adbba1a5879) as a common ancestor with the main branch. If we want to use the "disjoint branch" method, might as well go all-in and remove the common ancestor entirely (we won't ever be merging the `mongo-c-tools` branch itself into the main branch anyways, right?).

vector-of-bool: No reason at all other than "I didn't bother to create a new repo for it" :). At `$prior_job` we used a separate repo for the subrepo target, and it works great with that. It'd probably actually be better, since there wouldn't be need to juggle the remote URL.

vector-of-bool: Oh also I don't have the permission to create a new repo in the org, so I just used what I had available.

rcsanchez97: We already have a project at `10gen/mongo-c-driver-tools.git` which might be a good candidate, assuming that a project which isn't public is acceptable in this case.

vector-of-bool: That could very well work. It does not need to be a publicly accessible repo.

tools/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Modifying these Files
+
+Before modifying the contents of this directory, be sure that you have the
+latest changes from the remote branch according to the
+[*pulling changes*](#pulling-changes) section. This will avoid annoying merge
+conflicts.
+
+To modify the contents of this directory, simply update and commit them as one
+would do normally. If you want your changes to undergo review:
```
kevinAlbs, line 76: A pattern I find helpful when updating drivers-evergreen-tools is to test the changes in a driver by cloning the updated branch. Example: https://github.com/mongodb-labs/drivers-evergreen-tools/pull/359 was tested in the C driver with an Evergreen patch of the C driver with temporary changes to clone the branch in review:

```
git clone --depth=1 git@github.com:kevinAlbs/drivers-evergreen-tools.git --branch pykmip_failure
```

If `mongo-c-driver-tools` is used, consider adding instructions for testing changes in review in the C driver, C++ driver, and/or libmongocrypt. Something like the following:

```markdown
To test changes in review, scripts may be updated from a separate remote branch with:
`git subrepo pull --remote git@github.com:$user/mongo-c-driver-tools.git --branch $branch_in_review $this_subdir`
```

tools/README.md
kevinAlbs: I support use of `10gen/mongo-c-driver-tools`. AFAIK the only actively used script in the repo is `release.py`.

kevinAlbs (commented):
I support trying `git subrepo` to share scripts. I agree with using a separate repo, rather than a disjoint branch.

PR #1474, opened 2023-11-09
Drop Debian Sphinx patch

kkloberdanz:


rcsanchez97 (changes requested):
This needs an new entry in `debian/changelog` along these lines:

```
  * drop 0002_sphinx_api_fix.diff (fixed upstream)
```

BTW, the "right" version to use is 1.25.1-2 in this instance, but leave it marked `UNRELEASED`.

rcsanchez97 (approved):
LGTM

PR #1475, opened 2023-11-13
CDRIVER-4781 Avoid intermediate uint32_t for socketTimeoutMS

eramongodb:
Resolves CDRIVER-4781. Verified by [this patch](https://spruce.mongodb.com/version/65525bff0305b9e2e2b97975).

https://github.com/mongodb/mongo-c-driver/pull/1229 introduced range representability checks for conversion from `int64_t` -> `int32_t` timeout parameters in internal read/write utility functions, where due to historical reasons, `int64_t` timeout parameters for several read/write operations are forwarded to underlying `int32_t` timeout parameters (narrowing conversion). The new range checks are (correctly but unexpectedly) triggered by `mongoc_cluster_t::sockettimeoutms` when the URI option is set to `-1`, which has also been (implicitly) supported for historical reasons. The negative value was being implicitly converted to a large unsigned value before being passed as an `int64_t`, before being converted back to `-1` when cast back to `int32_t`:

```
-1         (int32_t: MONGOC_URI_SOCKETTIMEOUTMS)
4294967295 (uint32_t: mongoc_cluster_t::sockettimeoutms)
4294967295 (int64_t: _mongoc_stream_writev_full())
  >int32_t range check failure< (the breaking change)
-1         (int32_t: mongoc_stream_writev(), impl-defined narrowing signed conversion)
```

This PR restores support for negative values for the `socketTimeoutMS` URI option for backward compatibility by replacing the intermediate `uint32_t` data members of `mongoc_cluster_t` with `int32_t` for better consistency with how the values are both assigned and used internally. This avoids the implicit conversion into a large unsigned value that triggers the range check before conversion back to an `int32_t`. This change also addresses 5 GCC and 5 Clang warnings for implicit sign conversion. Other instances of internal representation of timeout values (that aren't ignored, i.e. gridfs functions) do not seem to be affected: there are no other instances of intermediately-representing the timeout as an unsigned integer.

A regression test is added that ensures the range check is not triggered by a negative socketTimeoutMS URI value. This regression test + git bisect verified that the breaking change was triggered by https://github.com/mongodb/mongo-c-driver/commit/abb0bb7da8230d342dac5061623c1e4936ed46d9 (https://github.com/mongodb/mongo-c-driver/pull/1229).

As a drive-by fix, this PR also fixes the incorrect format specifier for error messages concerning `timeout_ms` (`PRId64` for `int64_t`).

Additionally, an attempt is made to document the unspecified/inconsistent behavior of non-positive timeout values for the URI option documentation and for functions that have a timeout parameter (i.e. `mongoc_stream_*()`). Scanning through how the timeout values are both represented and interpreted within libmongoc, it does not seem desirable to continue to support non-positive URI timeout values (even `0`) despite historical (implicit) support.

### DIFF
--- a/src/libmongoc/doc/mongoc_uri_t.rst
+++ b/src/libmongoc/doc/mongoc_uri_t.rst
@@ -104,8 +104,6 @@ MONGOC_URI_LOADBALANCED                    loadbalanced                      fal
 MONGOC_URI_SRVMAXHOSTS                     srvmaxhosts                       0                                 If zero, the number of hosts in DNS results is unlimited. If greater than zero, the number of hosts in DNS results is limited to being less than or equal to the given value.
 ========================================== ================================= ================================= ============================================================================================================================================================================================================================================
 
-Setting any of the \*timeoutMS options above to ``0`` will be interpreted as "use the default value".
-
 .. warning::
 
   Setting any of the \*timeoutMS options above to either ``0`` or a negative value is discouraged due to unspecified and inconsistent behavior.
--- a/src/libmongoc/doc/mongoc_uri_t.rst
+++ b/src/libmongoc/doc/mongoc_uri_t.rst
@@ -106,6 +106,13 @@ MONGOC_URI_SRVMAXHOSTS                     srvmaxhosts                       0  
 
 Setting any of the \*timeoutMS options above to ``0`` will be interpreted as "use the default value".
 
+.. warning::
+
+  Setting any of the \*timeoutMS options above to either ``0`` or a negative value is discouraged due to unspecified and inconsistent behavior.
+  The "default value" historically specified as a fallback for ``0`` or a negative value is NOT related to the default values for the \*timeoutMS options documented above.
+  The meaning of a timeout of ``0`` or a negative value may vary depending on the operation being executed, even when specified by the same URI option.
+  To specify the documented default value for a \*timeoutMS option, use the `MONGOC_DEFAULT_*` constants defined in ``mongoc-client.h`` instead.
+
 Authentication Options
 ----------------------
 
@@ -330,4 +337,3 @@ MONGOC_URI_SAFE                            safe                              {tr
     mongoc_uri_set_username
     mongoc_uri_set_write_concern
     mongoc_uri_unescape
-
--- a/src/libmongoc/doc/mongoc_stream_read.rst
+++ b/src/libmongoc/doc/mongoc_stream_read.rst
@@ -26,6 +26,13 @@ Parameters
 
 The :symbol:`mongoc_stream_read()` function shall perform a read from a :symbol:`mongoc_stream_t`. It's modeled on the API and semantics of ``read()``, though the parameters map only loosely.
 
+.. warning::
+
+  The "default timeout" indicated by a negative value is both unspecified and
+  unrelated to the documented default values for ``*TimeoutMS`` URI options.
+  To specify a default timeout value for a ``*TimeoutMS`` URI option, use the
+  ``MONGOC_DEFAULT_*`` constants defined in ``mongoc-client.h``.
+
 Returns
 -------
 
@@ -38,4 +45,3 @@ The :symbol:`mongoc_stream_read` function returns the number of bytes read on su
   | :symbol:`mongoc_stream_write()`
 
   | :symbol:`mongoc_stream_writev()`
-
--- a/src/libmongoc/doc/mongoc_stream_readv.rst
+++ b/src/libmongoc/doc/mongoc_stream_readv.rst
@@ -26,6 +26,13 @@ Parameters
 
 This function is identical to :symbol:`mongoc_stream_read()` except that it takes a :symbol:`mongoc_iovec_t` to perform gathered I/O.
 
+.. warning::
+
+  The "default timeout" indicated by a negative value is both unspecified and
+  unrelated to the documented default values for ``*TimeoutMS`` URI options.
+  To specify a default timeout value for a ``*TimeoutMS`` URI option, use the
+  ``MONGOC_DEFAULT_*`` constants defined in ``mongoc-client.h``.
+
 Returns
 -------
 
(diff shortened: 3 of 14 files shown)


src/libmongoc/doc/mongoc_uri_t.rst
```diff
@@ -106,6 +106,13 @@ MONGOC_URI_SRVMAXHOSTS                     srvmaxhosts                       0
 
 Setting any of the \*timeoutMS options above to ``0`` will be interpreted as "use the default value".
 
+.. warning::
+
+  Setting any of the \*timeoutMS options above to either ``0`` or a negative value is discouraged due to unspecified and inconsistent behavior.
```
eramongodb, line 111: Not sure how to reconcile the documentation above:

> Setting any of the \*timeoutMS options above to ``0`` will be interpreted as "use the default value".

With the documentation/use of timeout parameters:

> The number of milliseconds to wait before failure, a timeout of 0 will not block.

Therefore, included value of `0` in the new warning paragraph.

src/libmongoc/doc/mongoc_uri_t.rst
```diff
@@ -106,6 +106,13 @@ MONGOC_URI_SRVMAXHOSTS                     srvmaxhosts                       0
...
+  The "default value" historically specified as a fallback for ``0`` or a negative value is NOT related to the default values for the \*timeoutMS options documented above.
```
eramongodb, line 112: > is NOT related to the default values for the \*timeoutMS options documented above.

This is due to the observation that the "default value" set given a negative timeout [appears to be](https://github.com/mongodb/mongo-c-driver/blob/fc1cba77b671f5c8c942d450a20c7c306fe909d4/src/libmongoc/src/mongoc/mongoc-stream.c#L160) with the undocumented `MONGOC_DEFAULT_TIMEOUT_MSEC` constant which appears to have no direct relation to any of the documented URI default values. Furthermore, this is specific to "write" paths; it is unclear to me how a negative value is interpreted in the "read" paths, which have [no equivalent check](https://github.com/mongodb/mongo-c-driver/blob/fc1cba77b671f5c8c942d450a20c7c306fe909d4/src/libmongoc/src/mongoc/mongoc-stream.c#L237) for negative timeouts before invoking `readv()`.

src/libmongoc/doc/mongoc_uri_t.rst
kevinAlbs: I suggest removing:
> Setting any of the \*timeoutMS options above to ``0`` will be interpreted as "use the default value".

I expect the actual behavior does not match. Setting `socketTimeoutMS=0` passes 0 for the timeout to `readv` and `writev`, not the value for `MONGOC_DEFAULT_SOCKETTIMEOUTMS`.

src/libmongoc/doc/mongoc_uri_t.rst
kevinAlbs: I expect the timeout value passed to `writev` needs to be preserved for backwards compatibility.
A custom `mongoc_stream_t` implementation can be set in `mongoc_client_set_stream_initiator`. An implementation may not expect negative values passed in `writev`.

Since negatives are now documented as "unspecified and inconsistent", consider adding a test to ensure the `writev` handling of negatives is kept for backwards compatibility:

```c
typedef struct {
   mongoc_stream_t vtable;
   struct {
      bool set;
      uint32_t value;
   } last_timeout_applied;
} timeout_tracker_stream_t;

static ssize_t
timeout_tracker_stream_writev (mongoc_stream_t *stream,
                               mongoc_iovec_t *iov,
                               size_t iovcnt,
                               int32_t timeout_msec)
{
   timeout_tracker_stream_t *ttstream = (timeout_tracker_stream_t *) stream;

   BSON_UNUSED (iov);
   BSON_UNUSED (iovcnt);
   ttstream->last_timeout_applied.set = true;
   ttstream->last_timeout_applied.value = timeout_msec;
   return 0;
}

void
timeout_tracker_stream_destroy (mongoc_stream_t *stream)
{
   bson_free (stream);
}

static timeout_tracker_stream_t *
timeout_tracker_stream_new (void)
{
   timeout_tracker_stream_t *ttstream;

   ttstream = bson_malloc0 (sizeof *ttstream);
   ttstream->vtable.type = 999;
   ttstream->vtable.writev = timeout_tracker_stream_writev;
   ttstream->vtable.destroy = failing_stream_destroy;

   return ttstream;
}

// Test special handling of negative write timeout in mongoc_stream_t API.
// Negative values in `mongoc_stream_writev` are applied as a private default.
// This behavior is tested for backwards compatibility with existing mongoc_stream_t implementations.
static void
test_stream_negative_write_timeout (void)
{
   char buf[1] = {'a'};
   mongoc_iovec_t iov[1] = {{.iov_base = buf, .iov_len = sizeof (buf)}};
   timeout_tracker_stream_t *ttstream = timeout_tracker_stream_new ();

#define MONGOC_DEFAULT_TIMEOUT_MSEC (60L * 60L * 1000L)

   // writev with -1 timeout applies a private "default".
   ttstream->last_timeout_applied.set = false;
   mongoc_stream_writev ((mongoc_stream_t *) ttstream,
                         iov,
                         sizeof (iov) / sizeof (iov[0]),
                         -1 /* timeout */);
   ASSERT (ttstream->last_timeout_applied.set);
   ASSERT_CMPINT32 (
      ttstream->last_timeout_applied.value, ==, MONGOC_DEFAULT_TIMEOUT_MSEC);

#undef MONGOC_DEFAULT_TIMEOUT_MSEC

   mongoc_stream_destroy ((mongoc_stream_t *) ttstream);
}
```

kevinAlbs (approved):
LGTM with minor suggestions.

src/libmongoc/doc/mongoc_uri_t.rst
eramongodb: Removed as requested.

src/libmongoc/doc/mongoc_uri_t.rst
eramongodb: Added a writev test as requested.

PR #1478, opened 2023-11-26
CDRIVER-4746 update x509 certificates

rcsanchez97:
This updates the x509 certificates used for our SSL/TLS-client certificate authentication test to match the new certificates which have just been deployed to replace the about-to-expire certificates. Note that this PR does not re-enable the auth test on macOS. The new certificates do not meet the requirements of macOS and so the test still fails there. I have addressed the issue to Zakhar in DEVPROD-1209.

Full Evergeen patch build: https://spruce.mongodb.com/version/6563cb357742ae45ccf67875


kevinAlbs (approved):

rcsanchez97:
I have been communicating with Zakhar about revising the certificates so that macOS will like them. I expect the updated certificates Real Soon Now(TM), and once those are deployed, I will re-test with macOS and, assuming it works, push one final commit to re-enable the test on macOS.

PR #1479, opened 2023-11-27
libbson: prevent -Werror=conversion with GCC 12

rohieb:
Building on ARM32 fails with GCC 12.3:

    bson-iter.h:434:33: error: conversion from 'int64_t' {aka 'long long int'} to '__suseconds_t' {aka 'long int'} may change value [-Werror=conversion]
    434 | tv->tv_usec = (value % 1000) * 1000;
    | ~~~~~~~~~~~~~~~^~~~~~
    cc1plus: all warnings being treated as errors

Do the same as with tv->tv_sec, and explicitely cast it to suseconds_t on non-Win32 systems and to long on Win32.

### DIFF
--- a/src/libbson/src/bson/bson-iter.h
+++ b/src/libbson/src/bson/bson-iter.h
@@ -440,10 +440,11 @@ bson_iter_timeval_unsafe (const bson_iter_t *iter, struct timeval *tv)
    int64_t value = bson_iter_int64_unsafe (iter);
 #ifdef BSON_OS_WIN32
    tv->tv_sec = (long) (value / 1000);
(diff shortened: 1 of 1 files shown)


rohieb:
> Applied [c81e4ce](https://github.com/mongodb/mongo-c-driver/commit/c81e4ce24383ca8488a5d5a05d7f288de8cda746) to cast the assigned value to `tv_sec` to `time_t` to match what is specified in POSIX 2008: https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/

Ah yes, thanks, looks correct to me!

PR #1481, opened 2023-11-27
CDRIVER-4790 AIX has a glibc-compatible strerror_r w/ diff name

NattyNarwhal:
The AIX strerror_r doesn't have the same semantics as glibc's (diff type signature), but it does provide one under a different name. This name is used if you use -D_LINUX_SOURCE_COMPAT, but it's a big hammer. We can just use it directly.

### DIFF
--- a/src/libbson/src/bson/bson-error.c
+++ b/src/libbson/src/bson/bson-error.c
@@ -116,6 +116,10 @@ bson_strerror_r (int err_code,                    /* IN */
    if (strerror_s (buf, buflen, err_code) != 0) {
       ret = buf;
    }
+#elif defined(_AIX)
+   // AIX does not provide strerror_l, and its strerror_r isn't glibc's.
+   // But it does provide a glibc compatible one called __linux_strerror_r
+   ret = __linux_strerror_r (err_code, buf, buflen);
 #elif defined(__APPLE__)
    // Apple does not provide `strerror_l`, but it does unconditionally provide
    // the XSI-compliant `strerror_r`, but only when compiling with Apple Clang.


kkloberdanz (approved):
Thank you for the contribution!

Since AIX is not officially, supported, I wanted to give some disclaimer information: We do not officially support AIX, and we also do not test on AIX, so we cannot guarantee that the driver will behave correctly on AIX. We also cannot guarantee that the AIX build will not be broken in the future. However, we are open to accepting small fixes like this if the build ever does break in the future.

Once again, thanks for the contribution!

src/libbson/src/bson/bson-error.c
```diff
@@ -116,6 +116,10 @@ bson_strerror_r (int err_code,                    /* IN */
    if (strerror_s (buf, buflen, err_code) != 0) {
       ret = buf;
    }
+#elif defined(_AIX)
+   // AIX does not provide strerror_l, and its strerror_r isn't glibc's.
+   // But it does provide a glibc compatible one called __linux_strerror_r
+   ret = __linux_strerror_r (err_code, buf, buflen);
```
jmikola, line 122: @NattyNarwhal: did you guys only come across this after upgrading to libmongoc 1.24? It looks like it may be related to the recent changes in [CDRIVER-4679](https://jira.mongodb.org/browse/CDRIVER-4679), which were prompted by [PHPC-2263](https://jira.mongodb.org/browse/PHPC-2263).

NattyNarwhal: I finally had the opportunity to upgrade from whatever the vendored version in 1.12 of the PHP driver was to whatever the version in 1.16 was, so I'd be running in context of that release of the PHP driver's configure script.

jmikola: Thanks. If you jumped from 1.12 to 1.16 I think it's safe to assume this issue was introduced by the aforementioned fixes for Alpine Linux. :v:

PR #1483, opened 2023-11-30
CDRIVER-4788 do not require sphinx-design for man pages

kevinAlbs:
# Summary

Do not require `sphinx-design` to build man pages.

Add a `make-docs-manonly` task to ensure man pages can be built without `sphinx-design`.

Verified with this patch build: https://spruce.mongodb.com/version/6568b9d12fbabeee1a08b52e

# Background & Motivation

The Fedora package configures to install man pages [with `ENABLE_MAN_PAGES`](https://src.fedoraproject.org/rpms/mongo-c-driver/blob/1defc9d660f0b672eaa3853adcd31061d15a55f6/f/mongo-c-driver.spec#_130).

CDRIVER-4767 reports EPEL does not currently include `sphinx-design`. Building man pages without `sphinx-design` results in the error `ModuleNotFoundError: No module named 'sphinx_design'` when built as follows:

```bash
cmake \
    -DENABLE_MAN_PAGES:BOOL=ON \
    -S . -B cmake-build
cmake --build cmake-build --target mongoc-man bson-man
```

This PR removes the requirement of `sphinx-design` to build man pages. The directives requiring sphinx-design (`tab-set` and `ad-dropdown`) do not appear in any documentation built for man pages (containing `:man_page:`). Instead, an error is raised if attempting to build HTML without `sphinx-design`.

The proposed `make-docs-manonly` task expectedly [fails](https://spruce.mongodb.com/task/mongo_c_driver_smoke_make_docs_manonly_patch_5e27ad6b21f1cba6c08120e6123f317bbfc67e5c_6568b7f5d1fe0772c2a1fe11_23_11_30_16_27_34/logs?execution=0) without changes to conditionally import `sphinx-design`.

### DIFF
--- a/.evergreen/config_generator/components/make_docs.py
+++ b/.evergreen/config_generator/components/make_docs.py
@@ -27,6 +27,26 @@ class MakeDocs(Function):
     def call(cls, **kwargs):
         return cls.default_call(**kwargs)
 
+class MakeDocsManOnly(Function):
+    name = "make-docs-manonly"
+    commands = [
+        bash_exec(
+            command_type=EvgCommandType.TEST,
+            working_dir="mongoc",
+            include_expansions_in_env=["distro_id"],
+            script="""\
+                set -o errexit
+                # Build man pages only using the dependencies required for man pages.
+                bash tools/poetry.sh install --with=docs-manonly
+                bash tools/poetry.sh run bash .evergreen/scripts/build-docs-manonly.sh
+                """,
+        ),
+    ]
+
+    @classmethod
+    def call(cls, **kwargs):
+        return cls.default_call(**kwargs)
+
 
 class UploadDocs(Function):
     name = "upload-docs"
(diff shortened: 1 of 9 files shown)


.evergreen/etc/spec.patch
```diff
@@ -18,3 +18,13 @@
  Release:   1%{?dist}
  # See THIRD_PARTY_NOTICES
  License:   Apache-2.0 AND ISC AND MIT AND Zlib
+@@ -50,7 +50,8 @@ BuildRequires: perl-interpreter
+ # From man pages
+ BuildRequires: python3
+ BuildRequires: python3-sphinx
+-BuildRequires: python3-sphinx-design
++# python3-sphinx-design is intentionally removed to test build succeeds without python3-sphinx-design.
```
rcsanchez97, line 26: Do you mean something like "to test build success" or "to test that the build succeeds"?

rcsanchez97 (approved):
LGTM

.evergreen/etc/spec.patch
kevinAlbs: Updated to "to test that the build succeeds".

PR #1496, opened 2023-12-26
Bson-string: Removed redundant assignement

SaltyMilk:
Hello,
Minor change, I just noticed a redundant assignment in bson-string.c.
Have a nice day.


kevinAlbs (approved):
Thank you for the contribution.

PR #1498, opened 2023-12-28
Replace custom null check with strlen

adriandole:
strlen is faster than a manual loop and will be optimized away for compile-time strings.

Use less-than rather than equality to preserve the behavior where the user can get the first x characters of a string by setting a shorter length.

### DIFF
--- a/src/libbson/src/bson/bson.c
+++ b/src/libbson/src/bson/bson.c
@@ -1559,13 +1559,7 @@ bson_append_regex_w_len (bson_t *bson,
       regex_length = (int) strlen (regex);
    } else {
       /* Necessary to validate embedded NULL is not present in key. */
-<<<<<<< HEAD
       HANDLE_KEY_LENGTH (regex, regex_length);
-=======
-      if (strlen (regex) < regex_length) {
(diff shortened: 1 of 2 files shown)


src/libbson/src/bson/bson.c
```diff
@@ -416,24 +416,13 @@ _bson_append (bson_t *bson,              /* IN */
...
-   return false;
-}
-
 #define HANDLE_KEY_LENGTH(key, key_length)                                \
    do {                                                                   \
       if (key_length < 0) {                                               \
          key_length = (int) strlen (key);                                 \
       } else {                                                            \
          /* Necessary to validate embedded NULL is not present in key. */ \
-         if (_string_contains_null (key, key_length)) {                   \
```
vector-of-bool, line 436: You likely want `strnlen`, as `key` might not be null-terminated at `key_length`.

kevinAlbs:
Closing. https://github.com/mongodb/mongo-c-driver/pull/1738 removes `HANDLE_KEY_LENGTH` and uses `memchr` to check for embedded NULL.

PR #1501, opened 2024-01-02
CDRIVER-4754 pre-4.4 mongos writeConcernError does not determine retryability

jmikola:
https://jira.mongodb.org/browse/CDRIVER-4754

POC for https://github.com/mongodb/specifications/pull/1486

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-error-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-error-private.h
@@ -19,6 +19,8 @@
 #include <bson/bson.h>
 #include <stddef.h>
 
+#include "mongoc-server-description.h"
+
 BSON_BEGIN_DECLS
 
 typedef enum {
@@ -72,7 +74,7 @@ void
 _mongoc_write_error_handle_labels (bool cmd_ret,
                                    const bson_error_t *cmd_err,
                                    bson_t *reply,
-                                   int32_t server_max_wire_version);
+                                   mongoc_server_description_t *sd);
 
 bool
 _mongoc_error_is_shutdown (bson_error_t *error);
--- a/src/libmongoc/src/mongoc/mongoc-error.c
+++ b/src/libmongoc/src/mongoc/mongoc-error.c
@@ -20,6 +20,7 @@
 #include "mongoc-error-private.h"
 #include "mongoc-rpc-private.h"
 #include "mongoc-client-private.h"
+#include "mongoc-server-description-private.h"
 
 bool
 mongoc_error_has_label (const bson_t *reply, const char *label)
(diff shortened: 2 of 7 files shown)


src/libmongoc/src/mongoc/mongoc-error.c
```diff
@@ -20,6 +20,7 @@
 #include "mongoc-error-private.h"
 #include "mongoc-rpc-private.h"
 #include "mongoc-client-private.h"
+#include "mongoc-server-description-private.h"
 
 bool
 mongoc_error_has_label (const bson_t *reply, const char *label)
```
kevinAlbs, line 26: Aside: are error labels expected in `writeConcernError`?

DRIVERS-1641 suggests `writeConcernError` may have error labels:
> Should drivers consider the code/label within the writeConcernError field from a mongos reponse?
> - YES for labels from a 4.4+ mongos. The spec already says that mongos will avoid adding the label if it already retried to help avoid "excessive retries" by the driver.

DRIVERS-1385 suggests `writeConcernError` may not have error labels:
> The server always returns error labels as top-level fields

`mongoc_error_has_label` checks `writeConcernError` for labels. That may not be necessary. Though I do not think it needs to be addressed in this PR.

jmikola: > Aside: are error labels expected in writeConcernError?

No. [DRIVERS-1385](https://jira.mongodb.org/browse/DRIVERS-1385) is correct, since that was confirmed in [this comment](https://jira.mongodb.org/browse/DRIVERS-1385?focusedCommentId=3406472&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-3406472). I'll revise the description of [DRIVERS-1641](https://jira.mongodb.org/browse/DRIVERS-1641) accordingly.

I had no intention of adding label-checking behavior for the `writeConcernError` document, but also missed that libmongoc was already doing so. Thanks for catching that.

I created [CDRIVER-4808](https://jira.mongodb.org/browse/CDRIVER-4808) to track the `mongoc_error_has_label` cleanup.

kevinAlbs (commented):
C changes LGTM. I expect test failures will be resolved with spec test changes of DRIVERS-2802.

kkloberdanz (approved):
Looks good to me!

src/libmongoc/src/mongoc/mongoc-error-private.h
```diff
@@ -72,7 +74,7 @@ void
 _mongoc_write_error_handle_labels (bool cmd_ret,
                                    const bson_error_t *cmd_err,
                                    bson_t *reply,
-                                   int32_t server_max_wire_version);
+                                   mongoc_server_description_t *sd);
```
jmikola, line 77: @kevinAlbs @kkloberdanz: I just thought of this after merging (of course), but would you have preferred I use `const mongoc_server_description_t *sd` here or does it make no difference?

kkloberdanz: I would have a slight preference to making it `const`, but not a hard requirement in my book.

PR #1505, opened 2024-01-04
CDRIVER-4808 do not check writeConcernError in mongoc_error_has_label()

jmikola:
https://jira.mongodb.org/browse/CDRIVER-4808

Server responses do not return errorLabels within writeConcernError. It is only a top-level field.


kevinAlbs (approved):
Thank you. LGTM

PR #1507, opened 2024-01-05
CDRIVER-4717 sync legacy transaction tests for LB runOnRequirements

jmikola:
https://jira.mongodb.org/browse/CDRIVER-4717

Synced with mongodb/specifications@38e65fc8e59fedea351d9393ae43d305cde6ce00

This includes error-labels-blockConnection.json for client-side operations timeout, which is not yet implemented in libmongoc.

Additional tests have been skipped since libmongoc does not pin connections on load-balanced topologies.


kevinAlbs (commented):
To include these tests in CI, add: `test_args+=("-l" "/transactions/legacy/*")` to `run-tests.sh` [here](https://github.com/mongodb/mongo-c-driver/blob/6f4d492aacc31bcfd81e0244456b58de2ef9c84c/.evergreen/scripts/run-tests.sh#L220-L233).

`run-tests.sh` runs the subset of tests noted in https://github.com/mongodb/specifications/tree/eef59d9d386bb054e0eb59e6bd540a030186f899/source/load-balancers/tests#tests

jmikola:
@kevinAlbs: Thanks. I'd have had no idea that was required.

Why do some other directories just run `<spec_name>/*`, like `retryable-writes/*`, despite having "legacy" and "unified" sub-directories? It leads me to wonder why we should add `test_args+=("-l" "/transactions/legacy/*")` instead of changing `test_args+=("-l" "/transactions/unified/*")` to `test_args+=("-l" "/transactions/*")`.

jmikola:
> If all transactions tests are expected to be run against load balancers, I suggest updating the Load Balancer Support test README to add a bullet for "All transactions tests".

Legacy tests for transactions are not the only inconsistency. I just reported [DRIVERS-2806](https://jira.mongodb.org/browse/DRIVERS-2806) so we have a record of this but I don't intend to take any action on that as the priority should be on porting all of these legacy tests to the unified format.

jmikola:
@kevinAlbs: I'm giving up on the rabbit hole chase to get libmongoc's legacy test runner working with LBs. Additionally, the skips I had previously added would have resulted in skipping the tests on _all_ environments (in contast to what we previously did in PHPLIB to only skip tests on LBs).

None of this should be worth the effort given that the transactions tests should be ported to the unified format within the coming quarter ([DRIVERS-1709](https://jira.mongodb.org/browse/DRIVERS-1709)).

kevinAlbs (approved):

PR #1514, opened 2024-01-16
CDRIVER-4512 remove legacy shell from AWS test scripts

kevinAlbs:
# Summary

This PR migrates AWS tests to use scripts from drivers-evergreen-tools added in DRIVERS-2328.

This removes use of the legacy `mongo` shell and resulted in passing the `.*aws.*` tasks. Verified by this patch build: https://spruce.mongodb.com/version/65a6e2ad0ae60677917cf830

I suggest reviewing individual commits.

## Additional changes

https://github.com/mongodb/mongo-c-driver/pull/1514/commits/072eace3a1d00a69d3c61158fbd38bd04125e114 migrates AWS tests using MongoDB 6.0+ to Ubuntu 22.04. This is intended to better follow [Guidelines around Evergreen distros](https://wiki.corp.mongodb.com/display/DBDEVPROD/Guidelines+around+Evergreen+distros):

> Prefer 22.04 when possible

https://github.com/mongodb/mongo-c-driver/pull/1514/commits/6840e85fae72a884cfbbd720ca8d0834765b172b enables tracing in `test-awsauth` intended to help diagnose future test failures. A spurious [test failure](https://spruce.mongodb.com/task/mongo_c_driver_aws_ubuntu2004_test_aws_openssl_lambda_6.0_patch_0da4704d32f80a9eac7d264f1fda2bd70ebe399d_65a167d0850e61bbdb91b2d6_24_01_12_16_24_51/logs?execution=0) was observed on the LAMBDA test case. I was unable to determine the cause.

ec6a689071f1c68717723ec3d0f9db58087120d9 adds a test case for `LAMBDA` to test without a session token. [The AWS test README](https://github.com/mongodb/specifications/blob/2176bbac035263d9397b585c5e70ac1a97359757/source/auth/tests/mongodb-aws.rst) describes two test cases in AWS Lambda:

> Sample URIs both with and without optional session tokens set are shown below. Drivers MUST test both cases.

# Additional background

The `aws_test_secrets_role` variable is now defined on https://spruce.mongodb.com/project/mongo-c-driver/settings/variables. [Using AWS Secrets Manager to Store Testing Secrets](https://wiki.corp.mongodb.com/display/DRIVERS/Using+AWS+Secrets+Manager+to+Store+Testing+Secrets) provides instructions for obtaining secrets.

[Configuration Scripts for End-to-end Testing](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/8e2f560511bbd0a532a84eef6528a12f01890143/.evergreen/auth_aws/README.md) describes the flow for running drivers AWS tests in Evergreen.

### DIFF
--- a/.evergreen/config_generator/components/funcs/fetch_det.py
+++ b/.evergreen/config_generator/components/funcs/fetch_det.py
@@ -21,11 +21,7 @@ class FetchDET(Function):
         bash_exec(
             command_type=EvgCommandType.SETUP,
             working_dir="drivers-evergreen-tools",
-            script='''\
-                for file in $(find .evergreen -type f -name "*.sh"); do
-                    chmod +rx "$file" || exit
-                done
-            ''',
+            script='find .evergreen -type f -name "*.sh" -execdir chmod +rx "{}" \;',
         ),
 
         # python is used frequently enough by many tasks that it is worth
--- a/.evergreen/config_generator/components/funcs/fetch_det.py
+++ b/.evergreen/config_generator/components/funcs/fetch_det.py
@@ -17,6 +17,17 @@ class FetchDET(Function):
             ''',
         ),
 
+        # Make shell scripts executable.
+        bash_exec(
+            command_type=EvgCommandType.SETUP,
+            working_dir="drivers-evergreen-tools",
+            script='''\
+                for file in $(find .evergreen -type f -name \*.sh); do
+                    chmod +x "$file" || exit
+                done
+            ''',
+        ),
+
         # python is used frequently enough by many tasks that it is worth
         # running find_python3 once here and reusing the result.
         bash_exec(
--- a/.evergreen/config_generator/components/funcs/fetch_det.py
+++ b/.evergreen/config_generator/components/funcs/fetch_det.py
@@ -22,8 +22,8 @@ class FetchDET(Function):
             command_type=EvgCommandType.SETUP,
             working_dir="drivers-evergreen-tools",
             script='''\
-                for file in $(find .evergreen -type f -name \*.sh); do
-                    chmod +x "$file" || exit
+                for file in $(find .evergreen -type f -name "*.sh"); do
+                    chmod +rx "$file" || exit
                 done
             ''',
         ),
--- a/.evergreen/generated_configs/functions.yml
+++ b/.evergreen/generated_configs/functions.yml
@@ -177,6 +177,17 @@ functions:
       type: setup
       params:
         binary: bash
+        working_dir: drivers-evergreen-tools
+        args:
+          - -c
+          - |
+            for file in $(find .evergreen -type f -name \*.sh); do
+                chmod +x "$file" || exit
+            done
+    - command: subprocess.exec
+      type: setup
+      params:
+        binary: bash
         args:
           - -c
           - |
--- a/.evergreen/generated_configs/functions.yml
+++ b/.evergreen/generated_configs/functions.yml
@@ -181,8 +181,8 @@ functions:
         args:
           - -c
           - |
-            for file in $(find .evergreen -type f -name \*.sh); do
-                chmod +x "$file" || exit
+            for file in $(find .evergreen -type f -name "*.sh"); do
+                chmod +rx "$file" || exit
             done
     - command: subprocess.exec
       type: setup
--- a/.evergreen/generated_configs/functions.yml
+++ b/.evergreen/generated_configs/functions.yml
@@ -180,7 +180,7 @@ functions:
         working_dir: drivers-evergreen-tools
         args:
           - -c
-          - find .evergreen -type f -name "*.sh" -execdir chmod +x "{}" \;
+          - find .evergreen -type f -name "*.sh" -execdir chmod +rx "{}" \;
     - command: subprocess.exec
       type: setup
       params:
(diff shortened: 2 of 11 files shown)


.evergreen/legacy_config_generator/evergreen_config_lib/variants.py
```diff
@@ -422,25 +422,25 @@ def days(n: int) -> int:
         ],
         {"CC": "clang"},
     ),
+    # Test with Ubuntu 20.04. MongoDB 7.0 no longer ships binaries for Ubuntu 18.04. AWS setup scripts expect Ubuntu 20.04+. MongoDB 4.4 and 5.0 are not available on Ubuntu 20.04.
```
eramongodb, line 425: Confused by this comment vs. matrix below. Should it be something like the following instead?

```
# AWS setup scripts expect Ubuntu 20.04+. MongoDB 4.4 and 5.0 are not available on 22.04.
Variant("aws-ubuntu2004", ...),

# AWS setup scripts expect Ubuntu 20.04+. MongoDB 7.0 and newer are not available on Ubuntu 22.04.
Variant("aws-ubuntu2204", ...),
```

.evergreen/config_generator/components/funcs/fetch_det.py
```diff
@@ -17,6 +17,17 @@ class FetchDET(Function):
...
         ),
 
+        # Make shell scripts executable.
+        bash_exec(
+            command_type=EvgCommandType.SETUP,
+            working_dir="drivers-evergreen-tools",
+            script='''\
+                for file in $(find .evergreen -type f -name \*.sh); do
+                    chmod +x "$file" || exit
+                done
```
eramongodb, line 27:
```suggestion
                for file in $(find .evergreen -type f -name "*.sh"); do
                    chmod +rx "$file" || exit
                done
```

Minor improvements + read permission is also necessary to execute scripts.

eramongodb (changes requested):

.evergreen/config_generator/components/funcs/fetch_det.py
```diff
@@ -17,6 +17,17 @@ class FetchDET(Function):
...
+            ''',
```
rcsanchez97, line 28:
```suggestion
            script='find .evergreen -type f -name "*.sh" -execdir chmod +x "{}" \;'
```

You can also include `+r` as Ezra suggested, but I'm not convinced that's necessary.

.evergreen/generated_configs/functions.yml
```diff
@@ -173,6 +173,17 @@ functions:
...
+      type: setup
+      params:
+        binary: bash
+        working_dir: drivers-evergreen-tools
+        args:
+          - -c
+          - |
+            for file in $(find .evergreen -type f -name \*.sh); do
+                chmod +x "$file" || exit
+            done
```
rcsanchez97, line 186:
```suggestion
            script='find .evergreen -type f -name "*.sh" -execdir chmod +x "{}" \;'
```

rcsanchez97 (approved):
Overall, LGTM. I just have two very minor suggestions around the `chmod` invocations.

.evergreen/config_generator/components/funcs/fetch_det.py
eramongodb: > I'm not convinced it's necessary.

Absence of read permission bit me a few times during CDRIVER-3620 work. Apparently execute-only permission is (normally) only supported for binaries due to how they are executed; it does not work with interpreted shell scripts (see: https://unix.stackexchange.com/questions/34202/can-a-script-be-executable-but-not-readable).

rcsanchez97: I'm confused about how it would even be possible for the current user to fetch DET (presumably by cloning) and end up with files not readable by the current user. I get the problem with execute permissions (which could either be absent in the repo, or get lost based on clone options, target FS, OS, and who knows what else). I've just never encountered a scenario where user read permissions were absent (in a normal file, that is).

kevinAlbs: Updated to make more concise. I included the read permission in case it is needed in the future.

kevinAlbs (commented):
Latest changes verified with this patch: https://evergreen.mongodb.com/version/65a6fd987742ae6308101aa3

eramongodb (approved):

.evergreen/generated_configs/functions.yml
```diff
@@ -180,10 +180,7 @@ functions:
         working_dir: drivers-evergreen-tools
         args:
           - -c
-          - |
-            for file in $(find .evergreen -type f -name \*.sh); do
-                chmod +x "$file" || exit
-            done
+          - find .evergreen -type f -name "*.sh" -execdir chmod +x "{}" \;
```
rcsanchez97, line 183:
```suggestion
          - find .evergreen -type f -name "*.sh" -execdir chmod +rx "{}" \;
```

Add the `+r` here too, for consistency.

rcsanchez97 (approved):

.evergreen/generated_configs/functions.yml
kevinAlbs: Done. `functions.yml` is generated. I forgot to regenerate.

PR #1516, opened 2024-01-17
CDRIVER-4512 do not use `execdir`

kevinAlbs:
This is a follow-up to https://github.com/mongodb/mongo-c-driver/pull/1514 intended to resolve errors on some tasks running `find` with `-execdir`. [Example](https://spruce.mongodb.com/task/mongo_c_driver_cse_matrix_openssl_cse_sasl_cyrus_openssl_ubuntu2004_arm64_gcc_test_latest_server_auth_7f1a88b42a311e6db8edd4adafcce55aa3b8a094_24_01_17_13_28_12/logs?execution=0):

> [2024/01/17 08:36:31.247] find: The relative path '.local/bin' is included in the PATH environment variable, which is insecure in combination with the -execdir action of find.  Please remove that entry from $PATH

Verified with this patch build: https://spruce.mongodb.com/version/65a80624a4cf47738e04ec81

### DIFF
--- a/.evergreen/config_generator/components/funcs/fetch_det.py
+++ b/.evergreen/config_generator/components/funcs/fetch_det.py
@@ -21,7 +21,7 @@ class FetchDET(Function):
         bash_exec(
             command_type=EvgCommandType.SETUP,
             working_dir="drivers-evergreen-tools",
-            script='find .evergreen -type f -name "*.sh" -execdir chmod +rx "{}" \;',
+            script='find .evergreen -type f -name "*.sh" -exec chmod +rx "{}" \;',
         ),
 
         # python is used frequently enough by many tasks that it is worth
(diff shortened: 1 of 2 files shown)


rcsanchez97 (changes requested):
Could you try with `-exec` rather than `-execdir`? That should not be subject to the same complaint from `find`.

kevinAlbs:
> Could you try with `-exec` rather than `-execdir`? That should not be subject to the same complaint from `find`.

`-exec` also appears to work. Patch: https://spruce.mongodb.com/version/65a82a95562343d7efdc4843

rcsanchez97 (approved):
LGTM

rcsanchez97:
Thanks for checking. I tend to make a habit of preferring `-execdir`, but that means that every once in a while a situation like this will make it the wrong choice. Thanks for checking `-exec` and going with that.

PR #1520, opened 2024-01-22
CDRIVER-4663 Eagerly create minPoolSize connections

kkloberdanz:


src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -373,6 +373,77 @@ mongoc_client_pool_try_pop (mongoc_client_pool_t *pool)
...
 
+static bool
+_mongoc_connect_client_to_servers (mongoc_client_t *client)
+{
+   bool ok = true;
+   bson_error_t error = {0};
+   bson_t *ping_command = BCON_NEW ("ping", BCON_INT32 (1));
+   size_t num_servers = 0;
+   mongoc_server_description_t **descriptions =
+      mongoc_client_get_server_descriptions (client, &num_servers);
```
kevinAlbs, line 384: I suggest sending a `ping` to one server before calling `mongoc_client_get_server_descriptions`.

`mongoc_client_get_server_descriptions` does not connect to a server. If server monitoring has not yet returned, I expect it will return the initial set of server descriptions from the URI.

Example: if the URI is `mongodb://localhost:27017,localhost:27018`, but there are three nodes in the replica set, `mongoc_client_get_server_descriptions` may only return the two servers in the URI (if monitoring has not yet occurred). Once a `hello` response is processed from the two known servers, the third server will be added to the topology description.

This is applied in the suggested C++ implementation.

kevinAlbs (commented):
I am concerned about the generic usefulness of the warmup API. I see this as being primarily motivated by Genny's need to benchmark the server. I expect that is not typical use of the C driver.

Rust has precedent for a client warmup method (https://github.com/mongodb/mongo-rust-driver/pull/932). Though it may not be an apples-to-apples comparison. The Rust driver implements CMAP, which limits the rate of connection creation with `maxConnecting`. The C driver does not limit the rate of connection creation.

I recommend implementing a `pool::warmup` in the C++ driver only to limit the new API. This may be a convenience for general users, but I prefer limiting API additions until there is more of a known need. Here is a suggested implementation:
https://github.com/mongodb/mongo-cxx-driver/compare/master...kevinAlbs:mongo-cxx-driver:CDRIVER-4663?expand=1

kevinAlbs:
No longer needed. See CDRIVER-4663.

PR #1525, opened 2024-01-26
Refactor Earthly EVG tasks for better consistency with C Driver tasks

eramongodb:
Compare task and variant names in [this patch](https://spruce.mongodb.com/version/65b4134957e85a1e221d727e) vs. [base commit](https://spruce.mongodb.com/version/mongo_c_driver_5e5784d814b2e3912c2855ac2b9a172267b8235f/tasks?taskName=check%3A).

Notable changes:

- Grouped all Earthly tasks under a single "earthly" build variant, selected via a single `.earthly` tag.
- Prefixed all Earthly task names with `earthly-` (alphabetically grouped in tasks.yml).
- Moved platform selection into task name.
- All task names and fields are lowercase and delimited by simple dashes.
- Fixed list of `earthly-{env}` tags per task.
- Moved `run_on` field from tasks to build variant (specific distro doesn't matter + reduce tasks.yml verbosity).

Also includes the following drive-by improvements:

- Use ~Alpine 3~ Alpine 3.19 instead of Alpine 3.18 (see: https://github.com/mongodb/mongo-cxx-driver/pull/1088 and https://github.com/mongodb/mongo-cxx-driver/pull/1090).
- Removal the obsolete `early_termination` config field from root config.


kevinAlbs (approved):

vector-of-bool (commented):
This unfortunately runs counter to the intended design, that is: Earthly is the means to support additional targets, but it isn't the target itself (e.g. having an "Earthly" build variant is akin to having a "CMake" build variant).

I have changes in progress (soon to be an open PR) that will expand the platform set and capabilities of the Earthly execution, and hope to eventually migrate most CI tasks such that they can be executed this same way.

eramongodb:
Deferring/postponing this PR in favor of prioritizing https://github.com/mongodb/mongo-c-driver/pull/1537. May revisit suggestions in this PR later, relative to https://github.com/mongodb/mongo-c-driver/pull/1537 instead.

PR #1526, opened 2024-01-30
CDRIVER-4816 Error parsing large IPv6 literal

kevinAlbs:
# Summary

Error parsing IPv6 literals exceeding `BSON_HOST_NAME_MAX - 2` bytes.

Tested with this patch build: https://spruce.mongodb.com/version/65b91a525623434922184fa6. Test failures appear to be unrelated.

# Background & Motivation

See CDRIVER-4816.

Assertions are added to the return values of `bson_snprintf` in `_mongoc_host_list_from_hostport_with_err` to ensure the destination has enough capacity. CDRIVER-4820 was filed to check other calls to `bson_snprintf` in the codebase.

# Rejected alternatives

Increasing the size of `mongoc_host_list_t::host_and_port` was briefly considered, but decided against due to `mongoc_host_list_t` being included in the [public API](https://mongoc.org/libmongoc/current/mongoc_host_list_t.html). Changing the size is an ABI breaking change.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-host-list.c
+++ b/src/libmongoc/src/mongoc/mongoc-host-list.c
@@ -1,7 +1,7 @@
 /*
  * Copyright 2015 MongoDB Inc.
- *
  * Licensed under the Apache License, Version 2.0 (the "License");
+ *
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+#include <inttypes.h> // PRIu16
 
 #include "mongoc-host-list-private.h"
 /* strcasecmp on windows */
--- a/src/libmongoc/src/mongoc/mongoc-host-list.c
+++ b/src/libmongoc/src/mongoc/mongoc-host-list.c
@@ -358,7 +358,7 @@ _mongoc_host_list_from_hostport_with_err (mongoc_host_list_t *link_,
       mongoc_lowercase (link_->host, link_->host);
       int req = bson_snprintf (link_->host_and_port,
                                sizeof link_->host_and_port,
-                               "[%s]:%hu",
+                               "[%s]:%" PRIu16,
                                link_->host,
                                link_->port);
       BSON_ASSERT (bson_in_range_size_t_signed (req));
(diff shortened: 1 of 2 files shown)


src/libmongoc/src/mongoc/mongoc-host-list.c
```diff
@@ -358,11 +372,14 @@ _mongoc_host_list_from_hostport_with_err (mongoc_host_list_t *link_,
...
 
       mongoc_lowercase (link_->host, link_->host);
-      bson_snprintf (link_->host_and_port,
-                     sizeof link_->host_and_port,
-                     "%s:%hu",
-                     link_->host,
-                     link_->port);
+      int req = bson_snprintf (link_->host_and_port,
+                               sizeof link_->host_and_port,
+                               "%s:%hu",
```
eramongodb, line 377:
```suggestion
                               "%s:%" PRIu16,
```
For consistency with `uint16_t` (from `<inttypes.h>`).

src/libmongoc/src/mongoc/mongoc-host-list.c
```diff
@@ -343,13 +343,27 @@ _mongoc_host_list_from_hostport_with_err (mongoc_host_list_t *link_,
...
 
+      mongoc_lowercase (link_->host, link_->host);
+      int req = bson_snprintf (link_->host_and_port,
+                               sizeof link_->host_and_port,
+                               "[%s]:%hu",
+                               link_->host,
+                               link_->port);
+      BSON_ASSERT (bson_in_range_size_t_signed (req));
+      // Use `<`, not `<=` to account for NULL byte.
+      BSON_ASSERT ((size_t) req < sizeof link_->host_and_port);
```
eramongodb, line 366: Per snprintf spec:

> The snprintf function returns the number of characters that would have been written had n been sufficiently large, not counting the terminating null character, or a negative value if an encoding error occurred. Thus, the null-terminated output has been completely written if and only if the returned value is nonnegative and less than n.

Has the port length already been validated at this point? If not, this assertion should probably be made a call to `bson_set_error()` instead (e.g. if port ends up being 6 digits long, violating `host_and_port` length expectations).

This comment also applies to the branch below.

kevinAlbs: > Has the port length already been validated at this point?

I expect the port length cannot exceed 5 digits. UINT16_MAX is 65535.

src/libmongoc/src/mongoc/mongoc-host-list.c
```diff
@@ -343,13 +343,27 @@ _mongoc_host_list_from_hostport_with_err (mongoc_host_list_t *link_,
...
+            MONGOC_ERROR_STREAM_NAME_RESOLUTION,
+            "IPv6 literal provided in URI is too long, max is %d chars",
+            BSON_HOST_NAME_MAX - 2);
+         return false;
+      }
 
+      mongoc_lowercase (link_->host, link_->host);
+      int req = bson_snprintf (link_->host_and_port,
+                               sizeof link_->host_and_port,
+                               "[%s]:%" PRIu16,
```
eramongodb, line 361: IWYU: missing `#include <inttypes.h>` for `PRIu16`.

eramongodb (approved):

PR #1547, opened 2024-02-29
allow for 64-bit time_t on 32-bit architectures

rcsanchez97:
This was tested by executing a manual build on a Debian armhf porterbox (since the failure was revealed by a failed build on that architecture in Debian and we don't directly test 32-bit ARM architectures in our own CI).

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-log.c
+++ b/src/libmongoc/src/mongoc/mongoc-log.c
@@ -207,7 +207,7 @@ mongoc_log_default_handler (mongoc_log_level_t log_level, const char *log_domain
    fprintf (stream,
             "%s.%04" PRId64 ": [%5d]: %8s: %12s: %s\n",
             nowstr,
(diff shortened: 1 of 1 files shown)


src/libmongoc/src/mongoc/mongoc-log.c
```diff
@@ -205,9 +205,9 @@ mongoc_log_default_handler (mongoc_log_level_t log_level, const char *log_domain
 #endif
 
    fprintf (stream,
-            "%s.%04ld: [%5d]: %8s: %12s: %s\n",
+            "%s.%04" PRId64 ": [%5d]: %8s: %12s: %s\n",
             nowstr,
-            tv.tv_usec / 1000L,
+            (int64_t)(tv.tv_usec / 1000L),
```
eramongodb, line 210:
```suggestion
            (int64_t) (tv.tv_usec / 1000),
```
ClangFormat + removal of redundant literal suffix.

eramongodb (approved):

PR #1552, opened 2024-03-07
CDRIVER-5504 Use size_t when indexing arrays

eramongodb:
Address CDRIVER-5504.

### DIFF
--- a/src/libbson/src/bson/bson-string.c
+++ b/src/libbson/src/bson/bson-string.c
@@ -473,8 +473,8 @@ void
 bson_strfreev (char **str) /* IN */
 {
    if (str) {
-      for (size_t i = 0u; str[i] != NULL; ++i) {
-         bson_free (str[i]);
+      for (char **ptr = str; *ptr != NULL; ++ptr) {
+         bson_free (*ptr);
       }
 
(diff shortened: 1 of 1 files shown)


src/libbson/src/bson/bson-string.c
```diff
@@ -472,11 +472,11 @@ bson_strndup (const char *str, /* IN */
 void
 bson_strfreev (char **str) /* IN */
 {
-   int i;
-
    if (str) {
-      for (i = 0; str[i]; i++)
+      for (size_t i = 0u; str[i] != NULL; ++i) {
          bson_free (str[i]);
+      }
```
kevinAlbs, line 478: Suggest iterating with a `char **`:
```suggestion
      for (char **ptr = str; *ptr != NULL; ++ptr) {
         bson_free (*ptr);
      }
```

This may avoid a possible unsigned overflow with `i` if `str` has more than `SIZE_MAX` entries.

eramongodb: +1 to using pointer-based iteration over indexing (with all the complexities that come with it). Updated accordingly.

kevinAlbs (approved):

PR #1556, opened 2024-03-12
CDRIVER-5548 new mongoc_client_update_sockettimeoutms()

scotthunt:
so multiple pools are not necessary to accomodate different timeouts

### DIFF
--- a/src/libmongoc/doc/mongoc_client_set_sockettimeoutms.rst
+++ b/src/libmongoc/doc/mongoc_client_set_sockettimeoutms.rst
@@ -1,0 +1,26 @@
+:man_page: mongoc_client_set_sockettimeoutms
+
+mongoc_client_set_sockettimeoutms()
+===================================
+
+Synopsis
+--------
+
+.. code-block:: c
+
+  void
+  mongoc_client_set_sockettimeoutms (mongoc_client_t *client, int32_t timeoutms);
+
+Change the ``sockettimeoutms`` of the :symbol:`mongoc_client_t`.
+
+If ``client`` was obtained from a :symbol:`mongoc_client_pool_t`, the socket timeout is restored to the previous value when calling :symbol:`mongoc_client_pool_push`.
+
+Parameters
+----------
+
+* ``client``: A :symbol:`mongoc_client_t`.
+* ``timeoutms``: The requested timeout value in milliseconds.
+
+.. seealso::
+
+  | :ref:`URI Connection Options <connection_options>`
--- a/src/libmongoc/doc/mongoc_client_t.rst
+++ b/src/libmongoc/doc/mongoc_client_t.rst
@@ -87,6 +87,7 @@ Example
     mongoc_client_set_read_concern
     mongoc_client_set_read_prefs
     mongoc_client_set_server_api
+    mongoc_client_set_sockettimeoutms
     mongoc_client_set_ssl_opts
     mongoc_client_set_stream_initiator
     mongoc_client_set_write_concern
--- a/src/libmongoc/doc/mongoc_uri_t.rst
+++ b/src/libmongoc/doc/mongoc_uri_t.rst
@@ -84,6 +84,8 @@ IPv4 and IPv6
 
 .. include:: includes/ipv4-and-ipv6.txt
 
+.. _connection_options:
+
 Connection Options
 ------------------
 
--- a/src/libmongoc/src/mongoc/mongoc-client-pool.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-pool.c
@@ -366,6 +366,9 @@ mongoc_client_pool_push (mongoc_client_pool_t *pool, mongoc_client_t *client)
    BSON_ASSERT_PARAM (pool);
    BSON_ASSERT_PARAM (client);
 
+   /* reset sockettimeoutms to the default in case it was changed with mongoc_client_set_sockettimeoutms() */
+   mongoc_cluster_reset_sockettimeoutms (&client->cluster);
+
    bson_mutex_lock (&pool->mutex);
    _mongoc_queue_push_head (&pool->queue, client);
 
--- a/src/libmongoc/src/mongoc/mongoc-client.c
+++ b/src/libmongoc/src/mongoc/mongoc-client.c
@@ -1125,6 +1125,13 @@ mongoc_client_destroy (mongoc_client_t *client)
 }
 
 
+void
+mongoc_client_set_sockettimeoutms (mongoc_client_t *client, int32_t timeoutms)
+{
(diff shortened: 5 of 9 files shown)


kevinAlbs:
Hi @scotthunt  thank you for the PR. Can you elaborate on the motivation? E.g. what operations require a different `socketTimeoutMS` from what was set on the pool?

`socketTimeoutMS` is specified as a [URI option](https://github.com/mongodb/specifications/blob/510a1e7ce2b232cab310f1fc932988ffe878c01a/source/client-side-operations-timeout/client-side-operations-timeout.md#deprecations) for all drivers.

There is future planned work that may conflict with this change. Client-Side Operation Timeout proposes a unified timeout (CDRIVER-3786) that [deprecates socketTimeoutMS](https://github.com/mongodb/specifications/blob/510a1e7ce2b232cab310f1fc932988ffe878c01a/source/client-side-operations-timeout/client-side-operations-timeout.md#deprecations).

There may be preferable alternatives: e.g. specifying `maxTimeMS` on an operation to have the server time out.

scotthunt:
Our application encompasses various database-related activities, some of which are stores or simple retrieves while others are scans across millions of rows with regex matches applied.
These varied uses do not fit well with a single timeout value constraint.
We are currently dealing with this by creating 3 `mongoc_pool_t` (with different timeout values: 5s, 180s, 600s) for each of our mongo clusters (currently: 8).  However, that leads to some code ugliness as well as (I believe) inefficiencies in connection management.  (I/we believe that 8 connection pools would be "better" than 24?)

I'm certainly open to alternatives.  In my quick perusal, however, it does not appear that `maxTimeMS` applies to everything that `sockerTimeoutMS` does?
That said, a revamped client timeout setup would definitely welcome... though that ticket has been open for literally years?

kevinAlbs:
> However, that leads to some code ugliness as well as (I believe) inefficiencies in connection management. (I/we believe that 8 connection pools would be "better" than 24?)

I expect less pools would result in less total connections. Each pool shares connections for monitoring. And two pools cannot share connections for operations. If connection count is a concern, the [`maxPoolSize` option](https://mongoc.org/libmongoc/current/mongoc_uri_t.html#connection-pool-options) can also be set to limit the number of clients in a pool.

> I'm certainly open to alternatives. In my quick perusal, however, it does not appear that `maxTimeMS` applies to everything that `sockerTimeoutMS` does?

Many server operations support `maxTimeMS` (refer: [command reference](https://www.mongodb.com/docs/manual/reference/command)). An advantage of specifying `maxTimeMS` is to avoid connection churn.

If the server returns an error due to the operation exceeding `maxTimeMS`, the socket can continue to be used. If a socket times out due to `socketTimeoutMS` it must be closed and re-established.

Here is [an example](https://github.com/kevinAlbs/c-bootstrap/blob/7de693f2908f210c629c97229f7ad5eb522b2728/investigations/timeouts/main.c#L14-L53) of setting different `maxTimeMS` on `insert` and `find` operations.

kevinAlbs:
@scotthunt does using `maxTimeMS` solve for your use case?  I prefer not to accept this PR since `socketTimeoutMS` is planned for future deprecation.

scotthunt:
Here are my thoughts, for whatever they're worth:
1. Not all server operations appear to support maxTimeMS (including perhaps all write operations?)
2. I do not view a server-side timeout as a replacement for a client-side timeout.  Ideally, I would use both, with the server-side value set somewhat smaller than the client-side one?

So: What is the plan for client-side timeouts going forward?  And am I going to be able to change them on a connection object after popping it from a connection pool?  And how soon might those changes actually come about?
Thanks.

kevinAlbs (changes requested):
Thank you for the additional feedback. Please see suggested additional commits on [this branch](https://github.com/mongodb/mongo-c-driver/compare/master...kevinAlbs:mongo-c-driver:rebased.sockettimeoutms?expand=1). 

---

> Not all server operations appear to support maxTimeMS (including perhaps all write operations?)

I expect most do. [This example](https://github.com/kevinAlbs/c-bootstrap/blob/7de693f2908f210c629c97229f7ad5eb522b2728/investigations/timeouts/main.c#L14-L26) shows setting `maxTimeMS` on an insert operation. 

> I do not view a server-side timeout as a replacement for a client-side timeout. Ideally, I would use both, with the server-side value set somewhat smaller than the client-side one?

I agree. Plus, some `maxTimeMS` behavior is subtle. For example: `maxTimeMS` set on `find` applies to the cumulative lifetime of the returned cursor (i.e. including all `getMore` commands).

Client-Side Operation Timeout (CSOT) specifies similar behavior: a socket timeout is used in combination with a smaller `maxTimeMS` (taking the estimated RTT into account).

> And how soon might those changes actually come about?

I expect CSOT is a large project and do not expect it to be completed in the near future. This PR seems like a reasonable interim solution.

scotthunt:
1. your additional commits look great.  removing the extra public function seems fine, can't imagine it would be used, I had only added it for completeness.  would you like me to squash them into this PR?  or something else?

2. on the `maxTimeMS` problems (running tests against MongoDB v6.0.4-enterprise) in case you care:
    a. I get query errors when I try to use it on any bulk operation (`mongoc_bulk_operation_t`).  I tried it both (separately) in the opts passed to `mongoc_collection_create_bulk_operation_with_opts()` and in `mongoc_bulk_operation_insert_with_opts()`, `mongoc_bulk_operation_update_one_with_opts()`, `mongoc_bulk_operation_update_one_with_opts()`, and `mongoc_bulk_operation_remove_one_with_opts()`
    b. I get query errors when I try to use it on: `mongoc_collection_update_one()`
    c. I get what appears to be a non-functioning query (no error, but the update doesn't seem to work?) when I try it with `mongoc_collection_update_many()`
    d. It does appear to work (or at least not complain or break?) on: `mongoc_collection_find_with_opts()`, `mongoc_collection_count_documents()`, and `mongoc_database_get_collection_names_with_opts()`

kevinAlbs:
> would you like me to squash them into this PR? or something else?

Yes please. Or I can try committing directly to this branch if you prefer.

scotthunt:
rebased on master and squashed all your commits in (and adjusted the commit message with a Co-Author message)
feel free to ask for any other changes, I'm happy to oblige

kevinAlbs (approved):

rcsanchez97 (approved):
LGTM

kevinAlbs:
Thank you for the contribution @scotthunt. This is planned to be included in the 1.27.0 release later this week.

bisht2050:
@scotthunt 
[**1.27.0**](https://github.com/mongodb/mongo-c-driver/releases) with this change is now available.

PR #1558, opened 2024-03-12
CDRIVER-4828 skip `rangePreview` tests on server 8.0+

kevinAlbs:
This PR does not resolve CDRIVER-4828, but adds skips for `rangePreview` tests on MongoDB Server 8.0+ to avoid expected test failures on latest server builds.

See https://github.com/mongodb/specifications/pull/1548 for motivation.

Verified with this patch build: https://spruce.mongodb.com/version/65f06c270ae60677761ed38f. Though server changes expected to fail are not-yet-available.


kkloberdanz (approved):
LGTM!

PR #1563, opened 2024-03-25
CDRIVER-4817 update compile-libmongocrypt for KMIP delegated

adriandole:


src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -1073,6 +1073,15 @@ _state_machine_run (_state_machine_t *state_machine, bson_t *result, bson_error_
          break;
       case MONGOCRYPT_CTX_DONE:
          goto success;
+         break;
+      case MONGOCRYPT_CTX_NEED_MONGO_COLLINFO_WITH_DB:
+         bson_set_error (error,
+                         MONGOC_ERROR_CLIENT_SIDE_ENCRYPTION,
+                         0,
```
kevinAlbs, line 1080:
```suggestion
                         MONGOC_ERROR_CLIENT_INVALID_ENCRYPTION_STATE,
```

Suggest using a non-zero error code. Callers may assume if the code is 0, no error occurred.

kevinAlbs (approved):
LGTM with an update to the error code.

src/libmongoc/src/mongoc/mongoc-crypt.c
```diff
@@ -1073,6 +1073,15 @@ _state_machine_run (_state_machine_t *state_machine, bson_t *result, bson_error_
          break;
       case MONGOCRYPT_CTX_DONE:
          goto success;
+         break;
+      case MONGOCRYPT_CTX_NEED_MONGO_COLLINFO_WITH_DB:
```
alcaeus, line 1077: @adriandole this introduced breakage in the PHP driver, where we test with the latest version of libmongoc. The issue is that the current stable libmongocrypt version doesn't have `MONGOCRYPT_CTX_NEED_MONGO_COLLINFO_WITH_DB` defined, causing a compilation error. This should be wrapped in an `#ifdef` block unless there is proper documentation that libmongocrypt 1.10 (which hasn't been released yet) or newer is required to compile libmongoc 1.27.

kevinAlbs: libmongoc 1.27 is expected to release after libmongocrypt 1.10 (CDRIVER-5299 depends on MONGOCRYPT-665). https://github.com/mongodb/mongo-c-driver/pull/1573/files bumps the required libmongocrypt to 1.10.

If updating to a libmongoc 1.27 prerelease, consider also updating libmongocrypt to a 1.10 prerelease containing https://github.com/mongodb/libmongocrypt/commit/1793ff43d9c1849cfef76e8bfbafa86d41570eef (introduces `MONGOCRYPT_CTX_NEED_MONGO_COLLINFO_WITH_DB`).

PR #1564, opened 2024-03-27
CDRIVER-5506 fix `mongoc_gridfs_file_readv` with `min_bytes` <= 0

kevinAlbs:


kevinAlbs:
Closing. Upon further reading, I am not sure `min_bytes` of 0 is supported. 

[mongoc_stream_read()](https://mongoc.org/libmongoc/current/mongoc_stream_read.html) (which can wrap `mongoc_gridfs_file_readv`) documents:

> It returns `>= 0` and `< min_bytes` when end-of-file is encountered

That may suggest that passing `min_bytes=0` is not supported. If EOF is encountered, how could the return be `>= 0` and `< min_bytes`.

PR #1565, opened 2024-03-27
CDRIVER-5506 error if `mongoc_gridfs_file_readv` reads an incomplete chunk

kevinAlbs:
# Summary

Address a hang observed in `mongoc_gridfs_file_readv` when reading a corrupt incomplete chunk. If an incomplete chunk is detected, an error is now returned.

# Background & Motivation

Though the `mongoc_gridfs_t` API notably [does not conform](https://mongoc.org/libmongoc/current/mongoc_gridfs_t.html) to the GridFS spec, the [GridFS spec](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.md) requires returning an error for an unexpected length:

> As drivers stream the stored file they MUST check that each chunk received is the next expected chunk (i.e. it has the expected "n" value) and that the data field is of the expected length

The cause of the hang is an assumption that chunks (excluding the last) contain data of length chunk size. `_mongoc_gridfs_file_refresh_page` calculates the chunk to read [from the position of the last read byte](https://github.com/mongodb/mongo-c-driver/blob/a2dedb3ff2875a5ac75e3e0a8ded72bf838be1ec/src/libmongoc/src/mongoc/mongoc-gridfs-file.c#L753):

```c
file->n = (int32_t) (file->pos / file->chunk_size);
```

If an incomplete chunk is read, `file->pos` may be smaller than assumed, resulting in repeated setting of `file->n` to an already-read chunk.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,6 +5,10 @@ Cyrus SASL:
 
   * Disable plugin loading with Cyrus SASL on Windows by default. To re-enable, set the CMake option `CYRUS_PLUGIN_PATH_PREFIX` to the absolute path prefix of the Cyrus SASL plugins.
 
+Fixes:
+
+  * Fix possible hang if `mongoc_gridfs_file_readv` is called with a corrupt chunk with incomplete data.
+
 libmongoc 1.26.1
 ================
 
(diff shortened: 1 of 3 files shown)


kkloberdanz (approved):
Looks good!

PR #1576, opened 2024-04-17
CDRIVER-4363 refactor retryable writes into shared function

kevinAlbs:
This PR relocates the retryable writes logic into a new function: `mongoc_cluster_run_retryable_write`.

This enables sharing the retryable writes logic among the current three uses. It is planned to be used in the new bulk write helper for CDRIVER-4363.

No functional changes are expected.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-client.c
+++ b/src/libmongoc/src/mongoc/mongoc-client.c
@@ -1562,8 +1562,6 @@ _mongoc_client_retryable_write_command_with_stream (mongoc_client_t *client,
                                                     bson_error_t *error)
 {
    mongoc_server_stream_t *retry_server_stream = NULL;
-   bson_iter_t txn_number_iter;
-   bool is_retryable = true;
    bool ret;
 
    ENTRY;
@@ -1571,86 +1569,11 @@ _mongoc_client_retryable_write_command_with_stream (mongoc_client_t *client,
    BSON_ASSERT_PARAM (client);
    BSON_ASSERT (parts->is_retryable_write);
 
-   /* increment the transaction number for the first attempt of each retryable
-    * write command */
-   BSON_ASSERT (bson_iter_init_find (&txn_number_iter, parts->assembled.command, "txnNumber"));
-   bson_iter_overwrite_int64 (&txn_number_iter, ++parts->assembled.session->server_session->txn_number);
-
-   // Store the original error and reply if needed.
-   struct {
-      bson_t reply;
-      bson_error_t error;
-      bool set;
-   } original_error = {.reply = {0}, .error = {0}, false};
-
-retry:
-   ret = mongoc_cluster_run_command_monitored (&client->cluster, &parts->assembled, reply, error);
-
-   _mongoc_write_error_handle_labels (ret, error, reply, server_stream->sd);
-
-   if (is_retryable) {
-      _mongoc_write_error_update_if_unsupported_storage_engine (ret, error, reply);
-   }
-
-   /* If a retryable error is encountered and the write is retryable, select
-    * a new writable stream and retry. If server selection fails or the selected
-    * server does not support retryable writes, fall through and allow the
-    * original error to be reported. */
-   if (is_retryable && _mongoc_write_error_get_type (reply) == MONGOC_WRITE_ERR_RETRY) {
-      bson_error_t ignored_error;
-
-      // The write command may be retried at most once.
-      is_retryable = false;
-
-      {
-         mongoc_deprioritized_servers_t *const ds = mongoc_deprioritized_servers_new ();
-
-         mongoc_deprioritized_servers_add_if_sharded (ds, server_stream->topology_type, server_stream->sd);
-
-         BSON_ASSERT (!retry_server_stream);
-         retry_server_stream =
-            mongoc_cluster_stream_for_writes (&client->cluster, parts->assembled.session, ds, NULL, &ignored_error);
-
-         mongoc_deprioritized_servers_destroy (ds);
-      }
-
-      if (retry_server_stream) {
-         parts->assembled.server_stream = retry_server_stream;
-         {
-            // Store the original error and reply before retry.
-            BSON_ASSERT (!original_error.set); // Retry only happens once.
-            original_error.set = true;
-            bson_copy_to (reply, &original_error.reply);
-            if (error) {
-               original_error.error = *error;
-            }
-         }
-         bson_destroy (reply);
-         GOTO (retry);
-      }
-   }
+   ret = mongoc_cluster_run_retryable_write (
+      &client->cluster, &parts->assembled, true /* is_retryable */, &retry_server_stream, reply, error);
 
    if (retry_server_stream) {
       mongoc_server_stream_cleanup (retry_server_stream);
-   }
-
-   // If a retry attempt fails with an error labeled NoWritesPerformed,
-   // drivers MUST return the original error.
-   if (original_error.set && mongoc_error_has_label (reply, "NoWritesPerformed")) {
-      if (error) {
-         *error = original_error.error;
-      }
-      bson_destroy (reply);
-      bson_copy_to (&original_error.reply, reply);
-   }
-
-   if (original_error.set) {
-      bson_destroy (&original_error.reply);
-   }
-
-   if (ret && error) {
-      /* if a retry succeeded, clear the initial error */
-      memset (error, 0, sizeof (bson_error_t));
    }
 
    RETURN (ret);
--- a/src/libmongoc/src/mongoc/mongoc-cluster-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-cluster-private.h
@@ -186,6 +186,16 @@ mongoc_cluster_stream_valid (mongoc_cluster_t *cluster, mongoc_server_stream_t *
 bool
 mongoc_cluster_run_command_monitored (mongoc_cluster_t *cluster, mongoc_cmd_t *cmd, bson_t *reply, bson_error_t *error);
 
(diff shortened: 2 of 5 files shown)


src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -3636,3 +3636,98 @@ mcd_rpc_message_decompress_if_necessary (mcd_rpc_message *rpc, void **data, size
...
+                                    mongoc_cmd_t *cmd,
+                                    bool is_retryable_write,
+                                    mongoc_server_stream_t **retry_server_stream,
+                                    bson_t *reply,
+                                    bson_error_t *error)
+{
+   BSON_ASSERT_PARAM (cluster);
+   BSON_ASSERT_PARAM (cmd);
+   BSON_ASSERT_PARAM (retry_server_stream);
+   BSON_ASSERT_PARAM (reply);
```
eramongodb, line 3651:
```suggestion
   BSON_ASSERT_PARAM (reply);
   BSON_ASSERT (error || true);
```
For clarity of intent.

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -3636,3 +3636,98 @@ mcd_rpc_message_decompress_if_necessary (mcd_rpc_message *rpc, void **data, size
...
+      BSON_ASSERT (bson_iter_init_find (&txn_number_iter, cmd->command, "txnNumber"));
+      bson_iter_overwrite_int64 (&txn_number_iter, ++cmd->session->server_session->txn_number);
+   }
+
+   // Store the original error and reply if needed.
+   struct {
+      bson_t reply;
+      bson_error_t error;
+      bool set;
+   } original_error = {.reply = {0}, .error = {0}, .set = false};
```
eramongodb, line 3670:
```suggestion
   } original_error = {.reply = {0}, .error = {0}, .set = false};

   // Ensure `*retry_server_stream` is always valid or null.
   *retry_server_stream = NULL;
```
To avoid risks such as the following:
```c
// Uninitialized.
mongoc_server_stream_t *retry_server_stream;

// `retry_server_stream` is *maybe* assigned-to.
const bool ret = mongoc_cluster_run_retryable_write (..., &retry_server_stream, ...);

// *Maybe* undefined behavior (still uninitialized).
if (retry_server_stream) {
  // ...
}
```

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -3636,3 +3636,98 @@ mcd_rpc_message_decompress_if_necessary (mcd_rpc_message *rpc, void **data, size
...
+   BSON_ASSERT_PARAM (cmd);
+   BSON_ASSERT_PARAM (retry_server_stream);
+   BSON_ASSERT_PARAM (reply);
+
+   bool ret;
+   // `can_retry` is set to false on retry. A retry may only happen once.
+   bool can_retry = is_retryable_write;
+
+   /* increment the transaction number for the first attempt of each retryable
+    * write command */
```
eramongodb, line 3658:
```suggestion
   // Increment the transaction number for the first attempt of each retryable
   // write command.
```
Mix of `/**/` and `//` style comments is distracting. Recommend using `//` consistently.

src/libmongoc/src/mongoc/mongoc-cluster-private.h
```diff
@@ -186,6 +186,16 @@ mongoc_cluster_stream_valid (mongoc_cluster_t *cluster, mongoc_server_stream_t *
...
 mongoc_cluster_run_command_monitored (mongoc_cluster_t *cluster, mongoc_cmd_t *cmd, bson_t *reply, bson_error_t *error);
 
+// `mongoc_cluster_run_retryable_write` executes a write command and may apply retryable writes behavior.
+// `*retry_server_stream` is set to a new stream on retry. The caller must call `mongoc_server_stream_cleanup`.
+bool
+mongoc_cluster_run_retryable_write (mongoc_cluster_t *cluster,
+                                    mongoc_cmd_t *cmd,
+                                    bool is_retryable_write,
+                                    mongoc_server_stream_t **retry_server_stream,
+                                    bson_t *reply,
```
eramongodb, line 196: Document that `reply` must point to an _uninitialized_ `bson_t`, and that it is _always_ initialized by this function (must be destroyed by caller).

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -3636,3 +3636,98 @@ mcd_rpc_message_decompress_if_necessary (mcd_rpc_message *rpc, void **data, size
...
+         mongoc_deprioritized_servers_add_if_sharded (ds, cmd->server_stream->topology_type, cmd->server_stream->sd);
+
+         *retry_server_stream =
+            mongoc_cluster_stream_for_writes (cluster, cmd->session, ds, NULL /* reply */, &ignored_error);
+
+         mongoc_deprioritized_servers_destroy (ds);
+      }
+
+      if (*retry_server_stream) {
+         cmd->server_stream = *retry_server_stream;
```
eramongodb, line 3703: Is `retry_server_stream` as an _out parameter_ necessary if it is always obtainable via `cmd->server_stream` instead on retry?

Additionally, should this assignment be preceeded by a `mongoc_server_stream_destroy (cmd->server_stream)`? Recommend clarifying expectations of stream object lifetime in function documentation (e.g. is the caller required to have a reference to the old stream object with which they can destroy the stream object should a retry occur? or can they expect this function to destroy the old stream object for them on retry?).

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1562,97 +1562,20 @@ _mongoc_client_retryable_write_command_with_stream (mongoc_client_t *client,
...
-         }
-         bson_destroy (reply);
-         GOTO (retry);
-      }
-   }
+   ret = mongoc_cluster_run_retryable_write (
+      &client->cluster, &parts->assembled, true /* is_retryable */, &retry_server_stream, reply, error);
 
    if (retry_server_stream) {
       mongoc_server_stream_cleanup (retry_server_stream);
```
eramongodb, line 1576:
```suggestion
      mongoc_server_stream_cleanup (retry_server_stream);
      parts->assembled.server_stream = NULL;
```

This function cleans up `retry_server_stream` despite assigning it to `parts->assembled.server_stream`, which seems like opportunity for use-after-free by the caller. Given `mongoc_cluster_run_retryable_write` also unconditionally assigns-to `mongoc_cmd_t::server_stream` on retry, consider setting to `NULL` to reduce the possibility of undefined behavior.

eramongodb (changes requested):
Thank you very much for this refactor. This was long overdue. 👍

src/libmongoc/src/mongoc/mongoc-cluster-private.h
```diff
@@ -186,6 +186,18 @@ mongoc_cluster_stream_valid (mongoc_cluster_t *cluster, mongoc_server_stream_t *
 bool
 mongoc_cluster_run_command_monitored (mongoc_cluster_t *cluster, mongoc_cmd_t *cmd, bson_t *reply, bson_error_t *error);
 
+// `mongoc_cluster_run_retryable_write` executes a write command and may apply retryable writes behavior.
+// `cmd->server_stream` is expected to be non-owning and is overwritten on retry.
```
eramongodb, line 190:
```suggestion
// `cmd->server_stream` is set to `*retry_server_stream` on retry. Otherwise, it is unmodified.
```
Recommend documenting ownership in the function instead (it's really `mongoc_cmd_t`'s responsibility to document ownership, but alas):
```c
cmd->server_stream = *retry_server_stream; // Non-owning.
```
Non-ownership of `cmd->server_stream` is already implied by documentation of `*retry_server_stream` being required to be cleaned-up on retry.

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -3636,3 +3636,99 @@ mcd_rpc_message_decompress_if_necessary (mcd_rpc_message *rpc, void **data, size
...
+
+   // Ensure `*retry_server_stream` is always valid or null.
+   *retry_server_stream = NULL;
+
+retry:
+   ret = mongoc_cluster_run_command_monitored (cluster, cmd, reply, error);
+
+   if (is_retryable_write) {
+      _mongoc_write_error_handle_labels (ret, error, reply, cmd->server_stream->sd);
+      _mongoc_write_error_update_if_unsupported_storage_engine (ret, error, reply);
```
eramongodb, line 3680: Confirming: is inclusion of this call correct in all cases that are now defined in terms of `mongoc_cluster_run_retryable_write()`?

eramongodb (approved):

kkloberdanz (approved):

src/libmongoc/src/mongoc/mongoc-cluster.c
kevinAlbs: I expect "yes". `_mongoc_write_error_update_if_unsupported_storage_engine` implements [this spec requirement](https://github.com/mongodb/specifications/blob/b1e0274ab01183676ec3e21ecb2aa10807107bec/source/retryable-writes/retryable-writes.rst#supported-server-versions):

> Retryable writes are only supported by storage engines that support document-level locking. [...] Drivers MUST raise an actionable error message when the server returns code 20 with errmsg starting with "Transaction numbers".

Before 4d02878f70446968c499230c21f2c13ee0e805d9, `_mongoc_write_error_update_if_unsupported_storage_engine` was only called on the initial attempt (not the retry). If the error is rewritten, I expect no retry is attempted (server does not supporting retryable writes). Therefore, I expect `_mongoc_write_error_update_if_unsupported_storage_engine` is still only called on the initial attempt.

PR #1578, opened 2024-04-22
CDRIVER-4363 Support multiple OP_MSG doc sequences

kevinAlbs:
# Summary

This PR introduces private support for sending an OP_MSG with multiple document sequences (OP_MSG [Section with payloadType=1](https://github.com/mongodb/specifications/blob/fe860823426a7ab1d6e492e00e08439c8839ed71/source/message/OP_MSG.rst#op_msg-1))

# Background & Motivation

The C driver currently assumes outgoing commands include one OP_MSG document sequence. https://github.com/mongodb/specifications/pull/1534 is expected to require sending two document sequences:

> Drivers MUST use document sequences ([`OP_MSG`](../message/OP_MSG.rst) payload type 1) for the `ops` and `nsInfo` fields.

This PR tests sending two document sequences using `bulkWrite` with private API. No behavior changes are expected to public API.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-cluster.c
+++ b/src/libmongoc/src/mongoc/mongoc-cluster.c
@@ -3225,45 +3225,26 @@ _mongoc_cluster_run_opmsg_send (
       message_length += mcd_rpc_header_set_response_to (rpc, 0);
       message_length += mcd_rpc_header_set_op_code (rpc, MONGOC_OP_CODE_MSG);
 
-      size_t section_count = 1u;
-      if (cmd->payload) {
-         section_count++;
-      }
-      if (cmd->payload2) {
-         section_count++;
-      }
-      mcd_rpc_op_msg_set_sections_count (rpc, section_count);
+      BSON_ASSERT (cmd->payloads_count <= SIZE_MAX - 1u);
+      mcd_rpc_op_msg_set_sections_count (rpc, 1u + cmd->payloads_count);
 
       message_length += mcd_rpc_op_msg_set_flag_bits (rpc, flags);
       message_length += mcd_rpc_op_msg_section_set_kind (rpc, 0u, 0);
       message_length += mcd_rpc_op_msg_section_set_body (rpc, 0u, bson_get_data (cmd->command));
 
-      if (cmd->payload) {
-         BSON_ASSERT (bson_in_range_signed (size_t, cmd->payload_size));
+      for (size_t i = 0; i < cmd->payloads_count; i++) {
+         BSON_ASSERT (bson_in_range_signed (size_t, cmd->payloads[i].size));
 
          const size_t section_length =
-            sizeof (int32_t) + strlen (cmd->payload_identifier) + 1u + (size_t) cmd->payload_size;
+            sizeof (int32_t) + strlen (cmd->payloads[i].identifier) + 1u + (size_t) cmd->payloads[i].size;
          BSON_ASSERT (bson_in_range_unsigned (int32_t, section_length));
 
-         message_length += mcd_rpc_op_msg_section_set_kind (rpc, 1u, 1);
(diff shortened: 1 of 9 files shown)


src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -3225,24 +3225,26 @@ _mongoc_cluster_run_opmsg_send (
...
+      mcd_rpc_op_msg_set_sections_count (rpc, 1u + cmd->payloads_count);
 
       message_length += mcd_rpc_op_msg_set_flag_bits (rpc, flags);
       message_length += mcd_rpc_op_msg_section_set_kind (rpc, 0u, 0);
       message_length += mcd_rpc_op_msg_section_set_body (rpc, 0u, bson_get_data (cmd->command));
 
-      if (cmd->payload) {
-         BSON_ASSERT (bson_in_range_signed (size_t, cmd->payload_size));
+      for (size_t i = 0; i < cmd->payloads_count; i++) {
+         BSON_ASSERT (bson_in_range_signed (size_t, cmd->payloads[i].size));
```
eramongodb, line 3236:
```suggestion
      for (size_t i = 0; i < cmd->payloads_count; i++) {
         const mongoc_cmd_payload_t* const payload = cmd->payloads[i];

         BSON_ASSERT (bson_in_range_signed (size_t, payload->size));
```
Recommend defining a `mongoc_cmd_payload_t` struct to allow for code such as suggested (avoiding repetitive subscripts).

eramongodb (changes requested):

src/libmongoc/src/mongoc/mongoc-cluster.c
eramongodb: Fixed suggestion to take address of `.payloads` element. Alternatively:
```c
const mongoc_cmd_payload_t payload = cmd->payloads[i];

BSON_ASSERT (bson_in_range_signed (size_t, payload.size));
```
if the view-like ownership semantics of the local variable isn't considered too confusing.

src/libmongoc/src/mongoc/mongoc-cluster.c
```diff
@@ -3225,24 +3225,28 @@ _mongoc_cluster_run_opmsg_send (
       message_length += mcd_rpc_header_set_response_to (rpc, 0);
       message_length += mcd_rpc_header_set_op_code (rpc, MONGOC_OP_CODE_MSG);
 
-      mcd_rpc_op_msg_set_sections_count (rpc, cmd->payload ? 2u : 1u);
+      // Reserve one section for the body (kind 0).
+      BSON_ASSERT (cmd->payloads_count <= MONGOC_CMD_PAYLOADS_COUNT_MAX);
```
eramongodb, line 3229:
```suggestion
      BSON_ASSERT (cmd->payloads_count < MONGOC_CMD_PAYLOADS_COUNT_MAX + 1u);
```
Recommend using `<` and keeping the +1/-1 for clarity.

eramongodb (approved):

src/libmongoc/src/mongoc/mongoc-cluster.c
eramongodb: Scratch this suggestion: I think this assertion should simply be `cmd->payloads_count < MONGOC_CMD_PAYLOADS_COUNT_MAX` and be treated as no longer relevant to the body section reservation.

kevinAlbs: I think this should remain as-is with `<=`. `cmd->payloads_count` may be equal to `MONGOC_CMD_PAYLOADS_COUNT_MAX` (defined to 2) if there are two kind 1 payloads (aka document sequences). Updated comment to clarify.

eramongodb: Confused myself into thinking of the max as an exclusive bound rather than inclusive. LGTM.

PR #1584, opened 2024-04-25
CDRIVER-4363 add `nsInfo` builder

kevinAlbs:
# Summary

This PR adds a private component to build the `nsInfo` payload for CDRIVER-4363. No publicly visible changes are expected.

- Add private type `mcd_nsinfo_t` to build the `nsInfo` payload in the `bulkWrite` command.
- Add [uthash](https://troydhanson.github.io/uthash/) 2.3.0.

# Background & Motivation

The `mcd_nsinfo_t` is intended for upcoming support of the `bulkWrite` command. Each operation in a `bulkWrite` references an int32 index into an array of namespaces to reduce duplication of namespace strings:

```yml
{
    "bulkWrite": 1,
    "ops": [
        {
            "insert": 0, # references 'db.coll1'
            "document": { "_id": "foo" }
        },
        {
            "insert": 0, # references 'db.coll1'
            "document": { "_id": "bar" }
        },
        {
            "insert": 1, # references 'db.coll2'
            "document": { "_id": "baz" }
        }
    ],
    "nsInfo": [
        { "ns": "db.coll1" },
        { "ns": "db.coll2" }
    ]
}
```

The syntax of the `bulkWrite` command is further described in [Scope: Server improved bulk write command](https://docs.google.com/document/d/1z1SFTRy9O0i5FtXEwQwc0wfIjQniqbMEaIDonsphG9k/edit#heading=h.haavpvf17xzz).

Each operation can refer to a different namespace. There can be up to `maxWriteBatchSize` operations in one command. The server currently defines `maxWriteBatchSize` in the response to the `hello` command as 100,000.

My initial implementation stored namespaces in a `bson_t` and did linear look-up. This resulted in an O(n^2) algorithm which took very long to execute for a large number of namespaces (I stopped waiting after several minutes).

A hash table with faster look-up seemed needed. uthash was chosen for simplicity (small header-only library) and consistency with other dependencies used (utlist.h is already copied).

### DIFF
--- a/src/libmongoc/src/mongoc/mcd-nsinfo.c
+++ b/src/libmongoc/src/mongoc/mcd-nsinfo.c
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2024-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <mcd-nsinfo.h>
+
+#include <mongoc/mongoc-buffer-private.h>
+#include <mongoc/mongoc-error.h>
+#include <mongoc/uthash.h>
+
+typedef struct {
+   char *ns; // Hash key.
+   int32_t index;
+   UT_hash_handle hh;
+} ns_to_index_t;
+
+struct _mcd_nsinfo_t {
+   ns_to_index_t *n2i;
+   int32_t count;
+   mongoc_buffer_t payload;
+};
+
+mcd_nsinfo_t *
+mcd_nsinfo_new (void)
+{
+   mcd_nsinfo_t *self = bson_malloc0 (sizeof (*self));
+   _mongoc_buffer_init (&self->payload, NULL, 0, NULL, NULL);
+   return self;
+}
+
+void
+mcd_nsinfo_destroy (mcd_nsinfo_t *self)
+{
+   if (!self) {
+      return;
+   }
+   // Delete hash table.
+   ns_to_index_t *entry, *tmp;
+   HASH_ITER (hh, self->n2i, entry, tmp)
+   {
+      HASH_DEL (self->n2i, entry);
+      bson_free (entry->ns);
+      bson_free (entry);
+   }
+   _mongoc_buffer_destroy (&self->payload);
+   bson_free (self);
+}
+
+int32_t
+mcd_nsinfo_append (mcd_nsinfo_t *self, const char *ns, bson_error_t *error)
+{
+   BSON_ASSERT_PARAM (self);
+   BSON_ASSERT_PARAM (ns);
+   BSON_ASSERT (error || true);
+
+   int32_t ns_index = self->count;
+   if (self->count == INT32_MAX) {
+      bson_set_error (error,
+                      MONGOC_ERROR_COMMAND,
+                      MONGOC_ERROR_COMMAND_INVALID_ARG,
+                      "Only %" PRId32 " distinct collections may be used",
+                      INT32_MAX);
+      return -1;
+   }
+   self->count++;
+
+   // Add to hash table.
+   ns_to_index_t *entry = bson_malloc0 (sizeof (*entry));
+   entry->index = ns_index;
+   entry->ns = bson_strdup (ns);
+   HASH_ADD_KEYPTR (hh, self->n2i, entry->ns, strlen (entry->ns), entry);
+
+   // Append to buffer.
+   bson_t mcd_nsinfo_bson = BSON_INITIALIZER;
+   BSON_ASSERT (bson_append_utf8 (&mcd_nsinfo_bson, "ns", 2, ns, -1));
+   BSON_ASSERT (_mongoc_buffer_append (&self->payload, bson_get_data (&mcd_nsinfo_bson), mcd_nsinfo_bson.len));
+   bson_destroy (&mcd_nsinfo_bson);
+   return ns_index;
+}
+
+int32_t
+mcd_nsinfo_find (const mcd_nsinfo_t *self, const char *ns)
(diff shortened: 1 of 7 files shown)


src/libmongoc/src/mongoc/uthash.h
eramongodb, line 1: Can we consider moving this header (+ utlist.h) into a dedicated directory (like kms_message, utf8proc, and zlib) for the uthash "library" of header files so it is structurally clearer that these are bundled third-party source files that are not owned/maintained by us?

```
src/
├── common/
├── libbson/
├── libmongoc/
├── kms_message/
├── uthash-2.3.0/ # <-- New.
│   ├── utlist.h
│   └── uthash.h
├── utf8proc-2.8.0/
├── zlib-1.2.13/
└── ...
```

Also consider taking this opportunity to upgrade the `utlist.h` header (currently v1.9.9) to v2.3.0.

src/libmongoc/src/mongoc/uthash.h
```diff
@@ -0,0 +1,1145 @@
...
+#define UTHASH_H
+
+// libmongoc specific code ... begin
+// clang-format off
+#include <mongoc-prelude.h>
+#include <bson/bson.h>
+// Use libbson's malloc/free functions in uthash.
+#define uthash_malloc(sz) bson_malloc (sz)
+#define uthash_free(ptr, sz) bson_free (ptr)
+// libmongoc specific code ... end
```
eramongodb, line 34: Rather than modifying bundled source files, consider using an indirect header file instead:
```c
// src/libmongoc/src/mongoc/uthash.h

#ifndef MONGOC_UTHASH_H
#define MONGOC_UTHASH_H

#include <mongoc-prelude.h>
#include <bson/bson.h>

#define uthash_malloc(sz) bson_malloc (sz)
#define uthash_free(ptr, sz) bson_free (ptr)

#include <uthash-2.3.0/uthash.h>

#endif // MONGOC_UTHASH_H
```

Similarly for the existing `utlist.h` header.

src/libmongoc/tests/test-mcd-nsinfo.c
```diff
@@ -0,0 +1,95 @@
...
+   }
+
+   mcd_nsinfo_destroy (nsinfo);
+}
+
+static void
+test_nsinfo_handles_100k_namespaces (void)
+{
+   // Test repeated finding and adding 100,000 unique namespaces.
+   // A `bulkWrite` command supports a maximum of maxWriteBatchSize unique namespaces (currently 100,000).
```
eramongodb, line 48:
```suggestion
   // A `bulkWrite` command supports a maximum of maxWriteBatchSize unique namespaces (currently 100,000).
   const size_t ns_count = 100000;
```
Instead of repeated use of the `100000` literal.

src/libmongoc/src/mongoc/mcd-nsinfo.c
```diff
@@ -0,0 +1,127 @@
...
+}
+
+int32_t
+mcd_nsinfo_append (mcd_nsinfo_t *self, const char *ns, bson_error_t *error)
+{
+   BSON_ASSERT_PARAM (self);
+   BSON_ASSERT_PARAM (ns);
+   BSON_ASSERT (error || true);
+
+   int32_t ns_index = self->count;
```
eramongodb, line 68:
```suggestion
   const int32_t ns_index = self->count;
```

src/libmongoc/src/mongoc/mcd-nsinfo.c
```diff
@@ -0,0 +1,127 @@
...
+                      INT32_MAX);
+      return -1;
+   }
+   self->count++;
+
+   // Add to hash table.
+   ns_to_index_t *entry = bson_malloc0 (sizeof (*entry));
+   entry->index = ns_index;
+   entry->ns = bson_strdup (ns);
+   HASH_ADD_KEYPTR (hh, self->n2i, entry->ns, strlen (entry->ns), entry);
```
eramongodb, line 83:
```suggestion
   ns_to_index_t *entry = bson_malloc (sizeof (*entry));
   *entry = (ns_to_index_t){
      .index = ns_index,
      .ns = bson_strdup (ns),
      .hh = NULL,
   };
   HASH_ADD_KEYPTR (hh, self->n2i, entry->ns, strlen (entry->ns), entry);
```
Avoid `malloc + memset + assignment` initialization.

src/libmongoc/tests/test-mcd-nsinfo.c
```diff
@@ -0,0 +1,95 @@
...
+{
+   mcd_nsinfo_t *nsinfo = mcd_nsinfo_new ();
+
+   // Add several namespaces.
+   bson_error_t error;
+   ASSERT_OR_PRINT (0 <= mcd_nsinfo_append (nsinfo, "db.coll1", &error), error);
+   ASSERT_OR_PRINT (0 <= mcd_nsinfo_append (nsinfo, "db.coll2", &error), error);
+   ASSERT_OR_PRINT (0 <= mcd_nsinfo_append (nsinfo, "db.coll3", &error), error);
+
+   // Check resulting indexes.
```
eramongodb, line 17: Confirming: are the indices expected to be consistently sequenced in the order the nsinfo were appended? If so, recommend documenting this behavior (deterministic indices). If not, recommend changing these assertions to `0 <=` instead of asserting specific indices.

eramongodb (changes requested):

src/libmongoc/src/mongoc/uthash.h
kevinAlbs: Suggestion applied with an added `uthash` directory to include `uthash-2.3.0` in the include path.

`utlist.h` has also been updated to v2.3.0.

src/libmongoc/src/mongoc/uthash.h
kevinAlbs: Thank you for the suggestion. I very much like the idea of an indirect header. That enables keeping an unmodified `uthash.h` header. Suggestion has been applied for both `uthash.h` and `utlist.h`.

src/libmongoc/tests/test-mcd-nsinfo.c
kevinAlbs: Indexes are expected to be assigned in the order they are added. I expect it is needed to pass specification tests (example) that assert the contents of the `nsInfo` sent.

kevinAlbs (commented):
Thank you for the suggestions.

`utlist.h` has also been updated to 2.3.0.

Patch build to verify: https://spruce.mongodb.com/version/662bc956c099380007552450

eramongodb (approved):

PR #1585, opened 2024-04-26
fix crash if password is NULL

melisha-git:
If I have a custom connection string without password - 'mongodb://superuser@127.0.0.1:27030/admin', then mongo cursor crashes

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-scram.c
+++ b/src/libmongoc/src/mongoc/mongoc-scram.c
@@ -995,7 +995,7 @@ bool
 _mongoc_sasl_prep_required (const char *str)
 {
    unsigned char c;
-   while (*str) {
+   while (str != NULL && *str) {
       c = (unsigned char) *str;
       /* characters below 32 contain all of the control characters.
        * characters above 127 are multibyte UTF-8 characters.


melisha-git:
how i can restart first and second piplines? This is random? Fix is very small

kevinAlbs (commented):
Thank you for the contribution. This appears to fix the crash, but results in no error if auth fails with an empty password. Example:
```c
bson_error_t error = {0};
mongoc_client_t *client = mongoc_client_new ("mongodb://foobar@localhost:27017/admin");
mongoc_client_command_simple (
    client, "admin", tmp_bson ("{'ping': 1}"), NULL /* read prefs */, NULL /* reply */, &error);
if (error.code == 0) {
    fprintf (stderr, "Expected error to be set, but was not.\n");
    abort ();
}
```

https://github.com/mongodb/mongo-c-driver/pull/1586 is a suggested alternative fix and includes a regression test. If you would like to verify that change fixes the issue, that would be appreciated.

---

Aside: I was unable to successfully auth with a user with an empty password. I am not sure if it is expected to succeed. Local tests suggest servers do not support creating users with empty passwords:

```sh
$ mongosh --eval 'print("Server version: " + db.version()); db.createUser({"user":"foobar", "pwd":"", "roles": ["readWrite"]})' --quiet --username user --password password
Server version: 4.0.27
MongoServerError: User passwords must not be empty
```

Is authenticating with an empty password expected to succeed for your use case? If "yes", what server version are you running? How was the user created? Are you able to auth in another driver or `mongosh`? (I think the crash should be fixed regardless, but if an empty password is not expected to succeed, the driver could instead return error for an empty password).

---

> how i can restart first and second piplines? This is random? Fix is very small

External contributors are unable to start CI (Evergreen). The test failures are unrelated to the changes in this PR.

melisha-git:
but the driver crashes

melisha-git:
https://github.com/mongodb/mongo-c-driver/pull/1586 is a suggested alternative fix and includes a regression test. If you would like to verify that change fixes the issue, that would be appreciated.

Yes, i test it tomorrow

PR #1593, opened 2024-05-02
CDRIVER-5552 more robust string handling

rcsanchez97:


src/libbson/src/bson/bson-string.c
```diff
@@ -61,16 +62,23 @@ bson_string_t *
 bson_string_new (const char *str) /* IN */
 {
    bson_string_t *ret;
+   size_t len_sz;
 
    ret = bson_malloc0 (sizeof *ret);
-   ret->len = str ? (int) strlen (str) : 0;
+   if (str) {
+      len_sz = strlen (str);
+      BSON_ASSERT (bson_in_range_unsigned (uint32_t, len_sz + 1));
```
kevinAlbs, line 70: If `len_sz` is `SIZE_MAX`, then `len_sz + 1` may overflow. Suggest changing to:

```suggestion
      BSON_ASSERT (len_sz <= UINT32_MAX - 1);
```

src/libbson/src/bson/bson-string.c
```diff
@@ -126,17 +134,22 @@ bson_string_append (bson_string_t *string, /* IN */
...
    BSON_ASSERT (string);
    BSON_ASSERT (str);
 
-   len = (uint32_t) strlen (str);
+   len_sz = strlen (str);
+   BSON_ASSERT (bson_in_range_unsigned (uint32_t, len_sz));
+   len = (uint32_t) len_sz;
 
    if ((string->alloc - string->len - 1) < len) {
+      BSON_ASSERT (bson_in_range_unsigned (uint32_t, string->alloc + len));
```
kevinAlbs, line 147: Similar to above, `string->alloc + len` may overflow. Suggested alternative:

```suggestion
      BSON_ASSERT (string->alloc <= UINT32_MAX - len);
```

rcsanchez97:
@kevinAlbs Sorry about the conflict. I pushed a change before I saw that you had made suggested changes. I implemented the equivalents of your suggestions in a01e7db8f.

src/libbson/src/bson/bson-string.c
```diff
@@ -61,16 +62,23 @@ bson_string_t *
 bson_string_new (const char *str) /* IN */
 {
    bson_string_t *ret;
+   size_t len_sz;
 
    ret = bson_malloc0 (sizeof *ret);
-   ret->len = str ? (int) strlen (str) : 0;
+   if (str) {
+      len_sz = strlen (str);
+      BSON_ASSERT (len_sz <= INT32_MAX);
```
kevinAlbs, line 70: On closer read, though bsonspec.org implies the limit of a string within BSON is `INT32_MAX`, I expect `bson_string_t` does not have the same limit. A `bson_string_t` is a string type within libbson, not necessary used within a BSON document.

Suggest assuming limit `UINT32_MAX` to match the type of `bson_string_t::len`:

```suggestion
      BSON_ASSERT (len_sz <= UINT32_MAX);
```

src/libbson/src/bson/bson-string.c
```diff
@@ -61,16 +62,23 @@ bson_string_t *
...
+      ret->len = (uint32_t) len_sz;
+   } else {
+      ret->len = 0;
+   }
    ret->alloc = ret->len + 1;
 
    if (!bson_is_power_of_two (ret->alloc)) {
       ret->alloc = (uint32_t) bson_next_power_of_two ((size_t) ret->alloc);
```
kevinAlbs, line 78:
```suggestion
      len_sz = bson_next_power_of_two ((size_t) ret->alloc);
      BSON_ASSERT (len_sz <= UINT32_MAX);
      ret->alloc = (uint32_t) len_sz;
```

Suggest checking for possible overflow before cast.

src/libbson/src/bson/bson-string.c
```diff
@@ -126,17 +134,24 @@ bson_string_append (bson_string_t *string, /* IN */
...
 {
    uint32_t len;
+   size_t len_sz;
 
    BSON_ASSERT (string);
    BSON_ASSERT (str);
 
-   len = (uint32_t) strlen (str);
+   len_sz = strlen (str);
+   BSON_ASSERT (bson_in_range_unsigned (int32_t, len_sz));
```
kevinAlbs, line 143:
```suggestion
   BSON_ASSERT (bson_in_range_unsigned (uint32_t, len_sz));
```

src/libbson/src/bson/bson-string.c
```diff
@@ -126,17 +134,24 @@ bson_string_append (bson_string_t *string, /* IN */
...
+   len = (uint32_t) len_sz;
 
    if ((string->alloc - string->len - 1) < len) {
+      BSON_ASSERT (string->alloc <= INT32_MAX - len);
```
kevinAlbs, line 147:
```suggestion
      BSON_ASSERT (string->alloc <= UINT32_MAX - len);
```

src/libbson/src/bson/bson-string.c
rcsanchez97: I don't think we can do that. The `alloc` field is a `uint32_t`. If we allow a length anywhere from `INT32_MAX + 2` (2147483649) up to `UINT32_MAX` (4294967295), then the next power of two, 4294967296, will overflow the `uint32_t`. Setting the limit at `INT32_MAX` is the only way to ensure that the *next* power of 2 still fits in a `uint32_t`.

rcsanchez97: Put more succinctly, the root of the problem is that we want to allocate memory in increments of powers of 2, but the integer a `uint32_t` can hold is not a power of 2 (i.e., `2^32 - 1`)

rcsanchez97: OK. I ran some local tests, and it appears that I was only half right.

It seems like we can assert based on `UINT32_MAX` in all cases and in the event that the size of the string to create/append requires allocating all the way to 2^32, then the assertion fails.

kevinAlbs (approved):

PR #1595, opened 2024-05-04
Fix static build

jingjingxyk:


src/libmongoc/CMakeLists.txt
```diff
@@ -815,6 +837,61 @@ if (MONGOC_ENABLE_STATIC_BUILD)
    set_target_properties (mcd_rpc PROPERTIES OUTPUT_NAME "mcd-rpc")
 endif ()
 
+if (BUILD_SHARED_LIBS)
```
kevinAlbs, line 840: This block closely resembes the `if (ENABLE_SHARED)` block below. Is it needed?

kevinAlbs (commented):
@jingjingxyk is this PR still relevant? If "yes", can you describe the motivation?

This PR introduces the CMake options `BUILD_SHARED_LIBS` and `BUILD_STATIC_LIBS`. I am not sure what value they provide over the existing `ENABLE_STATIC` and `ENABLE_SHARED` CMake options.

PR #1605, opened 2024-05-17
Add mongoc_cursor_get_batch_num

Roman-Koshelev:


kevinAlbs:
Thank you for the PR @Roman-Koshelev. Can you describe the use case for `mongoc_cursor_get_batch_num`? There may be other ways to solve the problem (e.g. [command monitoring](https://mongoc.org/libmongoc/current/application-performance-monitoring.html#command-monitoring-example) can be used to get information about commands sent to the server).

Roman-Koshelev:
Usage example https://github.com/userver-framework/userver/blob/develop/mongo/src/storages/mongo/cdriver/cursor_impl.cpp#L68 .
If there is a more elegant solution then I'm happy to use it

Roman-Koshelev:
Did it in https://github.com/userver-framework/userver/commit/9860015811e6139991709579cb881b207f896764

PR #1606, opened 2024-05-17
CDRIVER-5575 Set error offset on subdocument sanity check failures

Roman-Koshelev:
in bson_iter_visit_all


kevinAlbs (changes requested):
Thank you for the contribution! LGTM with one requested change.

kevinAlbs (approved):

PR #1618, opened 2024-05-28
CDRIVER-3645 Close operation connections to removed servers

kevinAlbs:
# Summary

Close operation connections to removed servers.

Verified with this patch build: https://spruce.mongodb.com/version/665489593f686b000739adf0

# Background & Motivation

Motivated by an issue reported in [HELP-59749](https://jira.mongodb.org/browse/HELP-59749). Tests observed an unexpected increasing connection count. This may have been triggered by (misconfigured?) DNS records resulting in servers being frequently added/removed from the topology during SRV polling.

## Rejected alternative: callback

I considered adding a callback to `_mongoc_topology_description_remove_server` to more immediately signal to close connections. This was rejected to reduce risk of an AB-BA deadlock.

`_mongoc_topology_description_remove_server` expects the caller to hold `mongoc_topology_t::tpld_modification_mtx`. Modifying pooled client requires holding `mongoc_client_pool_t::mutex`.

There is an existing pattern of locking `mongoc_client_pool_t::mutex` then `mongoc_topology_t::tpld_modification_mtx` (via `mongoc_client_pool_try_pop` => `_start_scanner_if_needed` => `_mongoc_topology_background_monitoring_start` => `mc_tpld_modify_begin`).

Instead, this PR proposes lazily pruning stale connections on `mongoc_client_pool_push`. I expect multi-threaded applications to pop/push clients frequently to encourage reuse of clients among threads. A note is added to documentation to make this recommendation explicit.

## Performance Impact

[A benchmark](https://spruce.mongodb.com/version/6653e2ad7f76a90007f5b338) was run to compare throughput of repeated use of the pool through pop+command+push with pruning:

| Test                                   |   Baseline (mb/sec) | With simple pruning   | With conditional pruning   |
|----------------------------------------|------------|-----------------------|----------------------------|
| Parallel/Pool/ShareClients/Threads:1   |     251995 | 236063 (-6.32%)       | 248131 (-1.53%)            |
| Parallel/Pool/ShareClients/Threads:10  |    1364224 | 873991 (-35.93%)      | 1337326 (-1.97%)           |
| Parallel/Pool/ShareClients/Threads:100 |    2017530 | 1050534 (-47.93%)     | 2020829 (0.16%)            |
| Parallel/Pool/ShareClients/Threads:150 |    1540427 | 1571373 (2.01%)       | 1552121 (0.76%)            |

An initial solution ("With simple pruning") iterated over all pooled clients to prune on each call to `mongoc_client_pool_push`. Due to the performance impact, checks were added to only prune if topology description change was detected ("With conditional pruning"). This motivated storing the `last_known_serverids` in the `mongoc_client_pool_t` to quickly determine if the set of server IDs changes. I expect the set of server IDs changing is a "rare" event (e.g. HELP-59749 notes an extreme case where SRV records unexpectedly change. Even then, I expect the set of server IDs would change about once per minute).

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-client-pool.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-pool.c
@@ -370,8 +370,8 @@ typedef struct {
 static int
 server_id_cmp (const void *a_, const void *b_)
 {
-   uint32_t *a = (uint32_t *) a_;
-   uint32_t *b = (uint32_t *) b_;
+   const uint32_t *const a = (const uint32_t *) a_;
+   const uint32_t *const b = (const uint32_t *) b_;
 
    if (*a == *b) {
       return 0;
--- a/src/libmongoc/src/mongoc/mongoc-client-pool.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-pool.c
@@ -378,7 +378,7 @@ server_id_cmp (const void *a_, const void *b_)
       return 0;
    }
 
-   return a < b ? -1 : 1;
+   return *a < *b ? -1 : 1;
 }
 
 static bool
--- a/src/libmongoc/src/mongoc/mongoc-client-pool.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-pool.c
@@ -406,33 +406,6 @@ prune_client (mongoc_client_t *client, mongoc_array_t *known_server_ids)
    BSON_ASSERT_PARAM (known_server_ids);
 
    mongoc_cluster_t *cluster = &client->cluster;
-
-   bool needs_prune = false;
-   // Do a fast initial check to see if a prune is needed.
-   size_t idx1 = 0;
-   size_t idx2 = 0;
-   for (; idx1 < cluster->nodes->items_len; idx1++) {
-      // Compare both sorted lists in order. `cluster->nodes` may be a smaller list if not all servers were used.
-      mongoc_set_item_t *cn = &cluster->nodes->items[idx1];
-      bool found = false;
-      for (; idx2 < known_server_ids->len; idx2++) {
-         uint32_t last_known = _mongoc_array_index (known_server_ids, uint32_t, idx2);
-         if (cn->id == last_known) {
-            found = true;
-            break;
-         }
-      }
-      if (!found) {
-         // A server in the cluster is not in the last known server ids. Prune it.
-         needs_prune = true;
-         break;
-      }
-   }
-
-   if (!needs_prune) {
-      return;
-   }
-
    prune_ctx ctx = {.cluster = cluster, .known_server_ids = known_server_ids};
    mongoc_set_for_each (cluster->nodes, maybe_prune, &ctx);
 }
--- a/src/libmongoc/src/mongoc/mongoc-client-pool.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-pool.c
@@ -17,6 +17,7 @@
 
 #include "mongoc.h"
 #include "mongoc-apm-private.h"
+#include "mongoc-array-private.h"
 #include "mongoc-counters-private.h"
 #include "mongoc-client-pool-private.h"
 #include "mongoc-client-pool.h"
--- a/src/libmongoc/src/mongoc/mongoc-client-pool.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-pool.c
@@ -449,33 +449,29 @@ mongoc_client_pool_push (mongoc_client_pool_t *pool, mongoc_client_t *client)
    // Check if `last_known_server_ids` needs update.
    bool serverids_have_changed = false;
    {
-      mc_shared_tpld td = mc_tpld_take_ref (pool->topology);
-      const mongoc_set_t *servers = mc_tpld_servers_const (td.ptr);
+      mongoc_array_t current_serverids;
+      _mongoc_array_init (&current_serverids, sizeof (uint32_t));
 
-      if (pool->last_known_serverids.len != servers->items_len) {
-         serverids_have_changed = true;
-      } else {
-         // Check if any server IDs have changed.
+      {
+         mc_shared_tpld td = mc_tpld_take_ref (pool->topology);
+         const mongoc_set_t *servers = mc_tpld_servers_const (td.ptr);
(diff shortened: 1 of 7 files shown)


src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -370,6 +445,52 @@ mongoc_client_pool_push (mongoc_client_pool_t *pool, mongoc_client_t *client)
...
+      }
+
+      if (serverids_have_changed) {
+         // Store new set of server IDs.
+         _mongoc_array_destroy (&pool->last_known_serverids);
+         _mongoc_array_init (&pool->last_known_serverids, sizeof (uint32_t));
+         for (size_t i = 0; i < servers->items_len; i++) {
+            _mongoc_array_append_val (&pool->last_known_serverids, servers->items[i].id);
+         }
+      }
```
eramongodb, line 476: This algorithm iterates over `servers->items` twice: once for ID comparisons then once more for conditional new set storage.

Recommend refactoring so that the sorted array of "new" server IDs is unconditionally created first (which reduces time spend holding the topology mutex):

```c
mongoc_array_t current_serverids;
_mongoc_array_init (&current_serverids, sizeof (uint32_t));

{
   mc_shared_tpld td = mc_tpld_take_ref (pool->topology);
   const mongoc_set_t *servers = mc_tpld_servers_const (td.ptr);
   for (size_t i = 0; i < servers->items_len; i++) {
      _mongoc_array_append_val (&current_serverids, servers->items[i].id);
   }
   mc_tpld_drop_ref (&td);
}
```

`bool serverids_have_changed` may then set with a size-cmp + `memcmp`, taking advantage of the contiguous (already-sorted) array elements:

```c
serverids_have_changed = (current_serverids.len != pool->last_known_serverids.len) ||
                         memcmp (current_serverids.data,
                                 pool->last_known_serverids.data,
                                 current_serverids.len * current_serverids.element_size) != 0;
```

The conditional storage may then be a "move" of the array value:

```c
_mongoc_array_destroy (&pool->last_known_serverids);
pool->last_known_serverids = current_serverids; // Ownership transfer.
```

I expect this refactor to slightly improve the performance benchmarks for conditional pruning.

eramongodb (changes requested):

src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -357,6 +363,76 @@ mongoc_client_pool_try_pop (mongoc_client_pool_t *pool)
...
+server_id_cmp (const void *a_, const void *b_)
+{
+   const uint32_t *const a = (const uint32_t *) a_;
+   const uint32_t *const b = (const uint32_t *) b_;
+
+   if (*a == *b) {
+      return 0;
+   }
+
+   return a < b ? -1 : 1;
```
eramongodb, line 381: Confirming, is `a < b` intentional instead of `*a < *b`? Recommend adding a comment here documenting so due to its unusual behavior.

eramongodb (approved):

src/libmongoc/src/mongoc/mongoc-client-pool.c
kevinAlbs: No, that was a (very well spotted) bug. Fixed.

src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -357,6 +363,76 @@ mongoc_client_pool_try_pop (mongoc_client_pool_t *pool)
...
+{
+   mongoc_cluster_node_t *cn = (mongoc_cluster_node_t *) item;
+   prune_ctx *ctx = (prune_ctx *) ctx_;
+   uint32_t server_id = cn->handshake_sd->id;
+
+   if (!bsearch (&server_id, ctx->server_ids->data, ctx->server_ids->len, sizeof (uint32_t), server_id_cmp)) {
+      mongoc_cluster_disconnect_node (ctx->cluster, server_id);
+   }
+   return true;
+}
```
vector-of-bool, line 395: Recommend adding commentary on what/why this function is does what it does.

src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -357,6 +363,76 @@ mongoc_client_pool_try_pop (mongoc_client_pool_t *pool)
...
+      if (!found) {
+         // A server in the cluster is not in the last known server ids. Prune it.
+         needs_prune = true;
+         break;
+      }
+   }
+
+   if (!needs_prune) {
+      return;
+   }
```
vector-of-bool, line 430: Is it likely that this early-return saves a significant amount of time for the additional complexity that it adds?

vector-of-bool (approved):
LGTM with minor remarks

kkloberdanz (approved):

src/libmongoc/src/mongoc/mongoc-client-pool.c
kevinAlbs: Good question. Another benchmark does not show significant change with [this early-return removed](https://spruce.mongodb.com/task/mongo_c_driver_perf_c_driver_benchmark_mongo60_perf_BenchMarkTests_patch_9fe27e912525ca127ef889660cb9ebfa802bbe5c_665871d1450b780007ad9fde_24_05_30_12_32_18/logs?execution=0):


| Test                                   |   Baseline (all conditional pruning) | Without conditional client pruning   | Without conditional pruning (simple prune)   |
|----------------------------------------|--------------------------------------|--------------------------------------|----------------------------------------------|
| Parallel/Pool/ShareClients/Threads:1   |                               249159 | 246554 (-1.05%)                      | 235143 (-5.63%)                              |
| Parallel/Pool/ShareClients/Threads:10  |                              1312571 | 1313206 (0.05%)                      | 866941 (-33.95%)                             |
| Parallel/Pool/ShareClients/Threads:100 |                              1932369 | 1917569 (-0.77%)                     | 1194404 (-38.19%)                            |
| Parallel/Pool/ShareClients/Threads:150 |                              1555300 | 1564356 (0.58%)                      | 1564161 (0.57%)                              |

The "Baseline" is the current changeset. The "Without conditional client pruning" removes this early-return.

Updated to remove the early-return.

PR #1619, opened 2024-05-30
[CDRIVER-5535] Create asset groups in Silk

vector-of-bool:
This changeset includes scripting and EVG tasks to create Silk asset groups for branches in the repository. The EVG task is idempotent: Once the asset group is created for a branch, the task becomes a no-op.

The creation script is parameterized and can be re-used between projects. The EVG task has been grouped under a new "Miscellaneous" build variant for the sole purpose of such one-off CI tasks.

Currently, the EVG task fails as I do not yet have the Silk credentials loaded. This will be fixed when such credentials are obtained and defined for the CI process.


Earthfile
```diff
@@ -191,6 +191,37 @@ sbom-generate:
     # Save the result back to the host:
     SAVE ARTIFACT /s/cyclonedx.sbom.json AS LOCAL etc/cyclonedx.sbom.json
 
+# silk-create-asset-group :
```
kevinAlbs, line 194:
```suggestion
# create-silk-asset-group :
```

To match target name.

kevinAlbs (approved):

PR #1624, opened 2024-06-03
CDRIVER-4162 remove remaining solaris workarounds

laurelxiang:
[evergreen](https://spruce.mongodb.com/version/6659f980b0f8570007c19bdf/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

This pr removes preprocessor checks for __sun since solaris is not longer supported.

### DIFF
--- a/src/libbson/tests/test-json.c
+++ b/src/libbson/tests/test-json.c
@@ -1927,7 +1927,7 @@ test_bson_json_double (void)
 
 /* check that "x" is -0.0. signbit not available on Solaris, FreeBSD, or VS 2010
  */
-#if !defined(__sun) && !defined(__FreeBSD__) && (!defined(_MSC_VER) || (_MSC_VER >= 1800))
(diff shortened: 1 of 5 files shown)


src/libbson/tests/test-json.c
```diff
@@ -1927,7 +1927,7 @@ test_bson_json_double (void)
 
 /* check that "x" is -0.0. signbit not available on Solaris, FreeBSD, or VS 2010
```
kevinAlbs, line 1928:
```suggestion
/* check that "x" is -0.0. signbit not available on FreeBSD, or VS 2010
```

kevinAlbs (changes requested):
Suggest removing `__sun` references in mongoc-counters-private.h, mongoc-errno-private.h.

Recommend keeping the `__sun` check in mongoc-handhake-os-private.h since that may help identify an unexpected Solaris user, and IMO does not add much maintenance burden.

kevinAlbs (approved):

PR #1629, opened 2024-06-06
[CDRIVER-5535] Add an +sbom-download target for getting the augmented SBOM

vector-of-bool:
This changeset progresses along CDRIVER-5535 by adding automation to obtain an augmented SBOM for a particular branch. This is a follow-up to #1594 and #1619.

This defines an Earthly target `+sbom-download` that downloads the latest augmented SBOM file from Silk baesd on the specified Git branch. This artifact is not currently used, but will be part of upcoming changes related to a new signed release archive artifact for CDRIVER-5537.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -171,15 +171,19 @@ multibuild:
         --c_compiler=gcc --c_compiler=clang \
         --test_mongocxx_ref=master
 
+# This target is simply an environment in which the SilkBomb executable is available.
+silkbomb:
+    FROM artifactory.corp.mongodb.com/release-tools-container-registry-public-local/silkbomb:1.0
+    # Alias the silkbom executable to a simpler name:
+    RUN ln -s /python/src/sbom/silkbomb/bin /usr/local/bin/silkbomb
+
 # sbom-generate :
 #   Generate/update the etc/cyclonedx.sbom.json file from the etc/purls.txt file.
 #
 # This target will update the existing etc/cyclonedx.sbom.json file in-place based
 # on the content of etc/purls.txt.
 sbom-generate:
-    FROM artifactory.corp.mongodb.com/release-tools-container-registry-public-local/silkbomb:1.0
-    # Alias the silkbom executable to a simpler name:
-    RUN ln -s /python/src/sbom/silkbomb/bin /usr/local/bin/silkbomb
+    FROM +silkbomb
     # Copy in the relevant files:
     WORKDIR /s
     COPY etc/purls.txt etc/cyclonedx.sbom.json /s/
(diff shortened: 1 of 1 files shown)


eramongodb (approved):
Small thoughts; otherwise, LGTM.

Earthfile
```diff
@@ -191,6 +195,26 @@ sbom-generate:
...
+    FROM +silkbomb
+    # Set --no-cache, because the remote artifact could change arbitrarily over time
+    RUN --no-cache \
+        --secret SILK_CLIENT_ID \
+        --secret SILK_CLIENT_SECRET \
+        silkbomb download \
+            --sbom-out augmented-sbom.json \
+            --silk-asset-group mongo-c-driver-${branch}
+    # Export as /augmented-sbom.json
+    SAVE ARTIFACT augmented-sbom.json
```
eramongodb, line 216: Should this be saved to `etc/augmented-sbom.json` so it is placed alongside the SBOM Lite document?

Aside: I'm considering naming the Augmented SBOM `etc/augmented.sbom.json` for the CXX Driver. Thoughts? (There doesn't seem to be a standard/recommended naming scheme for this file.)

eramongodb (commented):
Whoops, forgot to post my "small thoughts".

Earthfile
vector-of-bool: IIUC, the augmented SBOM should be generated on-the-fly (since it can be updated asynchronously) as part of a release artifact, so it shouldn't matter where it goes. If my understanding is incorrect, please correct me.

I have no strong opinion on the naming of the file itself. In this case I just named it by whatever felt intuitive at the moment. If it _is_ an ephemeral artifact, then the name can be changed arbitrarily in the future.

Overall I don't like the way this particular artifact is generated and handled, but that may be unavoidable.

eramongodb: > should be generated on-the-fly [...] so it shouldn't matter where it goes

Correct, although if/when we eventually do distribute it as part of a release, it may be preferable to place it in a consistent location, thus my suggestion to place it under `etc` rather than project root.

vector-of-bool: Ah. There's some confusion. The `SAVE ARTIFACT` command does not place any files in the host, it only declares a file/directory to be exported so that they can be copied onto the host (with `earthly --artifact`) or into another target. The `+sbom-generate` target _does_ write a file into the host with the `AS LOCAL` clause. (Refer: https://docs.earthly.dev/docs/earthfile#save-artifact).

eramongodb: Ah, I see. Then there should be nothing to do for this PR. My suggestion should be handled by followup work. SGTM. 👍

PR #1630, opened 2024-06-06
CDRIVER-4618: updated transaction examples

Julia-Garland:
In with_transaction_example:
- Removed explicit setting of read concern options
- Removed explicit setting of read preferences
- Noted that Atlas connection strings include w=majority by default
- Removed timeout (of 1000ms) from write concern

Made same Atlas default note on line 3642


kkloberdanz (approved):
Looks good to me!

PR #1634, opened 2024-06-07
CDRIVER-4697: More efficient trace toggle

Julia-Garland:
Updated:
`if (MONGOC_TRACE_ENABLED)`
to
`if (MONGOC_TRACE_ENABLED && gLogTrace)`
via [_mongoc_log_trace_is_enabled](https://github.com/vector-of-bool/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/mongoc-log.c#L64) which has the same logic.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-trace-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-trace-private.h
@@ -33,41 +33,41 @@ BSON_BEGIN_DECLS
 
 #define TRACE(msg, ...)                                                                                          \
    do {                                                                                                          \
-      if (MONGOC_TRACE_ENABLED && gLogTrace) {                                                                                \
+      if (MONGOC_TRACE_ENABLED && gLogTrace) {                                                                   \
          mongoc_log (                                                                                            \
             MONGOC_LOG_LEVEL_TRACE, MONGOC_LOG_DOMAIN, "TRACE: %s():%d " msg, BSON_FUNC, __LINE__, __VA_ARGS__); \
       }                                                                                                          \
    } while (0)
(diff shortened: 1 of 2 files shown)


Julia-Garland:
As an additional note, I used the function call because gLogTrace is currently not included in mongoc-log-private.h

src/libmongoc/src/mongoc/mongoc-trace-private.h
```diff
@@ -33,41 +33,41 @@ BSON_BEGIN_DECLS
```
kevinAlbs, line 33: > Mild concern that moving the MONGOC_TRACE_ENABLED check into the function may impede the compiler's ability to deduce always-false conditions at usage sites.

Given that `ENTRY` and `EXIT` are called in many functions, I would rather err towards avoiding the extra runtime cost.

> this may be premature optimization

Agreed, but I expect it is a low complexity optimization. I suggest: remove `static` from `gLogTrace` in mongoc-log.c to enable external linkage and use as follows:

```c
// `gLogTrace` determines if tracing is enabled at runtime.
extern bool gLogTrace;

#define TRACE(msg, ...)                                                                                          \
   do {                                                                                                          \
      if (MONGOC_TRACE_ENABLED && gLogTrace) {                                                                   \
         mongoc_log (                                                                                            \
            MONGOC_LOG_LEVEL_TRACE, MONGOC_LOG_DOMAIN, "TRACE: %s():%d " msg, BSON_FUNC, __LINE__, __VA_ARGS__); \
      }                                                                                                          \
   } while (0)
```

kevinAlbs (approved):

PR #1637, opened 2024-06-07
[CDRIVER-5537] A Signed Release Archive + Augmented SBOM Publication

vector-of-bool:
# Summary 

This changeset re-introduces a release archive process as part of the release process. This is significantly simplified from the old `makedist` archive, and is mostly just an archive produced by Git with some additional goodies, with more to be added in later changes.

## What's Included

- New Earthly targets are added to automate aspects of the release process, including generation of the source archive and signing thereof. Currently, this creates an unfortunate split of the release tools between the c-driver-tools repo and this repository, but it is anticipated that future changes will progressively migrate additional components from the c-driver-tools repo back into the main source repository.
- More documentation has been added on using Earthly and the Earthfile targets, hopefully to accelerate future onboarding and troubleshooting for those that don't work with Earthly on a daily basis. Using separate documentation from the inline comments in the `Earthfile` allows for far more comprehensive descriptions of what is possible.
- A new semi-automated step has been introduced in the release process that utilizes the new Earthly targets to attach additional artifacts to the GitHub release for the C driver. This includes the new release archive, as well as a detached PGP signature, satisfying the requirements of CDRIVER-5537.
- In addition to the source repository, the release archive also includes the augmented SBOM, which is downloaded on-the fly as part of the archive generation. This will satisfy the requirements of CDRIVER-5535.

## Other Notes

- This PR builds upon the #1629 changes, so its commits are included in the commit history here, but can be ignored.
- The choice of Earthly was primarily made to just extend "what we already have" and to ensure isolation and uniformity of execution across different systems, but it was found that Earthly's secret-handling functionality is extremely convenient for these processes!
- No new EVG tasks are defined.

### DIFF
--- a/docs/dev/earthly.rst
+++ b/docs/dev/earthly.rst
@@ -227,7 +227,7 @@ enumerated using ``earthly ls`` or ``earthly doc`` in the root of the repository
 .. _earthly.secrets:
 
 Setting Earthly Secrets
-=======================
+***********************
 
 Some of the above targets require defining `earthly secrets`_\
 [#creds]_.
--- a/docs/dev/earthly.rst
+++ b/docs/dev/earthly.rst
@@ -96,7 +96,8 @@ enumerated using ``earthly ls`` or ``earthly doc`` in the root of the repository
 
    .. option:: --version <version>
 
-      Forwarded to `+release-archive --version`. Only affects output filenames.
+      Forwarded to `+release-archive --version` and used to name the artifact
+      files in `+signed-release/dist/`.
 
    .. rubric:: Secrets
 
--- a/docs/dev/earthly.rst
+++ b/docs/dev/earthly.rst
@@ -1,0 +1,263 @@
+Earthly
+#######
+
+Earthly_ is a CI and development tool that containerizes aspects of the CI
+pipeline so that they run consistently across hosts and across time.
+
+.. highlight:: console
+
+.. _earthly: https://earthly.dev
+.. _earthly secrets: https://docs.earthly.dev/docs/guides/secrets
+.. _docker: https://www.docker.com/
+.. _podman: https://podman.io/
+
+Running Earthly
+***************
+
+.. note::
+
+   Before you can run Earthly, you will need either Podman_ or Docker_ installed
+   on your system. If you have trouble getting Earthly to work with Podman,
+   refer to `the Earthly Podman Guide`__.
+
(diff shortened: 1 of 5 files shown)


docs/dev/earthly.rst
```diff
@@ -0,0 +1,264 @@
...
+
+   .. earthly-artifact:: +sign-file/signature.asc
+
+      The detached PGP signature for the input file.
+
+   .. rubric:: Parameters
+   .. option:: --file <filepath>
+
+      **Required**. Specify a path to a file (on the host) to be signed. This
+      file must be a descendent of the directory that contains the ``Earthfile``
```
rcsanchez97, line 186:
```suggestion
      file must be a descendant of the directory that contains the ``Earthfile``
```

rcsanchez97 (approved):
LGTM, w/ some small tweaks.

docs/dev/releasing.rst
```diff
@@ -244,6 +246,68 @@ Update the ``VERSION_CURRENT`` file on the release branch::
...
+
+1. :ref:`Authenticate with Artifactory <earthly.artifactory-auth>`
+2. Set the Earthly secrets required for the :any:`+sign-file` and
+   :any:`+sbom-download` targets.
+
+Once these prerequesites are met, creating the release archive can be done using
+the :any:`+signed-release` target. Let `$BRANCH` be the name of the Git branch
+from which the release is being made::
+
+   $ ./tools/earthly.sh --artifact +signed-release/dist dist --branch=$BRANCH --version=$NEW_VERSION
```
kevinAlbs, line 281: This may result in a tarball on a commit after the tag. The `post_release_bump` step before adds a commit to update `VERSION_CURRENT`.

Consider having the `+signed-release` target also do a `git checkout` to the `NEW_VERSION` to ensure the tarball is built with the correct `VERSION_CURRENT` file.

kevinAlbs (changes requested):
I like the approach to sign a (simpler) archive rather than signing a tag.

Left suggestion to ensure the tarball has the correct `VERSION_CURRENT` file.

docs/dev/releasing.rst
vector-of-bool: I've modified the workflow to archive the tagged version rather than the branch revision, and renamed some of the parameters as appropriate.

kevinAlbs (approved):

PR #1638, opened 2024-06-12
[CDRIVER-5535] SBOM Process Documentation

vector-of-bool:
This is a minor follow-up to #1637 that details how to update the SBOM-lite for the release process.


kevinAlbs (approved):
The documentation is appreciated (and I expect may be a helpful reference outside of the C driver).

PR #1645, opened 2024-06-14
[CDRIVER-5608] Clean up of estimated_document_count

vector-of-bool:
Control flow of estimated_document_count is needlessly complex when we have C99 available to mix declarations. This fixes a false-positive in static analysis on reply_local being destroyed while uninitialized.

**Note:** The `estimated_document_count` contained a call to `mongoc_cluster_stream_for_reads`, but the return value was simply discarded. There was no commentary on why this call occurs, and it isn't clear why it was there. Since it appears unused, it has been deleted. If it _should_ remain present, it should be given a comment to explain this "load bearing quirk". Do any reviewers have insight on the purpose of this call?

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-collection.c
+++ b/src/libmongoc/src/mongoc/mongoc-collection.c
@@ -17,6 +17,7 @@
 
 #include <stdio.h>
 
+#include "bson/bson.h"
 #include "mongoc-aggregate-private.h"
 #include "mongoc-bulk-operation.h"
 #include "mongoc-bulk-operation-private.h"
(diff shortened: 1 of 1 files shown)


eramongodb:
> Do any reviewers have insight on the purpose of this call?

The function initially [contained a wire version check](https://github.com/mongodb/mongo-c-driver/commit/aa35241409a10bcbb1b6990af6cde3c2b80203e6#diff-93fdba52325a171278699257557e1b1c1a0192b796aeba8b2aeb72eaf3ea5039R882) (CDRIVER-3905) which was [subsequently removed](https://github.com/mongodb/mongo-c-driver/commit/99e9893286767737d2e0cb5d55c7906c3f768656#diff-93fdba52325a171278699257557e1b1c1a0192b796aeba8b2aeb72eaf3ea5039L881) (CDRIVER-4309) following some [Drivers specification shenanigans](https://github.com/mongodb/specifications/blob/2f39993b453bf3b92610fd19e0ea5af8b829cb78/source/crud/crud.md#estimateddocumentcount) (DRIVERS-2228):

> The 5.0-compat versions of many drivers updated their `estimatedDocumentCount` implementations to use the `$collStats` aggregation stage instead of the `count` command. This had the unintended consequence of breaking `estimatedDocumentCount` on views, and so the change was seen as a backwards-incompatible regression and reverted.

The `server_stream` variable appears to have been left behind during that cleanup and is no longer required.

eramongodb (approved):

kevinAlbs (approved):

PR #1649, opened 2024-06-21
[CDRIVER-5535] Add a vulnerability report artifact to the release archive

vector-of-bool:
At long last, this final changeset should complete the work for CDRIVER-5535. It builds upon the previous PRs that introduced additional processes, scripts, and documentation, and defines the final 3rd party vulnerabilities report that is included as a release artifact.

# Summary

The process defined here expands upon the approaches from those taken in libmongocrypt and the C++ driver. It more explicitly defines the content of `third_party_vulnerabilities.md` and what and when items should be added or removed.

# Considerations

## Why not use Snyk `monitor` and rely on the augmented SBOM?

`snyk monitor` takes a snapshot of the working directory and posts it to Snyk, where Snyk effectively runs `snyk test` repeatedly on a schedule, with the test results posted to the Snyk target/project that corresponds to the snapshot. This allows for notifications when new vulnerabilities are issued on past snapshots of the repository.

The overall intended design was that Snyk would detect vulnerabilities, Silk would collect them together, and then Silk would filter the dependencies based on the content of the SBOM lite (i.e. to remove false-positives). The resulting filtered vulnerabilities from Snyk would be inserted into the augmented SBOM. The augmented SBOM would then be the primary release artifact for vulnerability reporting.

For creation of the vulnerability report, the following points were considered when deciding _not_ to rely on the augmented SBOM:

1. A Silk *asset group* must be created and correlated with a `snyk monitor` snapshot, and must correspond to a branch in the upstream repository. The association between a Snyk monitor and an asset group requires that snapshots be specifically tied to *branches* rather than release versions. We want to snapshot a release at the exact point it was taken, and not rely on the Snyk target reference name coinciding with an asset group branch name. We want to monitor past releases, and not just the tip of the release branch.
1. The augmented SBOM is updated asynchronously. It is unknown how soon after creating the Silk asset group and the Snyk snapshot the augmented SBOM will actually contain data (it can take up to 24 hours). In contrast, the vulnerability data from Snyk is immediately live and valid the moment it is requested.
1. This repository contains *four dependencies*, only _one_ of which is actually tracked in the Snyk database (Zlib). This makes the prospective value-add of Silk's vulnerability collection and filtering very very small.
1. Along with the previous point: At present, there is only one vulnerability currently found against the Snyk database. Manually reviewing the vulnerability report for false-positives is a trivial task.

## Does this make the SBOM pointless?

No. The SBOM still gives us structured data to represent our dependencies, including those that are not known to Snyk.

## Does this make the augmented SBOM pointless?

Maybe?

If `snyk monitor` is not executed to create a snapshot for the repository, then Silk will never accrue vulnerabilities to be added to the augmented SBOM.

IMO, we can do better with a semi-manual process that is enhanced by Snyk's automation.

## What about false-positives from Snyk?

When Snyk reports vulnerabilities and/or dependencies in the repository that are/would not actually be present in the release, it is plainly visible that something is amiss. We only have one (true positive) vulnerability at the moment, so this appears to be a non-issue. The problem can be revisited in the future if false positives become a common occurrence.

## What about false-negatives from Snyk?

If there are actual dependencies and/or vulnerabilities that are missed by Snyk, then they would never have appeared in the augmented SBOM anyway. Currently, `jsonsl`, `uthash`, and `utf8proc` are not detected by Snyk, despite being present in the repository. It appears that these libraries are not tracked by Snyk in their package database, so adding them is beyond the reach of just fiddling with the `snyk` CLI.

The correction of false-negatives has been ruled out as out-of-scope for the overall project.

## What about the MUST/SHOULD requirements from the linked ticket?

The following emphasized clause conflicts with these changes:

> Drivers with bundled dependencies MUST integrate with a supported tool (e.g. Snyk) that can perform vulnerability scanning and feed results into Silk for SBOM generation. If Snyk is used, **drivers SHOULD NOT rely on it to infer dependencies**, as it is prone to false-positives and version inaccuracies.

Over the past few weeks, issues have arisen as the implementation of Silk+Snyk was explored within the C driver that would make the process overly cumbersome and error-prone. Following discussion with various teams that are deploying these tools, the processes that are outlined in this changeset are intended to meet _the intent_ of the above requirements without necessarily matching _the wording_. Manually removing false-positives has been determined to be trivial for this repository.

## What About Automatic Jira Reporting via Silk?

One of the stated goals of using Silk is to support the automatic creation of Jira issues corresponding to vulnerabilities that appear in monitored dependencies. This integration is not currently working, and our dependency set is currently so small that we can get away with doing it all "by hand" for now.


tools/snyk-vulns.py
eramongodb, line 1: This script seems very excessive for fulfilling 3PD reporting requirements. Recommend its removal from this PR.

The `third_party_vulnerabilities.md` file as was initially proposed in https://github.com/mongodb/mongo-cxx-driver/pull/1143 is primarily meant to serve the purpose of minimally fulfilling the following requirements:

> If we choose not to ever fix a certain Critical/High/Medium vulnerability, we will disclose the reasons for doing so, as well as the scenarios in which a customer can be impacted by this vulnerability.

And:

> We will produce report(s) containing a list of direct third party component dependencies (and optionally their transitive dependencies) included in the release. This list will include (a) their version, (b) their license, (c) any detected CVEs not yet fixed and the date we discovered them[, and] (d) scanning tool used. Upon request, we will provide third party OSS CVEs fixed by the release and the date we discovered them. Information on any CVEs will include both the original CVSS score as well as any modifications to that score in the context of the MongoDB software.

The additional details recorded by `third_party_dependencies.md` entries (upstream status, fix status, for release, and additional notes) require _manual_ verification and updates as issues are remediated by the upstream dependencies and their corresponding fix status is updated by maintainers. If these details were captured by the Augmented SBOM file (which is also a release artifact) from which this file is being generated, there would not be a need for the separate 3PD Vulnerability report document: the Augmented SBOM would already fulfill the reporting requirements by itself (in a harder-to-read format, but nevertheless). This is especially important for correctly tracking the start and end of the remediation timeline within which vulnerabilities are expected to be addressed.

Furthermore, we currently do not have a clear process for applying remediation updates to either Silk or Snyk (IIRC _eventually_ via VULN JIRA tickets?) by which the vulnerability entries in the Augmented SBOM may be updated either, and there is as-of-yet no guarantees that I am aware of regarding the stability of the contents of the `"vulnerabilities"` field entries or the manner in which they are mapped to information reported by Snyk (even the guarantee of the presence of the `vulnerabilities` field itself was only determined as of about a week ago). I do not believe the value currently provided by this automation script outweights the inevitable maintenance cost that will arise from the ongoing development of SSDLC-related tooling and integrations.

eramongodb (changes requested):

tools/snyk-vulns.py
vector-of-bool: How would you recommend completing the vulnerability tracking and reporting requirements of CDRIVER-5535?

If the answer is to just rely on the augmented SBOM, I feel that there is equivalent (or greater) complexity involved (compared to the changes proposed here) while also being less reliable due to the asynchrony and additional levels of (not in-our-control) indirection involved. In testing, I was unable to get the augmented SBOM to even show any reported vulnerabilities that I knew should be present.

If the answer is that this should be updated/replaced with manual documentation: I would agree with that option. I was pursuing that as an extension to this changeset but decided to put this in review as-is so that I can get eyes on it before next week when I hope to perform the first SSDLC-ready release.

eramongodb: > this should be updated/replaced with manual documentation

Yes, this is my recommendation. Any/all known vulnerabilities (e.g. as reported by Snyk) should be _manually_ documented in `third_party_vulnerability.md`. See [libmongocrypt](https://github.com/mongodb/libmongocrypt/blob/713f8f774116e9382a65a0ddd18afd04530de5dd/etc/third_party_vulnerabilities.md#cve-2023-0437) as an example. The CXX Driver [doesn't have any issues to report atm](https://github.com/mongodb/mongo-cxx-driver/blob/master/etc/third_party_vulnerabilities.md). Note that both of those files are entirely hand-written.

Although Snyk (and the Augmented SBOM by proxy) provides some information such as "date detected" and the "upstream fix available" status (aka "Remediation Path" availability), we must still manually investigate and document values such as the date of upstream fix availability (that is, the date of the first official upstream release which provides the fix), fix status (for our product), any notes or rationale for the current fix status (e.g. when unable to fix by the remediation deadline), and the release version for which our latest fix status update is applicable (e.g. last decision made was for release 1.2.3; fix was applied to release 4.5.6, etc.).

> just rely on the augmented SBOM

As I understand, this appears to be the long-term ideal end goal that SSDLC tooling and integrations are aiming to achieve (e.g. aforementioned VULN JIRA tickets). But this is currently very much not the case right now.

> unable to get the augmented SBOM to even show any reported vulnerabilities that I knew should be present

Yes, the tooling and integration is still immature. This is why I proposed the `third_party_vulnerability.md` document, which is how we can minimally satisfy the reporting requirements of both CDRIVER-5558 (the actual vulnerability report) and CDRIVER-5535 (tooling integration and SBOM generation) despite current limitations. The ongoing issues with Silk <-> Snyk tracking and reporting of vulnerabilities are not our responsibility to resolve (outside the scope of both CDRIVER-5558 and CDRIVER-5535).

vector-of-bool:
Note: These changes are paused pending resolution of some ambiguity in the third-party vuln reporting requirements.

vector-of-bool:
One of the checklist items for conformance is:

> Vulnerabilities for third party libraries are monitored on all non-EOL versions.

I have added additional process and automation to address this goal. This is *separate* (but related) to the requirement for the vulnerability report itself.

vector-of-bool:
I have added and removed content significantly since the initial changeset. The new changes are based on ongoing conversations with team members about the stated goals of vulnerability tracking and what options we actually have available to us.

docs/dev/deps.rst
```diff
@@ -153,32 +152,158 @@ detect the other dependencies within the project. The `+snyk-test` Earthly
...
+   target within the **dev-prod** organization. Do not use the *GitHub target*:
+   That one is not currently useful to us.)
+
+   __ https://app.snyk.io/org/dev-prod/projects?searchQuery=mongo-c-driver&filters[Integrations]=cli
+3. Expand the **mongodb/mongo-c-driver** target, and then expand all **currently
+   supported release versions**. (If you are preparing for a new release, that
+   version should also be available and used after the `releasing.snyk` process
+   has been completed.)
+4. Take note of *all unique vulnerabilities amongst all supported versions'*
+   that are listed in Snyk. These will be the *relevant* vulnerabilities.
```
eramongodb, line 197: May want to add a note prior to this step to [ignore/deactivate/remove](https://github.com/mongodb/mongo-cxx-driver/blob/dc6da28357136028fc62fb15bec35fde92601e3b/etc/releasing.md#check-and-update-snyk) any unrelated projects Snyk may have detected on its own.

etc/third_party_vulnerabilities.md
```diff
@@ -0,0 +1,54 @@
...
+> documentation generators, and other miscellaneous files and artifacts are NOT
+> considered part of the standard release product, even if they are included in
+> the release distribution tarball. Vulnerabilities for such 3rd party
+> dependencies are NOT reported in this document.
+>
+> Details on packages that are not tracked tracked by Snyk Security will not
+> appear in this document.
+
+# Packages
+
```
eramongodb, line 34:
```suggestion
```
Redundant level-1 section header.

eramongodb (approved):

docs/dev/deps.rst
vector-of-bool: I haven't seen this occur in this repo? Running with `--unmanaged` seems to set the scanning mode that we want.

eramongodb: Snyk seems to auto-detect any "package"-like files it finds in the repository. If this doesn't affect the C Driver, I think it is fine to leave it to the other Snyk examination steps to handle if it does become an issue.

vector-of-bool: I think it is a non-issue here because we scan the only the `src/` directory (otherwise it fails to pick up Zlib), so those package-like files are not visible.

PR #1653, opened 2024-06-25
[CDRIVER-5540] Introduce an `ssdlc_compliance_report.md` to the release archive

vector-of-bool:
Refer: CDRIVER-5540. This simple change adds the `ssdlc_compliance_report.md` file to the repository and to the release archive. It is stored in `etc/ssdlc.md` for editing, but is copied as a top-level file `ssdlc_compliance_report.md` within the generate release archive.

The file itself is based on the similar file present in the C++ driver and libmongocrypt repositories.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -180,9 +180,17 @@ release-archive:
     ARG --required prefix
     ARG --required ref
 
+    WORKDIR /s
+    COPY --dir .git .
+
+    # Get the commit hash that we are archiving. Use ^{commit} to "dereference" tag objects
+    LET revision = $(git rev-parse "$ref^{commit}")
+    RUN git restore --quiet --source=$revision -- VERSION_CURRENT
+    LET version = $(cat VERSION_CURRENT)
+
     # Pick the waterfall project based on the tag
     COPY tools+tools-dir/__str /usr/local/bin/__str
-    IF __str test "$ref" -matches ".*\.0"
+    IF __str test "$version" -matches ".*\.0\$"
         # This is a minor release. Link to the build on the main project.
         LET base = "mongo_c_driver"
     ELSE
@@ -190,14 +198,8 @@ release-archive:
         LET base = "mongo_c_driver_latest_release"
     END
 
-    WORKDIR /s
-    COPY --dir .git .
     COPY (+sbom-download/augmented-sbom.json --branch=$sbom_branch) cyclonedx.sbom.json
 
-    # Get the commit hash that we are archiving. Use ^{commit} to "dereference" tag objects
-    LET revision = $(git rev-parse "$ref^{commit}")
-    RUN git restore --quiet --source=$revision -- VERSION_CURRENT
-    LET version=$(cat VERSION_CURRENT)
     # The full link to the build for this commit
     LET waterfall_url = "https://spruce.mongodb.com/version/${base}_${revision}"
     # Insert the URL into the SSDLC report
--- a/Earthfile
+++ b/Earthfile
@@ -179,13 +179,32 @@ release-archive:
     ARG --required sbom_branch
     ARG --required prefix
     ARG --required ref
+
+    # Pick the waterfall project based on the tag
+    COPY tools+tools-dir/__str /usr/local/bin/__str
+    IF __str test "$ref" -matches ".*\.0"
+        # This is a minor release. Link to the build on the main project.
+        LET base = "mongo_c_driver"
+    ELSE
+        # This is (probably) a patch release. Link to the build on the release branch.
+        LET base = "mongo_c_driver_latest_release"
+    END
+
     WORKDIR /s
     COPY --dir .git .
     COPY (+sbom-download/augmented-sbom.json --branch=$sbom_branch) cyclonedx.sbom.json
+
+    # Get the commit hash that we are archiving. Use ^{commit} to "dereference" tag objects
+    LET revision = $(git rev-parse "$ref^{commit}")
+    # The full link to the build for this commit
+    LET waterfall_url = "https://spruce.mongodb.com/version/${base}_${revision}"
+    # Insert the URL into the SSDLC report
     COPY etc/ssdlc.md ssldc_compliance_report.md
+    RUN sed -i "s|@waterfall_url@|$waterfall_url|g" ssldc_compliance_report.md
+    # Generate the archive
     RUN git archive -o release.tar.gz \
         --prefix="$prefix/" \ # Set the archive path prefix
-        "$ref" \ # Add the source tree
+        "$revision" \ # Add the source tree
         --add-file cyclonedx.sbom.json \ # Add the SBOM
         --add-file ssldc_compliance_report.md
     SAVE ARTIFACT release.tar.gz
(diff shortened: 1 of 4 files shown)


etc/ssdlc.md
```diff
@@ -0,0 +1,48 @@
...
+The tracking of security information for bundled third-party dependencies is
+performed using Snyk. Refer to the `etc/third_party_vulnerabilities.md` file
+within the repository or the release archive for vulnerability information known
+at the time of a release. Full bundled dependency information is available in
+the `cyclonedx.sbom.json` file within the release archive.
+
+## Static Analysis Findings
+
+Refer to the
+[SSDLC Static Analysis Reports](https://drive.google.com/drive/folders/17bjBnQ3mhEXvs6IK1rrTphJr0CUO2qZh)
```
eramongodb, line 30: May want to move this into the existing [C Driver](https://drive.google.com/drive/folders/1Q6U43xy8M_JF-jKVfz26uS1oGbAI-fBw) folder, e.g. as done for the [CXX Driver](https://drive.google.com/drive/folders/1Yq9wzMQ4O-7dvIeLp-dik-o9b5A4rzhs).

Earthfile
```diff
@@ -182,10 +182,12 @@ release-archive:
     WORKDIR /s
     COPY --dir .git .
     COPY (+sbom-download/augmented-sbom.json --branch=$sbom_branch) cyclonedx.sbom.json
+    COPY etc/ssdlc.md ssldc_compliance_report.md
```
eramongodb, line 185: Any reason not to simply name `etc/ssdlc.md` and leave it at its committed location? The copy into root directory seems superfluous (the files being in the root directory isn't a requirement AFAIK). Having two copies of the report in a release archive may be somewhat confusing.

eramongodb (approved):

kevinAlbs (approved):

etc/ssdlc.md
```diff
@@ -0,0 +1,48 @@
...
+folder (internal) for release-specific reports. Available as-needed from the
+MongoDB C Driver team.
+
+## Security Testing Report
+
+Refer to the
+[Driver Security Testing Summary](https://docs.google.com/document/d/1y2K_RY4GZVXpQvv4JH_35mSzFRTawNJ3mibpvSBU8H0)
+document (internal). Available as-needed from the MongoDB C Driver team.
+
```
kevinAlbs, line 39:
```suggestion
<!-- TODO: replace with link to Evergreen waterfall build for tagged commit. -->
Refer to the [Evergreen waterfall build](TODO) (internal) for test results.
```

I added the link to Evergreen in the [libmongocrypt Security Testing Summary](https://docs.google.com/document/d/1dc7uvBzu3okAIsA8LSW5sVQGkYIvwpBVdg5v4wb4c4s/edit#heading=h.7i6v0c6gf2w4). Since this links to the common drivers document, consider adding a link to the Evergreen waterfall build here to satisfy the [requirement noted in slack](https://mongodb.slack.com/archives/C06F31ZEKU1/p1718920120640689?thread_ts=1718047795.692869&cid=C06F31ZEKU1).

etc/ssdlc.md
vector-of-bool: The release instructions defer to the `Creating SSDLC static analysis reports` doc, which instructs placing those reports within the `SSDLC Static Analysis Reports` sub-folder (linked here). It may be preferable to inline the content of `Create SSDLC static analysis reports` into `releasing.rst` (it's not a big document) to get it into version control.

The folder structure may need some cleaning-up, because now we have:

```
- DBX Teams > C/C++ Team
  - C Driver/
    - Archive/
  - C++ Driver/
    - Archive/
    - SSDLC Reports/
  - libmongocrypt/
    - Archive/
  - SSDLC/
    - SSDLC Static Analysis Reports/
```

@kevinAlbs opinions?

Aside: I really really really don't like using "just a bunch of files in a shared folder" for things we want to track carefully. This feels like pre-CVS version control and I'm 90% sure this will come back to bite everyone.

Earthfile
vector-of-bool: I'll just rename the file and allow `git archive` to grab it.

etc/ssdlc.md
vector-of-bool: This is annoyingly circular, since the waterfall build URL for the release depends on the commit hash of the project, which depends on the content of this file, which will need to contain a link to the waterfall build for the release, which won't be ready until this file is committed for the release.

Alternatively, this file can be partially templated and have the URL inserted at the last moment by `+release-archive`. IIUC, the waterfall will eventually have a build for the commit of the Git tag?

eramongodb: Linking directly to the build corresponding to the release commit seems like an unreasonable requirement when the compliance report is meant to be a part of the release itself.

imo linking to a dedicated Google Doc as done for libmongocrypt, which may be updated post-release with the inclusion of a link to the mainline commit build, is the best intermediate solution for this requirement.

etc/ssdlc.md
eramongodb: > I really really really don't like using "just a bunch of files in a shared folder" for things we want to track carefully.

Ultimately, all we need to do is ensure the reports are stored and remain accessible on-demand (by customer request, which is the responsibility of the product manager(s) fulfilling that request). We do not need to actively maintain the contents once they've been archived. The actual final directory to which reports must be uploaded [isn't even under the DBX C/C++ path](https://drive.google.com/drive/folders/1ivg27K48Me720nrxXwddis6QNs5SlQTP) anyways, so how we organize the master/draft files is not really a serious concern.

The advice to use existing per-repo folders was to preserve separation of concerns. The parent `SSDLC` folder was created very early in SSDLC policy development stages before the current requirements were fully understood.

etc/ssdlc.md
vector-of-bool: Unfortunately the document linked to for C driver is intended to be the shared document for all drivers, so everyone would be piling their links together in the file all of the time.

I've found a solution that inserts the link as part of the release archive automatically. (See e18170e)

eramongodb: > the shared document for all drivers

I didn't mean the Driver Security Testing Summary (STS) document, I meant the [libmongocrypt-specific STS document](https://docs.google.com/document/d/1dc7uvBzu3okAIsA8LSW5sVQGkYIvwpBVdg5v4wb4c4s/edit#heading=h.7i6v0c6gf2w4) which defers to the Drivers STS document. The libmongocrypt STS document contains a link to the mainline commit build as proposed for its `1.10.1` release.

etc/ssdlc.md
kevinAlbs: `SSDLC Static Analysis Reports` was kept out of the per-project directories to support storing reports for all of C, C++, and libmongocrypt. I am not opposed to changing. Having separate `SSDLC Reports` directories for C and libmongocrypt for the static analysis reports may help organize.

vector-of-bool:
I renewed the review requests following my additional changes.

kevinAlbs (approved):

eramongodb (approved):
Willing to defer organization suggestions to post-deadline. Producing a release with artifacts is more important atm.

PR #1657, opened 2024-06-26
Post-1.23.7 updates

vector-of-bool:
This update pulls some NEWS changes from the 1.27.3 release, as well as some release-process updates to address issues discovered during the release.


eramongodb (approved):

NEWS
```diff
@@ -9,6 +9,27 @@ Deprecated:
...
+
+Notes:
+
+  * This patch release introduces changes surrounding the development processes,
+    including new code signing and dependency tracking. There are no behavioral
+    or API changes.
+
+Fixes:
+
+  * Upgraded bundled Zlib from 1.12.3 to 1.13.1
```
kevinAlbs, line 25:
```suggestion
  * Upgraded bundled Zlib from 1.2.3 to 1.3.1
```

kevinAlbs (approved):
LGTM with a fix to zlib version.

NEWS
vector-of-bool: Danget. That ended up in the release too :(

PR #1661, opened 2024-07-02
[CDRIVER-5604 + CDRIVER-5607] replace `inline` with `BSON_INLINE`

kevinAlbs:
To fix build on VS 2013. Replaces `inline` added in CDRIVER-5604 and CDRIVER-5607.

Patch with a VS 2013 task: https://spruce.mongodb.com/version/6683efce978b630007f6b801

I also added the `release-compile` task on the `windows-2013` variant as part of the [GitHub Patch Definitions](https://spruce.mongodb.com/project/mongo-c-driver/settings/github-commitqueue) to include a VS 2013 task in the Evergreen patches created in GitHub PRs.

### DIFF
--- a/src/libbson/src/bson/bson-oid.c
+++ b/src/libbson/src/bson/bson-oid.c
@@ -76,7 +76,7 @@ BSON_MAYBE_UNUSED static const uint16_t gHexCharPairs[] = {
 #endif
 };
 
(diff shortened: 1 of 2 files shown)


rcsanchez97 (approved):
LGTM

PR #1667, opened 2024-07-08
CDRIVER-5629 do not use `bool` in BSON DSL

kevinAlbs:
# Summary

Replace `bool` in BSON DSL with `boolean`.

Evergreen patch: https://spruce.mongodb.com/version/668c26d644a2bb0007db5a0f

# Background & Motivation

Intended to avoid possible build errors when `bool` is defined as a macro. This [forum post](https://www.mongodb.com/community/forums/t/error-when-build-mongo-c-driver-with-visual-studio-2019/287062/4) suggests the user's build environment defines `bool` (or `_Bool`) as `int`, resulting in a build error due to BSON DSL macros evaluating to an undeclared identifier:

```
mongoc-uri.c:1635:4: error: use of undeclared identifier '_bsonDSL_Type_int'
 1635 |    bsonParse (uri->options,
      |    ^
```

To reproduce the reported issue, a `stdbool.h` was added locally to define `bool` as `int`:

```c
#ifndef __STDBOOL_CUSTOM_H
#define __STDBOOL_CUSTOM_H

#define bool int
#define true 1
#define false 0
#define __bool_true_false_are_defined 1

#endif /* __STDBOOL_CUSTOM_H */
```

In the process of testing, other misuse of `bool` variables in tests were discovered and addressed as drive-by fixes.

### DIFF
--- a/src/common/bson-dsl.h
+++ b/src/common/bson-dsl.h
@@ -242,9 +242,8 @@ BSON_IF_GNU_LIKE (_Pragma ("GCC diagnostic ignored \"-Wshadow\""))
       bsonBuildError = "Error while appending bool(" _bsonDSL_str (b) ")"; \
    } else                                                                  \
       ((void) 0)
-#define _bsonArrayOperation_bool(X) _bsonArrayAppendValue (bool (X))
-#define _bsonValueOperation__Bool(b) _bsonValueOperation_bool (b)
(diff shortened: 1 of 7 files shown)


vector-of-bool (approved):
LGTM. This is a good fix.

PR #1668, opened 2024-07-08
remove extra artifact from `release-archive`

kevinAlbs:
This PR includes two changes to the release process:

- Do not save report in the `dist` directory. The report is already included in the release tarball.
- Fix typo in report name: `ssldc` => `ssdlc`.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -215,7 +215,6 @@ release-archive:
         --add-file cyclonedx.sbom.json \ # Add the SBOM
         --add-file ssldc_compliance_report.md
     SAVE ARTIFACT release.tar.gz
-    SAVE ARTIFACT ssldc_compliance_report.md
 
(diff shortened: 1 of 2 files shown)


vector-of-bool (approved):
LGTM, but do we yet have confirmation on whether the report files need to be submitted to the shared SSDLC folder for every release?

If "yes", then I recommend adding such details to the release instructions, as well as where those files are located within the release archive.

If "no", then this is fine as-is.

kevinAlbs:
> LGTM, but do we yet have confirmation on whether the report files need to be submitted to the shared SSDLC folder for every release?

Not yet. No response yet in [slack thread](https://mongodb.slack.com/archives/C05DMVCT9TN/p1720210798548399?thread_ts=1719222780.699799&cid=C05DMVCT9TN).

PR #1673, opened 2024-07-16
CDRIVER-4656: Driver re-initializes OpenSSL Context on every new socket

Julia-Garland:
### Summary
For OpenSSL versions >= 1.1.0, single-threaded clients and client pools own a single OpenSSL context that is shared among all connections a client makes over TLS.

### The issue
The C driver creates a new OpenSSL context for every new connection that a client makes over TLS. Creating an OpenSSL context is a relatively expensive operation that involves parsing the root certificate store for the operating system and building an internal data structure representing those certificates.

This performance issue was discovered as part of [PERF-4166](https://jira.mongodb.org/browse/DEVPROD-2215). When making 10,000 clients, it was parsing certificates a total of 50,000 times, multiple times for each client. This was enough to make the test essentially not generate any traffic and keep all the cores busy just parsing certificate files repeatedly.

image-2023-05-24-14-56-50-179

### What changed?
_SSL context sharing:_
- In the original behavior, a new `SSL_CTX` was being created locally from a client’s `ssl_opts` in each call to `mongoc_stream_tls_openssl_new()`, even though the SSL configuration is unlikely to change between connections.
- Upon single-threaded client or client pool creation, an `SSL_CTX` is created based on the `ssl_opts` field and is owned by the topology scanner. In the pooled case, any clients popped from the client pool will inherit the pool’s topology and thus its `SSL_CTX`.
- The `SSL_CTX` is used for all connections the client makes over TLS. The context is not modified after it is created, so there is no danger in multiple clients/connections sharing it.

_Updates to_ `ssl_opts`:
- Changing the `ssl_opts` of a single-threaded client or a client pool with  `mongoc_client_set_ssl_opts()` and `mongoc_client_pool_set_ssl_opts()` respectively will result in the `SSL_CTX` owned by that client or pool to be re-initialized to a context matching the new `opts`. 
- The old `SSL_CTX` stays in existence for any connections still using it and is automatically freed when all connections using it have been cleaned up. Note this relies on reference counting (with `SSL_CTX_up_ref()`) which was first introduced in OpenSSL 1.1.0. 

### Running performance benchmarks
A multi-threaded benchmark I used to test the performance improvement of this PR can be found in
I used two benchmarks to test the performance improvement of this PR. Both can be found in `src/libmongoc/tests`.
- `benchmark-tls-pooled` checks out 200 clients from a client pool configured with the driver's built-in TLS files, each of which sends a single ping command. The number of clients can be modified by an optional integer argument. The `maxPoolSize` will be set to the number of clients.

To run, build the C driver with `-DENABLE_SSL=OPENSSL` and set the target as the benchmark's name.

I used macOS's Instruments to do a time profile on the resulting executable. See the video: [Profile libmongoc with Instruments](https://youtu.be/Yh4-g9fNc44).

### Performance improvement

| | Total runtime | Time in `mongoc_openssl_ctx_new()` | % Spent in `mongoc_openssl_ctx_new()` |
|--------|--------|--------|--------|
| Baseline | 28.2s | 11.81s | 41.8% |
| With `ssl_ctx` sharing | 2.9s | 1.0 ms | 0.034% |


Julia-Garland:
For single-threaded clients, the `SSL_CTX` is passed down from the client to the connection stream in the topology scanner node initiator (`_mongoc_topology_scanner_node_setup_stream_for_tls()`). From there, the call stack has two new functions that also take an `SSL_CTX` as a parameter:
- `mongoc_stream_tls_new_with_hostname_and_openssl_context()`
- `mongoc_stream_tls_openssl_new_with_context()`

For pooled clients, the `SSL_CTX` is passed down from the client to the connection stream in the default client stream initiator (`mongoc_client_default_stream_initiator()`). From there, the call stack has three new functions that also take an `SSL_CTX` as a parameter:
- `mongoc_client_connext_with_openssl_context()`
- `mongoc_stream_tls_new_with_hostname_and_openssl_context()`
- `mongoc_stream_tls_openssl_new_with_context()`

src/libmongoc/examples/mongoc-pooled-ping.c
```diff
@@ -0,0 +1,117 @@
...
+   }
+
+   pool = mongoc_client_pool_new (uri);
+   mongoc_client_pool_set_error_api (pool, 2);
+
+   for (i = 0; i < num_clients; i++) {
+      pthread_create (&threads[i], NULL, worker, pool);
+   }
+
+   sleep (30);
```
kevinAlbs, line 102: Rather than waiting 30 seconds, suggest only running one `ping` command in each thread. I expect connections are created only on the first `ping`. That may help simplify the benchmark (no need for a mutex).

kevinAlbs (commented):
Fantastic work. The improvement is impressive.

Including review of the benchmarks and tests. I have not-yet reviewed the rest.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -822,6 +825,97 @@ mongoc_client_connect (bool buffered,
...
+ */
+mongoc_stream_t *
+mongoc_client_connect_with_openssl_context (bool buffered,
+                                            bool use_ssl,
+                                            void *ssl_opts_void,
+                                            SSL_CTX *ssl_ctx,
+                                            const mongoc_uri_t *uri,
+                                            const mongoc_host_list_t *host,
+                                            bson_error_t *error)
+{
```
kevinAlbs, line 856: Suggest adding a `void *openssl_ctx_void` argument to `mongoc_client_connect` and moving the check of `MONGOC_ENABLE_SSL_OPENSSL` within the function body:

```c
mongoc_stream_t *
mongoc_client_connect (bool buffered,
                       bool use_ssl,
                       void *ssl_opts_void,
                       const mongoc_uri_t *uri,
                       void *openssl_ctx_void, // New!
                       bson_error_t *error)
{
    // ... Same as before ...

#if defined(MONGOC_ENABLE_SSL_OPENSSL) && OPENSSL_VERSION_NUMBER >= 0x10100000L
         // Use shared OpenSSL context.
         base_stream =
            mongoc_stream_tls_new_with_hostname_and_openssl_context (base_stream, host->host, ssl_opts, true, (SSL_CTX*) openssl_ctx_void);
#else
         base_stream = mongoc_stream_tls_new_with_hostname (base_stream, host->host, ssl_opts, true);
#endif

    // ... Same as before ...
}
```

This follows a similar pattern to `ssl_opts_void` (which is similarly only defined if SSL is enabled). I expect this may reduce duplicate code.

src/libmongoc/src/mongoc/mongoc-stream-tls-openssl.c
```diff
@@ -853,6 +853,151 @@ mongoc_stream_tls_openssl_new (mongoc_stream_t *base_stream, const char *host, m
...
+ *       NULL on failure, otherwise a mongoc_stream_t.
+ *
+ * Side effects:
+ *       None.
+ *
+ *--------------------------------------------------------------------------
+ */
+
+mongoc_stream_t *
+mongoc_stream_tls_openssl_new_with_context (
```
kevinAlbs, line 880: Suggest having `mongoc_stream_tls_openssl_new_with_context` and `mongoc_stream_tls_openssl_new` pass the `SSL_CTX` to a shared implementation:

```c
// `create_stream_with_openssl_ctx` creates a stream with `ssl_ctx`.
static mongoc_stream_t *
create_stream_with_openssl_ctx (
   mongoc_stream_t *base_stream, const char *host, mongoc_ssl_opt_t *opt, int client, SSL_CTX *ssl_ctx)
{
   // TODO.
}

// `mongoc_stream_tls_openssl_new` creates a stream with a new OpenSSL context.
mongoc_stream_t *
mongoc_stream_tls_openssl_new (mongoc_stream_t *base_stream, const char *host, mongoc_ssl_opt_t *opt, int client)
{
   SSL_CTX *ssl_ctx = _mongoc_openssl_ctx_new (opt);

   if (!ssl_ctx) {
      return NULL;
   }

   return create_stream_with_openssl_ctx (base_stream, host, opt, client, ssl_ctx); // Takes ownership of `ssl_ctx`.
}

#if OPENSSL_VERSION_NUMBER >= 0x10100000L
// `mongoc_stream_tls_openssl_new_with_context` creates a stream from an existing OpenSSL context.
mongoc_stream_t *
mongoc_stream_tls_openssl_new_with_context (
   mongoc_stream_t *base_stream, const char *host, mongoc_ssl_opt_t *opt, int client, SSL_CTX *ssl_ctx)
{
   BSON_ASSERT_PARAM (ssl_ctx);
   SSL_CTX_up_ref (ssl_ctx);                                                        // Increment reference counter.
   return create_stream_with_openssl_ctx (base_stream, host, opt, client, ssl_ctx); // Takes ownership of reference.
}
#endif
```

src/libmongoc/src/mongoc/mongoc-stream-tls-openssl.c
```diff
@@ -853,6 +815,81 @@ mongoc_stream_tls_openssl_new (mongoc_stream_t *base_stream, const char *host, m
...
+ * mongoc_stream_tls_openssl_new --
+ *
+ *       Creates a new mongoc_stream_tls_openssl_t to communicate with a remote
+ *       server using a TLS stream.
+ *
+ *       @base_stream should be a stream that will become owned by the
+ *       resulting tls stream. It will be used for raw I/O.
+ *
+ *       @trust_store_dir should be a path to the SSL cert db to use for
+ *       verifying trust of the remote server.
```
kevinAlbs, line 830:
```suggestion
```

Though not introduced in this PR, suggest removing non-applicable comment. There is no `trust_store_dir` argument.

src/libmongoc/src/mongoc/mongoc-client-private.h
```diff
@@ -30,10 +30,16 @@
...
+
 #ifdef MONGOC_ENABLE_SSL
 #include "mongoc-ssl.h"
 #endif
 
+#if defined(MONGOC_ENABLE_SSL_OPENSSL) && OPENSSL_VERSION_NUMBER >= 0x10100000L
+#include <openssl/ssl.h>
+#include "mongoc-openssl-private.h"
+#endif
+
```
kevinAlbs, line 42:
```suggestion
```

Remove unneeded includes. This header no longer uses OpenSSL symbols.

src/libmongoc/src/mongoc/mongoc-client-private.h
```diff
@@ -30,10 +30,16 @@
 #include "mongoc-read-prefs.h"
 #include "mongoc-rpc-private.h"
 #include "mongoc-opcode.h"
+
```
kevinAlbs, line 33:
```suggestion
```

Suggest removing unneeded whitespace changes to reduce diff. It may add some effort when investigating git history.

src/libmongoc/src/mongoc/mongoc-stream-tls-openssl.c
```diff
@@ -853,6 +815,81 @@ mongoc_stream_tls_openssl_new (mongoc_stream_t *base_stream, const char *host, m
...
+ *--------------------------------------------------------------------------
+ *
+ * mongoc_stream_tls_openssl_new_with_context --
+ *
+ *       Creates a new mongoc_stream_tls_openssl_t to communicate with a remote
+ *       server using a TLS stream, using an existing OpenSSL context.
+ *
+ *       Only called by mongoc_stream_tls_new_with_hostname_and_openssl_context.
+ *
+ *       @ssl_ctx is the global OpenSSL context for the mongoc_client_t
```
kevinAlbs, line 870:
```suggestion
 *       @ssl_ctx is the shared OpenSSL context for the mongoc_client_t
```

Suggest avoiding "global" since the shared OpenSSL context is not stored in a global variable.

kevinAlbs (approved):
LGTM with minor comments. Great work!

src/libmongoc/src/mongoc/mongoc-server-monitor.c
```diff
@@ -855,12 +855,23 @@ _server_monitor_setup_connection (mongoc_server_monitor_t *server_monitor,
...
+      void *openssl_ctx_void = NULL;
 
 #ifdef MONGOC_ENABLE_SSL
       ssl_opts_void = server_monitor->ssl_opts;
 #endif
-      server_monitor->stream = mongoc_client_connect (
-         false, ssl_opts_void != NULL, ssl_opts_void, server_monitor->uri, &server_monitor->description->host, error);
+
+#if defined(MONGOC_ENABLE_SSL_OPENSSL) && OPENSSL_VERSION_NUMBER >= 0x10100000L
+      openssl_ctx_void = (void *) server_monitor->topology->scanner->openssl_ctx;
```
vector-of-bool, line 865:
```suggestion
      openssl_ctx_void = server_monitor->topology->scanner->openssl_ctx;
```

Casting to `void*` is unnecessary.

vector-of-bool (approved):
LGTM

PR #1675, opened 2024-07-19
CDRIVER-5635 Exclude tests and examples from ALL target

eramongodb:
## Summary

Resolves CDRIVER-5635. Verified by [this patch](https://spruce.mongodb.com/version/669a9f647a1e2e0007b0e750).

## The ALL Target and Installation

Although tests and examples are correctly _disabled_ via `ENABLE_TESTS` and `ENABLE_EXAMPLES`, they are not being correctly _excluded_ from the `ALL` target when enabled. The `ALL` target should only build targets which are required by the `install` target (that is, targets which produce artifacts intended to be installed, with the exception of excluded components, which does not apply to our project. This is the same in spirit to the `make && make install` pattern).

Grepping the CMake build output for "Linking" is an approximate means of identifying the list of executable/library targets being built. The base commit currently emits the following list of artifacts for a simple `cmake --build <binary-dir>` (ALL target) configured with tests and examples enabled:

```
C shared library src/libbson/libbson-1.0.so.0.0.0
C static library src/libbson/libbson-static-1.0.a
C static library src/libmongoc/libmongoc-static-1.0.a
CXX static library src/libmongoc-cxx-check.a
C executable src/libmongoc/test-awsauth
C executable src/libmongoc/test-gcpkms
C executable src/libmongoc/test-azurekms
C executable src/libmongoc/test-mongoc-cache
C executable src/libmongoc/test-mongoc-gssapi
C static library src/libmongoc/libtest-libmongoc-lib.a
C shared library src/libmongoc/libmongoc-1.0.so.0.0.0
C executable src/libbson/creating
C executable src/libbson/bson-check-depth
C executable src/libbson/json-to-bson
C executable src/libbson/bson-validate
C executable src/libbson/bson-to-json
C executable src/libbson/bson-streaming-reader
C executable src/libbson/bson-metrics
C executable src/libbson/bcon-speed
C executable src/libbson/bcon-col-view
C executable src/libmongoc/test-atlas-executor
C executable src/libmongoc/test-libmongoc
C executable src/libmongoc/appending
C executable src/libmongoc/executing
C executable src/libmongoc/fam
C executable src/libmongoc/common-operations
C executable src/libmongoc/bulk6
C executable src/libmongoc/bulk5
C executable src/libmongoc/bulk4
C executable src/libmongoc/bulk3
C executable src/libmongoc/bulk2
C executable src/libmongoc/bulk1
C executable src/libmongoc/bulk-collation
C executable src/libmongoc/basic-aggregation
C executable src/libmongoc/aggregation1
C executable src/libmongoc/example-bulkwrite
C executable src/libmongoc/example-collection-command
C executable src/libmongoc/mongoc-tail
C executable src/libmongoc/mongoc-ping
C executable src/libmongoc/mongoc-dump
C executable src/libmongoc/hello_mongoc
C executable src/libmongoc/find-and-modify
C executable src/libmongoc/example-update
C executable src/libmongoc/example-transaction
C executable src/libmongoc/example-session
C executable src/libmongoc/example-sdam-monitoring
C executable src/libmongoc/example-scram
C executable src/libmongoc/example-pool
C executable src/libmongoc/example-gridfs-bucket
C executable src/libmongoc/example-gridfs
C executable src/libmongoc/example-manage-search-indexes
C executable src/libmongoc/example-manage-collection-indexes
C executable src/libmongoc/example-command-with-opts
C executable src/libmongoc/example-command-monitoring
C executable src/libmongoc/example-start-at-optime
C executable src/libmongoc/example-resume
C executable src/libmongoc/example-collection-watch
C executable src/libmongoc/example-client
C executable src/libmongoc/mongoc-stat
```

This PR proposes reducing the list to only the following:

```
Linking C shared library src/libbson/libbson-1.0.so.0.0.0
Linking C static library src/libbson/libbson-static-1.0.a
Linking C static library src/libmongoc/libmongoc-static-1.0.a
Linking C shared library src/libmongoc/libmongoc-1.0.so.0.0.0
Linking C executable src/libmongoc/mongoc-stat
```

Note: man pages and html docs are _not_ excluded, as they are also installed artifacts under `share/`.

The new `tests` and `examples` custom targets are defined as a convenient means of compiling all test and example targets. The `if (NOT TARGET)` pattern is used to ensure compatibility with parent projects (where C Driver is imported via `add_subdirectory()`) which may have already defined the `tests` or `examples` target. The new `tests` target is used by the updated EVG scripts to ensure test binaries are available for use by test tasks.

## Miscellaneous

- Use `CMAKE_BUILD_PARALLEL_LEVEL` instead of `--parallel`, `-- -j`, or `/m`.
- Set `UseMultiToolTask` and `EnforceProcessCountAcrossBuilds` for better MSBuild parallelism with VS2019+.
- Changed `mongoc-cxx-check` from STATIC to OBJECT to avoid the unnecessary "Linking" step.
- Remove unused `EXAMPLES` variable ([incorrect for its entire existence?](https://github.com/mongodb/libbson/commit/8086fcf575c6ebce6f68c8b08673dae160989b1a#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR282)).

### DIFF
--- a/.evergreen/scripts/compile-unix.sh
+++ b/.evergreen/scripts/compile-unix.sh
@@ -215,3 +215,6 @@ CCACHE_NOHASHDIR=1
 
 # For use by test tasks, which directly use the binary directory contents.
 "${cmake_binary}" --build . --target mongo_c_driver_tests
+
+# Also validate examples.
+"${cmake_binary}" --build . --target mongo_c_driver_examples
--- a/.evergreen/scripts/compile-unix.sh
+++ b/.evergreen/scripts/compile-unix.sh
@@ -181,6 +181,8 @@ if [[ "${OSTYPE}" == darwin* ]]; then
     sysctl -n hw.logicalcpu
   }
 fi
+export CMAKE_BUILD_PARALLEL_LEVEL
+CMAKE_BUILD_PARALLEL_LEVEL="$(nproc)"
 
 declare -a extra_configure_flags
 IFS=' ' read -ra extra_configure_flags <<<"${EXTRA_CONFIGURE_FLAGS:-}"
@@ -208,5 +210,10 @@ CCACHE_BASEDIR="$(pwd)"
 CCACHE_NOHASHDIR=1
 
 "${cmake_binary}" "${configure_flags[@]}" ${extra_configure_flags[@]+"${extra_configure_flags[@]}"} .
-"${cmake_binary}" --build . -- -j "$(nproc)"
-"${cmake_binary}" --build . --target install
+"${cmake_binary}" --build .
+"${cmake_binary}" --install .
+
+# For use by test tasks, which directly use the binary directory contents.
+"${cmake_binary}" --build . --target tests || {
+  echo "Ignoring build error which may be caused by ENABLE_TESTING=OFF" 1>&2
+}
--- a/.evergreen/scripts/compile-unix.sh
+++ b/.evergreen/scripts/compile-unix.sh
@@ -214,6 +214,4 @@ CCACHE_NOHASHDIR=1
 "${cmake_binary}" --install .
 
 # For use by test tasks, which directly use the binary directory contents.
-"${cmake_binary}" --build . --target tests || {
-  echo "Ignoring build error which may be caused by ENABLE_TESTING=OFF" 1>&2
-}
+"${cmake_binary}" --build . --target mongo_c_driver_tests
--- a/.evergreen/scripts/compile-windows.sh
+++ b/.evergreen/scripts/compile-windows.sh
@@ -151,4 +151,4 @@ fi
 "${cmake_binary}" --build . --config "${build_config}" --target mongo_c_driver_tests
 
 # Also validate examples.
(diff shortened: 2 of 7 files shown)


kevinAlbs (approved):
LGTM with minor comments.

> The `if (NOT TARGET)` pattern is used to ensure compatibility with parent projects (where C Driver is imported via `add_subdirectory()`) which may have already defined the `tests` or `examples` target.

I would like to eventually consider prefixing target names (e.g. with `mongoc` or similar) to avoid collisions. But I think that is outside of the scope of this PR.

eramongodb:
Good call on prefixing target names. The CXX Driver may not be the only users using `add_subdirectory()` to import the C Driver; they may also define their own `tests` or `examples` target (a common pattern).

Renamed to use the `mongo_c_driver_*` prefix and unconditionally defined them as top-level targets for use by both libbson and libmongoc, removing the need for the `tests` target existence check+workaround in scripts.

kevinAlbs (approved):

PR #1681, opened 2024-07-25
CDRIVER-5645 skip auth tests running against MongoDB 3.6.

kevinAlbs:
# Summary

Skip authentication tests running against 3.6 servers.

Verified by [this patch](https://spruce.mongodb.com/version/66a29f2adf4e6a00076a62fa)

# Background & Motivation

CDRIVER-4815 dropped support for 3.6 servers. Some test servers are running 3.6, resulting in task failures. [Example](https://spruce.mongodb.com/task/mongo_c_driver_gcc48rhel_authentication_tests_openssl_75ec9c31cd974e9bfab0fecd8aa2eca92399543f_24_07_23_19_57_35/logs?execution=0):

```
Ping failure: Server at <REDACTED:auth_host>:27017 reports wire version 6, but this version of libmongoc requires at least 7 (MongoDB 4.0)
```

[DEVPROD-9029](https://jira.mongodb.org/browse/DEVPROD-9029) tracks upgrading test servers. This PR proposes skipping tests in the short-term as they are expected to fail.

### DIFF
--- a/.evergreen/scripts/run-auth-tests.sh
+++ b/.evergreen/scripts/run-auth-tests.sh
@@ -100,12 +100,15 @@ elif command -v otool >/dev/null; then
 fi
 
 if [[ "${ssl}" != "OFF" ]]; then
+  if true; then
+    echo "Skipping tests on server 3.6 (using 'auth_host' or 'auth_gssapi') until DEVPROD-9029 is addressed."
+  else
   # FIXME: CDRIVER-2008 for the cygwin check
   if [[ "${OSTYPE}" != "cygwin" ]]; then
     echo "Authenticating using X.509"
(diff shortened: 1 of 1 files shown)


.evergreen/scripts/run-auth-tests.sh
```diff
@@ -100,12 +100,15 @@ elif command -v otool >/dev/null; then
 fi
 
 if [[ "${ssl}" != "OFF" ]]; then
+  if true; then
```
eramongodb, line 103: Recommend using `{} || {}` instead of `if/else` to better distinguish the conditional blocks from existing code:

```bash
{
  # Skip 3.6. Remove this block when resolving CDRIVER-5645.
  echo "Skipping tests on server 3.6 (using 'auth_host' or 'auth_gssapi') until DEVPROD-9029 is addressed."
} || {
# FIXME: CDRIVER-2008 for the cygwin check
if [[ "${OSTYPE}" != "cygwin" ]]; then
  echo "Authenticating using X.509"
  ...
fi
}
```

eramongodb (approved):
Style suggestion; otherwise, LGTM.

PR #1684, opened 2024-08-05
CDRIVER-5634: SCRAM-SHA-256 FIPS Compliance

Julia-Garland:
### Summary
(Mostly) ensures the driver implementation of SCRAM-SHA-256 (as well as SCRAM-SHA-1) delegates implementation of key derivation is delegated to a cryptographic provider that already is or can be configured to be FIPS-compliant.

### What changed?
Previously the driver implemented the PBKDF2 (Password-based-Key-Derivative-Function) itself in `_mongoc_salt_password()`. 
- When MONGOC_ENABLE_CRYPTO=LIBCRYPTO is enabled, the driver uses `PKCS5_PBKDF2_HMAC()` with `EVP_sha256()` (or `EVP_sha1()`) from the OpenSSL EVP library.
- When MONGOC_ENABLE_CRYPTO=COMMON_CRYPTO is enabled, the driver uses `CCKeyDerivationPBKDF()` with `kCCPRFHmacAlgSHA256` (or `kCCPRFHmacAlgSHA1`) from the CommonCrypto KeyDerivation library.
- When MONGOC_ENABLE_CRYPTO=CNG is enabled, the driver uses `BCryptDeriveKeyPBKDF2()` with a handle to HMAC-SHA-256 from the Bcrypt library. If the symbol is not supported, the driver reverts to the original key derivation function.

### MinGW-w64
BCryptDeriveKeyPBKDF2 was added in MinGW-w64 6.0.0, but some hosts for Evergreen tests use older versions. I filed [CDRIVER-5649](https://jira.mongodb.org/browse/CDRIVER-5649) to “Require MinGW 6.0.0 or higher”, but until then only users with more modern MinGW versions get FIPS compliance on Windows.


src/libmongoc/CMakeLists.txt
```diff
@@ -1345,6 +1345,11 @@ endif ()
 
 include (CMakePackageConfigHelpers)
 
+# Check if BCryptDeriveKeyPBKDF2 is defined in bcrypt.h
+if (WIN32 AND MONGOC_ENABLE_CRYPTO_CNG)
+   check_symbol_exists (BCryptDeriveKeyPBKDF2 bcrypt.h MONGOC_HAVE_BCRYPT_PBKDF2)
```
kevinAlbs, line 1350: `MONGOC_HAVE_BCRYPT_PBKDF2` is defined as a CMake variable, but is not-yet applied to the C code.

Update `mongoc-config.h.in` with [these instructions](https://github.com/mongodb/mongo-c-driver/blob/46ceebacf6b36683aa20f2ba90db1ebd2ac55b8f/src/libmongoc/src/mongoc/mongoc-config.h.in#L27-L35). `mongoc-config.h.in` is configured by CMake and to generate `mongoc-config.h`. CMake variables (between `@`) are substituted.

Julia-Garland: I pushed new updates for this but I think I might have missed something. In line 73 of `mongoc-config.h` (`#if MONGOC_HAVE_BCRYPT_PBKDF2 != 1`) I see the error "operator has no left operand" which I believe means the macro is not defined?

kevinAlbs: Try moving this block above `configure_file` on line 725, which generates the `mongoc-config.h` file:

```c++
configure_file (
   "${PROJECT_SOURCE_DIR}/src/mongoc/mongoc-config.h.in"
   "${PROJECT_BINARY_DIR}/src/mongoc/mongoc-config.h"
)
```

src/libmongoc/src/mongoc/mongoc-crypto-cng.c
```diff
@@ -141,6 +143,101 @@ _mongoc_crypto_cng_hmac_or_hash (
...
+   }
+   bson_free (password_copy);
+   bson_free (salt_copy);
+   return (int) status;
+}
+#endif
+
+/* Compute the SCRAM step Hi() as defined in RFC5802 */
+static void
+mongoc_crypto_cng_derive_key_pbkdf2 (mongoc_crypto_t *crypto,
```
kevinAlbs, line 193:
```suggestion
_bcrypt_derive_key_pbkdf2 (mongoc_crypto_t *crypto,
```

Suggest using the same name as the alternative that uses `BCryptDeriveKeyPBKDF2` and wrap in `#else` / `#endif` since this is only called when the alternative is unavailable.

kevinAlbs (commented):
Looks good overall. Suggested changing returns from `int` to `bool` and a possible fix to check for `BCryptDeriveKeyPBKDF2`.

Nice spot of the duplicate call to `BCryptCloseAlgorithmProvider (&_sha256_hmac_algo, 0);`.

src/libmongoc/CMakeLists.txt
```diff
@@ -532,6 +532,17 @@ else ()
    set (MONGOC_HAVE_SS_FAMILY 1)
 endif ()
 
+# Check if BCryptDeriveKeyPBKDF2 is defined in bcrypt.h
+if (WIN32 AND MONGOC_ENABLE_CRYPTO_CNG)
+   check_symbol_exists (BCryptDeriveKeyPBKDF2 bcrypt.h HAVE_BCRYPT_PBKDF2)
```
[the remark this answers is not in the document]
Julia-Garland: Looks like that resolved the issue!

src/libmongoc/src/mongoc/mongoc-crypto-common-crypto.c
```diff
@@ -20,7 +20,21 @@
...
+mongoc_crypto_common_crypto_pbkdf2_hmac_sha1 (mongoc_crypto_t *crypto,
+                                              const char *password,
+                                              size_t password_len,
+                                              const uint8_t *salt,
+                                              size_t salt_len,
+                                              uint32_t iterations,
+                                              size_t output_len,
+                                              unsigned char *output)
+{
+   return !CCKeyDerivationPBKDF (
```
kevinAlbs, line 35:
```suggestion
   return kCCSuccess == kCCKeyDerivationPBKDF (
```

Suggest checking returned `int` to avoid unchecked cast to bool. The error codes [are defined as negatives](https://github.com/apple-oss-distributions/CommonCrypto/blob/CommonCrypto-60178.40.2/include/CommonCryptoError.h#L57). Add `#include <CommonCrypto/CommonCryptoError.h>` to include definition for `kCCSuccess`.

src/libmongoc/src/mongoc/mongoc-crypto-common-crypto.c
```diff
@@ -46,6 +60,20 @@ mongoc_crypto_common_crypto_sha1 (mongoc_crypto_t *crypto,
...
+mongoc_crypto_common_crypto_pbkdf2_hmac_sha256 (mongoc_crypto_t *crypto,
+                                                const char *password,
+                                                size_t password_len,
+                                                const uint8_t *salt,
+                                                size_t salt_len,
+                                                uint32_t iterations,
+                                                size_t output_len,
+                                                unsigned char *output)
+{
+   return !CCKeyDerivationPBKDF (
```
kevinAlbs, line 73:
```suggestion
   return kCCSuccess == CCKeyDerivationPBKDF (
```

kevinAlbs (approved):
LGTM with another check.

PR #1691, opened 2024-08-08
backport the secutity patch of CVE-2024-6383

Crispy-fried-chicken:
Here is a vulnerability which is mentioned in https://github.com/mongodb/mongo-c-driver/pull/1593 and is fixed in the r1.27 branch in https://github.com/mongodb/mongo-c-driver/commit/7c34461863211be172e6317221d72e4429bed45e, but is not fixed in the branch of r1.26, maybe it should be backported?

### DIFF
--- a/src/libbson/src/bson/bson-string.c
+++ b/src/libbson/src/bson/bson-string.c
@@ -20,6 +20,7 @@
 
 #include <bson/bson-compat.h>
 #include <bson/bson-config.h>
+#include <bson/bson-cmp.h>
 #include <bson/bson-string.h>
 #include <bson/bson-memory.h>
 #include <bson/bson-utf8.h>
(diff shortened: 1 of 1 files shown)


kevinAlbs:
Hello @Crispy-fried-chicken, CDRIVER-5552 is included in the 1.27.1 release. The r1.26 branch is no longer under active development, and no further releases are planned. Is there a reason for requesting a backport to r1.26? I suggest upgrading to 1.27.5.

kevinAlbs:
> due to specific dependencies in our project, upgrading to the 1.27.x series is not feasible at the moment

Can you elaborate? Is 1.27.x incompatible with some of these dependencies?

How is the C driver being obtained? E.g. is it built from source?

Is this PR requesting a 1.26.x release with this fix, or is committing to the r1.26 branch sufficient?

I apologize for the questions, but I would like to better understand the problem and see if there are other solutions.

kevinAlbs:
> I don't know whether 1.27.x is incompatible with some of these dependencies

To confirm: was there an unexpected issue upgrading to 1.27.x? If there was an unexpected issue, that may be better addressed with a fix in r1.27.

In general, I prefer not to set a precedent for cherry-picking commits to unmaintained branches. I expect we may close this PR.

SunBK201:
We also used version 1.26.x, and this patch is very important to us!

catch22out:
Dear Maintenance Team,

Thank you for your ongoing efforts in maintaining the mongo-c-driver. I would like to provide further clarification regarding the request to backport the security patch for CVE-2024-6383 to the 1.26.x branch. Our project currently relies on the 1.26.x version, and due to some critical dependencies, we are unable to upgrade to the 1.27.x series in the near term. While we understand that the 1.26.x branch is no longer under active development, the complexity of our project and the need for system stability make a direct upgrade to 1.27.x potentially problematic, as it could introduce supply chain security issues and destabilize our system. Additionally, I am aware that there are still many users relying on the 1.26.x version, so ensuring the security of this version is crucial for us and for other users.

I fully understand the reasons for avoiding cherry-picking commits into unmaintained branches, but in this specific case, considering the importance of this patch and the broad user base still depending on this version, backporting the fix to the 1.26.x branch would significantly enhance security and stability for existing users without impacting other versions.

kevinAlbs:
Closing. Please upgrade to 1.27.1+ to get the fix of CDRIVER-5552. 1.27 is expected to be backwards compatible with 1.26. If there is an unexpected issue upgrading that suggests a driver bug, please [file a bug report](https://github.com/mongodb/mongo-c-driver?tab=readme-ov-file#bugs--feature-requests).

PR #1693, opened 2024-08-08
CDRIVER-5667 Add missing BSON type checks in special key handlers

eramongodb:
Resolves CDRIVER-5667. Verified by [this patch](https://spruce.mongodb.com/version/66b53769e4b2620007435635).

Traced the issue to a code pattern discrepancy in `_bson_json_read_map_key` when handling special keys in the `BSON_JSON_IN_BSON_TYPE` read state. The BSON type field is reset to `BSON_TYPE_EOD` (value: `0`) when transitioning from `BSON_JSON_IN_START_MAP` -> `BSON_JSON_IN_BSON_TYPE`. This PR identified that when subsequently identifying the special key value on re-entry into this function (such as when parsing the key of a subdocument), only some branches performed a check for a pre-existing non-EOD BSON type. Simplified:

```c
if (bson->read_state == BSON_JSON_IN_BSON_TYPE) {
   // Via HANDLE_OPTION expansion.
   if (len == strlen ("$regex") && strncmp ((const char *) val, ("$regex"), len) == 0) {
      if (bson->bson_type && bson->bson_type != BSON_TYPE_REGEX) { // BSON type check. (!)
         _bson_json_read_set_error (...);
         return;
      }
      bson->bson_type = BSON_TYPE_REGEX;
      bson->bson_state = BSON_JSON_LF_REGEX; // LF: "Looking For"
   }

   // Similar pattern for "$options", "$oid", ..., "$numberDecimal".
   else if ...

   // Pattern break: `strcmp` instead of `strlen` + `strncmp`.
   else if (!strcmp ("$timestamp", (const char *) val)) {
      // Pattern break: no BSON type check. (!!)
      bson->bson_type = BSON_TYPE_TIMESTAMP;
      bson->read_state = BSON_JSON_IN_BSON_TYPE_TIMESTAMP_STARTMAP;
   }

   // Similar pattern for "$regularExpression", "$dbPointer", "$code", and "$scope".
   else if ...

   else {
      _bson_json_bad_key_in_type (reader, val);
   }
}
```

Adding the missing BSON type checks was sufficient to detect and handle the malformed input described in CDRIVER-5667. These checks prevent the state machine from reaching the point where `_bson_json_read_append_code` is invoked with a precondition violation (the top of JSON stack frame is not a previously-parsed scope document) by detecting+forbidding the presence of any intermediate and unrelated BSON types that could overwrite the current JSON stack frame with their own state.

Some drive-by improvements include:

- Use of an X macro for special key strings to better identify and maintain the list of special keys (e.g. this PR removes several redundant `strlen` + `memcmp` calls for duplicate key checks).
- Refactor of the `strlen` + `memcmp` (formerly `strncmp`) pattern into a dedicated macro for consistent reuse.
- Refactor the BSON type check into a macro for consistent reuse.
- Update JSON error test input values with `BSON_STR` to reduce need for escape characters.

### DIFF
--- a/src/libbson/src/bson/bson-json.c
+++ b/src/libbson/src/bson/bson-json.c
@@ -362,20 +362,27 @@ _noop (void)
       return;                                                                                                \
    } else                                                                                                    \
       (void) 0
-#define HANDLE_OPTION(_selection_statement, _key, _type, _state)                                 \
-   _selection_statement (len == strlen (_key) && strncmp ((const char *) val, (_key), len) == 0) \
-   {                                                                                             \
-      if (bson->bson_type && bson->bson_type != (_type)) {                                       \
-         _bson_json_read_set_error (reader,                                                      \
-                                    "Invalid key \"%s\".  Looking for values "                   \
-                                    "for type \"%s\", got \"%s\"",                               \
-                                    (_key),                                                      \
-                                    _bson_json_type_name (bson->bson_type),                      \
-                                    _bson_json_type_name (_type));                               \
-         return;                                                                                 \
-      }                                                                                          \
-      bson->bson_type = (_type);                                                                 \
-      bson->bson_state = (_state);                                                               \
+
+#define HANDLE_OPTION_KEY_COMPARE(_key) (len == strlen (_key) && memcmp (key, (_key), len) == 0)
+
+#define HANDLE_OPTION_TYPE_CHECK(_key, _type)                               \
+   if (bson->bson_type && bson->bson_type != (_type)) {                     \
+      _bson_json_read_set_error (reader,                                    \
+                                 "Invalid key \"%s\".  Looking for values " \
+                                 "for type \"%s\", got \"%s\"",             \
+                                 (_key),                                    \
+                                 _bson_json_type_name (bson->bson_type),    \
+                                 _bson_json_type_name (_type));             \
+      return;                                                               \
+   }                                                                        \
+   ((void) 0)
+
+#define HANDLE_OPTION(_selection_statement, _key, _type, _state) \
+   _selection_statement (HANDLE_OPTION_KEY_COMPARE (_key))       \
+   {                                                             \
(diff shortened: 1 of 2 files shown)


kevinAlbs (approved):
LGTM.

I am happy to learn "LF" (presumably) means Looking For.

src/libbson/src/bson/bson-json.c
```diff
@@ -362,20 +362,27 @@ _noop (void)
...
+   if (bson->bson_type && bson->bson_type != (_type)) {                     \
+      _bson_json_read_set_error (reader,                                    \
+                                 "Invalid key \"%s\".  Looking for values " \
+                                 "for type \"%s\", got \"%s\"",             \
+                                 (_key),                                    \
+                                 _bson_json_type_name (bson->bson_type),    \
+                                 _bson_json_type_name (_type));             \
+      return;                                                               \
+   }                                                                        \
+   ((void) 0)
```
vector-of-bool, line 378: Should there be an `else` before the cast-to-void?

vector-of-bool (approved):
Good fix and clean-up! LGTM

src/libbson/src/bson/bson-json.c
eramongodb: I omitted the `else` because it didn't appear to be necessary. The `if` is scoped with `{}`, so there is (AFAIK) no opportunity for `else` hijacking or other (plausible) mistakes which may produce unexpected behavior at usage sites. A plain `((void)0)` seems sufficient to force a semicolon regardless of the macro definition (`if` or otherwise). If I am incorrect and it is advisable to keep the `else` anyways, a followup PR can patch it in.

vector-of-bool: Since this is so contained, I don't think there's any risk of anything goofy happening. My only thought was about:

```c
if (x)
    HANDLE_OPTION_TYPE_CHECK(...)
else
    lol();
```

But that would just be a hard compile-error anyway.

PR #1694, opened 2024-08-09
CDRIVER-4479 drop VS 2013

kevinAlbs:
# Summary

Resolves CDRIVER-4479. Drop Evergreen tasks for VS 2013. Document dropping support.

# Future work

This PR does not drop workarounds to support VS 2013 (e.g. https://github.com/mongodb/mongo-c-driver/commit/77bbfee6a927af39a8040f18a6a2c37624b5791c and https://github.com/mongodb/mongo-c-driver/pull/1229/commits/61a4a8d80c254791b190aafdcd39ffefe5229d53).

### DIFF
--- a/.evergreen/config_generator/components/sasl/winssl.py
+++ b/.evergreen/config_generator/components/sasl/winssl.py
@@ -15,8 +15,6 @@ TAG = f'sasl-matrix-{SSL}'
 # pylint: disable=line-too-long
 # fmt: off
 COMPILE_MATRIX = [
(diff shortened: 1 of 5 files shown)


eramongodb (approved):
LGTM. The `windows-2013` GitHub patch definition regex also needs to be removed from the EVG project.

kevinAlbs:
> The `windows-2013` GitHub patch definition regex also needs to be removed from the EVG project.

Good catch. Removed from [Patch Definitions](https://spruce.mongodb.com/project/mongo-c-driver/settings/github-commitqueue).

PR #1697, opened 2024-08-12
Reduce Warnings - platform-dependent integer conversions for PBKDF functions

eramongodb:
Followup to https://github.com/mongodb/mongo-c-driver/pull/1684. Verified by [this patch](https://spruce.mongodb.com/version/66ba33965fd3510007b8997d).

Addresses integral type conversion warnings for the new PBKDF2-related functions due to platform- and library-specific crypto library function parameter types.


kevinAlbs (approved):
LGTM. The follow-up fixes are much appreciated.

PR #1702, opened 2024-08-12
CDRIVER-4656 check if server is running with TLS in tests

kevinAlbs:
Followup to https://github.com/mongodb/mongo-c-driver/pull/1673. Fixes failing `retry-true-latest-server` task. [Example failure](https://spruce.mongodb.com/task/mongo_c_driver_gcc75_retry_true_latest_server_f6971f8c49ccca3c840f65233ae514ccf1df8709_24_08_12_18_29_56/logs?execution=0):

```
FAIL:tests/test-mongoc-client.c:2240  test_mongoc_client_change_openssl_ctx_before_ops()
  ret
  No suitable servers found: `serverSelectionTimeoutMS` expired: [TLS handshake failed: Connection timed out calling hello on 'localhost:27017']. Topology type: Single
```

Tests require the server to be running with TLS enabled. The `retry-true-latest-server` uses a C driver built with OpenSSL, but does not run against a server with TLS enabled.

Verified with [this patch](https://spruce.mongodb.com/version/66ba6e28866ccc0007091cd2/).

### DIFF
--- a/src/libmongoc/tests/test-mongoc-client.c
+++ b/src/libmongoc/tests/test-mongoc-client.c
@@ -2222,14 +2222,9 @@ test_client_buildinfo_hang (void)
 #if defined(MONGOC_ENABLE_SSL_OPENSSL)
 
 static void
-test_mongoc_client_change_openssl_ctx_before_ops (void)
+test_mongoc_client_change_openssl_ctx_before_ops (void *unused)
 {
-   if (!test_framework_get_ssl ()) {
-      MONGOC_DEBUG ("Skipping test. Test requires server to be running with TLS enabled, but MONGOC_SSL_* environment "
-                    "variables are not set.");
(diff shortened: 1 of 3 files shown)


src/libmongoc/tests/test-mongoc-client.c
```diff
@@ -2224,6 +2224,12 @@ test_client_buildinfo_hang (void)
 static void
 test_mongoc_client_change_openssl_ctx_before_ops (void)
 {
+   if (!test_framework_get_ssl ()) {
```
vector-of-bool, line 2227: Add `@ctest-skipped@@` somewhere in message, which CTest will scan for and consider the test skipped if it sees it.

vector-of-bool (approved):
LGTM, minor note

PR #1703, opened 2024-08-13
CDRIVER-5672 Fix C23 compile; don't return bool when return type is a pointer

adriandole:
C23 removed the conversion between boolean and pointers. But this is just a typo and should've been NULL originally.


kevinAlbs (approved):
Neat. Nice find. Filed CDRIVER-5672 since this is a possible build fix. I think it is worth cherry-picking to the r1.27 branch to include in the 1.27.6 release.

PR #1705, opened 2024-08-22
CDRIVER-5676 implement bson_validate libfuzzer entrypoint

spencerjackson:
This patch implements a fuzzer for bson_validate. It works be performing the following steps:
1. Implement a new entrypoint
2. Adding the entrypoint to CMakeLists
3. Adding '-fno-sanitize-recover=all' to sanitizer flags, so that UBSan faults become unrecoverable.

### DIFF
--- a/src/libbson/fuzz/fuzz_test_validate.c
+++ b/src/libbson/fuzz/fuzz_test_validate.c
@@ -9,7 +9,7 @@ LLVMFuzzerTestOneInput (const uint8_t *data, size_t size)
    bson_t b;
    if (bson_init_static (&b, data, size)) {
       bson_validate (&b,
-                     BSON_VALIDATE_NONE | BSON_VALIDATE_UTF8 | BSON_VALIDATE_DOLLAR_KEYS | BSON_VALIDATE_DOT_KEYS |
+                     BSON_VALIDATE_UTF8 | BSON_VALIDATE_DOLLAR_KEYS | BSON_VALIDATE_DOT_KEYS |
                         BSON_VALIDATE_UTF8_ALLOW_NULL | BSON_VALIDATE_EMPTY_KEYS,
                      NULL);
    }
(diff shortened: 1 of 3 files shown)


src/libbson/fuzz/fuzz_test_validate.c
```diff
@@ -0,0 +1,18 @@
...
+#include <stdint.h>
+#include <bson/bson.h>
+
+int
+LLVMFuzzerTestOneInput (const uint8_t *data, size_t size)
+{
+   bson_t b;
+   if (bson_init_static (&b, data, size)) {
+      bson_validate (&b,
+                     BSON_VALIDATE_NONE | BSON_VALIDATE_UTF8 | BSON_VALIDATE_DOLLAR_KEYS | BSON_VALIDATE_DOT_KEYS |
```
kevinAlbs, line 12:
```suggestion
                     BSON_VALIDATE_UTF8 | BSON_VALIDATE_DOLLAR_KEYS | BSON_VALIDATE_DOT_KEYS |
```

Suggest removing `BSON_VALIDATE_NONE` (evaluates to 0).

kevinAlbs (approved):
Thank you for the contribution. LGTM with removal of an unnecessary flag.

src/libbson/fuzz/fuzz_test_validate.c
```diff
@@ -0,0 +1,18 @@
...
+{
+   bson_t b;
+   if (bson_init_static (&b, data, size)) {
+      bson_validate (&b,
+                     BSON_VALIDATE_UTF8 | BSON_VALIDATE_DOLLAR_KEYS | BSON_VALIDATE_DOT_KEYS |
+                        BSON_VALIDATE_UTF8_ALLOW_NULL | BSON_VALIDATE_EMPTY_KEYS,
+                     NULL);
+   }
+
+   return 0;
```
eramongodb, line 17:
```suggestion
      return 0;
   }
   return -1;
```
Do not add inputs which fail to initialize the bson_t object to the corpus. We are only interested in inputs which lead to `bson_validate()` being invoked.

eramongodb (approved):

PR #1708, opened 2024-08-26
Add fuzz_test_properties for common target options for fuzz executables

eramongodb:
Followup to https://github.com/mongodb/mongo-c-driver/pull/1705.

A small refactor that extracts common target properties of fuzz test executables into an interface library for reusability as more fuzz test executables are expected to be introduced.


kevinAlbs (approved):

spencerjackson (approved):
Oh, neat!

PR #1710, opened 2024-08-27
Miscellaneous EVG compilation script improvements

eramongodb:
Followup to https://github.com/mongodb/mongo-c-driver/pull/1524. Verified by [this patch](https://spruce.mongodb.com/version/66ce1342e0dea500075ee87d).

Addresses the following error on some Windows distros when used with the old `CCache.cmake` module (discovered while testing the C++ Driver; this does not break builds but does impact ccache efficiency):

```
ccache: error: CCACHE_BASEDIR: not an absolute path: "/cygdrive/c/data/mci/.../mongo-cxx-driver"
```

Applies a proper transformation to Windows path form on Windows distros to avoid Windows-side cmake or ccache confusion of file paths.

Additional improvements include:

- removing a pre-1.11.0 libmongocrypt workaround now that 1.11.0 is released.
- simplifying DEBUG vs. RELEASE variable checks in compile scripts (mutually exclusive options). We should consider refactoring the config to use only one or the other.
- Avoiding in-source builds of the C Driver by using the `build` subdirectory as the binary directory for CMake.

### DIFF
--- a/.evergreen/legacy_config_generator/evergreen_config_lib/tasks.py
+++ b/.evergreen/legacy_config_generator/evergreen_config_lib/tasks.py
@@ -735,7 +735,7 @@ all_tasks = chain(
             commands=[
                 shell_mongoc(
                     """
-            env SANITIZE=address DEBUG=ON SASL=AUTO SSL=OPENSSL EXTRA_CONFIGURE_FLAGS='-DENABLE_EXTRA_ALIGNMENT=OFF' .evergreen/scripts/compile.sh
+            env SANITIZE=address SASL=AUTO SSL=OPENSSL EXTRA_CONFIGURE_FLAGS='-DENABLE_EXTRA_ALIGNMENT=OFF' .evergreen/scripts/compile.sh
             """,
                     add_expansions_to_env=True,
                 ),
@@ -807,7 +807,7 @@ class SSLTask(Task):
         if cflags:
             script += f" CFLAGS={cflags}"
 
-        script += " DEBUG=ON SASL=OFF"
+        script += " SASL=OFF"
 
         if enable_ssl is not False:
             script += " SSL=" + enable_ssl
--- a/.evergreen/legacy_config_generator/evergreen_config_lib/tasks.py
+++ b/.evergreen/legacy_config_generator/evergreen_config_lib/tasks.py
@@ -70,9 +70,8 @@ class CompileTask(NamedTask):
 
         # Environment variables for .evergreen/scripts/compile.sh.
         self.compile_sh_opt: dict[str, str] = {}
-        if config == "debug":
-            self.compile_sh_opt["DEBUG"] = "ON"
-        else:
+
+        if config != "debug":
             assert config == "release"
             self.compile_sh_opt["RELEASE"] = "ON"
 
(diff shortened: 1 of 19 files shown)


.evergreen/scripts/compile-unix.sh
```diff
@@ -82,8 +82,14 @@ configure_flags_append_if_not_null SRV "-DENABLE_SRV=${SRV}"
 configure_flags_append_if_not_null TRACING "-DENABLE_TRACING=${TRACING}"
 configure_flags_append_if_not_null ZLIB "-DENABLE_ZLIB=${ZLIB}"
 
+if [[ "${DEBUG:-}" == "ON" && "${RELEASE:-}" == "ON" ]]; then
```
kevinAlbs, line 85: > We should consider refactoring the config to use only one or the other.

Consider removing `RELEASE` since it appears unused. Though it could be done in a future PR.

.evergreen/scripts/compile-unix.sh
```diff
@@ -208,12 +210,12 @@ fi
 . "${script_dir:?}/find-ccache.sh"
 find_ccache_and_export_vars "$(pwd)" || true
 
-"${cmake_binary}" "${configure_flags[@]}" ${extra_configure_flags[@]+"${extra_configure_flags[@]}"} .
-"${cmake_binary}" --build .
-"${cmake_binary}" --install .
+"${cmake_binary}" -S . -B build "${configure_flags[@]}" ${extra_configure_flags[@]+"${extra_configure_flags[@]}"} .
```
kevinAlbs, line 213:
```suggestion
"${cmake_binary}" -S . -B cmake-build "${configure_flags[@]}" ${extra_configure_flags[@]+"${extra_configure_flags[@]}"} .
```

Consider using `cmake-build` to avoid possibly interfering with existing files in the `build` directory.

kevinAlbs (approved):
LGTM with minor suggestions.

.evergreen/scripts/compile-unix.sh
eramongodb: `RELEASE` is used by the `release-compile` EVG task. It may be preferable to remove `DEBUG` instead and let all EVG tasks use Debug build type by default.

eramongodb:
Latest changes verified by [this patch](https://spruce.mongodb.com/version/66ce3047b8626b000798dadd).

.evergreen/legacy_config_generator/evergreen_config_lib/tasks.py
```diff
@@ -736,7 +736,7 @@ def pre_commands(self) -> Iterable[Value]:
             commands=[
                 shell_mongoc(
                     """
-            env SANITIZE=address DEBUG=ON SASL=AUTO SSL=OPENSSL EXTRA_CONFIGURE_FLAGS='-DENABLE_EXTRA_ALIGNMENT=OFF' .evergreen/scripts/compile.sh
```
kevinAlbs, line 739: Remove reference to no-longer-needed `DEBUG` environment variable on line 74:

```python
        if config == "debug":
            self.compile_sh_opt["DEBUG"] = "ON"
        else:
            assert config == "release"
            self.compile_sh_opt["RELEASE"] = "ON"
```

kevinAlbs (approved):
LGTM with another `DEBUG` removal.

PR #1712, opened 2024-08-28
CDRIVER-4800 return from trace functions if no logger

joshbsiegel:
# Summary
This ticket adds a macro to return early from `mongoc_log_trace_bytes()` and `mongoc_log_trace_iovec()` if no log handler is registered.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-log-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-log-private.h
@@ -34,14 +34,12 @@ mongoc_log_trace_bytes (const char *domain, const uint8_t *_b, size_t _l);
 void
 mongoc_log_trace_iovec (const char *domain, const mongoc_iovec_t *_iov, size_t _iovcnt);
 
-BSON_BEGIN_DECLS
-#define STOP_LOGGING_CHECK                                          \
-   int stop_logging;                                                \
-   stop_logging = !gLogFunc;                                        \
-   stop_logging = stop_logging || !_mongoc_log_trace_is_enabled (); \
-   if (stop_logging) {                                              \
-      return;                                                       \
-   }
+#define STOP_LOGGING_CHECK                                 \
+   if (1) {                                                \
+      if (!gLogFunc || !_mongoc_log_trace_is_enabled ()) { \
+         return;                                           \
+      }                                                    \
+   } else                                                  \
+      (void) 0
 
-BSON_END_DECLS
 #endif /* MONGOC_LOG_PRIVATE_H */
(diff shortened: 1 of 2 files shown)


src/libmongoc/src/mongoc/mongoc-log-private.h
```diff
@@ -34,4 +34,14 @@ mongoc_log_trace_bytes (const char *domain, const uint8_t *_b, size_t _l);
 void
 mongoc_log_trace_iovec (const char *domain, const mongoc_iovec_t *_iov, size_t _iovcnt);
 
+BSON_BEGIN_DECLS
+#define STOP_LOGGING_CHECK                                          \
```
joshbsiegel, line 38: This macro is only used in `mongoc_log_trace_bytes()` and `mongoc_log_trace_iovec()`, not `mongoc_log()` as the JIRA ticket recommends. This is because the current check in `mongoc_log()` needs information on whether the log level is `TRACE`, whereas this information is implicit in the two functions that this macro is used in.

src/libmongoc/src/mongoc/mongoc-log-private.h
```diff
@@ -34,4 +34,14 @@ mongoc_log_trace_bytes (const char *domain, const uint8_t *_b, size_t _l);
 void
 mongoc_log_trace_iovec (const char *domain, const mongoc_iovec_t *_iov, size_t _iovcnt);
 
+BSON_BEGIN_DECLS
```
kevinAlbs, line 37: Suggest removing `BSON_BEGIN_DECLS` and `BSON_END_DECLS`. Those are not needed for private definitions.

Aside: the macros conditionally define `extern "C" { ... }` to tell C++ compilers to treat the symbols as C declarations. But private headers are not expected to be used directly by C++ code.

src/libmongoc/src/mongoc/mongoc-log-private.h
```diff
@@ -34,4 +34,14 @@ mongoc_log_trace_bytes (const char *domain, const uint8_t *_b, size_t _l);
...
+#define STOP_LOGGING_CHECK                                          \
+   int stop_logging;                                                \
+   stop_logging = !gLogFunc;                                        \
+   stop_logging = stop_logging || !_mongoc_log_trace_is_enabled (); \
+   if (stop_logging) {                                              \
+      return;                                                       \
+   }
```
kevinAlbs, line 44: Suggest removing `int stop_logging` to simplify. Wrap with `if (1) { ... } else (void) 0` so `STOP_LOGGING_CHECK;` not result in an empty statement (This pattern was adopted somewhat recently.  See [Handling control flow inside macros](https://gustedt.wordpress.com/2011/02/02/handling-control-flow-inside-macros/)):
```c
#define STOP_LOGGING_CHECK                                 \
   if (1) {                                                \
      if (!gLogFunc || !_mongoc_log_trace_is_enabled ()) { \
         return;                                           \
      }                                                    \
   } else                                                  \
      (void) 0
```

kevinAlbs (approved):

PR #1713, opened 2024-08-29
CDRIVER-4726 make delimiting slash in URI optional

joshbsiegel:
# Summary
This ticket removes the requirement for URIs to have a delimiting slash in between the hosts and options (e.g. `mongodb://example.com?w=1` is equivalent to `mongodb://example.com/?w=1`).

# Testing
Relevant spec tests were updated according to [this](https://github.com/mongodb/specifications/pull/1447) PR.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-uri.c
+++ b/src/libmongoc/src/mongoc/mongoc-uri.c
@@ -1410,7 +1410,9 @@ mongoc_uri_parse (mongoc_uri_t *uri, const char *str, bson_error_t *error)
 {
    BSON_ASSERT_PARAM (str);
 
-   char *before_options = NULL;
+   char *before_slash = NULL;
+   char *userpass = NULL;
+   char *hostname = NULL;
    const char *tmp;
 
    if (!bson_utf8_validate (str, strlen (str), false /* allow_null */)) {
@@ -1423,20 +1425,34 @@ mongoc_uri_parse (mongoc_uri_t *uri, const char *str, bson_error_t *error)
       goto error;
    }
 
-   // Check for delimiting slash (not required)
-   before_options = scan_to_unichar (str, '/', "", &tmp);
-   if (!before_options) {
-      before_options = scan_to_unichar (str, '?', "", &tmp);
+   before_slash = scan_to_unichar (str, '/', "", &tmp);
+   if (!before_slash) {
+      // Handle cases of optional delimiting slash
+
+      // Skip any "?" that exist is the userpass
+      userpass = scan_to_unichar (str, '@', "", &tmp);
+      if (!userpass) {
+         // If none found, we can safely check for "?" indicating beginning of options
+         before_slash = scan_to_unichar (str, '?', "", &tmp);
+      } else {
+         // Otherwise, see if options exist after username/password and concatenate result
+         hostname = scan_to_unichar (tmp, '?', "", &tmp);
+
+         if (hostname) {
+            before_slash = bson_strdup (str);
+            before_slash[strlen (userpass) + strlen (hostname)] = '\0';
+         }
+      }
    }
 
-   if (!before_options) {
-      before_options = bson_strdup (str);
-      str += strlen (before_options);
+   if (!before_slash) {
+      before_slash = bson_strdup (str);
+      str += strlen (before_slash);
    } else {
       str = tmp;
    }
 
-   if (!mongoc_uri_parse_before_slash (uri, before_options, error)) {
+   if (!mongoc_uri_parse_before_slash (uri, before_slash, error)) {
       goto error;
    }
 
@@ -1475,11 +1491,15 @@ mongoc_uri_parse (mongoc_uri_t *uri, const char *str, bson_error_t *error)
       goto error;
    }
 
-   bson_free (before_options);
+   bson_free (before_slash);
+   bson_free (userpass);
+   bson_free (hostname);
    return true;
 
 error:
-   bson_free (before_options);
+   bson_free (before_slash);
+   bson_free (userpass);
+   bson_free (hostname);
    return false;
 }
 
--- a/src/libmongoc/src/mongoc/mongoc-uri.c
+++ b/src/libmongoc/src/mongoc/mongoc-uri.c
@@ -1427,7 +1427,7 @@ mongoc_uri_parse (mongoc_uri_t *uri, const char *str, bson_error_t *error)
    if (!before_slash) {
       // Handle cases of optional delimiting slash
       char *userpass = NULL;
-      char *hostname = NULL;
+      char *hosts = NULL;
 
       // Skip any "?"s that exist in the userpass
       userpass = scan_to_unichar (str, '@', "", &tmp);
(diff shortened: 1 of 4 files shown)


src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -1423,23 +1423,28 @@ mongoc_uri_parse (mongoc_uri_t *uri, const char *str, bson_error_t *error)
...
    }
 
-   before_slash = scan_to_unichar (str, '/', "", &tmp);
-   if (!before_slash) {
-      before_slash = bson_strdup (str);
-      str += strlen (before_slash);
+   // Check for delimiting slash (not required)
+   before_options = scan_to_unichar (str, '/', "", &tmp);
+   if (!before_options) {
+      before_options = scan_to_unichar (str, '?', "", &tmp);
```
kevinAlbs, line 1429: This change may prevent including `?` in the userinfo. Example:

```c
bson_error_t error;
mongoc_uri_t *uri = mongoc_uri_new_with_error ("mongodb://user:pa?s@localhost", &error);
ASSERT_OR_PRINT (uri, error); // Passes before PR. With PR, fails with "Invalid host string in URI".
ASSERT_CMPSTR (mongoc_uri_get_password (uri), "pa?s");
mongoc_uri_destroy (uri);
```

The [Connection String spec](https://github.com/mongodb/specifications/blob/e9f02f328a93d6a0321694cc30d031079c555e4a/source/connection-string/connection-string-spec.md#userinfo-optional) notes:

> drivers SHOULD require connection string user information to follow the "userinfo" encoding rules of RFC 3986 and SHOULD throw an exception if disallowed characters are detected. However, for backwards-compatibility reasons, drivers MAY allow reserved characters other than "@" and ":" to be present in user information without percent-encoding.

RFC 3968 prohibits a `?` in the userinfo, but the C driver permits it. I expect `?` is used in passwords, and given the [difficulties encountered](https://jira.mongodb.org/browse/CDRIVER-4119?focusedCommentId=5518486&focusedId=5518486&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-5518486) with non-ASCII passwords, it would not surprise me if there is real-world use passing literal `?` in the password.

I suggest adding this test case, and changing to preserve accepting `?` in userinfo. Consider trying to find the `?` after a possible `@`.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -1423,23 +1423,28 @@ mongoc_uri_parse (mongoc_uri_t *uri, const char *str, bson_error_t *error)
...
       goto error;
    }
 
    BSON_ASSERT (str);
 
    if (*str) {
-      if (*str == '/') {
-         str++;
+      // Check for valid end of hostname delimeter (skip slash if necessary)
+      if (*str == '?' || *str++ == '/') {
```
kevinAlbs, line 1447: Suggest splitting the conditions for simplicity. IMO the conditionally run `*str++` adds some difficulty to reason about. Example:

```c
if (*str != '/' && *str != '?') {
   MONGOC_URI_ERROR (error, "%s", "Expected end of hostname delimiter");
   goto error;
}

if (*str == '/') {
   // Try to parse database.
   str++;
   if (*str) {
      if (!mongoc_uri_parse_database (uri, str, &str)) {
         MONGOC_URI_ERROR (error, "%s", "Invalid database name in URI");
         goto error;
      }
   }
}

if (*str == '?') {
   // Try to parse options.
   str++;
   if (*str) {
      if (!mongoc_uri_parse_options (uri, str, false /* from DNS */, error)) {
         goto error;
      }
   }
}
```

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -1424,6 +1426,25 @@ mongoc_uri_parse (mongoc_uri_t *uri, const char *str, bson_error_t *error)
...
+      if (!userpass) {
+         // If none found, we can safely check for "?" indicating beginning of options
+         before_slash = scan_to_unichar (str, '?', "", &tmp);
+      } else {
+         // Otherwise, see if options exist after username/password and concatenate result
+         hostname = scan_to_unichar (tmp, '?', "", &tmp);
+
+         if (hostname) {
+            before_slash = bson_strdup (str);
+            before_slash[strlen (userpass) + strlen (hostname)] = '\0';
```
joshbsiegel, line 1443: Open for suggestions on how to accomplish this in a more efficient way. I tried messing around with `bson_strncpy` but couldn't get it to work.

eramongodb: `scan_to_unichar` returns a pointer past the last scanned character via `&tmp` when the return value is not null. Therefore, the length of the scanned string given `scan_to_unichar (begin, ..., &end)` can be computed as `end - begin`.

```c
// Handle cases of optional delimiting slash

// Skip any "?"s that exist in the userpass
userpass = scan_to_unichar (str, '@', "", &tmp);
if (!userpass) {
   // If none found, safely check for "?" indicating beginning of options
   before_slash = scan_to_unichar (str, '?', "", &tmp);
} else {
   const size_t userpass_len = (size_t) (tmp - str);

   // Otherwise, see if options exist after userpass and concatenate result
   hostname = scan_to_unichar (tmp, '?', "", &tmp);

   if (hostname) {
      const size_t hostname_len = (size_t) (tmp - str) -userpass_len;

      before_slash = bson_strndup (str, userpass_len + hostname_len);
   }
}
```

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -1438,26 +1459,31 @@ mongoc_uri_parse (mongoc_uri_t *uri, const char *str, bson_error_t *error)
...
          }
+      }
 
-         if (*str == '?') {
-            str++;
-            if (*str) {
-               if (!mongoc_uri_parse_options (uri, str, false /* from DNS */, error)) {
-                  goto error;
-               }
+      if (*str == '?') {
```
eramongodb, line 1479:
```suggestion
      else if (*str == '?') {
```

Recommend using `else if` to denote related and exclusive conditional branches.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -1466,10 +1492,14 @@ mongoc_uri_parse (mongoc_uri_t *uri, const char *str, bson_error_t *error)
    }
 
    bson_free (before_slash);
+   bson_free (userpass);
+   bson_free (hostname);
```
eramongodb, line 1496: `userpass` and `hostname` are not used outside of the "Handle cases of optional delimiting slash" block, and there are no jumps out of the block. Move the `bson_free()` calls to the end of that block + reduce scope of the `userpass` and `hostname` variables.

```c
if (!before_slash) {
   // Handle cases of optional delimiting slash

   const char* const userpass = ...;
   ...

   bson_free (userpass);
   bson_free (hostname);
}
```

eramongodb (approved):

src/libmongoc/src/mongoc/mongoc-uri.c
joshbsiegel: I think keeping this conditional as an `if` is necessary to ensure that a URI can first be parsed for its database name and then separately be checked for options (e.g. `mongodb://localhost:27017/test?replicaset=foo`). Changing to an `else if` will parse the database name but then skip parsing the options completely. Let me know if there's a misunderstanding on my end.

eramongodb: Good catch. I neglected to consider that `str` at this point depends on the preceeding `str++` and `mongoc_uri_parse_database()` even in the `*str == '/' -> true` case.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -1424,6 +1424,32 @@ mongoc_uri_parse (mongoc_uri_t *uri, const char *str, bson_error_t *error)
    }
 
    before_slash = scan_to_unichar (str, '/', "", &tmp);
+   if (!before_slash) {
+      // Handle cases of optional delimiting slash
+      char *userpass = NULL;
+      char *hostname = NULL;
```
kevinAlbs, line 1430: Suggest renaming `hostname` to `hosts` since it may contain multiple comma delimited [Host Identifiers](https://github.com/mongodb/specifications/blob/e651ebe819743f1cffae938f5488c69d5277c21f/source/connection-string/connection-string-spec.md#general-syntax), e.g. `mongodb://host1,host2`.

kevinAlbs (approved):
LGTM with a suggested rename.

PR #1719, opened 2024-09-05
CDRIVER-5681 Address potential buffer overflow when writing null terminator after decimal128 significand digits

eramongodb:
Resolves CDRIVER-5681. See ticket for details.

### DIFF
--- a/src/libbson/tests/test-decimal128.c
+++ b/src/libbson/tests/test-decimal128.c
@@ -108,6 +108,7 @@ test_decimal128_to_string__regular (void)
    bson_decimal128_t trailing_zeros;
    bson_decimal128_t all_digits;
    bson_decimal128_t full_house;
+   bson_decimal128_t untruncated_significand;
    bson_decimal128_t truncated_significand;
 
    DECIMAL128_FROM_ULLS (one, 0x3040000000000000, 0x0000000000000001);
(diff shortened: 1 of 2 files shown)


kevinAlbs (approved):
Great catch. LGTM.

Suggest adding a regression test, in case `bson_decimal128_to_string` is modified again in the future.

src/libbson/tests/test-decimal128.c
```diff
@@ -108,57 +108,77 @@ test_decimal128_to_string__regular (void)
...
 
-   /* 5192296858534827628530496329220095 */
+   // 5192296858534827628530496329220095
    DECIMAL128_FROM_ULLS (full_house, 0x3040ffffffffffff, 0xffffffffffffffff);
 
+   // 0.0000010384593717069655257060992658440191
+   DECIMAL128_FROM_ULLS (untruncated_significand, 0x2ff1ffffffffffff, 0xffffffffffffffff);
+
+   // -0.000001038459371706965525706099265844019
+   DECIMAL128_FROM_ULLS (truncated_significand, 0xaff1ffffffffffff, 0xffffffffffffffff);
```
eramongodb, line 142: After much effort trying to craft a human-readable/easily-understandable test case for this bug, it is now unclear to me whether this "significand truncation" behavior is even the correct behavior to begin with. I am having great difficulty finding any detailed specification or reference for the decimal128 (BID) -> string conversion algorithm against which the results can be compared. Nevertheless, I added an `untruncated_significand` test case to hopefully make clear how the inclusion of the negative sign leads to the omission of the trailing `1` digit at the end due to length limitations.

eramongodb:
Added annotations for values where inexact rounding errors may occur when attempting to convert the string representation to a Decimal128 value. On that note, added missing assertions of the return value of `bson_decimal128_from_string*()` and converted `BSON_ASSERT -> ASSERT` for better error messages on assertion failure.

vector-of-bool (approved):
LGTM

PR #1720, opened 2024-09-06
WIP: CDRIVER-5577 add serverMonitoringMode URI option

joshbsiegel:
# Summary 
This ticket adds functionality and tests for a new URI option `serverMonitoringMode`.

# Background
Changes motivated by [DRIVERS-2578](https://github.com/mongodb/specifications/commit/14a6c8134d0c3038cdaa34bb33bcebbb3a545ca8). In summary, there's a new URI option `serverMonitoringMode` that the user can pass through with one of three values: `stream`, `poll`, or `auto`. The behavior is as follows:
- `poll`: use poll monitoring
- `stream`: use stream monitoring if the server supports it (version >= 4.4)
- `auto` (default): `stream` if the server supports it AND the application isn't in a FAAS environment, `poll` otherwise

# Testing
There are two new spec tests (`sdam-options.json` and `serverMonitoringMode.json`) as specified in the drivers ticket. Functionality for new events `serverHeartbeat{Started/Succeeded/Failed}Event` and operations `waitForEvent` and `createEntities` was implemented. The event implementations were modeled after the command event callback structure.

The test runner was modified to create a client pool and pop a client to run tests on rather than run tests on a single-threaded client. This is because the stream monitoring functionality could not be tested on a single-threaded client due to the blocking that occurs during the procedure. This change results in all unified tests being ran on a pooled client. 

# What's New
- `serverMonitoringMode` logic in `mongoc-server-monitor.c`
- Functionality to support new URI option
- 2 new spec tests
- Support for 3 new events and 2 new operations
- Support for asserting `awaited` in a server event


src/libmongoc/tests/json/server_discovery_and_monitoring/unified/serverMonitoringMode.json
```diff
@@ -0,0 +1,459 @@
...
+      "runOnRequirements": [
+        {
+          "minServerVersion": "4.4.0"
+        }
+      ],
+      "operations": [
+        {
+          "name": "createEntities",
+          "object": "testRunner",
+          "ignoreResultAndError": true,
```
joshbsiegel, line 26: `"ignoreResultAndError": true` was added across this spec test to prevent the test runner from checking the result of an operation that doesn't have one (all instances of `createEntities` and `waitForEvent`). Open to suggestions on how to fix this issue without modifying the spec test. Maybe we can add a conditional in `operation.c` where `ignore_result_and_error` is checked but this feels a little clunky.

PR #1722, opened 2024-09-10
CDRIVER-5706 spawn pooled client in unified test runner

joshbsiegel:
# Description
This ticket updates the unified test runner to run tests on a pooled client instead of a single-threaded client.

# Background
Some unified tests require the client to be from a pool in order to test expected behavior (e.g. stream monitoring tests in SDAM). This motivated the change of having all unified tests run on a pooled client. Existing spec tests shouldn't be affected by this behavior.

# `entity->callback_disabled`
Because `mongoc_client_pool_set_apm_callbacks` prevents callbacks from getting set more than once on a pool, functionality in `entity_map_disable_event_listeners` wouldn't work with a simple call to disable the APM callbacks in a pool. To mimic this functionality, an atomic int was added to `entity_t` that is set in `entity_map_disable_event_listeners` and checked in `apm_command_callback` before observing events.

# What's New
- `callback_disabled` atomic int in `entity_t`
- Pooled client used to run unified tests

### DIFF
--- a/src/libmongoc/tests/unified/entity-map.c
+++ b/src/libmongoc/tests/unified/entity-map.c
@@ -161,6 +161,7 @@ entity_new (entity_map_t *em, const char *type)
    entity = bson_malloc0 (sizeof (entity_t));
    entity->type = bson_strdup (type);
    entity->entity_map = em;
+   entity->callbacks_disabled = 0;
    _mongoc_array_init (&entity->observe_events, sizeof (observe_event_t));
    _mongoc_array_init (&entity->store_events, sizeof (store_event_t));
    return entity;
@@ -478,6 +479,11 @@ apm_command_callback (command_callback_funcs_t funcs, const char *type, const vo
 
    BSON_ASSERT (funcs.get_context);
    entity_t *const entity = (entity_t *) funcs.get_context (apm_command);
+
+   // Check if callbacks are disabled
+   if (bson_atomic_int_fetch (&entity->callbacks_disabled, bson_memory_order_relaxed)) {
+      return;
+   }
 
    observe_event (entity, funcs, type, apm_command);
    store_event_to_entities (entity, funcs, type, apm_command);
@@ -2228,9 +2234,7 @@ entity_map_disable_event_listeners (entity_map_t *em)
    LL_FOREACH (em->entities, eiter)
    {
       if (0 == strcmp (eiter->type, "client")) {
-         mongoc_client_t *client = (mongoc_client_t *) eiter->value;
-
-         mongoc_client_set_apm_callbacks (client, NULL, NULL);
+         bson_atomic_int_fetch_add (&eiter->callbacks_disabled, 1, bson_memory_order_acquire);
       }
    }
 }
--- a/src/libmongoc/tests/unified/entity-map.c
+++ b/src/libmongoc/tests/unified/entity-map.c
@@ -617,6 +617,7 @@ entity_t *
 entity_client_new (entity_map_t *em, bson_t *bson, bson_error_t *error)
 {
    entity_t *entity = NULL;
+   mongoc_client_pool_t *pool = NULL;
    mongoc_client_t *client = NULL;
    mongoc_uri_t *uri = NULL;
    bool ret = false;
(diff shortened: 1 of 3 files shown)


src/libmongoc/tests/unified/entity-map.c
```diff
@@ -479,6 +480,11 @@ apm_command_callback (command_callback_funcs_t funcs, const char *type, const vo
    BSON_ASSERT (funcs.get_context);
    entity_t *const entity = (entity_t *) funcs.get_context (apm_command);
 
+   // Check if callbacks are disabled
+   if (bson_atomic_int_fetch (&entity->callbacks_disabled, bson_memory_order_relaxed)) {
```
kevinAlbs, line 484: On further read, I am concerned about accessing shared data within callbacks.

I do not think it is a problem yet within this PR, since the only monitored events are command events. But I expect adding SDAM events may result in shared access:

With a client pool, callbacks may be called from separate threads. Example:
- A server monitor thread calls: `_server_monitor_thread` => `_server_monitor_check_server` => `_server_monitor_heartbeat_started` (calls heartbeat started callback)
- Meanwhile, the main thread calls: `mongoc_client_command_simple` => `_mongoc_client_command_with_stream` => `mongoc_cluster_run_command_monitored` (calls command started callback)

This may result in new inconsistent test failures (e.g. missed events due to a race).

Identifying each case of shared access and locking may be a significant effort, and difficult to get right. Since using a client pool is only needed for the spec test `serverMonitoringMode.yml`, I would rather implement the functionality of that spec test separately. Example:

```c
// `smm_t` is a test fixture for serverMonitoringMode tests.
#define MAX_EVENTS 3
typedef struct {
   bson_mutex_t lock;
   bool awaited[MAX_EVENTS];
   size_t awaited_len;
   mongoc_client_pool_t *pool;
} smm_t;

static void
heartbeat_started (const mongoc_apm_server_heartbeat_started_t *event)
{
   smm_t *t = mongoc_apm_server_heartbeat_started_get_context (event);
   bson_mutex_lock (&t->lock);
   // Store the most recent awaited value (if room).
   if (t->awaited_len < MAX_EVENTS) {
      bool awaited = mongoc_apm_server_heartbeat_started_get_awaited (event);
      t->awaited[t->awaited_len] = awaited;
      t->awaited_len++;
   }
   bson_mutex_unlock (&t->lock);
}

static smm_t *
smm_new (const char *mode)
{
   smm_t *t = bson_malloc0 (sizeof (smm_t));
   bson_mutex_init (&t->lock);

   // Create client pool.
   {
      mongoc_uri_t *uri = test_framework_get_uri ();
      mongoc_uri_set_option_as_int32 (uri, MONGOC_URI_HEARTBEATFREQUENCYMS, 500); // To speed up test.
      mongoc_uri_set_option_as_utf8 (uri, "serverMonitoringMode", mode);
      t->pool = mongoc_client_pool_new (uri);
      test_framework_set_pool_ssl_opts (t->pool);
      mongoc_uri_destroy (uri);
   }

   // Capture heartbeat started events.
   {
      mongoc_apm_callbacks_t *cbs = mongoc_apm_callbacks_new ();
      mongoc_apm_set_server_heartbeat_started_cb (cbs, heartbeat_started);
      mongoc_client_pool_set_apm_callbacks (t->pool, cbs, t);
      mongoc_apm_callbacks_destroy (cbs);
   }

   // Pop and push a client to start monitoring.
   {
      mongoc_client_t *client = mongoc_client_pool_pop (t->pool);
      mongoc_client_pool_push (t->pool, client);
   }

   return t;
}

// `smm_wait` waits for `count` heartbeat started events.
static bool
smm_wait (smm_t *t, size_t count)
{
   int64_t started = bson_get_monotonic_time ();
   while (true) {
      bson_mutex_lock (&t->lock);
      if (t->awaited_len >= count) {
         bson_mutex_unlock (&t->lock);
         return true;
      }
      bson_mutex_unlock (&t->lock);

      int64_t now = bson_get_monotonic_time ();
      if (now - started > 10 * 1000 * 1000) {
         break;
      }
      _mongoc_usleep (500 * 1000); // Sleep for 500ms.
   }
   return false;
}

static void
smm_destroy (smm_t *t)
{
   if (!t) {
      return;
   }
   mongoc_client_pool_destroy (t->pool);
   bson_mutex_destroy (&t->lock);
   bson_free (t);
}

// `test_serverMonitoringMode` implements spec tests from serverMonitoringMode.yml.
// The spec test needs a client pool for stream monitoring. The unified test runner only uses single-threaded clients.
static void
test_serverMonitoringMode (void *unused)
{
   BSON_UNUSED (unused);

   if (test_framework_is_replset ()) {
      printf ("Test is skipped. SDAM events are non-deterministic when monitoring multiple servers.");
      return;
   }

   if (test_framework_get_server_version () >= test_framework_str_to_version ("4.4.0")) {
      printf ("'connect with serverMonitoringMode=auto >=4.4' ... begin\n");
      smm_t *t = smm_new ("auto");
      ASSERT (smm_wait (t, 2));
      // Check events.
      {
         bson_mutex_lock (&t->lock);
         ASSERT_CMPSIZE_T (t->awaited_len, >=, 2);
         ASSERT (!t->awaited[0]); // Polls.
         ASSERT (t->awaited[1]); // Streams.
         bson_mutex_unlock (&t->lock);
      }
      smm_destroy (t);
      printf ("'connect with serverMonitoringMode=auto >=4.4' ... end\n");
   }

   // TODO: other tests.
}
```

PR #1723, opened 2024-09-10
CDRIVER-5707 ignore unknown servers in aggregate-with-write check

kevinAlbs:
# Summary

Ignore unknown servers when checking to override read preferences for an aggregate-with-write.

Verified by this patch build: https://spruce.mongodb.com/version/66e0ad1ebdf6d3000745d0ba

# Background & Motivation

See CDRIVER-5707.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-topology-description.c
+++ b/src/libmongoc/src/mongoc/mongoc-topology-description.c
@@ -726,7 +726,7 @@ _check_any_server_less_than_wire_version_13 (const void *sd_, void *any_too_old_
  * @brief Calculate the read mode that we should be using, based on what was
  * requested and what is available in the topology.
  *
- * Per the CRUD spec, if the requested read mode is *not* primary, and *any*
+ * Per the CRUD spec, if the requested read mode is *not* primary, and *any* available
  * server in the topology has a wire version < server v5.0, we must override the
  * read mode preference with "primary." Server v5.0 indicates support on a
  * secondary server for using aggregate pipelines that contain writing stages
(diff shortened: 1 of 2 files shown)


src/libmongoc/src/mongoc/mongoc-topology-description.c
```diff
@@ -714,7 +714,7 @@ _check_any_server_less_than_wire_version_13 (const void *sd_, void *any_too_old_
 {
    const mongoc_server_description_t *sd = sd_;
    bool *any_too_old = any_too_old_;
-   if (sd->max_wire_version < WIRE_VERSION_5_0) {
+   if (sd->type != MONGOC_SERVER_UNKNOWN && sd->max_wire_version < WIRE_VERSION_5_0) {
```
joshbsiegel, line 717: I think a similar additional check is needed in `mongoc_topology_description_select` here:
```cpp
if (optype == MONGOC_SS_AGGREGATE_WITH_WRITE && sd->max_wire_version < WIRE_VERSION_5_0) {
   /* The single server may be part of an unseen replica set that may not
    * support aggr-with-write operations on secondaries. Force the read
    * preference to use a primary. */
    if (must_use_primary) {
      *must_use_primary = true;
    }
  }
```
Sharded clusters were failing [here](https://spruce.mongodb.com/version/66e0ac5813acb00007aa8b0a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) without this check but [pass](https://spruce.mongodb.com/version/66e0b9b95889c300071c0e7a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) with the addition.

joshbsiegel (changes requested):

joshbsiegel (approved):

vector-of-bool (approved):
LGTM

PR #1725, opened 2024-09-13
CDRIVER-5713 fix bson_utf8_escape_for_json crash

joshbsiegel:
# Summary
This ticket patches an issue in `bson_utf8_escape_for_json` that causes a crash for some invalid inputs. 

See [CDRIVER-5713](https://jira.mongodb.org/browse/CDRIVER-5713)

### DIFF
--- a/src/libbson/doc/bson_utf8_escape_for_json.rst
+++ b/src/libbson/doc/bson_utf8_escape_for_json.rst
@@ -32,3 +32,7 @@ Returns
 -------
 
 A newly allocated string that should be freed with :symbol:`bson_free()`.
+
+.. warning::
+  If ``utf8`` represents invalid UTF-8 data :symbol:`bson_utf8_escape_for_json` may return NULL or have undefined behavior.
+  UTF-8 may be validated with :symbol:`bson_utf8_validate`.
--- a/src/libbson/doc/bson_utf8_escape_for_json.rst
+++ b/src/libbson/doc/bson_utf8_escape_for_json.rst
@@ -31,8 +31,4 @@ byte is found before ``utf8_len`` bytes, it is converted to
 Returns
 -------
 
-A newly allocated string that should be freed with :symbol:`bson_free()`.
-
-.. warning::
-  If ``utf8`` represents invalid UTF-8 data :symbol:`bson_utf8_escape_for_json` may return NULL or have undefined behavior.
-  UTF-8 may be validated with :symbol:`bson_utf8_validate`.
+A newly allocated string that should be freed with :symbol:`bson_free()` when ``utf8`` is a valid UTF-8 string, or ``NULL`` if the (possibly invalid UTF-8) string could not be escaped.
--- a/src/libbson/fuzz/fuzz_test_bson_utf8_escape_for_json.c
+++ b/src/libbson/fuzz/fuzz_test_bson_utf8_escape_for_json.c
@@ -1,0 +1,16 @@
+#include <bson/bson.h>
+
+int
+LLVMFuzzerTestOneInput (const uint8_t *data, size_t size)
+{
+   // Check if input crashes program (ok if return NULL)
+   char *got = bson_utf8_escape_for_json ((const char *) data, (ssize_t) size);
+
+   // If UTF-8 is valid, we should get some non-null response
+   if (bson_utf8_validate ((const char *) data, size, false)) {
+      BSON_ASSERT (got);
+   }
+
+   bson_free (got);
+   return 0;
+}
--- a/src/libbson/fuzz/fuzz_test_bson_utf8_escape_for_json.c
+++ b/src/libbson/fuzz/fuzz_test_bson_utf8_escape_for_json.c
@@ -3,11 +3,16 @@
 int
 LLVMFuzzerTestOneInput (const uint8_t *data, size_t size)
 {
+   // Reject inputs with lengths greater than ssize_t
+   if (bson_cmp_greater_us (size, SSIZE_MAX)) {
+      return -1;
+   }
+
    // Check if input crashes program (ok if return NULL)
    char *got = bson_utf8_escape_for_json ((const char *) data, (ssize_t) size);
 
    // If UTF-8 is valid, we should get some non-null response
-   if (bson_utf8_validate ((const char *) data, size, false)) {
+   if (bson_utf8_validate ((const char *) data, size, true)) {
       BSON_ASSERT (got);
    }
 
--- a/src/libbson/src/bson/bson-utf8.c
+++ b/src/libbson/src/bson/bson-utf8.c
@@ -283,9 +283,7 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
       // Check if expected char length goes past end
       _bson_utf8_get_sequence (utf8, &length_of_char, &mask);
       if (utf8 + length_of_char > end) {
-         // Invalid UTF-8
-         bson_string_free (str, true);
-         return NULL;
+         goto invalid_utf8;
       }
 
       c = bson_utf8_get_char (utf8);
@@ -323,18 +321,25 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
       if (c) {
          utf8 = bson_utf8_next_char (utf8);
       } else {
-         if (length_provided && !*utf8) {
-            /* we escaped nil as '\u0000', now advance past it */
-            utf8++;
+         if (length_provided) {
+            // Check if current character is null character,
+            // if so skip past it as we've already added '\\u0000' above
+            if (*utf8 == '\0' || (*utf8 == '\xc0' && *(utf8 + 1) == '\x80')) {
+               utf8 += (*utf8 == '\0') ? 1 : 2;
+            } else {
+               goto invalid_utf8;
+            }
          } else {
-            /* invalid UTF-8 */
-            bson_string_free (str, true);
-            return NULL;
+            goto invalid_utf8;
          }
       }
    }
 
    return bson_string_free (str, false);
+
+invalid_utf8:
+   bson_string_free (str, true);
+   return NULL;
 }
 
 
--- a/src/libbson/src/bson/bson-utf8.c
+++ b/src/libbson/src/bson/bson-utf8.c
@@ -265,6 +265,8 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
    bson_string_t *str;
    bool length_provided = true;
    const char *end;
+   uint8_t mask;
+   uint8_t length_of_char;
 
    BSON_ASSERT (utf8);
 
@@ -278,6 +280,14 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
    end = utf8 + utf8_len;
 
    while (utf8 < end) {
+      // Check if expected char length goes past end
+      _bson_utf8_get_sequence (utf8, &length_of_char, &mask);
+      if (utf8 + length_of_char > end) {
+         // Invalid UTF-8
+         bson_string_free (str, true);
+         return NULL;
+      }
+
       c = bson_utf8_get_char (utf8);
 
       switch (c) {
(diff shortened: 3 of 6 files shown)


src/libbson/doc/bson_utf8_escape_for_json.rst
```diff
@@ -32,3 +32,7 @@ Returns
 -------
 
 A newly allocated string that should be freed with :symbol:`bson_free()`.
+
+.. warning::
+  If ``utf8`` represents invalid UTF-8 data :symbol:`bson_utf8_escape_for_json` may return NULL or have undefined behavior.
+  UTF-8 may be validated with :symbol:`bson_utf8_validate`.
```
eramongodb, line 38:
```suggestion
A newly allocated string that should be freed with :symbol:`bson_free()` when ``utf8`` is a valid UTF-8 string, or ``NULL`` if the (possibly invalid UTF-8) string could not be escaped.
```

We should not expose undefined behavior to users when it is (easily) avoidable. Given this function is already capable of handling invalid inputs, recommend documenting the `NULL` result condition as a wider and simpler "could not be escaped" instead.

src/libbson/fuzz/fuzz_test_bson_utf8_escape_for_json.c
```diff
@@ -0,0 +1,16 @@
+#include <bson/bson.h>
+
+int
+LLVMFuzzerTestOneInput (const uint8_t *data, size_t size)
+{
```
eramongodb, line 5:
```suggestion
{
   if (bson_cmp_greater_us (size, SSIZE_MAX)) {
      return -1;
   }

```
Reject inputs whose length exceeds `ssize_t`.

src/libbson/fuzz/fuzz_test_bson_utf8_escape_for_json.c
```diff
@@ -0,0 +1,16 @@
...
+   // Check if input crashes program (ok if return NULL)
+   char *got = bson_utf8_escape_for_json ((const char *) data, (ssize_t) size);
+
+   // If UTF-8 is valid, we should get some non-null response
+   if (bson_utf8_validate ((const char *) data, size, false)) {
```
eramongodb, line 10:
```suggestion
   if (bson_utf8_validate ((const char *) data, size, true)) {
```
Allow for escaped null characters in UTF-8 strings given `bson_utf8_escape_for_json` is capable of supporting it.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -265,6 +265,8 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
    bson_string_t *str;
    bool length_provided = true;
    const char *end;
+   uint8_t mask;
+   uint8_t length_of_char;
```
eramongodb, line 269:
```suggestion
```
Reduce scope of local variables.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -245,7 +245,7 @@ bson_utf8_validate (const char *utf8, /* IN */
  *       two byte UTF-8 sequence.
  *
  * Parameters:
- *       @utf8: A UTF-8 encoded string.
+ *       @utf8: A validated UTF-8 encoded string.
```
eramongodb, line 248:
```suggestion
 *       @utf8: A UTF-8 encoded string.
```
Adding a validity requirement is a breaking change given the function is written to handle [invalid inputs](https://github.com/mongodb/mongo-c-driver/blob/8a85033b212dabf140442dd094f636ef2d44b8e5/src/libbson/src/bson/bson-utf8.c#L320). See other comment regarding "could not be escaped".

src/libbson/src/bson/bson-utf8.c
```diff
@@ -278,6 +280,14 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
    end = utf8 + utf8_len;
 
    while (utf8 < end) {
+      // Check if expected char length goes past end
+      _bson_utf8_get_sequence (utf8, &length_of_char, &mask);
+      if (utf8 + length_of_char > end) {
```
eramongodb, line 285:
```suggestion
      if (utf8 > end - length_of_char) {
```
Avoid forming invalid out-of-range pointers (pointer equivalent to integer overflow).

src/libbson/tests/test-utf8.c
```diff
@@ -81,6 +81,29 @@ test_bson_utf8_escape_for_json (void)
...
+
+   // 4 bytes expected
+   str = bson_utf8_escape_for_json ("\xf0", -1);
+   BSON_ASSERT (!str);
+
+   str = bson_utf8_escape_for_json ("\xf0\x9f", -1);
+   BSON_ASSERT (!str);
+
+   str = bson_utf8_escape_for_json ("\xf0\x9f\x9f", -1);
+   BSON_ASSERT (!str);
```
eramongodb, line 106:
```suggestion
   // 3 bytes expected
   {
      str = bson_utf8_escape_for_json ("\xed", -1);
      BSON_ASSERT (!str);

      str = bson_utf8_escape_for_json ("\xed\x90", -1);
      BSON_ASSERT (!str);
   }

   // 4 bytes expected
   {
      str = bson_utf8_escape_for_json ("\xf0", -1);
      BSON_ASSERT (!str);

      str = bson_utf8_escape_for_json ("\xf0\x9f", -1);
      BSON_ASSERT (!str);

      str = bson_utf8_escape_for_json ("\xf0\x9f\x9f", -1);
      BSON_ASSERT (!str);
   }
```
Scope related assertions with corresponding comment.

src/libbson/tests/test-utf8.c
```diff
@@ -81,6 +81,29 @@ test_bson_utf8_escape_for_json (void)
    str = bson_utf8_escape_for_json (unescaped, -1);
    BSON_ASSERT (0 == memcmp (str, "\\u000e", 7));
    bson_free (str);
+
+   // Invalid UTF-8 strings (should return null)
+
+   // 2 bytes expected
+   str = bson_utf8_escape_for_json ("\xc2", -1);
+   BSON_ASSERT (!str);
```
eramongodb, line 89:
```suggestion
   str = bson_utf8_escape_for_json ("\xc2", -1);
   BSON_ASSERT (!str);
   
   str = bson_utf8_escape_for_json ("...", -1);
   BSON_ASSERT (str);
   bson_free (str);
```
Include a test for the other side of the boundary condition.

Also applies to the other tests added below.

eramongodb (changes requested):

src/libbson/src/bson/bson-utf8.c
```diff
@@ -313,18 +322,25 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
...
-            /* we escaped nil as '\u0000', now advance past it */
-            utf8++;
+         if (length_provided) {
+            // Check if current character is null character,
+            // if so skip past it as we've already added '\\u0000' above
+            if (*utf8 == '\0' || (*utf8 == '\xc0' && *(utf8 + 1) == '\x80')) {
+               utf8 += (*utf8 == '\0') ? 1 : 2;
+            } else {
+               goto invalid_utf8;
+            }
```
joshbsiegel, line 332: Modifying the fuzz tester to allow valid UTF-8 null-terminated strings to be passed through to this function exposed another bug. The null character `0xC0 0x80` is a valid character according to `bson_utf8_validate`, but previously this function assumed that the only allowed null character is `0x0`. As such, it would return `NULL` when it came across `0xC0 0x80`. This change fixes this logic to account for `0xC0 0x80`.

Let me know if you'd prefer this change to occur in a separate PR.

src/libbson/tests/test-utf8.c
```diff
@@ -81,6 +85,47 @@ test_bson_utf8_escape_for_json (void)
    str = bson_utf8_escape_for_json (unescaped, -1);
    BSON_ASSERT (0 == memcmp (str, "\\u000e", 7));
    bson_free (str);
+
+   // Invalid UTF-8 strings (should return null)
+
```
eramongodb, line 90:
```suggestion
```
Comment is now stray. Suggest its removal (or scope as done below).

src/libbson/tests/test-utf8.c
```diff
@@ -81,6 +85,47 @@ test_bson_utf8_escape_for_json (void)
...
+      BSON_ASSERT (!str);
+
+      str = bson_utf8_escape_for_json ("\xf0\x9f", -1);
+      BSON_ASSERT (!str);
+
+      str = bson_utf8_escape_for_json ("\xf0\x9f\x9f", -1);
+      BSON_ASSERT (!str);
+
+      str = bson_utf8_escape_for_json ("\xf0\xa0\x9c\x8f", -1);
+      BSON_ASSERT (str);
```
eramongodb, line 126:
```suggestion
      ASSERT_CMPSTR (str, "𠜏");
```
Assert against the valid UTF-8 character being constructed.

src/libbson/tests/test-utf8.c
```diff
@@ -81,6 +85,47 @@ test_bson_utf8_escape_for_json (void)
...
+   // 3 bytes expected
+   {
+      str = bson_utf8_escape_for_json ("\xed", -1);
+      BSON_ASSERT (!str);
+
+      str = bson_utf8_escape_for_json ("\xed\x90", -1);
+      BSON_ASSERT (!str);
+
+      str = bson_utf8_escape_for_json ("\xe4\xb8\xad", -1);
+      BSON_ASSERT (str);
```
eramongodb, line 110:
```suggestion
      ASSERT_CMPSTR (str, "中");
```
Assert against the valid UTF-8 character being constructed.

src/libbson/tests/test-utf8.c
```diff
@@ -81,6 +85,47 @@ test_bson_utf8_escape_for_json (void)
...
+
+   // Invalid UTF-8 strings (should return null)
+
+   // 2 bytes expected
+   {
+      str = bson_utf8_escape_for_json ("\xc2", -1);
+      BSON_ASSERT (!str);
+
+      str = bson_utf8_escape_for_json ("\xc3\xa9", -1);
+      BSON_ASSERT (str);
```
eramongodb, line 97:
```suggestion
      ASSERT_CMPSTR (str, "é");
```
Assert against the valid UTF-8 character being constructed.

src/libbson/tests/test-utf8.c
```diff
@@ -66,6 +66,10 @@ test_bson_utf8_escape_for_json (void)
    BSON_ASSERT (0 == memcmp (str, "my\\u0000key", 7));
    bson_free (str);
 
+   str = bson_utf8_escape_for_json ("my\xc0\x80key", 7);
+   BSON_ASSERT (0 == memcmp (str, "my\\u0000key", strlen (str)));
```
eramongodb, line 70:
```suggestion
   ASSERT_CMPSTR (str, "my\\u0000key");
```
Use `ASSERT_CMPSTR` when comparing strings in tests (which includes approprate null checks).

Drive-by update to other assertions in this function/file to use `ASSERT_CMPSTR` is optional.

src/libbson/tests/test-utf8.c
```diff
@@ -81,6 +85,47 @@ test_bson_utf8_escape_for_json (void)
    str = bson_utf8_escape_for_json (unescaped, -1);
    BSON_ASSERT (0 == memcmp (str, "\\u000e", 7));
    bson_free (str);
+
+   // Invalid UTF-8 strings (should return null)
+
+   // 2 bytes expected
+   {
+      str = bson_utf8_escape_for_json ("\xc2", -1);
+      BSON_ASSERT (!str);
```
eramongodb, line 94:
```suggestion
      ASSERT (!str);
```
Use `ASSERT` in tests instead of `BSON_ASSERT`.

Drive-by update to other assertions in this function/file to use `ASSERT` is optional.

eramongodb (approved):
Minor suggestions remaining; otherwise, LGTM.

Note: there's a small chance that the suggestion to compare against UTF-8 string literals might fail depending on platform/compiler support. If it causes trouble, recommend moving it into the related comment instead, e.g.:

```c
// 4 bytes expected: 𠜏 (U+2070F)
{
   ...

   str = bson_utf8_escape_for_json ("\xf0\xa0\x9c\x8f", -1);
   ASSERT (str);
   bson_free (str);
}
```

kevinAlbs (approved):
Great find and fix.

PR #1727, opened 2024-09-16
CDRIVER-5719 fix build for macOS without pthread_threadid_np

barracuda156:
`pthread_threadid_np` is not present prior to 10.6 and it is also missing on `ppc` (including 10.6.8 Rosetta), because it is made x86-only in Apple libc.
This is the only issue to be fixed in order for the build to succeed.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-log.c
+++ b/src/libmongoc/src/mongoc/mongoc-log.c
@@ -21,6 +21,7 @@
 #include <process.h>
 #elif defined(__APPLE__)
 #include <pthread.h>
+#include <AvailabilityMacros.h>
 #elif defined(__FreeBSD__)
 #include <sys/thr.h>
 #elif defined(__NetBSD__)
@@ -197,9 +198,15 @@ mongoc_log_default_handler (mongoc_log_level_t log_level, const char *log_domain
 #elif defined(__NetBSD__)
    pid = (int) _lwp_self ();
 #elif defined(__APPLE__)
+#if MAC_OS_X_VERSION_MIN_REQUIRED < 1070
+   // libc on macOS < 10.7 does not support `pthread_threadid_np`
+   mach_port_t tid = pthread_mach_thread_np (pthread_self ());
+   pid = (int) tid;
+#else
    uint64_t tid;
    pthread_threadid_np (0, &tid);
    pid = (int) tid;
+#endif
 #else
    pid = (int) getpid ();
 #endif


src/libmongoc/src/mongoc/mongoc-log.c
```diff
@@ -198,7 +199,17 @@ mongoc_log_default_handler (mongoc_log_level_t log_level, const char *log_domain
    pid = (int) _lwp_self ();
 #elif defined(__APPLE__)
    uint64_t tid;
+#if MAC_OS_X_VERSION_MAX_ALLOWED < 1060 || defined(__ppc__)
```
kevinAlbs, line 202: macOS 10.7 appears to remove the restriction that `pthread_threadid_np` is defined only on x86. Compare the libc linked versions linked from https://opensource.apple.com/releases/.

[From 10.6.0](https://github.com/apple-oss-distributions/Libc/blob/Libc-583/pthreads/pthread.h#L475-L477):

```c
#if defined(__i386__) || defined(__x86_64__)
int		pthread_threadid_np(pthread_t,__uint64_t*);
#endif
```

[From 10.7.0](https://github.com/apple-oss-distributions/Libc/blob/Libc-763.11/pthreads/pthread.h#L481):
```c
int pthread_threadid_np(pthread_t,__uint64_t*) __OSX_AVAILABLE_STARTING(__MAC_10_6, __IPHONE_3_2);
```

To simplify, suggest always using `pthread_mach_thread_np` for if the min version of macOS <= 10.7:
```c
#if MAC_OS_X_VERSION_MIN_REQUIRED <= MAC_OS_X_VERSION_10_7
   // libc on macOS <= 10.7 does not support `pthread_threadid_np` for all architectures.
   mach_port_t tid = pthread_mach_thread_np (pthread_self ());
   pid = (int) tid;
#else
   uint64_t tid;
   pthread_threadid_np (0, &tid);
   pid = (int) tid;
#endif
```

kevinAlbs (commented):
Thank you for the PR @barracuda156.

Versions of macOS older than 11.0 are not tested or supported (see [mongo-c-driver Platform Support](https://www.mongodb.com/docs/languages/c/c-driver/current/libmongoc/ref/platforms/)). I prefer not to add complexity to support older versions. This PR seems like a small enough change to be acceptable.

Left a comment with a suggested change to the check.

src/libmongoc/src/mongoc/mongoc-log.c
barracuda156: @kevinAlbs Thank you. Sure, this can be done, `pthread_mach_thread_np` will work. I will rebase the PR accordingly.

barracuda156: @kevinAlbs I have updated the fix. On 10.7 we can use `pthread_threadid_np`, since there it is supported for all archs in fact, so < 10.7 works fine.
We need to use a numerical macro though, since `MAC_OS_X_VERSION_10_7` is not defined prior to 10.7, so the effect will otherwise be exactly the opposite of what is needed.

barracuda156: P. S. Do you know which header declares `mach_port_t` and whether it gets pulled in?

kevinAlbs (approved):
Thanks again @barracuda156. Merging.

src/libmongoc/src/mongoc/mongoc-log.c
kevinAlbs: On 10.6, it appears to be defined in the [pthread.h header](https://github.com/apple-oss-distributions/Libc/blob/36e20149daacb6cfafb07c68636d9c33b5509331/pthreads/pthread.h#L114-L117).

barracuda156: Great, then all good. Thank you for merging!

PR #1730, opened 2024-09-18
Update NEWS and release steps post-1.28.0

kevinAlbs:
**Summary**
- Update the NEWS file with 1.28.0 release notes.
- Update the release process.

**Details**

The"Create a New Clone" step was relocated before the "Start Snyk Monitoring" step so the referenced `$RELEASE_BRANCH` is created. Relocating the step resulted in an error from Sphinx:

```
Warning, treated as error:
/Users/kevin.albertson/code/mongo-c-driver-release-1.28.0/docs/dev/releasing.rst:101:'any' reference target not found: $RELEASE_CLONE
```

I was unable to determine the cause. This PR changes to use two backticks.


docs/dev/releasing.rst
```diff
@@ -95,6 +95,38 @@ Check that the `etc/purls.txt` file is up-to-date with the set of
 :term:`vendored dependencies <vendored dependency>`. If any items need to be
 updated, refer to `sbom-lite-updating`.
 
+Create a New Clone of ``mongo-c-driver``
+****************************************
+
+To prevent publishing unwanted changes and to preserve local changes, create a
+fresh clone of the C driver. We will clone into a new arbitrary directory which
+we will refer to as ``$RELEASE_CLONE`` ::
```
vector-of-bool, line 103: For future reference: The reason that this requires two backticks is because you moved it above the `.. default-role:: bash` line. The `default-role` changes the meaning of the single backticks when they appear without an inline role tag (i.e. ``:some-role:`my-text` `` has an explicit `:some-role:` role). In this scope, the default role is `:any:` (set in `conf.py`), which will attempt to generate a hyperlink reference to an arbitrary object.

The double-backticks is equivalent to the `:literal:` role, which just adds a monospace font to the text. The `:bash:` role simply adds syntax highlighting within the annotated text (it is a custom role we add in the `rst_prolog` in `conf.py`).

See: https://docutils.sourceforge.io/docs/ref/rst/directives.html#default-role and https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-default_role

vector-of-bool (approved):
LGTM!

docs/dev/releasing.rst
kevinAlbs: TIL thank you for the explanation.

PR #1732, opened 2024-09-19
CDRIVER-4740 bson_utf8_escape_for_json performance improvements

joshbsiegel:
# Summary
This PR optimizes `bson_utf8_escape_for_json` to reduce performance problems in `bson_as_json`.

# Motivation
Credit to @alcaeus for the initial find and [draft PR](https://github.com/mongodb/mongo-c-driver/pull/1419). Much of the work in this PR is adapted from his changes. To summarize, `bson_as_json` was performing about 5x as slow when compared to the PHP driver's equivalent function. It was recognized that `bson_utf8_escape_for_json` was the main reason for this.

# Background
In order to convert a UTF-8 string to JSON, some characters require [special treatment](https://www.json.org/json-en.html). For example, double quotes `"` must be converted into escaped double quotes `\"`. See [this diagram](https://www.json.org/img/string.png) for all of the necessary conversions. As such, when converting to JSON, strings must be parsed character by character to ensure that the resulting string is valid for JSON.

# `_is_special_char()`
Previously, `bson_utf8_escape_for_json` was checking, converting, and copying over each character one-by-one. Considering that most characters do not require any conversion, a bit mask is now used to see if a character is "special" or not. If not, we simply continue to iterate. When a special character is reached (or end of string), we append of ALL of the normal characters seen up to that point. This reduces the amount of appends to the string, improving performance. `bson_string_append_ex` was added as a helper for this piece-by-piece strategy.

# `bson_string_alloc()`
Previously, an empty string was created in `bson_utf8_escape_for_json` and then appended to in order to produce the JSON-safe string. Being that we can guarantee that the string produced from this function will be at least the size of the string passed in, this is wasteful due to the cost of growing the string from size 0. To fix this, a helper function `bson_string_alloc` was created. This reduces the amount of reallocations needed as we now create a initial string with the minimal known size.

# Testing
Unit tests and docs were added for the new public API `bson_string_alloc` and  `bson_string_append_ex`. In addition to that and running the `utf8` tests, the fuzz tester for `bson_utf8_escape_for_json` successfully ran without issues, which was the initial reason for backlogging the [draft PR](https://github.com/mongodb/mongo-c-driver/pull/1419). This difference in behavior is likely due to the fix from [this](https://github.com/mongodb/mongo-c-driver/pull/1725) PR.

# Performance Improvements
To test performance, [these](https://github.com/kevinAlbs/c-bootstrap/tree/master/investigations/profileme) profiling steps were followed using both a `bson_utf8_escape_for_json` [benchmark](https://gist.github.com/joshbsiegel/cd783aed9e80481c8a2fa2b7766399eb) and a `bson_as_json` [benchmark](https://gist.github.com/joshbsiegel/61b2847b34cc22cee20510dde20816af).

| Benchmark    | Time Before (s) |  Time After (s) | Improvement | 
| :------------: | :-------: | :-------: | :-------: |
| `bson_utf8_escape_for_json`  | 35.94  bson_utf8_escape_for_json before  | 13.16 bson_utf8_escape_for_json after | 63.38%
| `bson_as_json` | 56.04 bson_as_json_before     | 35.54 bson_as_json_after | 36.58%

# What's New
- `bson_string_alloc()` implementation, documentation, and tests
- `bson_string_append_ex()` implementation, documentation, and tests
- `bson_utf8_escape_for_json()` optimizations (see above) and refactor


src/libbson/src/bson/bson-string.c
```diff
@@ -145,9 +182,9 @@ bson_string_free (bson_string_t *string, /* IN */
 /*
  *--------------------------------------------------------------------------
  *
- * bson_string_append --
+ * bson_string_append_ex --
  *
- *       Append the UTF-8 string @str to @string.
+ *       Append the UTF-8 string @str of given length @len to @string.
```
joshbsiegel, line 187: The behavior when `len` > `strlen(str)` is undefined but doesn't crash. I don't think we want to add an assertion because that will require an `O(n)` check.

src/libbson/doc/bson_string_alloc.rst
eramongodb, line 1: In light of CDRIVER-5697 and CDRIVER-5698, recommend moving `bson_string_alloc` into an internal header and avoid exporting it in the ABI (for internal use only).

src/libbson/src/bson/bson-string.c
```diff
@@ -116,7 +116,44 @@ bson_string_new (const char *str) /* IN */
...
+ *       bson_string_free().
+ *
+ * Side effects:
+ *       None.
+ *
+ *--------------------------------------------------------------------------
+ */
+
+bson_string_t *
+bson_string_alloc (const uint32_t size)
```
eramongodb, line 146:
```suggestion
bson_string_alloc (const size_t size)
```
Use `size_t` in the parameter (as done for `bson_string_append_ex`) and assert as a documented precondition the given value is less than `UINT32_MAX`. This will address most of the new `-Wsign-conversion` and `-Wshorten-64-to-32` warnings in this PR.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -264,77 +376,104 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
...
    }
 
+   if (utf8_len == 0) {
+      return bson_strdup ("");
+   }
+
    end = utf8 + utf8_len;
 
-   while (utf8 < end) {
+   str = bson_string_alloc (utf8_len);
```
eramongodb, line 395:
```suggestion
   str = bson_string_alloc (utf8_len);
```
Consider using a local `size_t utf8_ulen` variable after performing a bounds check on the result of the `strlen (utf8)` call above. This will address most of the new `-Wsign-conversion` warnings in this PR.

See other comment regarding `size_t` parameter for `bson_string_alloc`.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -264,77 +376,104 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
...
-            }
-         } else {
-            goto invalid_utf8;
-         }
+         goto invalid_utf8;
       }
-   }
+
+      utf8_len--;
+   } while (utf8_len);
```
eramongodb, line 476:
```suggestion
   } while (utf8_len > 0);
```
Avoid treating integers as booleans.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -264,77 +376,104 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
...
-   while (utf8 < end) {
+   str = bson_string_alloc (utf8_len);
+
+   pos = 0;
+
+   do {
       uint8_t mask;
       uint8_t length_of_char;
 
+      c = (unsigned char) utf8[pos];
```
eramongodb, line 403:
```suggestion
const uint8_t current_byte = (uint8_t) utf8[pos];
```
A separate variable of type `uint8_t` may be required to distinguish it from the UTF-8 character `bson_unichar_t c` when used for single-byte operations. `c` as a `bson_unichar_t` isn't used until the `// Unicode character` block below by the call to `bson_utf8_get_char`.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -264,77 +376,104 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
...
    end = utf8 + utf8_len;
 
-   while (utf8 < end) {
+   str = bson_string_alloc (utf8_len);
+
+   pos = 0;
+
+   do {
       uint8_t mask;
       uint8_t length_of_char;
```
eramongodb, line 401: Reduce scope of local variables.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -230,6 +230,118 @@ bson_utf8_validate (const char *utf8, /* IN */
...
+                                             0x00000004, // double quote "
+                                             0x10000000, // backslash
+                                             0x00000000,
+                                             0xffffffff,
+                                             0xffffffff,
+                                             0xffffffff,
+                                             0xffffffff}; // non-ASCII
+   const int int_index = c / (sizeof (bson_unichar_t) * 8);
+   const int bit_index = c & (sizeof (bson_unichar_t) * 8 - 1);
+   return (charmap[int_index] >> bit_index) & 1;
```
eramongodb, line 271:
```suggestion
   const int int_index = ((int) c) / ((int) sizeof (bson_unichar_t) * 8);
   const int bit_index = ((int) c) & ((int) sizeof (bson_unichar_t) * 8 - 1);
   return ((charmap[int_index] >> bit_index) & ((bson_unichar_t) 1)) != 0u;
```
Signedness and type consistency.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -230,6 +230,118 @@ bson_utf8_validate (const char *utf8, /* IN */
...
+ *       None.
+ *
+ *--------------------------------------------------------------------------
+ */
+
+static BSON_INLINE void
+_bson_utf8_handle_special_char (bson_unichar_t c,   /* IN */
+                                bson_string_t *str) /* OUT */
+{
+   BSON_ASSERT (c < 0x80);
```
eramongodb, line 303:
```suggestion
   BSON_ASSERT (c < 0x80u);
```
Signedness consistency.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -230,6 +230,118 @@ bson_utf8_validate (const char *utf8, /* IN */
...
+      break;
+   case '\r':
+      bson_string_append (str, "\\r");
+      break;
+   case '\t':
+      bson_string_append (str, "\\t");
+      break;
+   default: {
+      // ASCII control character
+      BSON_ASSERT (c < 0x20);
```
eramongodb, line 330:
```suggestion
      BSON_ASSERT (c < 0x20u);
```
Signedness consistency.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -230,6 +230,118 @@ bson_utf8_validate (const char *utf8, /* IN */
...
+
+      const char digits[] = "0123456789abcdef";
+      char codepoint[6] = "\\u0000";
+
+      codepoint[2] = digits[(c >> 12) & 0xf];
+      codepoint[3] = digits[(c >> 8) & 0xf];
+      codepoint[4] = digits[(c >> 4) & 0xf];
+      codepoint[5] = digits[c & 0xf];
```
eramongodb, line 338:
```suggestion
      codepoint[2] = digits[(c >> 12) & 0x0fu];
      codepoint[3] = digits[(c >> 8) & 0x0fu];
      codepoint[4] = digits[(c >> 4) & 0x0fu];
      codepoint[5] = digits[c & 0x0fu];
```
Signedness consistency.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -230,6 +230,118 @@ bson_utf8_validate (const char *utf8, /* IN */
...
+
+      bson_string_append_ex (str, codepoint, 6);
```
eramongodb, line 340:
```suggestion
      bson_string_append_ex (str, codepoint, sizeof (codepoint));
```
Reduce magic numbers.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -264,77 +376,104 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
...
+      // bson_utf8_get_char will crash without this check
       _bson_utf8_get_sequence (utf8, &length_of_char, &mask);
       if (utf8 > end - length_of_char) {
          goto invalid_utf8;
       }
 
-      c = bson_utf8_get_char (utf8);
+      // Check for null character
+      // Null characters are only allowed if the length is provided
+      if (!*utf8 || (*utf8 == '\xc0' && *(utf8 + 1) == '\x80')) {
```
eramongodb, line 434:
```suggestion
      if (utf8[0] == '\0' || (utf8[0] == '\xc0' && utf8[1] == '\x80')) {
```
Use subscript syntax instead of pointer arithmetic + avoid treating integers as booleans.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -264,77 +376,104 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
...
+   if (utf8_len == 0) {
+      return bson_strdup ("");
+   }
+
    end = utf8 + utf8_len;
 
-   while (utf8 < end) {
+   str = bson_string_alloc (utf8_len);
+
+   pos = 0;
```
eramongodb, line 397:
```suggestion
   size_t pos = 0u;
```
Reduce scope of local variables.

Recommend renaming the variable or adding a comment documenting the purpose of this variable, not as a position of the _entire_ string being escaped, but for buffered write position tracking.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -264,77 +376,104 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
...
+            bson_string_append_ex (str, utf8, pos);
+            break;
+         }
+
+         continue;
+      }
+
+      // Reached a special character. Copy over all of normal characters
+      // we have passed so far
+      if (pos) {
```
eramongodb, line 419:
```suggestion
      if (pos > 0) {
```
Avoid treating integers as booleans.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -264,77 +376,104 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
...
-         } else {
-            bson_string_append_unichar (str, c);
+         bson_string_append (str, "\\u0000");
+         utf8_len -= *utf8 ? 2 : 1;
+         utf8 += *utf8 ? 2 : 1;
+         continue;
+      }
+
+      // Unicode character
+      if (c > 0x7f /* highest ASCII character */) {
```
eramongodb, line 446:
```suggestion
      // Highest ASCII character.
      if (c > 0x7fu) {
```
Signedness consistency + should the block scope comment be updated?

eramongodb (changes requested):

src/libbson/src/bson/bson-string.h
```diff
@@ -42,6 +42,8 @@ bson_string_new (const char *str);
 BSON_EXPORT (char *)
 bson_string_free (bson_string_t *string, bool free_segment);
 BSON_EXPORT (void)
+bson_string_append_ex (bson_string_t *string, const char *str, const size_t len);
```
eramongodb, line 45: Also move to `bson-string-private.h`.

src/libbson/src/bson/bson-string.c
eramongodb: It is enough for `len <= strlen(str)` to be an unasserted precondition of the function (as is currently written, "of given length `@len`"). A more featureful contracts API would allow for distinguishing "cheap" vs. "expensive" assertions, but we do not have such an API. It is fine to leave it to ASan to detect violations for now.

src/libbson/src/bson/bson-string.c
```diff
@@ -116,7 +116,46 @@ bson_string_new (const char *str) /* IN */
...
+ * Side effects:
+ *       None.
+ *
+ *--------------------------------------------------------------------------
+ */
+
+bson_string_t *
+_bson_string_alloc (const size_t size)
+{
+   BSON_ASSERT (bson_in_range_unsigned (uint32_t, size));
```
eramongodb, line 148:
```suggestion
   BSON_ASSERT (size < UINT32_MAX);
```

`bson_string_ensure_space` asserts `BSON_ASSERT (needed <= UINT32_MAX - 1u);`. Reuse that assertion here to leave room for guaranteed null terminator.

src/libbson/src/bson/bson-string.c
```diff
@@ -159,24 +198,48 @@ bson_string_free (bson_string_t *string, /* IN */
...
+ *       None.
+ *
+ *--------------------------------------------------------------------------
+ */
+
+void
+bson_string_append (bson_string_t *string, /* IN */
+                    const char *str)       /* IN */
+{
+   bson_string_append_ex (string, str, strlen (str));
```
eramongodb, line 239:
```suggestion
   BSON_ASSERT_PARAM (string);
   BSON_ASSERT_PARAM (str);
   bson_string_append_ex (string, str, strlen (str));
```
Assert preconditions + guard call to `strlen` (null argument is UB).

src/libbson/src/bson/bson-string.c
eramongodb, line 1: Include `<bson/bson-string-private.h>`.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -373,41 +372,37 @@ char *
...
+      utf8_ulen = (size_t) utf8_len;
    }
 
-   if (utf8_len == 0) {
+   if (utf8_ulen == 0) {
       return bson_strdup ("");
    }
 
-   end = utf8 + utf8_len;
+   const char *end = utf8 + utf8_ulen;
```
eramongodb, line 391:
```suggestion
   const char *const end = utf8 + utf8_ulen;
```
Const correctness.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -373,41 +372,37 @@ char *
...
 
-   str = bson_string_alloc (utf8_len);
+   bson_string_t *str = _bson_string_alloc (utf8_ulen);
```
eramongodb, line 393:
```suggestion
   bson_string_t *const str = _bson_string_alloc (utf8_ulen);
```
Const correctness.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -327,17 +328,15 @@ _bson_utf8_handle_special_char (bson_unichar_t c,   /* IN */
...
    default: {
       // ASCII control character
-      BSON_ASSERT (c < 0x20);
+      BSON_ASSERT (c < 0x20u);
 
       const char digits[] = "0123456789abcdef";
       char codepoint[6] = "\\u0000";
 
-      codepoint[2] = digits[(c >> 12) & 0xf];
-      codepoint[3] = digits[(c >> 8) & 0xf];
```
eramongodb: Were these deliberately removed?

src/libbson/src/bson/bson-utf8.c
```diff
@@ -416,64 +411,69 @@ bson_utf8_escape_for_json (const char *utf8, /* IN */
...
       // Null characters are only allowed if the length is provided
-      if (!*utf8 || (*utf8 == '\xc0' && *(utf8 + 1) == '\x80')) {
+      if (utf8[0] == '\0' || (utf8[0] == '\xc0' && utf8[1] == '\x80')) {
          if (!length_provided) {
             goto invalid_utf8;
          }
 
          bson_string_append (str, "\\u0000");
-         utf8_len -= *utf8 ? 2 : 1;
+         utf8_ulen -= *utf8 ? 2 : 1;
```
eramongodb, line 440:
```suggestion
         utf8_ulen -= *utf8 ? 2u : 1u;
```
Sign consistency.

src/libbson/tests/test-string.c
```diff
@@ -356,9 +357,73 @@ test_bson_string_capacity (void *unused)
       large_str[UINT32_MAX - 2u] = 's'; // Restore.
    }
 
+   // Test allocating the largest possible string.
+   {
+      bson_string_t *str = _bson_string_alloc (2147483648);
```
eramongodb, line 362:
```suggestion
      bson_string_t *str = _bson_string_alloc (UINT32_MAX - 1u);
```
Reduce magic numbers.

src/libbson/tests/test-string.c
```diff
@@ -356,9 +357,73 @@ test_bson_string_capacity (void *unused)
...
+test_bson_string_append_ex (void)
+{
+   bson_string_t *str;
+   char *s;
+
+   str = bson_string_new (NULL);
+   bson_string_append_ex (str, "the quick brown fox jumps over the lazy dog", 10);
+   s = bson_string_free (str, false);
+   BSON_ASSERT (s);
+   BSON_ASSERT (!strcmp (s, "the quick "));
```
eramongodb, line 381:
```suggestion
   ASSERT_CMPSTR (s, "the quick ");
```
Use `ASSERT` and `ASSERT_CMPSTR` in test code.

eramongodb (changes requested):
Rename of `pos -> normal_chars_seen` and use of `current_byte` vs. `unichar` is a great readability improvement.

src/libbson/src/bson/bson-utf8.c
joshbsiegel: Sorry I forgot to add a comment about this. At this block of code `c` is between 0x01 and 0x1f (see `BSON_ASSERT (c < 0x20u);` above). So the maximum value of `c` is 0b00011111. Shifting `c` 8 or 12 bits to the right is therefore unneeded as `digits[2]` and `digits[3]` should always be 0.

src/libbson/tests/test-string.c
```diff
@@ -356,9 +357,70 @@ test_bson_string_capacity (void *unused)
       large_str[UINT32_MAX - 2u] = 's'; // Restore.
    }
 
+   // Test allocating the largest possible string.
+   {
+      bson_string_t *str = _bson_string_alloc (UINT32_MAX - 4u);
```
eramongodb, line 362:
```suggestion
      bson_string_t *str = _bson_string_alloc (UINT32_MAX - 1u);
```
Should this be `1u` (for null terminator)?

src/libbson/tests/test-string.c
```diff
@@ -356,9 +357,70 @@ test_bson_string_capacity (void *unused)
...
+test_bson_string_append_ex (void)
+{
+   bson_string_t *str;
+   char *s;
+
+   str = bson_string_new (NULL);
+   _bson_string_append_ex (str, "the quick brown fox jumps over the lazy dog", 10);
+   s = bson_string_free (str, false);
+   ASSERT_CMPSTR (s, "the quick ");
+   bson_free (s);
```
eramongodb, line 381:
```suggestion
   s = bson_string_free (str, false);
   ASSERT_CMPSTR (str->str, "the quick ");
   bson_string_free (str, true);
```
Simplification. Local variable `s` is not needed.

eramongodb (approved):

src/libbson/src/bson/bson-utf8.c
```diff
@@ -261,80 +372,108 @@ char *
...
-         if (length_provided) {
-            // Check if current character is null character,
-            // if so skip past it as we've already added '\\u0000' above
-            if (*utf8 == '\0' || (*utf8 == '\xc0' && *(utf8 + 1) == '\x80')) {
-               utf8 += (*utf8 == '\0') ? 1 : 2;
-            } else {
-               goto invalid_utf8;
-            }
-         } else {
+      // Highest ASCII character
```
kevinAlbs, line 445:
```suggestion
      // Multi-byte UTF-8 sequence
```

I expect this block only applies to multi-byte UTF-8 sequences.

src/libbson/src/bson/bson-utf8.c
```diff
@@ -230,6 +231,116 @@ bson_utf8_validate (const char *utf8, /* IN */
...
+/*
+ *--------------------------------------------------------------------------
+ *
+ * _bson_utf8_handle_special_char --
+ *
+ *       Appends a special character in the correct format when converting
+ *       from UTF-8 to JSON. This includes characters that should be escaped
+ *       as well as ASCII control characters.
+ *
+ *       Normal ASCII characters are handled in bson_utf8_escape_for_json,
```
kevinAlbs, line 284:
```suggestion
 *       Normal ASCII characters and multi-byte UTF-8 sequences are handled in bson_utf8_escape_for_json,
```

Suggest noting this function does not apply to multi-byte UTF-8 sequences. `_is_special_char` returns true for the first byte of a multi-byte UTF-8 sequence, but multi-byte UTF-8 sequences are not expected to be passed to `_bson_utf8_handle_special_char`

src/libbson/src/bson/bson-utf8.c
```diff
@@ -230,6 +231,116 @@ bson_utf8_validate (const char *utf8, /* IN */
...
+ *       where this function is called from.
+ *
+ * Parameters:
+ *       @c: A bson_unichar_t.
```
kevinAlbs, line 288:
```suggestion
 *       @c: A uint8_t ASCII codepoint.
```

kevinAlbs (approved):
Fantastic. I found the comments and organization helpful for understanding. LGTM with very minor comments.

PR #1736, opened 2024-09-23
CDRIVER-5733 fix behavior of `bson_string_truncate`

kevinAlbs:
Fixes behavior of `bson_string_truncate` with large lengths. Verified with [this patch build](https://spruce.mongodb.com/version/66f186c226f05a00072fb0a4). See CDRIVER-5733 for details.

Also adds tests for behavior of `bson_string_truncate`. In the process of this fix, other unexpected behavior was observed: `bson_string_truncate` [documents](https://mongoc.org/libbson/1.28.0/bson_string_truncate.html) the truncated length "must be smaller or equal to the current length of the string.". However, `bson_string_truncate` does not reject greater lengths.
```c
bson_string_t *str = bson_string_new ("a");
bson_string_truncate (str, 2u); // Is not rejected.
ASSERT_CMPUINT32 (str->len, ==, 2u);
ASSERT_CMPUINT32 (str->alloc, ==, 4u);
bson_string_free (str, true);
```
This behavior is preserved (and now tested) for backwards compatibility. I expect there may be a possible use-case for calling `bson_string_truncate` with larger lengths to reserve space.

### DIFF
--- a/src/libbson/doc/bson_string_truncate.rst
+++ b/src/libbson/doc/bson_string_truncate.rst
@@ -24,4 +24,4 @@ Truncates the string so that it is ``len`` bytes in length. This must be smaller
 
 A ``\0`` byte will be placed where the end of the string occurs.
 
-.. warning:: This function will abort if the length of the resulting string (including the NULL terminator) would exceed ``UINT32_MAX``.
+.. warning:: The length of the resulting string (including the ``NULL`` terminator) MUST NOT exceed ``UINT32_MAX``.
--- a/src/libbson/doc/bson_string_truncate.rst
+++ b/src/libbson/doc/bson_string_truncate.rst
@@ -24,3 +24,4 @@ Truncates the string so that it is ``len`` bytes in length. This must be smaller
 
 A ``\0`` byte will be placed where the end of the string occurs.
 
+.. warning:: This function will abort if the length of the resulting string (including the NULL terminator) would exceed ``UINT32_MAX``.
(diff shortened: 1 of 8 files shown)


src/libbson/doc/bson_string_truncate.rst
```diff
@@ -24,3 +24,4 @@ Truncates the string so that it is ``len`` bytes in length. This must be smaller
 
 A ``\0`` byte will be placed where the end of the string occurs.
 
+.. warning:: This function will abort if the length of the resulting string (including the NULL terminator) would exceed ``UINT32_MAX``.
```
eramongodb, line 27:
```suggestion
.. warning:: The length of the resulting string (including the ``NULL`` terminator) MUST NOT exceed ``UINT32_MAX``.
```
Wording suggestion.

src/libbson/src/bson/bson-string.c
```diff
@@ -307,19 +307,17 @@ void
...
-      alloc = 16;
+   BSON_ASSERT_PARAM (string);
+   if (len == string->len) {
+      return;
    }
-
-   if (!bson_is_power_of_two (alloc)) {
-      alloc = (uint32_t) bson_next_power_of_two ((size_t) alloc);
+   uint32_t needed = len;
+   BSON_ASSERT (needed <= UINT32_MAX - 1u);
```
eramongodb, line 315:
```suggestion
   BSON_ASSERT (needed < UINT32_MAX);
```

eramongodb (approved):

src/libbson/doc/bson_string_truncate.rst
kevinAlbs: Applied suggestion to other `bson_string_append*` functions with the same warning.

rcsanchez97 (approved):

rcsanchez97 (approved):

PR #1739, opened 2024-09-24
CDRIVER-5715 deprecate defining `BSON_MEMCHECK`

kevinAlbs:
Deprecate defining `BSON_MEMCHECK` when building libbson. See CDRIVER-5715 for rationale.

When building with the C flag `-DBSON_MEMCHECK=1`, a warning is now produced:
```
/Users/kevin.albertson/code/tasks/mongo-c-driver-2.0/src/libbson/src/bson/bson.c:31:9: warning: Do not define BSON_MEMCHECK. BSON_MEMCHECK changes the data layout of bson_t. BSON_MEMCHECK is deprecated may be removed in a future major release [-W#pragma-messages]
   31 | #pragma message( \
      |         ^
1 warning generated.
```

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,10 @@
 libmongoc 1.29.0 (unreleased)
 =============================
 
+Deprecated:
+
+  * Compiling with `BSON_MEMCHECK` defined is deprecated.
+
 Platform Support:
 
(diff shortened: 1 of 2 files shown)


vector-of-bool (approved):
LGTM

PR #1740, opened 2024-09-24
CDRIVER-5734 assert length in bson_value_copy to integer overflow

joshbsiegel:
# Summary
This PR will add asserts to `bson_value_copy` to check the length of `src` before allocating the memory.

# Background
See [CDRIVER-5734](https://jira.mongodb.org/browse/CDRIVER-5734). Integer overflow was occurring in certain cases when allocating memory for both the value and null terminator.

# Testing
The changes were compiled and ran with [this](https://gist.github.com/joshbsiegel/a94db6d276d388429ff5e97452d8ac06) program to confirm that the new asserts were failing instead of a segmentation fault.

### DIFF
--- a/src/libbson/src/bson/bson-value.c
+++ b/src/libbson/src/bson/bson-value.c
@@ -35,6 +35,7 @@ bson_value_copy (const bson_value_t *src, /* IN */
       dst->value.v_double = src->value.v_double;
       break;
    case BSON_TYPE_UTF8:
+      BSON_ASSERT (src->value.v_utf8.len <= UINT32_MAX - 1);
       dst->value.v_utf8.len = src->value.v_utf8.len;
       dst->value.v_utf8.str = bson_malloc (src->value.v_utf8.len + 1);
       memcpy (dst->value.v_utf8.str, src->value.v_utf8.str, dst->value.v_utf8.len);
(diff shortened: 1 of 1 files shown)


src/libbson/src/bson/bson-value.c
```diff
@@ -35,6 +35,7 @@ bson_value_copy (const bson_value_t *src, /* IN */
       dst->value.v_double = src->value.v_double;
       break;
    case BSON_TYPE_UTF8:
+      BSON_ASSERT (src->value.v_utf8.len <= UINT32_MAX - 1);
```
kevinAlbs, line 38: `bson_malloc` accepts a `size_t`, so the resulting value is already cast to a `size_t`. I suggest casting to `size_t` first:

```c
BSON_ASSERT (bson_in_range_size_t_unsigned (src->value.v_utf8.len));
size_t len_sz = (size_t) src->value.v_utf8.len;
BSON_ASSERT (len_sz <= SIZE_MAX - 1);
dst->value.v_utf8.len = src->value.v_utf8.len;
dst->value.v_utf8.str = bson_malloc (len_sz + 1);
```

On most 64-bit systems, I expect `size_t` has range greater than `uint32_t`. That may make a failed assert less likely.

Suggestion requires `#include <bson/bson-cmp.h>` for `bson_in_range_size_t_unsigned`.

kevinAlbs (changes requested):

src/libbson/src/bson/bson-value.c
joshbsiegel: Completely missed that `bson_malloc` takes in a `size_t`. This is definitely a better solution to prevent failed asserts.

kevinAlbs (approved):

PR #1742, opened 2024-10-01
CDRIVER-5740 Use run-orchestration.sh

eramongodb:
## Summary

This PR resolves CDRIVER-5740. Verified by [this patch](https://spruce.mongodb.com/version/66fb1cdda45ff700074a7d13/).

Fixes the purple wall currently blocking C Driver EVG tasks.

Observed that recent updates to DET scripts (in order to keep up with updates to mongo-orchestration) were frequently breaking C Driver EVG tasks, but not C++ Driver EVG tasks. The difference appeared to be due to the C++ Driver EVG scripts using `run-orchestration.sh` to start mongo-orchestration (as intended/expected by DET scripts), whereas the C Driver is using its own manual implementation of the same initialization routines.

This PR updates `integration-tests.sh` to use `run-orchestration.sh` to better keep up with DET script updates in a compatible manner.

## run-orchestration.sh

Using the C++ Driver's implementation as reference, `run-orchestration.sh` has been updated as follows:

- `MONGO_ORCHESTRATION_HOME`: defined as `"$DRIVERS_TOOLS/.evergreen/orchestration"` (instead of `C:\data\MO` or `$(pwd)/MO` depending on the distro) for consistency with DET scripts and the C++ Driver.
    - All subsequent path handling of the MO home directory are left to `run-orchestration.sh`.
- `ORCHESTRATION_FILE`: defined by `run-orchestration.sh` except when an explicit config is required (i.e. versioned API tasks, Auth AWS tasks, and `ocsp-*` tasks).
    - The `.json` file extension needed to be included for compatibility.
    - Config files are under `$DRIVERS_TOOLS/.evergreen/orchestration` (the new `MONGO_ORCHESTRATION_HOME`).
- `MONGODB_BINARIES`: defined as `$DRIVERS_TOOLS/mongodb/bin` for consistency with DET scripts and the C++ Driver.
    - This path also provides crypt_shared and mongocryptd. `run-tests.sh` needed to be updated accordingly.

## Versioned API Tests

A currently-unaddressed issue with `run-orchestration.sh` (which is blocked on a fix to mongo-orchestration) does not allow testing Stable API with SSL/TLS enabled in mongo-orchestration due to lack of a `tls` URI option in the connection string.

This PR uses a workaround that adds `--tls --tlsAllowInvalidCertificates` to the `mongosh` command used to set `"requireApiVersion": 1` in `run-orchestration.sh`. This should hopefully be made unnecessary in the near future.

The `test-versioned-api-*` tasks are updated to use a build with SSL/TLS support enabled, but does not utilize these features during test execution. Using `AUTH` without `SSL` is a bit unusual for C Driver tests, so comments were added to the task definition accordingly.

## Other Notes

- The `AUTHSOURCE` env var appears to be obsolete and was therefore removed.
- Removed references to the non-existent `snappy-zlib-zstd` orchestration config (logs indicate the `basic` config file was being used instead...?).
- Added `--max-time` to the MO wait routine's curl command.
- The single task failure in the veryfing patch appears to be due to `mongocryptd --fork`, but it is unclear to me what the root cause may be. It appears to be a bit flaky.


eramongodb:
The `--tls --tlsAllowInvalidCertificates` workaround appears to be sufficient to support SSL/TLS testing by the C Driver as verified by [this patch](https://spruce.mongodb.com/version/66fc1217014db40007d17b04/). Reverted the task definition changes to their original state.

eramongodb:
Latest patch reveals Ubuntu 18.04 still has some issues with killing existing processes occupying server port(s). This issue should eventually be addressed by DET scripts (not us).

.evergreen/scripts/integration-tests.sh
```diff
@@ -7,11 +7,9 @@
 # MONGODB_VERSION: latest, 4.2, 4.0
 # TOPOLOGY: server, replica_set, sharded_cluster
 # AUTH: auth, noauth
-# AUTHSOURCE
 # IPV4_ONLY: off, on
 # SSL: openssl, darwinssl, winssl, nossl
-# ORCHESTRATION_FILE: <file name in orchestration_configs/ without .json>
-#   If this is not set, the file name is constructed from other options.
+# ORCHESTRATION_FILE: <file name in DET configs/${TOPOLOGY}s/>
```
kevinAlbs, line 12: The C driver contains orchestration configs that are not in DET:
- `orchestration_configs/servers/basic-ipv4-only.json`
- `orchestration_configs/servers/snappy.json`
- `orchestration_configs/servers/zstd.json`

On further look, I do not the extra configs provide much meaningful test coverage.

`snappy` and `zstd` may not be relevant for modern servers. The default [enables all three](https://www.mongodb.com/docs/manual/reference/program/mongod/#std-option-mongod.--networkMessageCompressors).

Tasks specifying `IPV4_ONLY` (to use `basic-ipv4-only`) also [specify the URI `mongodb://127.0.0.1`](https://github.com/mongodb/mongo-c-driver/blob/56cbe883be2fe84812e35ea8fae181dc7e0be8ce/.evergreen/generated_configs/legacy-config.yml#L1706-L1707), so I expect those tasks connect with IPv4 without extra server configuration.

Suggest:
- Remove C driver `orchestration_configs`.
- Remove mention of `IPV4_ONLY`.

.evergreen/scripts/integration-tests.sh
```diff
@@ -7,11 +7,9 @@
...
 # OCSP: off, on
```
kevinAlbs, line 13: Suggest removing this comment block. The OCSP variable is no-longer-used. `run-orchestration.sh` [documents supported environment variables](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/a69349171a454a12d9c584d55ace8f00cef49ab6/.evergreen/run-orchestration.sh#L5-L19).

kevinAlbs (commented):
Exciting! This has been a long-standing want (recalling CDRIVER-2707).

.evergreen/scripts/integration-tests.sh
eramongodb: Thank you. Done as suggested.

I preserved the `MONGOC_CHECK_IPV6` env var (formerly conditioned on `"$IPV4_ONLY" != "on"`; now unconditionally defined in EVG scripts) in case there are downstream users that still depend on it. LMK if you would like this removed from the test suite as well.

.evergreen/scripts/integration-tests.sh
eramongodb: Done.

eramongodb:
Latest changes verified by [this patch](https://spruce.mongodb.com/version/66fc28525ef8c0000768c49c/).

.evergreen/scripts/integration-tests.sh
kevinAlbs: I slightly prefer removing `MONGOC_CHECK_IPV6`. I do not think preserving test environment variables is needed, and I expect removing `MONGOC_CHECK_IPV6` would not be disruptive.

kevinAlbs (approved):

.evergreen/scripts/integration-tests.sh
eramongodb: Done.

PR #1743, opened 2024-10-01
CDRIVER-5740 fix fprintf call

kevinAlbs:
Minor follow-up to https://github.com/mongodb/mongo-c-driver/pull/1742 to address:

```
../../../src/libmongoc/tests/TestSuite.c: In function 'TestSuite_PrintJsonHeader':
../../../src/libmongoc/tests/TestSuite.c:862:13: error: format '%s' expects a matching 'char *' argument [-Werror=format=]
             getenv_or ("MONGODB_API_VERSION", "null"));
             ^
```


eramongodb (approved):
Whoops. Thank you.

PR #1748, opened 2024-10-02
Update Debian packaging instructions

rcsanchez97:


docs/dev/debian.rst
```diff
@@ -25,40 +25,53 @@ To publish a new release Debian package, perform the following:
...
-3. Verify that there are no extraneous differences from the release tag,
+2. Because it is possible to have divergences between release branches, the next
+   step depends on whether the release ``1.xx.y`` is a minor release (i.e.,
+   ``y=0``) or a patch release (i.e., ``y>0``). In the case of a minor release
+   perform step 3 and then continue with step 5, and in the case of a patch
+   release, skip to step 4 and continue from there.
+3. For a minor release, replace the upstream sources with those from the tagged
+   release, ``git checkout --no-overlay 1.xx.0 . ':!debian'``. This operation
+   should never produce a conflict. Commit the changes,
+   ``git commit -m "Checkout upstream sources for '1.xx.0' into debian/unstable"``.
```
eramongodb, line 36: In the interest of preserving branch history as with the former merge strategy, consider:

```bash
git merge --no-commit --no-ff 1.xx.0          # Conflicts may occur.
git checkout --no-overlay 1.xx.0 . ':!debian' # "Resolves" conflicts.
git add . && git commit
```
This should have the effect of creating merge commit while preserving the "overwrite" behavior of file checkout. Disregard if you would prefer the simpler checkout-only method as currently written.

rcsanchez97: So, I really like your proposal since it preserves the connection between the upstream history and the resulting packages (which is why I developed the initial procedure), but your proposal has a bit of a gap:

Here I am working the C++ driver repo and "re-doing" the 3.11.0 as a test.

```
roberto@build01:~/src/mongodb/mongo-cxx-driver.git (debian/unstable =)$  git checkout -b debian/unstable-test 8b009e4de2
Switched to a new branch 'debian/unstable-test'
roberto@build01:~/src/mongodb/mongo-cxx-driver.git (debian/unstable-test)$  git merge --no-commit --no-ff r3.11.0
(this produces a number of merge conflicts, as expected)
roberto@build01:~/src/mongodb/mongo-cxx-driver.git (debian/unstable-test *+|MERGING)$  git checkout --no-overlay r3.11.0 . ':!debian'
error: path 'docs/content/mongocxx-v3/installation/advanced.md' is unmerged
error: path 'docs/content/mongocxx-v3/installation/linux.md' is unmerged
error: path 'docs/content/mongocxx-v3/installation/macos.md' is unmerged
error: path 'docs/content/mongocxx-v3/installation/windows.md' is unmerged
error: path 'etc/generate-all-apidocs.pl' is unmerged
roberto@build01:~/src/mongodb/mongo-cxx-driver.git (debian/unstable-test *+|MERGING)$  git merge --abort
roberto@build01:~/src/mongodb/mongo-cxx-driver.git (debian/unstable-test)$
```

Note that the merge of the `r3.11.0` tag also removed most of the files in the `debian/` subdirectory, which we actually need to keep.

Here is what I came up with instead (for this specific example).

```
git merge --no-commit --no-ff r3.11.0
git checkout HEAD debian
git add .
git checkout --no-overlay r3.11.0 . ':!debian'
git add . && git commit
```

The output from this approach.

```
roberto@build01:~/src/mongodb/mongo-cxx-driver.git (debian/unstable-test)$  git merge --no-commit --no-ff r3.11.0
(this produces a number of merge conflicts, as expected)
roberto@build01:~/src/mongodb/mongo-cxx-driver.git (debian/unstable-test *+|MERGING)$  git checkout HEAD debian
Updated 25 paths from 46346fef27
roberto@build01:~/src/mongodb/mongo-cxx-driver.git (debian/unstable-test *+|MERGING)$  git add .
roberto@build01:~/src/mongodb/mongo-cxx-driver.git (debian/unstable-test +|MERGING)$  git checkout --no-overlay r3.11.0 . ':!debian'
Updated 68 paths from efba21d707
roberto@build01:~/src/mongodb/mongo-cxx-driver.git (debian/unstable-test +|MERGING)$  git add . && git commit
[debian/unstable-test 74991e51ef] Merge tag 'r3.11.0' into debian/unstable-test
roberto@build01:~/src/mongodb/mongo-cxx-driver.git (debian/unstable-test)$  git diff r3.11.0..HEAD --stat -- . ':!debian'
roberto@build01:~/src/mongodb/mongo-cxx-driver.git (debian/unstable-test)$
```

After performing this sequence of steps I cherry-picked the commits that were between the original checkout of `r3.11.0` onto `debian/unstable` and the final package release and then I rebuild the packages. The resulting packages weren't bit-for-bit identical (I think that our builds incorporate date/time built into one of the headers), but there was no source-level difference between the packages I created and released yesterday and the packages I created after following the above sequence.

If what I am proposing seems like it would make sense, then I am going further propose making this the procedure for all releases (minor and patch) in order to have a consistent outcome each time and to avoid the need to unnecessarily deal with any unforeseen merge conflicts.

eramongodb: SGTM.

> Note that the merge of the r3.11.0 tag also removed most of the files in the debian/ subdirectory, which we actually need to keep.

Should the `checkout HEAD -- debian` come after the branch checkout (see below)? Might not matter given the exclusion rule as long as the files are restored to `HEAD`.

Aside: recommend using `--` in the checkout commands to better visually distinguish them from branch checkouts. Unless I missed something, the `git add` should only be needed once.

```bash
git merge --no-commit --no-ff r3.11.0
git checkout --no-overlay r3.11.0 -- . ':!debian'
git checkout HEAD -- debian
git add .
git commit
```

rcsanchez97: I think that the ordering of the `git checkout HEAD` operation is not that important. However, `git add .` is required between the merge and the checkout.

Without the `git add .`:

```
roberto@build01:~/src/mongodb/mongo-cxx-driver.git (debian/unstable-test *+|MERGING)$  git checkout --no-overlay r3.11.0 -- . ':!debian'
error: path 'docs/content/mongocxx-v3/installation/advanced.md' is unmerged
error: path 'docs/content/mongocxx-v3/installation/linux.md' is unmerged
error: path 'docs/content/mongocxx-v3/installation/macos.md' is unmerged
error: path 'docs/content/mongocxx-v3/installation/windows.md' is unmerged
error: path 'etc/generate-all-apidocs.pl' is unmerged
```

Then:

```
roberto@build01:~/src/mongodb/mongo-cxx-driver.git (debian/unstable-test *+|MERGING)$  git add .
roberto@build01:~/src/mongodb/mongo-cxx-driver.git (debian/unstable-test +|MERGING)$  git checkout --no-overlay r3.11.0 -- . ':!debian'
```

eramongodb: Interesting. Thank you for the thorough testing of suggestions.

eramongodb (approved):

PR #1750, opened 2024-10-04
CDRIVER-5736 do not assert on failure to create OpenSSL context

kevinAlbs:
Changes in [CDRIVER-4656](https://github.com/mongodb/mongo-c-driver/pull/1673) resulted in a regression. Failure to create an OpenSSL context resulted in an abort, rather than an expected failure.

This PR updates `mongoc_stream_tls_openssl_new_with_context` to return NULL if the passed context is NULL to signal failure.

The context is created in `mongoc_client_pool_set_ssl_opts` and `mongoc_client_set_ssl_opts`. Neither have a way to return an error.


vector-of-bool (approved):
LGTM

PR #1751, opened 2024-10-04
CDRIVER-4822

joshbsiegel:


src/CMakeLists.txt
```diff
@@ -19,3 +19,21 @@ if (CMAKE_CXX_COMPILER)
...
+
+target_compile_options(public-header-warnings PRIVATE
+  $<$<C_COMPILER_ID:AppleClang,Clang>:-Weverything>
+  $<$<C_COMPILER_ID:GNU>:-Wextra>
+  $<$<C_COMPILER_ID:AppleClang,Clang,GNU>:-Werror>
+  $<$<C_COMPILER_ID:AppleClang,Clang,GNU>:-Wno-declaration-after-statement>
+  $<$<C_COMPILER_ID:AppleClang,Clang,GNU>:-Wno-disabled-macro-expansion>
+  $<$<C_COMPILER_ID:AppleClang,Clang>:-Wno-reserved-identifier>
+  $<$<C_COMPILER_ID:AppleClang,Clang>:-Wno-documentation-unknown-command>
+  $<$<C_COMPILER_ID:AppleClang,Clang>:-Wno-unknown-warning-option>
```
joshbsiegel, line 36: Some versions of Clang don't have certain warnings yet, this flag will silence the errors about unknown warnings.

src/libbson/src/bson/bson-atomic.h
```diff
@@ -32,6 +32,8 @@
 
 BSON_BEGIN_DECLS
 
+MC_DISABLE_COVERED_SWITCH_DEFAULT_BEGIN
```
kevinAlbs, line 35: Confirming: `-Wcovered-switch-default` is suppressed, rather than `default` removed, because it contradicts the warnings from `-Wswitch-default`, correct?

src/libbson/src/bson/bson-atomic.h
```diff
@@ -32,6 +32,8 @@
 
 BSON_BEGIN_DECLS
```
kevinAlbs, line 34: Since `bson-atomic.h` is the only header that requires `MC_DISABLE_COVERED_SWITCH_DEFAULT_*` and `MC_DISABLE_ATOMIC_IMPLICIT_SEQ_CST_*`, suggest defining them (and undefining) within this file:

```c
// ... #define macros ...

// ... rest of header ...

// ... #undef macros ...
```

That removes the macros from public API. `bson-atomic.h` is planned for removal in 2.0 (CDRIVER-5703).

src/libbson/src/bson/bcon.h
```diff
@@ -167,17 +168,21 @@ typedef enum {
    BCON_TYPE_ERROR,
 } bcon_type_t;
 
+MC_DISABLE_PADDED_WARNING_BEGIN
```
kevinAlbs, line 171: I would rather build with `-Wno-padded` than suppress. This seems like a case of "we know this causes a warning, but we choose not to address it".

Suppressing `-Wunsafe-buffer-usage` seems like case of "we know this causes a warning, but this is deliberate" as is the case of `-Wunsafe-buffer-usage`

src/libbson/src/bson/bson-atomic.h
joshbsiegel: Yes, removing `default` will then trigger the `-Wswitch-default` warning (even though all of the switch statements deal with enums that are fully covered by the cases). I tried removing the `default` case and instead `ASSERT`ing that the switched value is one of the enums, but the compiler didn't like that either. 

I think the only solution would be to remove the final `case` of each `switch` and replace it with the `default`. I don't love this idea though because it makes the code less readable and could cause issues later on if another value is added to the enum (new enum value would be covered by `default` instead of throwing an error, putting the ownership on the developer to remember they have to add a case to the switch statement).

kevinAlbs: Interesting. Thank you for the explanation. I agree with the current disabling the warning. Resolving.

src/CMakeLists.txt
```diff
@@ -19,3 +19,23 @@ if (CMAKE_CXX_COMPILER)
       target_link_libraries (mongoc-cxx-check PRIVATE mongoc_shared)
    endif ()
 endif ()
+
+# public-header-warnings
+add_executable(public-header-warnings EXCLUDE_FROM_ALL ${CMAKE_CURRENT_SOURCE_DIR}/c-check.c)
+
+target_link_libraries(public-header-warnings PRIVATE mongoc_static)
```
kevinAlbs, line 26:
```suggestion

if (TARGET mongoc_static)
   target_link_libraries(public-header-warnings PRIVATE mongoc_static)
else ()
   target_link_libraries(public-header-warnings PRIVATE mongoc_shared)
endif ()
```

Suggest checking if `mongoc_static` target is defined first. The static libraries may not be defined if CMake is configured with `ENABLE_STATIC=OFF`.

.evergreen/scripts/compile-unix.sh
```diff
@@ -216,3 +216,6 @@ build_dir="cmake-build"
 
 # Also validate examples.
 "${cmake_binary}" --build "${build_dir:?}" --target mongo_c_driver_examples
+
+# Check public headers for extra warnings.
+"${cmake_binary}" --build "${build_dir:?}" --target public-header-warnings
```
kevinAlbs, line 221:
```suggestion
if [[ "$EXTRA_CONFIGURE_FLAGS" != *"ENABLE_MONGOC=OFF"* ]]; then
  # Check public headers for extra warnings.
  "${cmake_binary}" --build "${build_dir:?}" --target public-header-warnings
fi
```

Suggestion may fix the [install-libmongoc-after-libbson](https://spruce.mongodb.com/task/mongo_c_driver_smoke_install_libmongoc_after_libbson_patch_e2b54b1be31f795b6d238c740ff1388679917ad7_670543f299a8ed000756ab50_24_10_08_14_38_45/logs?execution=0) task. The failure logs:

```
gmake: *** No rule to make target 'public-header-warnings'.  Stop.
```

That task first installs libbson without libmongoc by configuring with `ENABLE_MONGOC=OFF`.

`src/CMakeLists.txt` is only loaded if `ENABLE_MONGOC` is true [here](https://github.com/mongodb/mongo-c-driver/blob/060a759c228b63f09908a6b9a508c4447f89b911/CMakeLists.txt#L562).

To reproduce locally:
```
% cmake -DENABLE_MONGOC=OFF -S. -Bcmake-build
% cmake --build cmake-build --target public-header-warnings
make: *** No rule to make target `public-header-warnings'.  Stop.
```

.evergreen/scripts/compile-windows.sh
```diff
@@ -143,3 +143,6 @@ fi
 
 # Also validate examples.
 "${cmake_binary:?}" --build "${build_dir:?}" --config "${build_config:?}" --target mongo_c_driver_examples
+
+# Check public headers for extra warnings.
+"${cmake_binary:?}" --build "${build_dir:?}" --config "${build_config:?}" --target public-header-warnings
```
kevinAlbs, line 148:
```suggestion
if [[ "$EXTRA_CONFIGURE_FLAGS" != *"ENABLE_MONGOC=OFF"* ]]; then
  # Check public headers for extra warnings.
  "${cmake_binary:?}" --build "${build_dir:?}" --config "${build_config:?}" --target public-header-warnings
fi
```

kevinAlbs (approved):
LGTM with minor comments and a fix to the `install-libmongoc-after-libbson` task.

PR #1756, opened 2024-10-08
Support Python 3.12 in EVG config scripts

eramongodb:
- Python 3.12 removed the deprecated `pkg_resources` package:
  ```
  ModuleNotFoundError: No module named 'pkg_resources'
  ```
  We were only using it for `parse_version`, so it has been replaced with `package.version.Version` instead (changes to tasks.py).

- The `typing-extensions` package has compatibility issues with Python 3.12 prior to version 4.6.0:
  ```
  TypeError: type 'typing.TypeVar' is not an acceptable base type
  ```
  It has been upgraded alongside other dependencies of required packages with `poetry update` (changes to poetry.lock).

- The pydantic package makes breaking changes in version 2.0 which appears to be incompatible with the shrub-py package:
  ```
  pydantic_core._pydantic_core.ValidationError: 2 validation errors for EvgProject
  ```
  Downgraded and pinned the pydantic package version to 1.X until the shrub-py package supports pydantic 2.0 (changes to pyproject.toml).

### DIFF
--- a/.evergreen/config_generator/generate.py
+++ b/.evergreen/config_generator/generate.py
@@ -1,12 +1,6 @@
 #!/usr/bin/env python3
 
 # Generates *.yml files under .evergreen/generated_configs.
-#
-# Install dependencies by running (preferably in a virtual environment):
-#     python -m pip install -r .evergreen/config_generator/requirements.txt
-#
-# Invoke this using the command:
-#     python3 .evergreen/config_generator/generate-config.py
 
 
 import sys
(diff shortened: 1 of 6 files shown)


joshbsiegel (approved):
LGTM, `poetry run mc-evg-generate` now runs locally on Python 3.12

eramongodb:
Also took the opportunity to tidy up the Python package dependency specification. `pyyaml<5.4` is no longer required as `shrub.py` added PyYAML 6.0 support in https://github.com/evergreen-ci/shrub.py/commit/391f17f2cb98d8ade7b5493f56bda6f378ec7efb. The `requirements.txt` file is removed given it has been superceeded by pyproject.toml.

kevinAlbs (approved):

PR #1757, opened 2024-10-08
CDRIVER-5745 Enable and address aggressive -Wformat warnings

eramongodb:
## Summary

Resolves CDRIVER-5745. Verified by [this patch](https://spruce.mongodb.com/version/670549aca9b10b0007419c7d) (excluding the addition of `--verbose` for debugging).

Addresses 1007 (550 unique) `-Wformat` warnings in library, test, and example code.

## Scope of mongoc Warnings

The `MongoC-Warnings` module applied warning flags to all targets defined in the current directory and below using `add_compile_options()`. Unfortunately, this included external library sources such as kms-message and bundled zlib. This led to `-Wformat` warnings triggering build failures due to zlib sources:

```
src/zlib-1.3.1/gzwrite.c:407:40: error: format string is not a string literal [-Werror,-Wformat-nonliteral]
    len = vsnprintf(next, state->size, format, va);
                                       ^~~~~~
```

To avoid applying our warning flags to external library sources, the `MongoC-Warnings` module is refactored to set a `${mongoc-warning-options}` variable instead of using `add_compile_options()`. The `${mongoc-warning-options}` variable is used to apply the warning options to _only_ our sources and targets.

For libbson, the options are added to target library properties via `mongo_target_requirements()`. (This helper function should probably be reused by/extended to libmongoc someday).

For libmongoc, due to the libraries being built with sources included both directly (as source files, a la kms-message) and indirectly (as object files, a la bundled zlib), the options are added as source file properties (specifically for `${MONGOC_SOURCES}`) instead of library target properties.

> [!NOTE]
> Directory properties would apply warning options to the zlib object library target (currently defined in the `src/libmongoc` directory). Target properties would apply warning options to the kms-message sources (included as source files by `${SOURCE}`). Therefore, warning options are applied to `${MONGOC_SOURCES}` only. In the future, a refactor may enable setting warning options at the directory level, but a source-based/target-based approach is likely still preferable as done here.

For test and example targets, the options are added as target properties.

## -Wformat Options

Investigation revealed that neither `-Wformat` (currently as a maintainer flag) nor `-Wformat=2` (a subset of which is currently partially enabled as maintainer flags) are sufficient to fully enable all format warning options across GCC and Clang. For GCC, refer to [Warning Options](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html). For Clang, refer to [Diagnostic Flags in Clang](https://clang.llvm.org/docs/DiagnosticsReference.html).

To ensure full and proper format warning coverage, all format flags which are not explicitly documented as being included by `-Wformat` or `-Wformat=2` are explicitly specified. This required the addition of several compiler version checks for compatibility to avoid "unrecognized option" errors. The redundancy of some flags is deliberate (no harm).

> [!NOTE]
> `-Wformat-signedness` not being enabled by `-Wformat` appears to be a major reason behind the lack of valuable format warnings enforcement in the current EVG config despite maintainer flags being enabled.

## -Wformat Fixes

Most changes required a simple fix of either the specifier to match the provided argument. Enumerators are cast to `int` to avoid implementation-defined underlying type compatibility issues. As a special case, saturation is used for `expires_in_len` in `mcd_azure_access_token_try_init_from_json_str` to print as much of the string as possible (although this is probably very unlikely to happen).

The implementation-defined type of `__LINE__` was exposed by MinGW builds, where `unsigned int` (`DWORD`) is used to represent `__LINE__` instead of `int` as expected by format strings:

```
error: format '%ld' expects argument of type 'long int', but argument has type 'DWORD' {aka 'long unsigned int'} [-Werror=format=]
```

Therefore, all cases of `__LINE__` being passed directly as an argument to a format string (rather than to a function parameter or variable of type `int`) are now explicitly cast to `int`. I do not expect we will ever need to support a source file that exceeds `int` representability.

A workaround was required for the formatting of `uint8_t` due to `PRIu8` on some platforms/build configurations (specifically release builds with GCC...?) conflicting with `-Wformat` due to implicit integer promotion rules:

```
error: format '%u' expects argument of type 'unsigned int', but argument has type 'int' [-Werror=format=]
  [...]
/usr/include/inttypes.h:102:18: note: format string is defined here
  102 | # define PRIu8  "u"
```

It is not entirely clear whether the stdlib implementation (should it be using `%hhu`...?) or the compiler (all `uint8_t` is representable by `int`; `uint8_t` -> `int` -> `unsigned int` should be perfectly fine...?) is to blame, but nevertheless, `%d` is used instead of `PRIu8` to address this (false-positive?) warning. (Relevant discussion: https://sourceware.org/bugzilla/show_bug.cgi?id=5747.)

> [!NOTE]
> Investigating this issue revealed a possibly obsolete definition of [__USE_MINGW_ANSI_STDIO](https://github.com/mongodb/mongo-c-driver/blob/58c3836e2c03361a49a914fbe82e67212d54a41a/src/libbson/src/bson/bson-compat.h#L27) which should now be handled by our defining `_XOPEN_SOURCE=700` + `_DEFAULT_SOURCE` in the CMake config. This is left untouched for now.

This PR also applies a drive-by fix for a new `-Wconversion` warning in mongoc-bulkwrite.c introduced by https://github.com/mongodb/mongo-c-driver/commit/84728bdb4e7def11a792e3bcb1b94811feff7b62.

Additionally, as a drive-by improvement and part of reducing the scope of mongoc warning options, `${mongo-c-driver_SOURCE_DIR}` is used in place of `*_SOURCES` variables (defined at root level) and `../`-relative paths when specifying source files for bundled external library sources.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -458,28 +458,6 @@ if (ENABLE_MONGOC)
       message (FATAL_ERROR "ENABLE_TESTS requires ENABLE_STATIC or ENABLE_STATIC_BUILD")
    endif ()
 
-   set (UTF8PROC_SOURCES
-      ${mongo-c-driver_SOURCE_DIR}/src/utf8proc-2.8.0/utf8proc.c
-   )
-
-   set (ZLIB_SOURCES
-      ${mongo-c-driver_SOURCE_DIR}/src/zlib-1.3.1/adler32.c
-      ${mongo-c-driver_SOURCE_DIR}/src/zlib-1.3.1/crc32.c
-      ${mongo-c-driver_SOURCE_DIR}/src/zlib-1.3.1/deflate.c
-      ${mongo-c-driver_SOURCE_DIR}/src/zlib-1.3.1/infback.c
-      ${mongo-c-driver_SOURCE_DIR}/src/zlib-1.3.1/inffast.c
-      ${mongo-c-driver_SOURCE_DIR}/src/zlib-1.3.1/inflate.c
-      ${mongo-c-driver_SOURCE_DIR}/src/zlib-1.3.1/inftrees.c
-      ${mongo-c-driver_SOURCE_DIR}/src/zlib-1.3.1/trees.c
-      ${mongo-c-driver_SOURCE_DIR}/src/zlib-1.3.1/zutil.c
-      ${mongo-c-driver_SOURCE_DIR}/src/zlib-1.3.1/compress.c
-      ${mongo-c-driver_SOURCE_DIR}/src/zlib-1.3.1/uncompr.c
-      ${mongo-c-driver_SOURCE_DIR}/src/zlib-1.3.1/gzclose.c
-      ${mongo-c-driver_SOURCE_DIR}/src/zlib-1.3.1/gzlib.c
-      ${mongo-c-driver_SOURCE_DIR}/src/zlib-1.3.1/gzread.c
-      ${mongo-c-driver_SOURCE_DIR}/src/zlib-1.3.1/gzwrite.c
-   )
-
    set (CPACK_RESOURCE_FILE_LICENSE "${mongo-c-driver_SOURCE_DIR}/COPYING")
 
    include (CPack)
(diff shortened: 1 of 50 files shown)


kevinAlbs (approved):
Warning fixes are much appreciated. LGTM

joshbsiegel (approved):
LGTM

PR #1758, opened 2024-10-08
CDRIVER-5736 do not assert

kevinAlbs:
Follow-up to https://github.com/mongodb/mongo-c-driver/pull/1750 to fix failing Windows tasks.

The Windows hosts appear to have the expected CA file installed:

```
PS C:\Users\Administrator> Get-ChildItem -Recurse Cert:
(...)
Subject      : C=US, S=New York, L=New York City, O=MongoDB, OU=Drivers, CN=Drivers Testing CA
```

So no error is returned for the test.

Verified with this [patch build](https://spruce.mongodb.com/version/670586e5dac49200074a5795).

### DIFF
--- a/src/libmongoc/tests/test-mongoc-ssl.c
+++ b/src/libmongoc/tests/test-mongoc-ssl.c
@@ -205,7 +205,7 @@ static void
 test_non_existant_cafile (void)
 {
    mongoc_client_t *client = mongoc_client_new ("mongodb://localhost:27017/?tls=true&tlsCAFile=/nonexistant/ca.pem");
-   ASSERT (!mongoc_client_command_simple (client, "admin", tmp_bson ("{'ping': 1}"), NULL, NULL, NULL));
+   mongoc_client_command_simple (client, "admin", tmp_bson ("{'ping': 1}"), NULL, NULL, NULL);
    mongoc_client_destroy (client);
 }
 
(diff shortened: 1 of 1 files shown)


src/libmongoc/tests/test-mongoc-ssl.c
```diff
@@ -205,7 +205,7 @@ static void
 test_non_existant_cafile (void)
 {
    mongoc_client_t *client = mongoc_client_new ("mongodb://localhost:27017/?tls=true&tlsCAFile=/nonexistant/ca.pem");
-   ASSERT (!mongoc_client_command_simple (client, "admin", tmp_bson ("{'ping': 1}"), NULL, NULL, NULL));
+   mongoc_client_command_simple (client, "admin", tmp_bson ("{'ping': 1}"), NULL, NULL, NULL);
```
vector-of-bool, line 208: Should this fail even if the Windows has the cert file? If the `tlsCAFile` names a non-existent file, then do we really want to fall back to the Windows cert store? And if so, should it fail anyway since the user is explicitly requesting a non-existing CA cert?

kevinAlbs: That may be a worthwhile later improvement. But at present, failure to load the CA [file logs](https://github.com/mongodb/mongo-c-driver/blob/023df6a377683c3b3621713b3901cc1f995651c9/src/libmongoc/src/mongoc/mongoc-secure-channel.c#L292) but does not propagate the error.

kevinAlbs: Filed CDRIVER-5747 to propose propagating TLS set-up errors. To unblock the 1.28.1 release, I would rather keep the behavior as-is for now (no worse than the status quo).

vector-of-bool (approved):
LGTM

PR #1767, opened 2024-10-12
CDRIVER-5754 always write output on error in bson_validate_with_error_and_offset

mdbmes:
This fixes an edge case in `bson_validate_with_error_and_offset` that would cause unexpectedly uninitialized output parameters when a failure occurs at offset zero.

Includes the test case from CDRIVER-5754, now passing.

### DIFF
--- a/src/libbson/tests/test-bson.c
+++ b/src/libbson/tests/test-bson.c
@@ -1120,6 +1120,18 @@ test_bson_validate_dbpointer (void)
 
 
 static void
+test_bson_validate_with_error_and_offset (void)
+{
+   size_t err_offset = 12345;
+   bson_error_t err = { 67890 };
+   bson_t bson = { 0 };
+   ASSERT (!bson_validate_with_error_and_offset (&bson, BSON_VALIDATE_NONE, &err_offset, &err));
+   ASSERT (err_offset == 0);
+   ASSERT (err.domain != 67890);
+}
+
+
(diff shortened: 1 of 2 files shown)


src/libbson/tests/test-bson.c
```diff
@@ -1119,6 +1119,18 @@ test_bson_validate_dbpointer (void)
...
 
+static void
+test_bson_validate_with_error_and_offset (void)
+{
+   size_t err_offset = 12345;
+   bson_error_t err = {67890};
+   bson_t bson = {0};
+   ASSERT (!bson_validate_with_error_and_offset (&bson, BSON_VALIDATE_NONE, &err_offset, &err));
+   ASSERT (err_offset == 0);
+   ASSERT (err.domain != 67890);
```
kevinAlbs, line 1130:
```suggestion
   ASSERT_CMPSIZE_T (err_offset, ==, 0);
   ASSERT_CMPUINT32 (err.domain, !=, 67890); // domain is overwritten.
```

Suggest using `ASSERT_CMP*`. On failure, the compared numbers are printed:

```
Assert Failure: 0 == 123
```

kevinAlbs (approved):
Nice find! LGTM with a test suggestion.

After merging to `master`, I suggest cherry-picking to the `r1.28` branch to include this fix in the next 1.28 release.

PR #1769, opened 2024-10-14
CDRIVER-5721 document bson_validate_with_error_and_offset()

mdbmes:
Adds the missing documentation page for `bson_validate_with_error_and_offset()`. This also moves `bson_validate_flags_t` to its own page, and clarifies the different meanings of `0` here in libbson vs libmongoc.

This patch re-runs the `build/generate-opts` script, which also caused minor unrelated diffs in `mongoc-opts.c`.


src/libmongoc/src/mongoc/mongoc-opts-private.h
rcsanchez97, line 1: The license header needs to be restored.

rcsanchez97 (approved):
LGTM, modulo the one license header that needs to be restored.

src/libmongoc/src/mongoc/mongoc-opts-private.h
mdbmes: This file is automatically generated, I just re-ran the generator. It looks like someone added this header without including it on the template as well. I'll fix that. Thanks.

mdbmes:
Thanks for the review, Roberto!

PR #1770, opened 2024-10-15
CDRIVER-3769 Deprecate unimplemented URI options

mdbmes:
- Removes maxidletimems and waitqueuemultiple from the schema of allowed int options
- Adds the suggested unit test from CDRIVER-3769
- Adds an additional similar deprecation unit test for waitqueuemultiple
- Updates other URI unit tests to account for the new expected behavior


src/libmongoc/tests/json/uri-options/connection-options.json
```diff
@@ -2,7 +2,7 @@
   "tests": [
     {
       "description": "Valid connection and timeout options are parsed correctly",
-      "uri": "mongodb://example.com/?appname=URI-OPTIONS-SPEC-TEST&connectTimeoutMS=20000&heartbeatFrequencyMS=5000&localThresholdMS=3000&maxIdleTimeMS=50000&replicaSet=uri-options-spec&retryWrites=true&serverSelectionTimeoutMS=15000&socketTimeoutMS=7500",
```
kevinAlbs, line 5: `connection-options.json` is copied from the specifications repo (written [in YAML](https://github.com/mongodb/specifications/blob/master/source/uri-options/tests/connection-options.yml) then converted to JSON).

Suggest leaving this test file unchanged, so future changes to the spec tests can be copied in without interfering with libmongoc-specific modifications. Instead, skip no-longer-applicable tests. Example in `test_connection_uri_cb`:

```c
// Check if test must be skipped.
{
    const char *description = bson_lookup_utf8 (&test_case, "description");
    test_skip_t skips[] = {{.description = "Valid connection pool options are parsed correctly",
                            .reason = "libmongoc does not support maxIdleTimeMS"},
                        {.description = "Valid connection and timeout options are parsed correctly",
                            .reason = "libmongoc does not support maxIdleTimeMS"}};
    bool must_skip = false;
    for (size_t i = 0; i < sizeof (skips) / sizeof (skips[0]); i++) {
    if (0 == strcmp (description, skips[i].description)) {
        printf ("  - Skipping: '%s' with reason: '%s'\n", description, skips[i].reason);
        must_skip = true;
    }
    }
    if (must_skip) {
    break;
    }
}
```

And add libmongoc-specific tests to `additional-nonspec-tests.json`.

mdbmes: Thanks! I've implemented effectively the same thing in the patch now, but it seemed worth factoring the test and the message into a new utility function.

kevinAlbs (approved):

mdbmes:
evergreen retry

PR #1774, opened 2024-10-18
CDRIVER-4602 regenerate build variants config

eramongodb:
Followup to https://github.com/mongodb/mongo-c-driver/pull/1760. Addresses the warnings observed in EVG build pages:

image

Regenerates build variants config YAML file, which appears to have only been partially updated in https://github.com/mongodb/mongo-c-driver/pull/1760.


kevinAlbs (approved):
Thank you. LGTM

PR #1775, opened 2024-10-18
CDRIVER-4602 build fix for clang >= 3.1 and < 3.7

mdbmes:
This fixes the CentOS 7 build on Earthly that uses Clang 3.4.2.

We still require at least clang 3.1, due to "format-non-iso".

"format-pedantic" was added in 3.7, but since the sequence of compatibility tests here uses major versions only I'm testing against version 4.0.

CentOS 7 normally does not ship with clang (the test gets clang from EPEL).
It also normally has a cmake too old to build mongo-c-driver; our CI builds bring their own cmake.


eramongodb (approved):
Nice spot. LGTM.

PR #1776, opened 2024-10-21
CDRIVER-5700 deprecate `bson_as_json` and `bson_array_as_json`

kevinAlbs:
# Summary

- Deprecate `bson_as_json` for newly added `bson_as_legacy_extended_json`.
- Deprecate `bson_array_as_json` for newly added `bson_array_as_legacy_extended_json`.
- Document libbson's Legacy Extended JSON format.
- Test side-by-side comparison of Canonical, Relaxed, and Legacy Extended JSON.

Verified with this [patch build](https://spruce.mongodb.com/version/6716a86d6f27b60007428eeb).

# Background & Motivation

Deprecating intends to better inform users that `bson_as_json` and `bson_array_as_json` use Legacy Extended JSON. `bson_as_legacy_extended_json` and `bson_array_as_legacy_extended_json` are added as non-deprecated alternatives to ease migration. Users wanting to continue using Legacy Extended JSON can find+replace to the non-deprecated alternatives.

Significant existing use of `bson_as_json` is expected (see [scope](https://docs.google.com/document/d/13-pxZ9LPpIcoYSxkkQ7IfDhKCaw_cxyjIZfcVrC5J3g/edit#heading=h.i81lzqug228b)). Deprecation of encoding Legacy Extended JSON can be considered for future releases. 

Documenting the libbson's Legacy Extended JSON format is intended to inform users wanting to migrate to non-Legacy formats.

Example code snippets in docs have been moved to a new `extended-json.c` source file to ease modifying and ensure they compile.

Examples are updated to use `bson_as_relaxed_extended_json` to encourage use of non-Legacy formats.

libmongoc tests are updated to use `bson_as_relaxed_extended_json`.

libbson tests intending to test encoding are updated to use `bson_as_legacy_extended_json` to clarify Legacy Extended JSON is intended.

## Testing

A new test is added to compare encoding of Canonical, Relaxed, and Legacy Extended JSON. This led to the discovery of CDRIVER-5759.

Notably, libbson's Legacy Extended JSON encoding of double Infinity and NaN is implementation defined. libbson's Legacy Extended JSON uses `"%.20g"` to encode all double values:

```c
bson_t *b = BCON_NEW ("double_inf", BCON_DOUBLE (INFINITY));
char *as_legacy = bson_as_json (b, NULL);
printf ("%s\n", as_legacy);
// Prints either:
// { "double_inf" : inf }
// { "double_inf" : infinity }
```

Quoting [C99](https://port70.net/~nsz/c/c99/n1256.html#7.19.6):

> A double argument representing an infinity is converted in one of the styles [-]inf or [-]infinity -- which style is implementation-defined. A double argument representing a NaN is converted in one of the styles [-]nan or [-]nan(n-char-sequence) -- which style, and the meaning of any n-char-sequence, is implementation-defined.

Preserving this encoding seems somewhat deliberate. Quoting https://github.com/mongodb/libbson/commit/6929d34e7af2ccdc1ce283daf72b4d4dff952849

> Maintain legacy bson_as_json() output for non-finite values. Note that legacy mode will produce "nan" and "inf" literals, which are not valid JSON.

This PR does not propose changing this encoding to preserve backward compatibility. Instead, this behavior is documented to inform users.

### DIFF
--- a/src/libbson/tests/test-json.c
+++ b/src/libbson/tests/test-json.c
@@ -8,6 +8,7 @@
 #include "test-conveniences.h"
 #include <common-string-private.h>
 #include <common-cmp-private.h>
+#include <bson/bson-iso8601-private.h>
 
 static ssize_t
 test_bson_json_read_cb_helper (void *string, uint8_t *buf, size_t len)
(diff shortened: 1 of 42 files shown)


src/libbson/tests/test-json.c
```diff
@@ -3367,6 +3368,318 @@ test_parse_array (void)
...
+      {
+         bson_t *b = BCON_NEW ("double_nan", BCON_DOUBLE (NAN));
+         ASSERT_AS_JSON (b, BSON_JSON_MODE_CANONICAL, "{ 'double_nan' : { '$numberDouble' : 'NaN' } }");
+         ASSERT_AS_JSON (b, BSON_JSON_MODE_RELAXED, "{ 'double_nan' : { '$numberDouble' : 'NaN' } }");
+         // Legacy encoding is platform dependent. May encode to '[-]nan' or '[-]nan(n-char-sequence)' without quotes.
+         {
+            char *as_legacy = bson_as_legacy_extended_json (b, NULL);
+            char *ptr = as_legacy;
+            ASSERT_STARTSWITH (ptr, "{ \"double_nan\" : ");
+            ptr += strlen ("{ \"double_nan\" : ");
```
mdbmes, line 3626: This seems barely worth mentioning, but it makes me a bit nervous to have two copies of what's really the same prefix string. If they get out of sync the following comparison will be made at the wrong location, which should be detected immediately. Still, it might be a little less error prone to define a const local var for this.

mdbmes (approved):
Looks good to me!
One extremely minor comment, please feel free to ignore.

src/libbson/tests/test-json.c
kevinAlbs: Seems reasonable enough. Updated.

PR #1780, opened 2024-10-28
CDRIVER-5767, CDRIVER-5774: Update URI options tests

alcaeus:
Syncs URI options tests to include recent changes to the URI options spec tests. Tests for specifications that are not implemented in libmongoc are skipped with a reference to the corresponding ticket.


eramongodb (approved):
LGTM. Thank you for referencing related tickets in the skip reasons.

PR #1783, opened 2024-11-04
CDRIVER-3054 Revise docs from `mongoc_collection_(remove|insert|update)`

kevinAlbs:
# Summary

This PR intends to resolve CDRIVER-3054 with documentation, rather than deprecation.

- Remove "superseded by" in docs of `mongoc_collection_(remove|insert|update)`
- Document `mongoc_collection_remove` as an alternative to `mongoc_collection_delete`.
- Document and test `mongoc_collection_update` accepts a replacement document.
- Add missing `mongoc_init` and `mongoc_cleanup` to `example-update.c`.

Verified by [this patch build](https://spruce.mongodb.com/version/6728d2a45a7cbe000758099d).

# Background & Motivation

CDRIVER-3054 originally proposed deprecating `mongoc_collection_(remove|insert|update)`. Choosing not to deprecate was motivated by a non-trivial migration to the alternatives. Migrating calls with a write concern requires creating a new options BSON. Example:

Before:
```c
return mongoc_collection_insert (collection, MONGOC_INSERT_NONE, document, write_concern, error);
```

After:
```c
if (write_concern && !mongoc_write_concern_append (write_concern, &opts)) {
    bson_set_error (
        error, MONGOC_ERROR_COMMAND, MONGOC_ERROR_COMMAND_INVALID_ARG, "failed to append write concern");
    bson_destroy (&opts);
    return false;
}
bool ok = mongoc_collection_insert_one (collection, document, &opts, NULL /* reply */, error);
bson_destroy (&opts);
return ok;
```

I expect removing `mongoc_collection_(remove|insert|update)` in 2.0 would not reduce much maintenance burden. They are mostly implemented in terms of non-deprecated functions. I would rather err towards an easier upgrade. I do not think the deprecation/removal has a clear benefit.

Unlike `mongoc_collection_update_(one|many)`, `mongoc_collection_update` accepts both update documents (e.g. `{ "$set": { "foo": "bar" }}`) and replacement documents (e.g. `{ "foo": "bar" }`). Documentation was updated, and a test was added.

### DIFF
--- a/src/libmongoc/doc/mongoc_collection_delete.rst
+++ b/src/libmongoc/doc/mongoc_collection_delete.rst
@@ -6,7 +6,7 @@ mongoc_collection_delete()
 .. warning::
    .. deprecated:: 1.9.0
 
-      Use :symbol:`mongoc_collection_delete_one()` or :symbol:`mongoc_collection_delete_many()` instead.
+      Use :symbol:`mongoc_collection_remove`, :symbol:`mongoc_collection_delete_one()` or :symbol:`mongoc_collection_delete_many()` instead.
 
 Synopsis
 --------
--- a/src/libmongoc/doc/mongoc_collection_insert.rst
+++ b/src/libmongoc/doc/mongoc_collection_insert.rst
@@ -27,7 +27,9 @@ Parameters
 Description
 -----------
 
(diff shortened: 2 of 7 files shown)


vector-of-bool (approved):
LGTM

PR #1786, opened 2024-11-08
CDRIVER-3436 Document "ssl" APIs use TLS protocols

kevinAlbs:
# Summary

Document "ssl" APIs use TLS protocols

# Background & Motivation

CDRIVER-3436 originally proposed deprecating `mongoc_ssl_opt_t`. This PR does not deprecate to avoid unnecessary extra migration effort for users.

The motivation for deprecating `mongoc_ssl_opt_t` refers to the MongoDB URI options [preferring "tls" over "ssl" as a prefix](https://github.com/mongodb/specifications/blob/fc7996db26d0ea92091a5034c6acb287ef7282fe/source/uri-options/uri-options.md#why-use-tls-as-the-prefix-instead-of-ssl-for-related-options).

Deprecating "ssl" termed API was done on [this branch](https://github.com/mongodb/mongo-c-driver/compare/master...kevinAlbs:mongo-c-driver:deprecate-ssl.C3436?expand=1) but was abandoned. Though the migration is simple (find+replace), I think it is an unnecessary effort for users with little value gained. I expect there is significant existing use of `mongoc_ssl_opt_t`. [GitHub code search](https://github.com/search?q=mongoc_ssl_opt_t+path%3A*.c+-repo%3Amongodb%2Fmongo-c-driver+NOT+is%3Afork+NOT+path%3A%2Fmongoc%5C-.*%5C.c%2F+NOT+path%3A%2Fbson%5C-.*%5C.c%2F+NOT+path%3Abson.c++NOT+path%3A%2Fmongocrypt%5C-.*%5C.c%2F+NOT+path%3A%2F.*test.*%2F&type=code&ref=advsearch) shows 7 unique looking non-MongoDB uses.

There is precedent among drivers for "ssl" named APIs: Java's [SslSettings](https://mongodb.github.io/mongo-java-driver/5.2/apidocs/mongodb-driver-core/com/mongodb/connection/SslSettings.html) and C#'s [SslSettings](https://mongodb.github.io/mongo-csharp-driver/3.0.0/api/MongoDB.Driver/MongoDB.Driver.SslSettings.html).

### DIFF
--- a/src/libmongoc/doc/conf.py
+++ b/src/libmongoc/doc/conf.py
@@ -223,6 +223,9 @@ rst_prolog = rf"""
 
 .. _bson_lifetimes: https://www.mongodb.com/docs/languages/c/c-driver/current/libbson/guides/lifetimes/
 
+.. |ssl:naming| replace::
+    Though some API names include the term "ssl", the C driver only support TLS protocols, which supersede SSL.
+
 """
 
 
--- a/src/libmongoc/doc/mongoc_client_pool_set_ssl_opts.rst
+++ b/src/libmongoc/doc/mongoc_client_pool_set_ssl_opts.rst
@@ -14,6 +14,9 @@ Synopsis
                                    const mongoc_ssl_opt_t *opts);
   #endif
 
+.. note::
+   |ssl:naming|
+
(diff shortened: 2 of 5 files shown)


vector-of-bool (approved):
LGTM

PR #1787, opened 2024-11-12
Remove skip for loading crypt_shared on Windows

kevinAlbs:
When investigating a reported Windows crash unloading crypt shared ([JAVA-5648](https://jira.mongodb.org/browse/JAVA-5648)), I rediscovered that testing with the crypt shared library was skipped on Windows. The skip was added in https://github.com/mongodb/mongo-c-driver/pull/968. I am not sure what motivated the skip. Removing the skips appears to pass. The logs of the [patch build](https://spruce.mongodb.com/version/6733524718af6a0007318507/) include:

```
crypt_shared library will be loaded with cryptSharedLibPath: [C:/data/mci/5c60da88b4d482d9f4943604b2d9b92a/mongoc/mongo_crypt_v1.dll]
```


vector-of-bool (approved):
LGTM

PR #1794, opened 2024-11-18
Use Artifactory to obtain Docker images

eramongodb:
Verified by [this patch](https://spruce.mongodb.com/version/673b77880014770007fff149).

Appends a `docker login` command prior to Earthly/Docker commands and uses the custom Earthly buildkit image in `tools/earthly.sh` as done in https://github.com/mongodb/libmongocrypt/pull/911.

The Ubuntu 20.04 distro is dropped from the list due to the following obtuse error:

```
Error saving credentials: error storing credentials - err: exit status 1, out: `not implemented`
```

### DIFF
--- a/.evergreen/config_generator/components/earthly.py
+++ b/.evergreen/config_generator/components/earthly.py
@@ -228,6 +228,15 @@ def earthly_task(
     return EvgTask(
         name=name,
         commands=[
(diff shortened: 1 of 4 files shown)


kevinAlbs (approved):
LGTM, and thank you. This reminded me the step to authenticate was needed. Updated the libmongocrypt PR accordingly.

PR #1797, opened 2024-11-22
Added pragma for Advapi32.lib

MonkeybreadSoftware:
Without adding Advapi32.lib, the code won't link on Windows 64-bit ARM for me when building with clang. The #pragma is a simple way to tell linker to include the library later.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-socket.c
+++ b/src/libmongoc/src/mongoc/mongoc-socket.c
@@ -27,6 +27,7 @@
 #ifdef _WIN32
 #include <Mstcpip.h>
 #include <process.h>
+#pragma comment(lib, "Advapi32.lib")
 #endif
 #include <common-cmp-private.h>
 


MonkeybreadSoftware:
I may add that it builds for Intel 64-bit and 32-bit with clang and Visual Studio 2022 without this changes. 
Just for ARM we need it.

MonkeybreadSoftware:
I build using Visual Studio 2022 with the command line, so nmake  and cl command line tools.
Maybe clang doesn't need it as this is a library basically any app needs, so it may get added as default?

MonkeybreadSoftware:
Hello,

Thanks for asking.

I copy mongo-c-driver-1.29.0 to windows machine.
Then I open cmd for VS Studio 2022 with arm64 target.

I configure it here:

```
cmake . -G"NMake Makefiles" -DOPENSSL_ROOT_DIR=Y:\Libs\openssl\openssl/win64-arm -DENABLE_STATIC=ON -DENABLE_TESTS=OFF -DENABLE_EXAMPLES=OFF -DENABLE_MAN_PAGES=OFF -DENABLE_HTML_DOCS=OFF -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS=" -O2 -DMONGOC_NO_AUTOMATIC_GLOBALS=1 " -DCMAKE_SYSTEM_PROCESSOR=arm64 -DWITH_CRT_DLL=ON -DENABLE_TRACING=ON -BC:\TEMP\winARM

CD C:\TEMP\winARM

nmake all

```

I get an error about linking:

Bildschirmfoto 2024-12-04 um 07 53 33

This is when linking mongoc-1.0.dll and in the list of libraries to link, the Advapi32.lib is missing for Registry functions.

MonkeybreadSoftware:
did it build the DLL files?
Depending on what code paths you go, something else may include Advapi32.lib and then it works.
In my case it is missing.

MonkeybreadSoftware:
Thanks. Your branch builds just fine for Windows ARM.

For a test, I removed your target_link_libraries line and then it fails again. So seems to be the change needed.
I would love to see this change in the next release.

PR #1803, opened 2024-12-03
CDRIVER-5819 always assign operation_id when bulk client changes

jmikola:
https://jira.mongodb.org/browse/CDRIVER-5819


jmikola:
@kevinAlbs: I think the Evergreen "setup failed" errors are unrelated, but please confirm.

kevinAlbs (approved):
Thank you for the fix.

kevinAlbs:
> I think the Evergreen "setup failed" errors are unrelated, but please confirm.

Agreed. I expect these are unrelated.

PR #1805, opened 2024-12-04
CDRIVER-5816: mongoc_bulkwrite_new and mongoc_bulkwrite_set_client

jmikola:
https://jira.mongodb.org/browse/CDRIVER-5816
https://jira.mongodb.org/browse/CDRIVER-5817

Using this to POC the PHPC implementation in [PHPC-2478](https://jira.mongodb.org/browse/PHPC-2478), so there is no rush to review/merge.

PR #1806, opened 2024-12-05
CDRIVER-5824 remove EXPORT macros from source files

jmikola:
https://jira.mongodb.org/browse/CDRIVER-5824

@kevinAlbs: This dates back to the initial release in 1.28.0, but I'm not sure whether it warrants a patch release. I figured it easier to fix in master and then you can backport as desired.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-bulkwrite.c
+++ b/src/libmongoc/src/mongoc/mongoc-bulkwrite.c
@@ -1438,7 +1438,7 @@ _bulkwritereturn_apply_result (mongoc_bulkwritereturn_t *self,
    return true;
 }
 
-BSON_EXPORT (void)
+void
 mongoc_bulkwrite_set_session (mongoc_bulkwrite_t *self, mongoc_client_session_t *session)
 {
    BSON_ASSERT_PARAM (self);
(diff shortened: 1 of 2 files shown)


src/libmongoc/src/mongoc/mongoc-bulkwrite.c
```diff
@@ -1438,7 +1438,7 @@ _bulkwritereturn_apply_result (mongoc_bulkwritereturn_t *self,
    return true;
 }
 
-BSON_EXPORT (void)
+MONGOC_EXPORT (void)
```
kevinAlbs, line 1441:
```suggestion
void
```

kevinAlbs: I expect `MONGOC_EXPORT` / `BSON_EXPORT` is only needed in public headers, and is harmless (though unnecessary) in source files.

The only other occurrence is [mongoc_client_encryption_new](https://github.com/mongodb/mongo-c-driver/blob/2c4dd7581beb65fcf97b44b68f36e9ba7152733c/src/libmongoc/src/mongoc/mongoc-client-side-encryption.c#L893), which I expect can be removed too.

jmikola: Noted. I'll remove this here and in the new APIs introduced in https://github.com/mongodb/mongo-c-driver/pull/1805.

jmikola:
I repurposed CDRIVER-5824 to remove both of these `EXPORT` macros.

kevinAlbs (approved):
Thank you for the clean-up.

PR #1807, opened 2024-12-06
CDRIVER-5826 Fix result of `bson_strerror_r` on macOS

kevinAlbs:
Follow-up to https://github.com/mongodb/mongo-c-driver/pull/1370/files. On macOS `bson_strerror_r` always returned NULL and set the output buffer to "Unknown error message". This PR adds a missing `ret = buf;` assignment to fix.

Verified with [this patch build](https://spruce.mongodb.com/version/675320d2c9816e0007e9dca3).


eramongodb (approved):
Oh dear. Good catch. LGTM.

PR #1810, opened 2024-12-12
CDRIVER-5839 Unofficial ARM64 Windows support

adriandole:
Test system: ARM64 Windows VM running on an M1 Mac using [UTM](https://mac.getutm.app/).

Build in the ARM64 Native tools command prompt.

Configure for clang:
```
cmake -D ENABLE_EXTRA_ALIGNMENT=OFF -D ENABLE_AUTOMATIC_INIT_AND_CLEANUP=OFF -D CMAKE_BUILD_TYPE=Debug -D ENABLE_MONGOC=ON -G "Visual Studio 17 2022" -A ARM64 -T ClangCL -D CMAKE_C_FLAGS="-w" -S srcDir -B buildDir
```

Configure for MSVC:
```
cmake -G"NMake Makefiles" -DCMAKE_SYSTEM_PROCESSOR=arm64 -DENABLE_TRACING=ON -DENABLE_SSL=WINDOWS 
```

### DIFF
--- a/src/libmongoc/CMakeLists.txt
+++ b/src/libmongoc/CMakeLists.txt
@@ -154,6 +154,7 @@ endif()
 add_library(_mongoc-dependencies INTERFACE)
 add_library(mongo::detail::c_dependencies ALIAS _mongoc-dependencies)
 target_link_libraries(_mongoc-dependencies INTERFACE mongo::detail::c_platform)
+target_link_libraries(_mongoc-dependencies INTERFACE mongo::detail::c_platform $<$<PLATFORM_ID:Windows>:advapi32.lib>)
 install(TARGETS _mongoc-dependencies EXPORT mongoc-targets)
 set_property(TARGET _mongoc-dependencies PROPERTY EXPORT_NAME detail::c_dependencies)
 


kevinAlbs (commented):
Changes LGTM, but I expect may not resolve https://github.com/mongodb/mongo-c-driver/pull/1797. I expect Advapi32 is a required dependency on Windows. For example: `RegCloseKey` is used in [mongoc-socket.c](https://github.com/mongodb/mongo-c-driver/blob/7c2300c3638c660491e5b52c378f661757659c6b/src/libmongoc/src/mongoc/mongoc-socket.c#L411) and is [documented](https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regclosekey) in Advapi32.

Building on x86_64 shows Advapi32 is linked. Here is a screenshot of dependency walker:
Screenshot 2024-12-18 at 12 44 47 PM

I think the difference on ARM64 is CMake [does not include Advapi32 in the standard libraries](https://github.com/Kitware/CMake/blob/bacfc7c313a7cdcb15bc584fdb6350493aa84858/Modules/Platform/Windows-MSVC.cmake#L223-L227). I suggest making the dependency explicit:

```cmake
target_link_libraries(_mongoc-dependencies INTERFACE mongo::detail::c_platform $<$<PLATFORM_ID:Windows>:advapi32.lib>)
```

kevinAlbs (approved):

PR #1811, opened 2024-12-12
CDRIVER-5835 Rename `set_error` to `kms_set_error`

MonkeybreadSoftware:
set_error is an awful generic name for a global function.

Sadly this easily conflicts with other libraries if you statically link this into an application.

please have us rename set_error to kms_set_error to avoid link errors with libraries/applications that also have set_error as function.

### DIFF
--- a/.evergreen/scripts/kms-divergence-check.sh
+++ b/.evergreen/scripts/kms-divergence-check.sh
@@ -13,7 +13,7 @@ LIBMONGOCRYPT_DIR="$MONGOC_DIR/libmongocrypt-for-kms-divergence-check"
 
 # LIBMONGOCRYPT_GITREF is expected to refer to the version of libmongocrypt
 # where kms-message was last copied.
-LIBMONGOCRYPT_GITREF="a650d171ed3b552446095817ae2c5c4f7cec43a2"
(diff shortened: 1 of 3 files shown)


kevinAlbs:
Thank you for the contribution @MonkeybreadSoftware. The change LGTM. `src/kms-message` is vendored [from libmongocrypt](https://github.com/mongodb/libmongocrypt/tree/master/kms-message). I would rather make the change in libmongocrypt first, then sync the source in the C driver. The C driver tests check the copy of `src/kms-message` [does not diverge](https://github.com/mongodb/mongo-c-driver/blob/db54ccf2fe10f6577cf17feae46f9448cd21f3a4/.evergreen/scripts/kms-divergence-check.sh#L3) from libmongocrypt. If you are OK making the change in libmongocrypt, please do. Otherwise, we can make that change.

MonkeybreadSoftware:
Thanks.

I made a pull request over there:
https://github.com/mongodb/libmongocrypt/pull/923

I appreciate if you apply this and for future mongo-c-driver versions, I don't need to patch this locally.

kevinAlbs (approved):
Thank you for the fix. Changes were verified with this [patch build](https://spruce.mongodb.com/version/67605cd61437b0000740b236/).

PR #1816, opened 2024-12-30
New internal bounded string append APIs

mdbmes:
To efficiently support JSON serialization with bounded length, this is a new internal string API that can express bounded UTF-8-preserving append operations directly.

This patch on its own should have limited effect, but it's a prerequisite for work on CDRIVER-3775 and CDRIVER-4814.

The main benefit to this new API is that a single buffer can be shared by all nesting levels of a complex document. Previously, all nesting levels of a BSON document would allocate their own string append buffer, and truncation happened as a separate step on the way out of these nested levels.

Several other temporary string buffers were eliminated, including the temporary buffer that was needed for every printf. Now mcommon_string_append_printf writes directly to the destination, growing it as needed.

This work wasn't focused on performance, but it has an overall positive effect. The new mcommon_string implementations can be optimized more effectively at compile-time, and the updated string escape implementation treats fewer bytes as 'special'.

Other than the internal API change itself, the effects should be:
* bson_as_json_with_opts now always truncates output at a UTF-8 boundary (this behavior was unspecified before)
* bson_as_json serialization is overall faster, about 2x
* worst case performance of bson_utf8_escape_for_json is better by about 2x
* best case performance of bson_utf8_escape_for_json with plain ASCII data is worse by about 3x, due to an extra bson_utf8_validate pass that was needed to maintain API compatibility. The new internal equivalent for this API doesn't have the extra overhead.

The new internal APIs are documented in header comments. I've added test cases to cover UTF-8 truncation behavior, but most of the new code's test coverage is provided by existing tests.


eramongodb (changes requested):
Partial feedback. Still in review.

src/libmongoc/tests/TestSuite.c
```diff
@@ -1166,24 +1155,25 @@ MONGOC_PRINTF_FORMAT (1, 2)
...
-         fflush (gTestSuite->mock_server_log);
-      }
-
-      bson_free (formatted_msg);
+   } else if (gTestSuite->mock_server_log) {
+      va_list ap;
+      va_start (ap, msg);
+      vfprintf (gTestSuite->mock_server_log, msg, ap);
+      va_end (ap);
+      fwrite ("\n", 1, 1, gTestSuite->mock_server_log);
```
eramongodb, line 1175:
```suggestion
      fputc ('\n', gTestSuite->mock_server_log);
```

src/libmongoc/tests/TestSuite.c
```diff
@@ -203,7 +204,7 @@ TestSuite_Init (TestSuite *suite, const char *name, int argc, char **argv)
       } else if (!strcmp (mock_server_log, "stderr")) {
          suite->mock_server_log = stderr;
       } else if (!strcmp (mock_server_log, "json")) {
-         suite->mock_server_log_buf = mcommon_string_new (NULL);
+         suite->mock_server_log_buf = mcommon_string_new_with_capacity ("", 0, 4096);
```
eramongodb, line 207:
```suggestion
         suite->mock_server_log_buf = bson_malloc (sizeof (*suite->mock_server_log_buf));
         mcommon_string_append_new_fixed_capacity (&suite->mock_server_log_buf, 4096);
```
Consider refactoring `TestSuite::mock_server_log_buf` into an owning `mcommon_string_append_t *` to avoid underlying `bson_string_t` ownership confusion in other TestSuite-related functions below.

src/libmongoc/tests/TestSuite.c
```diff
@@ -506,8 +507,9 @@ TestSuite_RunFuncInChild (TestSuite *suite, /* IN */
    if (suite->mock_server_log_buf) {
       close (pipefd[1]);
       while ((nread = read (pipefd[0], buf, sizeof (buf) - 1)) > 0) {
-         buf[nread] = '\0';
-         mcommon_string_append (suite->mock_server_log_buf, buf);
+         mcommon_string_append_t append;
+         mcommon_string_append_init (&append, suite->mock_server_log_buf);
+         mcommon_string_append_bytes (&append, buf, nread);
```
eramongodb, line 512:
```suggestion
         mcommon_string_append_bytes (suite->mock_server_log_buf, buf, nread);
```
See suggestion above regarding `TestSuite::mock_server_log_buf`.

src/libmongoc/tests/TestSuite.c
```diff
@@ -1123,9 +1114,7 @@ TestSuite_Destroy (TestSuite *suite)
       fclose (suite->outfile);
    }
 
-   if (suite->mock_server_log_buf) {
-      mcommon_string_free (suite->mock_server_log_buf, true);
-   }
+   mcommon_string_destroy (suite->mock_server_log_buf);
```
eramongodb, line 1117:
```suggestion
   if (suite->mock_server_log_buf) {
      mcommon_string_append_destination_destroy (suite->mock_server_log_buf);
      bson_free (suite->mock_server_log_buf);
   }
```
See suggestion above regarding `TestSuite::mock_server_log_buf`.

src/libmongoc/src/mongoc/mongoc-write-command.c
```diff
@@ -989,15 +989,16 @@ _set_error_from_response (bson_t *bson_array,
...
    bson_iter_t array_iter;
    bson_iter_t doc_iter;
-   mcommon_string_t *compound_err;
    const char *errmsg = NULL;
    int32_t code = 0;
    uint32_t n_keys, i;
 
-   compound_err = mcommon_string_new (NULL);
+   mcommon_string_append_t compound_err;
+   mcommon_string_append_new_fixed_capacity (&compound_err, sizeof error->message - 1);
```
eramongodb, line 997:
```suggestion
   mcommon_string_append_new_fixed_capacity (&compound_err, sizeof (error->message) - 1u);
```

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -2239,7 +2238,9 @@ mongoc_uri_unescape (const char *escaped_string)
 
    ptr = escaped_string;
    end = ptr + len;
-   str = mcommon_string_new (NULL);
+
+   mcommon_string_append_t append;
+   mcommon_string_append_init (&append, mcommon_string_new_with_capacity ("", 0, len));
```
eramongodb, line 2243:
```suggestion
   mcommon_string_append_init_with_capacity (&append, len);
```
Consider adding a `mcommon_string_append_init_with_capacity` given the frequency of this pattern (default initialization with custom capacity).

src/libmongoc/src/mongoc/mongoc-topology-scanner.c
```diff
@@ -1171,30 +1171,30 @@ void
...
    mongoc_topology_scanner_node_t *node, *tmp;
-   bson_error_t *error = &ts->error;
-   mcommon_string_t *msg;
 
+   bson_error_t *error = &ts->error;
    memset (&ts->error, 0, sizeof (bson_error_t));
 
-   msg = mcommon_string_new (NULL);
+   mcommon_string_append_t msg;
+   mcommon_string_append_new_fixed_capacity (&msg, sizeof error->message - 1);
```
eramongodb, line 1179:
```suggestion
   mcommon_string_append_new_fixed_capacity (&msg, sizeof (error->message) - 1u);
```

src/libbson/src/bson/bson-iso8601.c
```diff
@@ -21,6 +21,7 @@
 #include <bson/bson-iso8601-private.h>
 #include <bson/bson-json.h>
 #include <bson/bson-timegm-private.h>
+#include <common-string-private.h>
```
eramongodb, line 24:
```suggestion
```
Stray include?

eramongodb (changes requested):

src/common/src/common-string-private.h
```diff
@@ -20,75 +20,537 @@
...
+ * @brief Truncate the string to zero length without deallocating the buffer
+ * @param string String to clear
+ */
+static BSON_INLINE void
+mcommon_string_clear (mcommon_string_t *string)
 {
-   BEGIN_IGNORE_DEPRECATIONS
-   return bson_string_free (string, free_segment);
-   END_IGNORE_DEPRECATIONS
+   string->len = 0;
```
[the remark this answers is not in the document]
mdbmes: Ok.

src/common/src/common-string-private.h
```diff
@@ -20,75 +20,537 @@
...
+ * The maximum string length will be set to the largest representable by the data type, UINT32_MAX - 1.
+ * The new string will be allocated with a small default capacity.
+ *
+ * This method is intended to be the most convenient way to start growing a string. If a reasonable guess
+ * can be made about the final size of the string, it's better to call mcommon_string_new_with_capacity()
+ * and mcommon_string_append_init() directly.
+ */
 static BSON_INLINE void
-mcommon_string_append_unichar (mcommon_string_t *string, bson_unichar_t unichar)
+mcommon_string_append_new (mcommon_string_append_t *new_append)
```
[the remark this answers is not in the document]
mdbmes: My intent with 'new' here was to focus on the new mcommon_string_t being created. A more verbose name might be mcommon_string_append_init_with_new_string, but at the time this seemed like a better naming compromise. I'm not sure the _init_with_string name serves well, because I'd expect nearly all of the mcommon_string_append constructors to take a mcommon_string_t as argument. Constructing a mcommon_string_t at the same time as the mcommon_string_append_t is a convenience, not a built-in feature of the string_append. I'll think about a clearer naming for these.

src/common/src/common-string-private.h
```diff
@@ -20,75 +20,537 @@
...
+   return !append->_max_len_exceeded;
+}
 
+/**
+ * @brief Get a reference to the string being appended to
+ * @param append Append operation, initialized with mcommon_string_append_init
+ * @returns Pointer to the mcommon_string_t being written to
+ */
+static BSON_INLINE mcommon_string_t *
+mcommon_string_append_destination (const mcommon_string_append_t *append)
```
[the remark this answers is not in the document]
mdbmes: I almost added this myself, then decided against it because it seemed wise to nudge code toward using mcommon_string_t more pervasively. I agree it would be a useful shortcut for the existing call sites.

src/common/src/common-string-private.h
```diff
@@ -20,75 +20,537 @@
...
-   BSON_ASSERT_PARAM (format);
+/**
+ * @brief Deallocate the mcommon_string_t destination associated with an mcommon_string_append_t and return its internal
+ * buffer
+ * @param append Append operation, initialized with mcommon_string_append_init
+ * @returns A freestanding NUL-terminated string in a buffer that must be freed with bson_free()
+ * The append operation will no longer be usable after this call.
+ */
+static BSON_INLINE char *
+mcommon_string_append_destination_destroy_into_buffer (const mcommon_string_append_t *append)
```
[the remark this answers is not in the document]
mdbmes: Sure

src/common/src/common-string-private.h
```diff
@@ -20,75 +20,537 @@
...
+
+/* Parameters and outcome for a bounded append operation on a mcommon_string_t. Individual type-specific append
+ * functions can consume this struct to communicate bounds info. "max_len_exceeded" can be tested any time an
+ * algorithmic exit is convenient; the actual appended content will be limited by max_len. Truncation is guaranteed not
+ * to split a valid UTF-8 byte sequence.
+ *
+ * Members are here to support inline definitions; not intended for direct access.
+ *
+ * Multiple mcommon_string_append_t may simultaneously refer to the same 'string' but this usage is not recommended.
+ *
```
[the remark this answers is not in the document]
mdbmes: I'm not sure this effectively supports existing use cases. I've been trying to steer toward a best practice where the mcommon_append_t is long-lived, but when that's not practical we need to be able to construct a mcommon_string_append_t on the stack just for a few append operations, using a pointer to an effectively borrowed mcommon_string_t.

src/common/src/common-string-private.h
```diff
@@ -20,75 +20,537 @@
...
+}
+
+/**
+ * @brief Deallocate the mcommon_string_t destination associated with an mcommon_string_append_t
+ * @param append Append operation, initialized with mcommon_string_append_init
+ * The append operation will no longer be usable after this call.
+ */
 static BSON_INLINE void
-mcommon_string_append_printf (mcommon_string_t *string, const char *format, ...)
+mcommon_string_append_destination_destroy (const mcommon_string_append_t *append)
```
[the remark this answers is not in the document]
mdbmes: I like code that tells you exactly what it's doing when possible, and the mcommon_string_append_t itself really does not need destroying, only the mcommon_string_t it may point to. I don't think it's appropriate to treat this like it's a deallocator for the string_append, but I'll think about a clearer naming.

src/libmongoc/tests/test-mongoc-ssl.c
```diff
@@ -186,7 +188,7 @@ test_mongoc_ssl_opts_from_bson (void)
       ASSERT (!ssl_opt.crl_file);
 
       _mongoc_ssl_opts_cleanup (&ssl_opt, true /* free_internal */);
-      mcommon_string_free (errmsg, true /* free_segment */);
+      mcommon_string_destroy (errmsg);
```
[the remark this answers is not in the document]
mdbmes: Sure

src/common/src/common-string-private.h
```diff
@@ -20,75 +20,537 @@
...
+ * and mcommon_string_append_init() directly.
+ */
 static BSON_INLINE void
-mcommon_string_append_unichar (mcommon_string_t *string, bson_unichar_t unichar)
+mcommon_string_append_new (mcommon_string_append_t *new_append)
 {
-   BEGIN_IGNORE_DEPRECATIONS
-   bson_string_append_unichar (string, unichar);
-   END_IGNORE_DEPRECATIONS
+   mcommon_string_append_init (new_append, mcommon_string_new_with_capacity ("", 0, 32));
```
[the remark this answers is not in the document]
mdbmes: I did consider adding an initializer macro, but I'm not sure it makes the code clearer. It's less typing, but it's also more opaque.

src/libmongoc/tests/TestSuite.c
mdbmes: The mcommon_string_t* here is optional, and it didn't seem worth moving its optional status to a separate flag. I'm not sure I know what you mean by 'ownership confusion'. Is there a flaw in the ownership as written, or do you mean that mcommon_string_append_t does not necessarily own the mcommon_string_t?

src/libmongoc/src/mongoc/mongoc-write-command.c
mdbmes: What's the reason for the extra parens here?

src/libmongoc/src/mongoc/mongoc-uri.c
mdbmes: Agree that a shortcut would be helpful. I'll think about a name.. I'd like it to be clear that a mcommon_string_t is being allocated.

src/libbson/src/bson/bson-iso8601.c
mdbmes: Yes, thanks.

src/libmongoc/src/mongoc/mongoc-write-command.c
eramongodb: I was going to say code consistency, but it seems around 25% of our use of `sizeof` does not use parenthesis. Anecdotally I believe most recently-committed code includes the parenthesis for better readability and avoiding syntactic confusion ("like a function", consistent with `_Alignof`/`_Alignas`/etc.). The C99 draft also [often](https://www.iso-9899.info/n1570.html#6.7.2.1p20) (but not quite always) uses the parenthesis form in its examples, e.g.:

```
22   After the further declaration:
              struct ss { int n; };
     the expressions:
              sizeof (struct s) >= sizeof (struct ss)
              sizeof (struct s) >= offsetof(struct s, d)
     are always equal to 1.
```

src/libmongoc/tests/TestSuite.c
eramongodb: "Ownership confusion" was referring to the use of `_init` without an accompanying `_destroy` in the while loop within `TestSuite_RunFuncInChild` (re: code smell; see the code suggestion below which removes the call to `_init`) and the suggestion to remove the allowance for non-unique-ownership of the underlying `mcommon_string_t` object (no "owning" vs. "non-owning" pointer ambiguity).

If this "deliberately view-like" usage of `mcommon_string_append_t` is sufficiently compelling, I suggest avoiding the `_init` function altogether and directly initialize the data members instead (the decision not to use `_init` communicating a deliberate intent to bypass usual semantics of `_init`, including ownership of the provided `mcommon_string_t` argument):

```c
while ((nread = read (pipefd[0], buf, sizeof (buf) - 1)) > 0) {
   mcommon_string_append_bytes (
      &(mcommon_string_append_t) {
         ._string = suite->mock_server_log_buf,
         ._max_len = UINT32_MAX - 1u,
      },
      buf,
      nread);
}
```

src/common/src/common-string-private.h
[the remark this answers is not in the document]
eramongodb: The "with_string" suggestion was to distinguish `_init(&astr)` (default initialization, newly allocated `mcommon_string_t`) vs. `_init(&astr, str)` (initialization that takes ownership of a user-provided `mcommon_string_t`).

```c
mcommon_string_append_t astr;
mcommon_string_append_init (&astr); // "Default" initialization.
```

```c
mcommon_string_append_t astr;
mcommon_string_append_init_with_string (
  &astr,
  mcommon_string_new (...) // User-provided `mcommon_string_t`.
);
```

Note, there are currently 45 instances of `mcommon_string_append_new` (proposed: `mcommon_string_append_init`) in the updated test code vs. 13 instances of `mcommon_string_append_init` (proposed: `mcommon_string_append_init_with_string`).

src/common/src/common-string-private.h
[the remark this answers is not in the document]
eramongodb: I do not consider this to be a blocker. The convenience macro can be added later if sufficiently compelled.

src/libmongoc/src/mongoc/mongoc-write-command.c
mdbmes: Is there a coding standard we're heading toward that prefers extra parens after sizeof? If not, I prefer to avoid them, and to emphasize that sizeof is an operator.

src/libmongoc/tests/TestSuite.c
mdbmes: I was really hoping to create an interface here and not just expose all the members of the struct mcommon_string_append_t. An inline function that writes to already-allocated stack memory really seems like the right implementation for this, but I'll rename it to avoid confusion with the expectations around 'init'.

eramongodb: To clarify, the suggestion to avoid `_init` was specifically for this instance (and other occasional instances that deliberately subvert unique ownership semantics), not a suggestion to remove `_init` altogether (which I am in favor of using by default).

The suggestion to initialize data members directly would help make clearly visible situations where unique ownership is expected (via the proposed, recommended interfaces) vs. comparatively exceptional (but nevertheless compelling) situations where `mcommon_string_append_t` is treated as a non-owning append handler (by taking advantage of the visibility of the struct data members in library internal code).

Regarding the initial suggestion to refactor `TestSuite::mock_server_log_buf`, I am fine with leaving it as-is in favor of the "deliberately view-like" usage pattern (bypassing `_init`) suggested above.

src/libmongoc/src/mongoc/mongoc-write-command.c
eramongodb: AFAIK no. Given this is not already being enforced either way by our ClangFormat settings, I am fine with treating this as a non-blocker for this PR.

src/common/src/common-json.c
```diff
@@ -0,0 +1,766 @@
...
+                                    const char *begin_non_empty,
+                                    const char *end_non_empty,
+                                    const char *omitted)
+{
+   // Note that the return value here is bson validity, not append status.
+   if (bson_empty (bson)) {
+      (void) mcommon_string_append (append, empty);
+      return true;
+   } else if (max_depth == 0) {
+      (void) mcommon_string_append (append, omitted);
```
kevinAlbs, line 384: Is `max_depth` expected to be 0? If no, consider using `BSON_ASSERT (max_depth > 0);` to remove an unused special case.

src/common/src/common-string-private.h
```diff
@@ -20,75 +20,582 @@
...
+typedef struct mcommon_string_t {
+   char *str;
+   uint32_t len;
+   uint32_t alloc;
+} mcommon_string_t;
+
+/* Parameters and outcome for a bounded append operation on a mcommon_string_t. Individual type-specific append
+ * functions can consume this struct to communicate bounds info. "max_len_exceeded" can be tested any time an
+ * algorithmic exit is convenient; the actual appended content will be limited by max_len. Truncation is guaranteed not
+ * to split a valid UTF-8 byte sequence.
```
kevinAlbs, line 51: I expect the behavior of truncation in `bson_as_json_with_opts` was previously unspecified. Request add an entry to `src/libbson/NEWS` to note the possible output change:
```md
libbson 1.30.0 (Unreleased)
===========================

Fixes:

  * Truncated output of `bson_as_json_with_opts` is changed to no longer split valid UTF-8 sequences.
```

src/common/src/common-json.c
```diff
@@ -0,0 +1,766 @@
...
+
+/**
+ * @brief Measure the number of consecutive non-special bytes.
+ */
+static BSON_INLINE uint32_t
+mcommon_json_append_escaped_count_non_special_bytes (const char *str, uint32_t len)
+{
+   uint32_t result = 0;
+   // Good candidate for architecture-specific optimizations.
+   // SSE4 strcspn is nearly what we want, but our table of special bytes would be too large (34 > 16)
```
kevinAlbs, line 453: Curious: what is significant about 16?

src/common/src/common-string.c
```diff
@@ -0,0 +1,374 @@
...
+   } else if (max_append_len == 0) {
+      // Truncation to a zero-length append
+      mcommon_string_append_set_overflow (append);
+      return false;
+   } else {
+      /* We expect to append at least one byte, and truncate.
+       * Encoding only produces single-byte UTF-8 sequences, so the result always has exactly the maximum length.
+       *
+       * mcommon_b64_ntop() can't truncate without failing. To do this without allocating a full size temporary buffer
+       * or rewriting mcommon_b64_ntop, we can parititon the write into three parts: a 'direct' portion made from entire
```
kevinAlbs, line 217:
```suggestion
       * or rewriting mcommon_b64_ntop, we can partiton the write into three parts: a 'direct' portion made from entire
```

kevinAlbs (approved):
Remarkably, I only have minor comments. LGTM. Great work.

src/common/src/common-json.c
mdbmes: In the case of strcspn(), the "reject" string can fit in a single 128-bit SSE4 register if it's 16 bytes or less. One SSE4 "pcmpistri" instruction can search a 16 byte register for any occurrence of each of the 16 bytes in the another register.

For reference, see the SSE4 text processing instructions (https://en.wikibooks.org/wiki/X86_Assembly/SSE#Text_Processing_Instructions) and the glibc source (./sysdeps/x86_64/multiarch/strcspn-sse4.c)

src/common/src/common-string-private.h
mdbmes: Will do. Yes, my understanding is that the behavior was unspecified before.

src/libbson/tests/test-string.c
```diff
@@ -358,6 +384,33 @@ test_bson_string_truncate (void)
...
+      ASSERT_CMPSTR (str->str, "\x41\xe2\x89");
+      ASSERT_CMPSIZE_T (str->len, ==, 3u);
+      ASSERT_CMPSIZE_T (str->alloc, ==, 4u);
+      bson_string_free (str, true);
+   }
+}
+
+// Compatibility wrapper; deprecated private API.
+bson_string_t *
+_bson_string_alloc (const size_t size)
```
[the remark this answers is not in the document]
kevinAlbs: I think remove. I see no reason to preserve test coverage. Test code using `_bson_string_alloc` and `_bson_string_append_ex` appear only to be test the functions themselves.

mdbmes:
Ok, I'll need a minute to write a polished patch for this but I have a proposal for a new naming to make the ownership of strings and string append operations clearer.

I considered renaming the string_append to be an "appender", but my main frustration with this is that it implies the existence of methods like "string_appender_append" and "string_appender_printf", whereas I'd rather treat the "append" as the conceptual object that various appending methods act on. The mcommon_string_append() method itself feels conceptually like operator() on a string_append object. I think this makes the naming of most actual append operations clearer, if the confusion about lifetimes can be cleared up.

I'm avoiding the word "init" in describing how we come up with a mcommon_string_append_t. Something like "begin" or "do" could work, describing the expected placement of the operation, but I'd prefer to be more straightforward about what's happening. So, I'm considering that instead of a mcommon_string_append_t constructor, this could be written as a mcommon_string method that "sets" an append to itself. So `mcommon_string_set_append(string, new_append)`, `mcommon_string_set_append_with_limit(string, new_append, max_len)`. With string as the first param. And this means when we actually allocate a string, it can be written as a special type of string_new, like `mcommon_string_new_as_append()`, `mcommon_string_new_as_fixed_capacity_append()`.

src/libmongoc/tests/TestSuite.c
mdbmes: Took another look at this and I still think the existing mock_server_log_buf pointer is preferable to a rewrite that propagates the string_append_t up higher. I think the concern about 'init' has been addressed by changing the naming convention to 'set'.

eramongodb (approved):
The subject-object reversal of `string` and `append` is a much appreciated improvement. Thank you very much! One last suggestion remaining; otherwise, LGTM.

PR #1822, opened 2025-01-09
[CDRIVER-5611] Refactor of validation fuzz test to expand coverage and speed

vector-of-bool:
# Summary

The existing `fuzz_test_validate` suffered from two issues that made fuzzing less than optimal:

1. We unconditionally pass all validation flags to the `bson_validate`, which will reduce the branch coverage that is reachable by the fuzzer.
2. The fuzz input is treated as a full BSON document, and would immediately reject if the header/trailer was invalid. This can massively slow down the fuzzer's ability to discover valid inputs, since inserting or removing any bytes from the test data would immediately reject in `bson_init_static` until the fuzzer would coincidentally discover the need to update the document header to match the new data size. Fuzzing would stagnate at around 475 corpus items.

This changeset addresses both issues:

1. The first byte of test data is treated as a set of flags that control the behavior of `bson_validate`. This allows the fuzzer to test multiple branches within validation.
2. The remaining bytes of test data are treated as the inner contents of a BSON object. A new valid header and trailer is constructed on-the-fly to wrap the data, ensuring that `bson_init_static` will always succeed and we can actually get to testing `bson_validate`. This allows the fuzzer to generate over 500 valid inputs in the first 30 seconds of running (with `-fork=32`).

Additionally:

1. This changeset introduces a [libFuzzer dictionary][dict] of valid/invalid BSON byte snippets that the fuzzer can use to insert into test data. Because generating these byte sequences by hand is non-trivial, a small Python script generates this file.
2. Because test input is no longer a simple BSON doc, an extra exe is added that can be used to execute `bson_validate` on a corpus item directly. This can be used to repro any issue discovered by the fuzz tester.

[dict]: https://llvm.org/docs/LibFuzzer.html#dictionaries


src/libbson/fuzz/CMakeLists.txt
```diff
@@ -5,8 +5,10 @@ set_property(TARGET fuzz_test_properties APPEND PROPERTY INTERFACE_LINK_OPTIONS
 add_executable(fuzz_test_init_from_json EXCLUDE_FROM_ALL fuzz_test_init_from_json.c)
 target_link_libraries(fuzz_test_init_from_json PRIVATE fuzz_test_properties)
 
-add_executable(fuzz_test_validate EXCLUDE_FROM_ALL fuzz_test_validate.c)
+add_executable(fuzz_test_validate EXCLUDE_FROM_ALL fuzz_test_validate.cpp)
```
eramongodb, line 8: Recommend updating the description of the `MONGO_FUZZ` CMake option to include "(requires C++) to document the additional toolchain dependency required to build fuzz testing targets. Also consider guarding these targets with `if (CMAKE_CXX_COMPILER)` as done for the `mongoc-cxx-check` target.

eramongodb (approved):
Mostly minor feedback; otherwise, LGTM.

src/libbson/fuzz/CMakeLists.txt
```diff
@@ -5,8 +5,10 @@ set_property(TARGET fuzz_test_properties APPEND PROPERTY INTERFACE_LINK_OPTIONS
...
+add_executable(validate-repro EXCLUDE_FROM_ALL validate-repro.cpp)
```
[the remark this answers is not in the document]
vector-of-bool: Would it make more sense to switch everything else to be hyphenated? Most components are hyphenated currently.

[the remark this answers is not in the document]
eramongodb: SGTM.

Aside: a quick grep reveals the list of user-defined CMake targets (excluding those generated by the build system) which use `_` is currently limited to the following list:

```
bson_shared
bson_static
hello_mongoc
mcd_rpc
mongo_c_driver_examples
mongo_c_driver_tests
mongoc_shared
mongoc_static
```

src/libbson/fuzz/make-dicts.py
```diff
@@ -0,0 +1,313 @@
...
+            s.encode("ascii")  # Try to encode as ASCII
+        except ValueError:
+            # Byte is not valid UTF-8, or not valid ASCII
+            pass
+        else:
+            if s.isprintable():
+                yield s
+                continue
+        # Byte is not valid ASCII, or is not a printable char
+        yield f"\\x{u8:0>2x}"
```
kevinAlbs, line 297: Suggested simplification:

```suggestion
def escape(b: bytes) -> Iterable[str]:
    for u8 in b:
        s = chr(u8)
        if s.isascii() and s.isprintable():
            yield s
            continue
        # Byte is not valid ASCII, or is not a printable char
        yield f"\\x{u8:0>2x}"
```

src/libbson/fuzz/validate-repro.cpp
```diff
@@ -0,0 +1,33 @@
...
+      return 3;
+   }
+   std::stringstream sbuf;
+   sbuf << infile.rdbuf ();
+   auto s = sbuf.str ();
+   int rc = validate_one_input (reinterpret_cast<const std::uint8_t *> (s.data ()), s.size ());
+   if (rc == 0) {
+      std::fprintf (stderr, "Validation returned normally (no bug?)\n");
+   } else if (rc == -1) {
+      std::fprintf (stderr, "Test case is was rejected by the fuzzer (not a valid corpus item?)\n");
```
kevinAlbs, line 28:
```suggestion
      std::fprintf (stderr, "Test case is/was rejected by the fuzzer (not a valid corpus item?)\n");
```

PR #1825, opened 2025-01-13
CDRIVER-5850 prefix all error messages with source information

eramongodb:
## Summary

> [!IMPORTANT]
> This is an experimental PR exploring a possible design space. It is not necessarily expected to be merged. An alternative solution may be pursued instead if the changes proposed by this PR are considered undesirable.

Resolves CDRIVER-5850 as a prerequisite for [CXX-834](https://jira.mongodb.org/browse/CXX-834).

This PR proposes adding a prefix to all error messages returned by the C Driver libraries. This prefix may be used to reliably disambiguate the proper error category of the associated error code. This solution does not introduce any breaking changes beyond the contents of error messages themselves.

This PR is initially split into two separate commits: the first implements the proposed solution (prefixing all error messages), the second applies corresponding required changes in response to the modified error messages. This second commit may be used as a "case study" for the means and likelihood by which user code may be broken by this solution. If sufficiently concerning, we may defer this solution to an upcoming major release.

## Terminology

For clarity, the following terms are used in this PR description:

- **Error Value**: the integer value or enumerator describing an error.
    - e.g. `MONGOC_ERROR_QUERY_FAILURE`, `mongocxx::v_noabi::error_code::k_ssl_not_supported`.
- **Error Category**: the identity used to associate an error code value with its original domain.
    - e.g. `MONGOC_ERROR_SERVER` (`mongoc_error_domain_t`), `mongocxx::v_noabi::error_category()`.
- **Error Code**: an error value and its corresponding error category.
    - e.g. `bson_error_t` (`.domain` + `.code`), `std::error_code` (`.category()` + `.value()`).
- **Error Message**: a string describing an associated error code.

> [!NOTE]
> Integer values and enumerators which are typically referred to as "error codes" are deliberately referred to as "error values" in this PR description for consistency, e.g. ["server error values"](https://www.mongodb.com/docs/manual/reference/error-codes/) rather than "server error codes" when referring specifically to the list of documented integer values.

## Motivation

The goal of this PR is to explore a least-effort-required solution to address CXX-834: that is, to be able to reliably disambiguate the category of error code values.

The meaning of an error value can only be correctly interpreted if its category is known. The C Driver uses the `bson_error_t` type to communicate the triplet of category, value, and message as the `.domain`, `.code`, and `.message` data members respectively. Unfortunately, although the C Driver correctly ensures through this design that a value is always accompanied by its category, it did not sufficiently prepare to accomodate error codes originating from multiple libraries.

The C Driver currently uses `bson_error_t` to describe error codes coming from _at least_ four distinct sources:

- The bson library.
- The mongoc library.
- The libmongocrypt library.
- The MongoDB Server.

[CDRIVER-644](https://jira.mongodb.org/browse/CDRIVER-644) made an effort to disambiguate values coming from the server from those coming from mongoc by introducing an Error API Version which, when set to `MONGOC_ERROR_API_VERSION_2`, uses a dedicated `MONGOC_ERROR_SERVER` domain for server error values. Although this is an improvement, the situation remains complicated by further developments over time, such as Write Concern Errors ([CDRIVER-2900](https://jira.mongodb.org/browse/CDRIVER-2900)) and Client-Side Field Level Encryption through libmongocrypt ([CDRIVER-2875](https://jira.mongodb.org/browse/CDRIVER-2875)).

Most troubling however is the ever-growing list of domains that is supported by the `mongoc_error_domain_t` enumeration, as well as the overloaded list of error values described by `mongoc_error_code_t`. These enumerations do not sufficiently or reliably disambiguate error values coming from the various sources listed above. Issues include but are not limited to:

- Domain values for error codes from bson and mongoc are in conflict.
    - e.g. `BSON_ERROR_JSON` and `MONGOC_ERROR_CLIENT` both have a domain value of `1` ([CDRIVER-3660](https://jira.mongodb.org/browse/CDRIVER-3660)).
- Domain values are overloaded and describe values coming from multiple sources.
    - e.g. `MONGOC_ERROR_CLIENT_SIDE_ENCRYPTION` is used both for values coming from libmongocrypt (via `mongocrypt_status_code`) and from mongoc (e.g. `MONGOC_ERROR_CLIENT_INVALID_ENCRYPTION_ARG`).
    - e.g. `MONGOC_ERROR_SERVER` is used both for values coming from the server (both normal values _and_ write concern error values in `_parse_error_reply`) and from mongoc (e.g. `MONGOC_ERROR_QUERY_COMMAND_NOT_FOUND` in `_mongoc_cmd_check_ok`).
- Multiple domain values are used to describe unique error values coming the same source.
    - e.g. `MONGOC_ERROR_SERVER` and `MONGOC_ERROR_WRITE_CONCERN` domain values are both used for equivalent error values coming from the server.
    - e.g. `MONGOC_ERROR_CLIENT`, `MONGOC_ERROR_COLLECTION`, etc. redundantly document the _component_ from which an error value is associated despite the error values used with the domain value `MONGOC_ERROR_CLIENT_NOT_READY`, `MONGOC_ERROR_COLLECTION_DOES_NOT_EXIST`, etc. being unique to the source (mongoc).

This situation led to [CXX-834](https://jira.mongodb.org/browse/CXX-834), which documents the troubles plaguing the C++ Driver (and, by extension, users of the C Driver libraries) concerning error category disambiguation. I do not believe the "Error API Version" approach to extending the `mongoc_error_domain_t` enumeration with further domain values is desirable. Instead, this PR proposes a "least-effort-required" solution of embedding the error category within the error message as a prefix.

Although knowing the component from which an error originates may be situationally useful, this is often sufficiently obvious from context alone. Instead, being able to _easily_ and _unambiguously_ know how to _interpret_ an error value is of far greater importance for programmable error handling purposes. We do not want users (or ourselves, i.e. test code) to be pattern-matching the contents of error messages (which should be "for display only") to make decisions in code.

## Proposed Solution

At the moment, four error categories appear to be sufficient to describe all error codes returned by the C Driver API. These four categories are:

- `bson: ...`: maps to `BSON_ERROR_*` error values.
- `mongoc: ...`: maps to (most) `MONGOC_ERROR_*` error values.
- `server: ...`: maps to server error values (`MONGOC_ERROR_SERVER` and `MONGOC_ERROR_WRITE_CONCERN`).
- `crypt: ...`: maps to libmongocrypt error values (`MONGOC_ERROR_CLIENT_SIDE_ENCRYPTION`).

The documented source of a given error is _not_ used to identify the error category (see: overloaded domain values above). Instead, the _origin_ of the value used to set the `bson_error_t::code` data member is used instead.

All instances of assigning an error value to `bson_error_t::code` were found by grepping either for a call to `bson_set_error` or the pattern `(\.|->)code =` (direct assignment).

### bson

This error category was the most straightforward to identify: any call to `bson_set_error` within bson library code naturally uses the bson error category. No other error category is present within the bson library, aside from `errno` and POSIX error codes, neither of which is directly assigned to `bson_error_t::code`.

The error messages in all such bson library code were manually prefixed with `"bson: "`. The error messages have otherwise been left unchanged.

### mongoc

This error category required the most effort to identify and prefix. Rather than manually prefixing all error messages in mongoc, a new `mongoc_set_error` internal function is used instead to consistently apply the `"mongoc: "` prefix to error messages (a straightforward substitution of calls to `bson_set_error`).

> [!NOTE]
> `mongoc_set_error` cannot be a macro due to some calls providing no argument for `...`, e.g.:
> ```c
> // error: passing no argument for the '...' parameter of a variadic macro is a C23 extension [-Werror,-Wvariadic-macro-arguments-omitted]
> mongoc_set_error(error, domain, code, "mongoc: message");
> ```

Some areas of concern include:

- Several authentication-related error handling code sets `MONGOC_ERROR_CLIENT_AUTHENTICATE` without explicitly setting a corresponding error message (already set beforehand in another context?):
    - `_stream_run_hello`
    - `_mongoc_cluster_auth_node_plain`
    - `_mongoc_cluster_auth_node_x509`
    - `_mongoc_cluster_run_scram_command`
    - `_mongoc_cluster_auth_scram_start`
    - `mongoc_server_description_handle_hello`
- Server error codes are overwritten by mongoc error codes in `_mongoc_cmd_check_ok`.
- The origin of the error message reported by `mongoc_topology_select_server_id` when setting `MONGOC_ERROR_SERVER_SELECTION_FAILURE` is uncertain.

### server

This error category is used whenever an error value coming directly from the server is assigned to `bson_error_t::code`. There appears to be the case in three distinct locations:

- `_mongoc_cmd_check_ok_no_wce` when `_parse_error_reply` returns a non-zero error value obtained from either `reply: {"code": <value> }` or `reply: {"writeConcernError": {"code": <value>}}`.
- `_mongoc_write_result_complete` when `_set_error_from_response` obtains an error value from the first operation which responded with either `reply: {"writeErrors": {"code": <value>}}` or `reply: {"writeConcernErrors": {}}`.
- `_mongoc_client_command_with_opts`, `mongoc_collection_create_index_with_opts`, and `_mongoc_cursor_run_command` when `_mongoc_parse_wc_err` obtains an error value from `reply: {"WriteConcernError": {"code": <value>}}`.

### crypt

This error category is only used in `_status_to_error`, where `mongocrypt_status_code` is used to set `bson_error_t::code`. Use of this function is limited to the `*_check_error` functions provided by `mongoc-crypt.c`. All other usage of the `MONGOC_ERROR_CLIENT_SIDE_ENCRYPTION` domain value appear to describe mongoc error codes, not libmongocrypt error codes.

## Concerns

This proposed solution may potentially break any user code which depends on error message stability for error handling. Examples of such code in our own codebase can be seen as diffs in the second commit, "Update dependent code with new error message prefixes".

Most code which searches for _substrings_ within the error message are unaffected, as their contents remain unchanged. However, code which performs _equality-comparison_ or _starts-with_ operations on error messages are broken by the presence of the prefix, either at the start of the error message or embedded within nested error messages, e.g. `mongoc: invalid argument for update: bson: keys cannot contain ".": "a.a"`.

It is also likely that, given this solution is deliberately designed to support error categories reported by the C++ Driver, users may become reliant on this behavior as well (this is not a problem limited to the C++ Driver), which directly conflicts with our desire to avoid encouraging error handling programmed on error messages. We may need to add documentation loudly warning users against depending on the stability of error messages (reserved only for ourselves).

## Alternatives

An alternative solution, which would require far more significant API breaking changes and internal refactors (thus not demonstrated via PR as done here), but may nevertheless be considered preferable, would be to add a new data member whose purpose is to function as an equivalent to `std::error_code::category()` that is properly distinct from `mongoc_error_domain_t`. This addition can be limited to the mongoc library by the use of a new `mongoc_error_t` type that wraps or replaces `bson_error_t`. This alternative solution is tracked by CDRIVER-5854.

```c
// `bson_error_t` implies `bson_category()`.

typedef struct _mongoc_error_t {
   const void *category; // a la `std::error_category const&`.
   uint32_t domain;      // Do we still want/need this?
   uint32_t code;
   char message[MONGOC_ERROR_BUFFER_SIZE]; // Up to 512? Higher? Alloc instead?
} mongoc_error_t;

BSON_EXPORT (const void*) mongoc_category();
BSON_EXPORT (const void*) server_category();
BSON_EXPORT (const void*) crypt_category();
```

As we probably do not have interest in fully reimplementing all the features of `std::error_code` in C++ (in particular, the flexibility of `std::error_condition`), the `mongoc_error_t::domain` data member could be repurposed/used as the error category specifier instead, e.g.:

```c
typedef struct _mongoc_error_category_t {
  MONGOC_ERROR_CATEGORY_MONGOC = 1,
  MONGOC_ERROR_CATEGORY_SERVER,
  MONGOC_ERROR_CATEGORY_CRYPT,
};

// For all mongoc error values.
// "Source component" info is embedded in the value/identifier, not in a "domain" value.
// Discard "neat" grouping of values, just expand with new values on-demand
// as new errors are required by the library and deprecate old values when no longer used.
typedef struct _mongoc_error_code_t {
  MONGOC_ERROR_CODE_CLIENT_REASON_A = 1,
  MONGOC_ERROR_CODE_CLIENT_REASON_B,
  MONGOC_ERROR_CODE_COLLECTION_REASON_A,
  MONGOC_ERROR_CODE_COLLECTION_REASON_B,
  ...
  MONGOC_ERROR_CODE_MAX_VALUE = UINT32_MAX, // or INT32_MAX if we want signed values?
};

typedef struct _mongoc_error_t {
   _mongoc_error_category_t category;
   uint32_t code; // or `int32_t` for signedness consistency with `int`?
   char message[MONGOC_ERROR_BUFFER_SIZE]; // Up to 512? Higher? Alloc instead?
} mongoc_error_t;
```

> [!NOTE]
> The 512 byte size of `bson_error_t` with an embedded message array has been stable since its introduction in [Aug 2013](https://github.com/mongodb/libbson/commit/efd80c2a897af20d8ebee1f1d8de0e090d685af3). We may want to expand the size of this structure to support longer error messages (i.e. embedded BSON documents) or revert to using an allocated resizeable string buffer instead.

### DIFF
--- a/build/opts_templates/mongoc-opts.c.template
+++ b/build/opts_templates/mongoc-opts.c.template
@@ -1,6 +1,7 @@
 #include "mongoc-opts-helpers-private.h"
 #include "mongoc-opts-private.h"
 #include "mongoc-error.h"
+#include "mongoc-error-private.h"
 #include "mongoc-util-private.h"
 #include "mongoc-client-private.h"
 
@@ -54,10 +55,10 @@ _{{ struct_name }}_parse (
    }
 
    if (!bson_iter_init (&iter, opts)) {
-      bson_set_error (error,
-                      MONGOC_ERROR_BSON,
-                      MONGOC_ERROR_BSON_INVALID,
-                      "Invalid 'opts' parameter.");
+      _mongoc_set_error (error,
+                        MONGOC_ERROR_BSON,
+                        MONGOC_ERROR_BSON_INVALID,
+                        "Invalid 'opts' parameter.");
       return false;
    }
 
@@ -100,18 +101,18 @@ _{{ struct_name }}_parse (
                &{{ struct_name }}->extra,
                bson_iter_key (&iter),
                bson_iter_value (&iter))) {
-            bson_set_error (error,
-                            MONGOC_ERROR_BSON,
-                            MONGOC_ERROR_BSON_INVALID,
-                            "Invalid 'opts' parameter.");
+            _mongoc_set_error (error,
+                              MONGOC_ERROR_BSON,
+                              MONGOC_ERROR_BSON_INVALID,
+                              "Invalid 'opts' parameter.");
             return false;
          }
 {% else %}
-         bson_set_error (error,
-                         MONGOC_ERROR_COMMAND,
-                         MONGOC_ERROR_COMMAND_INVALID_ARG,
-                         "Invalid option '%s'",
-                         bson_iter_key (&iter));
+         _mongoc_set_error (error,
+                           MONGOC_ERROR_COMMAND,
+                           MONGOC_ERROR_COMMAND_INVALID_ARG,
+                           "Invalid option '%s'",
+                           bson_iter_key (&iter));
          return false;
 {% endif %}
       }
--- a/src/libbson/src/bson/bson-iso8601.c
+++ b/src/libbson/src/bson/bson-iso8601.c
@@ -126,9 +126,9 @@ _bson_iso8601_date_parse (const char *str, int32_t len, int64_t *out, bson_error
 
    struct bson_tm posix_date = {0};
 
-#define DATE_PARSE_ERR(msg)                                                                                     \
-   bson_set_error (                                                                                             \
-      error, BSON_ERROR_JSON, BSON_JSON_ERROR_READ_INVALID_PARAM, "Could not parse \"%s\" as date: " msg, str); \
+#define DATE_PARSE_ERR(msg)                                                                                           \
+   bson_set_error (                                                                                                   \
+      error, BSON_ERROR_JSON, BSON_JSON_ERROR_READ_INVALID_PARAM, "bson: Could not parse \"%s\" as date: " msg, str); \
    return false
 
 #define DEFAULT_DATE_PARSE_ERR                                                 \
--- a/src/libbson/src/bson/bson-json.c
+++ b/src/libbson/src/bson/bson-json.c
@@ -417,15 +417,21 @@ _bson_json_read_set_error (bson_json_reader_t *reader, /* IN */
                            const char *fmt,            /* IN */
                            ...)
 {
-   va_list ap;
+   const char prefix[] = "bson: ";
+   const size_t prefix_len = sizeof (prefix) - 1u; // Exclude null terminator.
 
    if (reader->error) {
       reader->error->domain = BSON_ERROR_JSON;
       reader->error->code = BSON_JSON_ERROR_READ_INVALID_PARAM;
+
+      size_t remaining = sizeof reader->error->message;
+      bson_snprintf (reader->error->message, remaining, "%s", prefix);
+      remaining -= prefix_len;
+
+      va_list ap;
       va_start (ap, fmt);
-      bson_vsnprintf (reader->error->message, sizeof reader->error->message, fmt, ap);
+      bson_vsnprintf (reader->error->message + prefix_len, remaining, fmt, ap);
       va_end (ap);
-      reader->error->message[sizeof reader->error->message - 1] = '\0';
    }
 
    reader->bson.read_state = BSON_JSON_ERROR;
@@ -442,13 +448,20 @@ _bson_json_read_corrupt (bson_json_reader_t *reader, /* IN */
                          const char *fmt,            /* IN */
                          ...)
 {
-   va_list ap;
+   const char prefix[] = "bson: ";
+   const size_t prefix_len = sizeof (prefix) - 1u; // Exclude null terminator.
 
    if (reader->error) {
       reader->error->domain = BSON_ERROR_JSON;
       reader->error->code = BSON_JSON_ERROR_READ_CORRUPT_JS;
+
+      size_t remaining = sizeof reader->error->message;
+      bson_snprintf (reader->error->message, remaining, "%s", prefix);
+      remaining -= prefix_len;
+
+      va_list ap;
       va_start (ap, fmt);
-      bson_vsnprintf (reader->error->message, sizeof reader->error->message, fmt, ap);
+      bson_vsnprintf (reader->error->message + prefix_len, remaining, fmt, ap);
       va_end (ap);
       reader->error->message[sizeof reader->error->message - 1] = '\0';
    }
@@ -1815,7 +1828,7 @@ _bson_json_unescape (bson_json_reader_t *reader, struct jsonsl_state_st *state, 
       bson_set_error (reader->error,
                       BSON_ERROR_JSON,
                       BSON_JSON_ERROR_READ_CORRUPT_JS,
-                      "error near position %d: \"%s\"",
+                      "bson: error near position %d: \"%s\"",
                       (int) state->pos_begin,
                       jsonsl_strerror (err));
       return false;
@@ -1970,7 +1983,7 @@ _error_callback (jsonsl_t json, jsonsl_error_t err, struct jsonsl_state_st *stat
    bson_set_error (reader->error,
                    BSON_ERROR_JSON,
                    BSON_JSON_ERROR_READ_CORRUPT_JS,
-                   "Got parse error at \"%c\", position %d: \"%s\"",
+                   "bson: Got parse error at \"%c\", position %d: \"%s\"",
                    *errat,
                    (int) json->pos,
                    jsonsl_strerror (err));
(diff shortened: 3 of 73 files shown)


mdbmes:
This is neat. I will come back with more detailed feedback, but my initial impressions:

It seems really important at this point to know exactly how much stability apps expect for error messages. It does seem inadvisable for apps to rely on the exact message contents. Assuming we can change the messages at all, I like your proposed solution of adding a human-readable prefix.

While I was reading this, an additional solution occurred to me: There could be a new variant of bson_error_t which has the same size but a new interpretation, where 'message' is a nul-terminated list of nul-terminated strings, rather than a single string. We could use this to supply extensible error information in a fully ABI-compatible way. Old apps would see the first string in the list only. Newer apps would know to read additional strings.

eramongodb:
> There could be a new variant of bson_error_t which has the same size but a new interpretation, where 'message' is a nul-terminated list of nul-terminated strings, rather than a single string. We could use this to supply extensible error information in a fully ABI-compatible way.

I am very interested in this idea and consider it an improvement over the currently-proposed prefix method. The only concern I have at the moment is that this may further exacerbate error message truncation issues, especially for error messages containing embedded BSON documents, but this might not be a situation that "typically" occurs with normal error messages returned by the library (limited to our test code). I will explore this solution to see if it can be implemented in a manner similar to this PR.

eramongodb:
Succeeded by https://github.com/mongodb/mongo-c-driver/pull/1834.

PR #1826, opened 2025-01-15
CDRIVER-4485 string append and truncation fixes for structured logging

mdbmes:
This PR is a follow-up for #1795 which depends on the new internal string append and json building API from #1816. Note that #1821 (still out for review) will also need a similar follow-up that isn't included in this patch.

* Refines the concept of "maximum max document length" internally as a specific value near INT_MAX, to account for both ellipsis and bson overhead. Currently we choose not to expose the specific value of this constant publicly.
* Adds public APIs for the max_document_length within a mongoc_structured_log_opts_t.
* Addresses the remainder of CDRIVER-4485: JSON documents within structured log messages are truncated correctly according to the rules set out in the Logging specification.
* Addresses the remainder of CDRIVER-4486 by implementing prose tests.
* Addresses CDRIVER-4814. Command or payload data beyond the truncation limit no longer degrades logging performance.

### DIFF
--- a/src/libmongoc/doc/mongoc_structured_log_opts_get_max_document_length.rst
+++ b/src/libmongoc/doc/mongoc_structured_log_opts_get_max_document_length.rst
@@ -1,0 +1,27 @@
+:man_page: mongoc_structured_log_opts_get_max_document_length
+
+mongoc_structured_log_opts_get_max_document_length()
+====================================================
+
+Synopsis
+--------
+
+.. code-block:: c
+
+  size_t
+  mongoc_structured_log_opts_get_max_document_length (const mongoc_structured_log_opts_t *opts);
+
+Parameters
+----------
+
+* ``opts``: Structured log options, allocated with :symbol:`mongoc_structured_log_opts_new`.
+
+Returns
+-------
+
+Returns the current maximum document length set in ``opts``, as a ``size_t``. Always succeeds.
+This may be the last value set with :symbol:`mongoc_structured_log_opts_set_max_document_length` or it may be an environment variable captured by :symbol:`mongoc_structured_log_opts_set_max_document_length_from_env` or :symbol:`mongoc_structured_log_opts_new`.
+
+.. seealso::
+
+  | :doc:`structured_log`
--- a/src/libmongoc/src/mongoc/mongoc-structured-log.c
+++ b/src/libmongoc/src/mongoc/mongoc-structured-log.c
@@ -270,25 +270,46 @@ mongoc_structured_log_get_named_component (const char *name, mongoc_structured_l
    return false;
 }
 
-static uint32_t
-_mongoc_structured_log_get_max_document_length_from_env (void)
+size_t
+mongoc_structured_log_opts_get_max_document_length (const mongoc_structured_log_opts_t *opts)
 {
+   return (size_t) opts->max_document_length;
+}
+
+bool
+mongoc_structured_log_opts_set_max_document_length (mongoc_structured_log_opts_t *opts, size_t max_document_length)
+{
+   if (max_document_length <= (size_t) MONGOC_STRUCTURED_LOG_MAXIMUM_MAX_DOCUMENT_LENGTH) {
+      opts->max_document_length = (uint32_t) max_document_length;
+      return true;
+   } else {
+      return false;
+   }
+}
+
+bool
+mongoc_structured_log_opts_set_max_document_length_from_env (mongoc_structured_log_opts_t *opts)
+{
+   BSON_ASSERT_PARAM (opts);
+
    const char *variable = "MONGODB_LOG_MAX_DOCUMENT_LENGTH";
    const char *max_length_str = getenv (variable);
 
    if (!max_length_str) {
-      return MONGOC_STRUCTURED_LOG_DEFAULT_MAX_DOCUMENT_LENGTH;
+      return true;
    }
 
    if (!strcasecmp (max_length_str, "unlimited")) {
-      return MONGOC_STRUCTURED_LOG_MAXIMUM_MAX_DOCUMENT_LENGTH;
+      BSON_ASSERT (
+         mongoc_structured_log_opts_set_max_document_length (opts, MONGOC_STRUCTURED_LOG_MAXIMUM_MAX_DOCUMENT_LENGTH));
+      return true;
    }
 
    char *endptr;
    long int_value = strtol (max_length_str, &endptr, 10);
-   if (int_value >= 0 && int_value <= (long) MONGOC_STRUCTURED_LOG_MAXIMUM_MAX_DOCUMENT_LENGTH &&
-       endptr != max_length_str && !*endptr) {
-      return (uint32_t) int_value;
+   if (int_value >= 0 && endptr != max_length_str && !*endptr && mcommon_in_range_signed (size_t, int_value) &&
+       mongoc_structured_log_opts_set_max_document_length (opts, (size_t) int_value)) {
+      return true;
    }
 
    // Only log the first instance of each error per process
@@ -296,7 +317,7 @@ _mongoc_structured_log_get_max_document_length_from_env (void)
    if (0 == mcommon_atomic_int_fetch_add (&err_count_atomic, 1, mcommon_memory_order_seq_cst)) {
       MONGOC_WARNING ("Invalid length '%s' read from environment variable %s. Ignoring it.", max_length_str, variable);
    }
-   return MONGOC_STRUCTURED_LOG_DEFAULT_MAX_DOCUMENT_LENGTH;
+   return false;
 }
 
 bool
@@ -422,7 +443,8 @@ mongoc_structured_log_opts_new (void)
     * If environment variables can't be parsed, warnings will be logged once but we must, by specification,
     * continue to provide structured logging using whatever valid or default settings remain. */
    opts->default_handler_path = bson_strdup (getenv ("MONGODB_LOG_PATH"));
-   opts->max_document_length = _mongoc_structured_log_get_max_document_length_from_env ();
+   opts->max_document_length = MONGOC_STRUCTURED_LOG_DEFAULT_MAX_DOCUMENT_LENGTH;
+   (void) mongoc_structured_log_opts_set_max_document_length_from_env (opts);
    (void) mongoc_structured_log_opts_set_max_level_for_all_components (opts, MONGOC_STRUCTURED_LOG_DEFAULT_LEVEL);
    (void) mongoc_structured_log_opts_set_max_levels_from_env (opts);
 
--- a/src/libmongoc/src/mongoc/mongoc-structured-log.c
+++ b/src/libmongoc/src/mongoc/mongoc-structured-log.c
@@ -16,6 +16,8 @@
 
 #include "common-atomic-private.h"
 #include "common-oid-private.h"
+#include "common-string-private.h"
+#include "common-json-private.h"
 #include "common-thread-private.h"
 #include "mongoc-apm-private.h"
 #include "mongoc-error-private.h"
@@ -55,7 +57,7 @@ struct mongoc_structured_log_opts_t {
    mongoc_structured_log_func_t handler_func;
    void *handler_user_data;
    mongoc_structured_log_level_t max_level_per_component[STRUCTURED_LOG_COMPONENT_TABLE_SIZE];
-   int32_t max_document_length;
+   uint32_t max_document_length;
    char *default_handler_path;
 };
 
@@ -268,7 +270,7 @@ mongoc_structured_log_get_named_component (const char *name, mongoc_structured_l
    return false;
 }
 
-static int32_t
+static uint32_t
 _mongoc_structured_log_get_max_document_length_from_env (void)
 {
    const char *variable = "MONGODB_LOG_MAX_DOCUMENT_LENGTH";
@@ -279,13 +281,14 @@ _mongoc_structured_log_get_max_document_length_from_env (void)
    }
 
    if (!strcasecmp (max_length_str, "unlimited")) {
-      return BSON_MAX_LEN_UNLIMITED;
+      return MONGOC_STRUCTURED_LOG_MAXIMUM_MAX_DOCUMENT_LENGTH;
    }
 
    char *endptr;
    long int_value = strtol (max_length_str, &endptr, 10);
-   if (int_value >= 0 && int_value <= INT32_MAX && endptr != max_length_str && !*endptr) {
-      return (int32_t) int_value;
+   if (int_value >= 0 && int_value <= (long) MONGOC_STRUCTURED_LOG_MAXIMUM_MAX_DOCUMENT_LENGTH &&
+       endptr != max_length_str && !*endptr) {
+      return (uint32_t) int_value;
    }
 
    // Only log the first instance of each error per process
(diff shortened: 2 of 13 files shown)


src/libmongoc/tests/test-mongoc-structured-log.c
```diff
@@ -699,6 +737,22 @@ test_structured_log_component_names (void)
    ASSERT_CMPSTR (mongoc_structured_log_get_component_name (component), "connection");
 }
 
+void
+test_structured_log_max_document_length (void)
+{
+   mongoc_structured_log_opts_t *opts = mongoc_structured_log_opts_new ();
+
+   // Test assumes no environment variable option is set
```
eramongodb, line 745: Consider guarding on `std::getenv("MONGODB_LOG_MAX_DOCUMENT_LENGTH") != NULL` instead.

src/libmongoc/src/mongoc/mongoc-structured-log.c
```diff
@@ -269,32 +271,54 @@ mongoc_structured_log_get_named_component (const char *name, mongoc_structured_l
...
+      BSON_ASSERT (
+         mongoc_structured_log_opts_set_max_document_length (opts, MONGOC_STRUCTURED_LOG_MAXIMUM_MAX_DOCUMENT_LENGTH));
+      return true;
    }
 
    char *endptr;
    long int_value = strtol (max_length_str, &endptr, 10);
-   if (int_value >= 0 && int_value <= INT32_MAX && endptr != max_length_str && !*endptr) {
-      return (int32_t) int_value;
+   if (int_value >= 0 && endptr != max_length_str && !*endptr && mcommon_in_range_signed (size_t, int_value) &&
```
eramongodb, line 311:
```suggestion
   if (int_value >= 0 && endptr != max_length_str && *endptr == '\0' && mcommon_in_range_signed (size_t, int_value) &&
```
Recommend against treating characters as booleans.

eramongodb (changes requested):

src/libmongoc/tests/test-mongoc-command-logging-and-monitoring.c
```diff
@@ -0,0 +1,359 @@
...
+      ASSERT (bson_iter_init (&iter, log));
+      ASSERT (bson_iter_find (&iter, "message"));
+      ASSERT (BSON_ITER_HOLDS_UTF8 (&iter));
+      ASSERT_CMPSTR (bson_iter_utf8 (&iter, NULL), "Command succeeded");
+      ASSERT (bson_iter_find (&iter, "reply"));
+      ASSERT (BSON_ITER_HOLDS_UTF8 (&iter));
+      uint32_t reply_len;
+      const char *reply = bson_iter_utf8 (&iter, &reply_len);
+      ASSERT (reply);
+      printf ("rep %s\n", reply);
```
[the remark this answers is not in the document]
mdbmes: Whoops, that line wasn't supposed to be in the diff at all.

src/libmongoc/tests/test-mongoc-structured-log.c
mdbmes: Good idea.

src/libmongoc/src/mongoc/mongoc-structured-log.c
```diff
@@ -269,32 +271,54 @@ mongoc_structured_log_get_named_component (const char *name, mongoc_structured_l
...
+       mongoc_structured_log_opts_set_max_document_length (opts, (size_t) int_value)) {
+      return true;
    }
 
    // Only log the first instance of each error per process
    static int err_count_atomic;
    if (0 == mcommon_atomic_int_fetch_add (&err_count_atomic, 1, mcommon_memory_order_seq_cst)) {
```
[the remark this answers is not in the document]
mdbmes: Sure. Error counters seemed like a useful concept but without any way to read the value a flag is just as good and a little cheaper. This is identical to _mongoc_structured_log_get_log_level_from_env, I'll change both.

src/libmongoc/doc/mongoc_structured_log_opts_get_max_document_length.rst
```diff
@@ -0,0 +1,26 @@
...
+
+Parameters
+----------
+
+* ``opts``: Structured log options, allocated with :symbol:`mongoc_structured_log_opts_new`.
+
+Returns
+-------
+
+Returns the current maximum document length set in ``opts``, as a ``size_t``. Always succeeds.
```
eramongodb, line 22:
```suggestion
Returns the current maximum document length set in ``opts``, as a ``size_t``.
```
Redundant? (No "failure" mode possible with the declared interface.)

src/libmongoc/tests/test-mongoc-structured-log.c
```diff
@@ -742,9 +742,11 @@ test_structured_log_max_document_length (void)
 {
    mongoc_structured_log_opts_t *opts = mongoc_structured_log_opts_new ();
 
-   // Test assumes no environment variable option is set
-   ASSERT_CMPINT (
-      MONGOC_STRUCTURED_LOG_DEFAULT_MAX_DOCUMENT_LENGTH, ==, mongoc_structured_log_opts_get_max_document_length (opts));
+   // Test requires the environment variable option is unset, so we see the default.
+   ASSERT (getenv ("MONGODB_LOG_MAX_DOCUMENT_LENGTH") == NULL);
```
eramongodb, line 746: This will cause the test to fail if the environment (i.e. downstream packages and/or users) happen to have this env var enabled. Recommend runtime-conditioning the subsequent assertions (`if (getenv (...) == NULL) { ASSERT_CMPINT (...); ... }`) rather than asserting it is undefined.

eramongodb (approved):
One last suggestion; otherwise, LGTM.

src/libmongoc/doc/mongoc_structured_log_opts_get_max_document_length.rst
mdbmes: I thought it was a nice confirmation but yes it's also redundant. I'll remove it.

src/libmongoc/tests/test-mongoc-structured-log.c
mdbmes: I'm reluctant to have this test "pass" when it's not fully executing. I'll look for a way to mark it as skipped instead.

eramongodb: Oh, I see. In that case, I recommend using the `MONGOC_DEBUG` message pattern used by the [unified spec test runner](https://github.com/mongodb/mongo-c-driver/blob/1e4d7fbaa00b8529309e430d0dccebf4275fe5cf/src/libmongoc/tests/unified/runner.c#L1959) as reference.

mdbmes: We could go that route but I was hoping to have TestSuite actually mark the test as a skip rather than a pass. I think my evergreen is green enough, I'll go ahead and push what I've got. It's a little verbose but I think it has the right effect. I also found some getenv issues, and the patch fixes call sites to consistently use _mongoc_getenv() and to bson_free() the result.

eramongodb: Alas, unlike the C++ Driver which has Catch2 `SKIP()`, the C Driver test suite AFAIK does not have a uniform mechanism to support skipping individual test cases (aside from the pattern used for unified spec tests on load). It would be nice if we can eventually improve our test framework at some point so that it also works better with EVG's test case reporter as well.

mdbmes: What do you think about the change I pushed using TestSuite_AddFullWithTestFn? It's the same method we use to skip tests that are missing a live mongodb server. Results look like:

```
Begin /structured_log/env_defaults, seed 1693407775
2025/01/16 14:29:22.0774: [30660]:    DEBUG:       mongoc: Skipping test because environment var 'MONGODB_LOG_MAX_DOCUMENT_LENGTH' is set
    { "status": "skip", "test_file": "/structured_log/env_defaults" }
  ]
}
```

src/libmongoc/tests/test-mongoc-structured-log.c
```diff
@@ -714,4 +821,11 @@ test_structured_log_install (TestSuite *suite)
    TestSuite_Add (suite, "/structured_log/duration", test_structured_log_duration);
    TestSuite_Add (suite, "/structured_log/level_names", test_structured_log_level_names);
    TestSuite_Add (suite, "/structured_log/component_names", test_structured_log_component_names);
+   TestSuite_Add (suite, "/structured_log/max_document_length", test_structured_log_max_document_length);
+   TestSuite_AddFullWithTestFn (suite,
+                                "/structured_log/env_defaults",
+                                TestSuite_AddHelper,
+                                NULL,
+                                test_structured_log_env_defaults,
+                                check_test_structured_log_env_defaults);
```
eramongodb, line 830:
```suggestion
   TestSuite_AddFull (suite,
                      "/structured_log/env_defaults",
                      test_structured_log_env_defaults,
                      NULL,
                      NULL,
                      check_test_structured_log_env_defaults);
```
Recommend using `TestSuite_AddFull` with `_dtor` and `_ctx` set to `NULL` due to no test context data being required. This avoids the need to expose `TestSuite_AddHelper`.

src/libmongoc/tests/test-mongoc-structured-log.c
```diff
@@ -699,6 +738,74 @@ test_structured_log_component_names (void)
...
+         MONGOC_DEBUG ("Skipping test because environment var '%s' is set", var);
+         return 0;
+      }
+   }
+   return 1;
+}
+
+void
+test_structured_log_env_defaults (void)
+{
```
eramongodb, line 783:
```suggestion
test_structured_log_env_defaults (void *ctx)
{
   BSON_UNUSED (ctx);

```
`AddFull*` test functions must accept `void *ctx` for function type consistency:
```
runtime error: call to function test_structured_log_env_defaults through pointer to incorrect function type 'void (*)(void)'
```

src/libmongoc/tests/test-mongoc-structured-log.c
```diff
@@ -699,6 +738,74 @@ test_structured_log_component_names (void)
...
+   ASSERT (mongoc_structured_log_opts_set_max_document_length (opts, 0));
+   ASSERT (!mongoc_structured_log_opts_set_max_document_length (opts, INT_MAX));
+   ASSERT (mongoc_structured_log_opts_set_max_document_length (opts, INT_MAX / 2));
+   ASSERT_CMPINT (INT_MAX / 2, ==, mongoc_structured_log_opts_get_max_document_length (opts));
+
+   mongoc_structured_log_opts_destroy (opts);
+}
+
+int
+check_test_structured_log_env_defaults (void)
```
eramongodb, line 757:
```suggestion
test_structured_log_skip_if_env_not_default (void)
```
Recommend including "skip" in the name of test suite skip functions for consistency.

src/libmongoc/tests/test-mongoc-structured-log.c
```diff
@@ -714,4 +821,11 @@ test_structured_log_install (TestSuite *suite)
    TestSuite_Add (suite, "/structured_log/duration", test_structured_log_duration);
    TestSuite_Add (suite, "/structured_log/level_names", test_structured_log_level_names);
    TestSuite_Add (suite, "/structured_log/component_names", test_structured_log_component_names);
+   TestSuite_Add (suite, "/structured_log/max_document_length", test_structured_log_max_document_length);
```
eramongodb, line 824:
```suggestion
   TestSuite_AddFull (suite,
                      "/structured_log/max_document_length",
                      test_structured_log_max_document_length,
                      NULL,
                      NULL,
                      check_test_structured_log_env_defaults);
```

Register other structured logging tests which may be impacted by env vars at runtime with the same skip condition used for the `/structured_log/env_defaults` test below.

eramongodb (changes requested):

src/libmongoc/tests/test-mongoc-structured-log.c
eramongodb: I'd completely forgot about the `AddFull*` skip functions. Good call. I am in favor.

src/libmongoc/tests/test-mongoc-structured-log.c
mdbmes: Thanks, didn't notice that this layer was unnecessary. AddFull does require a NULL terminator on the va list too.

src/libmongoc/tests/test-mongoc-structured-log.c
mdbmes: It should be fine as it, I moved the default test for max_document_length into test_structured_log_env_defaults(). test_structured_log_max_document_length() is just the remaining tests which don't depend on env settings.

eramongodb: Given the env var is read on every construction of the `mongoc_structured_log_opts_new` object, all tests which construct the object may be affected, e.g.:
```
$ MONGODB_LOG_MAX_DOCUMENT_LENGTH=1 ./test-libmongoc -d -l "/structured_log/json"
...
Begin /structured_log/json, seed 1057888386
FAIL

Assert Failure:
  "{ "message" : "Log entry with deferred BSON-to-JSON", "kJSON" : "{...", "kNull" : null }"
  !=
  "{ "message" : "Log entry with deferred BSON-to-JSON", "kJSON" : "{ \"k\" : \"v\" }", "kNull" : null }"
 src/libmongoc/tests/test-mongoc-structured-log.c:54  structured_log_func()
```
```
$ MONGODB_LOG_MAX_DOCUMENT_LENGTH=1 ./test-libmongoc -d -l "/structured_log/command"
...
Begin /structured_log/command, seed 897720391
FAIL

Assert Failure:
  "{ "message" : "Log entry with command and reply fields", "commandName" : "Not a command", "databaseName" : "Some database", "commandName" : "Not a command", "operationId" : 1311768468877410321, "command" : "{...", "reply" : "{...", "reply" : "{...", "reply" : "{}", "failure" : { "r" : "s", "code" : 1 }, "failure" : { "r" : "s", "code" : 1 }, "failure" : { "code" : 1 }, "failure" : { "code" : 123, "domain" : 456, "message" : "oh no" } }"
  !=
  "{ "message" : "Log entry with command and reply fields", "commandName" : "Not a command", "databaseName" : "Some database", "commandName" : "Not a command", "operationId" : 1311768468877410321, "command" : "{ \"c\" : \"d\", \"first_payload\" : [ { \"i\" : 0, \"x\" : 0 }, { \"i\" : 0, \"x\" : 1 }, { \"i\" : 0, \"x\" : 2 }, { \"i\" : 0, \"x\" : 3 }, { \"i\" : 0, \"x\" : 4 } ], \"second_payload\" : [ { \"i\" : 1, \"x\" : 0 }, { \"i\" : 1, \"x\" : 1 }, { \"i\" : 1, \"x\" : 2 }, { \"i\" : 1, \"x\" : 3 }, { \"i\" : 1, \"x\" : 4 } ] }", "reply" : "{ \"r\" : \"s\", \"code\" : 1 }", "reply" : "{ \"r\" : \"s\", \"code\" : 1 }", "reply" : "{}", "failure" : { "r" : "s", "code" : 1 }, "failure" : { "r" : "s", "code" : 1 }, "failure" : { "code" : 1 }, "failure" : { "code" : 123, "domain" : 456, "message" : "oh no" } }"
 src/libmongoc/tests/test-mongoc-structured-log.c:54  structured_log_func()
```
```
$ MONGODB_LOG_ALL=off ./test-libmongoc -d -l "/structured_log/opts"
...
Begin /structured_log/opts, seed 3267837868
FAIL

Assert Failure: 4 == 0
src/libmongoc/tests/test-mongoc-structured-log.c:72  test_structured_log_opts()
```

src/libmongoc/tests/test-mongoc-structured-log.c
```diff
@@ -714,4 +823,12 @@ test_structured_log_install (TestSuite *suite)
...
+   TestSuite_AddFull (suite,
+                      "/structured_log/env_defaults",
+                      test_structured_log_env_defaults,
+                      NULL,
+                      NULL,
+                      test_structured_log_skip_if_env_not_default,
+                      NULL);
```
eramongodb, line 833:
```suggestion
                      test_structured_log_skip_if_env_not_default);
```
The `TestSuite_AddFull` macro (not to be confused with the `_TestSuite_AddFull` function it wraps) guarantees a `NULL` terminating argument is passed at the end of the argument list.

mdbmes: My mistake.

src/libmongoc/tests/test-mongoc-structured-log.c
mdbmes: It's not appropriate to skip these tests. These tests should be modified not to depend on the default.

In general, I'm happy to guard against environment settings that conflict with the actual test being performed, but it's not reasonable to exhaustively list environment settings that would cause the test to fail.

src/libmongoc/tests/test-mongoc-structured-log.c
```diff
@@ -142,6 +143,7 @@ test_structured_log_plain (void)
 
    mongoc_structured_log_opts_t *opts = mongoc_structured_log_opts_new ();
    mongoc_structured_log_opts_set_handler (opts, structured_log_func, &assumption);
+   ASSERT (mongoc_structured_log_opts_set_max_document_length (opts, 10000));
```
eramongodb, line 146:
```suggestion
   ASSERT (
      mongoc_structured_log_opts_set_max_document_length (opts, MONGOC_STRUCTURED_LOG_DEFAULT_MAX_DOCUMENT_LENGTH));
```
Should these values of `10000` be `MONGOC_STRUCTURED_LOG_DEFAULT_MAX_DOCUMENT_LENGTH` (1000) instead?

eramongodb (approved):
One last suggestion; otherwise, LGTM. 👍

src/libmongoc/tests/test-mongoc-structured-log.c
mdbmes: The specific value doesn't really matter as long as it's more than the small lengths used in the test, but my intent was to follow the advice from the CLAM spec:

"Drivers MUST run the logging tests with their max document
length setting (as described in the [logging specification](../../logging/logging.md#configurable-max-document-length))
set to a large value e.g. 10,000; this is necessary in order for the driver to emit the full server reply (and to allow
matching against that reply) on certain MongoDB versions and topologies."

It's mostly arbitrary but I'll add a comment explaining the choice just so it doesn't seem out of place.

PR #1827, opened 2025-01-21
copy server description without RTT for server stream

kevinAlbs:
The RTT is not needed. Report suggests the atomic operations may be costly.


kevinAlbs:
Closing for #1828

PR #1831, opened 2025-01-22
CDRIVER-3775 structured log: floating point durations

ghost:
The CLAM and SDAM specs both define `durationMS` values as integers, but in fact the unified tests are flexible about data types and the Python implementation returns a double-precision float value here.

This redefines libmongoc's `monotonic_time_duration` structured log item as double-precision floating point in place of the previous definition that used an integer number of milliseconds accompanied by a second higher-precision integer.

The new definition is consistent with the Python driver implementation and with the unified tests but it diverges from the text of the specification.

Also adds a new `double` structured log item, which underlies the implementation of `monotonic_time_duration` now.


kevinAlbs (approved):
> diverges from the text of the specification

Good spot. Reported DRIVERS-3089.

PR #1836, opened 2025-01-27
CDRIVER-5866 Fix 2 includes

jeroen:
This fixes two complication errors:

```
clang -arch arm64 -DNDEBUG -I. -Icommon -Ikms -Iutf8proc -DBSON_COMPILATION -DBSON_EXTRA_ALIGN -DMONGOC_COMPILATION -DMONGOC_HAVE_SASL_CLIENT_DONE -DMONGOC_ENABLE_SSL_SECURE_TRANSPORT -DMONGOC_ENABLE_CRYPTO_COMMON_CRYPTO -DKMS_MESSAGE_ENABLE_CRYPTO -DKMS_MESSAGE_ENABLE_CRYPTO_COMMON_CRYPTO -D_DARWIN_C_SOURCE  -I/opt/R/arm64/include   -mmacosx-version-min=10.8 -fPIC  -falign-functions=64 -Wall -g -O2  -c mongoc/mongoc-cluster.c -o mongoc/mongoc-cluster.o
In file included from mongoc/mongoc-cluster.c:55:
mongoc/utlist.h:20:10: error: 'mongoc-prelude.h' file not found with <angled> include; use "quotes" instead
   20 | #include <mongoc-prelude.h>
      |          ^~~~~~~~~~~~~~~~~~
      |          "mongoc-prelude.h"
1 error generated.
```

and 

```
clang -arch arm64 -DNDEBUG -I. -Icommon -Ikms -Iutf8proc -DBSON_COMPILATION -DBSON_EXTRA_ALIGN -DMONGOC_COMPILATION -DMONGOC_HAVE_SASL_CLIENT_DONE -DMONGOC_ENABLE_SSL_SECURE_TRANSPORT -DMONGOC_ENABLE_CRYPTO_COMMON_CRYPTO -DKMS_MESSAGE_ENABLE_CRYPTO -DKMS_MESSAGE_ENABLE_CRYPTO_COMMON_CRYPTO -D_DARWIN_C_SOURCE  -I/opt/R/arm64/include   -mmacosx-version-min=10.8 -fPIC  -falign-functions=64 -Wall -g -O2  -c mongoc/mongoc-init.c -o mongoc/mongoc-init.o
mongoc/mongoc-init.c:55:10: error: 'mongoc-cyrus-private.h' file not found with <angled> include; use "quotes" instead
   55 | #include <mongoc-cyrus-private.h> // _mongoc_cyrus_verifyfile_cb
      |          ^~~~~~~~~~~~~~~~~~~~~~~~
      |          "mongoc-cyrus-private.h"
1 error generated.
```

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-init.c
+++ b/src/libmongoc/src/mongoc/mongoc-init.c
@@ -52,7 +52,7 @@
 
 #ifdef MONGOC_ENABLE_SASL_CYRUS
 #include <sasl/sasl.h>
-#include <mongoc-cyrus-private.h> // _mongoc_cyrus_verifyfile_cb
+#include "mongoc-cyrus-private.h" // _mongoc_cyrus_verifyfile_cb
 
 static void *
 mongoc_cyrus_mutex_alloc (void)
--- a/src/libmongoc/src/mongoc/utlist.h
+++ b/src/libmongoc/src/mongoc/utlist.h
@@ -17,7 +17,7 @@
 #ifndef MONGOC_UTLIST_H
 #define MONGOC_UTLIST_H
 
-#include <mongoc-prelude.h>
+#include "mongoc-prelude.h"
 
 #include <uthash-2.3.0/utlist.h>
 


jeroen:
I maintain the R bindings and we vendor mongo-c-driver. We have a custom makefile so this might explain some subtle differences with your cmake config.

What makes the above lines a bug, is that they use bracket notation but the include path is not prefixed with `<mongoc/..>` as is done everywhere else in the code. These are the only two lines that assume the mongoc dir is on the include path, which I think is bad practice.

So you should either switch them to `"quotes"` or alternatively prefix the mongoc includes with `mongoc/`.

kevinAlbs:
Thank you for reporting @jeroen. Filed CDRIVER-5865 to apply a consistent include pattern in the codebase.

> or alternatively use `#include <mongoc/....>`

I would rather this PR be updated to use this form to avoid relative look-up.

jeroen:
@kevinAlbs it's fine by me but literally all other mongoc headers in this file use quotes so for consistency that seems to make more sense? This line is the only odd one out...


https://github.com/mongodb/mongo-c-driver/blob/69d04ae70cca48f2035f6b40ae3347057edd43e4/src/libmongoc/src/mongoc/mongoc-init.c#L18-L55

PR #1837, opened 2025-01-27
Fix a compiler warning on Windows

jeroen:
Fixes a warning with mingw-w64 on Windows:

```
  kms/kms_crypto_windows.c:171:35: warning: pointer targets in passing argument 3 of 'CryptDecodeObjectEx' differ in signedness [-Wpointer-sign]
  kms/kms_crypto_windows.c:185:35: warning: pointer targets in passing argument 3 of 'CryptDecodeObjectEx' differ in signedness [-Wpointer-sign]
```


kevinAlbs:
Thank you for the fix @jeroen. Can you instead make this change to https://github.com/mongodb/libmongocrypt? kms-message is vendored, and the C driver tests [check the vendored code matches](https://github.com/kevinAlbs/mongo-c-driver/blob/master/.evergreen/scripts/kms-divergence-check.sh#L3-L4).

jeroen:
Moved to https://github.com/mongodb/libmongocrypt/pull/944

PR #1839, opened 2025-01-28
CDRIVER-5873 reduce Debian 9 task coverage to compile-only

eramongodb:
Addresses task failures on Evergreen due to the following error thrown by DET's mongodl.py script on Debian 9.2 distros:

```
ValueError: No download was found for version="latest" target="debian92" arch="x86_64" edition="enterprise" component="crypt_shared"
```

Debian 9 is EOL and should probably have been removed from the EVG matrix as part of CDRIVER-4602.

All removed tasks either already have a Debian 10 or Debian 11 equivalent, or were extended accordingly (i.e. Clang coverage), with the sole exception of the `clang38` build variant which ran the `release-compile`, `debug-compile-nosasl-nossl`, and `test-latest-server-*` tasks (there are no equivalent for Clang on Debian 10+, only GCC equivalents).

### DIFF
--- a/.evergreen/config_generator/components/cse/openssl.py
+++ b/.evergreen/config_generator/components/cse/openssl.py
@@ -21,7 +21,6 @@ COMPILE_MATRIX = [
     ('debian10', 'gcc',   None, ['cyrus']),
     ('debian11', 'clang', None, ['cyrus']),
     ('debian11', 'gcc',   None, ['cyrus']),
-
     ('rhel80',            'gcc',       None, ['cyrus']),
     ('rhel8-zseries',     'gcc',       None, ['cyrus']),
     ('ubuntu2004',        'clang',     None, ['cyrus']),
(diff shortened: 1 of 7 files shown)


.evergreen/config_generator/components/c_std_compile.py
```diff
@@ -17,7 +17,6 @@
 # pylint: disable=line-too-long
 # fmt: off
 MATRIX = [
-    ('debian92',          'clang',     None,   [11,   ]),
```
kevinAlbs, line 20: [Process to drop support of an OS+Architecture or Compiler](https://docs.google.com/document/d/1yt8kIw5098WGeUgYajRbCFCEkUsdhM33-P3fwVH_ehE/edit?tab=t.0) suggests documenting this in a minor version prior to dropping. Erring towards caution, I suggest add back compile-only tasks for Debian 9.2 and update NEWS to warn of the upcoming drop.

Aside: given that Debian 9.2 has been EOL for 2.5 years, I expect (hope) dropping has little impact. @bisht2050 do you have (or can point me to) usage data? If Debian 9.2 usage is very low, maybe warning in a minor release is overly cautious.

eramongodb:
Filed CDRIVER-5873 and CDRIVER-5874 tracking removal of Debian 9 and Debian 10 task coverage (both are EOL). Restored compile-only tasks for Debian 9 in config generator components (tasks in the legacy config remain removed). Added a deprecation entry to NEWS files accordingly.

kevinAlbs (approved):

PR #1841, opened 2025-01-28
CDRIVER-5877 avoid aligned_alloc on MacOS prior to 10.15

barracuda156:
@NattyNarwhal Could you please review this? Pre-Catalina macOS need the same fix as you have added for AIX in https://github.com/mongodb/mongo-c-driver/commit/9fe8af14e458523bbe57d9e2658295976f2aac08

### DIFF
--- a/src/libbson/src/bson/bson-memory.c
+++ b/src/libbson/src/bson/bson-memory.c
@@ -23,6 +23,10 @@
 #include <bson/bson-config.h>
 #include <bson/bson-memory.h>
 
+#ifdef __APPLE__
+#include <AvailabilityMacros.h>
+#endif
+
 
 // Ensure size of exported structs are stable.
 BSON_STATIC_ASSERT2 (bson_mem_vtable_t, sizeof (bson_mem_vtable_t) == sizeof (void *) * 8u);
@@ -32,7 +36,7 @@ BSON_STATIC_ASSERT2 (bson_mem_vtable_t, sizeof (bson_mem_vtable_t) == sizeof (vo
 static void *
 _aligned_alloc_impl (size_t alignment, size_t num_bytes)
 #if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L && !defined(_WIN32) && !defined(__ANDROID__) && \
-   !defined(_AIX)
+   !defined(_AIX) && !(defined(__APPLE__) & MAC_OS_X_VERSION_MIN_REQUIRED < 101500)
 {
    return aligned_alloc (alignment, num_bytes);
 }


eramongodb:
Created CDRIVER-5876 to track relocating `aligned_alloc` availability checks from preprocessor directives into the build system.

src/libbson/src/bson/bson-memory.c
```diff
@@ -32,7 +36,7 @@ BSON_STATIC_ASSERT2 (bson_mem_vtable_t, sizeof (bson_mem_vtable_t) == sizeof (vo
 static void *
 _aligned_alloc_impl (size_t alignment, size_t num_bytes)
 #if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L && !defined(_WIN32) && !defined(__ANDROID__) && \
-   !defined(_AIX)
+   !defined(_AIX) && !(defined(__APPLE__) & MAC_OS_X_VERSION_MIN_REQUIRED < 101500)
```
eramongodb, line 39:
```suggestion
   !defined(_AIX) && !(defined(MAC_OS_X_VERSION_MIN_REQUIRED) && MAC_OS_X_VERSION_MIN_REQUIRED < 101500)
```
Use logical OR rather than bitwise OR. Additionally recommend adjusting the condition to test for the definition of `MAC_OS_X_VERSION_MIN_REQUIRED` instead of `defined(__APPLE__)` (already used to condition the inclusion of `<AvailabilityMacros.h>` above) to guard against the possibility of `OSX_DEPLOYMENT_TARGET` and/or `-mmacosx-version-min` interfering with the presence of the macro (not a configuration space we control nor test).

eramongodb: Noting here for future reference that the value of `101500` (MacOS 10.15) is consistent with [libc++'s own checks](https://github.com/llvm/llvm-project/blob/db60244519023a2b083caa3ed3a27a6b59eb03d8/libcxx/include/__config#L662) for `aligned_alloc` support:
```cpp
// It is not yet possible to use aligned_alloc() on all Apple platforms since
// 10.15 was the first version to ship an implementation of aligned_alloc().
#  if defined(__APPLE__)
#    if (defined(__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__) &&                                                     \
         __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 101500) ||                                                    \
        /* ... */
#      define _LIBCPP_HAS_C11_ALIGNED_ALLOC 0
#    else
#      define _LIBCPP_HAS_C11_ALIGNED_ALLOC 1
#    endif
#  elif /* ... */
```

eramongodb (changes requested):
The difficulty of finding official documentation for `<AvailabilityMacros.h>` and its provided features makes me hesitant to depend on it. This may nevertheless suffice as a temporary solution until CDRIVER-5876 is resolved.

eramongodb:
Filed CDRIVER-5877 to track this PR.

kevinAlbs:
Thank you for alerting us of this issue @barracuda156. CDRIVER-5876 applies a fix within CMake to detect if `aligned_alloc` is available. CDRIVER-5876 was released in 1.30.0.

barracuda156:
@kevinAlbs Thank you very much, great! (Sorry for a delayed reply, a lot of stuff in the pipeline.)

kevinAlbs:
> @kevinAlbs Thank you very much, great! (Sorry for a delayed reply, a lot of stuff in the pipeline.)

Hah no worries. And thanks to @eramongodb for the change. I only reviewed it :)

PR #1842, opened 2025-01-29
CDRIVER-4500: SDAM structured log and unified test support

ghost:
* Resolves CDRIVER-4500 (SDAM structured logs + updated unified tests)
* Resolves CDRIVER-4758 (Prose test for heartbeat event order: /server_discovery_and_monitoring/prose/heartbeat)
* Resolves CDRIVER-4137 (Missing server close events, now tested by /server_discovery_and_monitoring/unified)
* Completes outstanding issues in the CDRIVER-3775 epic (libmongoc structured logging)

The central purpose of the patch is to add missing log events for heartbeats, topology, and server updates. Several other changes became necessary:

* Refactored APM callbacks into a new "log and monitor instance"

   * Previously, APM callbacks were duplicated in many places including each individual topology description instance. This moves both APM and Structured Log callbacks/settings to a new "log and monitor instance" (mongoc_log_and_monitor_instance_t) object with a clear internal interface. It's owned by mongoc_topology_t, and given as an explicit parameter to log and monitor functions.

   * Existing documented thread safety rules about pool settings (At most one write, prior to the first client) are implemented uniformly for both structured logs and APM callbacks. Edit: For compatibility, previously allowed unsafe uses are now still allowed. A new warning message will be reported as an unstructured log error.

* Refined topology lifecycle events to match standard behavior

   * Server 'opened' events are deferred until after the corresponding topology is 'opened', rather than until the APM callbacks have been set.

   * Added missing end-of-life SDAM events. Before monitoring is closed, topology descriptions will now transition to an "Unknown" state as required. Server monitoring will be closed. Closure events are only emitted if the same log_and_monitor instance has seen a corresponding open.

* Added test skips for failures corresponding to features that are currently out of scope:

   * CMAP events, especially pool clearing on various error conditions
   * Driver-generated connection IDs (optional in the spec, but some tests assume they are present)
   * Thread entities in unified tests

* Added new unified test runner features:

   * Support for "closing" some entity types without deleting (For example, to inspect events logged when closing a client)
   * SDAM APM events, and a new eventType filtering implementation
   * New operations: recordTopologyDescription, assertTopologyType, waitForPrimaryChange
   * Added a structured log filter stack (entity_map_log_filter_push/pop) used internally to refine the apparent behavior of waitForEvent

* Added new structured log items:

   * oid() for a plain ObjectID without hex representation
   * topology_description_as_json() for plain topology descriptions not inside a topology

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-topology-description.c
+++ b/src/libmongoc/src/mongoc/mongoc-topology-description.c
@@ -1395,8 +1395,12 @@ mongoc_topology_description_add_server (mongoc_topology_description_t *topology,
 
       mongoc_set_add (mc_tpld_servers (topology), server_id, description);
 
-      /* if we're in topology_new then no callbacks are registered and this is
-       * a no-op. later, if we discover a new RS member this sends an event. */
+      /* Note that libmongoc defers topology 'opening' until server selection or background monitoring begins,
+       * and server monitoring must correspondingly only be 'opened' after the API has seen topology monitoring open.
+       *
+       * If the topology is already opened, we will send server opening events immediately.
+       * Otherwise this has no effect, and server opening events will be sent later by
+       * _mongoc_topology_description_monitor_opening. */
       _mongoc_topology_description_monitor_server_opening (topology, log_and_monitor, description);
    }
 


src/libmongoc/src/mongoc/mongoc-topology-description-apm.c
```diff
@@ -148,38 +174,52 @@ _mongoc_topology_description_monitor_opening (mongoc_topology_description_t *td)
...
 /* TopologyClosedEvent */
 void
-_mongoc_topology_description_monitor_closed (const mongoc_topology_description_t *td)
+_mongoc_topology_description_monitor_closed (const mongoc_topology_description_t *td,
+                                             const mongoc_log_and_monitor_instance_t *log_and_monitor)
 {
-   if (td->apm_callbacks.topology_closed) {
-      mongoc_apm_topology_closed_t event;
+   BSON_ASSERT (td->type == MONGOC_TOPOLOGY_UNKNOWN);
+   BSON_ASSERT (mc_tpld_servers_const (td)->items_len == 0);
```
eramongodb, line 207: Suggest adding a comment referencing the function/code responsible for ensuring these conditions; these assertions feel somewhat out-of-place otherwise (beyond the responsibility of merely logging the event on request).

src/libmongoc/src/mongoc/mongoc-structured-log.c
```diff
@@ -1022,10 +1040,9 @@ _mongoc_structured_log_append_server_description (bson_t *bson,
    return stage + 1;
 }
 
-const mongoc_structured_log_builder_stage_t *
-_mongoc_structured_log_append_topology_as_description_json (bson_t *bson,
-                                                            const mongoc_structured_log_builder_stage_t *stage,
-                                                            const mongoc_structured_log_opts_t *opts)
+static mcommon_string_t *
+_mongoc_structured_log_topology_description_as_json (const mongoc_topology_description_t *td,
```
eramongodb, line 1044:
```suggestion
_mongoc_structured_log_topology_description_to_json (const mongoc_topology_description_t *td,
```
Perhaps "description_to_json" may help disambiguate these helper functions a bit better from their "as_description_json" counterparts.

eramongodb (changes requested):
Although the frequency of the `td` + `log_and_monitor` parameter pairs is unfortunate, the consolidation of both logging and event handlers into `mongoc_topology_t::log_and_monitor` is a welcome improvement.

eramongodb (approved):
Thank you for the explanations and clarifications. Re-suggesting some changes to coincide with additional motivations outlined in responses below, but otherwise, LGTM.

> Re the extra warnings you're reporting, is there any plan to include them in an evergreen check?

No, these are warning I enable for local development (Clang's `-Weverything` + set of `-Wno-*` to disable irrelevant/undesirable warnings) to assist with drive-by code improvements and avoid expanding the set of possible warnings which might be observed by users (comparing a diff of total warnings for proposed changes against the base commit).

> casting away const in an explicit conversion to (void*) seems to be a common enough pattern in the repository.

It is an unfortunately common pattern. The reason for the `-Wcast-qual` warning (which is emitted in spite of the explicit cast to `void*`) is to guard against the possibility of undefined behavior due to accessing a const-qualified object when the non-const void pointer is (expected to be) cast back to an object pointer. Recent code changes has been making efforts to improve the const-correctness of our interfaces even when type-erasure is involved (`*_const` functions).

>  Why do we need to disallow stateful filter functions?

The suggestion was based simply on the observation that no current usage of this pointer requires modifiable access to the pointed-to object: it is only being used for pointer equality comparison [here](https://github.com/mongodb/mongo-c-driver/blob/34b647b0dd61b18c96671623e93a735f994f2fc5/src/libmongoc/tests/unified/entity-map.c#L246) and string comparison [here](https://github.com/mongodb/mongo-c-driver/blob/34b647b0dd61b18c96671623e93a735f994f2fc5/src/libmongoc/tests/unified/operation.c#L3858). If your preference is to leave open the possibility for modifiable object pointers by using `void*`, that is fine.

> My preference was to avoid the "unsafe" version unless there was a reason to prefer it.

Suggestion was not for performance reasons, but instead to communciate that synchronization is not required in this context (there should only be one thread accessing the to-be-destroyed topology object). That is, TSAN should rightfully warn if there are multiple threads accessing (un)guarded resources at this point.

> I declined to document apm_mutex in additional detail here because I'm not convinced the current implementation is correct. [...] For this patch, I tried to preserve the existing behavior without changing it.

The proposed rationale was echoing the one given at the time the mutex was introduced (see: https://github.com/mongodb/mongo-c-driver/pull/607#discussion_r421725853).

My guess is that, at the time, thread-safe usage of the C Driver API may not have been an actively supported use case, therefore APM callback invocations were assumed to be always single-threaded (from the perspective of a user who registered callbacks) until the introduction of background server monitoring threads. This would explain why no other APM callback invocations are guarded.

> This is for skipping events of other event types, when the "eventType" parameter is being used. When there's no "eventType", this never skips events.

I overlooked the specified behavior of `eventType` event checks. Thank you.

The skipping of cmap events would still be required.

> It's been documented in the manpage.

Sorry, I was referencing outdated docs (generated locally). I see the Thread Safety section [on the live website](https://mongoc.org/libmongoc/current/mongoc_client_pool_set_apm_callbacks.html).

> It's a workaround for allowing applications to see events which would otherwise be emitted earlier than the APM callbacks could be attached.

Thank you for the explanation.

> It's not clear to me that an ASSERT here would help. [...] I don't think it would be an improvement to guarantee a crash after opening 2^64 topologies per process.

The proposed assertion is not to prevent collisions, but to avoid undefined behavior due to signed integer overflow during the atomic increment. The assertion would guard against undefined behavior by detecting the soon-to-occur condition and triggering well-defined erroneous behavior instead.

> This implies that we could switch to using type-specific event lists, instead of having one event list that we filter.

I am fine with deferring this potential improvement to a separate PR.

src/libmongoc/src/mongoc/mongoc-topology.c
```diff
@@ -733,23 +678,40 @@ mongoc_topology_destroy (mongoc_topology_t *topology)
...
-      bson_mutex_destroy (&topology->apm_mutex);
       bson_mutex_destroy (&topology->srv_polling_mtx);
       mongoc_cond_destroy (&topology->srv_polling_cond);
    }
 
-   if (topology->valid) {
-      /* Do not emit a topology_closed event. A topology opening event was not
-       * emitted. */
-      _mongoc_topology_description_monitor_closed (mc_tpld_unsafe_get_const (topology));
+   /* Before reporting this topology as closed, life cycle rules expect us to close
```
kevinAlbs, line 685: > Added missing end-of-life SDAM events.

Suggest noting this change in the NEWS since this impacts how callbacks are called.

I set `PHP Team Attention Needed` in CDRIVER-4500 to generate a ticket to inform the PHP team. (This required transitioning the ticket to `Ready for Work`, then back to `Code Review` to work around [PROG-12016](https://jira.mongodb.org/browse/PROG-12016).

src/libmongoc/src/mongoc/mongoc-topology-description-private.h
```diff
@@ -53,16 +53,16 @@ struct _mongoc_topology_description_t {
    bool stale;
    unsigned int rand_seed;
 
+   /* Has this been 'opened' by a log_and_monitor instance? If yes, this is that instance's version_id. Zero if not. */
+   bson_oid_t opened_by_log_and_monitor_version_id;
```
kevinAlbs, line 57: > Early SDAM events may be missed due to the relatively late stage at which callbacks can be attached to a non-pooled client.

Is this needed only for tests? I expect this is a change in events returned:

```c
mongoc_client_t *client = mongoc_client_new ("mongodb://localhost:27017");
ASSERT (mongoc_client_command_simple (client, "db", tmp_bson ("{'hello': 1}"), NULL, NULL, NULL));
set_callbacks (client); // Set callbacks after first command.
ASSERT (mongoc_client_command_simple (client, "db", tmp_bson ("{'ping': 1}"), NULL, NULL, NULL));
```

Before this PR, the topology opening callback is not called. If this change is only motivated by tests, can the test runner be modified to set callbacks earlier? That might avoid an unexpected subtle behavior change.

kevinAlbs: Resolving. Latest changes switch back to an `opened` bool flag.

kevinAlbs (approved):

PR #1843, opened 2025-01-29
CDRIVER-5869 fix bulk write crash with verbose results

kevinAlbs:
# Summary

Uses integral offsets, rather than a `bson_iter_t`, to refer to "_id" fields in the `bulkWrite` payload.

Verified with this patch build: https://spruce.mongodb.com/version/6798e244aed6b800073ef361

# Background

See CDRIVER-5869 for background.

### DIFF
--- a/src/libmongoc/tests/test-mongoc-bulkwrite.c
+++ b/src/libmongoc/tests/test-mongoc-bulkwrite.c
@@ -629,6 +629,58 @@ test_bulkwrite_execute_requires_client (void *ctx)
    mongoc_client_destroy (client);
 }
 
+// `test_bulkwrite_two_large_inserts` is a regression test for CDRIVER-5869.
+static void
+test_bulkwrite_two_large_inserts (void *unused)
+{
+   BSON_UNUSED (unused);
+
+   bson_error_t error;
+   mongoc_client_t *client = test_framework_new_default_client ();
+
+   // Drop prior collection:
+   {
+      mongoc_collection_t *coll = mongoc_client_get_collection (client, "db", "coll");
+      mongoc_collection_drop (coll, NULL);
+      mongoc_collection_destroy (coll);
+   }
+
+   // Allocate a large string:
+   size_t large_len = 2095652;
+   char *large_string = bson_malloc (large_len + 1);
+   memset (large_string, 'a', large_len);
+   large_string[large_len] = '\0';
+
+   // Create two large documents:
+   bson_t *docs[2];
+   docs[0] = BCON_NEW ("_id", "over_2mib_1");
+   bson_append_utf8 (docs[0], "unencrypted", -1, large_string, large_len);
+   docs[1] = BCON_NEW ("_id", "over_2mib_2");
+   bson_append_utf8 (docs[1], "unencrypted", -1, large_string, large_len);
+
+   mongoc_bulkwriteopts_t *bw_opts = mongoc_bulkwriteopts_new ();
+   mongoc_bulkwriteopts_set_verboseresults (bw_opts, true);
+
+   mongoc_bulkwrite_t *bw = mongoc_client_bulkwrite_new (client);
+   ASSERT_OR_PRINT (mongoc_bulkwrite_append_insertone (bw, "db.coll", docs[0], NULL, &error), error);
+   ASSERT_OR_PRINT (mongoc_bulkwrite_append_insertone (bw, "db.coll", docs[1], NULL, &error), error);
+
+   mongoc_bulkwritereturn_t bwr = mongoc_bulkwrite_execute (bw, bw_opts); // Crashes!
+   ASSERT_NO_BULKWRITEEXCEPTION (bwr);
+   ASSERT (bwr.res);
+   const bson_t *insertresults = mongoc_bulkwriteresult_insertresults (bwr.res);
+   ASSERT_MATCH (insertresults,
+                 BSON_STR ({"0" : {"insertedId" : "over_2mib_1"}}, {"1" : {"insertedId" : "over_2mib_2"}}));
+   bson_destroy (docs[0]);
+   bson_destroy (docs[1]);
+   mongoc_bulkwrite_destroy (bw);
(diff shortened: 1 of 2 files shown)


eramongodb (changes requested):
AFAICT the purpose of the `id_loc` (formerly `id_iter`) data member is ultimately to populate the `insertedId` field in the bulk write insert result document by [_bulkwriteresult_set_insertresult](https://github.com/mongodb/mongo-c-driver/blob/4b4aade5a07c62c917b1f3cd24dec7c34152608f/src/libmongoc/src/mongoc/mongoc-bulkwrite.c#L1013) in [_bulkwritereturn_apply_result](https://github.com/mongodb/mongo-c-driver/blob/4b4aade5a07c62c917b1f3cd24dec7c34152608f/src/libmongoc/src/mongoc/mongoc-bulkwrite.c#L1461), which must correlate to the index of a bulk write insert operation.

According to [db.collection.bulkWrite Examples (MongoDB Manual)](https://www.mongodb.com/docs/manual/reference/method/db.collection.bulkWrite/#examples) there should be an `insertedIds` field in the reply document containing the list of index-id pairs (regardless whether the bulk write operation is ordered or unordered). Can this field be parsed instead, [in a manner similar to](https://github.com/mongodb/mongo-c-driver/blob/4b4aade5a07c62c917b1f3cd24dec7c34152608f/src/libmongoc/src/mongoc/mongoc-bulkwrite.c#L1430) what is currently being done with the `upserted` field to populate the `upsertedId` field for an bulk write update operation result? This would avoid the need to retain an iterator to the `_id` field of the insert document and its resulting implicit reference semantics.

If the `insertedIds` field cannot be used (I could not find discussion/rationale in https://github.com/mongodb/mongo-c-driver/pull/1590), can we avoid implicit reference semantics entirely by instead storing an owning copy (via `bson_strdup`) of the BSON UTF-8 string (as already done for the `.ns` data member)?

kevinAlbs:
> According to [db.collection.bulkWrite Examples (MongoDB Manual)](https://www.mongodb.com/docs/manual/reference/method/db.collection.bulkWrite/#examples) there should be an `insertedIds` field in the reply document

That example uses the `mongosh` helper. The server reply does not return the `_id` fields. The driver is expected to generate the `_id` field for inserted documents. Using mongosh:
```js
db.coll.drop()
// Do an insert:
const reply = db.adminCommand({bulkWrite: "admin", ops: [{insert: 0, document: {_id: "foo" } }], nsInfo: [{ns: "test.coll"}] })
print (reply.cursor.firstBatch[0]) // { ok: 1, idx: 0, n: 1 }
```

For upserts, the server is expected to generate the `_id` and return it:
```js
db.coll.drop()
// Do an upsert:
const reply = db.adminCommand({bulkWrite: "admin", errorsOnly: false, ops: [{update: 0, filter: { _id: "foo" }, updateMods: { $set: { x: 1 }}, upsert: true }], nsInfo: [{ns: "test.coll"}] })
print (reply.cursor.firstBatch[0]) // { ok: 1, idx: 0, n: 1, nModified: 0, upserted: { _id: 'foo' } }
```

> can we avoid implicit reference semantics entirely by instead storing an owning copy

Possibly yes, but it may come at a non-neglible performance cost. Reporting inserted `_id`s is only needed when verbose results are requested. `mongoc_bulkwrite_append_insertone` does not know if the `_id`s will be needed:

```c
mongoc_bulkwrite_t *bw = mongoc_client_bulkwrite_new (client);
ASSERT_OR_PRINT (mongoc_bulkwrite_append_insertone (bw, "db.c", tmp_bson ("{'_id': 'foo'}"), NULL, &error), error);
mongoc_bulkwriteopts_t *bw_opts = mongoc_bulkwriteopts_new ();
mongoc_bulkwriteopts_set_verboseresults (bw_opts, true); // Report insertedId's in result.
mongoc_bulkwritereturn_t bwr = mongoc_bulkwrite_execute (bw, bw_opts);
ASSERT (bwr.res);
const bson_t *insertresults = mongoc_bulkwriteresult_insertresults (bwr.res);
ASSERT_MATCH (insertresults, BSON_STR ({"0" : {"insertedId" : "foo"}}));
```

`_id` can be an arbitrarily BSON value. I expect requesting verbose results is an uncommon use case. I would rather avoid the cost of copying the `_id` of each inserted document to only be used for verbose results.

I considered a simpler alternative to avoid tracking the `_id` offset and iterating later: https://github.com/kevinAlbs/mongo-c-driver/commit/126e45c947e0ad609a315a0beb2d46842f54dc05. But this does not avoid the implicit reference so I did not pursue it.

src/libmongoc/tests/test-mongoc-bulkwrite.c
```diff
@@ -629,6 +629,58 @@ test_bulkwrite_execute_requires_client (void *ctx)
...
+   char *large_string = bson_malloc (large_len + 1);
+   memset (large_string, 'a', large_len);
+   large_string[large_len] = '\0';
+
+   // Create two large documents:
+   bson_t *docs[2];
+   docs[0] = BCON_NEW ("_id", "over_2mib_1");
+   bson_append_utf8 (docs[0], "unencrypted", -1, large_string, large_len);
+   docs[1] = BCON_NEW ("_id", "over_2mib_2");
+   bson_append_utf8 (docs[1], "unencrypted", -1, large_string, large_len);
```
eramongodb, line 659:
```suggestion
   large_string[large_len] = '\0';
   ASSERT (mcommon_in_range_unsigned (int, large_len));

   // Create two large documents:
   bson_t *docs[2];
   docs[0] = BCON_NEW ("_id", "over_2mib_1");
   bson_append_utf8 (docs[0], "unencrypted", -1, large_string, (int) large_len);
   docs[1] = BCON_NEW ("_id", "over_2mib_2");
   bson_append_utf8 (docs[1], "unencrypted", -1, large_string, (int) large_len);
```
Address `-Wshorten-64-to-32` warnings.

eramongodb (approved):

PR #1850, opened 2025-02-04
Suppress -Wdeprecated-declarations warnings for Secure Transport (CDRIVER-5867, CDRIVER-5879)

eramongodb:
Resolves CDRIVER-5867 and CDRIVER-5879.

> [!NOTE]
> Many EVG tasks are currently failing due to unrelated reasons.

CDRIVER-5879 was filed to document the removal of a _repo-level_ `-Wno-deprecated-declarations` flag set by CDRIVER-520 (which seems to incorrectly use `add_definitions` instead of `add_compile_options`, but is nevertheless directly forwarded to command-line arguments as-is by CMake). This is removed in favor of more selective warning suppression methods.

Addressing CDRIVER-5879 revealed the warnings reported in CDRIVER-5867. This PR proposes suppressing these warnings with `(BEGIN|END)_IGNORE_DEPRECATIONS` until CDRIVER-2722 is resolved to avoid blocking builds using maintainer flags. The suppressions are conditioned on `defined(MONGOC_ENABLE_SSL_SECURE_TRANSPORT)` to avoid being applied with other configurations.


kevinAlbs (approved):

jmikola (approved):

src/libmongoc/src/mongoc/mongoc-cyrus.c
```diff
@@ -31,6 +31,11 @@
 #undef MONGOC_LOG_DOMAIN
 #define MONGOC_LOG_DOMAIN "CYRUS-SASL"
 
+// CDRIVER-2722: the entire Secure Transport library is deprecated.
+#if defined(MONGOC_ENABLE_SSL_SECURE_TRANSPORT)
```
jmikola, line 35: Missed this earlier, but the Cyrus warnings were unrelated to Secure Transport.

eramongodb: Added proper distinction between Cyrus SASL (cross-platform, but specifically for MacOS) and the Secure Transport library (MacOS only). Verified by [this patch](https://spruce.mongodb.com/version/67a6674c7e6ae50007bd47b2/).

jmikola (approved):

PR #1851, opened 2025-02-04
CDRIVER-5876 detect aligned_alloc via CMake instead of feature test macros

eramongodb:
Resolves CDRIVER-5876. Alternative (preferable?) solution to https://github.com/mongodb/mongo-c-driver/pull/1841.

When https://github.com/mongodb/mongo-c-driver/pull/1072 was implemented, `check_symbol_exists` was deliberately avoided due to difficulty with accurately setting the required `CMAKE_REQUIRED_*` arguments, which are handled independently from the actual definitions and options used with the libraries being built (i.e. the various `*_SOURCE` macros used to control POSIX, FreeBSD, and Darwin features). However, it seems there are enough cases of platform-specific edge cases concerning the actual availability of `aligned_alloc` despite efforts to use approprate feature test macros that a build system check is sufficiently well-motivated at this stage.

Therefore, this PR proposes adding a new `BSON_HAVE_ALIGNED_ALLOC` option to `bson-config.h`, which is detected using `check_symbol_exists` alongside other existing checks. This PR also applies a drive-by scan and update to `CMAKE_REQUIRED_*` variables to ensure better consistency of definitions/options being used for `check_*_exists` and the built libraries.

---

As a reminder, `check_symbol_exists` [behaves as follows](https://cmake.org/cmake/help/v3.15/module/CheckSymbolExists.html):

> If the header files define the symbol as a macro it is considered available and assumed to work. If the header files declare the symbol as a function or variable then the symbol must also be available for linking (so intrinsics may not be detected).

and:

> The following variables may be set before calling this macro to modify the way the check is run: `CMAKE_REQUIRED_*`

Although `add_definitions`, `add_compile_options`, and other directory-level and target-specific properties are _not_ observed by `check_symbol_exists`, configuration options such as `CMAKE_C_STANDARD` and `CMAKE_C_FLAGS` _are_ observed by `check_symbol_exists`. It is not entirely clear where/how the line is drawn. For our purposes, maintaining `CMAKE_REQUIRED_DEFINITIONS` and `CMAKE_REQUIRED_LIBRARIES` is _probably_ sufficient to ensure consistency between symbol checks and built libraries.


src/libbson/src/bson/bson-memory.c
```diff
@@ -31,8 +31,7 @@ BSON_STATIC_ASSERT2 (bson_mem_vtable_t, sizeof (bson_mem_vtable_t) == sizeof (vo
 // For compatibility with C standards prior to C11.
 static void *
 _aligned_alloc_impl (size_t alignment, size_t num_bytes)
-#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L && !defined(_WIN32) && !defined(__ANDROID__) && \
-   !defined(_AIX)
+#if defined(BSON_HAVE_ALIGNED_ALLOC)
 {
    return aligned_alloc (alignment, num_bytes);
```
kevinAlbs, line 36: Some task failures with clang look related. [Example](https://spruce.mongodb.com/task/mongo_c_driver_u22_clang_check:sasl=Cyrus%C2%A0%E2%80%A2%C2%A0tls=OpenSSL%C2%A0%E2%80%A2%C2%A0test_mongocxx_ref=r3.8.0_patch_0f7fcad4e513d83f3b4241461c3d30504a82dc59_67a2445a59ea7f0007bc0413_25_02_04_16_46_32/logs?execution=0):

```
/usr/bin/clang -DBSON_COMPILATION -DBSON_STATIC -DJSONSL_PARSE_NAN -DMCOMMON_NAME_PREFIX=_bson_mcommon -D_BSD_SOURCE -D_DEFAULT_SOURCE -D_XOPEN_SOURCE=700 -DCMAKE_INTDIR=\"RelWithDebInfo\" -I/opt/mongoc/source/src/libbson/src -I/opt/mongoc/build/src/libbson/src -I/opt/mongoc/source/src/common/src -I/opt/mongoc/build/src/common/src -O2 -g -DNDEBUG -std=c99 -fPIC -fvisibility=hidden -Werror=implicit -Werror=return-type -Werror=incompatible-pointer-types -Werror=int-conversion -Werror=ignored-qualifiers -Werror=uninitialized -Werror=format -Werror=format=2 -Werror=format-non-iso -Werror=format-pedantic -Werror=format-type-confusion -Werror -pedantic -Wall -Wempty-body -Winit-self -Wmissing-include-dirs -Wredundant-decls -Wshadow -Wstrict-prototypes -Wswitch-default -Wswitch-enum -Wundef -Wuninitialized -Wno-strict-aliasing --coverage -MD -MT src/libbson/CMakeFiles/bson_static.dir/RelWithDebInfo/src/bson/bson-memory.c.o -MF src/libbson/CMakeFiles/bson_static.dir/RelWithDebInfo/src/bson/bson-memory.c.o.d -o src/libbson/CMakeFiles/bson_static.dir/RelWithDebInfo/src/bson/bson-memory.c.o -c /opt/mongoc/source/src/libbson/src/bson/bson-memory.c
/opt/mongoc/source/src/libbson/src/bson/bson-memory.c:36:11: error: implicitly declaring library function 'aligned_alloc' with type 'void *(unsigned long, unsigned long)' [-Werror,-Wimplicit-function-declaration]
   return aligned_alloc (alignment, num_bytes);
          ^
/opt/mongoc/source/src/libbson/src/bson/bson-memory.c:36:11: note: include the header <stdlib.h> or explicitly provide a declaration for 'aligned_alloc'
```

kevinAlbs (commented):
I like the change to detect the symbol in CMake, but this appears to cause task failures with clang.

src/libbson/src/bson/bson-memory.c
eramongodb: Because `-Wimplicit-function-declaration` is not an error by default, and because CMake's `check_symbol_exists` considers the symbol (when not a macro) to be "found" when the (undeclared) symbol produces no linker error, `BSON_HAVE_ALIGNED_ALLOC` may therefore be defined even when `<stdlib.h>` does not declare `aligned_alloc` (the symbol really does exist in one of the linked libraries, but it is not declared), which conflicts with our `-Werror=implicit-function-declaration` maintainer flags when compiling our own libraries (the warning is ignored by `check_symbol_exists` even when emitted).

Opted to silence the warning for `aligned_alloc` to remain consistent with CMake rather than attempting to apply maintainer flags to `check_symbol_exists` which may pessimize the cases where the requested symbol is found and used. LMK if we would prefer to make `check_symbol_exists` behavior consistent with our libraries instead (applying `-Werror=implicit-function-declaration` to the check rather than silencing it in our library).

kevinAlbs (approved):

PR #1853, opened 2025-02-06
ccache: support MSVC by using /Z7 via MSVC_DEBUG_INFORMATION_FORMAT

eramongodb:
Attempt to improve support for building with ccache on Windows distros.

CMake uses `/Zi` in `CMAKE_<LANG>_FLAGS_DEBUG` by default. Unfortunately, ccache [does not support](/ccache/ccache/issues/1040) `/Zi`. This means building C projects with CMake and Visual Studio is incompatible with ccache by default when using debug configurations.

CMake 3.25 introduces [CMP0141](https://cmake.org/cmake/help/v3.25/policy/CMP0141.html) which moves the `/Zi` flag out of `CMAKE_<LANG>_FLAGS_DEBUG` into a new, dedicated [`MSVC_DEBUG_INFORMATION_FORMAT`](https://cmake.org/cmake/help/v3.25/prop_tgt/MSVC_DEBUG_INFORMATION_FORMAT.html) target property instead. This property is initialized by the new [`CMAKE_MSVC_DEBUG_INFORMATION_FORMAT`](https://cmake.org/cmake/help/v3.25/variable/CMAKE_MSVC_DEBUG_INFORMATION_FORMAT.html) variable, which still defaults to `/Zi` (aka `ProgramDatabase`) when the compiler supports it (which, for MSVC, it usually does). However, other debug information formats such as `/Z7` (aka `Embedded`) which _are_ compatible with ccache are now available to be specified directly in a controlled manner (rather than messing with `*_FLAGS` variables).

Therefore, this PR adds `find_ccache_and_export_vars` (from https://github.com/mongodb/mongo-c-driver/issues/1524) to `compile-windows.sh`, sets `CMP0141` to `NEW`, and `MSVC_DEBUG_INFORMATION_FORMAT` top `Embedded` (`/Z7`) (when debug info is requested by the build configuration) to fully opt-into ccache-compatible builds.

### DIFF
--- a/.evergreen/scripts/compile-std.sh
+++ b/.evergreen/scripts/compile-std.sh
@@ -134,7 +134,7 @@ echo "Installing libmongocrypt... done."
 # Use ccache if able.
 . "${script_dir:?}/find-ccache.sh"
 find_ccache_and_export_vars "$(pwd)" || true
-if [[ "${OSTYPE:?}" == cygwin ]]; then
+if command -v "${CMAKE_C_COMPILER_LAUNCHER:-}" && [[ "${OSTYPE:?}" == cygwin ]]; then
   configure_flags_append "-DCMAKE_POLICY_DEFAULT_CMP0141=NEW"
   configure_flags_append "-DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=$<$<CONFIG:Debug,RelWithDebInfo>:Embedded>"
 fi
--- a/.evergreen/scripts/compile-std.sh
+++ b/.evergreen/scripts/compile-std.sh
@@ -131,12 +131,16 @@ echo "Installing libmongocrypt..."
 }
 echo "Installing libmongocrypt... done."
 
-echo "CFLAGS: ${CFLAGS}"
-echo "configure_flags: ${configure_flags[*]}"
-
 # Use ccache if able.
 . "${script_dir:?}/find-ccache.sh"
 find_ccache_and_export_vars "$(pwd)" || true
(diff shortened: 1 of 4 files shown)


.evergreen/scripts/link-sample-program.sh
```diff
@@ -76,6 +76,9 @@ ZSTD="AUTO"
 if [[ -f $DIR/find-ccache.sh ]]; then
   . $DIR/find-ccache.sh
   find_ccache_and_export_vars "$SCRATCH_DIR" || true
+if [[ "${OSTYPE:?}" == cygwin ]]; then
```
kevinAlbs, line 79:
```suggestion
fi
if [[ "${OSTYPE:?}" == cygwin ]]; then
```

.evergreen/scripts/compile-std.sh
```diff
@@ -137,6 +137,10 @@ echo "configure_flags: ${configure_flags[*]}"
 # Use ccache if able.
 . "${script_dir:?}/find-ccache.sh"
 find_ccache_and_export_vars "$(pwd)" || true
+if [[ "${OSTYPE:?}" == cygwin ]]; then
```
kevinAlbs, line 140: Suggest invoking `find_ccache_and_export_vars` (I expect ccache is not applied in this file).

eramongodb: Call to `find_ccache_and_export_vars` is done above on line 136.

eramongodb: Done.

kevinAlbs (approved):

PR #1856, opened 2025-02-07
[CDRIVER-5859] Enable sign-compare warnings globally, and fix them all

vector-of-bool:
Refer: CDRIVER-5859 This changeset adds sign-compare warnings at the top-level, and fixes every occurrence. Most occurrences just swap for the new `mlib_cmp` macro, but a few more changes have been made to clear up sign-handling. 

# Summary

- Lots of use of `mlib_cmp`, from PR #1845
- c763a73f0fcd2cc8082b09ddd1a85b5585d54803: A statement-macro that tests whether a given (compound) statement raises `SIGABRT`. This is a no-op on Windows. This uses `fork` without `exec`, so it isn't super fast, usually incurring a few milliseconds for each use.
- e830cf55ea4d7cd75eb898d54496151f4b3b2aad: A macro for looping over a half-open range of integers.
- 93efe4ddeadba205fa3c8594a9ccdcd83ecdd460: A macro for looping over arrays.
- d01e7f678013e274958902204d6aff4e1f7ada10: A new header: `mlib/intencode.h`, which adds functions dedicated to decoding/encoding integer values into raw memory.
- b689d494669baee5844aab5f15775dcd8804aaeb: This commit actually fixes all instances of sign-comparison warnings in the codebase.


src/libbson/src/bson/bson-decimal128.c
```diff
@@ -171,6 +166,7 @@ bson_decimal128_to_string (const bson_decimal128_t *dec, /* IN  */
    /* Decode combination field and exponent */
    combination = (high >> 26) & COMBINATION_MASK;
 
+   int biased_exponent;
    if (BSON_UNLIKELY ((combination >> 3) == 3)) {
       /* Check for 'special' values */
       if (combination == COMBINATION_INFINITY) { /* Infinity */
```
eramongodb, line 172:
```suggestion
      if (mlib_cmp (combination, ==, COMBINATION_INFINITY)) { /* Infinity */
```
With GCC 13.3.0:
```
warning: comparison of integer expressions of different signedness: ‘uint32_t’ {aka ‘unsigned int’} and ‘int’ [-Wsign-compare]
```

Similarly for `combination == COMBINATION_NAN` below on line 175.

src/libbson/tests/test-bson.c
```diff
@@ -1797,8 +1795,7 @@ test_bson_steal (void)
    /* test stealing from a bson created with bson_new_from_buffer */
    buf = bson_malloc0 (5);
    len = 5;
-   len_le = BSON_UINT32_TO_LE (5);
-   memcpy (buf, &len_le, sizeof (len_le));
+   mlib_write_u32le (buf, len);
```
eramongodb, line 1798:
```suggestion
   mlib_write_u32le (buf, 5);
```
With Clang 20.0.0:
```
warning: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'uint32_t' (aka 'unsigned int') [-Wshorten-64-to-32]
```

src/libbson/src/bson/bson-decimal128.c
```diff
@@ -298,10 +298,12 @@ bson_decimal128_to_string (const bson_decimal128_t *dec, /* IN  */
             *(str_out++) = '0';
          }
 
-         const unsigned radix_stop = (unsigned) BSON_MAX (radix_position - 1, 0);
-         const unsigned n_to_write = significand_digits - radix_stop;
-         for (uint32_t i = 0; i < n_to_write && (str_out - str) < available_bytes; i++) {
-            *(str_out++) = *(significand_read++) + '0';
+         const unsigned dot_pos = (unsigned) BSON_MAX (radix_position - 1, 0);
+         const unsigned n_trailing_digits = num_significant_digits - dot_pos;
```
eramongodb, line 302:
```suggestion
         const unsigned n_trailing_digits = (unsigned) num_significant_digits - dot_pos;
```
With Clang 20.0.0:
```
warning: implicit conversion changes signedness: 'int32_t' (aka 'int') to 'unsigned int' [-Wsign-conversion]
```

src/libmongoc/src/mongoc/mongoc-scram.c
```diff
@@ -1027,7 +1029,7 @@ _mongoc_sasl_prep_impl (const char *name, const char *in_utf8, bson_error_t *err
    utf8_codepoints = bson_malloc (sizeof (uint32_t) * ((size_t) num_chars + 1u)); /* add one for trailing 0 value. */
    const char *c = in_utf8;
 
-   for (size_t i = 0; i < num_chars; ++i) {
+   mlib_foreach_urange (i, num_chars) {
```
eramongodb, line 1032:
```suggestion
   mlib_foreach_irange (i, num_chars) {
```
With Clang 20.0.0:
```
warning: implicit conversion changes signedness: 'ssize_t' (aka 'long') to 'uintmax_t' (aka 'unsigned long') [-Wsign-conversion]
```
Similarly for other calls to `mlib_foreach_irange` in this scope which use `num_chars` (which is `ssize_t`).

src/libmongoc/tests/benchmark-tls-pooled.c
```diff
@@ -74,11 +76,11 @@ main (int argc, char *argv[])
    pool = mongoc_client_pool_new (uri);
    mongoc_client_pool_set_error_api (pool, MONGOC_ERROR_API_VERSION_2);
 
-   for (i = 0; i < num_clients; i++) {
+   mlib_foreach_urange (i, num_clients) {
```
eramongodb, line 79:
```suggestion
   mlib_foreach_irange (i, num_clients) {
```
With GCC:
```
warning: conversion to ‘uintmax_t’ {aka ‘long unsigned int’} from ‘int’ may change the sign of the result [-Wsign-conversion]
```

src/libmongoc/tests/benchmark-tls-pooled.c
```diff
@@ -74,11 +76,11 @@ main (int argc, char *argv[])
...
       pthread_create (&threads[i], NULL, worker, pool);
    }
 
-   for (i = 0; i < num_clients; i++) {
+   mlib_foreach_urange (i, num_clients) {
```
eramongodb, line 83:
```suggestion
   mlib_foreach_irange (i, num_clients) {
```
With GCC:
```
warning: conversion to ‘uintmax_t’ {aka ‘long unsigned int’} from ‘int’ may change the sign of the result [-Wsign-conversion]
```

src/libmongoc/src/mongoc/mongoc-write-command.c
```diff
@@ -698,23 +699,20 @@ _mongoc_write_opmsg (mongoc_write_command_t *command,
    }
 
    do {
-      uint32_t ulen;
-      memcpy (&ulen, command->payload.data + payload_batch_size + payload_total_offset, 4);
-      ulen = BSON_UINT32_FROM_LE (ulen);
+      const int32_t len = mlib_read_u32le (command->payload.data + payload_batch_size + payload_total_offset);
```
eramongodb, line 702: This change produces:
```
warning: conversion to ‘int32_t’ {aka ‘const int’} from ‘uint32_t’ {aka ‘unsigned int’} may change the sign of the result [-Wsign-conversion]
```
Consider reverting/revising these changes to preserve the assertion of `ulen` before the cast to signed `len` to avoid implementation-defined behavior for unsigned->signed conversion.

src/common/tests/test-mlib.c
```diff
@@ -362,13 +390,108 @@ _test_in_range (void)
...
+      mlib_check (buf, streq, "\x04\x03\x02\x01*");
+
+      o = mlib_write_i64le (buf, 0x0102030405060708);
+      mlib_check (o, ptreq, buf + 8);
+      mlib_check (buf, streq, "\x08\x07\x06\x05\x04\x03\x02\x01");
+   }
+}
+
+void
+_test_foreach (void)
```
eramongodb, line 431:
```suggestion
static void
_test_foreach (void)
```
Also applies to other new `_test_*` functions.

eramongodb (changes requested):
Very nice changes so far.

Some additional suggestions:

- Also replace or remove use of `BSON_MIN` and `BSON_MAX` from `<bson/bson-macros.h>` to address the following warning(s) due to the internal comparison expression:

```
src/libbson/src/bson/bson-macros.h:121:30: warning: comparison of integer expressions of different signedness: ‘unsigned int’ and ‘int’ [-Wsign-compare]
  121 | #define BSON_MIN(a, b) (((a) < (b)) ? (a) : (b))
      |                              ^
src/libbson/src/bson/bson-decimal128.c:303:38: note: in expansion of macro ‘BSON_MIN’
  303 |          const unsigned n_to_write = BSON_MIN (n_trailing_digits, available_bytes);
      |                                      ^~~~~~~~
src/libbson/src/bson/bson-macros.h:121:45: warning: operand of ‘?:’ changes signedness from ‘int’ to ‘unsigned int’ due to unsignedness of other operand [-Wsign-compare]
  121 | #define BSON_MIN(a, b) (((a) < (b)) ? (a) : (b))
      |                                             ^~~
src/libbson/src/bson/bson-decimal128.c:303:38: note: in expansion of macro ‘BSON_MIN’
  303 |          const unsigned n_to_write = BSON_MIN (n_trailing_digits, available_bytes);
      |                                      ^~~~~~~~
```

- Also update `ASSERT_CMPINT_HELPER` in `TestSuite.h` to use `mlib_cmp` for its internal equality comparison. Note the non-integer `ASSERT_CMP*` macros currently implemented in terms of `ASSERT_CMPINT_HELPER` (for `double` and `void*`) will need to be special-cased to avoid using incorrect use with the new integer comparison utilities.

src/libbson/src/bson/bson-decimal128.c
vector-of-bool: I'll leave this one but fix the other `irange`/`urange` ones, as I want to address most arithmetic-related `sign-conversion` warnings in a later change that adds ckdint APIs.

vector-of-bool:
@eramongodb

- I considered how to modify or replace uses of `BSON_MIN/BSON_MAX`, but it's difficult in that one cannot statically determine whether the caller intends to have a signed/unsigned result in the case that the signs mismatch. For now, I just let `-Wsign-compare` find the issues and then update the call sites so that the signs agree with what the caller actually intends to happen.
- I can update `ASSERT_CMP*` functions to use this compare, but I could also replace all uses of `ASSERT_CMP<int-type>` with `mlib_check`, which will do the correct assertion with a diagnostic. This will touch hundreds of lines, but significantly simplifies integer assertions. Do you (or anyone else) have any opinions on whether to do this swap?

kevinAlbs (commented):
The consistent use of helpers to read/write integers is much appreciated.

eramongodb (commented):
> Do you (or anyone else) have any opinions on whether to do this swap?

I am fine with deferring those changes to a followup PR.

CMakeLists.txt
```diff
@@ -322,6 +322,9 @@ if(ENABLE_MAINTAINER_FLAGS)
       gnu-like:-Wuninitialized
       # Disabled, for now:
       gnu-like:-Wno-strict-aliasing
+      # Sign-comparison-mismatch:
+      gnu-like:-Wsign-compare
```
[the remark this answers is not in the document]
vector-of-bool: I'll leave that until a later PR related to `-Wconversion`

src/common/src/mlib/intencode.h
```diff
@@ -0,0 +1,161 @@
...
+   uint32_t ret = 0;
+   if (mlib_is_little_endian ()) {
+      // Optimize: The platform uses a LE encoding already
+      memcpy (&ret, buf, sizeof ret);
+   } else {
+      // Portable decode of an LE integer
+      const uint8_t *cptr = (const uint8_t *) buf;
+      mlib_foreach_urange (i, sizeof ret) {
+         ret <<= 8;
+         ret |= cptr[i];
```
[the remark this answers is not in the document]
vector-of-bool: You are correct. This loop tries to decode a BE integer, not an LE one. I can repro on LE platforms by just forcing it to take this branch instead of the `memcpy` one.

src/common/src/mlib/loop.h
```diff
@@ -0,0 +1,165 @@
...
+ * @brief Loop over the elements of a C array
+ *
+ * @param T the type of the array elements
+ * @param Var Identifier to declare as the pointer to the current element
+ * @param Array An expression of array type (not a pointer)
+ */
+#define mlib_foreach_arr(T, Var, Array) mlib_foreach (T, Var, Array, (sizeof Array / sizeof Array[0]))
+
+// clang-format off
+#define _mlibForeachRange(VarType, VarName, StartValue, StopValue, StartVar, StopVar, Counter) \
```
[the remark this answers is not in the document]
vector-of-bool: I can attest that all three main compilers will produce identical code with the macros as with a regular single-layer `for` loop. All loop heads except for the primary one will be flattened and deleted/unrolled, along with unused locals. I suspect that the compiler's initial lowering of `for` produces nearly identical code to the proposed alternative.

(For reference, I first found this idiom from Gustedt (again) ([blogpost](https://gustedt.wordpress.com/2010/08/14/scope-bound-resource-management-with-for-scopes/)))

kevinAlbs (approved):

eramongodb (approved):

src/libbson/src/bson/bson.c
```diff
@@ -328,7 +324,7 @@ BSON_STATIC_ASSERT2 (max_alloc_grow_fits_min_sizet, (uint64_t) BSON_MAX_SIZE * 2
    if (BSON_UNLIKELY ((_length) > BSON_MAX_SIZE - (_list).n_bytes)) { \
       goto append_failure;                                            \
    } else if ((_length) > 0) {                                        \
-      *(_list).current++ = (_bson_append_bytes_arg){                  \
+      *(_list).current++ = (_bson_append_bytes_arg) {                 \
```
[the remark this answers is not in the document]
vector-of-bool: Uggh. My local `clang-format` is updated past the version installed by Poetry, and it now adds a space after compound-initializers every time I save a file. I need to figure out what `.clang-format` setting does this, because it is a chore to remember not to commit them.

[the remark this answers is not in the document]
vector-of-bool: It looks like clang-format 19 now treats compound initializers as C-style casts, so our `SpaceAfterCStyleCast` option now inserts a space. Earlier Clang versions don't seem to have a way to control the whitespace here, and always deleted it.

PR #1859, opened 2025-02-11
CDRIVER-5889: Fix BSON_NORETURN for C23 standard

ghost:
The new BSON_NORETURN macro has implementations for C11, C23, GCC, and a fallback. The C23 version was untested in CI, and led to CDRIVER-5889 when the about-to-be-released GCC 15 switched to the C23 standard as default.

Failures are also observed in normal libmongoc build with a new enough gcc and -DCMAKE_C_STANDARD=23

With this patch, all branches of the BSON_NORETURN definition are working.

Tested on 1.30.0 with a fresh gcc build from master, change 4abac2ffdb071ca9337e4f31fa79cd38df1ac7c3.

Needs to be cherry-picked for 1.x

### DIFF
--- a/src/libbson/src/bson/bson-macros.h
+++ b/src/libbson/src/bson/bson-macros.h
@@ -200,7 +200,7 @@
 
 
 #if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L
-#define BSON_NORETURN noreturn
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
LGTM!

PR #1861, opened 2025-02-14
Fix example code for `bson_writer_t` in docs

LinRaymond2006:
1. `i` is not defined in the example program.
2. Invalid usage of pointers will lead to segmentation fault.


kevinAlbs (approved):
Thank you for the fix @LinRaymond2006

PR #1867, opened 2025-02-15
Coverity fixes

rcsanchez97:
This PR fixes some small Coverity issues (w/ full Evergreen [patch build](https://spruce.mongodb.com/version/67aff103ef57a100079ce47b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)).

@kevinAlbs I wasn't sure who would be available to review, so please feel free to adjust the reviewers as needed.

### DIFF
--- a/src/common/src/common-b64.c
+++ b/src/common/src/common-b64.c
@@ -263,14 +263,20 @@ static const uint8_t mongoc_b64rmap_invalid = 0xff;
 #if defined(BSON_OS_UNIX)
 #include <pthread.h>
 #define mongoc_common_once_t pthread_once_t
-#define mongoc_common_once pthread_once
+#define mongoc_common_once(o, c)             \
+   do {                                      \
+      Assert (pthread_once ((o), (c)) == 0); \
+   } while (0)
 #define MONGOC_COMMON_ONCE_FUN(n) void n (void)
 #define MONGOC_COMMON_ONCE_RETURN return
 #define MONGOC_COMMON_ONCE_INIT PTHREAD_ONCE_INIT
 #else
 #define mongoc_common_once_t INIT_ONCE
 #define MONGOC_COMMON_ONCE_INIT INIT_ONCE_STATIC_INIT
-#define mongoc_common_once(o, c) InitOnceExecuteOnce (o, c, NULL, NULL)
+#define mongoc_common_once(o, c)                            \
+   do {                                                     \
+      Assert (InitOnceExecuteOnce (o, c, NULL, NULL) != 0); \
+   } while (0)
 #define MONGOC_COMMON_ONCE_FUN(n) BOOL CALLBACK n (PINIT_ONCE _ignored_a, PVOID _ignored_b, PVOID *_ignored_c)
 #define MONGOC_COMMON_ONCE_RETURN return true
 #endif
--- a/src/libbson/src/bson/bson-json.c
+++ b/src/libbson/src/bson/bson-json.c
@@ -620,9 +620,10 @@ _bson_json_read_integer (bson_json_reader_t *reader, uint64_t val, int64_t sign)
 
    if (rs == BSON_JSON_REGULAR) {
       BASIC_CB_BAIL_IF_NOT_NORMAL ("integer");
+      BSON_ASSERT (mlib_in_range (int, len));
 
       if (val <= INT32_MAX || (sign == -1 && val <= (uint64_t) INT32_MAX + 1)) {
-         bson_append_int32 (STACK_BSON_CHILD, key, (int) len, (int) (val * sign));
+         bson_append_int32 (STACK_BSON_CHILD, key, (int) len, (int32_t) ((int64_t) val * sign));
       } else if (sign == -1) {
 #if defined(_WIN32) && !defined(__MINGW32__)
          // Unary negation of unsigned integer is deliberate.
--- a/src/libbson/src/jsonsl/jsonsl.c
+++ b/src/libbson/src/jsonsl/jsonsl.c
@@ -1149,6 +1149,9 @@ void jsonsl_jpr_match_state_init(jsonsl_t jsn,
         return;
     }
     jsn->jprs = (jsonsl_jpr_t *)malloc(sizeof(jsonsl_jpr_t) * njprs);
+    if (!jsn->jprs) {
+        return;
+    }
     jsn->jpr_count = njprs;
     jsn->jpr_root = (size_t*)calloc(1, sizeof(size_t) * njprs * jsn->levels_max);
     memcpy(jsn->jprs, jprs, sizeof(jsonsl_jpr_t) * njprs);
--- a/src/libmongoc/src/mongoc/mongoc-counters.c
+++ b/src/libmongoc/src/mongoc/mongoc-counters.c
@@ -111,7 +111,9 @@ mongoc_counters_calc_size (void)
            (n_cpu * n_groups * sizeof (mongoc_counter_slots_t)));
 
 #ifdef BSON_OS_UNIX
-   return BSON_MAX (sysconf (_SC_PAGESIZE), size);
+   long pg_sz = sysconf (_SC_PAGESIZE);
+   BSON_ASSERT (pg_sz > 0);
+   return BSON_MAX (pg_sz, size);
 #else
    return size;
 #endif
--- a/src/libmongoc/src/mongoc/mongoc-stream-tls-openssl.c
+++ b/src/libmongoc/src/mongoc/mongoc-stream-tls-openssl.c
@@ -843,6 +843,10 @@ create_stream_with_ctx (
 mongoc_stream_t *
 mongoc_stream_tls_openssl_new (mongoc_stream_t *base_stream, const char *host, mongoc_ssl_opt_t *opt, int client)
 {
+   if (!opt) {
+      RETURN (NULL);
+   }
+
    SSL_CTX *ssl_ctx = _mongoc_openssl_ctx_new (opt);
 
    if (!ssl_ctx) {
--- a/src/libmongoc/src/mongoc/mongoc-stream-tls.c
+++ b/src/libmongoc/src/mongoc/mongoc-stream-tls.c
@@ -194,13 +194,13 @@ mongoc_stream_tls_new_with_hostname (mongoc_stream_t *base_stream, const char *h
 
    /* !client is only used for testing,
     * when the streams are pretending to be the server */
-   if (!client || opt->weak_cert_validation) {
+   if (opt && (!client || opt->weak_cert_validation)) {
       opt->allow_invalid_hostname = true;
    }
 
 #ifndef _WIN32
    /* Silly check for Unix Domain Sockets */
-   if (!host || (host[0] == '/' && !access (host, F_OK))) {
+   if (opt && (!host || (host[0] == '/' && !access (host, F_OK)))) {
       opt->allow_invalid_hostname = true;
    }
 #endif
@@ -255,13 +255,13 @@ mongoc_stream_tls_new_with_hostname_and_openssl_context (
 
    /* !client is only used for testing,
     * when the streams are pretending to be the server */
-   if (!client || opt->weak_cert_validation) {
+   if (opt && (!client || opt->weak_cert_validation)) {
       opt->allow_invalid_hostname = true;
    }
 
 #ifndef _WIN32
    /* Silly check for Unix Domain Sockets */
-   if (!host || (host[0] == '/' && !access (host, F_OK))) {
+   if (opt && (!host || (host[0] == '/' && !access (host, F_OK)))) {
       opt->allow_invalid_hostname = true;
    }
 #endif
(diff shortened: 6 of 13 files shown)


src/tools/mongoc-stat.c
```diff
@@ -29,6 +29,11 @@
 #include <sys/stat.h>
 #include <unistd.h>
 
+#define BSON_INSIDE
+#define MONGOC_INSIDE
```
vector-of-bool, line 33: These wrapping macros are not needed for `mlib/cmp.h`

src/libmongoc/src/mongoc/mongoc-stream-tls.c
```diff
@@ -255,13 +255,13 @@ mongoc_stream_tls_new_with_hostname_and_openssl_context (
 
    /* !client is only used for testing,
     * when the streams are pretending to be the server */
-   if (!client || opt->weak_cert_validation) {
+   if (opt && (!client || opt->weak_cert_validation)) {
```
vector-of-bool, line 258: Should this check `opt`, or should it assert `opt`?

src/libbson/src/jsonsl/jsonsl.c
```diff
@@ -1149,6 +1149,9 @@ void jsonsl_jpr_match_state_init(jsonsl_t jsn,
         return;
     }
     jsn->jprs = (jsonsl_jpr_t *)malloc(sizeof(jsonsl_jpr_t) * njprs);
+    if (!jsn->jprs) {
+        return;
+    }
     jsn->jpr_count = njprs;
     jsn->jpr_root = (size_t*)calloc(1, sizeof(size_t) * njprs * jsn->levels_max);
```
vector-of-bool, line 1156: This `calloc` could also fail (in which cast the `malloc` above would need to be free'd as well)

src/libmongoc/src/mongoc/mongoc-counters.c
```diff
@@ -111,7 +111,9 @@ mongoc_counters_calc_size (void)
            (n_cpu * n_groups * sizeof (mongoc_counter_slots_t)));
 
 #ifdef BSON_OS_UNIX
-   return BSON_MAX (sysconf (_SC_PAGESIZE), size);
+   long pg_sz = sysconf (_SC_PAGESIZE);
+   BSON_ASSERT (pg_sz > 0);
+   return BSON_MAX (pg_sz, size);
```
vector-of-bool, line 116:
```suggestion
   return BSON_MAX ((size_t) pg_sz, size);
```

src/libmongoc/src/mongoc/mongoc-stream-tls-openssl.c
```diff
@@ -843,6 +843,10 @@ create_stream_with_ctx (
 mongoc_stream_t *
 mongoc_stream_tls_openssl_new (mongoc_stream_t *base_stream, const char *host, mongoc_ssl_opt_t *opt, int client)
 {
+   if (!opt) {
+      RETURN (NULL);
+   }
```
vector-of-bool, line 848: Should this check `opt` or assert `opt`?

src/libmongoc/src/mongoc/mongoc-topology-background-monitoring.c
```diff
@@ -311,10 +311,12 @@ _mongoc_topology_background_monitoring_stop (mongoc_topology_t *topology)
    }
 
    /* Signal all RTT monitors to shut down. */
+   bson_mutex_lock (&topology->tpld_modification_mtx);
    for (size_t i = 0u; i < n_rtt_monitors; i++) {
       server_monitor = mongoc_set_get_item (topology->rtt_monitors, i);
       mongoc_server_monitor_request_shutdown (server_monitor);
    }
+   bson_mutex_unlock (&topology->tpld_modification_mtx);
```
vector-of-bool, line 319: I'm not sure if this mutex is relevant here. Do you have a link to the Coverity warning related to these lines?

vector-of-bool (commented):
A few comments and questions.

src/tools/mongoc-stat.c
rcsanchez97: OK. I am very confused. I am nearly 100% certain that the first time I tried without the macros and the build failed. I had contemplated adding them, but I wasn't sure if it was OK. I asked Ezra and he showed me an example of adding the macros in a test for a case where the direct inclusion was needed. In any event, I removed them and ran a new patch build, and it worked. So I assume that something changed in the underlying code between when I first observed the failure and now. Any event, a commit is incoming that removes these.

src/libmongoc/src/mongoc/mongoc-topology-background-monitoring.c
rcsanchez97: I sent you the link in a DM.

src/libmongoc/src/mongoc/mongoc-stream-tls.c
rcsanchez97: Yeah, it seems that an assert works here. I have a clear recollection of attempting to use `BSON_ASSERT` and it triggering a bunch of failures, but I can't find it in my history of patch builds. I attempted another patch build with `BSON_ASSERT` rather than the Boolean check and that seems to work. Commit incoming.

src/libmongoc/src/mongoc/mongoc-stream-tls-openssl.c
rcsanchez97: Same as the other instance of checking `opt`. The assert seems to work (and is preferrable), so commit incoming.

src/libbson/src/jsonsl/jsonsl.c
rcsanchez97: Good catch. I missed that. Commit incoming.

src/libbson/src/jsonsl/jsonsl.c
```diff
@@ -1149,8 +1149,15 @@ void jsonsl_jpr_match_state_init(jsonsl_t jsn,
         return;
     }
     jsn->jprs = (jsonsl_jpr_t *)malloc(sizeof(jsonsl_jpr_t) * njprs);
+    if (!jsn->jprs) {
+        return;
+    }
```
kevinAlbs, line 1154:
```suggestion
    jsn->jprs = (jsonsl_jpr_t *) bson_malloc (sizeof (jsonsl_jpr_t) * njprs);
```

Suggest using `bson_malloc`. `bson_malloc` already checks (and aborts) if allocation fails. I expect that will fix the Coverity warning. Using `bson_malloc` ensures allocators set with `bson_mem_set_vtable` are used.

Also replace `free(jsn->jprs);` with `bson_free(jsn->jprs);`

src/libbson/src/jsonsl/jsonsl.c
```diff
@@ -1149,8 +1149,15 @@ void jsonsl_jpr_match_state_init(jsonsl_t jsn,
...
     jsn->jpr_count = njprs;
     jsn->jpr_root = (size_t*)calloc(1, sizeof(size_t) * njprs * jsn->levels_max);
+    if (!jsn->jpr_root) {
+        free(jsn->jprs);
+        return;
+    }
```
kevinAlbs, line 1160:
```suggestion
    jsn->jpr_root = (size_t *) bson_malloc0 (sizeof (size_t) * njprs * jsn->levels_max);
```

Similarly, use `bson_malloc0` in place of `calloc`. Suggest updating other `malloc` / `free` calls in this file.

rcsanchez97: OK. That makes sense. Should I do that as a drive-by in this PR, or a separate PR?

kevinAlbs: I think it is OK to include in this PR. Though the other `malloc` calls check their returns, I think it is a related enough change to include in this PR.

src/common/src/common-b64.c
```diff
@@ -262,14 +262,20 @@ static const uint8_t mongoc_b64rmap_invalid = 0xff;
 #if defined(BSON_OS_UNIX)
```
kevinAlbs, line 262: I expect this entire `#if / #else / #endif` block can be removed. Add `#include <common-thread-private.h>` and instead use the `bson_once*` and `BSON_ONCE*` APIs. `bson_once` already has an assert.

src/libbson/src/bson/bson-json.c
```diff
@@ -622,9 +622,10 @@ _bson_json_read_integer (bson_json_reader_t *reader, uint64_t val, int64_t sign)
 
    if (rs == BSON_JSON_REGULAR) {
       BASIC_CB_BAIL_IF_NOT_NORMAL ("integer");
+      BSON_ASSERT (mlib_in_range (int, len));
```
kevinAlbs, line 625: Rather than assert, suggest checking if `len` is greater than `INT_MAX` and returning an error. I expect this applies to all types being read. That may avoid a possible crash if the input JSON has a very large (over INT_MAX length) key.

Here is a commit with a test (which is only run when the environment variable `MONGOC_TEST_LARGE_ALLOCATIONS=ON` is defined) and [suggested change](https://github.com/mongodb/mongo-c-driver/commit/d433c227e44226701b7a823fa4bd30a09ac61bad).

rcsanchez97: done

src/libbson/src/jsonsl/jsonsl.c
```diff
@@ -1052,9 +1052,9 @@ void jsonsl_jpr_match_state_init(jsonsl_t jsn,
     if (njprs == 0) {
         return;
     }
-    jsn->jprs = (jsonsl_jpr_t *)malloc(sizeof(jsonsl_jpr_t) * njprs);
+    jsn->jprs = (jsonsl_jpr_t *) bson_malloc (sizeof (jsonsl_jpr_t) * njprs);
```
kevinAlbs, line 1055: Also update `jsonsl_jpr_match_state_cleanup` to use `bson_free`, rather than `free`.

src/libmongoc/src/mongoc/mongoc-stream-tls-openssl.c
```diff
@@ -828,6 +828,8 @@ create_stream_with_ctx (
 mongoc_stream_t *
 mongoc_stream_tls_openssl_new (mongoc_stream_t *base_stream, const char *host, mongoc_ssl_opt_t *opt, int client)
 {
+   BSON_ASSERT (opt);
```
kevinAlbs: Suggest using `BSON_ASSERT_PARAM` for a more specific error message on failure. Suggest also asserting `base_stream` for consistency (as is done in `create_stream_with_ctx`).

```suggestion
   BSON_ASSERT_PARAM (base_stream);
   BSON_ASSERT_PARAM (opt);
```

src/libmongoc/src/mongoc/mongoc-stream-tls.c
```diff
@@ -155,13 +155,13 @@ mongoc_stream_tls_new_with_hostname (mongoc_stream_t *base_stream, const char *h
...
```
kevinAlbs: Since this function previously dereferenced `opt`, I suggest instead adding:

```c
BSON_ASSERT_PARAM (base_stream);
BSON_ASSERT_PARAM (opt);
```

To indicate that `opt` is required. That may help simplify (do not need to handle a possibly NULL `opt`).

src/libmongoc/src/mongoc/mongoc-stream-tls.c
```diff
@@ -209,6 +209,7 @@ mongoc_stream_tls_new_with_hostname_and_openssl_context (
    mongoc_stream_t *base_stream, const char *host, mongoc_ssl_opt_t *opt, int client, SSL_CTX *ssl_ctx)
 {
    BSON_ASSERT (base_stream);
+   BSON_ASSERT (opt);
```
kevinAlbs:
```suggestion
   BSON_ASSERT_PARAM (base_stream);
   BSON_ASSERT_PARAM (opt);
```

src/libmongoc/src/mongoc/mongoc-topology-background-monitoring.c
kevinAlbs: I strongly suspect there are existing thread-safety issues in `_mongoc_topology_background_monitoring_stop`.

I suspect between storing the lengths and accessing the sets, the set of server monitors could possibly change (via `_mongoc_topology_background_monitoring_reconcile`.

I recommend reverting this change. A more comprehensive fix can be done in the future to address the coverity warning.

rcsanchez97:
Here is [the patch build](https://spruce.mongodb.com/version/687514b6c916d90007132fc7/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) for the latest changes.

rcsanchez97:
I discussed with @vector-of-bool in Slack and he said he doesn't feel like he needs to review this again, so I have removed the request for his review.

PR #1868, opened 2025-02-15
CDRIVER-5641: BSON Binary Vector Subtype Support

ghost:
* Resolves CDRIVER-5641 by documenting and implementing new BSON Binary Vector APIs
* Synced bson_binary_vector and bson_corpus spec tests from specifications commit 585b797c110b6709f81def6200b946b94c8d9c55
* tests: Full support for the existing version of the spec test suite.
* tests: TODOs mark spec test improvements that depend on DRIVERS-3095 and DRIVERS-3097.
* tests: Additional API usage and fuzzing.
* New BSON Binary APIs: bson_append_binary_uninit, bson_iter_binary_equal, bson_iter_binary_subtype, bson_iter_overwrite_binary
* Extend existing big-endian byte swapping API for 32-bit float
* Compiler support for restricted pointer aliasing using BSON_RESTRICT.

The concept: no new owned data types are defined. Instead, we encourage access to vector data that's already allocated inside a bson_t. When you append a vector to a bson_t, the result is a *view*. Views are effectively pointers decorated with type and length information, conferring a validity guarantee about the memory they reference.

The view data types are included in the ABI, and several performance-critical accessors are implemented as inline functions.

For review I'd recommend starting with the new `binary_vector.rst` documentation page and following the links from there. The documentation page for each view type includes a short example. These examples are also included in the test suite, in an adapted form.

It's tempting to automatically generate the repetitive portions of this patch, but so far I have opted instead to avoid the extra complexity.

Commits are separated into new files, synced spec files, and modified files.

Verified with:
* evergreen builds: https://spruce.mongodb.com/version/67b023355d3da6000759923e/tasks
* libbson tests compiled and run on powerpc debian in qemu, to cover 32-bit and big endian

### DIFF
--- a/src/libbson/src/bson/bson-macros.h
+++ b/src/libbson/src/bson/bson-macros.h
@@ -341,23 +341,39 @@ _bson_assert_failed_on_param (const char *param, const char *func)
 #define BSON_TYPEOF typeof
 #endif
 
+/**
+ * @brief Statically annotate an entity as deprecated, including the given deprecation message
+ *
+ * @param Message The message to be included in a deprecation warning. This
+ * should be a string literal.
+ */
+#define BSON_DEPRECATED(Message) _bsonDeprecatedImpl (Message)
 
-#if BSON_GNUC_CHECK_VERSION(3, 1)
-#define BSON_GNUC_DEPRECATED __attribute__ ((__deprecated__))
+// Pick the appropriate implementation of a deprecation attribute
+#if defined(_MSC_VER)
+// For MSVC, emit __declspec(deprecated(Msg))
+#define _bsonDeprecatedImpl(Msg) __declspec (deprecated (Msg))
+#elif defined(__GNUC__) && (defined(__clang__) || BSON_GNUC_CHECK_VERSION(4, 5))
+// For new enough Clang and GCC, emit __attribute__((__deprecated__(Msg)))
+#define _bsonDeprecatedImpl(Msg) __attribute__ ((__deprecated__ (Msg)))
+#elif defined(__GNUC__)
+// For older GCC, emit deprecation attribute without the message
+#define _bsonDeprecatedImpl(Msg) __attribute__ ((__deprecated__))
 #else
-#define BSON_GNUC_DEPRECATED
+// For other compilers, emit nothing
+#define _bsonDeprecatedImpl(Msg)
 #endif
+
+#define BSON_DEPRECATED_FOR(F) BSON_DEPRECATED ("This API is deprecated. Use " #F " instead.")
+
+#define BSON_GNUC_DEPRECATED BSON_DEPRECATED ("This API is deprecated")
+#define BSON_GNUC_DEPRECATED_FOR(F) BSON_DEPRECATED_FOR (F)
 
 #define BSON_CONCAT_IMPL(a, ...) a##__VA_ARGS__
 #define BSON_CONCAT(a, ...) BSON_CONCAT_IMPL (a, __VA_ARGS__)
 #define BSON_CONCAT3(a, b, c) BSON_CONCAT (a, BSON_CONCAT (b, c))
 #define BSON_CONCAT4(a, b, c, d) BSON_CONCAT (BSON_CONCAT (a, b), BSON_CONCAT (c, d))
 
-#if BSON_GNUC_CHECK_VERSION(4, 5)
-#define BSON_GNUC_DEPRECATED_FOR(f) __attribute__ ((deprecated ("Use " #f " instead")))
-#else
-#define BSON_GNUC_DEPRECATED_FOR(f) BSON_GNUC_DEPRECATED
-#endif
 
 /**
  * @brief String-ify the given argument
--- a/.clang-format
+++ b/.clang-format
@@ -53,6 +53,9 @@ AlwaysBreakBeforeMultilineStrings: false
 AlwaysBreakTemplateDeclarations: MultiLine
 AttributeMacros:
   - __capability
+  - BSON_GNUC_WARN_UNUSED_RESULT
+  - BSON_DEPRECATED
+  - BSON_DEPRECATED_FOR
 BinPackArguments: false
 BinPackParameters: false
 BitFieldColonSpacing: Both
--- a/.evergreen/config_generator/components/c_std_compile.py
+++ b/.evergreen/config_generator/components/c_std_compile.py
@@ -6,7 +6,7 @@ from config_generator.components.funcs.find_cmake_latest import FindCMakeLatest
 
 from config_generator.etc.distros import find_large_distro
 from config_generator.etc.distros import make_distro_str
-from config_generator.etc.distros import to_cc
+from config_generator.etc.distros import compiler_to_vars
 from config_generator.etc.function import Function
 from config_generator.etc.utils import bash_exec
 
@@ -17,14 +17,31 @@ TAG = 'std-matrix'
 # pylint: disable=line-too-long
 # fmt: off
 MATRIX = [
-    ('debian92',          'clang',     None,   [11,   ]),
-    ('debian10',          'clang',     None,   [11,   ]),
-    ('debian10',          'gcc',       None,   [11, 17]),
-    ('debian11',          'clang',     None,   [11,   ]),
-    ('debian11',          'gcc',       None,   [11, 17]),
-    ('ubuntu2004',        'clang',     None,   [11,   ]),
-    ('ubuntu2004',        'gcc',       None,   [11,   ]),
-    ('windows-vsCurrent', 'vs2017x64', None,   [11, 17]),
+    ('rhel80',     'clang',    None, [99, 11, 17,   ]), # Clang 7.0
+    ('ubuntu2004', 'clang-10', None, [99, 11, 17, 23]), # Clang 10.0 (max: C2x)
+    ('rhel84',     'clang',    None, [99, 11, 17, 23]), # Clang 11.0 (max: C2x)
+    ('ubuntu2204', 'clang-12', None, [99, 11, 17, 23]), # Clang 12.0 (max: C2x)
+    ('rhel90',     'clang',    None, [99, 11, 17, 23]), # Clang 13.0 (max: C2x)
+    ('rhel91',     'clang',    None, [99, 11, 17, 23]), # Clang 14.0 (max: C2x)
+    ('rhel92',     'clang',    None, [99, 11, 17, 23]), # Clang 15.0 (max: C2x)
+    ('rhel93',     'clang',    None, [99, 11, 17, 23]), # Clang 16.0 (max: C2x)
+    ('rhel94',     'clang',    None, [99, 11, 17, 23]), # Clang 17.0 (max: C2x)
+    ('rhel95',     'clang',    None, [99, 11, 17, 23]), # Clang 18.0 (max: C23)
+
+    ('rhel76',     'gcc',    None, [99, 11,       ]), # GCC 4.8 (max: C11)
+    ('rhel80',     'gcc',    None, [99, 11, 17,   ]), # GCC 8.2 (max: C17)
+    ('debian10',   'gcc-8',  None, [99, 11, 17,   ]), # GCC 8.3 (max: C17)
+    ('rhel84',     'gcc',    None, [99, 11, 17,   ]), # GCC 8.4 (max: C17)
+    ('ubuntu2004', 'gcc-9',  None, [99, 11, 17, 23]), # GCC 9.4 (max: C2x)
+    ('debian11',   'gcc-10', None, [99, 11, 17, 23]), # GCC 10.2 (max: C2x)
+    ('rhel90',     'gcc',    None, [99, 11, 17, 23]), # GCC 11.2 (max: C2x)
+    ('rhel92',     'gcc',    None, [99, 11, 17, 23]), # GCC 11.3 (max: C2x)
+    ('rhel94',     'gcc',    None, [99, 11, 17, 23]), # GCC 11.4 (max: C2x)
+    ('rhel95',     'gcc',    None, [99, 11, 17, 23]), # GCC 11.5 (max: C2x)
+
+    ('windows-vsCurrent', 'vs2017x64', None, [99, 11, 17, 'latest']), # Max: C17, clatest (C2x)
+    ('windows-vsCurrent', 'vs2019x64', None, [99, 11, 17, 'latest']), # Max: C17, clatest (C2x)
+    ('windows-vsCurrent', 'vs2022x64', None, [99, 11, 17, 'latest']), # Max: C17, clatest (C2x)
 ]
 # fmt: on
 # pylint: enable=line-too-long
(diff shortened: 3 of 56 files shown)


src/libbson/src/bson/bson-vector.h
```diff
@@ -0,0 +1,591 @@
...
+{
+   size_t length_bytes = bson_vector_packed_bits_view_length_bytes (view);
+   if (BSON_LIKELY (vector_offset_bytes <= length_bytes && byte_count <= length_bytes - vector_offset_bytes)) {
+      if (byte_count == length_bytes - vector_offset_bytes && byte_count >= 1u) {
+         // This write touches the last byte in the vector:
+         // special-case that byte so we can ensure unused bits remain set to zero.
+         size_t other_bytes = byte_count - 1u;
+         memcpy (view.binary.data + BSON_VECTOR_HEADER_LEN + vector_offset_bytes, packed_values, other_bytes);
+         view.binary.data[BSON_VECTOR_HEADER_LEN + vector_offset_bytes + other_bytes] =
+            ((uint8_t) 0xFF << bson_vector_packed_bits_view_padding (view)) & packed_values[other_bytes];
```
eramongodb, line 513:
```suggestion
            (UINT8_C (0xFF) << bson_vector_packed_bits_view_padding (view)) & packed_values[other_bytes];
```
Minor suggestion to avoid the explicit cast.

src/libbson/tests/test-bson-vector.c
```diff
@@ -0,0 +1,1400 @@
...
+   // - convert "-inf" to a double -Infinity
+
+   bson_iter_t iter_in;
+   BSON_ASSERT (bson_iter_init (&iter_in, array_in));
+   bson_array_builder_t *builder = bson_array_builder_new ();
+   while (bson_iter_next (&iter_in)) {
+      if (BSON_ITER_HOLDS_UTF8 (&iter_in) && 0 == strcmp ("inf", bson_iter_utf8 (&iter_in, NULL))) {
+         BSON_ASSERT (bson_array_builder_append_double (builder, INFINITY));
+      } else if (BSON_ITER_HOLDS_UTF8 (&iter_in) && 0 == strcmp ("-inf", bson_iter_utf8 (&iter_in, NULL))) {
+         BSON_ASSERT (bson_array_builder_append_double (builder, -INFINITY));
```
eramongodb, line 75:
```suggestion
      if (BSON_ITER_HOLDS_UTF8 (&iter_in) && 0 == strcmp ("inf", bson_iter_utf8 (&iter_in, NULL))) {
         BSON_ASSERT (bson_array_builder_append_double (builder, 1.0 / 0.0));
      } else if (BSON_ITER_HOLDS_UTF8 (&iter_in) && 0 == strcmp ("-inf", bson_iter_utf8 (&iter_in, NULL))) {
         BSON_ASSERT (bson_array_builder_append_double (builder, -1.0 / 0.0));
```
Consider using `double` arithmetic instead of `INFINITY` (`float`) to address `-Wdouble-promotion` warnings. Also applies to other instances of `INFINITY` in this file.

src/libbson/tests/test-bson-vector.c
```diff
@@ -0,0 +1,1400 @@
...
+      return true;
+   } else {
+      return false;
+   }
+}
+
+static void
+hex_str_to_bson (bson_t *bson_out, const char *hex_str)
+{
+   uint32_t size = strlen (hex_str) / 2;
```
eramongodb, line 159:
```suggestion
   uint32_t size = (uint32_t) strlen (hex_str) / 2u;
```
Addresses `-Wshorten-64-to-32` warnings.

src/libbson/tests/test-bson-vector.c
```diff
@@ -0,0 +1,1400 @@
...
+                      TEST_ERROR_DOMAIN,
+                      TEST_ERROR_CODE,
+                      "nonzero 'padding' parameter (%" PRId64
+                      ") for zero-length append_vector_packed_bits_from_packed_array",
+                      padding);
+      return false;
+   }
+
+   bson_vector_packed_bits_view_t view;
+   if (bson_append_vector_packed_bits (bson, key, key_length, byte_count * 8 - padding, &view)) {
```
eramongodb, line 143:
```suggestion
   if (bson_append_vector_packed_bits (bson, key, key_length, byte_count * 8u - (size_t) padding, &view)) {
```
Addresses `-Wsign-conversion` warnings (guarded by `padding < 0` above).

src/libbson/tests/test-bson-vector.c
```diff
@@ -0,0 +1,1400 @@
...
+                           (int) i);
+            }
+
+            bool is_sorta_equal = false;
+            if (expected_double != expected_double) {
+               // Expect NaN, any type is fine.
+               if (actual_double != actual_double) {
+                  is_sorta_equal = true;
+               }
+            } else if (expected_double == 0 || expected_double * 0 != 0) {
```
eramongodb, line 335:
```suggestion
            } else if (expected_double == 0.0 || expected_double * 0.0 != 0.0) {
```
Addresses `-Wdouble-promotion` warnings and better indicates these are floating-point equality comparisons (type consistency).

src/libbson/tests/test-bson-vector.c
```diff
@@ -0,0 +1,1400 @@
...
+
+static void
+test_bson_vector_view_api_fuzz_int8 (void)
+{
+   size_t current_length = 0;
+   bson_t vector_doc = BSON_INITIALIZER;
+   int8_t *expected_elements = bson_malloc (MAX_TESTED_VECTOR_LENGTH * sizeof *expected_elements);
+   int8_t *actual_elements = bson_malloc (MAX_TESTED_VECTOR_LENGTH * sizeof *actual_elements);
+   for (int fuzz_iter = 0; fuzz_iter < FUZZ_TEST_ITERS; fuzz_iter++) {
+      int r = rand ();
```
eramongodb, line 968:
```suggestion
      size_t r = _mongoc_simple_rand_size_t ();
```
Recommend using the `_mongoc_simple_rand_*` functions provided by `mongoc-util-private.h` for better type consistency and to avoid increasing our raw `rand()` footprint.

Suggestion also applies to other functions below.

src/libbson/tests/test-bson-vector.c
```diff
@@ -0,0 +1,1400 @@
...
+      int r_operation = r & 0xF;
+      int r_param = r >> 4;
```
eramongodb, line 970:
```suggestion
      size_t r_operation = r & 0xFu;
      size_t r_param = r >> 4;
```
To reduce the need for casts to `size_t` below (given `size_t r`).

Suggestion also applies to other functions below.

src/libbson/src/bson/bson-vector.h
```diff
@@ -0,0 +1,591 @@
...
+static BSON_INLINE bool
+bson_vector_packed_bits_view_pack_bool (bson_vector_packed_bits_view_t view,
+                                        const bool *BSON_RESTRICT unpacked_values,
+                                        size_t element_count,
+                                        size_t vector_offset_elements)
+{
+   size_t length = bson_vector_packed_bits_view_length (view);
+   if (BSON_LIKELY (vector_offset_elements <= length && element_count <= length - vector_offset_elements)) {
+      while (element_count > 0) {
+         uint8_t *BSON_RESTRICT packed_byte = &view.binary.data[BSON_VECTOR_HEADER_LEN + (vector_offset_elements >> 3)];
```
eramongodb, line 563:
```suggestion
         uint8_t *BSON_RESTRICT packed_byte = view.binary.data + BSON_VECTOR_HEADER_LEN + (vector_offset_elements >> 3);
```
Minor suggestion to avoid an intermediate subscript.

src/libbson/src/bson/bson-vector.h
```diff
@@ -0,0 +1,591 @@
...
+         if (element_count >= 8 && (vector_offset_elements & 7) == 0) {
+            uint8_t complete_byte = 0;
+            unsigned i;
+            for (i = 0; i < 8; i++) {
```
eramongodb, line 567:
```suggestion
            for (int i = 0; i < 8; i++) {
```
The C Driver requires C99 or newer since 1.24.0 (CDRIVER-4136). Separation of declaration and code is not necessary.

Suggestion also applies to other instances of stray induction variables in this PR.

src/libbson/src/bson/bson-vector.h
```diff
@@ -0,0 +1,591 @@
...
+
+BSON_EXPORT (bool)
+bson_array_builder_append_vector_elements (struct _bson_array_builder_t *builder, const bson_iter_t *iter);
+
+
+BSON_EXPORT (bool)
+bson_append_vector_int8 (
+   bson_t *bson, const char *key, int key_length, size_t element_count, bson_vector_int8_view_t *view_out);
+
+#define BSON_APPEND_VECTOR_INT8(b, key, count, view) bson_append_vector_int8 (b, key, (int) strlen (key), count, view)
```
eramongodb, line 198:
```suggestion
BSON_EXPORT (bool)
bson_append_vector_int8_uninit (
   bson_t *bson, const char *key, int key_length, size_t element_count, bson_vector_int8_view_t *view_out);

#define BSON_APPEND_VECTOR_INT8_UNINIT(b, key, count, view) \
   bson_append_vector_int8_uninit (b, key, (int) strlen (key), count, view)
```
Consider suffixing `_uninit` and `_UNINIT` to this function (and the other two functinons below for float32 and packed_bits) for consistency with calls to `bson_append_binary_uninit` and better communicate our expectation for users to fill in the uninitialized bytes.

src/libbson/src/bson/bson-iter.h
```diff
@@ -40,6 +40,15 @@ BSON_BEGIN_DECLS
...
 #define BSON_ITER_HOLDS_BINARY(iter) (bson_iter_type ((iter)) == BSON_TYPE_BINARY)
 
+#define BSON_ITER_HOLDS_VECTOR(iter) \
+   (BSON_ITER_HOLDS_BINARY (iter) && bson_iter_binary_subtype (iter) == BSON_SUBTYPE_VECTOR)
+
+#define BSON_ITER_HOLDS_VECTOR_INT8(iter) (bson_vector_int8_const_view_from_iter (NULL, iter))
+
+#define BSON_ITER_HOLDS_VECTOR_FLOAT32(iter) (bson_vector_float32_const_view_from_iter (NULL, iter))
+
+#define BSON_ITER_HOLDS_VECTOR_PACKED_BITS(iter) (bson_vector_packed_bits_const_view_from_iter (NULL, iter))
```
eramongodb, line 50:
```suggestion
#define BSON_ITER_HOLDS_VECTOR_PACKED_BIT(iter) (bson_vector_packed_bit_const_view_from_iter (NULL, iter))
```

The "BSON Binary Subtype 9" spec always refers to "PACKED_BIT" in the singular (i.e. "a PACKED_BIT vector"). Should we also be using the singular in our API?

eramongodb (changes requested):

src/libbson/src/bson/bson-vector.h
eramongodb: Closer examination of the current state of our public headers reveals we do not have any such for loop (with a declaration) in a public header even after the C99 minimum requirement was implemented. I would not have expected this to break builds since we should be enforcing C99 or newer standards by default, but I guess its attempted usage expose an implementation bug or compatibility issue with an old compiler version. Since this falls outside the scope of this PR, I am happy to defer this issue.

src/libbson/src/bson/bson-macros.h
```diff
@@ -210,6 +210,18 @@
...
+#if defined(__GNUC__)
+#define BSON_RESTRICT __restrict__
+#elif defined(_MSC_VER)
+#define BSON_RESTRICT __restrict
+#elif !defined(__cplusplus)
+// C99 (not C++)
+#define BSON_RESTRICT restrict
+#else
+#define BSON_RESTRICT
+#endif
```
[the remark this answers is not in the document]
eramongodb: That was far more detailed than I was expected. Examining the quality of generated code by decompiling the assembly back into C a very cool idea. 👀 Thank you very much!

src/libbson/tests/test-bson-vector.c
eramongodb: Did not realize UBSan would prohibit these operations. That is unfortunate. 😢
I am fine with leaving it as-written and deferring the "`INFINITY` for `double`" problem to another PR.

src/libbson/tests/test-bson-vector.c
eramongodb: The rand functions definitely leave a lot to be desired w.r.t. performance. Their addition in https://github.com/mongodb/mongo-c-driver/pull/898 was very much a "correctness first, performance later" decision. The suggestion to use them here was only for `rand()`-avoidance and reuse of existing API to easily achieve `size_t` type consistency without requiring explicit casts (i.e. in module expressions). imo given the execution of these tests are still under a second in total, I do not think the performance tradeoff is especially significant given the overall execution time of the test suite. If you still prefer the explicit `rand()`, I am fine with that decision.

src/libbson/tests/test-bson-vector.c
```diff
@@ -0,0 +1,1581 @@
...
+         for (size_t i = 0; i < new_length; i++) {
+            expected_elements[i] = (int8_t) (uint8_t) rand ();
+         }
+         BSON_ASSERT (bson_vector_int8_view_write (view, expected_elements, new_length, 0));
+         current_length = new_length;
+
+      } else if (r_operation < 7) {
+         // Partial write
+         size_t element_count = r_param % current_length;
+         size_t offset = rand () % (current_length - element_count);
```
eramongodb, line 988:
```suggestion
         size_t offset = (size_t) rand () % (current_length - element_count);
```

Explicit invocation of `rand ()` in modulo expressions still require an explicit cast to address `-Wsign-conversion` warnings for `int` -> `size_t`.

src/libbson/tests/test-bson-vector.c
```diff
@@ -0,0 +1,1581 @@
...
+         // Partial read
+         size_t element_count = r_param % current_length;
+         size_t offset = rand () % (current_length - element_count);
+         bson_vector_float32_const_view_t view;
+         bson_iter_t iter;
+         BSON_ASSERT (bson_iter_init_find (&iter, &vector_doc, "vector"));
+         BSON_ASSERT (bson_vector_float32_const_view_from_iter (&view, &iter));
+         BSON_ASSERT (bson_vector_float32_const_view_read (view, actual_elements, element_count, offset));
+         for (size_t i = 0; i < element_count; i++) {
+            BSON_ASSERT (actual_elements[i] == expected_elements[i + offset]);
```
eramongodb, line 1064:
```suggestion
            ASSERT_EQUAL_DOUBLE (actual_elements[i], expected_elements[i + offset]);
```
Dedicated test assertion macro + addresses `-Wfloat-equal` warnings.

On the other hand, `ASSERT_EQUAL_DOUBLE`'s allowance of up to 20% error (according to its failure message) may be a bit much... 🤔

src/libbson/tests/test-bson-vector.c
```diff
@@ -0,0 +1,1581 @@
...
+      unsigned r = (unsigned) rand ();
+      unsigned r_operation = r & 0xFu;
+      size_t r_param = r >> 4;
+
+      if (current_length == 0 || r_operation == 15) {
+         // Resize and fill
+         size_t new_length = (size_t) r_param % MAX_TESTED_VECTOR_LENGTH;
+         bson_reinit (&vector_doc);
+         bson_vector_int8_view_t view;
+         BSON_ASSERT (BSON_APPEND_VECTOR_INT8_UNINIT (&vector_doc, "vector", new_length, &view));
```
eramongodb, line 978:
```suggestion
         ASSERT (BSON_APPEND_VECTOR_INT8_UNINIT (&vector_doc, "vector", new_length, &view));
```
Recommend using `ASSERT*` test assertion macros in test code (test-specific error messages + includes some [additional operations](https://github.com/mongodb/mongo-c-driver/blob/d6f975d175ec42487f2a1e19f55c87a98f996cc2/src/libmongoc/tests/TestSuite.h#L115) for test failure messages) rather than `BSON_ASSERT`.

eramongodb (approved):
Some minor suggestions remaining; otherwise, LGTM.

src/libbson/src/bson/bson-vector.c
vector-of-bool, line 1: My only concern is the large amount of code duplication between vector types and the `const`/non-`const` variants. I can't think of a pretty solution that doesn't involve macro trickery, though.

vector-of-bool (approved):
LGTM

PR #1870, opened 2025-02-17
Fix signedness mismatch in `_bson_reader_data_read` and `_bson_reader_handle_read`

LinRaymond2006:
In `_bson_reader_data_read` and `_bson_reader_handle_read`, `blen` is a signed integer type, which will wrap around when handling BSON documents larger than 2GB. This occurs because `BSON_MAX_SIZE` is defined as `(1U << 31) - 1`, which is 2GB larger than maximum value a signed integer can represent.


LinRaymond2006:
Okay, sorry for the confusion caused.

PR #1874, opened 2025-02-19
Add description for `dst` parameter of `bson_value_copy`

LinRaymond2006:
`dst` should be uninitialized to prevent memory leak.


kevinAlbs (approved):
Thank you!

PR #1881, opened 2025-02-20
[CDRIVER-5881] Remove the "forwarding" headers

vector-of-bool:
Refer: CDRIVER-5881

The "forwarding" headers were added in 1.13.0 to support programs that include "mongoc.h" or "bson.h" directly. We now require that the header names be prefixed with their parent directory name.

This removes the header files, the install rules, and the uninstaller's test cases for them.


kevinAlbs (approved):

rcsanchez97 (approved):
LGTM

PR #1882, opened 2025-02-20
CDRIVER-5904 update scripts and release instructions for SilkBomb 2.0

eramongodb:
Resolves CDRIVER-5904 for the master branch (EVG project: `mongo-c-driver`). Verified by [this patch](https://spruce.mongodb.com/version/67b7819a21b2e60007435fc8).

Applies similar changes to https://github.com/mongodb/mongo-cxx-driver/pull/1344 but for the C Driver. Unlike the C++ Driver, the C Driver's `sbom` task only performs `validate` + `augment` and simply uploads the Augmented SBOM file (no diffs) so it may be examined and downloaded during release steps when needed.

Due to the current inability to download the Augmented SBOM outside of Evergreen with SilkBomb 2.0, the `+sbom-download` task and related commands, flags, and instructions are removed. In their place, a new release requirement is that `etc/augmented-sbom.json` is present when running `+release-archive` or `+signed-release` (this file does not need to be committed into the repo, it just needs to be present).

This file must be downloaded from a recent execution of the `sbom` task (both commit builds and patch builds are OK!) during the release process prior to executing `+release-archive` or `+signed-release`. If the Augmented SBOM is not present, the `COPY` command in `+release-archive` will fail with the following error (paraphrased for brevity):

```
prefix=[...] ref=HEAD
--> COPY etc/augmented-sbom.json cyclonedx.sbom.json
[no output]
ERROR Earthfile:202:4
      The command
          COPY etc/augmented-sbom.json cyclonedx.sbom.json
      failed: failed to walk [...]/etc: lstat [...]/etc: no such file or directory
```

The `+sbom-update` and the new `+sbom-validate` Earthly commands do not require being run inside Evergreen.


Earthfile
```diff
@@ -269,7 +270,7 @@ signed-release:
     LET rel_tgz = "$rel_dir/$stem.tar.gz"
     LET rel_asc = "$rel_dir/$stem.tar.gz.asc"
     # Make the release archive:
-    COPY (+release-archive/ --branch=$sbom_branch --prefix=$stem --ref=$ref) $rel_dir/
+    COPY (+release-archive/ --prefix=$stem --ref=$ref) $rel_dir/
```
vector-of-bool, line 273: Can delete the `ARG` `sbom_branch` above since it is now unused.

Earthfile
```diff
@@ -296,63 +297,27 @@ sbom-generate:
...
-# sbom-download :
-#   Download an augmented SBOM from the Silk server for the given branch. Exports
-#   the artifact as /augmented-sbom.json
-#
-# Requires credentials for silk access.
-sbom-download:
+# sbom-validate:
+#   Validate the SBOM Lite for the given branch.
+sbom-validate:
     FROM artifactory.corp.mongodb.com/dockerhub/library/alpine:3.20
```
vector-of-bool, line 311: This `FROM` line can be deleted since the next line overrides it immediately.

docs/dev/releasing.rst
```diff
@@ -381,15 +381,14 @@ running :any:`+signed-release`, one will need to set up some environment that is
 required for it to succeed:
 
 1. :ref:`Authenticate with Artifactory <earthly.artifactory-auth>`
-2. Set the Earthly secrets required for the :any:`+sign-file` and
-   :any:`+sbom-download` targets.
+2. Set the Earthly secrets required for the :any:`+sign-file` target.
+3. Download a recent augmented SBOM from Evergreen and save it to `etc/augmented-sbom.json`.
```
vector-of-bool, line 385:
```suggestion
3. Download a recent augmented SBOM from Evergreen, created by the standalone `sbom` task in the "SBOM" variant, and save it to `etc/augmented-sbom.json`.
```

.evergreen/config_generator/components/misc.py
vector-of-bool, line 1: I assume that silk asset groups are no longer required with Silkbomb 2?

.evergreen/scripts/sbom.sh
```diff
@@ -0,0 +1,50 @@
...
+
+command -v podman >/dev/null || {
+  echo "missing required program podman" 1>&2
+  exit 1
+}
+
+command -v jq >/dev/null || {
+  echo "missing required program jq" 1>&2
+  exit 1
+}
```
vector-of-bool, line 19: Is `jq` actually required? It doesn't appear to be used.

.evergreen/scripts/sbom.sh
vector-of-bool, line 1: This could be moved to an `Earthfile` task, but would require another step to auth podman/Docker with Artifactory. It's fine as a standalone script since it's fairly trivial.

vector-of-bool (approved):
LGTM with minor tweaks. The failing tasks appear to be unrelated caused by flakiness in the docker hub gateway (sad).

.evergreen/config_generator/components/misc.py
eramongodb: Correct. The `--repo mongodb/mongo-c-driver` and `--branch "${branch_name}"` are now sufficient to associate uploaded/augmented SBOMs with a given project and branch. Asset group management is no longer necessary. 👏

.evergreen/scripts/sbom.sh
eramongodb: It is not (used for diffs in the C++ Driver, which are not present here).

eramongodb:
Added the `+sbom-generate-new-serial-number` Earthly target, which may be used to generate a new unique serial number and reset the SBOM version to `1` following a release via the new SilkBomb 2.0 `--generate-new-serial-number` flag.

docs/dev/earthly.rst
```diff
@@ -249,32 +211,15 @@ enumerated using ``earthly ls`` or ``earthly doc`` in the root of the repository
...
-      The repository branch for which to create the new asset group. If not
-      specified, the branch name will be inferred by asking Git.
-
-   .. rubric:: Secrets
-   .. envvar::
-         SILK_CLIENT_ID
-         SILK_CLIENT_SECRET
-      :noindex:
+   Equivalent to `+sbom-generate` but uses the `--generate-new-serial-number`
+   flag to generate a new unique serial number and reset the SBOM version to 1.
```
vector-of-bool, line 217: I think this docs line may fail, requiring two backticks around `--generate-new-serial-number`. Recommend testing docs changes with `make -C docs/dev serve`.

eramongodb: Good catch and thank you for the tip. Testing doc generation caught additional obsolete references to the old `+sbom-download` target. Inspected the generated HTMLs and updated accordingly.

docs/dev/deps.rst
```diff
@@ -100,36 +99,9 @@ augmented SBOM's generation. [#asbom-vulns]_
 
 The augmented SBOM is produced automatically and asynchronously as part of an
 external process that is not contained within the repository itself. The
```
kevinAlbs, line 101:
```suggestion
The augmented SBOM is produced as part of an
external process that is not contained within the repository itself. The
```

I expect producing the augmented SBOM is no longer asynchronous with the `silkbomb augment` command.

kevinAlbs (approved):

PR #1889, opened 2025-02-26
[CDRIVER-5916] Deprecate the Hedged-Reads APIs

vector-of-bool:
# Background 

Refer: CDRIVER-5916

DRIVERS-2929 is be a spec update that deprecates Hedge Reads in driver APIs. This changeset adds such deprecation warnings, among other things.

# Summary

This does not change any library behavior, it only adds deprecation warnings, and updates those that already exist. Per-commit review is recommended:

- c17b30f1 Adds new deprecation attribute macros that work on MSVC, to replace the `BSON_GNUC_DEPRECATED` macros. This macro always requires a deprecation message string, which will be included in the emitted attribute if it is supported by the compiler.
- ef8347bb Updates all existing deprecation attributes to use the new macros:
  - This also applies a formatting change and repositions the attributes for MSVC compatibility.
  - We also add several macros to `.clang-format`'s `AttributeMacros` to format them as declaration specifiers on functions.
  - Lots of headers have had whitespace inserted between function declarations, where it was becoming nigh unreadable.
  - Because deprecation warnings are now applied on MSVC, a new deprecation-warning-suppression had to be added to `bson-atomic.h` and `bson-cmp.h`
  - This commit is the largest change, and was largely a find-replace job.
- 0da23b06 Adds an internal declaration-like macro that emits the appropriate `_Pragma` to disable deprecation warnings.
- 4cc8a974 Adds deprecation annotations to "hedge" APIs, and updates the documentation to note the deprecation. Warning suppressions were added in internal locations where the hedged-read APIs are called.

## Note on Attribute Placement:

MSVC attributes are declaration-specifiers (thus "__declspec"), and thus need to be positioned correctly. The following is ill-formed:

```c
int*
BSON_DEPRECATED("oof")
foo();
```

because the `__declspec` is placed between the asterisk and the function name (not a valid position for declaration specifiers). GCC accepts this form, but we cannot use it for compatibility. One possible correct formatting for this declaration would be:

```c
int
BSON_DEPRECATED("oof")
* foo();
```

but that is a extremely weird, so it's easier to just place the deprecation before the return type:

```c
BSON_DEPRECATED("oof")
int* foo();
```

This has been applied throughout the codebase in ef8347bb.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -53,6 +53,9 @@ AlwaysBreakBeforeMultilineStrings: false
 AlwaysBreakTemplateDeclarations: MultiLine
 AttributeMacros:
   - __capability
+  - BSON_GNUC_WARN_UNUSED_RESULT
+  - BSON_DEPRECATED
+  - BSON_DEPRECATED_FOR
 BinPackArguments: false
 BinPackParameters: false
 BitFieldColonSpacing: Both
--- a/src/common/src/mlib/config.h
+++ b/src/common/src/mlib/config.h
@@ -292,6 +292,13 @@
 #define _mlibTestBuildConfig_MinSizeRel_MinSizeRel()
 
 /**
+ * @brief Emit a _Pragma that will disable warnings about the use of deprecated entities.
+ */
+#define mlib_disable_deprecation_warnings()                \
+   mlib_gnu_warning_disable ("-Wdeprecated-declarations"); \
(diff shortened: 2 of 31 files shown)


kevinAlbs (approved):
The MSVC deprecation warnings and additional deprecation warning strings are much appreciated. LGTM with comments addressed.

PR #1891, opened 2025-02-27
CDRIVER-5915: Fix for allocation of bson_t larger than half max size

ghost:
This PR addresses CDRIVER-5915, fixing three closely related problems:
* rounding allocation size to the next power of two could cause BSON_MAX_SIZE to be exceeded
* bson_reserve_buffer allocated more space than requested, equal to the previous document length
* test_bson_reserve_buffer_errors had a flawed "too big" case which was masked by the other two issues

This fixes allocation for bson_t that are exactly max size, and adds a test which exercises this case on 64-bit systems.

### DIFF
--- a/src/libbson/doc/bson_reserve_buffer.rst
+++ b/src/libbson/doc/bson_reserve_buffer.rst
@@ -24,6 +24,8 @@ Grow the internal buffer of ``bson`` to ``total_size`` and set the document leng
 
 First, initialize ``bson`` with :symbol:`bson_init` or :symbol:`bson_new`, then call this function. After it returns, the length of ``bson`` is set to ``total_size`` but its contents are uninitialized memory: you must fill the contents with a BSON document of the correct length before any other operations.
 
+The document must be freed with :symbol:`bson_destroy`.
+
 Note that, in this usage, the BSON header and footer bytes will not be verified or used by Libbson.
 The ``bson_t`` document length and buffer size limit are both set to ``total_size`` regardless of the value encoded in the document header.
 


src/libbson/src/bson/bson.c
```diff
@@ -2140,15 +2179,24 @@ bson_reserve_buffer (bson_t *bson, uint32_t size)
       return NULL;
    }
 
-   if (!_bson_grow (bson, size)) {
+   /* The caller wants a total document size of "size".
```
kevinAlbs, line 2182: Consider renaming the `size` parameters in `_bson_grow` and `bson_reserver_buffer` (e.g. to `grow_size` and `total_size`?) to help clarify the distinction.

src/libbson/src/bson/bson.c
```diff
@@ -143,14 +175,20 @@ _bson_impl_alloc_grow (bson_impl_alloc_t *impl, /* IN */
    /*
     * Determine how many bytes we need for this document in the buffer
     * including necessary trailing bytes for parent documents.
+    *
+    * Note that the buffer offset and nesting depth are not available
+    * outside bson_impl_alloc_t, meaning it's not possible for callers to
+    * fully rule out BSON_MAX_SIZE overflow before _bson_grow().
```
kevinAlbs, line 181: I expect a `bson_impl_inline_t` has an implicit zero `offset` and `depth`.

Does this comment mean a `bson_impl_inline_t` could call `_bson_grow()` and exceed `BSON_MAX_SIZE`?

kevinAlbs (approved):
Nice catch. Changes LGTM. Left a question and possible suggestion for a parameter rename.

src/libbson/src/bson/bson.c
```diff
@@ -60,6 +61,36 @@ typedef struct {
...
+ *
+ * _bson_round_up_alloc_size --
+ *
+ *       Given a potential allocation length in bytes, round up to the
+ *       next power of two without exceeding BSON_MAX_SIZE.
+ *
+ * Returns:
+ *       If the input is <= BSON_MAX_SIZE, returns a value >= the input
+ *       and still <= BSON_MAX_SIZE. If the input was greater than
+ *       BSON_MAX_SIZE, it is returned unmodified.
```
eramongodb, line 75: Perhaps this operation should be referred to as [clamping](https://en.wikipedia.org/wiki/Clamp_%28function%29) rather than rounding?

src/libbson/src/bson/bson.c
```diff
@@ -60,6 +61,36 @@ typedef struct {
...
+ * Side effects:
+ *       None.
+ *
+ *--------------------------------------------------------------------------
+ */
+
+static BSON_INLINE size_t
+_bson_round_up_alloc_size (size_t size)
+{
+   if (size <= BSON_MAX_SIZE) {
```
eramongodb, line 86: Given the precondition on `bson_next_power_of_two`:

> It is up to the caller to guarantee this will not overflow.

suggest adding a static assertion to document and assert the call to `bson_next_power_of_two` with `BSON_MAX_SIZE` is always valid, e.g.:

```c
// BSON_MAX_SIZE ^ 2 must be representable for computational purposes even if not a valid BSON size.
BSON_STATIC_ASSERT2 (bson_max_size_is_doubleable, (SIZE_MAX / 2u) >= BSON_MAX_SIZE);
```

Aside: the "32-bit unsigned integer of required bytes" description of the argument to `bson_next_power_of_two` seems to be outdated.

src/libbson/src/bson/bson.c
```diff
@@ -60,6 +61,36 @@ typedef struct {
...
+      size_t power_of_two = bson_next_power_of_two (size);
+      return BSON_MIN (power_of_two, BSON_MAX_SIZE);
+   } else {
+      return size;
```
eramongodb, line 90: Allowing `size > BSON_MAX_SIZE` to be returned unmodified feels like an odd special-case for this function. Can this be made a precondition for this function and the `size > BSON_MAX_SIZE` handling be moved into the call sites instead (where other bounds checks are already being performed)?

eramongodb: The call to `bson_next_power_of_two` in `bson_copy_to` is suspect given `bson_t` creation functions should be enforcing `len <= BSON_MAX_SIZE` via checks or assertions as an invariant of `bson_t` objects (+ why should a "copy" reserve more space than the value being copied from?), but I suppose this does not protect against manually-created `bson_t` objects which may have a buffer length greater than `BSON_MAX_SIZE`. Given this is falls outside the scope of this PR, I am okay with deferring this item.

src/libbson/src/bson/bson.c
eramongodb: I do not think `mcommon_next_power_of_two_u32` is equivalent, as `bson_next_power_of_two ` currently supports `sizeof(size_t) == sizeof(uint64_t)` with the `BSON_WORD_SIZE` special-case in its definition. This assertion is to statically assert (rather than runtime checks) the precondition for the call to `bson_next_power_of_two` which does not perform overflow safety checks.

src/libbson/src/bson/bson.c
```diff
@@ -2140,15 +2179,24 @@ bson_reserve_buffer (bson_t *bson, uint32_t size)
...
 
    if (bson->flags & BSON_FLAG_INLINE) {
       /* bson_grow didn't spill over */
       ((bson_impl_inline_t *) bson)->len = size;
+      BSON_ASSERT (size <= BSON_INLINE_DATA_SIZE);
    } else {
-      ((bson_impl_alloc_t *) bson)->len = size;
+      bson_impl_alloc_t *impl = (bson_impl_alloc_t *) bson;
+      impl->len = size;
+      BSON_ASSERT (impl->offset <= *impl->buflen && *impl->buflen - impl->offset >= (size_t) size);
```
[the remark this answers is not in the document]
eramongodb: Given this is an internal assertion rather than a test assertion, I am fine with leaving it as-written.

src/libbson/src/bson/bson.c
eramongodb: > Do we intend to provide a way to increase BSON_MAX_SIZE?

Likely no. That would be a major API breaking change. The static assertion (or any similar runtime checks on `BSON_MAX_SIZE`) is just to both document and protect against any such future modifications to `BSON_MAX_SIZE`. I believe it is fine to defer changes that extend forward compatibility for related functions to support larger `BSON_MAX_SIZE` than what it is currently set to (which is sufficiently small enough to be handled even by 32-bit `size_t`).

src/libbson/src/bson/bson.c
kevinAlbs: Thank you. The updated comment helped me understand.

kevinAlbs (approved):

eramongodb (approved):

PR #1892, opened 2025-02-28
sync RPM spec from downstream

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/67c2338323a40400075ec53d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Note that the task fails because the `master` branch requires the latest (not yet packaged) libmongocrypt. It will eventually turn green once the new libmongocrypt packages land.


eramongodb (approved):
Can/should the `BuildRequires: cmake(mongocrypt) >= 1.12.0` be updated by our spec.patch in advance to 1.13.0?

PR #1894, opened 2025-03-03
CDRIVER-5914: Proposed fix to address internal issue

ghost:
Full description in CDRIVER-5914.


kevinAlbs (approved):
The clearer tests are much appreciated. LGTM

PR #1899, opened 2025-03-06
CDRIVER-5641: Build fix for _FORTIFY_SOURCE

ghost:
This is a follow-up for #1868 to fix a build break in the abi-compliance-check task.

The break isn't straightforward to reproduce. It seems to require gcc-9(?), -D_FORTIFY_SOURCE=2 and -Og.

The cause is a set of tests that intentionally generate an oversized memcpy within dead code, to ensure the surrounding guard prevents the binary vector read/write operation. When _FORTIFY_SOURCE is set, additional checks are enabled for string functions like memcpy, enforcing a maximum object size of SIZE_MAX/2. In this patch, any time we build with _FORTIFY_SOURCE the maximum counts for these memcpy are adjusted so we don't overflow even for the largest element type (float32).

### DIFF
--- a/src/libbson/tests/test-bson-vector.c
+++ b/src/libbson/tests/test-bson-vector.c
@@ -1379,14 +1379,24 @@ test_bson_vector_example_packed_bit_view (void)
    } else                                                                         \
       ((void) 0)
 
+#if defined(__USE_FORTIFY_LEVEL) && __USE_FORTIFY_LEVEL > 0
+// Prevent memcpy size overflows even in dead code
+#define MAX_TESTABLE_COPY_COUNT (SIZE_MAX / 2u / sizeof (float))
+#else
+// Allow dead code to contain an oversized or overflowing memcpy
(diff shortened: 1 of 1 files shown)


eramongodb (approved):
Interesting find. It looks like the details involve use of [Object Size Checking Built-in Functions](https://gcc.gnu.org/onlinedocs/gcc-8.1.0/gcc/Object-Size-Checking.html) (e.g. for [memcpy](https://github.com/gcc-mirror/gcc/blob/releases/gcc-8.1.0/libssp/ssp/string.h#L57)) and selecting "a closest surrounding subobject" as "the object a pointer points to" when [`_FORTIFY_SOURCE > 1`](https://github.com/gcc-mirror/gcc/blob/406c2abec3f998e9064919b22db62f38a7c0e7b9/libssp/ssp/ssp.h.in#L51). However, I am unable to determine why this behavior may be restricted to specific GCC versions.

PR #1901, opened 2025-03-07
CDRIVER-5880: Remove workaround for DRIVERS-3095, defer other test improvements

ghost:
This includes the proposed spec change for DRIVERS-3095, and it removes the workarounds which are no longer needed due to the change. (Extended JSON infinities now pass through without special treatment.)

Spec PR:
https://github.com/mongodb/specifications/pull/1767

The C driver test code had noted several other opportunities to improve spec tests. These are no longer associated with any tickets, so I've updated the comments to be generic TODOs rather than referencing closed tickets.

### DIFF
--- a/src/libbson/tests/test-bson-vector.c
+++ b/src/libbson/tests/test-bson-vector.c
@@ -58,34 +58,11 @@ typedef struct vector_json_test_case_t {
    bool *test_valid;
 } vector_json_test_case_t;
 
-static void
-translate_json_test_vector (bson_t *array_in, bson_t *array_out)
-{
-   // Patches JSON tests until DRIVERS-3095 is resolved:
-   // - convert "inf" to a double +Infinity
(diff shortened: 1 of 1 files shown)


src/libbson/tests/test-bson-vector.c
```diff
@@ -357,6 +320,7 @@ test_bson_vector_json_case (vector_json_test_case_t *test_case)
          // understand it, they're operating on bytes rather than elements. This is the inverse of
          // append_vector_packed_bit_from_packed_array() above, and it bypasses the vector-to-array conversion.
          // 'array_from_vector' is ignored on this path.
+         // (Spec test improvement TODO)
```
eramongodb, line 323: Stray TODO comment? (Not accompanied by detail like the others are.)

eramongodb (approved):

PR #1903, opened 2025-03-07
CDRIVER-5743: NUL terminate string, handle more error cases

ghost:
In `mongoc_secure_channel_setup_ca()`, `pem_key` is used as a NUL terminated string even though it never has a NUL terminator. This patch fixes that by allocating an extra zeroed byte. While I was here, I also fixed a leaking `FILE` in the `fseek()`/`ftell()` error case, and I added a check for read errors.

I haven't been able to reproduce the crash locally yet, to verify that this is for sure the same issue. So far I don't have a memory dump to go on, just a backtrace similar to the one in CDRIVER-5743. (It would be useful to have more details about the fault. I'm assuming the address in the backtrace is the instruction causing the fault and not the data, due to the lack of page alignment.) Without a NUL byte, the `strstr()` will still terminate at the begin-certificate marker, but the `CryptStringToBinaryA()` will read past the end of the allocated buffer. Subsequent behavior depends on what memory is allocated adjacent to the `bson_malloc0()` above.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-secure-channel.c
+++ b/src/libmongoc/src/mongoc/mongoc-secure-channel.c
@@ -297,14 +297,14 @@ mongoc_secure_channel_setup_ca (mongoc_stream_tls_secure_channel_t *secure_chann
    fseek (file, 0, SEEK_END);
    length = ftell (file);
    fseek (file, 0, SEEK_SET);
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
Thank you for the fix. LGTM

PR #1905, opened 2025-03-10
CDRIVER-5924 test on Graviton

kevinAlbs:
# Summary
Test on Graviton with latest server

# Background & Motivation

Testing Graviton is requested in DRIVERS-2436. MongoDB servers on Graviton is expected to be [more common](https://docs.google.com/document/d/117jiFR_JH6tgNBC5QeoXRpTxJoV308M1Q-JxImN30mM/edit?tab=t.0). Additional testing appears motivated by server issues encountered on Graviton. [HELP-37322](https://jira.mongodb.org/browse/HELP-37322) references two behavior differences:
1. SERVER-69612: [`char` is unsigned](https://github.com/aws/aws-graviton-getting-started/blob/main/c-c%2B%2B.md#signed-vs-unsigned-char) on ARM.
2. SERVER-69611: gcc may enable a default for `-ffp-contract` (which differs from clang).

Neither appear specific to Graviton. (1) applies to other ARM CPUs. The C driver already tests on arm (via `ubuntu2004-arm64`). (2) appears non-applicable to the C driver. Quoting https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html:

> The default is -ffp-contract=off for C in a standards compliant mode

The C driver [specifies C99 by default](https://github.com/mongodb/mongo-c-driver/blob/915277362a01f7304a6e119fb7c2030f5401dca9/CMakeLists.txt#L406). So I expect the default C build agrees with `clang`. Regardless, I expect there are few floating point algorithms in the driver that would be affected.

Though the server supports arm on [4.2-latest](https://docs.google.com/spreadsheets/d/1-sZKW70HbVt2yHOWa18qwBwi-gBcn54tWmronnn89kI/), this PR proposes only testing the latest server.  The C driver already tests on ARM (via `ubuntu2004-arm64`). And a patch build testing [4.2-latest servers](https://spruce.mongodb.com/version/67cee246f1647400074b55b0/) showed no issues.

### DIFF
--- a/.evergreen/config_generator/components/sasl/openssl.py
+++ b/.evergreen/config_generator/components/sasl/openssl.py
@@ -26,6 +26,7 @@ COMPILE_MATRIX = [
     ('ubuntu2004-arm64',  'gcc',        None, ['cyrus']),
     ('ubuntu2004',        'gcc',        None, ['cyrus']),
     ('windows-vsCurrent', 'vs2017x64',  None, ['cyrus']),
+    ('amazon2023-arm64-latest-large-m8g', 'gcc', None, ['cyrus']),
 ]
 
 TEST_MATRIX = [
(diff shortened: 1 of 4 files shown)


vector-of-bool (approved):
LGTM

PR #1913, opened 2025-03-13
CDRIVER-2837 remove deprecated `bson_oid_init_sequence`

kevinAlbs:


vector-of-bool (approved):
LGTM. Failing tasks appear to be Artifactory timeouts.

PR #1925, opened 2025-03-17
Fix upload of ABI Compatibility Report HTML on failure

eramongodb:
The `compat_report.html` file is not being uploaded properly to AWS on task failure. This PR fixes the upload routine such that the task will have a "failed" status (via the curl command) while proceeding with subsequent `s3.put` operations. However, this curl command is temporarily disabled given the deliberate breaking changes targeting the 2.0.0 release. CDRIVER-5930 tracks re-enabling task failure after the 2.0.0 release.


kevinAlbs (approved):
Thank you for the fix. LGTM

PR #1926, opened 2025-03-17
CDRIVER-5701 remove references to removed symbols

kevinAlbs:
Follow-up to https://github.com/mongodb/mongo-c-driver/pull/1920:
- Remove references to removed symbols in documentation to fix `make-docs` task.
- Update C++ driver reference from r3.9.0 to r4.0.0. This includes CXX-2995 to removes use of `bson_as_json`. Fixes Earthly tasks.


vector-of-bool (approved):
LGTM

PR #1931, opened 2025-03-18
CDRIVER-5693 drop support for LibreSSL

kevinAlbs:
Tested with [this patch build](https://spruce.mongodb.com/version/67d880e355691e000720c05e/ ) run with `evergreen patch --project=mongo-c-driver --auto-description --yes --finalize -v all -t all`. Patch includes https://github.com/mongodb/mongo-c-driver/pull/1927, and has a broken format task fixed in this PR.


vector-of-bool (approved):
LGTM

PR #1933, opened 2025-03-18
CDRIVER-5514 drop support for Cyrus-SASL on Windows

kevinAlbs:
# Summary
- Warn if configuring with `ENABLE_SASL=CYRUS` on Windows
- Drop tests with Cyrus-SASL on Windows.
- Remove Windows-specific option `CYRUS_PLUGIN_PATH_PREFIX`.

# Background & Motivation
See CDRIVER-5696. Cyrus-SASL on Windows was untested and has little expected use.


NEWS
```diff
@@ -23,6 +23,8 @@ Unreleased (2.0.0)
 * `mongoc_cursor_is_alive` is removed. Use the equivalent `mongoc_cursor_more` instead.
 * `mongoc_collection_delete` is removed. Use `mongoc_collection_delete_one` or `mongoc_collection_delete_many` instead.
 * `mongoc_delete_flags_t` and `mongoc_reply_flags_t` are removed.
+* Support for `ENABLE_SASL=CYRUS` on Windows is removed. Use `ENABLE_SASL=SSPI` or `ENABLE_SASL=OFF` instead.
```
vector-of-bool, line 26: Is there a need to support disabling SASL on Windows if SSPI ships with Windows?

vector-of-bool (approved):

NEWS
kevinAlbs: I am not sure if there is a compelling use-case, but it could be used to avoid linking to unnecessary Windows DLLs (e.g. with `ENABLE_SASL=OFF` and `ENABLE_SSL=OFF`, I see SECURE32.DLL, CRYPT32.DLL, and BCRYPT.DLL are no longer shown in Dependency Walker.

PR #1935, opened 2025-03-18
Reduce Warnings - address -Wunused-macros warnings

eramongodb:
Removes all `-Wunused-macros` warnings locally observed with Clang 21 on Linux.


src/libbson/src/jsonsl/jsonsl.c
```diff
@@ -75,7 +75,6 @@ void jsonsl_dump_global_metrics(void)
 #else
 #define INCR_METRIC(m)
 #define INCR_GENERIC(c)
```
kevinAlbs, line 77: `INCR_STRINGY_CATCH` is also conditionally defined above. Consider removing the entire: `#ifdef JSONSL_USE_METRICS` block (I expect is used).

src/libmongoc/src/mongoc/mongoc-async-cmd.c
```diff
@@ -35,9 +35,6 @@
 #include <mongoc/mongoc-stream-tls.h>
 #endif
 
-#undef MONGOC_LOG_DOMAIN
-#define MONGOC_LOG_DOMAIN "async"
```
kevinAlbs, line 39: Suggest also removing the above unused `#include <mongoc/mongoc-log.h>`.

Defining `MONGOC_LOG_DOMAIN` before the include can override the logged domain. I initially thought the placement of this `#define` should instead be moved up, but this file does not appear to log.

kevinAlbs (approved):
LGTM with minor suggestions.

src/libbson/src/jsonsl/jsonsl.c
eramongodb: Removed all unnecessary and unused code in jsonsl conditioned on macros which we never define or use.

eramongodb:
Requesting another review in case the changes to jsonsl may be too drastic.

kevinAlbs (approved):
> Requesting another review in case the changes to jsonsl may be too drastic.

LGTM. I was mildly concerned about the jsonsl code going further out-of-sync with the upstream copy. But I expect it is unlikely jsonsl will be updated from upstream (project appears unmaintained).

PR #1938, opened 2025-03-19
CDRIVER-1330 remove automatic init/cleanup

kevinAlbs:
Removes deprecated behavior enabled by `ENABLE_AUTOMATIC_INIT_AND_CLEANUP`.

To help users identify a missing call to `mongoc_init` a check is added in the construction of `mongoc_client_t` and `mongoc_client_pool_t` to return an error if `mongoc_init` was not called. For example:
```c
// mongoc_init() was not called previously!
mongoc_client_t *client = mongoc_client_new_from_uri_with_error (uri, &error);
if (!client) {
    fprintf (stderr, "error creating client: %s\n", error.message);
    return EXIT_FAILURE;
}
```

Before the check, an assert fails:
```
./cmake-build/src/libmongoc/example-client       
/Users/kevin.albertson/code/tasks/mongo-c-driver-2.0/src/libmongoc/src/mongoc/mongoc-handshake.c:709 _mongoc_handshake_freeze(): assertion failed: pthread_mutex_lock (&(&gHandshakeLock)->wrapped_mutex) == 0
```

With the check, this becomes:
```
error creating client: Attempting to create client, but libmongoc not initialized. Call mongoc_init
```


vector-of-bool (approved):
Yay!

PR #1939, opened 2025-03-19
 CDRIVER-4489 promote empty username to client error 

eramongodb:
Followup to https://github.com/mongodb/mongo-c-driver/pull/1896. Enforces that even optional usernames are non-empty when specified (there are no cases where an empty username is desirable).

### DIFF
--- a/src/libmongoc/tests/test-mongoc-uri.c
+++ b/src/libmongoc/tests/test-mongoc-uri.c
@@ -704,9 +704,11 @@ test_mongoc_uri_auth_mechanism_mongodb_x509 (void)
          mongoc_uri_t *const uri =
             mongoc_uri_new_with_error ("mongodb://@localhost/?" MONGOC_URI_AUTHMECHANISM "=MONGODB-X509", &error);
          ASSERT_NO_CAPTURED_LOGS ("mongoc_uri_new_with_error"); // CDRIVER-1959
-         ASSERT_OR_PRINT (uri, error);
-         ASSERT_CMPSTR (mongoc_uri_get_username (uri), "");
-         mongoc_uri_destroy (uri);
+         ASSERT (!uri);
+         ASSERT_ERROR_CONTAINS (error,
+                                MONGOC_ERROR_COMMAND,
+                                MONGOC_ERROR_COMMAND_INVALID_ARG,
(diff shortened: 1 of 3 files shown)


src/libmongoc/tests/test-mongoc-uri.c
```diff
@@ -313,6 +313,59 @@ _auth_mechanism_username_required (const char *mechanism)
    }
 }
 
+static void
+_auth_mechanism_password_prohibited (const char *mechanism, const char *user_prefix, const char *uri_suffix)
+{
+   BSON_ASSERT_PARAM (mechanism);
+   BSON_ASSERT_PARAM (user_prefix);
+
+   bson_error_t error;
```
kevinAlbs, line 322: Move declaration inside scope block to fix warnings on Evergreen tasks:

```
../../../src/libmongoc/tests/test-mongoc-uri.c: In function '_auth_mechanism_password_prohibited':
../../../src/libmongoc/tests/test-mongoc-uri.c:339:20: error: declaration of 'error' shadows a previous local [-Werror=shadow]
  339 |       bson_error_t error;
      |                    ^~~~~
```

src/libmongoc/tests/test-mongoc-uri.c
```diff
@@ -313,6 +313,59 @@ _auth_mechanism_username_required (const char *mechanism)
...
+
+   // None.
+   {
+      mongoc_uri_t *const uri = mongoc_uri_new_with_error (
+         tmp_str ("mongodb://localhost/?" MONGOC_URI_AUTHMECHANISM "=%s%s", mechanism, uri_suffix), &error);
```
kevinAlbs, line 327:
```suggestion
      mongoc_uri_t *const uri = mongoc_uri_new_with_error (
         tmp_str ("mongodb://%s@localhost/?" MONGOC_URI_AUTHMECHANISM "=%s%s", user_prefix, mechanism, uri_suffix),
         &error);
```

Suggest including `user_prefix`. IIUC this helper only intends to test changes in the password and keep the rest of the URI fixed.

If accepted, also update or remove the `ASSERT_CMPSTR (mongoc_uri_get_username (uri), NULL);` below.

kevinAlbs (approved):
LGTM with minor comments addressed.

PR #1941, opened 2025-03-19
CDRIVER-4489 update URI documentation for authMechanismProperties

eramongodb:
Followup to https://github.com/mongodb/mongo-c-driver/pull/1896 which updates API documentation for relevant URI options. Additionally deprecates old GSSAPI URI options in favor of using `authMechanismProperties` instead.

This PR also preemptively adds [the warning](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#mongocredential-properties-6) required by the Authentication spec:

> Note: because the `TOKEN_RESOURCE` is often itself a URL, drivers MUST document that a `TOKEN_RESOURCE` with a comma `,` must be given as a `MongoClient` configuration and not as part of the connection string, and that the `TOKEN_RESOURCE` value can contain a colon `:` character.

The wording is adjusted to indicate that mongoc will interpret the presence of a comma as a key-value pair delimiter. More importantly, in contrast to the spec and associated spec tests, _percent-encoded_ commas will be permitted as an extension to improve user experience, as they do not interfere with the parsing of `authMechanismProperties` or the connection string as implied by the spec. This requires upcoming changes which are separate from this PR.

Aside: ReStructuredText is painpainpain. 🫠

### DIFF
--- a/src/libmongoc/doc/mongoc_uri_t.rst
+++ b/src/libmongoc/doc/mongoc_uri_t.rst
@@ -122,9 +122,55 @@ Authentication Options
 Constant                                   Key                               Description
 ========================================== ================================= =========================================================================================================================================================================================================================
 MONGOC_URI_AUTHMECHANISM                   authmechanism                     Specifies the mechanism to use when authenticating as the provided user. See `Authentication <authentication_>`_ for supported values.
-MONGOC_URI_AUTHMECHANISMPROPERTIES         authmechanismproperties           Certain authentication mechanisms have additional options that can be configured. These options should be provided as comma separated option_key:option_value pair and provided as authMechanismProperties. Specifying the same option_key multiple times has undefined behavior.
-MONGOC_URI_AUTHSOURCE                      authsource                        The authSource defines the database that should be used to authenticate to. It is unnecessary to provide this option the database name is the same as the database used in the URI.
+MONGOC_URI_AUTHMECHANISMPROPERTIES         authmechanismproperties           Additional properties for the specified mechanism using key-value pair format, e.g. ``key1:value1,key2:value2``.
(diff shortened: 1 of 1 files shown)


src/libmongoc/doc/mongoc_uri_t.rst
```diff
@@ -122,10 +122,56 @@ Authentication Options
...
+SERVICE_HOST           Optional. May be needed to use a service host that differs from the initial role.
+====================== ===================================================================================================================================================================================================================================
+
+MONGODB-AWS
+^^^^^^^^^^^
+
+=========== ====================================================================================
+Key         Value
+=========== ====================================================================================
+MONGODB-AWS Optional. An AWS session token to use for authentication with temporary credentials.
```
kevinAlbs, line 172: Change to `AWS_SESSION_TOKEN`.

src/libmongoc/doc/mongoc_uri_t.rst
```diff
@@ -122,20 +122,67 @@ Authentication Options
...
+
+Deprecated Mechanism Property Options
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following options have been deprecated and may be removed from future releases of libmongoc.
 
+========================================== ================================= ==================================================================== =======================
+Constant                                   Key                               Deprecated For                                                       Key
+========================================== ================================= ==================================================================== =======================
+MONGOC_URI_CANONICALIZEHOSTNAME            canonicalizehostname              MONGOC_URI_AUTHMECHANISMPROPERTIES (CANONICALIZE_HOST_NAME)          authmechanismproperties
```
kevinAlbs, line 183: Suggest adding a `MONGOC_WARNING` in `mongoc_uri_apply_options` to signal these options deprecated (similar to how unsupported URI options warn). That may help to inform users.

kevinAlbs (approved):

PR #1942, opened 2025-03-20
CDRIVER-2813 remove `BSON_EXTRA_ALIGNMENT` option

kevinAlbs:
# Summary
- Remove CMake option `ENABLE_EXTRA_ALIGNMENT`.
- Align `bson_t` and `bson_iter_t` to the size of a pointer.

Verified with this patch build: https://spruce.mongodb.com/version/67dc0db1caab950007b0f764

# Motivation

Addresses long-standing ask to remove default over-alignment: CDRIVER-2813.

Removing alignment specifiers entirely was considered. However, removing the alignment specifier on `bson_t` resulted in a runtime error in calls to `BSON_ALIGNED_ALLOC (bson_t)`:

> Failure to allocate memory in bson_aligned_alloc()

I expect this is due to `bson_t` having alignment of 4. `bson_t` does not contain a pointer.

From `man aligned_alloc` (emphasis mine):
> In addition, aligned_alloc() returns a NULL pointer and sets errno to EINVAL if size is not an integral multiple of alignment, or if alignment is not a power of 2 **at least as large as sizeof(void *)**.

To limit the risk of causing runtime errors in user upgrades, this PR proposes keeping existing alignment specifiers. Only the over-alignment of `bson_t` and `bson_iter_t` is changed. 

## Alignment changes

The following were printed on my 64-bit machine to compare alignment changes.

Prior to this PR. With `ENABLE_EXTRA_ALIGNMENT=ON`:
```
alignof(bson_t):           128
alignof(bson_iter_t):      128
alignof(bson_visitor_t):   8
alignof(bson_reader_t):    8
alignof(bson_value_t):     8
alignof(bson_error_t):     8
```

With all alignment specifiers removed (not done in this PR):
```
alignof(bson_t):           4
alignof(bson_iter_t):      8
alignof(bson_visitor_t):   8
alignof(bson_reader_t):    4
alignof(bson_value_t):     8
alignof(bson_error_t):     4
```

With this PR:
```
alignof(bson_t):           8
alignof(bson_iter_t):      8
alignof(bson_visitor_t):   8
alignof(bson_reader_t):    8
alignof(bson_value_t):     8
alignof(bson_error_t):     8
```

### DIFF
--- a/src/libbson/NEWS
+++ b/src/libbson/NEWS
@@ -14,6 +14,7 @@ libbson 2.0.0 (Unreleased)
 * Remove deprecated atomic and threading interfaces: `bson_atomic_*` and `bson_thrd_yield`.
 * The deprecated `ENABLE_EXTRA_ALIGNMENT` CMake option is removed.
   * `bson_t` and `bson_iter_t` are now aligned to the size of a pointer instead of `128`.
+  * `bson_error_t` is now aligned to the size of a pointer instead of `8`.
   * `BSON_ALIGNED_BEGIN` and `BSON_ALIGNED_END` now unconditionally apply their requested alignment.
 
 
--- a/src/libbson/NEWS
+++ b/src/libbson/NEWS
@@ -13,7 +13,7 @@ libbson 2.0.0 (Unreleased)
 * Remove deprecated integral comparison interfaces: `bson_in_range_*` and `bson_cmp_*`.
 * Remove deprecated atomic and threading interfaces: `bson_atomic_*` and `bson_thrd_yield`.
 * The deprecated `ENABLE_EXTRA_ALIGNMENT` CMake option is removed.
-  * `bson_t` and `bson_iter_t` are now aligned to the size of a pointer.
+  * `bson_t` and `bson_iter_t` are now aligned to the size of a pointer instead of `128`.
   * `BSON_ALIGNED_BEGIN` and `BSON_ALIGNED_END` now unconditionally apply their requested alignment.
 
 
--- a/src/libbson/NEWS
+++ b/src/libbson/NEWS
@@ -12,6 +12,10 @@ libbson 2.0.0 (Unreleased)
 * Compiling with `BSON_MEMCHECK` defined now has no effect.
 * Remove deprecated integral comparison interfaces: `bson_in_range_*` and `bson_cmp_*`.
 * Remove deprecated atomic and threading interfaces: `bson_atomic_*` and `bson_thrd_yield`.
+* The deprecated `ENABLE_EXTRA_ALIGNMENT` CMake option is removed.
+  * `bson_t` and `bson_iter_t` are now aligned to the size of a pointer.
+  * `BSON_ALIGNED_BEGIN` and `BSON_ALIGNED_END` now unconditionally apply their requested alignment.
+
 
 libbson 1.30.2
 ==============
--- a/src/libbson/NEWS
+++ b/src/libbson/NEWS
@@ -14,7 +14,7 @@ libbson 2.0.0 (Unreleased)
 * Remove deprecated atomic and threading interfaces: `bson_atomic_*` and `bson_thrd_yield`.
 * The deprecated `ENABLE_EXTRA_ALIGNMENT` CMake option is removed.
   * `bson_t` and `bson_iter_t` are now aligned to the size of a pointer instead of `128`.
-  * `bson_error_t` and `bson_value_t` are now aligned to the size of a pointer instead of `8`.
+  * `bson_error_t`, `bson_value_t`, and `bson_visitor_t` are now aligned to the size of a pointer instead of `8`.
   * `BSON_ALIGNED_BEGIN` and `BSON_ALIGNED_END` now unconditionally apply their requested alignment.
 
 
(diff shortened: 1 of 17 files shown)


src/libbson/doc/bson_t.rst
```diff
@@ -127,12 +127,11 @@ Synopsis
   #define BSON_APPEND_VALUE(b, key, val) \
      bson_append_value (b, key, (int) strlen (key), (val))
 
-  BSON_ALIGNED_BEGIN (128)
   typedef struct {
      uint32_t flags;       /* Internal flags for the bson_t. */
      uint32_t len;         /* Length of BSON data. */
      uint8_t padding[120]; /* Padding for stack allocation. */
-  } bson_t BSON_ALIGNED_END (128);
+  } bson_t;
```
eramongodb, line 134:
```suggestion
  } bson_t BSON_ALIGNED_END (BSON_ALIGN_OF_PTR);
```

Preserve documentation of (updated) alignment requirements.

src/libbson/NEWS
```diff
@@ -12,6 +12,10 @@ libbson 2.0.0 (Unreleased)
 * Compiling with `BSON_MEMCHECK` defined now has no effect.
 * Remove deprecated integral comparison interfaces: `bson_in_range_*` and `bson_cmp_*`.
 * Remove deprecated atomic and threading interfaces: `bson_atomic_*` and `bson_thrd_yield`.
+* The deprecated `ENABLE_EXTRA_ALIGNMENT` CMake option is removed.
+  * `bson_t` and `bson_iter_t` are now aligned to the size of a pointer.
```
eramongodb, line 16:
```suggestion
  * `bson_t` and `bson_iter_t` are now aligned to the size of a pointer instead of `128`.
```

src/libbson/src/bson/bson-types.h
```diff
@@ -120,11 +120,11 @@ typedef struct _bson_json_opts_t bson_json_opts_t;
  *
  * This structure is meant to fit in two sequential 64-byte cachelines.
  */
-BSON_ALIGNED_BEGIN (128) typedef struct _bson_t {
+BSON_ALIGNED_BEGIN (BSON_ALIGN_OF_PTR) typedef struct _bson_t {
```
vector-of-bool, line 123: Could we instead fix `bson_aligned_alloc` to increase the alignment size if it is below pointer alignment?

src/libbson/src/bson/bson-types.h
eramongodb, line 1: The following types do not require alignment specifiers:

- `bson_impl_alloc_t` (contains pointers)
- `bson_error_t` (no pointers, no polymorphism)
- `bson_iter_t` (contains pointers)
- `bson_value_t` (contains pointers)
- `bson_visitor_t` (contains pointers)

I believe only `bson_t`, `bson_impl_inline_t`, and `bson_reader_t` require alignment specifiers (they are polymorphic without a pointer in the base type).

eramongodb (changes requested):
Suggest also removing the `debug-compile-no-align` tasks from the legacy config.

src/libbson/src/bson/bson-types.h
kevinAlbs: I am concerned removing the alignment specifier could break existing code allocating without libbson functions. Example:
```
bson_t *b = aligned_alloc (alignof(bson_t), sizeof(bson_t));
```
Upgrading to 2.0 may make this a runtime error. Maybe (hopefully) it is unlikely. I see no usage on GitHub code search [for `align.*bson_t`](https://github.com/search?q=%2Falign.*bson_t%2F+path%3A*.c+-repo%3Amongodb%2Fmongo-c-driver+NOT+is%3Afork+NOT+path%3A%2Fmongoc%5C-.*%5C.c%2F+NOT+path%3A%2Fbson%5C-.*%5C.c%2F+NOT+path%3Abson.c++NOT+path%3A%2Fmongocrypt%5C-.*%5C.c%2F+NOT+path%3A%2F.*test.*%2F&type=code&ref=advsearch). Though there is [use of `malloc.*bson_t`](https://github.com/search?q=%2Fmalloc.*bson_t%2F+path%3A*.c+-repo%3Amongodb%2Fmongo-c-driver+NOT+is%3Afork+NOT+path%3A%2Fmongoc%5C-.*%5C.c%2F+NOT+path%3A%2Fbson%5C-.*%5C.c%2F+NOT+path%3Abson.c++NOT+path%3A%2Fmongocrypt%5C-.*%5C.c%2F+NOT+path%3A%2F.*test.*%2F&type=code&ref=advsearch).

AFAIK types being aligned to a pointer size (or 8) is less of an issue. If there is little observed benefit, it might not be worth the risk (even if small).

eramongodb: Polymorphic types such as `bson_t` and `bson_reader_t` require a pointer alignment specifier in their "base" type due to pointer data members in their "derived" types (`bson_impl_alloc_t` and `bson_reader_handle_t` respectively). The alignment specifier is probably preferable to awkwardly inserting a stub pointer data member in the "base" type.

eramongodb (approved):

vector-of-bool (approved):
LGTM

eramongodb:
Note: partially reverting https://github.com/mongodb/mongo-cxx-driver/pull/1329 (reverting only the addition and use of `bsoncxx::aligned_storage<T>`) reveals a single benign warning when compiled with MSVC and `/W4` (analogous to `-Wpadded` warnings for GCC and Clang):

```
warning C4324: 'bsoncxx::v_noabi::builder::core::impl::frame': structure was padded due to alignment specifier
```

Runtime alignment errors are not observed when running C++ Driver UBSan tasks with the partial revert using the changes in this PR branch. 👍

PR #1945, opened 2025-03-21
CDRIVER-4689 add mongoc_oidc_callback_t

eramongodb:
Implements the abstraction of an OIDC Callback as required for CDRIVER-4689.

This PR does not yet implement _usage_ of the OIDC Callback by `mongoc_client_t` or `mongo_client_pool_t` as is documented by the proposed API doc examples (`mongoc_client_set_oidc_callback` and `mongoc_client_pool_set_oidc_callback` do not exist).

---

This PR implements an abstraction for the [OIDC Callback](https://github.com/mongodb/specifications/blob/78916887729bf0d3088541bd9a7390770fae5f77/source/auth/auth.md#oidc-callback) feature:

> Drivers MUST allow users to provide a callback that returns an OIDC access token. The purpose of the callback is to allow users to integrate with OIDC providers not supported by the Built-in Provider Integrations.

The spec requires the abstraction is capable of extending parameters and return values in a backward compatible manner:

> The signature and naming of the callback API is up to the driver's discretion. Drivers MUST ensure that additional optional input parameters and return values can be added to the callback signature in the future without breaking backward compatibility.

Therefore, the parameters (`mongoc_oidc_callback_params_t`) and the return values (`mongoc_oidc_credential_t`) are opaque types. All "parameters" and "return values" must be accessed using functions to ensure the addition of new parameters and return types will not break API or ABI compatibility. Therefore, the callback function as implemented by a user must be declared as follows:

```c
mongoc_oidc_credential_t *
example_oidc_callback_fn (mongoc_oidc_callback_params_t *params);
```

The callback function type is specified as `mongoc_oidc_callback_fn_t`.

---

The Authentication spec requires:

> Drivers MUST provide a way for the callback to be either automatically canceled, or to cancel itself. This can be as a timeout argument to the callback, a cancellation context passed to the callback, or some other language-appropriate mechanism.

To distinguish between success, cancellation due to a timeout, and cancellation due to an error, the callback function is expected to return with one of three conditions, expressed as pseudocode (this would be implemented within the handshake algorithm):

```python
cred = callback.fn (params) # mongoc_oidc_callback_fn_t

if cred:
    use(cred.access_token, cred.expires_in) # Success.
else if params.cancel_with_timeout:
    handle_timeout() # CSOT Timeout.
else:
    handle_error() # Authentication failure.
```

Although the current implementation proposes ignoring `cancel_with_timeout` when `cancel_with_error` is set (prioritize error over timeout), this can be changed into a Client API usage error if preferable. It would admittedly be unusual for both to ever be set at the same time.

> [!NOTE]
> `cancel_with_timeout` is implemented as an out parameter rather than as return value to avoid requiring users to create and return a `mongoc_oidc_credential_t` object during cancellation. This should help users better distinguish success vs. cancellation.

---

The proposed API deliberately avoids allowing the user to specify any details regarding the error or timeout. Note the absence of a `bson_error_t` parameter or even support for a `const char*` error message string:

```c
if (...) {
  // No arguments other than `params`.
  return mongoc_oidc_callback_params_cancel_with_timeout (params);
}

if (...) {
  // No arguments at all.
  return NULL;
}
```

This is to avoid locking the API into supporting `bson_error_t` codes, error messages, and/or any other details that are irrelevant to satisfying the minimal requirements of the OIDC Callback API. Instead, support for any additional details is completely deferred to the `user_data` in/out parameter, whose behavior may be defined entirely by the user as they require, as shown in the API example code:

```c
typedef struct {
   const char *error_message;
} user_data_t;

mongoc_oidc_credential_t *
example_callback_fn (mongoc_oidc_callback_params_t *params)
{
   user_data_t *user_data = mongoc_oidc_callback_params_get_user_data (params);

   // This callback function was implemented for OIDC callback API version 1.
   if (mongoc_oidc_callback_params_get_version (params) > 1) {
      user_data->error_message = "OIDC callback API has changed: update example_callback_fn!";
      return NULL;
   }

   // ...
}
```

Therefore, the mongoc library will not need to handle propagation of user-provided error details. It can focus simply on implementing the MONGODB-OIDC authentication's mechanism algorithmic requirements only.

> [!NOTE]
> The `cancel_with_*` functions always return `NULL` to support the `return` syntax above. This is primarily for convenience and to help avoid confusion with return value null vs. not-null requirements. However, users may choose to manually return `NULL` instead.

---

The `timeout` and `expires_in` parameter types are `int64_t`. This is to be consistent with `bson_get_monotonic_time()`, `expiration_ms_to_timer()`, `mongoc_async_cmd_t::timeout_ms`, `timeout_msec` parameters, etc. and to avoid signedness inconsistency due to `uint64_t` and `uint32_t`, avoid narrowing conversions due to `int32_t`.

The `timeout` parameter (when set) is meant to be directly compared against `bson_get_monotonic_time()` by the user to determine when to report a timeout, as shown in API example code.

The `expires_in` parameter is meant to be used after the callback function has (successfully) returned to compute the timestamp when the associated access token will expire (in the cache) and the OIDC callback function must be re-invoked. This will likely be computed as `bson_get_monotonic_time() + to_microseconds (*expires_in)` (internally represented as an `mcd_time_point`). Details are TBD and will depend on the implementation of Credential Caching and the overall MONGODB-OIDC authentication algorithm.


src/libmongoc/doc/mongoc_oidc_callback_fn_t.rst
```diff
@@ -0,0 +1,39 @@
...
+
+A :symbol:`mongoc_oidc_credential_t` object created with :symbol:`mongoc_oidc_credential_new()`, or ``NULL`` to indicate cancellation.
+
+* The function MUST return a :symbol:`mongoc_oidc_credential_t` object to indicate success.
+* The function MUST call :symbol:`mongoc_oidc_callback_params_cancel_with_timeout()` and return ``NULL`` to indicate a timeout.
+* The function MUST call :symbol:`mongoc_oidc_callback_params_cancel_with_error()` and return ``NULL`` to indicate an error.
+
+The ``cancel_with_timeout`` and ``cancel_with_error`` out parameters are ignored if the return value is not ``NULL``.
+
+A ``NULL`` return value without setting either ``cancel_with_timeout`` or ``cancel_with_error`` will be interpreted as a client error.
```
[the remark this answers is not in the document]
eramongodb: The `cancel_with_error` function was included for symmetry with `cancel_with_timeout`, e.g.:

```c
if (...) {
  return cancel_with_timeout (params); // NULL
}

if (...) {
  return cancel_with_error (params); // NULL
}
```

but I can remove it if avoiding the redundancy is preferable.

src/libmongoc/doc/mongoc_oidc_callback_fn_t.rst
```diff
@@ -0,0 +1,35 @@
...
+
+Parameters
+----------
+
+* ``params``: A :symbol:`mongoc_oidc_callback_params_t` object.
+
+Returns
+-------
+
+A :symbol:`mongoc_oidc_credential_t` object created with :symbol:`mongoc_oidc_credential_new()`, or ``NULL`` to indicate cancellation.
```
kevinAlbs, line 23:
```suggestion
A :symbol:`mongoc_oidc_credential_t` object created with :symbol:`mongoc_oidc_credential_new()`, or ``NULL`` to indicate cancellation or error.
```

src/libmongoc/doc/mongoc_oidc_callback_params_t.rst
```diff
@@ -0,0 +1,215 @@
...
+       const int64_t timeout = mongoc_oidc_callback_params_get_timeout (params);
+
+       if (bson_get_monotonic_time () > timeout) {
+          return mongoc_oidc_callback_params_cancel_with_timeout (params);
+       }
+
+       // ...
+    }
+
+Error Cancellation
```
kevinAlbs, line 184: Suggest avoiding "cancellation". I expect an error could indicate many things (failure to connect to OIDC server, error in credentials, etc.). I interpret "cancellation" to mean something external requested cancellation.

src/libmongoc/doc/mongoc_oidc_callback_fn_t.rst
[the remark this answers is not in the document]
kevinAlbs: I prefer avoiding the redundancy. My initial read was are errors indicated with `cancel_with_error` treated specially by the driver (e.g. clear the cached token?).

kevinAlbs (approved):
LGTM. I expect the newly added API will not be ready before the upcoming 2.0 release. I think keep this PR open until after the 2.0 release.

eramongodb:
On further thought, given the removal of the `cancel_with_error` out parameter, decided to move the `access_token` and optional `expires_in` return values into the constructors for `mongoc_oidc_credential_t`. This is to ensure validity of the provided argument(s) (i.e. `access_token` must not be `NULL`) and avoid potential user confusion between a "null return value" (to indicate an error or timeout) vs. "a return value with a null access token" (forbidden).

src/libmongoc/doc/mongoc_oidc_credential_new_with_expires_in.rst
```diff
@@ -0,0 +1,38 @@
+:man_page: mongoc_oidc_credential_new
+
+mongoc_oidc_credential_new()
+============================
```
kevinAlbs, line 4:
```suggestion
mongoc_oidc_credential_new_with_expiry()
========================================
```

kevinAlbs (approved):

src/libmongoc/doc/mongoc_oidc_credential_new_with_expires_in.rst
eramongodb: Interesting, Sphinx does not appear to emit a warning for this mistake.

src/libmongoc/src/mongoc/mongoc-oidc-callback.c
```diff
@@ -0,0 +1,248 @@
+/*
+ * Copyright 2009-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License") {}
```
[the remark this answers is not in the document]
eramongodb: Whoops, good catch.

src/libmongoc/doc/mongoc_oidc_callback_t.rst
```diff
@@ -0,0 +1,155 @@
...
+
+    void
+    with_many_clients (void)
+    {
+       mongoc_client_t *client_a = /* ... */;
+       mongoc_client_t *client_b = /* ... */;
+
+       {
+          mongoc_oidc_callback_t *callback = mongoc_oidc_callback_new ();
+          mongoc_oidc_callback_set_fn (callback, &many_threads_possible);
```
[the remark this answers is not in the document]
eramongodb: Oh, I understand now. Thank you for the explanation. Tweaked the wording and examples to emphasize the use of single-threaded clients _in multiple threads_ to exercise the thread-safe callback requirement.

src/libmongoc/tests/test-mongoc-oidc-callback.c
```diff
@@ -0,0 +1,167 @@
...
+   ASSERT (!mongoc_oidc_callback_params_get_cancelled_with_timeout (params));
+
+   // Input parameters.
+   {
+      // Normal values.
+      {
+         mongoc_oidc_callback_params_set_timeout (params, 123);
+         {
+            const char username[] = "username";
+            mongoc_oidc_callback_params_set_username (params, username); // Ensure a copy is made.
```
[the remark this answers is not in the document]
eramongodb: Makes sense. Updated the test case to modify the copied-from object to ensure it is not observed later (no reference semantics).

src/libmongoc/doc/mongoc_oidc_callback_params_t.rst
```diff
@@ -0,0 +1,215 @@
...
+    mongoc_oidc_credential_t *
+    example_callback_fn (mongoc_oidc_callback_params_t *params)
+    {
+       user_data_t *user_data = mongoc_oidc_callback_params_get_user_data (params);
+
+       // This callback function was implemented for OIDC callback API version 1.
+       if (mongoc_oidc_callback_params_get_version (params) > 1) {
+          user_data->error_message = "OIDC callback API has changed: update example_callback_fn!";
+          return NULL;
+       }
```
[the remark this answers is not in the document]
eramongodb: > Would the example make more sense reporting a non-fatal warning?

Adjusted the example to one which merely alerts the user (via `printf`) that new features are available. This should help avoid implying a version bump must be treated as an error.

src/libmongoc/src/mongoc/mongoc-oidc-callback.c
```diff
@@ -0,0 +1,257 @@
...
+   int64_t timeout; // Guarded by timeout_is_set.
+   int32_t version;
+   bool cancelled_with_timeout;
+   bool timeout_is_set;
+};
+
+struct _mongoc_oidc_credential_t {
+   char *access_token;
+   int64_t expires_in; // Guarded by expires_in_set.
+   bool expires_in_set;
```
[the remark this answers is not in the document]
eramongodb: Good point. This was due to initially treating `0` as an instant (rather than infinite) expiry duration while also conforming to the Auth spec, which states:

> Drivers with optional parameters MAY interpret a missing value as infinite. 

Since both unset and 0 are now equivalent, `expires_in_set` is no longer necessary. This should also help avoid the `const int64_t *` return type situation discussed earlier. Removed `expires_in_set`.

src/libmongoc/src/mongoc/mongoc-oidc-callback.h
```diff
@@ -0,0 +1,86 @@
...
+MONGOC_EXPORT (mongoc_oidc_callback_t *)
+mongoc_oidc_callback_new_with_user_data (mongoc_oidc_callback_fn_t fn, void *user_data);
+
+MONGOC_EXPORT (void)
+mongoc_oidc_callback_destroy (mongoc_oidc_callback_t *callback);
+
+MONGOC_EXPORT (mongoc_oidc_callback_fn_t)
+mongoc_oidc_callback_get_fn (const mongoc_oidc_callback_t *callback);
+
+MONGOC_EXPORT (void *)
```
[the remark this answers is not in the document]
eramongodb: Good call. I had not considered C++ code.

eramongodb:
@kevinAlbs Re-requesting review due to significant changes, in summary:

- Required (not-null) arguments are now enforced by constructors (`*_new()`) which return `NULL` for invalid arguments.
	- `mongoc_oidc_callback_t::fn` (removed `set_fn()`).
	- `mongoc_oidc_credential_t::access_token` (`set_access_token()` was already removed).
- `mongoc_oidc_credential_t::expires_in` is no longer optional; instead, a value of `0` (infinite expiry duration) is the default. Its validity is now enforced by constructors alongside `access_token`.
- Doc pages are restructured by "Functions" toctrees in each `*_t` type doc page. "Lifecycle" sections are also added.
- The API version check example is adjusted to print a message rather than error to avoid implying a version bump should be treated as an error condition by the user.

src/libmongoc/doc/mongoc_oidc_callback_params_t.rst
```diff
@@ -0,0 +1,215 @@
...
+
+          if (error.code != 0) {
+             printf ("client error message: %s\n", error.message);
+          }
+
+          if (user_data->error_message) {
+             printf ("custom error message: %s\n", user_data->error_message);
+          }
+
+          printf ("The callback function was invoked %n times!", user_data->counter);
```
kevinAlbs, line 146:
```suggestion
          printf ("The callback function was invoked %d times!", user_data->counter);
```

src/libmongoc/doc/mongoc_oidc_callback_params_t.rst
```diff
@@ -0,0 +1,215 @@
...
+    void
+    example (void)
+    {
+       mongoc_client_t *client = /* ... */;
+       bson_error_t error;
+
+       {
+          user_data_t *user_data = malloc (sizeof (*user_data));
+          *user_data = (user_data_t){.counter = 0, .error_message = NULL};
+          mongoc_oidc_callback_t *callback = mongoc_oidc_callback_new (&example_callback_fn, (void *) user_data);
```
kevinAlbs, line 127:
```suggestion
          mongoc_oidc_callback_t *callback = mongoc_oidc_callback_new_with_user_data (&example_callback_fn, (void *) user_data);
```

src/libmongoc/doc/mongoc_oidc_credential_new_with_expires_in.rst
```diff
@@ -0,0 +1,39 @@
...
+  mongoc_oidc_credential_t *
+  mongoc_oidc_credential_new_with_expires_in (const char *access_token, int64_t expires_in)
+
+Create a new :symbol:`mongoc_oidc_credential_t` object which stores a copy of the provided OIDC access token and its expiry duration (in milliseconds).
+
+The expiry duration will be evaluated relative to the value returned by :symbol:`bson_get_monotonic_time()` immediately after the callback function has returned.
+
+.. important::
+
+    A value of ``0`` is interpreted as an infinite expiry duration.
```
kevinAlbs, line 20:
```suggestion
```

Suggest removing interpretation of 0 as infinite for expiration. If accepted, also change `mongoc_oidc_credential_get_expires_in` to return a pointer.

`mongoc_oidc_credential_new` can be used to signal infinite expiration.

The auth spec links to https://openid.net/specs/openid-connect-core-1_0.html, which does not appear to note 0 is handled specially for `expires_in`.

There might be a future use-case for 0 to mean "do not cache". Quoting [Auth spec](https://github.com/mongodb/specifications/blob/0aee4aad0bc6710a8fae5910c36d41b8a60a0688/source/auth/auth.md):

> Note that the access token expiry value is currently not used in [Credential Caching](https://github.com/mongodb/specifications/blob/0aee4aad0bc6710a8fae5910c36d41b8a60a0688/source/auth/auth.md#credential-caching), but is intended to support future caching optimizations.

eramongodb: `expires_in` was initially defined as `int64_t*` such that:

```c
if (!expires_in_ptr) {
  // Infinite expiry duration.
} else if (*expires_in_ptr == 0) {
  // Immediate expiration.
} else {
  // Expiration relative to "now" immediately after callback returns.
}
```

After [this suggestion](https://github.com/mongodb/mongo-c-driver/pull/1945#discussion_r2012342667):

> Suggest removing interpretation of 0 to avoid inconsistency with CSOT's [interpretation of 0](https://github.com/mongodb/specifications/blob/cef66236cac7872852d1fb749bb34842eb7008fe/source/client-side-operations-timeout/client-side-operations-timeout.md#timeoutms) meaning "infinite timeout".

it's meaning was adjusted to:

```c
if (!expires_in_ptr || *expires_in_ptr == 0) {
  // Infinite expiry duration.
} else {
  // Expiration relative to "now" immediately after callback returns.
}
```

Therefore, after [this suggestion](https://github.com/mongodb/mongo-c-driver/pull/1945#discussion_r2023616735), this redundancy between "unset" vs. `0` was removed such that:

```c
if (expires_in == 0) {
  // Infinite expiry duration.
} else {
  // Expiration relative to "now" immediately after callback returns.
}
```

However, this feedback suggests "unset" may not necessarily mean "infinite expiry duration" as suggested by the spec:

> Drivers with optional parameters MAY interpret a missing value as infinite.

but instead may be interpreted as "immediate expiration" (aka "do not cache")?

```c
// Note: meaning NULL and 0 cases are swapped from initial design.
if (!expires_in_ptr) {
  // Immediate expiry duration. (?)
} else if (*expires_in_ptr == 0) {
  // Infinite expiry duration.
} else {
  // Expiration relative to "now" immediately after callback returns.
}
```

kevinAlbs: Since the Auth spec and OIDC spec do not appear to treat 0 specially, I propose:
```c
if (!expires_in_ptr) {
  // Infinite expiry duration.
} else {
  // Expiration relative to "now" immediately after callback returns.
}
```

I expect 0 would have the effect of immediate expiration. But I do not think that necessarily needs to be noted as a special case (it is a duration of 0).


> After [this suggestion](https://github.com/mongodb/mongo-c-driver/pull/1945#discussion_r2012342667):
> 
> > Suggest removing interpretation of 0 to avoid inconsistency with CSOT's [interpretation of 0](https://github.com/mongodb/specifications/blob/cef66236cac7872852d1fb749bb34842eb7008fe/source/client-side-operations-timeout/client-side-operations-timeout.md#timeoutms) meaning "infinite timeout".

Sorry for the roundabout feedback. On further thought, `expiresIn` does not seem comparable to CSOT's `timeoutMS`. CSOT is a driver timeout for "how much time remains before returning control to the caller". `expiresIn` is "how much time before the token is no longer valid". So I do not think they necessarily need to agree on what 0 means.

eramongodb: Reverted to `bool expires_in_set` + `const int64_t *` return value for `get_expires_in`. Updated documentation to avoid special-casing the value of `0` (for now). A value of "unset" (default) is once again interpreted as "infinite expiry duration".

src/libmongoc/src/mongoc/mongoc-oidc-callback.c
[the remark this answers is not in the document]
eramongodb: Following [this suggestion](https://github.com/mongodb/mongo-c-driver/pull/1945#discussion_r2025487499), reverted back to `optional<int64_t>` for the `expires_in` return value, which once again treats "unset" as an infinite expiry duration. The value of `0` is no longer treated specially (just a simple duration of `0`), but in the future may be interpreted as a "do not cache" request (special case for handling immediate expiration).

kevinAlbs (approved):

PR #1946, opened 2025-03-24
CDRIVER-5935 remove deprecated `bson_copy_to_excluding`

kevinAlbs:


vector-of-bool (approved):
LGTM.

Wondering if adopting something like [documentation-only attributes](https://mongodb-labs.github.io/mongo-c-driver-async/ref/meta-attrs/#attr-storage) may be useful for differentiating pointer-to-storage parameters.

kevinAlbs:
> Wondering if adopting something like [documentation-only attributes](https://mongodb-labs.github.io/mongo-c-driver-async/ref/meta-attrs/#attr-storage) may be useful for differentiating pointer-to-storage parameters.

I support that idea. I expect an attribute could help consistently document `NULL`-able parameters too.

PR #1952, opened 2025-03-24
CDRIVER-4489 revert aggressive validation of credential field requirements

eramongodb:
Followup to https://github.com/mongodb/mongo-c-driver/pull/1896 and https://github.com/mongodb/mongo-c-driver/pull/1939.

Reverts some URI auth finalization validation due to being too aggressive with authentication credential field requirements. In particular, requirements of the form "when A, then B must ..." where A and B are different credential fields (e.g. `authMechanism` and `authMechanismProperties`) are deferred to the Authentication Handshake implementation (i.e. `_mongoc_cluster_auth_node`). This is to support "late" setting of credential fields by the user, as demonstrated by the (partially-reverted) test cases in this PR which needed to be modified by https://github.com/mongodb/mongo-c-driver/pull/1896 due to the (aggressive) requirement that `authMechanism` must be specified given `authMechanismProperties`. This is no longer the case: a credential field is now only validated when it is specified and is not validated against other credential fields, as the other credential fields may be set later after finalization.

Username and password requirements for a given authentication mechanism are still validated and may return a client error during finalization. This is consistent with [prior behavior](https://github.com/mongodb/mongo-c-driver/blob/915277362a01f7304a6e119fb7c2030f5401dca9/src/libmongoc/src/mongoc/mongoc-uri.c#L1319) as well as Connection String spec tests (e.g. spec test: "should throw an exception if supplied a password (MONGODB-X509)"). Similarly, the non-empty authSource requirement is also validated for spec test conformance (e.g. spec test: "should throw an exception if authSource is invalid (MONGODB-X509)").


NEWS
```diff
@@ -10,9 +10,13 @@ Unreleased (2.0.0)
 * `mongoc_server_description_host` changes the return type from `mongoc_host_list_t *` to `const mongoc_host_list_t *`.
 * URI authentication credentials validation (only applicable during creation of a new `mongoc_uri_t` object from a connection string):
   * The requirement that a username is non-empty when specified is now enforced regardless of authentication mechanism.
+  * Username and password specification requirements are now validated and returns a client error for the specified authentication mechanism.
+    * e.g. it is a client error to not specify a username or a password for SCRAM-SHA-1, SCRAM-SHA-256, and GSSAPI.
```
kevinAlbs, line 14:
```suggestion
    * e.g. it is a client error to not specify a username or a password for SCRAM-SHA-1, SCRAM-SHA-256, and PLAIN.
    * e.g. it is a client error to not specify a username for GSSAPI.
```

kevinAlbs (approved):
LGTM with minor comments.

PR #1954, opened 2025-03-25
evg: remove macos-11-arm64 tasks for server 8.0+

eramongodb:
Resolves ongoing task setup failures for the macos-11-arm64 distro with server 8.0+ since around Feb 24 (latest) and March 12 (8.0). This appears to be due to the required server binaries now being built for MacOS 14. These binaries require dynamic linking with symbols which do not appear to be available on MacOS 11 (i.e. `std::__1::pmr::monotonic_buffer_resource`). Since we already have 8.0+ coverage on MacOS 14 (for which these binaries are built), this PR proposes simply removing coverage of 8.0+ on MacOS 11.

> [!NOTE]
> The [Platform Support Matrix](https://www.mongodb.com/docs/manual/administration/production-notes/#platform-support-matrix) entries for MacOS 11 (both x86_64 and ARM64) do not have checkmarks for 8.0+, which support the findings above.


kevinAlbs (approved):
Thank you. 

> The [Platform Support Matrix](https://www.mongodb.com/docs/manual/administration/production-notes/#platform-support-matrix) entries for MacOS 11 (both x86_64 and ARM64) do not have checkmarks for 8.0+, which support the findings above.

[MongoDB Platform Roadmap](https://docs.google.com/spreadsheets/d/1-sZKW70HbVt2yHOWa18qwBwi-gBcn54tWmronnn89kI/edit?gid=1790519872#gid=1790519872) similarly lists macOS 11 as EOL for MongoDB 8.0.


> require dynamic linking with symbols

Related: SERVER-101020 tracks this error. This error appears on macOS 13 too. This might be fixed to avoid (unintentionally?) dropping support on newer macOS versions. Regardless of the outcome, this PR seems like an improvement. Server 8.0 is documented as not supporting macOS 11.

PR #1955, opened 2025-03-25
[CDRIVER-5931] New forward-compatible CMake package files

vector-of-bool:
Refer: CDRIVER-5931 

# Summary

This changeset adds two new CMake config-mode packages: `bson` and `mongoc` [^bike], without any version suffix. These are intended to act as the new package names for the 2.x release, but we want to have them available for 1.x versions so that projects can target either version in their build without needing to try and detect which version is available.

[^bike]: Package names subject to bikeshedding.

The intent is that these package files will be included in future 1.x releases (or even backported and installed with older versions). For 1.x versions (and in this commit), the package files are relatively simple shims around the `{bson,mongoc}-1.0` packages. The new packages generate shim library targets that point back to the 1.x package targets. During the 2.x release, these package files will be modified to import the CMake targets directly.

# Details

## Package Import Behavior

Users are expected to modify their `find_package` calls to import the new package names, and link with the new target names:

```cmake
# Old:
find_package(bson-1.0 1.28.1)
target_link_libraries(my-app PRIVATE mongo::bson_static)

# New:
find_package(bson 1.28.1)
target_link_libraries(my-app PRIVATE bson::static)
```

If the caller specifies a `1.x` version, then these new packages will only ever import the 1.x version of the library, and will not work with the upcoming 2.x packages. Instead, importers can either omit the version (not recommended), or specify a version range that includes the 2.x version:

```cmake 
find_package(bson 1.28...2.999)
```

The above `find_package` call will work with both the 1.x packages and the upcoming 2.x packages. For either version, the `bson::static`/`bson::shared` imported targets will be available.

## Custom Version Checks

The `PackageConfigHelpers` functions are not used for the new package files, because they do not support version ranges that span major versions. The default `write_basic_package_version_file` will not support a version ranges like `1.0...2.0` in the way that we would expect. Instead, our own `pkgConfigVersion.cmake` files are provided, which are based on the default helpers, but implement the version range matching that we expect.

## Package File Names 

Previously, the packages were named as `<package>-config.cmake` (kebab-case). The new package files use `<package>Config.cmake` (camelCase). Using `kebab-case` for CMake package files results in case-insensitive name matching. The camelCase package filenames are case-sensitive.

## Package File Install Paths

The new package files are installed (by default) in `<prefix>/lib/cmake/<package>-<version>/` where `<version>` comes directly from the `project()`, and `<package>` is the package name intended for `find_package`. Using this naming scheme allows multiple versions of the library to be installed side-by-side in the same prefix. CMake will scan through all installed versions and pick one that accepts the version specified for `find_package()`. For example:

```
<prefix>/lib/cmake/mongoc-1.30.2/mongocConfig.cmake
<prefix>/lib/cmake/mongoc-2.0.1/mongocConfig.cmake
<prefix>/lib/cmake/mongoc-2.1.3/mongocConfig.cmake
<prefix>/lib/cmake/mongoc-2.2.0/mongocConfig.cmake
```

A call to `find_package(mongoc 2.1.0...<2.2)` will scan each package and select `2.1.3`.

The infix `lib/cmake/<name>-<version>` is chosen because it will work properly on both Windows and Unix systems according to [the `find_package` search procedure][fpsearch]. Downstream packagers can override this infix by setting `MONGOC_INSTALL_CMAKEDIR`.

[fpsearch]: https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure

## Other Targets

The new packages define two other targets: `bson::bson` and `mongoc::mongoc`, which are stubs that point to either the static library or the shared library, whichever is available, and whichever the user specifies according to an import-time configuration setting.

# Testing

A CMake module was added `TestProject.cmake`. This came from `amongoc` and allows defining CTest cases that execute CMake configure, build, and install automatically.

This PR adds test cases for several different import modes to assert that `find_package` plays nicely with the new packages depending on how the importer requests a certain version.

These test cases are not yet run in CI, as they are only defined in CTest and we do not yet have a good way to execute CTest as part of CI, and "just adding CTest" to a random build script will only exacerbate the convoluted CI process further.

# Other Changes

The following unrelated changes were made:

1. The CTest fixtures for our fake KMS HTTP server were misnamed, preventing them from starting up properly when CTest executes.
2. ~~The `bottle.py` Python script needs to be updated for newer Python compatibility (required for the KMS HTTP servers).~~ Reverted to keep the file history clean. Will do this in a separate PR.
3. A CTest label was added to all tests generated from `test-libmongoc`, so that it can be selected/excluded from a CTest run.
4. The install process writes temporary files into the build tree. This is usually fine, but if more than one installation is executing concurrently, then they can race and conflict with each other. This prevented running the CMake import-tests in parallel. This has been worked around by using `file(LOCK)` when generating and installing the relevant files in the build tree.


vector-of-bool:
The RPM build task now fails because the new package fails dare to utter the forbidden words `bson_static` and `mongoc_static`. Will require changes to the RPM spec to either patch out those references or be more lenient in checking for those strings.

src/libmongoc/examples/cmake/find_package/CMakeLists.txt
```diff
@@ -28,8 +28,8 @@ project (hello_mongoc LANGUAGES C)
 # building libmongoc.
 # -- sphinx-include-start --
 # Specify the minimum version you require.
-find_package (mongoc-1.0 1.7 REQUIRED)
+find_package (mongoc 1.7 REQUIRED)
```
kevinAlbs, line 31: Update `NEWS` and `src/libbson/NEWS` to describe the new package names.

kevinAlbs (approved):
Exciting! LGTM with updated NEWS and an option tweak.

vector-of-bool:
@kevinAlbs Added NEWS entries. Where else might we want to document these changes?

kevinAlbs (approved):
> @kevinAlbs Added NEWS entries. Where else might we want to document these changes?

I think documenting on https://www.mongodb.com/docs/languages/c/c-driver/current/get-started/ would be fitting. I do not see documentation there for including the C driver with CMake. Filed DOCSP-48728 and noted the upcoming package name changes.

PR #1956, opened 2025-03-26
[CDRIVER-3414] Remove deprecated legacy CMake package files

vector-of-bool:
Refer: CDRIVER-3414

This changeset deletes the CMake config-package files that were deprecated with CDRIVER-3413. New programs must use the new names (or the names that will be introduced in #1955)


kevinAlbs (approved):
At last! LGTM with additional removals of tests to fix failing smoke tests. Search for `BUILD_SAMPLE_WITH_CMAKE_DEPRECATED`.

rcsanchez97 (approved):

PR #1966, opened 2025-03-31
Apply `__cdecl` to all exported callbacks

kevinAlbs:
# Summary

Apply missing `BSON_CALL` to all exported callbacks to ensure `__cdecl` is applied when building with MSVC.

Resolves CDRIVER-2737 and CDRIVER-5678.

Verified with this [patch build](https://spruce.mongodb.com/version/67ea8bcf3bab8700076d6888).

# Background & Motivation

## Existing expectations

This PR assumes libbson and libmongoc do not support building with a non-cdecl calling convention. Attempting to build libbson with [/Gv](https://learn.microsoft.com/en-us/cpp/build/reference/gd-gr-gv-gz-calling-convention) to specify the `__vectorcall` calling convention results in errors:

```
src\libbson\src\bson\bson-clock.c(45): error C2373: 'bson_gettimeofday': redefinition; different type modifier
(... and many more ...)
```

This PR intends to ensure consumers building with MSVC see `__cdecl` in the function declaration when building headers to match the expected calling convention of libmongoc and libbson.

## Skipping CI testing

Similar to https://github.com/mongodb/mongo-cxx-driver/pull/1202, an attempt was made to build test-libmongoc with `/Gv` to test a non-cdecl consumer.

Doing so expectedly appears to require marking all callback functions defined in tests with `__cdecl` to avoid errors like:
```
src\libmongoc\tests\test-mongoc-command-monitoring.c(131): error C2440: 'function': cannot convert from 'void (__vectorcall *)(const mongoc_apm_command_failed_t *)' to 'mongoc_apm_command_failed_cb_t'
(... and many more ...)
```

And also appears to require marking functions in private headers used by tests to be marked `__cdecl`:
```
test-libmongoc-lib.lib(runner.obj) : error LNK2001: unresolved external symbol _mongoc_array_append_vals@@24
(... and many more ...)
```

To expedite the fix, tests in CI are not added.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -21,6 +21,7 @@ Unreleased (2.0.0)
     * Old behavior: `authMechanismProperties=A:B,C:D:E,F:G` is parsed as `{'A': 'B', 'C': 'D:E,F:G'}`.
     * New behavior: `authMechanismProperties=A:B,C:D:E,F:G` is parsed as `{'A': 'B': 'C': 'D:E', 'F': 'G'}`.
 * Calling `mongoc_bulk_operation_execute` on the same `mongoc_bulk_operation_t` repeatedly is an error. Previously this was only discouraged in documentation.
+* Consistently apply `__cdecl` calling convention to declarations in public API. Intended to support consumers built with other [calling conventions](https://learn.microsoft.com/en-us/cpp/build/reference/gd-gr-gv-gz-calling-convention) with MSVC. libmongoc and libbson only support building with `__cdecl`.
 
 ## Removals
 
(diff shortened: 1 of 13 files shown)


eramongodb (changes requested):
The pointer-to-function data members in [mongoc_stream_t](https://github.com/mongodb/mongo-c-driver/blob/845e1de2259041184fe4b741c4ee0f8d9ba02a8e/src/libmongoc/src/mongoc/mongoc-stream.h#L38) also require `__cdecl`.

kevinAlbs (commented):
> The pointer-to-function data members in [mongoc_stream_t](https://github.com/mongodb/mongo-c-driver/blob/845e1de2259041184fe4b741c4ee0f8d9ba02a8e/src/libmongoc/src/mongoc/mongoc-stream.h#L38) also require `__cdecl`.

Great catch. Updated.

eramongodb (approved):

PR #1967, opened 2025-03-31
CDRIVER-5956 note future min supported server version will be 4.2

kevinAlbs:


NEWS
```diff
@@ -23,6 +23,10 @@ New Features:
 Unreleased (2.0.0)
 ==================
 
+## Notes
+
+* A future minor release plans to raise the minimum supported MongoDB Server version from 4.0 to 4.2. This is in accordance with [MongoDB Software Lifecycle Schedules](https://www.mongodb.com/legal/support-policy/lifecycles).
```
rcsanchez97, line 28:
```suggestion
* In a future minor release the minimum supported MongoDB Server version will be raised from 4.0 to 4.2. This is in accordance with [MongoDB Software Lifecycle Schedules](https://www.mongodb.com/legal/support-policy/lifecycles).
```

Maybe I'm being overly pedantic, but your wording just felt a bit awkward. Feel free to discard my suggestion, though, as it is somewhat nitpicky.

kevinAlbs: Rewording SGTM. Thank you for the suggestion.

PR #1970, opened 2025-04-01
CDRIVER-5959 temporarily restrict updateDescription tests to 8.1 or older

eramongodb:
Preemptive changes mirroring those in https://github.com/mongodb/mongo-cxx-driver/pull/1365 until a proper solution is implemented for [DRIVERS-3151](https://jira.mongodb.org/browse/DRIVERS-3151).


kevinAlbs (approved):
Thank you.

PR #1972, opened 2025-04-01
CDRIVER-5968 Fix compiler warning on Windows

jeroen:
Fixes

```
  mongoc/mongoc-openssl.c:150:21: warning: format '%X' expects argument of type 'unsigned int', but argument 4 has type 'DWORD' {aka 'long unsigned int'} [-Wformat=]
```


kevinAlbs (approved):
Thank you for the fix @jeroen.

PR #1984, opened 2025-04-04
CDRIVER-5979 Update libmongoc for DRIVERS-3123 packed_bit change

ghost:
Updates libmongoc for compatibility with a proposed spec and test change affecting the packed_bit BSON Binary Vector format:

https://github.com/mongodb/specifications/pull/1783

* Copy updated spec test
* Remove obsolete workarounds
* modification to `append_vector_packed_bit_from_packed_array()`, asking it to treat masked writes as a conversion failure. (This "from packed array" operation is only for spec tests and doesn't otherwise exist in our API.)

### DIFF
--- a/src/libbson/tests/json/bson_binary_vector/packed_bit.json
+++ b/src/libbson/tests/json/bson_binary_vector/packed_bit.json
@@ -14,11 +14,11 @@
     {
       "description": "Simple Vector PACKED_BIT",
       "valid": true,
-      "vector": [127, 8],
+      "vector": [127, 7],
       "dtype_hex": "0x10",
       "dtype_alias": "PACKED_BIT",
       "padding": 0,
-      "canonical_bson": "1600000005766563746F7200040000000910007F0800"
+      "canonical_bson": "1600000005766563746F7200040000000910007F0700"
     },
     {
       "description": "PACKED_BIT with padding",


src/libbson/tests/test-bson-vector.c
```diff
@@ -123,6 +123,21 @@ append_vector_packed_bit_from_packed_array (
...
+         // Read back the packed byte, interpret any masking as a conversion failure.
+         uint8_t packed_byte_check;
+         ASSERT (bson_vector_packed_bit_view_read_packed (view, &packed_byte_check, 1, i));
+         if (packed_byte != packed_byte_check) {
+            bson_set_error (error,
+                            TEST_ERROR_DOMAIN,
+                            TEST_ERROR_CODE,
+                            "byte at index %" PRId64
+                            " with value 0x%02X included write to masked bits (reads as 0x%02X)",
+                            (int64_t) i,
```
eramongodb, line 136:
```suggestion
                            "byte at index %zu with value 0x%02X included write to masked bits (reads as 0x%02X)",
                            i,
```
`%zu` may be used to avoid the cast from `size_t`.

eramongodb (approved):

kevinAlbs (approved):

PR #1989, opened 2025-04-09
Reduce Warnings: address C4127 "conditional expression is constant" warnings

eramongodb:
Addresses 2448 C4127 "conditional expression is constant" warnings (1232 unique!) with VS 2022.

Most of these warnings are due the deliberate use of compile-time constants in conditional expressions either as configuration parameters, arguments to macro expansions, or platform-specific property checks.

The application of the `if (1) { ... } else ((void)0)` pattern to updated macros revealed some cases of branches without `{}`-scoping, which were applied accordingly as revealed by the `-Wdangling-else` warning for GCC and Clang.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-trace-private.h
+++ b/src/libmongoc/src/mongoc/mongoc-trace-private.h
@@ -28,53 +28,77 @@
 #include <mongoc/mongoc-log-private.h>
 #include <mongoc/mongoc-config.h>
 
+#include <mlib/config.h>
+
 
 BSON_BEGIN_DECLS
 
 // `gLogTrace` determines if tracing is enabled at runtime.
 extern bool gLogTrace;
 
-#define TRACE(msg, ...)                        \
-   do {                                        \
-      if (MONGOC_TRACE_ENABLED && gLogTrace) { \
-         mongoc_log (MONGOC_LOG_LEVEL_TRACE,   \
-                     MONGOC_LOG_DOMAIN,        \
-                     "TRACE: %s():%d " msg,    \
-                     BSON_FUNC,                \
-                     (int) (__LINE__),         \
-                     __VA_ARGS__);             \
-      }                                        \
-   } while (0)
+#define TRACE(msg, ...)                                         \
+   if (1) {                                                     \
+      mlib_diagnostic_push ();                                  \
+      mlib_disable_constant_conditional_expression_warnings (); \
+      if (MONGOC_TRACE_ENABLED && gLogTrace) {                  \
+         mongoc_log (MONGOC_LOG_LEVEL_TRACE,                    \
+                     MONGOC_LOG_DOMAIN,                         \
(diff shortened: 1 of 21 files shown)


src/libmongoc/src/mongoc/mongoc-trace-private.h
```diff
@@ -28,53 +28,82 @@
...
-                     BSON_FUNC,                \
-                     (int) (__LINE__),         \
-                     __VA_ARGS__);             \
-      }                                        \
-   } while (0)
+#define TRACE(msg, ...)                                         \
+   if (1) {                                                     \
+      mlib_diagnostic_push ();                                  \
+      mlib_disable_constant_conditional_expression_warnings (); \
+      if (MONGOC_TRACE_ENABLED && gLogTrace) {                  \
```
kevinAlbs, line 43: Confirming: does this trigger C4127? I expect the check with `gLogTrace` would make the conditional non-constant.

eramongodb: Yes, given `MONGOC_TRACE_ENABLED = 0` (false), short-circuiting skips `gLogTrace`, thus the condition is "always false" at compile-time.

kevinAlbs: Ah, right. Resolving.

src/libmongoc/src/mongoc/mongoc-trace-private.h
```diff
@@ -28,53 +28,82 @@
...
       }                                                                                                         \
       return;                                                                                                   \
-   } while (0)
+      mlib_diagnostic_pop ();                                                                                   \
+   } else                                                                                                       \
+      ((void) 0)
+
 #define RETURN(ret)                                                                                             \
-   do {                                                                                                         \
+   if (1) {                                                                                                     \
```
kevinAlbs, line 79: Heads up: this may need to be tweaked. Seems to be triggering a warning on the [mock-server-test](https://spruce.mongodb.com/task/mongo_c_driver_mock_server_test_mock_server_test_patch_66f23a9d463e842510fdb8f335ac0d8da1907a31_67f6ce212755bd0007484205_25_04_09_19_44_36/logs?execution=0) task:

```
mongoc-aggregate.c:318:1: error: control reaches end of non-void function [-Werror=return-type]
```

eramongodb: Yep, investigating. It appears to be a false-positive where it is not observing the `return` within the branch despite the `if (1)` constant conditional.

eramongodb: Looks like the if-else pattern triggers false-positives in multiple contexts, not just in `mongoc-aggregate.c` with GCC 11. Reverted trace macros containing `return` statements to their former do-while form to avoid triggering this issue.

PR #1998, opened 2025-04-24
CDRIVER-5956 raise min wire version from 7 to 8 (server 4.2)

kevinAlbs:
# Summary

- Raise minimum supported wire version from 7 (server 4.0) to 8 (server 4.2).
- Remove behavior and tests specific to maxWireVersion < 8.

Verified with this [patch build](https://spruce.mongodb.com/version/68090d0103e0920007b5cec5).

Removing no-longer-needed cursor implementations is not done in this PR and planned for CDRIVER-4722.

Auth tests using the test server are temporarily skipped until the test server is upgraded. Upgrading is tracked in DEVPROD-9029.

## Added `killCursors` in tests

Some tests that destroy cursors (`test_change_stream_resumable_error`) are updated to reflect behavior for server > 4.0. Quoting [SDAM](https://github.com/mongodb/specifications/blob/43d2c7bacd62249de8d2173bf8ee39e6fd7a686e/source/server-discovery-and-monitoring/server-discovery-and-monitoring.md#L939):

> If the old primary server version is 4.2 or newer, the client MUST NOT clear its connection pool for the old primary.

Previously the connection was closed after error. With the connection persisting, destroying the cursor [sends `killCursors`](https://github.com/mongodb/mongo-c-driver/blob/1955cffe9106d46957d0558753dcd869a04b07ac/src/libmongoc/src/mongoc/mongoc-client.c#L2121-L2129). Tests are updated to expect to send `killCursors` when destroying a cursor with a non-zero cursor ID and open connection.

### DIFF
--- a/.evergreen/config_generator/components/sasl/darwinssl.py
+++ b/.evergreen/config_generator/components/sasl/darwinssl.py
@@ -22,7 +22,7 @@ COMPILE_MATRIX = [
 ]
 
 TEST_MATRIX = [
-    ('macos-14', 'clang', None, 'cyrus', ['auth'], ['server'], ['4.0', '4.2', '4.4', '5.0']),
+    ('macos-14', 'clang', None, 'cyrus', ['auth'], ['server'], ['4.2', '4.4', '5.0']),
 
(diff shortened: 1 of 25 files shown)


vector-of-bool (approved):
LGTM. Yay for deleting code!

PR #1999, opened 2025-04-25
[CDRIVER-5996] Prevent multi-evaluation of arithmetic macro-function operands

vector-of-bool:
Currently, `mlib_upsize_integer` uses the expression `(0 & V)` to obtain a zero-value for the expression `V`. This has the unfortunate effect that it must evaluate `V` in order to obtain this value.

There's a better way: `1 ? 0 : V`, which will follow the same integer promotion rules as the infix `&`, but does not evaluate the right-hand operand. This now allows `mlib_upsize_integer` and all function-macros based on it to be used safely with operands that are expensive or produce side effects.


kevinAlbs (approved):
Neat, LGTM!

eramongodb (approved):
Although I am in favor of the improvement to avoiding (re)evaluation of the argument, I do not observe it addressing `-Wtype-limit` warnings [as suggested](https://github.com/mongodb/mongo-c-driver/pull/1992#issuecomment-2830979484) with GCC 13.3.0:

```
src/common/src/mlib/intutil.h:95:72: warning: comparison of unsigned expression in ‘< 0’ is always false [-Wtype-limits]
```

eramongodb:
It looks there may also be a GCC 14 [regression](https://godbolt.org/z/EfWMqn9br) where the `-Wtype-limits` warning is no longer emitted for the same code. 🤔

vector-of-bool:
@eramongodb I was able to repro and fix `-Wtype-limits`. It appears that `1 - 2 < 1` is enough to trick it 🙃.

PR #2003, opened 2025-04-30
DEVPROD-17323 - Replace perf.send with new data submission end point

MAhmadShah:
We replace the perf.send command with new shell scripts that submit the performance data to the new performance data end point available in signal processing service.

Successful patch with changes here, data can be viewed on the Trend Charts tab as before:
https://spruce.mongodb.com/version/681271b6fead0f0007f143bf/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/.evergreen/benchmark.yml
+++ b/.evergreen/benchmark.yml
@@ -170,10 +170,44 @@ functions:
       file_location: mongo-c-driver/report.json
 
   "send dashboard data" :
-    command: perf.send
-    params:
-      name: perf
-      file: mongo-c-driver/results.json
+    # Here we begin to generate the request to send the data to SPS
+    - command: shell.exec
+      params:
+        script: |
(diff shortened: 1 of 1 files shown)


.evergreen/benchmark.yml
```diff
@@ -170,10 +170,44 @@ functions:
       file_location: mongo-c-driver/report.json
 
   "send dashboard data" :
-    command: perf.send
-    params:
-      name: perf
-      file: mongo-c-driver/results.json
+    # Here we begin to generate the request to send the data to SPS
```
kevinAlbs, line 173:
```suggestion
    # Here we begin to generate the request to send the data to Signal Processing Service (SPS)
```

kevinAlbs (approved):
Thank you for the PR!

PR #2016, opened 2025-05-18
CDRIVER-5638 Record both FaaS and container metadata when both are present

mdb-ad:
Add environment variable checks for being in both a container and FaaS environment. Add that info to the handshake document. Sync spec test.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-handshake.c
+++ b/src/libmongoc/src/mongoc/mongoc-handshake.c
@@ -538,7 +538,7 @@ _mongoc_handshake_cleanup (void)
    bson_free (h->compiler_info);
    bson_free (h->flags);
    bson_free (h->env_region);
-   *h = (mongoc_handshake_t){0};
+   *h = (mongoc_handshake_t) {0};
 
    bson_mutex_destroy (&gHandshakeLock);
 }


src/libmongoc/src/mongoc/mongoc-handshake.c
```diff
@@ -351,39 +352,31 @@ _get_system_info (mongoc_handshake_t *handshake)
...
+_get_container_info (mongoc_handshake_t *handshake)
 {
-   bson_free (handshake->env_region);
+   char *kubernetes_env = _mongoc_getenv ("KUBERNETES_SERVICE_HOST");
+   handshake->kubernetes = kubernetes_env;
+
+   handshake->docker = false;
+#ifdef _POSIX_VERSION
+   handshake->docker = (access ("/.dockerenv", F_OK) == 0);
+#endif
```
kevinAlbs, line 377: Is a separate check needed if the Docker container is running Windows? Example:

```c
#ifdef _WIN32
   handshake->docker = (_access_s ("C:\\.dockerenv", 0) == 0);
#else
   handshake->docker = (access ("/.dockerenv", F_OK) == 0);
#endif
```

On non-Windows, I expect `access` can be assumed to be available since [POSIX features are enabled](https://github.com/mongodb/mongo-c-driver/blob/1b0e008d3aab07431827433c783f92680cb0cffe/CMakeLists.txt#L370).

kevinAlbs (approved):
LGTM with a test suggestion and a question about the Windows path.

src/libmongoc/src/mongoc/mongoc-handshake.c
mdb-ad: Right, I forgot Windows Docker containers exist

PR #2019, opened 2025-05-20
[CDRIVER-6017] Tweak visitor behavior when finding corrupt BSON

vector-of-bool:
This is a minimal fix for some edge cases around BSON iteration/visitation and text validation. A larger change could be made to make the APIs more complete and easier to use, but this is enough to fix a bug related to UTF-8 validation.

We now invoke `visit_corrupt` in more cases where `bson_visit_all` cannot properly decode elements, rather than just stopping with the boolean error code.

This also adds an error code to `bson_validate` for encountering corruption. Previously, it would just set `error->code = 0` and the message to `corrupt BSON`, which is counter-intuitive to those that expect `error->code == 0` to indicate success.

See also: CDRIVER-4448

### DIFF
--- a/src/libbson/src/bson/bson-types.h
+++ b/src/libbson/src/bson/bson-types.h
@@ -233,7 +233,7 @@ typedef enum {
     *
     * Passing this as a flag has no effect.
     */
-   BSON_VALIDATE_CORRUPT = (1 << 31),
+   BSON_VALIDATE_CORRUPT = (1 << 30),
 } bson_validate_flags_t;
 
 
--- a/src/libbson/src/bson/bson-types.h
+++ b/src/libbson/src/bson/bson-types.h
@@ -199,7 +199,6 @@ typedef enum {
     */
    BSON_VALIDATE_NONE = 0,
    /**
-    * @deprecated
     * @brief This flag has no effect
     *
     * @note Because invalid UTF-8 text is always invalid in BSON, the `bson_validate`
--- a/src/libbson/src/bson/bson-types.h
+++ b/src/libbson/src/bson/bson-types.h
@@ -185,25 +185,55 @@ typedef struct {
 
 
 /**
- * bson_validate_flags_t:
+ * @brief Flags an error code for BSON validation functions.
  *
- * This enumeration is used for validation of BSON documents. It allows
- * selective control on what you wish to validate.
+ * Pass these flags bits to control the behavior of the `bson_validate` family
+ * of functions.
  *
- * %BSON_VALIDATE_NONE: No additional validation occurs.
- * %BSON_VALIDATE_UTF8: Check that strings are valid UTF-8.
- * %BSON_VALIDATE_DOLLAR_KEYS: Check that keys do not start with $.
- * %BSON_VALIDATE_DOT_KEYS: Check that keys do not contain a period.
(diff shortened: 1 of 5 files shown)


src/libbson/src/bson/bson-types.h
```diff
@@ -185,25 +185,54 @@ typedef struct {
...
+    */
    BSON_VALIDATE_EMPTY_KEYS = (1 << 4),
+   /**
+    * @brief This is not a flag that controls behavior, but is instead used to indicate
+    * that a BSON document is corrupted in some way. This is the value that will
+    * appear as an error code.
+    *
+    * Passing this as a flag has no effect.
+    */
+   BSON_VALIDATE_CORRUPT = (1 << 30),
```
eramongodb, line 235: Can we reuse `BSON_VALIDATE_UTF8` instead? It already "has no effect" as a behavior control flag and already describes the very UTF-8 validity checks which `BSON_VALIDATE_CORRUPT` is meant to describe (appears redundant).

```
s/BSON_VALIDATE_CORRUPT/BSON_VALIDATE_UTF8/g
```

vector-of-bool: Maybe, but `VISIT_CORRUPT` is now (as of this PR) also used for when a subdocument is invalid (previously this was also just `return true;`, meaning that `bson_validate` _was also missing this case_).

I considered adding a `.visit_invalid_utf8` callback, which is potentially useful in general, but there's ambiguity in the semantics of "corruption", because the spec technically says that BSON strings are UTF-8, so a non-utf-8 string could arguably be considered corruption.

kevinAlbs (commented):
This change results in rejecting inserts that previously succeeded:
```c
bson_t to_insert = BSON_INITIALIZER;
ASSERT (bson_append_utf8 (&to_insert, "v", 1, "a\xFF\b", 3));
bool ok = mongoc_collection_insert_one (coll, &to_insert, NULL, NULL, &error);
ASSERT_OR_PRINT (ok, error);
// Before PR : OK
// After PR  : invalid document for insert: corrupt BSON
```

Write helpers validate UTF-8 [by default](https://github.com/mongodb/mongo-c-driver/blob/f84e270b2f0f75ee62c18ad3a08ec1f2ac14b801/src/libmongoc/src/mongoc/mongoc-util.c#L38-L45). However, this validation does not appear documented, and appears to have always been broken.

I would rather remove the UTF-8 validation in write helpers. I expect that would be a performance improvement, and remove the (low?) risk of breaking applications inserting invalid UTF-8.

Notably: UTF-8 key validation appeared to work (and is tested in PHP https://github.com/mongodb/mongo-php-driver/pull/1819#discussion_r2046696331).

It seems like most drivers do not validate UTF-8 on write. See [survey](https://docs.google.com/spreadsheets/d/1TUpq_o9w4FVAIv__k6dYLe3sDH6DuDN1KoKRtz1BNvo/edit?gid=0#gid=0) + [slack discussion](https://mongodb.slack.com/archives/C72LB5RPV/p1747931691652199).

**Proposal:**
- Do not validate UTF-8 values (but validate keys) if `BSON_VALIDATE_UTF8` is not set.
- Remove `BSON_VALIDATE_UTF8` from the default write helper flags (since UTF-8 validation never worked and the write helpers did not document expecting UTF-8 validation).
- Do validate UTF-8 in `bson_validate` if `BSON_VALIDATE_UTF8` is set (to match documented behavior of `bson_validate`).

PR #2024, opened 2025-05-29
CDRIVER-4269 support PKCS#8 keys with Secure Channel

kevinAlbs:
# Summary

Support PKCS#8 RSA keys with Windows Secure Channel

Patch build: https://spruce.mongodb.com/version/683853839983780007e8a268

# Background

Intended to ease using MONGODB-X509 with Atlas. Atlas creates client certificates in PKCS#8 with RSA keys. Prior to these changes, attempting to load a PKCS#8 key with the C driver built with Secure Channel results in this error:

```
Failed to parse private key. (0x8009310B) ASN1 bad tag value met.
```

This PR only adds support for unencrypted RSA keys. This is similar to the [server limitation loading from file](https://github.com/10gen/mongo/blob/0504451db32278eafb495cdde7f32f6304d1b256/src/mongo/util/net/ssl_manager_windows.cpp#L881-L885). I expect future support of non-RSA keys will require a more substantial change to switch to CNG (see CDRIVER-4530). This PR is intended to be backported to enable X509 auth to Atlas without the needed workaround described in CDRIVER-4269 (convert to PKCS#1).

The test client certificate, [pkcs8-unencrypted.pem](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/32fd8eb5de7be062c5750619c232189263371f98/.evergreen/x509gen/client-pkcs8-unencrypted.pem) is copied from drivers-evergreen-tools.


src/libmongoc/tests/test-mongoc-x509.c
```diff
@@ -143,6 +143,29 @@ test_x509_auth (void *unused)
...
+            uri, MONGOC_URI_TLSCERTIFICATEKEYFILE, CERT_TEST_DIR "/client-pkcs8-unencrypted.pem"));
+         ASSERT (mongoc_uri_set_option_as_utf8 (uri, MONGOC_URI_TLSCAFILE, CERT_CA));
+      }
+
+      // Try auth:
+      bson_error_t error = {0};
+      bool ok;
+      {
+         mongoc_client_t *client = test_framework_client_new_from_uri (uri, NULL);
+         ok = try_insert (client, &error);
```
eramongodb, line 161: Do the `/X509/*` tests need to be conditioned on _both_ `test_framework_skip_if_no_auth` and `test_framework_skip_if_no_server_ssl` given use of both user+pass and certs?

src/libmongoc/src/mongoc/mongoc-secure-channel.c
```diff
@@ -149,6 +149,56 @@ read_file_and_null_terminate (const char *filename, size_t *out_len)
...
 
+// `decode_object` decodes a cryptographic object from a blob.
+// Returns NULL on error.
+static LPBYTE
+decode_object (const char *structType,
+               const LPBYTE data,
+               DWORD data_len,
+               DWORD *out_len,
+               const char *descriptor,
+               const char *filename)
```
eramongodb, line 160: Missing `BSON_ASSERT_PARAM` / `BSON_OPTIONAL_PARAM` for pointer type parameters.

src/libmongoc/src/mongoc/mongoc-secure-channel.c
```diff
@@ -208,43 +248,47 @@ mongoc_secure_channel_setup_certificate_from_file (const char *filename)
...
-                                  0,
-                                  NULL,
-                                  blob_private,
-                                  &blob_private_len);
-   if (!success) {
-      char *msg = mongoc_winerr_to_string (GetLastError ());
-      MONGOC_ERROR ("Failed to parse private key: %s", msg);
-      bson_free (msg);
+      blob_private = decode_object (
+         PKCS_PRIVATE_KEY_INFO, encoded_private, encoded_private_len, &blob_private_rsa_len, "private key", filename);
```
eramongodb:
```suggestion
         PKCS_PRIVATE_KEY_INFO, encoded_private, encoded_private_len, &blob_private_len, "private key", filename);
```
This should be `blob_private_len` (for `blob_private`) rather than `blob_private_rsa_len` (for `blob_private_rsa`).

src/libmongoc/src/mongoc/mongoc-secure-channel.c
```diff
@@ -161,9 +211,9 @@ mongoc_secure_channel_setup_certificate_from_file (const char *filename)
    LPBYTE encoded_cert = NULL;
    const char *pem_public;
    const char *pem_private;
-   LPBYTE blob_private = NULL;
+   LPBYTE blob_private_rsa = NULL, blob_private = NULL;
```
eramongodb: Recommend avoiding multi-variable declarations with initializers.

```c
LPBYTE blob_private = NULL;
DWORD blob_private_len = 0;
LPBYTE blob_private_rsa = NULL;
DWORD blob_private_rsa_len = 0;
```

src/libmongoc/tests/test-mongoc-x509.c
kevinAlbs: Yes I think so. Updated to add `test_framework_skip_if_no_server_ssl`.

Aside: I expect this did not cause test failures because installing the test requires `MONGOC_ENABLE_SSL` be defined (to use SSL-only functions like `mongoc_client_set_ssl_opts`). `MONGOC_ENABLE_SSL` is defined if libmongoc is built with SSL support. I do not think there are Evergeen tasks testing libmongoc built with SSL to a server with SSL disabled (nor do I think that is needed IMO). But updated regardless to refine requirements.

src/libmongoc/src/mongoc/mongoc-secure-channel.c
kevinAlbs: Great spot. Updating.

src/libmongoc/tests/test-mongoc-x509.c
```diff
@@ -431,7 +454,13 @@ void
...
 #ifdef MONGOC_ENABLE_SSL
-   TestSuite_AddFull (suite, "/X509/auth", test_x509_auth, NULL, NULL, test_framework_skip_if_no_auth);
+   TestSuite_AddFull (suite,
+                      "/X509/auth",
+                      test_x509_auth,
+                      NULL,
+                      NULL,
+                      test_framework_skip_if_no_auth,
+                      test_framework_skip_if_no_server_ssl);
    TestSuite_AddFull (suite, "/X509/crl", test_crl, NULL, NULL, test_framework_skip_if_no_server_ssl);
```
eramongodb, line 464: Does `/509/crl` also need `skip_if_no_auth`?

kevinAlbs: No. The CRL test only verifies TLS rejects the server certificate. It does not try to do authentication.

PR #2026, opened 2025-06-03
[CDRIVER-6017] BSON Validation Refactor

vector-of-bool:
This changeset is a large refactor of the `bson_validate` implementation.

Existing validation test cases were rewritten to use a uniform code generation mechanism that notes the purpose of each case. Many more validation scenarios were added.

The `BSON_VALIDATE_UTF8` flags have been removed from the default flags used by the CRUD API. Because of a bug, UTF-8 validation wasn't being applied to the CRUD API, so removing the flag is necessary since it would otherwise be a breaking change.

The new validation is written as a straightforward recursive algorithm, and doesn't use the error-prone `bson_visit` API.

The following changes are of note:

1. The error code `BSON_VALIDATE_CORRUPT` is now set on corruption. Previously, this was set to `BSON_VALIDATE_NONE`, which had a value of zero. This was very confusing, as it was also set to zero in case of success.
2. Doc comments are added to better explain the behavior of the validation flags.
3. The "error offset" parameter is now adjusted properly in the case of document nesting.
4. Error message strings have been updated and made more explanatory.
5. BSON data that has over 1000 levels of nesting is now rejected, unconditionally.

### DIFF
--- a/src/libbson/src/bson/bson-types.h
+++ b/src/libbson/src/bson/bson-types.h
@@ -185,7 +185,7 @@ typedef struct {
 
 
 /**
- * @brief Flags an error code for BSON validation functions.
+ * @brief Flags and error codes for BSON validation functions.
  *
  * Pass these flags bits to control the behavior of the `bson_validate` family
  * of functions.
--- a/src/libbson/src/bson/bson-types.h
+++ b/src/libbson/src/bson/bson-types.h
@@ -235,18 +235,6 @@ typedef enum {
    BSON_VALIDATE_CORRUPT = (1 << 5),
 } bson_validate_flags_t;
 
-enum {
-   /**
-    * @brief This compile-time constant represents the maximum document nesting
-    * depth permitted by the `bson_validate` family of functions. If the nesting
-    * depth exceeds this limit, the data will be rejected.
-    *
-    * The depth limit of 100 is chosen to match the limit enforced by MongoDB server.
-    * Refer: https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Nested-Depth-for-BSON-Document
-    */
-   BSON_VALIDATION_MAX_NESTING_DEPTH = 100,
-};
-
 
 /**
  * bson_type_t:
--- a/src/libbson/src/bson/bson-types.h
+++ b/src/libbson/src/bson/bson-types.h
@@ -185,25 +185,54 @@ typedef struct {
 
 
 /**
- * bson_validate_flags_t:
+ * @brief Flags an error code for BSON validation functions.
  *
- * This enumeration is used for validation of BSON documents. It allows
- * selective control on what you wish to validate.
+ * Pass these flags bits to control the behavior of the `bson_validate` family
+ * of functions.
  *
- * %BSON_VALIDATE_NONE: No additional validation occurs.
- * %BSON_VALIDATE_UTF8: Check that strings are valid UTF-8.
- * %BSON_VALIDATE_DOLLAR_KEYS: Check that keys do not start with $.
- * %BSON_VALIDATE_DOT_KEYS: Check that keys do not contain a period.
- * %BSON_VALIDATE_UTF8_ALLOW_NULL: Allow NUL bytes in UTF-8 text.
- * %BSON_VALIDATE_EMPTY_KEYS: Prohibit zero-length field names
+ * Additionally, if validation fails, then the error code set on a `bson_error_t`
+ * will have the value corresponding to the reason that validation failed.
  */
 typedef enum {
+   /**
+    * @brief No special validation behavior specified.
+    */
    BSON_VALIDATE_NONE = 0,
+   /**
+    * @brief Check that all text components of the BSON data are valid UTF-8.
+    *
+    * Note that this will also cause validation to reject valid text that contains
+    * a null character. This can be changed by also passing
+    * `BSON_VALIDATE_UTF8_ALLOW_NULL`
+    */
    BSON_VALIDATE_UTF8 = (1 << 0),
+   /**
+    * @brief Check that element keys do not begin with an ASCII dollar `$`
+    */
    BSON_VALIDATE_DOLLAR_KEYS = (1 << 1),
+   /**
+    * @brief Check that element keys do not contain an ASCII period `.`
+    */
    BSON_VALIDATE_DOT_KEYS = (1 << 2),
+   /**
+    * @brief If set then it is *not* an error for a UTF-8 string to contain
+    * embedded null characters.
+    *
+    * This has no effect unless `BSON_VALIDATE_UTF8` is also passed.
+    */
    BSON_VALIDATE_UTF8_ALLOW_NULL = (1 << 3),
+   /**
+    * @brief Check that no element key is a zero-length empty string.
+    */
    BSON_VALIDATE_EMPTY_KEYS = (1 << 4),
+   /**
+    * @brief This is not a flag that controls behavior, but is instead used to indicate
+    * that a BSON document is corrupted in some way. This is the value that will
+    * appear as an error code.
+    *
+    * Passing this as a flag has no effect.
+    */
+   BSON_VALIDATE_CORRUPT = (1 << 30),
 } bson_validate_flags_t;
 
 
--- a/src/libbson/src/bson/bson-types.h
+++ b/src/libbson/src/bson/bson-types.h
@@ -232,7 +232,7 @@ typedef enum {
     *
     * Passing this as a flag has no effect.
     */
-   BSON_VALIDATE_CORRUPT = (1 << 30),
+   BSON_VALIDATE_CORRUPT = (1 << 5),
 } bson_validate_flags_t;
 
 
--- a/src/libbson/src/bson/bson-types.h
+++ b/src/libbson/src/bson/bson-types.h
@@ -235,6 +235,18 @@ typedef enum {
    BSON_VALIDATE_CORRUPT = (1 << 5),
 } bson_validate_flags_t;
 
+enum {
+   /**
+    * @brief This compile-time constant represents the maximum document nesting
+    * depth permitted by the `bson_validate` family of functions. If the nesting
+    * depth exceeds this limit, the data will be rejected.
+    *
+    * The depth limit of 100 is chosen to match the limit enforced by MongoDB server.
+    * Refer: https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Nested-Depth-for-BSON-Document
+    */
+   BSON_VALIDATION_MAX_NESTING_DEPTH = 100,
+};
+
 
 /**
  * bson_type_t:
--- a/src/libbson/src/bson/bson.c
+++ b/src/libbson/src/bson/bson.c
@@ -2502,6 +2502,10 @@ bson_validate_with_error_and_offset (const bson_t *bson,
                                      size_t *offset,
                                      bson_error_t *error)
 {
+   BSON_ASSERT_PARAM (bson);
+   BSON_OPTIONAL_PARAM (offset);
+   BSON_OPTIONAL_PARAM (error);
+
    size_t offset_local = 0;
    if (!offset) {
       offset = &offset_local;
--- a/src/libbson/src/bson/bson.c
+++ b/src/libbson/src/bson/bson.c
@@ -15,6 +15,7 @@
  */
 
 
+#include <bson/validate-private.h>
 #include <mlib/intencode.h>
 #include <bson/bson.h>
 #include <bson/bson-config.h>
@@ -2491,10 +2492,6 @@ bson_validate_with_error (const bson_t *bson, bson_validate_flags_t flags, bson_
 {
    return bson_validate_with_error_and_offset (bson, flags, NULL, error);
 }
-
-// Defined in validate.c
-extern bool
-_bson_validate_impl_v2 (const bson_t *bson, bson_validate_flags_t flags, size_t *offset, bson_error_t *error);
 
 bool
 bson_validate_with_error_and_offset (const bson_t *bson,
--- a/src/libbson/src/bson/bson.c
+++ b/src/libbson/src/bson/bson.c
@@ -31,29 +31,6 @@
 #include <mlib/config.h>
 
 
-typedef enum {
-   BSON_VALIDATE_PHASE_START,
-   BSON_VALIDATE_PHASE_TOP,
-   BSON_VALIDATE_PHASE_LF_REF_KEY,
-   BSON_VALIDATE_PHASE_LF_REF_UTF8,
-   BSON_VALIDATE_PHASE_LF_ID_KEY,
-   BSON_VALIDATE_PHASE_LF_DB_KEY,
-   BSON_VALIDATE_PHASE_LF_DB_UTF8,
-   BSON_VALIDATE_PHASE_NOT_DBREF,
-} bson_validate_phase_t;
-
-
-/*
- * Structures.
- */
-typedef struct {
-   bson_validate_flags_t flags;
-   ssize_t err_offset;
-   bson_validate_phase_t phase;
-   bson_error_t error;
-} bson_validate_state_t;
-
-
 /*
  * Globals.
  */
@@ -2502,196 +2479,6 @@ bson_array_as_canonical_extended_json (const bson_t *bson, size_t *length)
 }
 
 
-#define VALIDATION_ERR(_flag, _msg, ...) bson_set_error (&state->error, BSON_ERROR_INVALID, _flag, _msg, __VA_ARGS__)
-
-static bool
-_bson_iter_validate_utf8 (const bson_iter_t *iter, const char *key, size_t v_utf8_len, const char *v_utf8, void *data)
-{
-   bson_validate_state_t *state = data;
-   bool allow_null;
-
-   if ((state->flags & BSON_VALIDATE_UTF8)) {
-      allow_null = !!(state->flags & BSON_VALIDATE_UTF8_ALLOW_NULL);
-
-      if (!bson_utf8_validate (v_utf8, v_utf8_len, allow_null)) {
-         state->err_offset = iter->off;
-         VALIDATION_ERR (BSON_VALIDATE_UTF8, "invalid utf8 string for key \"%s\"", key);
-         return true;
-      }
-   }
-
-   if ((state->flags & BSON_VALIDATE_DOLLAR_KEYS)) {
-      if (state->phase == BSON_VALIDATE_PHASE_LF_REF_UTF8) {
-         state->phase = BSON_VALIDATE_PHASE_LF_ID_KEY;
-      } else if (state->phase == BSON_VALIDATE_PHASE_LF_DB_UTF8) {
-         state->phase = BSON_VALIDATE_PHASE_NOT_DBREF;
-      }
-   }
-
-   return false;
-}
-
-
-static void
-_bson_iter_validate_corrupt (const bson_iter_t *iter, void *data)
-{
-   bson_validate_state_t *state = data;
-
-   state->err_offset = iter->err_off;
-   VALIDATION_ERR (BSON_VALIDATE_NONE, "%s", "corrupt BSON");
-}
-
-
-static bool
-_bson_iter_validate_before (const bson_iter_t *iter, const char *key, void *data)
-{
-   bson_validate_state_t *state = data;
-
-   if ((state->flags & BSON_VALIDATE_EMPTY_KEYS)) {
-      if (key[0] == '\0') {
-         state->err_offset = iter->off;
-         VALIDATION_ERR (BSON_VALIDATE_EMPTY_KEYS, "%s", "empty key");
-         return true;
-      }
-   }
-
-   if ((state->flags & BSON_VALIDATE_DOLLAR_KEYS)) {
-      if (key[0] == '$') {
-         if (state->phase == BSON_VALIDATE_PHASE_LF_REF_KEY && strcmp (key, "$ref") == 0) {
-            state->phase = BSON_VALIDATE_PHASE_LF_REF_UTF8;
-         } else if (state->phase == BSON_VALIDATE_PHASE_LF_ID_KEY && strcmp (key, "$id") == 0) {
-            state->phase = BSON_VALIDATE_PHASE_LF_DB_KEY;
-         } else if (state->phase == BSON_VALIDATE_PHASE_LF_DB_KEY && strcmp (key, "$db") == 0) {
-            state->phase = BSON_VALIDATE_PHASE_LF_DB_UTF8;
-         } else {
-            state->err_offset = iter->off;
-            VALIDATION_ERR (BSON_VALIDATE_DOLLAR_KEYS, "keys cannot begin with \"$\": \"%s\"", key);
-            return true;
-         }
-      } else if (state->phase == BSON_VALIDATE_PHASE_LF_ID_KEY || state->phase == BSON_VALIDATE_PHASE_LF_REF_UTF8 ||
-                 state->phase == BSON_VALIDATE_PHASE_LF_DB_UTF8) {
-         state->err_offset = iter->off;
-         VALIDATION_ERR (BSON_VALIDATE_DOLLAR_KEYS, "invalid key within DBRef subdocument: \"%s\"", key);
-         return true;
-      } else {
-         state->phase = BSON_VALIDATE_PHASE_NOT_DBREF;
-      }
-   }
-
-   if ((state->flags & BSON_VALIDATE_DOT_KEYS)) {
-      if (strstr (key, ".")) {
-         state->err_offset = iter->off;
-         VALIDATION_ERR (BSON_VALIDATE_DOT_KEYS, "keys cannot contain \".\": \"%s\"", key);
-         return true;
-      }
-   }
-
-   return false;
-}
-
-
-static bool
-_bson_iter_validate_codewscope (
-   const bson_iter_t *iter, const char *key, size_t v_code_len, const char *v_code, const bson_t *v_scope, void *data)
-{
-   bson_validate_state_t *state = data;
-   size_t offset = 0;
-
-   BSON_UNUSED (key);
-   BSON_UNUSED (v_code_len);
-   BSON_UNUSED (v_code);
-
-   if (!bson_validate (v_scope, state->flags, &offset)) {
-      state->err_offset = iter->off + offset;
-      VALIDATION_ERR (BSON_VALIDATE_NONE, "%s", "corrupt code-with-scope");
-      return false;
-   }
-
-   return true;
-}
-
-
-static bool
-_bson_iter_validate_document (const bson_iter_t *iter, const char *key, const bson_t *v_document, void *data);
-
-
-static const bson_visitor_t bson_validate_funcs = {
-   _bson_iter_validate_before,
-   NULL, /* visit_after */
-   _bson_iter_validate_corrupt,
-   NULL, /* visit_double */
-   _bson_iter_validate_utf8,
-   _bson_iter_validate_document,
-   _bson_iter_validate_document, /* visit_array */
-   NULL,                         /* visit_binary */
-   NULL,                         /* visit_undefined */
-   NULL,                         /* visit_oid */
-   NULL,                         /* visit_bool */
-   NULL,                         /* visit_date_time */
-   NULL,                         /* visit_null */
-   NULL,                         /* visit_regex */
-   NULL,                         /* visit_dbpoint */
-   NULL,                         /* visit_code */
-   NULL,                         /* visit_symbol */
-   _bson_iter_validate_codewscope,
-};
-
-
-static bool
-_bson_iter_validate_document (const bson_iter_t *iter, const char *key, const bson_t *v_document, void *data)
-{
-   bson_validate_state_t *state = data;
-   bson_iter_t child;
-   bson_validate_phase_t phase = state->phase;
-
-   BSON_UNUSED (key);
-
-   if (!bson_iter_init (&child, v_document)) {
-      state->err_offset = iter->off;
-      return true;
-   }
-
-   if (state->phase == BSON_VALIDATE_PHASE_START) {
-      state->phase = BSON_VALIDATE_PHASE_TOP;
-   } else {
-      state->phase = BSON_VALIDATE_PHASE_LF_REF_KEY;
-   }
-
-   (void) bson_iter_visit_all (&child, &bson_validate_funcs, state);
-
-   if (state->phase == BSON_VALIDATE_PHASE_LF_ID_KEY || state->phase == BSON_VALIDATE_PHASE_LF_REF_UTF8 ||
-       state->phase == BSON_VALIDATE_PHASE_LF_DB_UTF8) {
-      if (state->err_offset <= 0) {
-         state->err_offset = iter->off;
-      }
-
-      return true;
-   }
-
-   state->phase = phase;
-
-   return false;
-}
-
-
-static void
-_bson_validate_internal (const bson_t *bson, bson_validate_state_t *state)
-{
-   bson_iter_t iter;
-
-   state->err_offset = -1;
-   state->phase = BSON_VALIDATE_PHASE_START;
-   memset (&state->error, 0, sizeof state->error);
-
-   if (!bson_iter_init (&iter, bson)) {
-      state->err_offset = 0;
-      VALIDATION_ERR (BSON_VALIDATE_NONE, "%s", "corrupt BSON");
-   } else {
-      _bson_iter_validate_document (&iter, NULL, bson, state);
-   }
-}
-
-
 bool
 bson_validate (const bson_t *bson, bson_validate_flags_t flags, size_t *offset)
 {
@@ -2705,6 +2492,9 @@ bson_validate_with_error (const bson_t *bson, bson_validate_flags_t flags, bson_
    return bson_validate_with_error_and_offset (bson, flags, NULL, error);
 }
 
+// Defined in validate.c
+extern bool
+_bson_validate_impl_v2 (const bson_t *bson, bson_validate_flags_t flags, size_t *offset, bson_error_t *error);
 
 bool
 bson_validate_with_error_and_offset (const bson_t *bson,
@@ -2712,22 +2502,16 @@ bson_validate_with_error_and_offset (const bson_t *bson,
                                      size_t *offset,
                                      bson_error_t *error)
 {
-   bson_validate_state_t state;
-
-   state.flags = flags;
-   _bson_validate_internal (bson, &state);
-
-   if (state.err_offset >= 0) {
-      if (offset) {
-         *offset = (size_t) state.err_offset;
-      }
-      if (error) {
-         memcpy (error, &state.error, sizeof *error);
-      }
-      return false;
+   size_t offset_local = 0;
+   if (!offset) {
+      offset = &offset_local;
+   }
+   bson_error_t error_local;
+   if (!error) {
+      error = &error_local;
    }
 
-   return true;
+   return _bson_validate_impl_v2 (bson, flags, offset, error);
 }
 
 
--- a/src/libbson/src/bson/validate-private.h
+++ b/src/libbson/src/bson/validate-private.h
@@ -1,0 +1,22 @@
+#ifndef BSON_VALIDATE_PRIVATE_H_INCLUDED
+#define BSON_VALIDATE_PRIVATE_H_INCLUDED
+
+#include <bson/bson-types.h>
+
+/**
+ * @brief Private function backing the implementation of validation.
+ *
+ * Validation was previously defined in the overburdened `bson-iter.c`, but it
+ * is now defined in its own file.
+ *
+ * @param bson The document to validate. Must be non-null.
+ * @param flags Validation control flags
+ * @param offset Receives the offset at which validation failed. Must be non-null.
+ * @param error Receives the error describing why validation failed. Must be non-null.
+ * @return true If the given document has no validation errors
+ * @return false Otherwise
+ */
+extern bool
+_bson_validate_impl_v2 (const bson_t *bson, bson_validate_flags_t flags, size_t *offset, bson_error_t *error);
+
+#endif // BSON_VALIDATE_PRIVATE_H_INCLUDED
--- a/src/libbson/src/bson/validate-private.h
+++ b/src/libbson/src/bson/validate-private.h
@@ -16,7 +16,7 @@
  * @return true If the given document has no validation errors
  * @return false Otherwise
  */
-extern bool
+bool
 _bson_validate_impl_v2 (const bson_t *bson, bson_validate_flags_t flags, size_t *offset, bson_error_t *error);
 
 #endif // BSON_VALIDATE_PRIVATE_H_INCLUDED
--- a/src/libbson/src/bson/validate-private.h
+++ b/src/libbson/src/bson/validate-private.h
@@ -3,6 +3,21 @@
 
 #include <bson/bson-types.h>
 
+enum {
+   /**
+    * @brief This compile-time constant represents the maximum document nesting
+    * depth permitted by the `bson_validate` family of functions. If the nesting
+    * depth exceeds this limit, the data will be rejected.
+    *
+    * This limit is intentionally larger than the default limit of MongoDB
+    * server, since we cannot anticipate what a libbson user might actually want
+    * to do with BSON, and to prevent accidentally rejecting data that the
+    * server might accept. The main purpose of this limit is to prevent stack
+    * overflow, not to reject invalid data.
+    */
+   BSON_VALIDATION_MAX_NESTING_DEPTH = 1000,
+};
+
 /**
  * @brief Private function backing the implementation of validation.
  *
--- a/src/libbson/src/bson/validate.c
+++ b/src/libbson/src/bson/validate.c
@@ -1,0 +1,506 @@
+/**
+ * @file bson/validate.c
+ * @brief Implementation of BSON document validation
+ * @date 2025-05-28
+ *
+ * This file implements the backend for the `bson_validate` family of functions.
+ *
+ * @copyright Copyright 2009-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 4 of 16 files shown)


src/libbson/src/bson/bson-types.h
```diff
@@ -185,25 +185,54 @@ typedef struct {
...
+    */
    BSON_VALIDATE_EMPTY_KEYS = (1 << 4),
+   /**
+    * @brief This is not a flag that controls behavior, but is instead used to indicate
+    * that a BSON document is corrupted in some way. This is the value that will
+    * appear as an error code.
+    *
+    * Passing this as a flag has no effect.
+    */
+   BSON_VALIDATE_CORRUPT = (1 << 30),
```
kevinAlbs: Also add `BSON_VALIDATE_CORRUPT` to RST docs.

src/libbson/src/bson/bson-types.h
```diff
@@ -185,25 +185,54 @@ typedef struct {
 
 
 /**
- * bson_validate_flags_t:
+ * @brief Flags an error code for BSON validation functions.
```
kevinAlbs:
```suggestion
 * @brief Flags and error codes for BSON validation functions.
```

src/libbson/tests/validate-tests.py
```diff
@@ -0,0 +1,1376 @@
...
+
+
+@dataclass(frozen=True)
+class TestCase:
+    """
+    Defines a single validation test case.
+    """
+
+    name: str
+    """The name of the test case, as displayed in test runners, which will valid a "/bson/validate" prefix"""
```
kevinAlbs:
```suggestion
    """The name of the test case, as displayed in test runners, which will have a "/bson/validate" prefix"""
```

src/libbson/tests/validate-tests.py
```diff
@@ -0,0 +1,1376 @@
...
+    byte offset.
+
+    Note that this won't match if the error message is something other
+    than "corrupt BSON".
+    """
+    return ErrorInfo("BSON_VALIDATE_CORRUPT", "corrupt BSON", off)
+
+
+BSON_VALIDATE_DOLLAR_KEYS = "BSON_VALIDATE_DOLLAR_KEYS"
+BSON_VALIDATE_UTF8 = "BSON_VALIDATE_UTF8"
```
kevinAlbs, line 264: Minor: consider defining variables for all flags for consistency (currently there is a mix of variables and string literals):
```python
BSON_VALIDATE_NONE="BSON_VALIDATE_NONE"
BSON_VALIDATE_UTF8="BSON_VALIDATE_UTF8"
BSON_VALIDATE_DOLLAR_KEYS="BSON_VALIDATE_DOLLAR_KEYS"
BSON_VALIDATE_DOT_KEYS="BSON_VALIDATE_DOT_KEYS"
BSON_VALIDATE_UTF8_ALLOW_NULL="BSON_VALIDATE_UTF8_ALLOW_NULL"
BSON_VALIDATE_EMPTY_KEYS="BSON_VALIDATE_EMPTY_KEYS"
BSON_VALIDATE_CORRUPT="BSON_VALIDATE_CORRUPT"
```

The auto-complete may slightly help write tests.

src/libbson/tests/validate-tests.py
```diff
@@ -0,0 +1,1376 @@
...
+        this because we don't pass BSON_VALIDATE_UTF8_ALLOW_NULL.
+        """,
+        flags=BSON_VALIDATE_UTF8,
+        error=ErrorInfo("BSON_VALIDATE_UTF8_ALLOW_NULL", "UTF-8 string contains a U+0000 (null) character", 4),
+    ),
+    TestCase(
+        "utf8-key/invalid/accept",
+        doc(utf8elem(b"abc\xffdef", "bar")),
+        """
+        The element key is not valid UTf-8, but we acecpt it if we don't do
```
kevinAlbs:
```suggestion
        The element key is not valid UTf-8, but we accept it if we don't do
```

src/libbson/tests/validate-tests.py
```diff
@@ -0,0 +1,1376 @@
...
+    TestCase(
+        "code-with-scope/invalid-scope-key",
+        doc(
+            elem(
+                "code",
+                Tag.CodeWithScope,
+                code_with_scope("void 0;", doc(utf8elem("", "some string"))),
+            )
+        ),
+        "A code-with-scope element, but the socpe contains an empty element key",
```
kevinAlbs:
```suggestion
        "A code-with-scope element, but the scope contains an empty element key. The scope is validated with stricter parameters and expected to error even without BSON_VALIDATE_EMPTY_KEYS.",
```

Suggest noting that BSON_VALIDATE_EMPTY_KEYS is intentionally missing.

src/libbson/tests/validate-tests.py
```diff
@@ -0,0 +1,1376 @@
...
+        is delimited by its null terminator, the iterator will stop early before
+        the actual EOD.
+        """,
+        error=corruption_at(21),
+    ),
+    TestCase(
+        "regex/invalid-utf8/accept",
+        doc(elem("regex", Tag.Regex, b"foo\xffbar\0gi\0")),
+        """
+        A regular expression that contains invalid UTf-8.
```
kevinAlbs:
```suggestion
        A regular expression that contains invalid UTF-8.
```

src/libbson/tests/validate-tests.py
```diff
@@ -0,0 +1,1376 @@
...
+        """,
+    ),
+    TestCase(
+        "regex/invalid-utf8/reject",
+        doc(elem("regex", Tag.Regex, b"foo\xffbar\0gi\0")),
+        """
+        A regular expression that contains invalid UTf-8.
```
kevinAlbs:
```suggestion
        A regular expression that contains invalid UTF-8.
```

src/libmongoc/CMakeLists.txt
```diff
@@ -998,7 +998,9 @@ set (test-libmongoc-sources
    ${mongo-c-driver_SOURCE_DIR}/src/libbson/tests/test-reader.c
    ${mongo-c-driver_SOURCE_DIR}/src/libbson/tests/test-string.c
    ${mongo-c-driver_SOURCE_DIR}/src/libbson/tests/test-utf8.c
+   ${mongo-c-driver_SOURCE_DIR}/src/libbson/tests/test-validate.generated.c
    ${mongo-c-driver_SOURCE_DIR}/src/libbson/tests/test-value.c
+   ${mongo-c-driver_SOURCE_DIR}/src/libbson/tests/test-validate.generated.c
```
eramongodb:
```suggestion
```
Duplicate entry.

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+
+// Validate the value of an element, without checking its key
+static bool
+_validate_element_value (validator *self, bson_iter_t const *iter, int depth)
+{
+   const bson_type_t type = bson_iter_type (iter);
+   switch (type) {
+   default:
+   case BSON_TYPE_EOD:
+      mlib_check (false && "Unreachable");
```
eramongodb:
```suggestion
      BSON_UNREACHABLE("BSON_TYPE_EOD");;
```
Address the following GCC warning:
```
warning: this statement may fall through [-Wimplicit-fallthrough=]
  132 |    _mlibCheckConditionSimple (Condition, ArgString, mlib_this_source_location ())
      |    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/common/src/mlib/config.h:40:24: note: in definition of macro ‘MLIB_JUST’
   40 | #define MLIB_JUST(...) __VA_ARGS__
      |                        ^~~~~~~~~~~
src/common/src/mlib/config.h:54:29: note: in expansion of macro ‘_mlib_check_argc_2’
   54 | #define _mlibPaste1(A, ...) A##__VA_ARGS__
      |                             ^
src/common/src/mlib/config.h:44:28: note: in expansion of macro ‘_mlibPaste1’
   44 | #define MLIB_PASTE(A, ...) _mlibPaste1 (A, __VA_ARGS__)
      |                            ^~~~~~~~~~~
src/common/src/mlib/config.h:49:33: note: in expansion of macro ‘MLIB_PASTE’
   49 | #define MLIB_PASTE_3(A, B, ...) MLIB_PASTE (A, MLIB_PASTE (B, __VA_ARGS__))
      |                                 ^~~~~~~~~~
src/common/src/mlib/config.h:151:15: note: in expansion of macro ‘MLIB_PASTE_3’
  151 |    MLIB_JUST (MLIB_PASTE_3 (Prefix, _argc_, MLIB_ARG_COUNT (__VA_ARGS__)) (__VA_ARGS__))
      |               ^~~~~~~~~~~~
src/common/src/mlib/test.h:129:25: note: in expansion of macro ‘MLIB_ARGC_PICK’
  129 | #define mlib_check(...) MLIB_ARGC_PICK (_mlib_check, #__VA_ARGS__, __VA_ARGS__)
      |                         ^~~~~~~~~~~~~~
src/libbson/src/bson/validate.c:329:7: note: in expansion of macro ‘mlib_check’
  329 |       mlib_check (false && "Unreachable");
      |       ^~~~~~~~~~
src/libbson/src/bson/validate.c:330:4: note: here
  330 |    case BSON_TYPE_DOUBLE:
      |    ^~~~
```

Should this really be an assertion? Can this be treated as a validation error instead?

src/libbson/tests/validate-tests.py
```diff
@@ -0,0 +1,1376 @@
...
+        Simple binary data { "binary": Binary(0x80, b'12345') }
+        """,
+    ),
+    TestCase(
+        "binary/bad-length-zero-subtype-2",
+        doc(
+            elem(
+                "binary",
+                Tag.Binary,
+                i32le(0),  # Invalid: Zero length
```
kevinAlbs, line 1106: I expect this test intended to set the invalid length on the inner length (separate from the `binary/bad-length-too-small` test below). Suggest updating as:

```python
    TestCase(
        "binary/bad-length-zero-subtype-2",
        doc(
            elem(
                "binary",
                Tag.Binary,
                i32le(8), # Length of 8
                b"\x02",  # subtype two
                i32le(0), # Invalid: Zero inner length
                b"1234",  # payload
            ),
        ),
        """
        Binary data that has an invalid inner length header. It is subtype 2,
        which means it contains an additional length header.
        """,
        error=corruption_at(17),
    ),
```

src/libbson/tests/validate-tests.py
```diff
@@ -0,0 +1,1376 @@
...
+    TestCase(
+        "boolean/simple-true",
+        doc(elem("bool", Tag.Boolean, b"\x01")),
+        """A simple boolean 'true'""",
+    ),
+    TestCase(
+        "boolean/invalid",
+        doc(elem("bool", Tag.Boolean, b"\xc3")),
+        """
+        In invalid boolean octet. Must be '0' or '1', but is 0xc3.
```
kevinAlbs:
```suggestion
        An invalid boolean octet. Must be '0' or '1', but is 0xc3.
```

src/libbson/tests/test-bson.c
```diff
@@ -854,272 +857,60 @@ test_bson_append_deep (void)
...
+{
+   // Needed size: 5 bytes for doc header/trailer, 2 bytes for each tag and empty
+   // key, minus 2 because the outer document has no tag and key
+   const size_t buffer_size = (depth * (5 + 2)) - 2;
+   uint8_t *const buffer = calloc (buffer_size, 1);
+   mlib_check (buffer);
+   uint8_t *out = buffer;
+   mlib_foreach_urange (i, depth) {
+      // Bytes we have already written:
+      const size_t begin_offset = out - buffer;
```
eramongodb:
```suggestion
      const size_t begin_offset = (size_t) (out - buffer);
```
`ptrdiff_t` -> `size_t` to address `-Wsign-conversion` warnings.

src/libmongoc/CMakeLists.txt
kevinAlbs:
```suggestion
```

Remove duplicate.

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+   mlib_check (!okay);
+   return false;
+}
+
+static bool
+_validate_doc (validator *self, const bson_t *bson, int depth)
+{
+   // We increment the depth here, otherwise we'd have `depth + 1` in several places
+   ++depth;
+   require_with_error (depth < 1000, 0, BSON_VALIDATE_CORRUPT, "BSON document nesting depth is too deep");
```
eramongodb:
```suggestion
   // https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Nested-Depth-for-BSON-Documents
   require_with_error (depth <= 100, 0, BSON_VALIDATE_CORRUPT, "BSON document nesting depth is too deep");
```
I think we can reduce this to 100 (inclusive):

> MongoDB supports no more than 100 levels of nesting for BSON documents. Each object or array adds a level.

It may also be worth defining an internal constant (e.g. `BSON_PRIVATE_MAX_DEPTH`) in `validate-private.h` for reuse by relevant test code (help avoid a lot of magic numbers).

src/libbson/tests/validate-tests.py
```diff
@@ -0,0 +1,1376 @@
+"""
+This script generates a C source file containing test cases for BSON validation
+and iteration.
+
+Run this script with Python 3.12+, and pipe the output into a file.
+
+This script takes no command-line arguments.
+"""
+
+import argparse
```
eramongodb:
```suggestion
"""

# /// script
# requires-python = ">=3.12"
# dependencies = []
# ///

import argparse
```

Add [inline script metadata](https://packaging.python.org/en/latest/specifications/inline-script-metadata/), which allows `uv run validate-tests.py` [without requiring `--no-project`](https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies) and documents (+ optionally enforces) the Python 3.12+ requirement mentioned above.

Example (only relevant lines included):

```
$ uv run -p 3.12 src/libbson/tests/validate-tests.py >/dev/null
(success, no diagnostic output)

$ uv run -v -p 3.12 src/libbson/tests/validate-tests.py >/dev/null
DEBUG uv 0.7.10
DEBUG Reading inline script metadata from `src/libbson/tests/validate-tests.py`
DEBUG Using Python request Python 3.12 from explicit request
DEBUG The script environment's Python version satisfies the request: `Python 3.12`
DEBUG No requirements to install
DEBUG Using Python 3.12.7 interpreter at: [...]
DEBUG Running `python src/libbson/tests/validate-tests.py`
DEBUG Command exited with code: 0

$ uv run -p 3.11 src/libbson/tests/validate-tests.py >/dev/null
warning: The requested interpreter resolved to Python 3.11.10, which is incompatible with the script's Python requirement: `>=3.12`
  File "src/libbson/tests/validate-tests.py", line 52
    type _ByteIter = bytes | Iterable[_ByteIter]
         ^^^^^^^^^
SyntaxError: invalid syntax

$ uv run -v -p 3.11 src/libbson/tests/validate-tests.py >/dev/null
DEBUG uv 0.7.10
DEBUG Reading inline script metadata from `src/libbson/tests/validate-tests.py`
DEBUG Using Python request Python 3.11 from explicit request
DEBUG The script environment's Python version satisfies the request: `Python 3.11`
DEBUG The script environment's Python version does not meet the Python requirement: `>=3.12`
warning: The requested interpreter resolved to Python 3.11.10, which is incompatible with the script's Python requirement: `>=3.12`
DEBUG No requirements to install
DEBUG Using Python 3.11.10 interpreter at: [...]
DEBUG Running `python src/libbson/tests/validate-tests.py`
  File "src/libbson/tests/validate-tests.py", line 52
    type _ByteIter = bytes | Iterable[_ByteIter]
         ^^^^^^^^^
SyntaxError: invalid syntax
DEBUG Command exited with code: 1
```

src/libbson/tests/validate-tests.py
```diff
@@ -0,0 +1,1376 @@
...
+        # The error object to be filled
+        "  bson_error_t error = {0};\n",
+        # The error offset. Expected to be reset to zero on success.
+        "  size_t offset = 999999;\n"
+        # Do the actual validation:
+        f"  const bool is_valid = bson_validate_with_error_and_offset(&doc, {case.flags}, &offset, &error);\n",
+    ]
+    is_error = case.error.code != "0"
+    yield from [
+        f"  mlib_check({'!is_valid' if is_error else 'is_valid'});\n",
```
kevinAlbs:
```suggestion
        "  mlib_check(!is_valid);\n" if is_error else "  ASSERT_OR_PRINT (is_valid, error);\n",
```

Suggest using `ASSERT_OR_PRINT` to print the error on failure. When forcing a failure:

Before:
```
test-validate.generated.c:64: in [_test_case_invalid_type]: Check condition <is_valid> failed
```

After:
```
test-validate.generated.c:64  _test_case_invalid_type()
  is_valid
  corrupt BSON
```

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+         return _validate_dollar_doc (self, &iter, depth);
+      }
+   }
+
+   return _validate_remaining_elements (self, &iter, depth);
+}
+
+// This private function is called by `bson_validate_with_error_and_offset`
+bool
+_bson_validate_impl_v2 (const bson_t *bson, bson_validate_flags_t flags, size_t *offset, bson_error_t *error)
```
eramongodb, line 540: Recommend adding a `validate-private.h` containing the declaration of this function (to address `-Wmissing-prototypes` Clang warnings), avoid the explicit non-defining declaration in `validate.c`, and better document by component structure (rather than by a comment) that `bson.c` uses `validate.c` via `#include <bson/validate-private.h>`.

src/libbson/src/bson/bson.c
```diff
@@ -2705,29 +2492,26 @@ bson_validate_with_error (const bson_t *bson, bson_validate_flags_t flags, bson_
...
-
-   if (state.err_offset >= 0) {
-      if (offset) {
-         *offset = (size_t) state.err_offset;
-      }
-      if (error) {
-         memcpy (error, &state.error, sizeof *error);
-      }
-      return false;
+   size_t offset_local = 0;
```
eramongodb, line 2506:
```suggestion
   BSON_ASSERT_PARAM (bson);
   BSON_OPTIONAL_PARAM (offset);
   BSON_OPTIONAL_PARAM (error);

   size_t offset_local = 0;
```
Missing precondition assertions.

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+{
```
eramongodb, line 541:
```suggestion
{
   BSON_ASSERT_PARAM (bson);
   BSON_ASSERT_PARAM (offset);
   BSON_ASSERT_PARAM (error);

```
Missing precondition assertions.

src/libbson/src/bson/bson-types.h
eramongodb:
```suggestion
   BSON_VALIDATE_CORRUPT = (1 << 5),
```

I am slightly concerned this might change the size of `bson_validate_flags_t` on _some_ target where the implementation-defined underlying integral type is chosen to be compatible with `char` by default. Perhaps `1 << 5` or `1 << 6` is a safer choice (remain within range of `signed char`)?

We can disregard this concern if all our currently supported targets use at least `(unsigned|signed) int` by default, but I haven't thoroughly verified this yet (note: the `abi-compliance-check` task only runs on Ubuntu 20.04).

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+ *
+ * @param Condition The condition that should evaluate to `true`
+ * @param Offset The byte offset where an error should be indicated.
+ * @param Code The error code that should be set if the condition fails
+ * @param ... The error string and format arguments to be used in the error message
+ *
+ * This macro assumes a `validator* self` is in scope. This macro will evaluate `return false`
+ * if the given condition is not true.
+ */
+#define require_with_error(Condition, Offset, Code, ...)                    \
```
eramongodb, line 115: Although we do not yet use pragma push/pop in the C Driver (except [this one case](https://github.com/mongodb/mongo-c-driver/blob/190fe9f48c7c82887c73d5577ba723cd633bed98/src/libmongoc/tests/test-mongoc-log.c#L79-L85)???), it may nevertheless be worthwhile to at least `#undef` these macros prior to use given how simple their identifiers are (likelihood of conflicting with global user-defined macros, e.g. `CMAKE_C_FLAGS` + `add_subdirectory(mongo-c-driver)` 😢).

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+   bson_t scope;
+   require_with_error (
+      bson_init_static (&scope, doc, doc_len), iter->off, BSON_VALIDATE_CORRUPT, "corrupt scope document");
+
+   // Validate the code string
+   require (_maybe_validate_utf8 (self, iter->off, u8, u8len));
+
+   // Now we validate the scope object.
+   // Don't validate the scope document using the parent parameters, because it should
+   // be treated as an opaque closure of JS variables.
```
kevinAlbs, line 277: The stricter validation for "code with scope" appears new. This seems to match the [spec expectations](https://bsonspec.org/spec.html):

> The string is JavaScript code. The document is a mapping from identifiers to values

"code with scope" is deprecated. I expect (hope) there is very little risk in users relying on the more relaxed validation.

Suggest noting the behavior changes in NEWS to help inform:

```
- Fix validation in `bson_validate` and related functions:
    - Invalid UTF-8 is now correctly rejected when `BSON_VALIDATE_UTF8` is passed.
    - The scope of a "code with scope" element is more strictly validated to match Javascript identifiers.
    - A nesting limit is added.
```

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+static bool
+_validate_doc (validator *self, const bson_t *bson, int depth)
+{
+   // We increment the depth here, otherwise we'd have `depth + 1` in several places
+   ++depth;
+   require_with_error (depth < 1000, 0, BSON_VALIDATE_CORRUPT, "BSON document nesting depth is too deep");
+
+   // Initialize an iterator into the document to be validated
+   bson_iter_t iter;
+   require_with_error (bson_iter_init (&iter, bson), 0, BSON_VALIDATE_CORRUPT, "Unable to initialize iterator");
```
eramongodb: Given this is a user-facing error message (user: "what iterator?"), perhaps a more straightforward message like "failed to satisfy minimal requirements for BSON binary data" or simply "corrupt BSON" might be preferable here.

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+   }
+   // Have the element key validator issue an error message about the bad $-key
+   bool okay = _validate_element_key (self, iter);
+   mlib_check (!okay);
+   return false;
+}
+
+static bool
+_validate_doc (validator *self, const bson_t *bson, int depth)
+{
```
eramongodb, line 507:
```suggestion
{
   BSON_ASSERT_PARAM (self);
   BSON_ASSERT_PARAM (bson);

```
Missing precondition assertions.

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+   }
+   // All subsequent elements should be validated as normal, and we don't expect
+   // any more $-keys
+   return _validate_remaining_elements (self, iter, depth);
+}
+
+// If we are validating special $-keys, validate a document whose first element is a $-key
+static bool
+_validate_dollar_doc (validator *self, bson_iter_t *iter, int depth)
+{
```
eramongodb, line 493:
```suggestion
{
   BSON_ASSERT_PARAM (self);
   BSON_ASSERT_PARAM (iter);
```

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+   if (_key_is (iter, "$ref")) {
+      return _validate_dbref (self, iter, depth);
+   }
```
eramongodb, line 498: For future reference: this exclusion for `$ref` (DBRef) seems to goes back to https://github.com/mongodb/libbson/commit/947c95982fac21e7affc94c77096be59d182c022 (Feb 2017) and https://github.com/mongodb/libbson/commit/4a71f87cb691beb200c4bd3e119c7a9190d1d353 (May 2017), probably motivated by compatibility with the new BSON corpus tests. See also [this warning](https://www.mongodb.com/docs/manual/core/dot-dollar-considerations/dollar-prefix/#std-label-dollar-prefix-field-names) (in particular, the last sentence):

> [MongoDB Extended JSON v2](https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/#std-label-extended-json-high-level-ref-v2) cannot differentiate between type wrappers and fields that happen to have the same name as type wrappers. Do not use Extended JSON formats in contexts where the corresponding BSON representations might include dollar ($) prefixed keys. **The [DBRef](https://www.mongodb.com/docs/manual/reference/database-references/#std-label-dbref-explanation) mechanism is an exception to this general rule.**

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+         /* The iterator stopped because of a decoding error */                                         \
+         require_with_error (false, (IteratorPointer)->err_off, BSON_VALIDATE_CORRUPT, "corrupt BSON"); \
+      }                                                                                                 \
+   } else                                                                                               \
+      ((void) 0)
+
+// Test if the element's key is equal to the given string
+static bool
+_key_is (bson_iter_t const *iter, const char *const key)
+{
```
eramongodb, line 157:
```suggestion
{
   BSON_ASSERT_PARAM (iter);
   BSON_ASSERT_PARAM (key);
```

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+      require (_validate_element (self, iter, depth));
+      require_advance (done, iter);
+   }
+   return true;
+}
+
+// Do validation for a DBRef document, indicated by a leading $ref key
+static bool
+_validate_dbref (validator *self, bson_iter_t *iter, int depth)
+{
```
eramongodb, line 445:
```suggestion
{
   BSON_ASSERT_PARAM (self);
   BSON_ASSERT_PARAM (iter);

```

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+   uint32_t len = mlib_read_u32le (iter->raw + iter->d1);
+   uint8_t const *data = (uint8_t const *) iter->raw + iter->d1;
+   return bson_init_static (subdoc, data, len);
+}
+
+// Validate the value of an element, without checking its key
+static bool
+_validate_element_value (validator *self, bson_iter_t const *iter, int depth)
+{
```
eramongodb, line 357:
```suggestion
{
   BSON_ASSERT_PARAM (self);
   BSON_ASSERT_PARAM (iter);

```

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+
+/**
+ * @brief Validate a string-like element (UTF-8, Symbol, or Code)
+ *
+ * This function relies on the representation of the text-like elements within
+ * the iterator struct to reduce code dup around text validation.
+ */
+static bool
+_validate_stringlike_element (validator *self, bson_iter_t const *iter)
+{
```
eramongodb, line 233:
```suggestion
{
   BSON_ASSERT_PARAM (self);
   BSON_ASSERT_PARAM (iter);
```

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+ * @param offset The byte-offset of the string, used to set the error offset
+ * @param u8 Pointer to the first byte in a UTF-8 string
+ * @param u8len The length of the array pointed-to by `u8`
+ * @return true If the UTF-8 string is valid, or if UTF-8 validation is disabled
+ * @return false If UTF-8 validation is requested, AND (the UTF-8 string is invalid OR (UTF-8 strings should not contain
+ * null characters and the UTf-8 string contains a null character))
+ */
+static bool
+_maybe_validate_utf8 (validator *self, size_t offset, const char *u8, size_t u8len)
+{
```
eramongodb, line 188:
```suggestion
{
   BSON_ASSERT_PARAM (self);
   BSON_ASSERT_PARAM (u8);
```

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+   case BSON_TYPE_ARRAY: {
+      bson_t doc;
+      require_with_error (_get_subdocument (&doc, iter), iter->off, BSON_VALIDATE_CORRUPT, "corrupt BSON");
+      if (_validate_doc (self, &doc, depth)) {
+         // No error
+         return true;
+      }
+      // Error in subdocument. Adjust the error offset for the current iterator position,
+      // plus the key length, plus 2 for the tag and key's null terminator.
+      self->error_offset += iter->off + strlen (bson_iter_key (iter)) + 2;
```
eramongodb:
```suggestion
      self->error_offset += iter->off + bson_iter_key_len (iter) + 2;
```
Avoid redundant `strlen()`.

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+   // Check that $ref is a UTF-8 element
+   require_with_error (
+      BSON_ITER_HOLDS_UTF8 (iter), iter->off, BSON_VALIDATE_DOLLAR_KEYS, "$ref element must be a UTF-8 element");
+   require (_validate_element_value (self, iter, depth));
+
+   // We require an $id as the next element
+   bool done;
+   require_advance (done, iter);
+   require_with_error (
+      !done && _key_is (iter, "$id"), iter->off, BSON_VALIDATE_DOLLAR_KEYS, "Expected an $id element following $ref");
```
eramongodb, line 460: Update: suggest adding comment documenting that `$id` may have any type, therefore `BSON_ITER_HOLDS_*` is deliberately omitted for this case.

Original comment below:

---

```suggestion
   require_with_error (!done && _key_is (iter, "$id") && BSON_ITER_HOLDS_OID (iter),
                       iter->off,
                       BSON_VALIDATE_DOLLAR_KEYS,
                       "Expected an $id element following $ref");
```
Should the type of `"$id"` be validated as an ObjectID? Every other `_key_is` is followed by a `BSON_ITER_HOLDS_*`. Per [docs](https://www.mongodb.com/docs/manual/reference/database-references/#format):

> `$id`: The `$id` field contains the value of the `_id` field in the referenced document.

and [docs](https://www.mongodb.com/docs/manual/core/document/#document-structure):

> `_id` holds an [ObjectId.](https://www.mongodb.com/docs/manual/reference/bson-types/#std-label-objectid)

If so, `validate-tests.py` needs to be updated to always specify `$id` as an ObjectID, e.g. `dbref/non-string-db` currently uses `utf8elem("$id", "bar")`.

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+                   scope_validator.error.code,
+                   "Error in scope document for element \"%s\": %s",
+                   bson_iter_key (iter),
+                   scope_validator.error.message);
+   // Adjust the error offset by the offset of the iterator
+   self->error_offset = scope_validator.error_offset + iter->off;
+   return false;
+}
+
+// Validate an element's key string according to the valiation rules
```
eramongodb:
```suggestion
// Validate an element's key string according to the validation rules
```

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+static bool
+_validate_element_key (validator *self, bson_iter_t const *iter)
+{
```
eramongodb, line 312:
```suggestion
{
   BSON_ASSERT_PARAM (self);
   BSON_ASSERT_PARAM (iter);

```
etc. for other functions.

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+   const char *const key = bson_iter_key (iter);
+   mlib_check (key);
+
+   // Check the UTF-8 of the key
+   require (_maybe_validate_utf8_cstring (self, iter->off, key));
```
eramongodb:
```suggestion

   uint32_t const key_len = bson_iter_key_len (iter);

   // Check the UTF-8 of the key
   require (_maybe_validate_utf8 (self, iter->off, key, key_len));
```
Avoids redundant `strlen()` here + below for `allow_empty_keys`.

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+      ((void) 0)
+
+/**
+ * @brief Advance the pointed-to iterator, check for errors, and test whether we are done.
+ *
+ * @param DoneVar An l-value of type `bool` that is set to `true` if the iterator hit the end of
+ * the document, otherwise `false`
+ * @param IteratorPointer An expression of type `bson_iter_t*`, which will be advanced.
+ *
+ * If advancing the iterator resulting in a decoding error, then this macro sets an error
```
kevinAlbs:
```suggestion
 * If advancing the iterator results in a decoding error, then this macro sets an error
```

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+/**
+ * @brief Validate a UTF-8 string, if-and-only-if UTF-8 validation is requested
+ *
+ * @param self Pointer to the validator object
+ * @param offset The byte-offset of the string, used to set the error offset
+ * @param u8 Pointer to the first byte in a UTF-8 string
+ * @param u8len The length of the array pointed-to by `u8`
+ * @return true If the UTF-8 string is valid, or if UTF-8 validation is disabled
+ * @return false If UTF-8 validation is requested, AND (the UTF-8 string is invalid OR (UTF-8 strings should not contain
+ * null characters and the UTf-8 string contains a null character))
```
kevinAlbs:
```suggestion
 * null characters and the UTF-8 string contains a null character))
```

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+/**
+ * @brief Validate a string-like element (UTF-8, Symbol, or Code)
+ *
+ * This function relies on the representation of the text-like elements within
+ * the iterator struct to reduce code dup around text validation.
+ */
+static bool
+_validate_stringlike_element (validator *self, bson_iter_t const *iter)
+{
+   // iter->d1 is the offset to the string header. Subtract 1 to exclude the null termintaor
```
kevinAlbs:
```suggestion
   // iter->d1 is the offset to the string header. Subtract 1 to exclude the null terminator
```

src/libbson/src/bson/validate.c
kevinAlbs:
```suggestion
// Validate an element's key string according to the validation rules
```

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+   const char *const key = bson_iter_key (iter);
+   mlib_check (key);
+
+   // Check the UTF-8 of the key
+   require (_maybe_validate_utf8_cstring (self, iter->off, key));
+
+   // Check for special keys
+   if (self->params->check_special_dollar_keys) {
+      // dollar-keys are checked during the startup of _validate_doc. If we get here, there's a problem.
+      require_with_error (key[0] != '$', iter->off, BSON_VALIDATE_DOLLAR_KEYS, "Disallowed element key: \"%s\"", key);
```
kevinAlbs:
```suggestion
      require_with_error (key[0] != '$', iter->off, BSON_VALIDATE_DOLLAR_KEYS, "Disallowed '$' in element key: \"%s\"", key);
```

Suggest including offending character in error.

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+   }
+
+   if (!self->params->allow_empty_keys) {
+      require_with_error (
+         strlen (key) != 0, iter->off, BSON_VALIDATE_EMPTY_KEYS, "Element key cannot be an empty string");
+   }
+
+   if (!self->params->allow_dot_in_keys) {
+      require_with_error (!strstr (key, "."), iter->off, BSON_VALIDATE_DOT_KEYS, "Disallowed element key: \"%s\"", key);
```
kevinAlbs:
```suggestion
      require_with_error (!strstr (key, "."), iter->off, BSON_VALIDATE_DOT_KEYS, "Disallowed '.' in element key: \"%s\"", key);
```

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,509 @@
...
+    */
+   bool check_special_dollar_keys;
+} validation_params;
+
+/**
+ * @brief State for a validator.
+ */
+typedef struct {
+   /// The parameters that control validation behavior
+   const validation_params *params;
```
eramongodb, line 86: Why `const validation_params *parmas;` instead of `validation_params params;`?

eramongodb (changes requested):

src/libbson/src/bson/validate.c
kevinAlbs: > Should the type of `"$id"` be validated as an ObjectID? 

I think "no". `_id` can be any BSON type (ObjectID is just the default if a document does not specify `_id`). Example:
```bash
$ mongosh --eval "db.coll.insertOne({'_id': 'foo'})"
{ acknowledged: true, insertedId: 'foo' }
```

eramongodb: Thank you. Then I'd like to adjust my suggestion here to include a comment documenting why there is deliberately no `BSON_ITER_HOLDS_*` for `$id`, unlike the other cases within this function.

src/libbson/src/bson/bson-types.h
vector-of-bool: That did cross my mind. `enum` pain.

src/libbson/src/bson/validate.c
vector-of-bool: I considered both, and chose this shape purely to prohibit attempting to modify the param values. Either way would be fine, though.

src/libbson/src/bson/validate.c
vector-of-bool: `bson_iter_next` checks against the zero tag byte. If execution reaches this point, then some invariants are violated somewhere else and I don't feel like it would be safe to continue.

src/libbson/src/bson/validate.c
vector-of-bool: I added changes that relax the validation rules further. I realized that we don't want reject a scope like

```js
{
    foo: {
        'dot.key': 42
    }
}
```

becuase `dot.key` is a valid object key. We *could* do validation of the outermost keys to match identifier rules, but that would be a lot more work to do it properly and would require more testing. I don't think it's worth it for this deprecated BSON feature when a JS runtime will do the validation for anyone that really needs to use it.

vector-of-bool:
Accidental rebase plz ignore.

I believe I've address all outstanding change requests, but may have missed some as the GitHub UI has been very uncooperative this week.

src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,568 @@
...
+}
+
+static bool
+_validate_doc (validator *self, const bson_t *bson, int depth)
+{
+   BSON_ASSERT_PARAM (self);
+   BSON_ASSERT_PARAM (bson);
+
+   // The depth limit of 100 is chosen to match the limit enforced by MongoDB server.
+   // Refer: https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Nested-Depth-for-BSON-Documen
```
eramongodb:
```suggestion
   // Refer: https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Nested-Depth-for-BSON-Documents
```
End of URL is clipped.

src/libbson/src/bson/validate-private.h
```diff
@@ -0,0 +1,22 @@
...
+ * is now defined in its own file.
+ *
+ * @param bson The document to validate. Must be non-null.
+ * @param flags Validation control flags
+ * @param offset Receives the offset at which validation failed. Must be non-null.
+ * @param error Receives the error describing why validation failed. Must be non-null.
+ * @return true If the given document has no validation errors
+ * @return false Otherwise
+ */
+extern bool
```
eramongodb:
```suggestion
bool
```
`extern` should (now) be unnecessary.

src/libbson/src/bson/bson-types.h
```diff
@@ -185,27 +185,68 @@ typedef struct {
...
+enum {
+   /**
+    * @brief This compile-time constant represents the maximum document nesting
+    * depth permitted by the `bson_validate` family of functions. If the nesting
+    * depth exceeds this limit, the data will be rejected.
+    *
+    * The depth limit of 100 is chosen to match the limit enforced by MongoDB server.
+    * Refer: https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Nested-Depth-for-BSON-Document
+    */
+   BSON_VALIDATION_MAX_NESTING_DEPTH = 100,
```
eramongodb:
```suggestion
   BSON_VALIDATE_MAX_NESTING_DEPTH = 100,
```

All other validation-related constants use a prefix of `BSON_VALIDATE_*`.

Do we really need to expose this in the public API? Can it be moved to `<bson/validate-private.h>` instead? Test code is able to include private headers.

eramongodb (approved):

src/libbson/src/bson/bson-types.h
vector-of-bool: I put it in the public header because I think it might be useful to users to know the limit. Because it's public, I intentionally chose the differing name to prevent a user from thinking that this is meaningfull:

```c
bson_validate(d, BSON_VALIDATE_UTF8 | BSON_VALIDATE_MAX_NESTING_DEPTH, NULL);
```

It may actually be better to have a function `bson_get_max_nesting_depth()`, in case it is changed in the future and may vary depending on what library they load at runtime?

eramongodb: Thank you for the explanation.

> It might be useful to users to know the limit.

I'm not sure about that, but in that case, perhaps `BSON_MAX_NESTING_DEPTH` via `bson/bson.h` (alongside `BSON_MAX_SIZE`) might be better, since it is not a limit that is restricted to validation only? (Apparently we have `#define BCON_STACK_MAX 100` in `bson/bcon.h`, but that doesn't seem to be meant to be used outside of the component.)

> It may actually be better to have a function `bson_get_max_nesting_depth()`

I do not expect the limit to change anytime soon, so even if we expose it in public API, I do not mind leaving it as a compile-time constant.

PR #2031, opened 2025-06-06
[CDRIVER-6017] BSON Validation Refactor (#2026) (Cherry-pick for 1.30.x)

vector-of-bool:
This is a cherry-pick of changes from #2026 for the 1.30 release.

Because the new code makes use of files in the `mlib/` directory, those files have been copied from `master` into this PR, as well as a `.clang-format` file to keep their formatting since the formatting has diverged since. The code in this directory is not used anywhere except as part of the new validation code and its relevant test cases.

The original commit message is retained below and should be included in the final merge.

---

* New BSON validation routine rewrite

The new `bson_validate` implementation does not
make use of the error-prone `bson_visit` APIs. Instead, it is written as a simple recursive validator. The new validator respects requests for UTF-8 validation properly.

* Stop validating at 1000 depth, preventing stack overflow
* Replace most BSON validation tests with generated ones

The existing test cases used BSON files, and didn't have any commentary on what they were actually testing. New test cases are generated from a Python shorthand and contain the tested bytes inline, with a distinct test case for each actual validation scenario.

* Disable UTF-8 validation by default on CRUD APIs
* Document and tweak the value of BSON_VALIDATE_CORRUPT
* Add test cases related to the overlong null encoding
* Tweak JS scope validation to permit more obj keys
* Add a NEWS entry for validation changes.
* Allow `-private.h` headers to not include the prelude header

---------

### DIFF
--- a/src/libbson/NEWS
+++ b/src/libbson/NEWS
@@ -1,3 +1,28 @@
+Unreleased
+==========
+
+Fixes:
+
+* Various fixes have been applied to the `bson_validate` family of functions,
+  with some minor behavioral changes.
+  * Previously accepted invalid UTF-8 will be rejected when `BSON_VALIDATE_UTF8`
+    is specified.
+  * The scope document in a deprecated "code with scope" element is now
+    validated with a fixed set of rules and is treated as an opaque JavaScript
+    object.
+  * A document nesting limit is now enforced during validation.
+
+libbson 2.0.1
+=============
+
+Fixes:
+
+  * Fix CMake configure error on macOS with CMake 4.
+
+Thanks to everyone who contributed to the development of this release.
+
+  * Kevin Albertson
+
 libbson 1.30.4
 ==============
 
--- a/src/libbson/src/bson/validate.c
+++ b/src/libbson/src/bson/validate.c
@@ -1,0 +1,566 @@
+/**
+ * @file bson/validate.c
+ * @brief Implementation of BSON document validation
+ * @date 2025-05-28
+ *
+ * This file implements the backend for the `bson_validate` family of functions.
+ *
+ * The `_validate_...` functions all accept `validator* self` as their first parameter,
+ * and must `return false` AND set `self->error` if-and-only-if they encounter a validation error.
+ * If a function returns true, it is assumed that validation of that item succeeded.
+ *
+ * For brevity, the `require...` macros are defined, which check conditions, set errors,
+ * and `return false` inline.
+ *
+ * @copyright Copyright 2009-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <bson/validate-private.h>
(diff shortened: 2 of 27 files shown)


src/libbson/src/bson/validate.c
```diff
@@ -0,0 +1,566 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <bson/validate-private.h>
+#include <mlib/intencode.h>
```
kevinAlbs: At the risk of having diverging changes: can mlib be removed to reduce the size of the backport? The release branches are intended to be stable. I am mildly concerned mlib could introduce some unexpected compile change.

If tests are removed I expect all that is needed is to replace `mlib_read_u32le` with `   BSON_UINT32_FROM_LE` and `mlib_check` with `BSON_ASSERT`.

src/libbson/NEWS
```diff
@@ -1,3 +1,28 @@
...
+=============
+
+Fixes:
+
+  * Fix CMake configure error on macOS with CMake 4.
+
+Thanks to everyone who contributed to the development of this release.
+
+  * Kevin Albertson
+
```
eramongodb:
```suggestion
```
Stray entries specific to 2.0.1?

src/libbson/tests/validate-tests.py
```diff
@@ -0,0 +1,1479 @@
...
+            subsequent_indent=" " * 4,
+            initial_indent=" " * 4,
+            width=80,
+        )
+    )
+    yield "\n  };\n"
+    yield from [
+        # Initialize a BSON doc that points to the byte array
+        "  bson_t doc;\n",
+        "  BSON_ASSERT(bson_init_static(&doc, bytes, sizeof bytes));\n",
```
eramongodb:
```suggestion
        "  ASSERT(bson_init_static(&doc, bytes, sizeof bytes));\n",
```
Prefer test suite assertions.

src/libbson/tests/validate-tests.py
```diff
@@ -0,0 +1,1479 @@
...
+        "  size_t offset = 999999;\n"
+        # Do the actual validation:
+        f"  const bool is_valid = bson_validate_with_error_and_offset(&doc, {case.flags}, &offset, &error);\n",
+    ]
+    is_error = case.error.code != "0"
+    yield from [
+        "  BSON_ASSERT(!is_valid);\n" if is_error else "  ASSERT_OR_PRINT(is_valid, error);\n",
+        f"  BSON_ASSERT(error.code == {case.error.code});\n",
+        f"  BSON_ASSERT(!strcmp(error.message, {json.dumps(case.error.message)}));\n",
+        f"  BSON_ASSERT(offset == {case.error.offset});\n" if is_error else "",
```
eramongodb:
```suggestion
        "  ASSERT(!is_valid);\n" if is_error else "  ASSERT_OR_PRINT(is_valid, error);\n",
        f"  ASSERT_CMPINT32(error.code, ==, {case.error.code});\n",
        f"  ASSERT_CMPSTR(error.message, {json.dumps(case.error.message)});\n",
        f"  ASSERT_CMPSIZE_T(offset, ==, {case.error.offset});\n" if is_error else "",
```
Prefer test suite assertions.

src/libbson/NEWS
vector-of-bool: ~~I believe this waits for the release process, maybe? @kevinAlbs~~

I misread. Plz ignore

faridguitouni00-debug:
Faridguitouni00@gmail.com

faridguitouni00-debug:
changements programmation YouTube profils il limitée provisionnelle avec IA complet dernier 2025

PR #2047, opened 2025-06-26
[CDRIVER-5983] Refactor String Handling Around URI Parsing

vector-of-bool:
# Background

Refer: CDRIVER-5983

This PR is a subset of the changes that were original slated for CDRIVER-5983. The URI parameter refactor turned out to be significantly more impractical than initially though, so the changes related to parameter handling have been deferred until the relevant spec documents can consolidate on a cohesive behavior.

Regardless, some of the changes from that work can still be extracted and is useful in general throughout the codebase, specifically related to string handling and parsing.

This PR introduces minor behavioral changes surrounding some edge cases related to parsing.

While initially the commit history was going to be clean for the overall changes, the partial rollback around parameter handling means that commits themselves are not entirely useful in isolation.

# Change Summary

## String Views 

This PR introduces the `mlib/str.h` header. This header has been re-introduced several times in libmongocrypt and amongoc, and this is the latest incarnation, which I believe is the most solid as it can build upon existing work regarding safe arithmetic and integer utilities.

- `mstr_view` is a "string view" type. It consists of a `const char* data` pointer and a `size_t len`, wheren `len` refers to the length of the pointed-to string view.
- Importantly, the data in `mstr_view` NOT necessarily null terminated! This means that `mstr_view` requires new string handling routines that respect the length.
- Because we can pass slices of strings, this greatly reduces the need to `strndup` every time we want to take a string slice to pass to another function, further reducing the occurrences of memory leaking, `goto fail`, and buffer overruns. Code using `mtsr_view` is almost always far simpler than trying to juggle C strings.
- There are several string algorithms: `mstr_find`, `mstr_find_first_of`, `mlib_substr`, `mstr_contains`, which inspect and manipulate string views.
- `mstr_cmp` is a string comparison function like `mlib_cmp`, which supports natural syntax like `mstr_cmp(a, !=, b)`
- `mstr_split_at(S, Pos, [Drop,] *Pfx, *Sfx)` is the most useful algorithm. It takes an index and optional skip parameter, and produces two new string views from an input.
- `mstr_split_around(S, Infix, *Pfx, *Sfx) -> bool` builds upon `mstr_split_at`, and splits the string `S` around the first occurrence of `Infix`, returning `true` if it found `Infix`. This one API produces the greatest complexity reduction of all. For parsing a CSV string:

  ```c
  mstr_view remain = csv;
  while (remain.len) {
    mstr_view entry;
    mstr_split_around(remain, mlib_cstring(","), &entry, &remain);
    // Handle `entry`
  }
  ```

- `mstr_cstring` creates an `mstr_view` from a C string. This is unfortunately necessary for calling a lot of APIs, but may be unnecessary in the future (see the note on `_Generic` in `mlib/str.h`)
- All algorithms that accept a string position index also accept a negative value to index from the end of the string. They also use `upsize_integer` in order to inhibit any integer conversion warnings. No need to worry about passing the right type to these functions, it will just Do the Right Thing without any implicit conversions.
- To printf-format a sized string, one must use the `%.*s` printf-specifier. This expect an `int` to specify the string length, followed by the string pointer. Because this is common, the macro `MSTR_FMT(X)` will expand to the two arguments required for the `%.*s` specifier.

## Case Sensitivity

- `mlib_latin_tolower`: Transforms a Basic Latin codepoint to a lowercase letter if it is an uppercase letter. Unlike standard `tolower`, this function is total in 32-bit integers, is very clear about what "lowercasing" it does, is not sensitive to the locale, and has no undefined behavior.
- `mlib_latin_charcasecmp`: Compare two codepoints for equivalence with case-insensitivity in Basic Latin. Uses `mlib_latin_tolower`.
- `mstr_latin_casecmp`: Like `mstr_cmp`, but case-insensitive using `mlib_latin_charcasecmp`.

### `bson_error_clear`

This function takes a `bson_error_t*` and resets it to a non-error value if the pointer is not null.

### `bson_error_reset`

The new function-like macro `bson_error_reset` is defined. It is equivalent to `bson_error_clear` and takes an l-value to a pointer to `bson_error_t`. If the pointer is null, it is made to point to an anonymous `bson_error_t` in the local scope, allowing the function to rely on the present of an error object, even if the caller didn't pass one:

```c
bool foo(bson_error_t* err) {
  bson_error_reset(err);  // Clears any "err"
  // `err` is now guaranteed to be non-null
}
```

This feature relies on C99 compound literals, so this function-like macro cannot be used in C++ 😢.

### `bson/mongoc_set_error`

The `set_error` functions that do a string-format into the error message now support passing the error message of the object as an input parameter to the format string itself:

```c
bson_set_error(&error, Cat, Code, "Got an error while creating a widget: %s", error.message);
```

Previously, this would garble the content of `error.message`. Now, the code uses a temporary buffer to format the message, then copies it over the `error.message`.

## Integer Parsing 

Because we are now using sized strings and not C strings, we need new integer parsing, because `strtoll` and friends expect a null terminator. Additionally, we can do better than `strtoll`:

- `mlib_i64_parse(Str, Base, *Out) -> int` parses a 64-bit signed integer in `Str`. It differs from `strtoll` in a few important ways:
  - It does not skip leading whitespace. That's an error.
  - It doesn't immediately stop on a non-digit. That's an error.
  - Doesn't silently reject empty strings. That's an error.
  - It doesn't require a null terminator, of course.
- `mlib_i32_parse(Str, Base, *Out) -> int` `int32_t` version of `mlib_i64_parse`
- `mlib_nat64_parse(Str, Base, *Out) -> int` like `mlib_i64_parse`, but does not allow sign or base prefixes. `Base` must be non-zero. The input must be a string of only digits. This is the core of integer parsing.

Previousy, our code using `strtoll` was silently relying on whitespace trimming and had clunky or incomplete detecting whether `strtoll` stopped early. These parsing APIs are much more strict.

The implementation may be less efficient than the hyper-optimized `strtoll`, but it's much more easy to use correctly. Optimizations can come later.

## Error Messages

Several APIs have been modified to return better error messages, including APIs that can fail with `true/false`. For example, %-decoding now explains why %-string is invalid, and that will be included when a URI is rejected.

## Other Changes

The following other minor changes were made:

- Passing an empty string to `mongoc_uri_set_compressors` was previously an error. The new behavior is to clear the compressors on the URI (as if passing a null pointer).
- The failure strings in test assertions macros were modified so that VSCode generates proper clickable file links in terminal output.
- Invalid UTF-8 following %-decoding generated a descriptive log warning in addition to a vague error message. This is changed to just generate a descriptive error without logging.

# Not Included 

## Owning Sized-String `mstr`

There was going to be a type `mstr` that owned a sized mutable null-terminated string, with its own set of useful "Does the Right Thing" algorithms (append, splice, copy, erase, replace, push, etc.). I have the code for this, but it can wait for a future PR.

## UTF-8 Validation

It may be useful to force `mstr_view` to only contain UTF-8 validation (i.e. you cannot create an `mstr_view` without checking that you have valid UTF-8), but that's a larger change for the potential future.

## Logging/Errors/Warning Mess

The behavior around logging, warning, and erroring in URI handling is a mess. There are cases where we test that an API logs, and sometimes we expect it to be silent and put the error on an error object instead, and sometimes we expect both!

This inconsistency has been mostly retained. In the future, any API that accepts a `bson_error_t*` probably *shouldn't* log anything.


src/common/src/mlib/intencode.h
```diff
@@ -165,3 +168,184 @@ mlib_write_f64le (void *out, double d)
...
+ *
+ * @param in The string of digits to be parsed.
+ * @param base Optional: The base to use for parsing. Use "0" to infer the base.
+ * @param out Optional storage for an int64 value to be updated with the result
+ * @return int Returns an errno value for the parse
+ *
+ * - A value of `0` indicates that the parse was successful.
+ * - A value of `EINVAL` indicates that the input string is not a valid
+ *   representation of an integer.
+ * - A value of `ERANGE` indicates thath the input string is a valid integer,
```
kevinAlbs:
```suggestion
 * - A value of `ERANGE` indicates that the input string is a valid integer,
```

src/common/src/mlib/intencode.h
```diff
@@ -165,3 +168,184 @@ mlib_write_f64le (void *out, double d)
...
+      // Just a plus. Drop it and do nothing with it.
+      in = mstr_substr (in, 1);
+   }
+   // Check for a negative prefix
+   else if (in.data[0] == '-') {
+      // Negative sign. We'll negate the value later.
+      in = mstr_substr (in, 1);
+      sign = -1;
+   }
+
```
kevinAlbs, line 270: Check length again after possibly removing sign:

```c
if (in.len == 0) {
  // String with only sign is invalid
  return EINVAL;
}
```

Avoids a possible buffer-overflow for a non-NULL terminated "+".

src/common/src/mlib/str.h
```diff
@@ -0,0 +1,554 @@
...
+ * You should use the format specifier `%.*s' for all mstr strings.
+ *
+ * This is just a convenience shorthand.
+ */
+#define MSTR_FMT(S) (int) mstr_view_from (S).len, mstr_view_from (S).data
+
+/**
+ * @brief Create an `mstr_view` that views the given array of `char`
+ *
+ * @param data Pointer to the beginning of the
```
kevinAlbs:
```suggestion
 * @param data Pointer to the beginning of the array
```

src/common/src/mlib/str.h
```diff
@@ -0,0 +1,554 @@
...
+ * where `-1` refers to the last character in the string, `-2` the penultimate,
+ * etc. If the absolute value is greater than the length of the string, the
+ * program will be terminated.
+ */
+static inline size_t
+_mstr_adjust_index (mstr_view s, mlib_upsized_integer pos, bool clamp_to_length)
+{
+   if (clamp_to_length) {
+      if (pos.is_signed) {
+         if (pos.i.s > 0 && (size_t) pos.i.s > s.len) {
```
kevinAlbs: Suggest using `mlib_cmp` to handle the possibility that `pos.i.s` (intmax_t) be greater than `SIZE_MAX`. Return early to avoid the (unsafe?) cast of `(intmax_t) s.len`:

```c
   if (clamp_to_length) {
      if (pos.is_signed) {
         if (pos.i.s > 0 && mlib_cmp (pos.i.s, >, s.len)) {
            return s.len;
         }
      } else {
         if (pos.i.u > s.len) {
            return s.len;
         }
      }
   }
```

src/common/src/mlib/str.h
```diff
@@ -0,0 +1,554 @@
...
+ * must not be freed or manipulated.
+ *
+ * @note The viewed string is NOT guaranteed to be null-terminated. It WILL
+ * be null-terminated if: Directly created from a string literal, a C string, or
+ * a null-terminated `mstr_view`.
+ * @note The viewed string MAY contain nul (zero-value) characters, so using them
+ * with C string APIs could truncate unexpectedly.
+ */
+typedef struct mstr_view {
+   // Pointer to the first character in the string
```
kevinAlbs: Suggest clarifying if `NULL` is OK for `data`. E.g. is it OK to create an empty string like `mstr_view_data (NULL, 0);`?

src/common/src/mlib/str.h
```diff
@@ -0,0 +1,554 @@
...
+}
+
+#define mstr_find(...) MLIB_ARGC_PICK (_mstr_find, __VA_ARGS__)
+#define _mstr_find_argc_2(Hay, Needle) _mstr_find_argc_3 (Hay, Needle, 0)
+#define _mstr_find_argc_3(Hay, Needle, Start) _mstr_find_argc_4 (Hay, Needle, Start, SIZE_MAX)
+#define _mstr_find_argc_4(Hay, Needle, Start, Stop) \
+   mstr_find (mstr_view_from (Hay), mstr_view_from (Needle), mlib_upsize_integer (Start), Stop)
+
+/**
+ * @brief Find the zero-based index of the first `char` in `hay` that also occurrs in `needles`
```
kevinAlbs:
```suggestion
 * @brief Find the zero-based index of the first `char` in `hay` that also occurs in `needles`
```

src/common/tests/test-mlib.c
```diff
@@ -658,6 +713,148 @@ _test_ckdint_partial (void)
...
+      mlib_check (mstr_cmp (b, ==, mstr_cstring ("bar baz")));
+      // Split again
+      mlib_check (mstr_split_around (b, space, &a, &b));
+      mlib_check (mstr_cmp (a, ==, mstr_cstring ("bar")));
+      mlib_check (mstr_cmp (b, ==, mstr_cstring ("baz")));
+      // Split again. This won't find a space, but will still do something
+      mlib_check (!mstr_split_around (b, space, &a, &b));
+      mlib_check (mstr_cmp (a, ==, mstr_cstring ("baz")));
+      mlib_check (mstr_cmp (b, ==, mstr_cstring ("")));
+      // Splitting on the final empty string does nothign
```
kevinAlbs:
```suggestion
      // Splitting on the final empty string does nothing
```

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -207,190 +205,190 @@ mongoc_uri_remove_host (mongoc_uri_t *uri, const char *host, uint16_t port)
...
+         // Not a % char, just append it
+         *out++ = remain.data[0];
+         remain = mstr_substr (remain, 1);
+         continue;
+      }
+      // %-sequence
+      if (remain.len < 3) {
+         MONGOC_URI_ERROR (error,
+                           "At offset %llu: Truncated %%-sequence \"%.*s\"",
+                           (long long unsigned) (sv.len - remain.len),
```
kevinAlbs: I expect since `sv.len` and `remain.len` are both `size_t` this can be slightly simplified to:

```c
MONGOC_URI_ERROR (
   error, "At offset %zu: Truncated %%-sequence \"%.*s\"", sv.len - remain.len, MSTR_FMT (remain));
```

Similar suggestion applies to the next error check.

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -207,190 +205,190 @@ mongoc_uri_remove_host (mongoc_uri_t *uri, const char *host, uint16_t port)
...
+         error, "Password \"%.*s\" must not have unescaped chars. %s", MSTR_FMT (password), escape_instructions);
       return false;
    }
 
-   mongoc_uri_do_unescape (&uri->username);
+   // Store the username and password on the URI
+   uri->username = _strdup_pct_decode (username, error);
    if (!uri->username) {
-      MONGOC_URI_ERROR (error, "Incorrect URI escapes in username. %s", escape_instructions);
+      MONGOC_URI_ERROR (error, "Invalid username \"%.*s\" in URI string: %s", MSTR_FMT (username), error->message);
```
kevinAlbs:
```suggestion
      MONGOC_URI_ERROR (error, "Invalid username in URI string: %s", escape_instructions);
```

Avoid including credentials in error codes to avoid rising leaking credentials into error logs etc. (Auth commands are [redacted from logs](https://github.com/mongodb/specifications/blob/668992950d975d3163e538849dd20383a214fc37/source/command-logging-and-monitoring/command-logging-and-monitoring.md#security)).

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -207,190 +205,190 @@ mongoc_uri_remove_host (mongoc_uri_t *uri, const char *host, uint16_t port)
...
-            error, "Password \"%s\" must not have unescaped chars. %s", uri->password, escape_instructions);
-         return false;
-      }
-
-      mongoc_uri_do_unescape (&uri->password);
+   if (has_password) {
+      uri->password = _strdup_pct_decode (password, error);
       if (!uri->password) {
-         MONGOC_URI_ERROR (error, "%s", "Incorrect URI escapes in password");
+         MONGOC_URI_ERROR (error, "Invalid password \"%.*s\" in URI string: %s", MSTR_FMT (password), error->message);
```
kevinAlbs:
```suggestion
         MONGOC_URI_ERROR (error, "Invalid password in URI string: %s", escape_instructions);
```

Avoid including credentials in error codes to avoid rising leaking credentials into error logs etc. (Auth commands are [redacted from logs](https://github.com/mongodb/specifications/blob/668992950d975d3163e538849dd20383a214fc37/source/command-logging-and-monitoring/command-logging-and-monitoring.md#security)).

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -414,149 +412,104 @@ mongoc_uri_parse_srv (mongoc_uri_t *uri, const char *str, bson_error_t *error)
...
 
-   /* invalid characters in database name */
-   for (c = "/\\. \"$"; *c; c++) {
-      invalid_c = scan_to_unichar (uri->database, (bson_unichar_t) *c, "", &tmp);
-      if (invalid_c) {
-         bson_free (invalid_c);
-         return false;
-      }
+   // Check if the database name contains and invalid characters after the %-decode
+   if (mstr_contains_any_of (mstr_cstring (uri->database), mstr_cstring ("/\\. \n\r\f\t\"$"))) {
```
kevinAlbs:
```suggestion
   if (mstr_contains_any_of (mstr_cstring (uri->database), mstr_cstring ("/\\. \"$"))) {
```

Suggest reverting disallowed characters. I appear to be able to create a database name with "\n\r\f\t":
```python
from pymongo import MongoClient
client = MongoClient()
for letter in "/\\. \n\r\f\t\"$":
    dbname = f"foo{letter}bar"
    try:
        client[dbname].command({"ping": 1})
        print ("{} ... ok".format(repr(dbname)))
    except Exception:
        print ("{} ... failed".format(repr(dbname)))
```

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -630,108 +583,110 @@ mongoc_uri_check_srv_service_name (mongoc_uri_t *uri, const char *str)
...
- *
- *       NOTE: This function keeps the order of the BSON keys.
- *
- *       NOTE: 'option' is case*in*sensitive.
- *
+/**
+ * @brief Remove a BSON element with the given key, case-insensitive
  *
- *--------------------------------------------------------------------------
+ * @param doc The document to be updtaed
```
kevinAlbs:
```suggestion
 * @param doc The document to be updated
```

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -630,108 +583,110 @@ mongoc_uri_check_srv_service_name (mongoc_uri_t *uri, const char *str)
...
+   bson_destroy (&tmp);
+}
 
-         if (!strcasecmp (bson_iter_key (&iter), option)) {
-            bson_append_utf8 (&tmp, option, -1, value, -1);
-            found = true;
-            continue;
-         }
+/**
+ * @brief Update a BSON document with a UTF-8 value, replacing it if it alread
```
kevinAlbs:
```suggestion
 * @brief Update a BSON document with a UTF-8 value, replacing it if it already
```

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -1892,168 +1815,211 @@ mongoc_uri_finalize_directconnection (mongoc_uri_t *uri, bson_error_t *error)
...
-      // String spec: for backwards-compatibility reasons, drivers MAY allow reserved characters other than "@" and
-      // ":" to be present in user information without percent-encoding.
-      char *userinfo = scan_to_unichar (cursor, '@', "", &tmp);
-
-      if (userinfo) {
-         cursor = tmp + 1; // Consume userinfo delimiter.
-         bson_free (userinfo);
+      size_t userinfo_end_pos = mstr_find (remain, AT);
+      if (userinfo_end_pos == SIZE_MAX) {
+         // There is no userinfo, so we don't need to do anything speical
```
kevinAlbs:
```suggestion
         // There is no userinfo, so we don't need to do anything special
```

src/libmongoc/src/mongoc/mongoc-uri.c
```diff
@@ -1892,168 +1815,211 @@ mongoc_uri_finalize_directconnection (mongoc_uri_t *uri, bson_error_t *error)
...
-         cursor = tmp; // Include the delimiter.
-         bson_free (hostinfo);
-      }
+   // * remain = "/path?query#fragment" (Each following component is optional, but this is the proper order)
+   const size_t path_end_pos = mstr_find_first_of (remain, mstr_cstring ("?#"));
+   mstr_split_at (remain, path_end_pos, &parts->path, &remain);
+   // * remain = "query#fragment" (Each following component is optional, but this is the proper order)
+   const size_t hash_pos = mstr_find_first_of (remain, mstr_cstring ("#"));
+   mstr_split_at (remain, hash_pos, &parts->query, &remain);
+   // * remain = "#fragment"
```
kevinAlbs: Suggest removing special handling for `#fragment`. The connection string spec does not appear to mention a fragment. And this may trigger a backwards breaking behavior change. Example:

```c
// Previously OK. Now a parse error:
mongoc_uri_t *uri = mongoc_uri_new ("mongodb://host/?appName=bar#baz");
const char *got = mongoc_uri_get_option_as_utf8 (uri, "appName", "");
ASSERT_CMPSTR (got, "bar#baz");
mongoc_uri_destroy (uri);
```

kevinAlbs (approved):
LGTM with comments addressed. I really like these changes. I think `mstr_view` significantly simplifies parsing.

Aside: I ran a local fuzz test with sanitizers on `mongoc_uri_new`, but did not discover anything.

> Owning Sized-String `mstr`

If / when that is done, it may be worth combining with the existing `mcommon_string_t`.

src/common/src/mlib/str.h
vector-of-bool: Added clarification about null string views as an explicitly supported use case

src/common/src/mlib/str.h
vector-of-bool: Heavily simplified this section

src/common/src/mlib/intencode.h
vector-of-bool: Tweaked with a check around the logic around base parsing. Added test cases for this that will be triggered during ASan builds.

PR #2050, opened 2025-06-30
[CDRIVER-5670] Include-directive Sorting and Formatting

vector-of-bool:
Refer: CDRIVER-5670

This changeset reformats and sorts our `#include` directives automatically.

The `#include` sorting is done by `clang-format`. The header style formatting is done with a custom script.

# `#include` Sorting

There has been a longstanding desire to sort out `#include` directives automatically, reducing the need to manually manage them, and also having a unified ordering across the codebase.

One problem was that our header files are not sanitary: They often accidentally assume that they are included in a certain order relative to each other. This PR includes minor changes so that headers support the chosen ordering. It is still possible (and almost certain) that headers still assume certain `#include` patterns, but the header ordering chosen in this PR works. The only changes needed were in certain `private.h` headers to `#include` their non-private counterpart, and some files in `bson/` relied on certain orders because they cyclicly included themselves. This was simply enough to fix by `#include`-ing the specific headers they used rather than including the catch-all `bson.h` header first.

The header ordering is chosen in order of specificity:

1. First the `prelude.h` header, because it forces itself to come first (unfortunate)
2. Then, `#include` files that reach into the same directory as the containing file. These directives are also reformatted (see below)
3. Then, all `-private.h` headers
4. Then, all `mongoc/` headers 
5. All `bson/` headers
6. All `mlib/` and `common/` headers
7. Then come platform-specific headers (`windows.h`, `unistd.h`)
8. Finally, standard library headers are chosen

The ordering is chosen to minimize implicit dependencies going forward. It isn't fool-proof, but it is a start. A future change should instead use CMake's functionality for header-checking, which will finally allow us to ditch `prelude.h` and allow us and users to `#include` specific components directly.

# Header Formatting

Before headers are sorted, they are reformatted according to some simple rules:

1. If a header ID uses double quotes and the first quoted char is a dot "`.`", it is left as-is. This allows only `"./foo.h"` and `"../foo.h"` headers to use quotes.
2. If a header uses quotes:
   1. If the named file is reachable within the directory of the containing file, it will have `./` prepended to the path, and remain in double-quotes
   2. Otherwise, the quotes are replaced with angle-brackets.

The result is that all relative "in this directory" headers have a path that starts with a dot or dot-dot. This has no actual effect on the compiler's file-finding algorithm, but makes it clear to the reader where the files live.

All other headers use angle brackets. This inhibits the "look in this directory" behavior in the compiler, and signals to the reader that the header lives at the base of some `-I` search directory.

# Formatting Script

The `clang-format-all.sh` script is replaced with a `tools/format.py` script. Since we already rely on `uv` to run the formatting task and install the correct `clang-format` version, we can assume that we have `uv` (and Python) installed.

`format.py` implements the header reformatting, but also splits the formatting work into parallel tasks. This makes formatting significantly faster, sufficient to use as a pre-commit hook:

```sh
# file: .git/hooks/pre-commit
uv run --only-group=format tools/format.py --mode=check
```


.evergreen/scripts/clang-format-all.sh
eramongodb, line 1: Also update [this comment](https://github.com/mongodb/mongo-cxx-driver/blob/72f650157d12b1fffeeee6549e6b9e6e49798bf3/pyproject.toml#L26) in pyproject.toml:

```toml
format = [
    # tools/format.py
    "clang-format~=20.1.0",
]
```

tools/format.py
```diff
@@ -0,0 +1,280 @@
...
+            case patterns:
+                files = list(itertools.chain.from_iterable(map(_get_files_matching, patterns)))
+    except Exception as e:
+        raise RuntimeError("Failed to collect files for formatting (See above)") from e
+    # Fail if no files matched
+    assert files
+    # Split the file list into groups to be dispatched
+    num_jobs: int = min(args.jobs, len(files))
+    groups = [files[n::num_jobs] for n in range(num_jobs)]
+    print(f"Formatting {len(files)} files with {num_jobs} workers...", file=sys.stderr)
```
eramongodb, line 79: Remarking that despite my findings [in this comment](https://github.com/mongodb/mongo-cxx-driver/pull/1299#issuecomment-2536656219), the changes in this PR appear to outperform basic `clang-format` invocation with all files. 👍 

```
# <files> | xargs -P 0 -n 1 uv run clang-format -i
            Mean        Std.Dev.    Min         Median      Max
real        7.417       0.338       6.786       7.520       7.714
user        80.042      2.296       77.068      79.555      83.523
sys         27.626      1.872       26.172      26.553      31.238
```

```
# <files> | xargs -P \"$(nproc)\" -n 1 uv run clang-format -i
            Mean        Std.Dev.    Min         Median      Max
real        6.985       0.371       6.390       6.927       7.491
user        74.266      3.582       68.196      75.638      78.603
sys         26.000      1.883       22.674      26.924      28.102
```

```
# uv run --frozen clang-format -i <files>
            Mean        Std.Dev.    Min         Median      Max
real        5.008       0.154       4.766       5.055       5.227
user        4.930       0.131       4.751       4.941       5.112
sys         0.095       0.011       0.081       0.099       0.108
```

```
# uv run tools/format.py
            Mean        Std.Dev.    Min         Median      Max
real        2.360       0.146       2.152       2.382       2.562
user        16.140      0.734       15.197      16.091      17.346
sys         0.920       0.072       0.817       0.929       1.010
```

tools/format.py
```diff
@@ -0,0 +1,280 @@
...
+    )
+    parser.add_argument(
+        "--files",
+        help="""
+        A globbing pattern to select the files to be formatted (Use "**" for recursive search).
+        If omitted, then all first-party source files will be selected
+        """,
+        metavar="<pattern>",
+        action="append",
+    )
```
eramongodb: I found this flag difficult to use correctly as-written. Can this script be adjusted to accept a list of files to format (as command line arguments) rather than a pattern? `pathlib.Path.glob()` is too limited in its behavior and imo not ideal for this purpose (`*`/`**` inconsistent with gitignore/etc. behavior, no OR pattern operator to match `foo|bar`, does not exclude not-normal-files such as directories, etc.)

```
# Matches `clang-format -i <list of files>` syntax.
$ uv run tools/format.py --mode=apply <list of files>
```

src/common/src/mlib/ckdint.test.cpp
```diff
@@ -1,9 +1,10 @@
-#include <mlib/test.h>
 #include <mlib/ckdint.h>
 
+#include <mlib/test.h>
```
eramongodb, line 3: Curious why `<mlib/test.h>` is being grouped separately from `<mlib/ckdint.h>`. Shouldn't both match the `<mlib/|<common` regex?

.clang-format
```diff
@@ -151,20 +151,35 @@ ForEachMacros:
...
-    CaseSensitive:   false
-  - Regex:           '.*'
-    Priority:        1
-    SortPriority:    0
-    CaseSensitive:   false
+  # The prelude header must always come first
+  - Regex: '-prelude.h>'
+    Priority: 1
+  # Relative "in this directory" headers next
+  - Regex: '"./|"../'
```
eramongodb:
```suggestion
  - Regex: '"\./|"\.\./'
```
Matching literal `'.'`.

.clang-format
```diff
@@ -151,20 +151,35 @@ ForEachMacros:
...
-  - Regex:           '^(<|"(gtest|gmock|isl|json)/)'
-    Priority:        3
-    SortPriority:    0
-    CaseSensitive:   false
-  - Regex:           '.*'
-    Priority:        1
-    SortPriority:    0
-    CaseSensitive:   false
+  # The prelude header must always come first
+  - Regex: '-prelude.h>'
```
eramongodb:
```suggestion
  - Regex: '-prelude\.h>'
```
Match literal `'.'`.

.clang-format
```diff
@@ -151,20 +151,35 @@ ForEachMacros:
...
+    Priority: 1
+  # Relative "in this directory" headers next
+  - Regex: '"./|"../'
+    Priority: 2
+  # Private API headers
+  - Regex: '-private.h>'
```
eramongodb:
```suggestion
  - Regex: '-private\.h>'
```
Match literal `'.'`.

.clang-format
```diff
@@ -151,20 +151,35 @@ ForEachMacros:
...
+  - Regex: '<mongoc/'
+    Priority: 10
+  # all `bson/` headers come after `mongoc` headers
+  - Regex: '<bson/|<bsonutil/'
+    Priority: 20
+  # Private API components
+  - Regex: '<mlib/|<common'
+    Priority: 30
+  # Platform headers 
+  - Regex: '<(unistd|netdb|[Ww]in|[W]s2|netinet/.*|sys/.*|netdb|fcntl).h>'
```
eramongodb:
```suggestion
  - Regex: '<(unistd|netdb|[Ww]in|[W]s2|netinet/.*|sys/.*|netdb|fcntl)\.h>'
```
Match literal `'.'`.

Some more platform-specific headers include (excluding the trailing `.h`):

- `pthread` (POSIX)
- `openssl/.*` (OpenSSL)
- `windows`, `bcrypt`, `sspi`, `WinDNS`, etc. (Windows)
- `jsonsl/.*` and `utf8proc` (potentially bundled external libraries)

The current groupings do not seem to account for test headers (e.g. `TestSuite.h`, `test-conveniences.h`, `mock_server/.*`, etc.). Are they deliberately excluded?

.clang-format
```diff
@@ -151,20 +151,35 @@ ForEachMacros:
...
+    Priority: 90
+  # Standard library headers are near last
+  - Regex: '<(assert|string|ctype|limits|time|math|errno|inttypes).h>|<std'
```
eramongodb:
```suggestion
  - Regex: '<(assert|string|ctype|limits|time|math|errno|inttypes)\.h>|<std'
```
Match literal `'.'`.

Can we simplify this regex by assuming any `<>`-include by this point is assumed to be a system header as a fallback similar to the "other relative headers" regex?

```
  - Regex: '^<'
    Priority: 100
```

eramongodb (changes requested):
I am concerned by the current groupings lack of special-casing the `bson-compat.h` and `bson-config.h` headers, which provide "global" configuration and compatibility macro (re)definitions that need to be consistent across all components. Diagnosing inconsistent macros definitions due to lack of including these headers (which I am not confident is enforced correctly by all current components) is terribly difficult, especially given many configuration and platform-dependent macros are existence-checked (`#if defined(FOO)`). Can we boost the priority of these two headers header in the groupings, and/or ensure they is unconditionally included "first" and "always" by including it within the prelude header?

tools/format.py
vector-of-bool: I'm still not entirely sure why it so _massively_ beats `xargs`, but that was a pleasant surprise.

src/common/src/mlib/ckdint.test.cpp
vector-of-bool: This is because it detects `ckdint.h` as the "main header" for this file, so it forces it to the top. Several other source files have this behavior appear, but I'm not entirely sure why some do and some don't.

.clang-format
vector-of-bool: I'll add groups for external libs. 

I couldn't decide on a good sort order for testing headers, so I didn't add one, but I think it would be good to add those before merging.

tools/format.py
vector-of-bool: YAGNI on selecting files by pattern. I'll remove it and just use positional file lists.

Re: `/**/` patterns: Is Python's `**` behavior significantly different from the `.gitignore` behavior? I haven't played around with `.gitignore` patterns enough to know.

vector-of-bool:
@eramongodb re: `#ifdef` issues: Just tried to fix up the preludes to `#include` the config/compat/macro headers, and while possible it makes things even more complicated than they already are.

OTOH, there aren't actually that many `ifdef` and `defined(X)` in the codebase, and I think it would be pretty easy to find-replace them with a new pattern that is resilient to typos and missing inclusions. For example, modifying `bson-config.h.in`:

```c
// Keep for compatibililty, but deprecate:
/*
 * Define to 1 if you have strings.h available on your platform.
 */
#define BSON_HAVE_STRINGS_H @BSON_HAVE_STRINGS_H@
#if BSON_HAVE_STRINGS_H != 1
# undef BSON_HAVE_STRINGS_H
#endif

// New and improved, always expands to 0 or 1:
#define BSON_HAVE_STRINGS_HEADER() @BSON_HAVE_STRINGS_H@
```

I opened CDRIVER-6046 to track this. I think it could be done in a very simple follow-up PR.

src/common/src/mlib/ckdint.test.cpp
eramongodb: Odd. I thought this behavior was disabled by [setting](https://github.com/mongodb/mongo-cxx-driver/blob/72f650157d12b1fffeeee6549e6b9e6e49798bf3/.clang-format#L205) `IncludeIsMainSourceRegex: ''`. Perhaps it isn't doing what I thought it should.

tools/format.py
eramongodb: The most notable difference is `pathlib.Path.glob()`'s `**` matching directories, but not files, (TIL [this changed in Python 3.13](https://docs.python.org/3.13/library/pathlib.html#pattern-language)), which is why the `**/*` is necessary instead of just `**`.

eramongodb (approved):

PR #2052, opened 2025-06-30
CDRIVER-5998 share `SCHANNEL_CRED`

kevinAlbs:
# Summary

Share `SCHANNEL_CRED` across connections.

Verified with this patch: https://spruce.mongodb.com/version/686298d51b4881000729652b/

# Background & Motivation

Sharing the `SCHANNEL_CRED` is motivated by an upcoming fix for CDRIVER-5998. The private key for a client certificate is currently imported as an ephemeral key stored in-process (via [call to CryptImportKey](https://github.com/mongodb/mongo-c-driver/blob/31c13b6080a8a153be9c849b7c4138bfa1cbc6d8/src/libmongoc/src/mongoc/mongoc-secure-channel.c#L320)). Supporting SHA2 signatures appears to require importing a persisted private key stored on the file-system (with a unique name). This will be described further in a follow-up PR. But to avoid importing many duplicate keys, I want to change the current behavior from "import key per connection" to "import key per client/pool and share". 

Though improving performance was not the main goal of this PR, it reduces file-reading from per-connection to per-client/per-pool. Sharing the credentials showed an improvement in a [one-off benchmark](https://parsley.mongodb.com/evergreen/mongo_c_driver_sasl_matrix_winssl_sasl_sspi_winssl_windows_2019_vs2017_x64_test_8.0_server_auth_patch_30ac9bf0d8b2677a1a1e92a4fe5a13584845f30e_68627721ab51140007ba7d44_25_06_30_11_38_11/0/task?bookmarks=0,7110&shareLine=5324) creating 100 connections in 10 threads:
```
No sharing took: 8766.00ms
Sharing took: 8093.00ms
```

Sharing `SCHANNEL_CRED` follows a similar pattern to the shared OpenSSL context implemented in https://github.com/mongodb/mongo-c-driver/pull/1673.

Microsoft documentation does not appear to document thread-safety of sharing `SCHANNEL_CRED`. However, mongod shares the same `SCHANNEL_CRED` for client connections (via [_clientCred](https://github.com/10gen/mongo/blob/43c23fb8f8c8de17fd6eed61124f5e0a5714a948/src/mongo/util/net/ssl_manager_windows.cpp#L1481-L1482)). I expect the TLS handshake only needs to read `SCHANNEL_CRED`.

`SCHANNEL_CRED` is documented as [deprecated](https://learn.microsoft.com/en-us/windows/win32/api/schannel/ns-schannel-schannel_cred). This is not addressed in this PR, but I expect can be replaced when adding support for TLS v1.3 (CDRIVER-6045).

## Shared pointer

A `mongoc_shared_ptr` is used to manage `SCHANNEL_CRED`. This is to support changing the SSL options after creating connections (may have a use case for rotating certificates?). This behavior is [tested](https://github.com/mongodb/mongo-c-driver/blob/c63bb91fcecffd101ce2f620dc8d25aa7c3b7f5c/src/libmongoc/tests/test-mongoc-client.c#L2070-L2080) in the OpenSSL implementation:

```c
/* any operation - ping the server */
ret = mongoc_client_command_simple (client, "admin", tmp_bson ("{'ping': 1}"), NULL, NULL, &error);
ASSERT_OR_PRINT (ret, error);

/* change ssl opts before a second connection */
ssl_opts = test_framework_get_ssl_opts ();
mongoc_client_set_ssl_opts (client, ssl_opts);

/* any operation - ping the server */
ret = mongoc_client_command_simple (client, "admin", tmp_bson ("{'ping': 1}"), NULL, NULL, &error);
ASSERT_OR_PRINT (ret, error);
```

`mongoc_client_pool_t` however [documents](https://mongoc.org/libmongoc/2.0.2/mongoc_client_pool_set_ssl_opts.html):

> This function can only be called once on a pool, and must be called before the first call to `mongoc_client_pool_pop()`.

Therefore, the updates to the `mongoc_shared_ptr` are assumed to not require thread-safety.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-client-pool.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-pool.c
@@ -87,9 +87,7 @@ mongoc_client_pool_set_ssl_opts (mongoc_client_pool_t *pool, const mongoc_ssl_op
 #if defined(MONGOC_ENABLE_SSL_OPENSSL) && OPENSSL_VERSION_NUMBER >= 0x10100000L
       SSL_CTX_free (pool->topology->scanner->openssl_ctx);
       pool->topology->scanner->openssl_ctx = _mongoc_openssl_ctx_new (&pool->ssl_opts);
-#endif
-
-#if defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
+#elif defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
       // Access to secure_channel_cred_ptr does not need the thread-safe `mongoc_atomic_*` functions.
       // secure_channel_cred_ptr is not expected to be modified by multiple threads.
       // mongoc_client_pool_set_ssl_opts documentation prohibits calling after threads start.
--- a/src/libmongoc/src/mongoc/mongoc-client-pool.c
+++ b/src/libmongoc/src/mongoc/mongoc-client-pool.c
@@ -39,6 +39,10 @@
 #include <mongoc/mongoc-openssl-private.h>
 #endif
 
+#if defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
+#include <mongoc/mongoc-stream-tls-secure-channel-private.h>
+#endif
+
 struct _mongoc_client_pool_t {
    bson_mutex_t mutex;
    mongoc_cond_t cond;
@@ -83,6 +87,15 @@ mongoc_client_pool_set_ssl_opts (mongoc_client_pool_t *pool, const mongoc_ssl_op
 #if defined(MONGOC_ENABLE_SSL_OPENSSL) && OPENSSL_VERSION_NUMBER >= 0x10100000L
       SSL_CTX_free (pool->topology->scanner->openssl_ctx);
       pool->topology->scanner->openssl_ctx = _mongoc_openssl_ctx_new (&pool->ssl_opts);
+#endif
+
+#if defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
+      // Access to secure_channel_cred_ptr does not need the thread-safe `mongoc_atomic_*` functions.
+      // secure_channel_cred_ptr is not expected to be modified by multiple threads.
+      // mongoc_client_pool_set_ssl_opts documentation prohibits calling after threads start.
+      mongoc_shared_ptr_reset (&pool->topology->scanner->secure_channel_cred_ptr,
+                               mongoc_secure_channel_cred_new (&pool->ssl_opts),
+                               mongoc_secure_channel_cred_deleter);
 #endif
    }
 
--- a/src/libmongoc/src/mongoc/mongoc-client.c
+++ b/src/libmongoc/src/mongoc/mongoc-client.c
@@ -70,6 +70,11 @@
 #include <mongoc/mongoc-stream-tls-private.h>
 #endif
 
+#if defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
+#include <mongoc/mongoc-stream-tls-secure-channel-private.h>
+#include <mongoc/mongoc-stream-tls-private.h>
+#endif
+
 #include <common-string-private.h>
 #include <mlib/cmp.h>
 
@@ -756,6 +761,7 @@ mongoc_client_connect (bool buffered,
                        const mongoc_uri_t *uri,
                        const mongoc_host_list_t *host,
                        void *openssl_ctx_void,
+                       mongoc_shared_ptr secure_channel_cred_ptr,
                        bson_error_t *error)
 {
    mongoc_stream_t *base_stream = NULL;
@@ -765,6 +771,7 @@ mongoc_client_connect (bool buffered,
    BSON_ASSERT (host);
 
    BSON_UNUSED (openssl_ctx_void);
+   BSON_UNUSED (secure_channel_cred_ptr);
 
 #ifndef MONGOC_ENABLE_SSL
    if (ssl_opts_void || mongoc_uri_get_tls (uri)) {
@@ -814,6 +821,9 @@ mongoc_client_connect (bool buffered,
          // Use shared OpenSSL context.
          base_stream = mongoc_stream_tls_new_with_hostname_and_openssl_context (
             base_stream, host->host, ssl_opts, true, (SSL_CTX *) openssl_ctx_void);
+#elif defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
+         // Use shared Secure Channel credentials.
+         base_stream = mongoc_stream_tls_new_with_secure_channel_cred (base_stream, ssl_opts, secure_channel_cred_ptr);
 #else
          base_stream = mongoc_stream_tls_new_with_hostname (base_stream, host->host, ssl_opts, true);
 #endif
@@ -881,9 +891,13 @@ mongoc_client_default_stream_initiator (const mongoc_uri_t *uri,
 
 #if defined(MONGOC_ENABLE_SSL_OPENSSL) && OPENSSL_VERSION_NUMBER >= 0x10100000L
    SSL_CTX *ssl_ctx = client->topology->scanner->openssl_ctx;
-   return mongoc_client_connect (true, use_ssl, ssl_opts_void, uri, host, (void *) ssl_ctx, error);
+   return mongoc_client_connect (
+      true, use_ssl, ssl_opts_void, uri, host, (void *) ssl_ctx, MONGOC_SHARED_PTR_NULL, error);
+#elif defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
+   mongoc_shared_ptr cred_ptr = client->topology->scanner->secure_channel_cred_ptr;
+   return mongoc_client_connect (true, use_ssl, ssl_opts_void, uri, host, NULL, cred_ptr, error);
 #else
-   return mongoc_client_connect (true, use_ssl, ssl_opts_void, uri, host, NULL, error);
+   return mongoc_client_connect (true, use_ssl, ssl_opts_void, uri, host, NULL, MONGOC_SHARED_PTR_NULL, error);
 #endif
 }
 
@@ -1027,6 +1041,12 @@ _mongoc_client_set_ssl_opts_for_single_or_pooled (mongoc_client_t *client, const
 #if defined(MONGOC_ENABLE_SSL_OPENSSL) && OPENSSL_VERSION_NUMBER >= 0x10100000L
       SSL_CTX_free (client->topology->scanner->openssl_ctx);
       client->topology->scanner->openssl_ctx = _mongoc_openssl_ctx_new (&client->ssl_opts);
+#endif
+
+#if defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
+      mongoc_shared_ptr_reset (&client->topology->scanner->secure_channel_cred_ptr,
+                               mongoc_secure_channel_cred_new (&client->ssl_opts),
+                               mongoc_secure_channel_cred_deleter);
 #endif
    }
 }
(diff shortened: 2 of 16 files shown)


src/libmongoc/tests/test-mongoc-secure-channel.c
```diff
@@ -0,0 +1,263 @@
...
+
+static bool
+connect_with_secure_channel_cred (mongoc_ssl_opt_t *ssl_opt, mongoc_shared_ptr cred_ptr, bson_error_t *error)
+{
+   mongoc_host_list_t host;
+   const int32_t connect_timout_ms = 10000;
+
+   *error = (bson_error_t) {0};
+
+   if (!_mongoc_host_list_from_string_with_err (&host, "localhost:27017", error)) {
```
vector-of-bool, line 23: Note: This line will break with #2047 because it changes the signature of this function. Whichever is merged first will require changing the other.

src/libmongoc/tests/test-mongoc-secure-channel.c
```diff
@@ -0,0 +1,263 @@
...
+      }
+
+      // Expect insert OK. Cert fails to load, but server configured with --tlsAllowConnectionsWithoutCertificates:
+      {
+         bool ok = try_ping (client, &error);
+         ASSERT_OR_PRINT (ok, error);
+      }
+
+      // Expect exactly one attempt to load the client cert:
+      ASSERT_CMPSIZE_T (1, ==, cf.failures);
```
vector-of-bool: Can use the type-generic `mlib_check` macros

```suggestion
      mlib_check (1, eq, cf.failures);
```

vector-of-bool (approved):
LGTM with only minor comments

src/libmongoc/tests/test-mongoc-secure-channel.c
```diff
@@ -0,0 +1,263 @@
...
+      mongoc_shared_ptr_reset_null (&cred_ptr);
+   }
+}
+
+typedef struct {
+   size_t failures;
+   size_t failures2;
+} cert_failures;
+
+void
```
eramongodb:
```suggestion
static void
```
Only needs internal linkage.

src/libmongoc/tests/test-mongoc-secure-channel.c
```diff
@@ -0,0 +1,263 @@
...
+count_cert_failures (mongoc_log_level_t log_level, const char *log_domain, const char *message, void *user_data)
+{
```
eramongodb, line 118:
```suggestion
{
   BSON_UNUSED (log_level);
   BSON_UNUSED (log_domain);
```

Unused parameters.

src/libmongoc/src/mongoc/mongoc-client-pool.c
```diff
@@ -84,6 +88,15 @@ mongoc_client_pool_set_ssl_opts (mongoc_client_pool_t *pool, const mongoc_ssl_op
       SSL_CTX_free (pool->topology->scanner->openssl_ctx);
       pool->topology->scanner->openssl_ctx = _mongoc_openssl_ctx_new (&pool->ssl_opts);
 #endif
+
+#if defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
```
eramongodb: Here:

```c
#if defined(MONGOC_ENABLE_SSL_OPENSSL) && ...
// ...
#endif

#if defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
// ...
#endif
```

However, in mongoc-client.c:

```c
#if defined(MONGOC_ENABLE_SSL_OPENSSL) && ...
// ...
#elif defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
// ...
#else
```

Recommend being consistent with how exclusive SSL configuration options are handled. Use of `#elif` seems to be more consistent with existing conditionals.

src/libmongoc/src/mongoc/mongoc-stream-tls.c
```diff
@@ -227,4 +227,18 @@ mongoc_stream_tls_new_with_hostname_and_openssl_context (
...
 #endif
 
+#if defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
+mongoc_stream_t *
+mongoc_stream_tls_new_with_secure_channel_cred (mongoc_stream_t *base_stream,
+                                                mongoc_ssl_opt_t *opt,
+                                                mongoc_shared_ptr secure_channel_cred_ptr)
+{
+   if (opt->weak_cert_validation) {
+      // For compatibility with `mongoc_stream_tls_new_with_hostname`, modify `opt` directly:
```
eramongodb, line 226: Requesting better (internal) documentation associating this function with `mongoc_stream_tls_new_with_hostname`, given this function is meant to be an extension of prior behavior with `mongoc_stream_tls_new_with_hostname` e.g. per `mongoc_client_connect`:

```c
// Before (given ENABLE_SSL=WINDOWS):
base_stream = mongoc_stream_tls_new_with_hostname (base_stream, host->host, ssl_opts, true);

// After (given ENABLE_SSL=WINDOWS):
base_stream = mongoc_stream_tls_new_with_secure_channel_cred (base_stream, ssl_opts, secure_channel_cred_ptr);
```

This is not immediately apparent given the function name or this comment + contextual code.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -881,9 +891,13 @@ mongoc_client_default_stream_initiator (const mongoc_uri_t *uri,
 
 #if defined(MONGOC_ENABLE_SSL_OPENSSL) && OPENSSL_VERSION_NUMBER >= 0x10100000L
    SSL_CTX *ssl_ctx = client->topology->scanner->openssl_ctx;
-   return mongoc_client_connect (true, use_ssl, ssl_opts_void, uri, host, (void *) ssl_ctx, error);
+   return mongoc_client_connect (
+      true, use_ssl, ssl_opts_void, uri, host, (void *) ssl_ctx, MONGOC_SHARED_PTR_NULL, error);
+#elif defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
+   mongoc_shared_ptr cred_ptr = client->topology->scanner->secure_channel_cred_ptr;
+   return mongoc_client_connect (true, use_ssl, ssl_opts_void, uri, host, NULL, cred_ptr, error);
```
eramongodb:
```suggestion
   return mongoc_client_connect (
      true, use_ssl, ssl_opts_void, uri, host, NULL, client->topology->scanner->secure_channel_cred_ptr, error);
```
Recommend avoiding the single-use local variable.

src/libmongoc/src/mongoc/mongoc-stream-tls.c
```diff
@@ -227,4 +227,18 @@ mongoc_stream_tls_new_with_hostname_and_openssl_context (
...
+      opt->allow_invalid_hostname = true;
+   }
```
eramongodb, line 228: Confirming, is the "silly check for Unix Domain Sockets" not needed here?

src/libmongoc/tests/test-mongoc-secure-channel.c
```diff
@@ -0,0 +1,263 @@
...
+#if defined(MONGOC_ENABLE_SSL_SECURE_CHANNEL)
+#include <mongoc/mongoc-stream-tls-secure-channel-private.h>
+#include <mongoc/mongoc-host-list-private.h> // _mongoc_host_list_from_string_with_err
+#include <mongoc/mongoc-client-private.h>    // mongoc_client_connect_tcp
+#include <mongoc/mongoc-log-private.h>       // _mongoc_log_get_handler
+#include <test-conveniences.h>
+#include <mlib/test.h>
+
+static bool
+connect_with_secure_channel_cred (mongoc_ssl_opt_t *ssl_opt, mongoc_shared_ptr cred_ptr, bson_error_t *error)
```
eramongodb: All calls to `connect_with_secure_channel_cred` are followed by an assertion that the result is true. Consider changing return type from `bool` to `void` and move assertions into this function.

src/libmongoc/tests/test-mongoc-secure-channel.c
```diff
@@ -0,0 +1,263 @@
...
+   {
+      mongoc_shared_ptr cred_ptr =
+         mongoc_shared_ptr_create (mongoc_secure_channel_cred_new (&ssl_opt), mongoc_secure_channel_cred_deleter);
+      ok = connect_with_secure_channel_cred (&ssl_opt, cred_ptr, &error);
+      ASSERT_OR_PRINT (ok, error);
+      // Use again.
+      ok = connect_with_secure_channel_cred (&ssl_opt, cred_ptr, &error);
+      ASSERT_OR_PRINT (ok, error);
+      mongoc_shared_ptr_reset_null (&cred_ptr);
+   }
```
eramongodb: Given this test component has visibility of `mongoc_stream_tls_secure_channel_t::cred_ptr`, suggest adding postcondition assertions that a reused `cred_ptr` is actually reflected by the internal `cred_ptr` (points to the same object).

src/libmongoc/src/mongoc/mongoc-stream-tls.c
kevinAlbs: Updated block comments on `mongoc_stream_tls_new_with_hostname` and the internal OpenSSL and SChannel variants. Added/updated argument asserts to clarify NULL-pointer expectations.

In the process, discovered and fixed a pre-existing leak in `mongoc_stream_tls_secure_channel_new_with_creds`.

src/libmongoc/src/mongoc/mongoc-stream-tls.c
kevinAlbs: I expect "no" since the existing check is guarded by a `_WIN32` macro.

src/libmongoc/tests/test-mongoc-secure-channel.c
kevinAlbs: I would rather keep as-is. If `connect_with_secure_channel_cred` fails, I want the assert message to show the line of the caller, e.g.
```
FAIL:[...]test-mongoc-secure-channel.c:57  test_secure_channel_shared_creds_stream()
  ok
```

Rather than which line within this function failed:

```
FAIL:[...]test-mongoc-secure-channel.c:26  connect_with_secure_channel_cred()
  Condition '!tcp_stream' failed.
```

src/libmongoc/tests/test-mongoc-secure-channel.c
kevinAlbs: I like that idea. Test has been updated to check the pointer and reference count of the shared credentials.

src/libmongoc/tests/test-mongoc-secure-channel.c
```diff
@@ -48,24 +48,63 @@ test_secure_channel_shared_creds_stream (void *unused)
 {
    BSON_UNUSED (unused);
 
-   bool ok;
+   mongoc_stream_t *stream;
+   mongoc_stream_tls_secure_channel_t *schannel_stream_view;
```
eramongodb:
```suggestion
```
Stray changes?

src/libmongoc/tests/test-mongoc-secure-channel.c
```diff
@@ -0,0 +1,304 @@
...
+// Test a TLS stream can be create with shared Secure Channel credentials.
+static void
+test_secure_channel_shared_creds_stream (void *unused)
+{
+   BSON_UNUSED (unused);
+
+   mongoc_stream_t *stream;
+   mongoc_stream_tls_secure_channel_t *schannel_stream_view;
+   bson_error_t error;
+   mongoc_ssl_opt_t ssl_opt = {.ca_file = CERT_TEST_DIR "/ca.pem", .pem_file = CERT_TEST_DIR "/client.pem"};
```
eramongodb:
```suggestion
   const mongoc_ssl_opt_t ssl_opt = {.ca_file = CERT_TEST_DIR "/ca.pem", .pem_file = CERT_TEST_DIR "/client.pem"};
```

I believe `ssl_opt` can be made `const`. It is used by:

- `connect_with_secure_channel_cred` (OK: no modification)
    - `mongoc_stream_tls_secure_channel_new_with_creds` (OK: no modification)
        - `mongoc_secure_channel_cred_new` (OK: already const)
- `mongoc_secure_channel_cred_new` (OK: already const)

This reveals the `opt` param of `mongoc_stream_tls_secure_channel_new` can also be made `const`, even though it is invoked by `mongoc_stream_tls_new_with_hostname` with a non-const `opt` (after modification).

eramongodb (approved):

src/libmongoc/tests/test-mongoc-secure-channel.c
kevinAlbs: Good spot. Updated to use more `const`.

PR #2053, opened 2025-07-02
[CDRIVER-4153] Beginning of Header Hygiene & Verification

vector-of-bool:
Refer: CDRIVER-4153

# Summary

This PR is an exploration of how we can incrementally begin to clean up our header hygiene, put together on a whim, when I realized the following:

1. We no longer allow users to include unqualified header names (e.g. No more `<bson.h>`, they must now do `<bson/bson.h>`)
2. With the additional directory prefix, all of the `bson-` and `mongoc-` prefixes are redundant and somewhat annoying, and I'd like to get rid of those prefixes
3. CMake 3.24 supports *header verification*, which will generate a compilation rule in an ephemeral target that *only* includes a single header.
4. We can selectively enable header verification on individual files based on a name pattern.
5. Because all `bson-` and `mongoc-` headers have guards against being included directly (via the `-prelude.h` headers), those files are not part of the public API, as users cannot access them directly. This gives us a filename pattern that we can use to *deselect* files for header verification.

Thus, this changeset. It works like this:

- For any header file in `bson/` or `mongoc/` that *does not* begin with `bson-` or `mongoc-`, we want to perform header verification on those files.
- To add a file to the verification check, rename it from e.g. `bson/bson-foo.h` to `bson/foo.h`. You can now run the build and header verification targets to check that the header can stand alone. This also gives the header a nicer name without a redundant `bson-` prefix, for public consumption as `bson/foo.h`
- We don't use header verification directly on the `bson_static` and `mongoc_static` targets, as it (1) requires a newer CMake than we require of users and (2) only checks as C or as C++, but not both. Instead, two new placeholder targets are created that have the headers attached: One checks the headers as C, and the other as C++. (See: `build/VerifyHeaders.cmake`)

> [!NOTE] 
> This method can also supplant the need for the separate `mongoc-cxx-check` and `public-header-warnings` targets, since we can now roll those into the header verification directly (this PR doesn't do that).

# Changes

The following is changed:

1. Add a `VerifyHeaders.cmake` module with a `mongo_verify_headers` command. This creates two blank targets whose only purpose is the creation of the verification rules.
2. Add selective header verification for `bson` and `mongoc` headers.
3. As a basic proof-of-concept, the following headers were made "public", which involved moving the relevant components into those standalone headers:
   - `bson/config.h` is the new filename of the generate libbson config header.
   - `bson/macros.h` contains only compile-time preprocessor macros.
   - `bson/compat.h` contains OS and C library compatibility shims
   - `bson/bson_t.h` contains only the definition of the `bson_t` struct
   - `bson/error.h` contains `bson_error_t` and its associated funtions
   - `bson/memory.h` contains `bson_malloc`, `bson_free`, etc.
   - (The `bson/bson.h` header still transitively includes all of the above headers, so users can continue to use the "get everything" header)
4. Fixed some minor C *and* C++ incompatibilities that were uncovered by enabling header verification.
5. Modify the `check-prelude.py` to only consider files that have the `bson-` or `mongoc-` prefix, to mirror the behavior of our selective header verification.
6. Add an Earthly target and Evergreen `verify-headers` task[^check-headers] that runs header verification (against a few platforms, since differences in C stdlib headers can cause issues).

[^check-headers]: Unfortunately, this may be confusing with the `check-headers` task, but I'd rather migrate to the name `verify-headers` since it directly corresponds to the name of the CMake functionality, and running them in an Earthly target runs faster in CI.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -12,7 +12,7 @@ IMPORT ./tools/ AS tools
 # COPY_SOURCE :
 #   Copy source files required for the build into the specified "--into" directory
 COPY_SOURCE:
-    COMMAND
+    FUNCTION
     ARG --required into
     COPY --dir \
         build/ \
@@ -28,7 +28,7 @@ COPY_SOURCE:
 # CONFIGURE :
 #   Configure the project in $source_dir into $build_dir with a common set of configuration options
 CONFIGURE:
-    COMMAND
+    FUNCTION
     ARG --required source_dir
     ARG --required build_dir
     ARG --required tls


build/cmake/VerifyHeaders.cmake
```diff
@@ -0,0 +1,117 @@
...
+        [EXCLUDE_REGEX [<pattern> ...]]
+    )
+
+Here `<tag>` is an arbitrary string that is used to qualify the internal target
+created for the verification. The `<glob>` expressions are used to automatically
+collect sources files (relative to the current source directory). All files
+collected by `<glob>` must live within the current source directory.
+
+After collecting sources according to the `<glob>` patterns, sources are
+excluded if they filepath contains any substring that matches any regular
```
rcsanchez97:
```suggestion
excluded if the filepath contains any substring that matches any regular
```

build/cmake/VerifyHeaders.cmake
```diff
@@ -0,0 +1,117 @@
...
+expression in the `<pattern>` list. Each `<pattern>` is tested against the
+relative path to the header file that was found by `<glob>`, and not the
+absolute path to the file file.
```
rcsanchez97:
```suggestion
absolute path to the file.
```

build/cmake/VerifyHeaders.cmake
```diff
@@ -0,0 +1,117 @@
...
+            )
+        # Warn if this pattern didn't match anything. It is probably a mistake
+        # in the caller's argument list.
+        if(NOT more)
+            message(WARNING "Globbing pattern “${pattern}” did not match any files")
+        endif()
+        list(APPEND headers_to_verify ${more})
+    endforeach()
+
+    # Excluce anything that matches any exclusion regex
```
rcsanchez97:
```suggestion
    # Exclude anything that matches any exclusion regex
```

build/cmake/VerifyHeaders.cmake
```diff
@@ -0,0 +1,117 @@
...
+        # C++ is available. define it
+        set(cxx_target ${tag}-verify-headers-cxx)
+        message(DEBUG "Defining header verification targets “${cxx_target}” (C++)")
+        add_library(${cxx_target} OBJECT "${CMAKE_CURRENT_FUNCTION_LIST_DIR}/empty.cpp")
+        target_sources(${cxx_target} PUBLIC FILE_SET HEADERS)
+    else()
+        message(AUTHOR_WARNING "No C++ compiler is available, so the header-check C++ targets won't be defined")
+        unset(cxx_target)
+    endif()
+    # Populate the properies and file sets.
```
rcsanchez97:
```suggestion
    # Populate the properties and file sets.
```

tools/build.earth
```diff
@@ -83,6 +83,12 @@ ADD_CXX_COMPILER:
             # Old Alpine 3.16 bundles libstdc++ in g++, so we need to pull GCC along anyway
             LET cxx_pkg = clang g++
         END
+    ELSE IF test -f /etc/redhat-release
+        IF test "$cxx_compiler" = g++
+            LET cxx_pkg = "gcc-c++"
+        ELSE IF test "$cxx_cmopiler" = clang++
```
eramongodb:
```suggestion
        ELSE IF test "$cxx_compiler" = clang++
```

src/libbson/src/bson/memory.h
eramongodb, line 1: The license header for this file is missing.

src/libbson/src/bson/bson_t.h
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * BSON_INITIALIZER:
+ *
+ * This macro can be used to initialize a #bson_t structure on the stack
+ * without calling bson_init().
+ *
+ * |[
+ * bson_t b = BSON_INITIALIZER;
+ * ]|
```
eramongodb, line 55: What is this syntax using `|[` and `]|`?

src/libbson/src/bson/bcon.h
```diff
@@ -24,8 +24,8 @@
 #define BCON_H_
 
 // Include specific headers first, because bson.h tries to include this header itself:
-#include <bson/bson-macros.h>
 #include <bson/bson-types.h>
+#include <bson/macros.h>
 // For other APIs, not properly grouped, but needed:
 #include <bson/bson.h>
```
eramongodb, line 30: Does this include of `<bson/bson.h>` + `<bson/bson-prelude.h>` above need to be updated?

build/cmake/empty.cpp
```diff
@@ -0,0 +1,2 @@
+// This is an empty placeholder file use for miscellaneous build system tasks
+#include <vector>
```
eramongodb: Consider `<cstddef>` instead.

build/cmake/empty.c
```diff
@@ -0,0 +1,2 @@
+/* This is an empty placeholder file use for miscellaneous build system tasks */
+#include <stdlib.h>
```
eramongodb: Consider `<stddef.h>` instead.

src/common/src/common-string-private.h
```diff
@@ -485,7 +485,7 @@ mcommon_string_append_overflow (mcommon_string_append_t *append)
  */
 bool
 mcommon_string_append_selected_chars (mcommon_string_append_t *append,
-                                      const char *template,
+                                      const char *template_,
```
eramongodb, line 488: Is this header being compiled as C++?

src/libbson/CMakeLists.txt
```diff
@@ -173,6 +173,25 @@ if(ENABLE_STATIC)
       # User wants static libs to use PIC code.
       set_property(TARGET bson_static PROPERTY POSITION_INDEPENDENT_CODE TRUE)
    endif()
+
+   if(MONGO_CAN_VERIFY_HEADERS)
```
eramongodb: I think we should guard `mongo_verify_headers` on `ENABLE_TESTS` to avoid defining relevant CMake targets during default configure-build-install workflows, even if they are not built as part of the `all` target.

build/cmake/VerifyHeaders.cmake
```diff
@@ -0,0 +1,117 @@
...
+    # Conditionally do the same thing for C++
+    if(CMAKE_CXX_COMPILER)
+        # C++ is available. define it
+        set(cxx_target ${tag}-verify-headers-cxx)
+        message(DEBUG "Defining header verification targets “${cxx_target}” (C++)")
+        add_library(${cxx_target} OBJECT "${CMAKE_CURRENT_FUNCTION_LIST_DIR}/empty.cpp")
+        target_sources(${cxx_target} PUBLIC FILE_SET HEADERS)
+    else()
+        message(AUTHOR_WARNING "No C++ compiler is available, so the header-check C++ targets won't be defined")
+        unset(cxx_target)
```
eramongodb, line 89: Unnecessary `unset()`? `cxx_target` shouldn't be set in this scope.

build/cmake/VerifyHeaders.cmake
```diff
@@ -0,0 +1,117 @@
...
+    endif()
+    # Populate the properties and file sets.
+    set_target_properties(${c_target} ${cxx_target} PROPERTIES
```
eramongodb, line 92:
```suggestion
    set_target_properties(${targets} PROPERTIES
```
Should this use `${targets}` instead (which I assume is meant to conditionally contain `${cxx_target}`)?

build/cmake/VerifyHeaders.cmake
```diff
@@ -0,0 +1,117 @@
...
+
+    # Drop duplicates since globs may grab a file more than once
+    list(REMOVE_DUPLICATES headers_to_verify)
+    list(SORT headers_to_verify)
+    foreach(file IN LISTS headers_to_verify)
+        message(DEBUG "Verify header file: ${file}")
+    endforeach()
+
+    # We create two targets: One for C and one for C++
+    set(targets)
```
eramongodb: Unused variable. Is this function missing `list(APPEND targets ${c-target})` and an optional `list(APPEND targets ${cxx-target})` below?

build/cmake/VerifyHeaders.cmake
```diff
@@ -0,0 +1,117 @@
...
+    endforeach()
+
+    # Exclude anything that matches any exclusion regex
+    foreach(pattern IN LISTS arg_EXCLUDE_REGEX)
+        list(FILTER headers_to_verify EXCLUDE REGEX "${pattern}")
+    endforeach()
+
+    # Drop duplicates since globs may grab a file more than once
+    list(REMOVE_DUPLICATES headers_to_verify)
+    list(SORT headers_to_verify)
```
eramongodb, line 66: Is this sort just for convenience?

build/cmake/VerifyHeaders.cmake
```diff
@@ -0,0 +1,117 @@
...
+    # We create two targets: One for C and one for C++
+    set(targets)
+    # C target
+    set(c_target ${tag}-verify-headers-c)
+    message(DEBUG "Defining header verification target “${c_target}” (C)")
+    # Create object libraries. They will only have one empty compiled source file.
+    # The source file language will tell CMake how to verify the associated header files.
+    add_library(${c_target} OBJECT "${CMAKE_CURRENT_FUNCTION_LIST_DIR}/empty.c")
+    # Define the file set
+    target_sources(${c_target} PUBLIC FILE_SET HEADERS)
```
eramongodb, line 79: Why not pass `${headers_to_verify}` here in this call to `target_sources()`? Using `set_target_properties()` to directly modify `HEADER_SET` seems odd.

eramongodb (changes requested):
Quite exciting changes.

Rather than defining `<name>-verify-headers-c` (and `<name>-verify-headers-cpp`) object library targets, can we consider conditionally (on CMake 3.24 and newer) and selectively (opt-in) adding the verified headers to the headers file set for the `bson_static` and `mongoc_static` library targets instead?

```cmake
if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.27")
  # When `CMAKE_VERIFY_INTERFACE_HEADER_SETS=ON`, creates the
  # `bson_static_verify_interface_header_sets` target and the
  # `all_verify_interface_header_sets` target.
  target_sources (bson_static PUBLIC FILE_SET HEADERS ...)
endif ()
```

In addition to slightly improving how we handle headers in the CMake configuration (e.g. for bson, currently bundled together with `.c` files in the glob-recursed `all_sources` variable), since file sets are completely opt-in, and since the `VERIFY_INTERFACE_HEADER_SETS` property is `false` by default (due to `CMAKE_VERIFY_INTERFACE_HEADER_SETS` being unset by default), adding I think this will better reflect the "for development purposes" nature of the `VERIFY_INTERFACE_HEADER_SETS` property, which we can enable when building test targets (i.e. in CI) using the `CMAKE_VERIFY_INTERFACE_HEADER_SETS` variable during CMake configuration. [Quoting](https://cmake.org/cmake/help/latest/variable/CMAKE_VERIFY_INTERFACE_HEADER_SETS.html) CMake docs:

> Projects should not normally set this variable, it is intended as a developer control to be set on the `cmake` command line or other equivalent methods. The developer must have the ability to enable or disable header set verification according to the capabilities of their own machine and compiler.

If we want better control over the set of headers to be verified (that is, if we want to add more headers to the `HEADERS` file set than just those which we want to be verified; I think we want to eventually these into header file sets), we can use custom header set names together with the [`INTERFACE_HEADER_SETS_TO_VERIFY`](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_HEADER_SETS_TO_VERIFY.html) target property.

build/cmake/VerifyHeaders.cmake
vector-of-bool: I was consider both designs and went with using two variables, but accidentally left `${targets}` unused. Either design is roughly equivalent, though.

build/cmake/VerifyHeaders.cmake
vector-of-bool: It's only to force the order of the `DEBUG` message below.

build/cmake/VerifyHeaders.cmake
vector-of-bool: It *shouldn't*, but I can reasonably foresee it being a variable name that someone (perhaps months or years from now) will set in a caller's scope, so I clear it here. The alternative design is to use the `${targets}` as a list, but I just decided on this shape instead, for no reason other than arbitrary choice.

build/cmake/VerifyHeaders.cmake
vector-of-bool: I was initially going to do everything in the `set_target_properties` just to make sure both targets got all equiv properties, but it turns out there is no way to add a new file set via `set_target_properties`, but I kept the `set_target_properties` to fill the fileset below, probably just as an oversight. Either shape is the same in the end.

src/common/src/common-string-private.h
vector-of-bool: Not currently. There was a brief moment where I fed this header into a C++ compiler by accident and it complained, so I just added this as a drive-by tweak since it's simple.

src/libbson/src/bson/bcon.h
vector-of-bool: I don't quite understand the question.

src/libbson/src/bson/bson_t.h
vector-of-bool: I have no idea. It was the original comment, so I just kept it as-is. These doc comments are ancient and could use some TLC.

eramongodb: Sorry, I misread the diff and thought it was new. I'm fine with leaving it as-is.

src/libbson/src/bson/bcon.h
eramongodb: AFAICT this is the only non-`bson-`-prefixed header which still includes `<bson/bson-prelude.h>` and `<bson/bson.h>`. All others (which are now defined as standalone-includable "public" headers) removed both includes in favor of a more streamlined list. I think this header should perhaps be renamed to `<bson/bson-bcon.h>` for consistency with the new header naming scheme.

src/libbson/CMakeLists.txt
```diff
@@ -116,13 +116,13 @@ else ()
 endif ()
```
kevinAlbs, line 117: Can update the above comment "bson-config.h as preprocessor values" to "bson/config.h as preprocessor values".

src/libbson/src/bson/bson-version-functions.c
```diff
@@ -14,11 +14,9 @@
  * limitations under the License.
  */
 
-
 #include <bson/bson-version-functions.h>
 
-#include <bson/bson-version.h>
-
+#include <bson/version.h>
```
kevinAlbs, line 19: `bson-version.h` is also referenced in `check-files.py`. I expect `check-files.py` can be removed (appears unused, and I think is no longer needed now that tarballs are not generated in CMake).

kevinAlbs (approved):
LGTM with minor comments.

PR #2055, opened 2025-07-08
Remove the `Language: Cpp` setting for formatting

vector-of-bool:
It seems that the `Language: Cpp` setting in `.clang-format` changes Clang's behavior in Clang 20.x, and sometimes restricts it from formatting C files, instead falling back to the default format.

This PR removes `Language: Cpp` and allows clang-format to apply the same formatting to all source files. This seems to also have minor changes to the way it typesets the `not` identifier in C files (it looks like clang-format was previously treating C as C++, where `not` is a special prefix operator rather than a bare identifier).

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,5 +1,7 @@
 ---
-Language:        Cpp
+# Note: The "Lanugage" option is omitted, as we want to apply the same formatting
+# to all files, otherwise clang-format may (sometimes) apply the default format
+# to other languages.
 AccessModifierOffset: -2
 AlignAfterOpenBracket: Align
 AlignArrayOfStructures: None
--- a/src/libbson/tests/test-bson.c
+++ b/src/libbson/tests/test-bson.c
@@ -1952,8 +1952,8 @@ test_bson_regex_lengths (void)
       &new, "0_________1_________2_________3___4", -1, "0_________1_________2_________3_________4_________5___4", "i");
 
    ASSERT (new.len == 121);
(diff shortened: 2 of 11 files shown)


src/libbson/tests/test-bson.c
```diff
@@ -1952,8 +1952,8 @@ test_bson_regex_lengths (void)
       &new, "0_________1_________2_________3___4", -1, "0_________1_________2_________3_________4_________5___4", "i");
 
    ASSERT (new.len == 121);
-   ASSERT (new.flags &BSON_FLAG_STATIC);
-   ASSERT (!(new.flags &BSON_FLAG_INLINE));
+   ASSERT (new.flags & BSON_FLAG_STATIC);
+   ASSERT (!(new.flags & BSON_FLAG_INLINE));
```
vector-of-bool, line 1956: This confirms some suspicion that clang-format was treating C as C++, because it looks like the `new` identifier here was confusing it and typesetting the infix `&` strangely.

.clang-format
```diff
@@ -1,5 +1,7 @@
 ---
-Language:        Cpp
+# Note: The "Lanugage" option is omitted, as we want to apply the same formatting
+# to all files, otherwise clang-format may (sometimes) apply the default format
+# to other languages.
```
eramongodb:
```suggestion
# Note: the "Language" option is omitted to apply this format style to all files
# regardless of the detected language.
```
Typo + wording tweak suggestions.

PR #2057, opened 2025-07-11
restore C++ driver tests

kevinAlbs:
Follow-up to https://github.com/mongodb/mongo-c-driver/pull/1957 to restore C++ driver tests that were removed due to backwards breaking changes in C driver 2.0. The C++ driver r4.1.0 migrated to C driver 2.0, and now successfully builds.

Verified with this patch build: https://spruce.mongodb.com/version/68713f991a0d3a0007c0f055

### DIFF
--- a/.evergreen/config_generator/components/earthly.py
+++ b/.evergreen/config_generator/components/earthly.py
@@ -5,8 +5,17 @@ import re
 from typing import Iterable, Literal, Mapping, NamedTuple, TypeVar
 
 from shrub.v3.evg_build_variant import BuildVariant
-from shrub.v3.evg_command import BuiltInCommand, EvgCommandType, subprocess_exec
+from shrub.v3.evg_command import (
+    BuiltInCommand,
+    EvgCommandType,
+    KeyValueParam,
+    ec2_assume_role,
+    expansions_update,
+    subprocess_exec,
+)
 from shrub.v3.evg_task import EvgTask, EvgTaskRef
+
+from config_generator.etc.function import Function
 
 from ..etc.utils import all_possible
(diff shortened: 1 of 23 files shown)


.evergreen/config_generator/components/earthly.py
```diff
@@ -38,7 +38,7 @@
 "Valid options for the SASL configuration parameter"
 TLSOption = Literal["OpenSSL", "off"]
 "Options for the TLS backend configuration parameter (AKA 'ENABLE_SSL')"
-CxxVersion = Literal["none"] # TODO: Once CXX-3103 is released, add latest C++ release tag.
+CxxVersion = Literal["r4.1.0", "none"]
```
vector-of-bool: Is `4.1.0` the only version we want to test at the moment? Do we care to test `master`?

vector-of-bool (approved):
Sorry, thought I approved this but apparently not. LGTM with a minor comment.

Aside: I realized that this is a suboptimal testing setup, since we're just duplicating a lot of work where we build the C driver for multiple different C++ library versions (or `none`, as in one case), when it would be simpler to do all the test versions within a single EVG task and reuse the same build between them. I'll create a Jira ticket for this.

.evergreen/config_generator/components/earthly.py
kevinAlbs: I think testing `master` seems reasonable (and was [done before](https://github.com/mongodb/mongo-c-driver/blob/86037268ccaab73a841219225d49a0c32c5024c4/.evergreen/config_generator/components/earthly.py#L31)).

PR #2058, opened 2025-07-11
CDRIVER-5971 Use Amazon ECR to obtain all OCI images

eramongodb:
Resolves CDRIVER-5971. These changes should improve the performance and reliability of Earthly tasks on Evergreen by avoiding both DockerHub rate limits and Artifactory's limited bandwidth.

@mdb-ad and @rcsanchez97: even if not review-requested, I recommended attempting the documented instructions to authenticate with Amazon ECR and pull images on your local machine and verify if it works as expected, since these changes and instructions are meant to preserve/support the ability to run Earthly tasks locally (not Evergreen only).

---

DevProd plans to decommission the Artifactory instance currently providing DevProd Release Tools images (e.g. garasign-gpg, silkbomb, etc.). We temporarily used Artifactory as a workaround for DockerHub rate limits (https://github.com/mongodb/mongo-c-driver/pull/1794) but reverted shortly after (https://github.com/mongodb/mongo-c-driver/pull/1927). The new [DevProd-provided Amazon ECR instance](https://docs.devprod.prod.corp.mongodb.com/devprod-platforms-ecr/DockerHub) is intended to the proper long-term solution going forward.

The Amazon ECR instance is both a host for internal images (e.g. DevProd Release Tools) as well as a pull-through cache for DockerHub images. Therefore, it serves all our present OCI image acquisition needs. However, using this Amazon ECR instance requires performing [IAM Identity Center authentication](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sso.html) with the `devprod-platforms-ecr` AWS account. This is simple in Evergreen using `ec2.assume_role`, but the steps are a bit more involved on local developer machines. These steps are documented in `earthly.rst`, but reiterated below for locality.

---

DockerHub images `<name>` and `<orgname>/<name>` are specified as `library/<name>` and `<orgname>/<name>` respectively. DevProd Release Tools images `release-tools-container-registry-local/<name>` are specified as `release-infrastructure/<name>`.

Using Amazon ECR requires authenticating with the `devprod-platforms-ecr` AWS account. Before pulling any OCI images (via `docker`, `podman`, or `earthly`), you must obtain AWS short-term credentials using SSO authentication with AWS CLI v2 (note: _not_ v1, which does not support `aws configure sso`). AWS CLI v2 can be obtained by running `snap install awscli --classic` (see: [install docs](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html)). Initial configuration can be done by running `aws configure sso` or directly setting `$HOME/.aws/config` to the following:

```
[profile <profile>]
sso_session = <session_name>
sso_account_id = 901841024863
sso_role_name = ECRScopedAccess
region = us-east-1
[sso-session <session_name>]
sso_start_url = https://d-9067613a84.awsapps.com/start#/
sso_region = us-east-1
sso_registration_scopes = sso:account:access
```

> [!NOTE]
> "\<profile>" is the AWS account, role, etc. "\<session_name>" is the short-term credentials. I recommend setting them to something like "amazon-ecr" and "\<your username>" respectively.

To modify an existing configuration, run `aws configure sso --profile <profile>`.

When short-term credentials expire, the following error is emitted:

```
denied: Your authorization token has expired. Reauthenticate and try again.
```

The expired credentials be refreshed by running either one of:

```bash
aws sso login --profile <profile>
aws sso login --sso-session <session_name>
```

The AWS credentials must be passed to `podman` and/or `docker` after (re)authentication by running:

```bash
aws ecr get-login-password --profile <profile> | podman login --username AWS --password-stdin 901841024863.dkr.ecr.us-east-1.amazonaws.com
```

At this point, you should be able to pull images from Amazon ECR, e.g. by running `podman pull 901841024863.dkr.ecr.us-east-1.amazonaws.com/release-infrastructure/garasign-gpg`.

> [!NOTE]
> I was unable to find a method of [configuring the default container image registry](https://github.com/containers/image/blob/main/docs/containers-registries.conf.5.md) that preserves the use of short-names (e.g. `alpine:3.20`), but which _doesn't_ involve system-level or user-level configuration, which would conflict with Evergreen task reproducibility. I am open to suggestions or a followup PR; otherwise, all images require the `901841024863.dkr.ecr.us-east-1.amazonaws.com/` prefix to be pulled from Amazon ECR. This is not necessary when manually running `pull` commands on local machines (such as to pre-pull images from DockerHub prior to their use by Earthly), e.g. `podman pull alpine` should resolve to DockerHub (by default) whereas `podman pull garasign-gpg` should resolve to Amazon ECR (when authenticated).

### DIFF
--- a/.evergreen/config_generator/components/earthly.py
+++ b/.evergreen/config_generator/components/earthly.py
@@ -5,8 +5,17 @@ import re
 from typing import Iterable, Literal, Mapping, NamedTuple, TypeVar
 
 from shrub.v3.evg_build_variant import BuildVariant
-from shrub.v3.evg_command import BuiltInCommand, EvgCommandType, subprocess_exec
+from shrub.v3.evg_command import (
+    BuiltInCommand,
+    EvgCommandType,
+    KeyValueParam,
+    ec2_assume_role,
+    expansions_update,
+    subprocess_exec,
+)
 from shrub.v3.evg_task import EvgTask, EvgTaskRef
+
+from config_generator.etc.function import Function
 
 from ..etc.utils import all_possible
 
@@ -38,7 +47,7 @@ SASLOption = Literal["Cyrus", "off"]
 "Valid options for the SASL configuration parameter"
 TLSOption = Literal["OpenSSL", "off"]
 "Options for the TLS backend configuration parameter (AKA 'ENABLE_SSL')"
-CxxVersion = Literal["none"] # TODO: Once CXX-3103 is released, add latest C++ release tag.
+CxxVersion = Literal["none"]  # TODO: Once CXX-3103 is released, add latest C++ release tag.
 "C++ driver refs that are under CI test"
 
 # A separator character, since we cannot use whitespace
@@ -136,6 +145,34 @@ class Configuration(NamedTuple):
         return _SEPARATOR.join(f"{k}={v}" for k, v in self._asdict().items())
 
 
+# Use DevProd-provided Amazon ECR instance to obtain earthly-buildkitd in advance.
+class DockerLoginAmazonECR(Function):
+    name = 'docker-login-amazon-ecr'
+    commands = [
+        # Avoid inadvertently using a pre-existing and potentially conflicting Docker config.
+        expansions_update(updates=[KeyValueParam(key='DOCKER_CONFIG', value='${workdir}/.docker')]),
+        ec2_assume_role(role_arn="arn:aws:iam::901841024863:role/ecr-role-evergreen-ro"),
+        subprocess_exec(
+            binary="bash",
+            command_type=EvgCommandType.SETUP,
+            include_expansions_in_env=[
+                "AWS_ACCESS_KEY_ID",
+                "AWS_SECRET_ACCESS_KEY",
+                "AWS_SESSION_TOKEN",
+                "DOCKER_CONFIG",
+            ],
+            args=[
+                "-c",
+                'aws ecr get-login-password --region us-east-1 | docker login --username AWS --password-stdin 901841024863.dkr.ecr.us-east-1.amazonaws.com',
+            ],
+        ),
+    ]
+
+    @classmethod
+    def call(cls, **kwargs):
+        return cls.default_call(**kwargs)
+
+
 def task_filter(env: EarthlyVariant, conf: Configuration) -> bool:
     """
     Control which tasks are actually defined by matching on the platform and
@@ -175,6 +212,7 @@ def earthly_exec(
             *(f"--{arg}={val}" for arg, val in (args or {}).items()),
         ],
         command_type=EvgCommandType(kind),
+        include_expansions_in_env=["DOCKER_CONFIG"],
         env=env if env else None,
         working_dir="mongoc",
     )
@@ -209,10 +247,12 @@ def earthly_task(
     return EvgTask(
         name=name,
         commands=[
+            DockerLoginAmazonECR.call(),
             # Ensure subsequent Docker commands are authenticated.
             subprocess_exec(
                 binary="bash",
                 command_type=EvgCommandType.SETUP,
+                include_expansions_in_env=["DOCKER_CONFIG"],
                 args=[
                     "-c",
                     r'docker login -u "${artifactory_username}" --password-stdin artifactory.corp.mongodb.com <<<"${artifactory_password}"',
@@ -247,6 +287,10 @@ CONTAINER_RUN_DISTROS = [
     "ubuntu2204-large",
     "ubuntu2404-large",
 ]
+
+
+def functions():
+    return DockerLoginAmazonECR.defn()
 
 
 def tasks() -> Iterable[EvgTask]:
--- a/Earthfile
+++ b/Earthfile
@@ -1,4 +1,4 @@
-VERSION --arg-scope-and-set --pass-args 0.7
+VERSION --arg-scope-and-set --pass-args --use-function-keyword 0.7
 LOCALLY
 
 IMPORT ./tools/ AS tools
@@ -114,7 +114,7 @@ test-cxx-driver:
 
 # PREP_CMAKE "warms up" the CMake installation cache for the current environment
 PREP_CMAKE:
-    COMMAND
+    FUNCTION
     LET scratch=/opt/mongoc-cmake
     # Copy the minimal amount that we need, as to avoid cache invalidation
     COPY tools/use.sh tools/platform.sh tools/paths.sh tools/base.sh tools/download.sh \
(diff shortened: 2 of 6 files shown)


Earthfile
```diff
@@ -1,4 +1,4 @@
-VERSION --arg-scope-and-set --pass-args 0.7
+VERSION --arg-scope-and-set --pass-args --use-function-keyword 0.7
 LOCALLY
```
vector-of-bool, line 2: Add a `--global` build arg to specify where to get containers by default:

```suggestion
LOCALLY

ARG --global default_container_registry = "docker.io/library"
```

(I'm not set on this parameter name, but just an example.)

Earthfile
```diff
@@ -148,7 +148,7 @@ multibuild:
 # release-archive :
 #   Create a release archive of the source tree. (Refer to dev docs)
 release-archive:
-    FROM alpine:3.20
+    FROM 901841024863.dkr.ecr.us-east-1.amazonaws.com/dockerhub/library/alpine:3.20
```
vector-of-bool: Parameterize the registry:

```suggestion
    FROM $default_container_registry/alpine:3.20
```

and so on for other updated image references

tools/earthly.sh
```diff
@@ -36,7 +36,7 @@ if ! is-file "$EARTHLY_EXE"; then
 fi
 
 run-earthly() {
-    "$EARTHLY_EXE" "$@"
+    "$EARTHLY_EXE" --buildkit-image 901841024863.dkr.ecr.us-east-1.amazonaws.com/dockerhub/earthly/buildkitd:v0.8.3 "$@"
```
vector-of-bool: In the suggested refactor, specifying `--buildkit-image=...` would need to be done in the generated `tools/earthly.sh` commands for the EVG tasks, rather than hardcoded in this file.

Earthfile
eramongodb: Replaced all DockerHub-provided image names with `$default_search_registry/library/<image>`. DevProd-provided release tool images remain fully-qualified with the Amazon ECR registry name. Omitted the `/library` suffix to support orgname-qualified image short-names.

Earthfile
eramongodb: Chose the name `default_search_registry` in reference to podman's [`unqualified-search-registries`](https://github.com/containers/image/blob/main/docs/containers-registries.conf.5.md) setting.

tools/earthly.sh
eramongodb: > EVG tasks that specify this may want to use the `.arg` file or `EARTHLY_ARGS` environment variable for setting this parameter, since it would probably be used everywhere and its easier than trying to pipe it through all the build commands in the file.

Moving both into `earthly_exec()` as an additional set of CLI arguments to `earthly.sh` seems to be sufficient.

Earthfile
```diff
@@ -148,7 +151,7 @@ multibuild:
 # release-archive :
 #   Create a release archive of the source tree. (Refer to dev docs)
 release-archive:
-    FROM alpine:3.20
+    FROM $default_search_registry/library/alpine:3.20
```
vector-of-bool, line 154: Move the trailing `/library` into the parameter, as that's not universal to container registries (e.g. [quay.io][] doesn't use the `/library` suffix, and [quay.io][] is a reasonable alternative default).

[quay.io]: https://quay.io/

.evergreen/config_generator/components/earthly.py
```diff
@@ -136,6 +145,34 @@ def suffix(self) -> str:
         return _SEPARATOR.join(f"{k}={v}" for k, v in self._asdict().items())
 
 
+# Use DevProd-provided Amazon ECR instance to obtain earthly-buildkitd in advance.
```
vector-of-bool: This comment appears unrelated to this class's purpose.

vector-of-bool (approved):
Minor comments, LGTM

Earthfile
eramongodb: Embedding the `/library` suffix in the variable rules out its use with image names containing an orgname, e.g. if the `/library` suffix is embedded:

- `$default_search_registry/ubuntu:24.04` -> `docker.io/library/ubuntu:24.04` (correct)
- `$default_search_registry/earthly/buildkitd` -> `docker.io/library/earthly/buildkitd` (incorrect)
- `$default_search_registry/mongodb/mongo-cxx-driver` -> `docker.io/library/mongodb/mongo-cxx-driver` (incorrect)

I was expecting use of non-DockerHub-provided images to use fully-qualified names which include the registry name, as currently done for DevProd Release Tools images, e.g. `quay.io/centos/centos:stream10` rather than `centos:stream10`, which is also consistent with [Red Hat recommendations](https://www.redhat.com/en/blog/be-careful-when-pulling-images-short-name):

> Users pulling without specifying the full path, commonly referred to as a short name, leaves them open to a particularly nasty form of attack, called image squatting. [...] Red Hat recommends that users always use fully qualified names when referring to container images in any context. [...] Consider removing all registries from the search list so that short names don’t resolve.

I do not think the ability to set `$default_search_registry` to anything but the default (DockerHub) or Amazon ECR (in CI) is really in scope for these changes, which are primarily to avoid Artifactory without hitting DockerHub rate limits again. Let me know if you'd like the suffix to be embedded regardless.

.evergreen/config_generator/components/earthly.py
eramongodb: Whoops, outdated comment from old changes. Fixed.

docs/dev/earthly.rst
```diff
@@ -175,23 +175,34 @@ enumerated using ``earthly ls`` or ``earthly doc`` in the root of the repository
...
+   - The SSO account ID is ``901841024863`` (aka ``devprod-platforms-ecr``).
+   - The SSO role name is ``ECRScopedAccess`` (default).
+
+   To refresh short-term credentials when they have expired, run
+   ``aws sso login --profile <profile>`` followed by the same
+   ``aws ecr get-login-password ... | podman login ...`` command described
+   above.
+
+   .. seealso:: `"DevProd Platforms Container Registry"
+   <https://docs.devprod.prod.corp.mongodb.com/devprod-platforms-ecr>`_ and `"Configuring IAM Identity Center authentication with the AWS CLI" <https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sso.html>_`.
```
kevinAlbs:
```suggestion
   <https://docs.devprod.prod.corp.mongodb.com/devprod-platforms-ecr>`_ and `"Configuring IAM Identity Center authentication with the AWS CLI" <https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sso.html>`_.
```

.evergreen/config_generator/components/earthly.py
```diff
@@ -136,6 +145,34 @@ def suffix(self) -> str:
         return _SEPARATOR.join(f"{k}={v}" for k, v in self._asdict().items())
 
 
+# Authenticate with DevProd-provided Amazon ECR instance to use as pull-through cache for DockerHub.
```
kevinAlbs, line 148: Suggest applying similar changes to `sbom.py` / `sbom.sh` since `sbom.sh` still refers to an artifactory image.

kevinAlbs (approved):
Thank you for investigating and making these changes. I expect this will help the reliability of Evergreen tasks. LGTM with similar changes applied to the SBOM scripts.

PR #2059, opened 2025-07-14
CDRIVER-5998 import persisted private keys for SChannel

kevinAlbs:
# Summary

Import private keys as persisted (not ephemeral) for Secure Channel.

# Background & Motivation

This PR is intended to address an error sending a client certificate to servers that do not support the SHA1 client certificate signature:
```
The client and server cannot communicate, because they do not possess a common algorithm.
```

See [comment](https://jira.mongodb.org/browse/CDRIVER-5998?focusedCommentId=7279125&focusedId=7279125&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-7279125) for details. In particular, this prevents MONGODB-X509 auth (which authenticates with a client certificate).

Importing a persisted key appears required to support the SHA2 signature (and TLS 1.3, but that is planned in CDRIVER-6045). Experiments and references suggest persisting keys is relevant:

Quoting [.NET runtime](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Security/src/System/Net/Security/SslStreamPal.Windows.cs#L261-L262):
> on Windows we do not support ephemeral keys.

Quoting [Private Key Lifetime on Windows and the January 2021 .NET Framework Security and Quality Rollup](https://support.microsoft.com/en-us/topic/private-key-lifetime-on-windows-and-the-january-2021-net-framework-security-and-quality-rollup-3f097a10-f798-4ffd-8511-4757ebaf2c70):
> The default behavior is that the private key gets loaded into a persisted (named) key via one of the system cryptography libraries

Quoting https://stackoverflow.com/a/72101855:
> The TLS layer on Windows requires that the private key be written to disk

Quoting [PFXImportCertStore](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-pfximportcertstore):
> if the handshake is performed externally to the calling process in LSASS.exe, it is not possible to use PKCS12_NO_PERSIST_KEY

## Key names

Some deprecated calls in CryptoAPI (CAPI) are migrated to Cryptography API: Next Generation (CNG). Quoting [About CNG](https://learn.microsoft.com/en-us/windows/win32/seccng/about-cng):
> Cryptography API: Next Generation (CNG) is the long-term replacement for the CryptoAPI

PKCS#8 keys now use CNG (`NCryptImportKey`). PKCS#1 keys still use CAPI (`CryptImportKey`) since CNG APIs do not appear to support PKCS#1. Atlas gives PKCS#8 certificates.

Keys are imported with a deterministic name. The imported key name has the form `libmongoc-<SHA256 thumbprint>-<pkcs1|pkcs8>`. The `<SHA256 thumbprint>` was chosen to be easily computable by a user (via `openssl x509` command). The `<pkcs1|pkcs8>` suffix is to distinguish the same key imported via PKCS#1 and PKCS#8 files (will end up in different providers). Since the key name is chosen by the C driver, I want to document the key name so users can identify and delete if desired. If merged, I plan to request [this documentation](https://docs.google.com/document/d/1JbwrNgHwC_ISuHXeEm03TvicKDLvyBxZEdnJReNtJSw/edit?tab=t.0) be added.

## Rejected Alternative: Import with GUID

Importing the key with GUID key name was considered. Each client/pool would import with a unique GUID and delete in the destructor. This appears to match the .NET runtime behavior, but I am concerned that a C application is more likely to abort early:
```c
const char* uri = "mongodb://localhost:27017/?tlsCertificateKeyFile=client.pem";
mongoc_client_t *client = mongoc_client_new(uri); // imports key with GUID name.
assert (returns_false()); // exits and does not delete imported key!
```
Each key "leak" could result in a build-up of orphaned imported keys.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-secure-channel.c
+++ b/src/libmongoc/src/mongoc/mongoc-secure-channel.c
@@ -19,12 +19,14 @@
 #ifdef MONGOC_ENABLE_SSL_SECURE_CHANNEL
 
 #include <common-string-private.h>
+#include <mongoc/mongoc-crypto-private.h> // mongoc_crypto_hash
 #include <mongoc/mongoc-errno-private.h>
 #include <mongoc/mongoc-error-private.h>
 #include <mongoc/mongoc-secure-channel-private.h>
 #include <mongoc/mongoc-stream-tls-private.h>
 #include <mongoc/mongoc-stream-tls-secure-channel-private.h>
 #include <mongoc/mongoc-trace-private.h>
+#include <mongoc/mongoc-util-private.h> // bin_to_hex
 
 #include <mongoc/mongoc-log.h>
 #include <mongoc/mongoc-ssl.h>
@@ -33,7 +35,6 @@
 #include <bson/bson.h>
 
 #include <mlib/cmp.h>
-
 
 #undef MONGOC_LOG_DOMAIN
 #define MONGOC_LOG_DOMAIN "stream-secure-channel"
@@ -207,6 +208,55 @@ decode_object (const char *structType,
    return out;
 }
 
+// `generate_key_name` generates a deterministic name for a key of the form: "libmongoc-<SHA256 thumbprint>-<suffix>".
+// Returns NULL on error.
+static LPWSTR
+generate_key_name (LPBYTE data, DWORD len, const char *suffix)
+{
+   bool ok = false;
+   char *hash_hex = NULL;
+   char *key_name = NULL;
+   LPWSTR key_name_wide = NULL;
+
+   BSON_ASSERT_PARAM (data);
+   BSON_ASSERT_PARAM (suffix);
+
+   // Compute a hash of the contents:
+   {
+      unsigned char hash[32];
+      mongoc_crypto_t crypto;
+      mongoc_crypto_init (&crypto, MONGOC_CRYPTO_ALGORITHM_SHA_256);
+      BSON_ASSERT (mlib_in_range (size_t, len));
+      if (!mongoc_crypto_hash (&crypto, (const unsigned char *) data, (size_t) len, hash)) {
+         goto fail;
+      }
+      // Use uppercase hex to match form of `openssl x509` command:
+      hash_hex = bin_to_hex ((const uint8_t *) hash, (uint32_t) sizeof (hash), true);
+   }
+
+   // Convert to a wide string:
+   {
+      key_name = bson_strdup_printf ("libmongoc-%s-%s", hash_hex, suffix);
+      size_t key_name_wide_chars = strlen (key_name) + 1;
+      key_name_wide = bson_malloc (sizeof (WCHAR) * (key_name_wide_chars));
+      if (0 ==
+          MultiByteToWideChar (CP_UTF8, 0, key_name, -1 /* NULL terminate */, key_name_wide, key_name_wide_chars)) {
+         goto fail;
+      }
+   }
+
+   ok = true;
+fail:
+   bson_free (key_name);
+   bson_free (hash_hex);
+   if (!ok) {
+      bson_free (key_name_wide);
+      key_name_wide = NULL;
+   }
+
+   return key_name_wide;
+}
+
 PCCERT_CONTEXT
 mongoc_secure_channel_setup_certificate_from_file (const char *filename)
 {
@@ -214,7 +264,7 @@ mongoc_secure_channel_setup_certificate_from_file (const char *filename)
    bool ret = false;
    bool success;
    size_t pem_length;
-   HCRYPTPROV provider;
+   HCRYPTPROV provider = 0u;
    DWORD encoded_cert_len;
    LPBYTE encoded_cert = NULL;
    const char *pem_public;
@@ -226,6 +276,10 @@ mongoc_secure_channel_setup_certificate_from_file (const char *filename)
    DWORD blob_private_rsa_len = 0;
    DWORD encoded_private_len = 0;
    LPBYTE encoded_private = NULL;
+   NCRYPT_PROV_HANDLE cng_provider = 0u;
+   LPWSTR key_name = NULL;
+
+   BSON_ASSERT_PARAM (filename);
 
    pem = read_file_and_null_terminate (filename, &pem_length);
    if (!pem) {
@@ -258,7 +312,10 @@ mongoc_secure_channel_setup_certificate_from_file (const char *filename)
       goto fail;
    }
 
+   // Import private key as a persisted (not ephemeral) key.
+   // Ephemeral keys do not appear to support modern signatures. See CDRIVER-5998.
    if (NULL != (pem_private = strstr (pem, "-----BEGIN RSA PRIVATE KEY-----"))) {
+      // Import PKCS#1 as a persisted CAPI key. Windows CNG API does not appear to support PKCS#1.
       encoded_private = decode_pem_base64 (pem_private, &encoded_private_len, "private key", filename);
       if (!encoded_private) {
          goto fail;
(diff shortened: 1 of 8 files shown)


src/libmongoc/src/mongoc/mongoc-util.c
```diff
@@ -1014,13 +1014,18 @@ hex_to_bin (const char *hex, uint32_t *len)
...
 {
    char *out = bson_malloc0 (2u * len + 1u);
 
    for (uint32_t i = 0u; i < len; i++) {
+      int req;
+      if (uppercase) {
+         req = bson_snprintf (out + (2u * i), 3, "%02X", bin[i]);
+      } else {
+         req = bson_snprintf (out + (2u * i), 3, "%02x", bin[i]);
+      }
```
eramongodb: Consider splitting `bin_to_hex()` into a `bin_to_hex_lowercase()` and `bin_to_hex_uppercase()` instead of using a `bool` parameter.

src/libmongoc/tests/test-mongoc-x509.c
```diff
@@ -158,7 +265,17 @@ test_x509_auth (void *unused)
...
+
          mongoc_client_t *client = test_framework_client_new_from_uri (uri, NULL);
+         SCHANNEL_ASSERT_PKCS8_KEY_IMPORTED (); // Imported.
+         ok = try_insert (client, &error);
+         mongoc_client_destroy (client);
+
+         // Auth again with key still imported on Secure Channel:
+         SCHANNEL_ASSERT_PKCS8_KEY_IMPORTED (); // Still imported.
+         client = test_framework_client_new_from_uri (uri, NULL);
          ok = try_insert (client, &error);
```
eramongodb:
```suggestion
         SCHANNEL_ASSERT_PKCS8_KEY_IMPORTED (); // Imported.
         ASSERT_OR_PRINT (try_insert (client, &error), error);
         mongoc_client_destroy (client);

         // Auth again with key still imported on Secure Channel:
         SCHANNEL_ASSERT_PKCS8_KEY_IMPORTED (); // Still imported.
         client = test_framework_client_new_from_uri (uri, NULL);
         ASSERT_OR_PRINT (try_insert (client, &error), error);
         mongoc_client_destroy (client);

         SCHANNEL_DELETE_PKCS8_KEY ();

         // Auth again with key no longer imported on Secure Channel:
         SCHANNEL_ASSERT_PKCS8_KEY_NOT_IMPORTED ();
         client = test_framework_client_new_from_uri (uri, NULL);
         ASSERT_WITH_ERROR (try_insert (client, &error), ...);
```

Recommend asserting the postcondition of each operation individually (import, reuse, delete).

src/libmongoc/tests/test-mongoc-x509.c
```diff
@@ -136,6 +136,113 @@ try_insert (mongoc_client_t *client, bson_error_t *error)
...
+      delete_imported_pkcs1_key (); \
+   else                             \
+      (void) 0
+#else
+#define SCHANNEL_ASSERT_PKCS8_KEY_IMPORTED() (void) 0
+#define SCHANNEL_ASSERT_PKCS8_KEY_NOT_IMPORTED() (void) 0
+#define SCHANNEL_DELETE_PKCS8_KEY() (void) 0
+#define SCHANNEL_ASSERT_PKCS1_KEY_IMPORTED() (void) 0
+#define SCHANNEL_ASSERT_PKCS1_KEY_NOT_IMPORTED() (void) 0
+#define SCHANNEL_DELETE_PKCS1_KEY() (void) 0
```
eramongodb:
```suggestion
#define SCHANNEL_ASSERT_PKCS8_KEY_IMPORTED() ((void) 0)
#define SCHANNEL_ASSERT_PKCS8_KEY_NOT_IMPORTED() ((void) 0)
#define SCHANNEL_DELETE_PKCS8_KEY() ((void) 0)
#define SCHANNEL_ASSERT_PKCS1_KEY_IMPORTED() ((void) 0)
#define SCHANNEL_ASSERT_PKCS1_KEY_NOT_IMPORTED() ((void) 0)
#define SCHANNEL_DELETE_PKCS1_KEY() ((void) 0)
```
Parenthesis should wrap the whole expression.

src/libmongoc/tests/test-mongoc-x509.c
```diff
@@ -136,6 +136,113 @@ try_insert (mongoc_client_t *client, bson_error_t *error)
...
+   ASSERT_WITH_MSG (
+      status == SEC_E_OK || status == NTE_BAD_KEYSET, "Failed to open key: %s", mongoc_winerr_to_string (status));
+   bool found = (keyHandle != 0);
+
+   NCryptFreeObject (keyHandle);
+   NCryptFreeObject (hProv);
+   return found;
+}
+
+// pkcs1_key_name is the deterministic name for the key: client.pem
```
eramongodb:
```suggestion
// The deterministic name for the certificate in src/libmongoc/tests/x509gen/client.pem.
```
Wording tweak suggestion to include full (relative) path to the key of interest + mention that it's the contents of the _certificate_ that is relevant (not the private key).

src/libmongoc/tests/test-mongoc-x509.c
```diff
@@ -136,6 +136,113 @@ try_insert (mongoc_client_t *client, bson_error_t *error)
...
+   // Open the software key storage provider:
+   NCRYPT_PROV_HANDLE hProv = 0;
+   SECURITY_STATUS status = NCryptOpenStorageProvider (&hProv, MS_KEY_STORAGE_PROVIDER, 0);
+   ASSERT_WITH_MSG (status == SEC_E_OK, "Failed to open key storage provider: %s", mongoc_winerr_to_string (status));
+
+   // Open the key handle:
+   NCRYPT_PROV_HANDLE keyHandle = 0;
+   status = NCryptOpenKey (hProv, &keyHandle, pkcs8_key_name, 0, 0);
+   ASSERT_WITH_MSG (
+      status == SEC_E_OK || status == NTE_BAD_KEYSET, "Failed to open key: %s", mongoc_winerr_to_string (status));
```
eramongodb, line 180: Request documenting why [`NTE_BAD_KEYSET`](https://learn.microsoft.com/en-us/windows/win32/api/ncrypt/nf-ncrypt-ncryptopenkey#return-value) is permitted in this context.

> The specified key was not found. 

Additionally, I suspect the `NCRYPT_KEY_HANDLE` out parameter (that is, `&keyHandle`) is not guaranteed to be non-zero on success, as documentation concerning success vs. failure detection only refers to the return value:

> Returns a status code that indicates the success or failure of the function.

src/libmongoc/tests/test-mongoc-x509.c
```diff
@@ -136,6 +136,113 @@ try_insert (mongoc_client_t *client, bson_error_t *error)
...
+   bool found = (keyHandle != 0);
+
+   NCryptFreeObject (keyHandle);
+   NCryptFreeObject (hProv);
```
eramongodb:
```suggestion
   ASSERT (NCryptFreeObject (keyHandle) == ERROR_SUCCESS);
   ASSERT (NCryptFreeObject (hProv) == ERROR_SUCCESS);
```
Discarded return values which can/should be asserted.

src/libmongoc/tests/test-mongoc-x509.c
```diff
@@ -136,6 +136,113 @@ try_insert (mongoc_client_t *client, bson_error_t *error)
...
+delete_imported_pkcs1_key (void)
+{
+   HCRYPTPROV provider;
+   bool success = CryptAcquireContextW (&provider,                          /* phProv */
+                                        pkcs1_key_name,                     /* pszContainer */
+                                        MS_ENHANCED_PROV_W,                 /* pszProvider */
+                                        PROV_RSA_FULL,                      /* dwProvType */
+                                        CRYPT_DELETEKEYSET | CRYPT_SILENT); /* dwFlags */
+   ASSERT_WITH_MSG (success, "Failed to delete key: %s", mongoc_winerr_to_string (GetLastError ()));
+   CryptReleaseContext (provider, 0);
```
eramongodb:
```suggestion
   ASSERT (CryptReleaseContext (provider, 0));
```
Discarded return value which can/should be asserted. Also applies to other instances of `CryptReleaseContext`.

src/libmongoc/tests/test-mongoc-x509.c
```diff
@@ -136,6 +136,113 @@ try_insert (mongoc_client_t *client, bson_error_t *error)
...
+   NCRYPT_PROV_HANDLE keyHandle = 0;
+   status = NCryptOpenKey (hProv, &keyHandle, pkcs8_key_name, 0, 0);
+   ASSERT_WITH_MSG (status == SEC_E_OK, "Failed to open key: %s", mongoc_winerr_to_string (status));
+
+   // Delete key:
+   status = NCryptDeleteKey (keyHandle, 0); // Also frees handle.
+   ASSERT_WITH_MSG (status == SEC_E_OK, "Failed to delete key: %s", mongoc_winerr_to_string (status));
+
+   // NCryptDeleteKey freed handle.
+   NCryptFreeObject (hProv);
```
eramongodb:
```suggestion
   ASSERT (NCryptFreeObject (hProv) == ERROR_SUCCESS);
```

Discarded return value which can/should be asserted. Also applies to other instances of `NCryptFreeObject`.

src/libmongoc/tests/test-mongoc-x509.c
```diff
@@ -158,7 +265,17 @@ test_x509_auth (void *unused)
       bson_error_t error = {0};
       bool ok;
       {
+         SCHANNEL_DELETE_PKCS8_KEY ();
+         SCHANNEL_ASSERT_PKCS8_KEY_NOT_IMPORTED ();
+
          mongoc_client_t *client = test_framework_client_new_from_uri (uri, NULL);
+         SCHANNEL_ASSERT_PKCS8_KEY_IMPORTED (); // Imported.
+         ok = try_insert (client, &error);
```
eramongodb: These tests might need to `capture_log()` and assert the following error is not present (e.g. when running without administrative privileges):

```
ERROR: stream-secure-channel: Error opening certificate store: (0x00000005) Access is denied.
```

Speaking of which, are [administrator privileges](https://learn.microsoft.com/en-us/windows/win32/SecBP/running-with-administrator-privileges) really necessary for client key storage management? Can the required privileges be reduced to user-level or lower? (Can defer this to a separate task if the required changes are substantial.)

src/libmongoc/src/mongoc/mongoc-secure-channel.c
```diff
@@ -207,14 +222,63 @@ decode_object (const char *structType,
...
+{
+   bool ok = false;
+   char *hash_hex = NULL;
+   char *key_name = NULL;
+   LPWSTR key_name_wide = NULL;
+
+   BSON_ASSERT_PARAM (data);
+   BSON_ASSERT_PARAM (suffix);
+
+   // Compute a hash of the contents:
```
eramongodb:
```suggestion
   // Compute a hash of the certificate:
```
Clarify it's the _certificate_ being hashed, not the private key.

src/libmongoc/src/mongoc/mongoc-secure-channel.c
```diff
@@ -207,14 +222,63 @@ decode_object (const char *structType,
    return out;
 }
 
+// `generate_key_name` generates a deterministic name for a key of the form: "libmongoc-<SHA256 thumbprint>-<suffix>".
```
eramongodb:
```suggestion
// `generate_key_name` generates a deterministic name for a key of the form: "libmongoc-<SHA256 fingerprint>-<suffix>".
```
I believe ["fingerprint"](https://en.wikipedia.org/wiki/Public_key_fingerprint) is the more familiar term to use here.

src/libmongoc/src/mongoc/mongoc-secure-channel.c
```diff
@@ -207,14 +222,63 @@ decode_object (const char *structType,
...
+      hash_hex = bin_to_hex ((const uint8_t *) hash, (uint32_t) sizeof (hash), true);
+   }
+
+   // Convert to a wide string:
+   {
+      key_name = bson_strdup_printf ("libmongoc-%s-%s", hash_hex, suffix);
+      size_t key_name_wide_chars = strlen (key_name) + 1;
+      key_name_wide = bson_malloc (sizeof (WCHAR) * (key_name_wide_chars));
+      if (0 ==
+          MultiByteToWideChar (CP_UTF8, 0, key_name, -1 /* NULL terminate */, key_name_wide, key_name_wide_chars)) {
```
eramongodb:
```suggestion
      if (0 == MultiByteToWideChar (
                  CP_UTF8, 0, key_name, -1 /* NULL terminate */, key_name_wide, (int) key_name_wide_chars)) {
```

```
warning C4267: 'function': conversion from 'size_t' to 'int', possible loss of data
```
This seems to be due to `key_name_wide_chars: size_t` being passed to the `ccWideCharint` parameter. Given the SHA256 and suffix are expected to be short strings bounded in length, recommend adding an explicit cast here (no range check necessary).

src/libmongoc/tests/test-mongoc-x509.c
```diff
@@ -136,6 +136,113 @@ try_insert (mongoc_client_t *client, bson_error_t *error)
...
+   HCRYPTPROV provider = 0;
+   bool success = CryptAcquireContextW (&provider,          /* phProv */
+                                        pkcs1_key_name,     /* pszContainer */
+                                        MS_ENHANCED_PROV_W, /* pszProvider */
+                                        PROV_RSA_FULL,      /* dwProvType */
+                                        CRYPT_SILENT);      /* dwFlags */
+   if (!success) {
+      DWORD lastError = GetLastError ();
+      ASSERT_WITH_MSG (
+         lastError = NTE_BAD_KEYSET, "Unexpected error in acquiring context: %s", mongoc_winerr_to_string (lastError));
```
eramongodb:
```suggestion
      ASSERT_WITH_MSG (
         lastError == NTE_BAD_KEYSET, "Unexpected error in acquiring context: %s", mongoc_winerr_to_string (lastError));
```
```
warning C4245: '=': conversion from 'HRESULT' to 'DWORD', signed/unsigned mismatch
```
Should this be equality comparison? (Recommend `const`-qualifying local variables.)

eramongodb (changes requested):
Recommend also updating [`connect_with_secure_channel_cred`](https://github.com/mongodb/mongo-c-driver/blob/afb406d5069306b07379caa7af40c11f691e931b/src/libmongoc/tests/test-mongoc-secure-channel.c#L18) to improve its handling of pre-existing credentials (e.g. due to prior test failure):

```
FAIL:src\libmongoc\tests\test-mongoc-secure-channel.c:58  test_secure_channel_shared_creds_stream()
  stream
  Failed to initialize security context: (0x80092010) The certificate is revoked.

test-libmongoc : test_secure_channel_shared_creds_stream src\libmongoc\tests\test-mongoc-secure-channel.c:58
```

src/libmongoc/src/mongoc/mongoc-util.c
kevinAlbs: Done. On further look, I made additional changes to the `bin_to_hex*` and `hex_to_bin` utilities:

- Use `size_t` for lengths to avoid casts with `uint32_t`.
- Use the recently added `mlib_nat64_parse` to parse hex. This avoids `sscanf` quirks (e.g. skipping whitespace).

I considered separating those changes to another PR. But since `bin_to_hex` and `hex_to_bin` were only previously used in tests, I expect this is a low impact drive-by improvement.

src/libmongoc/src/mongoc/mongoc-secure-channel.c
kevinAlbs: Updated here and in the [planned documentation](https://docs.google.com/document/d/1JbwrNgHwC_ISuHXeEm03TvicKDLvyBxZEdnJReNtJSw/edit?tab=t.0).

src/libmongoc/src/mongoc/mongoc-secure-channel.c
kevinAlbs: Added cast with an asserted range check to clarify the cast is expected not to overflow.

src/libmongoc/tests/test-mongoc-x509.c
kevinAlbs: > Request documenting

Added comment noting `NTE_BAD_KEYSET` is expected if key is not found.

> not guaranteed to be non-zero on success

Good spot. Updated to only check return status.

src/libmongoc/tests/test-mongoc-x509.c
kevinAlbs: Good spot. Updated.

src/libmongoc/tests/test-mongoc-x509.c
kevinAlbs: Updated others. Adding an assert to this call failed. This clued me to [CryptAcquireContextW](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptacquirecontexta) documentation for `CRYPT_DELETEKEYSET`:

> When this flag is set, the value returned in phProv is undefined, and thus, the CryptReleaseContext function need not be called afterward.

src/libmongoc/tests/test-mongoc-x509.c
kevinAlbs: Creating the `mongoc_client_t` imports the key. So the "import" and "delete" cases both test the same behavior (key not already imported). Reorganized tests to (hopefully) clarify.

src/libmongoc/tests/test-mongoc-x509.c
kevinAlbs: The `Error opening certificate store` error is from calls to `CertOpenStore`. Those are unchanged in this PR. Reported CDRIVER-6054 for a possible future improvement.

src/libmongoc/tests/test-mongoc-x509.c
kevinAlbs: Done. I did not add asserts to non-test code out of caution to avoid a possible runtime assert. [NCryptFreeObject documents](https://learn.microsoft.com/en-us/windows/win32/api/ncrypt/nf-ncrypt-ncryptfreeobject):

> Possible return codes include, but are not limited to

Calls to other free functions (e.g. `CryptReleaseContext`) also ignore the return, and I am unsure if some of the error codes may come up in normal scenarios (e.g. [ERROR_BUSY](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptreleasecontext)).

eramongodb: > I did not add asserts to non-test code out of caution to avoid a possible runtime assert.

Agreed, prior suggestions to assert return values were meant to apply to test code only.

src/libmongoc/tests/test-mongoc-util.c
```diff
@@ -95,24 +95,56 @@ static void
...
+   // Test non-even number of digits is an error:
+   ASSERT (!hex_to_bin ("666", &len));
 
-   uint32_t len;
-   uint8_t *got = hex_to_bin (hexstr, &len);
-   ASSERT_CMPSTR ((const char *) got, expect);
-   ASSERT_CMPUINT32 (len, ==, 7);
-   bson_free (got);
+   // Test non-hex digits is an error:
+   ASSERT (!hex_to_bin ("ZZ", &len));
```
eramongodb:
```c
   ASSERT_WITH_MSG (!hex_to_bin ("  66", &len), "whitespace is an error");
   ASSERT_WITH_MSG (!hex_to_bin ("666", &len), "non-even number of digits is an error");
   ASSERT_WITH_MSG (!hex_to_bin ("ZZ", &len), "non-hex digits is an error");
```

Move reason into assertion message.

eramongodb (approved):

src/libmongoc/src/mongoc/mongoc-secure-channel.c
```diff
@@ -207,14 +222,67 @@ decode_object (const char *structType,
...
+   BSON_ASSERT_PARAM (data);
+   BSON_ASSERT_PARAM (suffix);
+
+   // Compute a hash of the certificate:
+   {
+      unsigned char hash[32];
+      mongoc_crypto_t crypto;
+      mongoc_crypto_init (&crypto, MONGOC_CRYPTO_ALGORITHM_SHA_256);
+      BSON_ASSERT (mlib_in_range (size_t, len));
+      if (!mongoc_crypto_hash (&crypto, (const unsigned char *) data, (size_t) len, hash)) {
```
vector-of-bool: Can also use `mlib_assert_narrow` to do the check-and-cast inline, and it will print the bad value in case of assertion failure.

```suggestion
      if (!mongoc_crypto_hash (&crypto, 
                               (const unsigned char *) data,
                               mlib_assert_narrow (size_t, len), 
                               hash)) {
```

src/libmongoc/src/mongoc/mongoc-secure-channel.c
```diff
@@ -207,14 +222,67 @@ decode_object (const char *structType,
...
+      }
+   }
+
+   // Convert to a wide string:
+   {
+      key_name = bson_strdup_printf ("libmongoc-%s-%s", hash_hex, suffix);
+      size_t key_name_wide_chars = strlen (key_name) + 1;
+      key_name_wide = bson_malloc (sizeof (WCHAR) * (key_name_wide_chars));
+      BSON_ASSERT (mlib_in_range (int, key_name_wide_chars));
+      if (0 == MultiByteToWideChar (
```
vector-of-bool: `MultiByteToWideChar` (and other Win32 string conversion APIs) is a foot bazooka, and I get nervous every time I see it. In this case, it's _probably_ safe, but will possibly allocate more `WCHAR` than necessary.

We're only using it in a few places, but it may be good to refactor it into an internal API that "does the right thing" every time rather than needing to do the math and audit it every time.

src/libmongoc/src/mongoc/mongoc-util-private.h
```diff
@@ -246,11 +246,17 @@ _mongoc_rand_size_t (size_t min, size_t max);
...
 
+// `hex_to_bin` parses `hex` into bytes. Returns NULL on error.
 uint8_t *
-hex_to_bin (const char *hex, uint32_t *len);
+hex_to_bin (const char *hex, size_t *bin_len);
 
+// `bin_to_hex_uppercase` returns uppercase hex for `bin`. Returns NULL on error.
 char *
-bin_to_hex (const uint8_t *bin, uint32_t len);
+bin_to_hex_uppercase (const uint8_t *bin, size_t bin_len);
```
vector-of-bool: Is there a benefit to having separate uppercase/lowercase functions rather than using the `lowercase_inplace` (or `uppercase_inplace`) at the call-sites?

vector-of-bool (approved):
LGTM with minor comments

src/libmongoc/src/mongoc/mongoc-secure-channel.c
kevinAlbs: Neat. Thank you for the suggestion.

src/libmongoc/src/mongoc/mongoc-util-private.h
kevinAlbs: On second thought: removed `hex_to_bin_lowercase` in favor of consistently producing uppercase. The `hex_to_bin_lowercase` was only used for test output.

src/libmongoc/src/mongoc/mongoc-secure-channel.c
kevinAlbs: Refactored to an internal `utf8_to_wide` function which gets the required size first to avoid over-allocating.

I left the calls to `MultiByteToWideChar` in `_mongoc_cluster_sspi_new` unchanged. I was concerned about changing untested code. SSPI is only used for GSSAPI auth, which appears only tested in `run-auth-tests.sh` and currently skipped (pending CDRIVER-5995). Though the new `utf8_to_wide` function could be shared and reused in the future.

PR #2064, opened 2025-07-21
CDRIVER-6053 do not buffer pooled streams

kevinAlbs:
# Summary

Remove buffered stream on pooled connections

# Background & Motivation

See CDRIVER-6053 for a description of the problem. In short: long-lived buffered streams can significantly increase memory usage.

I initially thought buffering was intended to reduce allocations. But it actually increases allocations. When using a buffered stream to read, the caller still [allocates](https://github.com/mongodb/mongo-c-driver/blob/017cf329d7c9bf705cfa1d88eca15e884d13f7dc/src/libmongoc/src/mongoc/mongoc-buffer.c#L185) for the result, then [copies from the buffered stream](https://github.com/mongodb/mongo-c-driver/blob/017cf329d7c9bf705cfa1d88eca15e884d13f7dc/src/libmongoc/src/mongoc/mongoc-stream-buffered.c#L251).

[Comments](https://github.com/mongodb/mongo-c-driver/blob/017cf329d7c9bf705cfa1d88eca15e884d13f7dc/src/libmongoc/src/mongoc/mongoc-stream-buffered.c#L205-L209) indicate the original motivation for buffered streams:

> This isn't actually a huge savings since we never have more than one reply waiting for us, but perhaps someday that will be different. It should help for small replies, however that will reduce our read() syscalls by 50%.

I ran three patch builds of the `Parallel/Pool/*` performance tests to compare with buffering disabled. `Parallel/Pool/*` tests `mongoc_client_pool_t` running a `ping` command repeatedly. The performance [comparison against the stable baseline](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=687e4585eac28993fec65fbc&selected_tab=data-table&percent_filter=0%7C%7C100&z_filter=0%7C%7C10&filter_type=Default), summarized here:

| Test                      | Average Percent Change   |
|---------------------------|--------------------------|
| Parallel/Pool/Threads:1   | 0.39%                    |
| Parallel/Pool/Threads:10  | 0.31%                    |
| Parallel/Pool/Threads:100 | 4.29%                    |

Disabling buffering did not appear significant (and maybe slightly improved?). This PR proposes simply disabling buffered streams on `mongoc_client_pool_t`.

Additionally, single-threaded `mongoc_client_t` does not [enable buffered streams](https://github.com/mongodb/mongo-c-driver/blob/017cf329d7c9bf705cfa1d88eca15e884d13f7dc/src/libmongoc/src/mongoc/mongoc-topology-scanner.c#L835-L852) by default. A stream initiator could be set with `mongoc_client_set_stream_initiator` to create a buffered stream (though I expect this is rare). Regardless, documentation is added to warn about the behavior of buffered stream allocation.

### DIFF
--- a/src/libmongoc/doc/mongoc_stream_buffered_new.rst
+++ b/src/libmongoc/doc/mongoc_stream_buffered_new.rst
@@ -9,8 +9,7 @@ Synopsis
 .. code-block:: c
 
   mongoc_stream_t *
-  mongoc_stream_buffered_new (mongoc_stream_t *base_stream,
-                          size_t buffer_size);
+  mongoc_stream_buffered_new (mongoc_stream_t *base_stream, size_t buffer_size);
 
 Parameters
 ----------
@@ -21,6 +20,10 @@ Parameters
 This function shall create a new :symbol:`mongoc_stream_t` that buffers bytes to and from the underlying ``base_stream``.
 
 ``buffer_size`` will be used as the initial buffer size. It may grow past this size.
+
+.. warning::
+  
+  The internal buffer does not reduce in size once grown. Receiving a large message may result in a large allocation that persists until the returned :symbol:`mongoc_stream_t` is freed with :symbol:`mongoc_stream_destroy()`. The allocation strategy of the internal buffer is subject to change.
 
 Returns
 -------
--- a/src/libmongoc/doc/mongoc_stream_buffered_new.rst
+++ b/src/libmongoc/doc/mongoc_stream_buffered_new.rst
@@ -23,7 +23,7 @@ This function shall create a new :symbol:`mongoc_stream_t` that buffers bytes to
 
 .. warning::
   
-  The internal buffer does not reduce in size once grown. Receiving a large message may result in a large allocation that persists until the returned :symbol:`mongoc_stream_t` is freed with :symbol:`mongoc_stream_destroy()`. The allocation strategy of the internal buffer is subject to change.
+  The internal buffer does not reduce in size once grown. Receiving a large message may result in a large allocation that persists until the returned :symbol:`mongoc_stream_t` is freed with :symbol:`mongoc_stream_destroy()`.
 
 Returns
 -------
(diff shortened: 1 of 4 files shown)


src/libmongoc/doc/mongoc_stream_buffered_new.rst
```diff
@@ -22,6 +21,10 @@ This function shall create a new :symbol:`mongoc_stream_t` that buffers bytes to
 
 ``buffer_size`` will be used as the initial buffer size. It may grow past this size.
 
+.. warning::
+  
+  The internal buffer does not reduce in size once grown. Receiving a large message may result in a large allocation that persists until the returned :symbol:`mongoc_stream_t` is freed with :symbol:`mongoc_stream_destroy()`. The allocation strategy of the internal buffer is subject to change.
```
eramongodb:
```suggestion
  The internal buffer does not reduce in size once grown. Receiving a large message may result in a large allocation that persists until the returned :symbol:`mongoc_stream_t` is freed with :symbol:`mongoc_stream_destroy()`.
```
I don't think this last sentence provides much value.

eramongodb (approved):
Double-check: is there _any_ code path where the user can currently observe a `mongoc_stream_t` that is initialized as a `mongoc_stream_buffered_t` from a `mongoc_client_pool_t`? If so, I believe this is an API breaking change. Otherwise, LGTM. (I see a `mongoc_client_set_stream_initiator` but not a `mongoc_client_get_stream_initiator`, so I _think_ we're safe. The only other `mongoc_stream_t*`-returning API I can find are GridFS API, which do not use buffered streams.)

kevinAlbs:
> is there _any_ code path where the user can currently observe a `mongoc_stream_t` that is initialized as a `mongoc_stream_buffered_t` from a `mongoc_client_pool_t`?

No, I do not see a way to get a `mongoc_stream_t` connected to a MongoDB server from a `mongoc_client_pool_t` with public API.

PR #1682, opened 2024-07-30
Use rand_r in _mongoc_simple_rand

vdimir:

PR #2065, opened 2025-07-21
[CDRIVER-6058] Use thread-safe pseudo-random number generator API

azat:
Refs: https://github.com/mongodb/mongo-c-driver/pull/1682

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-util.c
+++ b/src/libmongoc/src/mongoc/mongoc-util.c
@@ -854,44 +854,51 @@ _mongoc_crypto_rand_size_t (void)
 
 #endif /* defined(MONGOC_ENABLE_CRYPTO) */
 
-static BSON_ONCE_FUN (_mongoc_simple_rand_init)
+#define _mongoc_thread_local BSON_IF_GNU_LIKE (__thread) BSON_IF_MSVC (__declspec (thread))
+
+// Use a thread-local random seed for calls to `rand_r`:
+static _mongoc_thread_local unsigned int _mongoc_simple_rand_seed = 0;
+static _mongoc_thread_local bool _mongoc_simple_rand_seed_initialized = false;
+
+static void
+_mongoc_simple_rand_init (void)
 {
+   if (_mongoc_simple_rand_seed_initialized) {
+      return;
+   }
+   _mongoc_simple_rand_seed_initialized = true;
    struct timeval tv;
-   unsigned int seed = 0;
 
    bson_gettimeofday (&tv);
 
-   seed ^= (unsigned int) tv.tv_sec;
-   seed ^= (unsigned int) tv.tv_usec;
-
-   srand (seed);
-
-   BSON_ONCE_RETURN;
+   _mongoc_simple_rand_seed ^= (unsigned int) tv.tv_sec;
+   _mongoc_simple_rand_seed ^= (unsigned int) tv.tv_usec;
 }
-
-static bson_once_t _mongoc_simple_rand_init_once = BSON_ONCE_INIT;
 
 uint32_t
 _mongoc_simple_rand_uint32_t (void)
 {
-   bson_once (&_mongoc_simple_rand_init_once, _mongoc_simple_rand_init);
+   _mongoc_simple_rand_init ();
 
    /* Ensure *all* bits are random, as RAND_MAX is only required to be at least
     * 32767 (2^15). */
-   return (((uint32_t) rand () & 0x7FFFu) << 0u) | (((uint32_t) rand () & 0x7FFFu) << 15u) |
-          (((uint32_t) rand () & 0x0003u) << 30u);
+   return (((uint32_t) _mongoc_rand_simple (&_mongoc_simple_rand_seed) & 0x7FFFu) << 0u) |
+          (((uint32_t) _mongoc_rand_simple (&_mongoc_simple_rand_seed) & 0x7FFFu) << 15u) |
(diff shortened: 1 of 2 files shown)


src/libmongoc/src/mongoc/mongoc-util.c
```diff
@@ -854,17 +854,17 @@ _mongoc_crypto_rand_size_t (void)
 
 #endif /* defined(MONGOC_ENABLE_CRYPTO) */
 
+static unsigned int _mongoc_simple_rand_seed = 0;
```
alexey-milovidov: For the functions to be thread safe, this variable should be thread local as well.

src/libmongoc/src/mongoc/mongoc-util.c
```diff
@@ -878,8 +878,8 @@ _mongoc_simple_rand_uint32_t (void)
 
    /* Ensure *all* bits are random, as RAND_MAX is only required to be at least
     * 32767 (2^15). */
-   return (((uint32_t) rand () & 0x7FFFu) << 0u) | (((uint32_t) rand () & 0x7FFFu) << 15u) |
-          (((uint32_t) rand () & 0x0003u) << 30u);
+   return (((uint32_t) rand_r (&_mongoc_simple_rand_seed) & 0x7FFFu) << 0u) | (((uint32_t) rand_r (&_mongoc_simple_rand_seed) & 0x7FFFu) << 15u) |
```
kevinAlbs: Suggest using `_mongoc_rand_simple` instead of `rand_r` to fix Windows builds:

```
mongoc-util.c(881): error C4013: 'rand_r' undefined
```

src/libmongoc/src/mongoc/mongoc-util.c
azat: It is initialized via `pthread_once`, so this should be safe.

src/libmongoc/src/mongoc/mongoc-util.c
```diff
@@ -889,9 +889,9 @@ _mongoc_simple_rand_uint64_t (void)
 
    /* Ensure *all* bits are random, as RAND_MAX is only required to be at least
     * 32767 (2^15). */
-   return (((uint64_t) rand () & 0x7FFFu) << 0u) | (((uint64_t) rand () & 0x7FFFu) << 15u) |
-          (((uint64_t) rand () & 0x7FFFu) << 30u) | (((uint64_t) rand () & 0x7FFFu) << 45u) |
-          (((uint64_t) rand () & 0x0003u) << 60u);
+   return (((uint64_t) rand_r (&_mongoc_simple_rand_seed) & 0x7FFFu) << 0u) | (((uint64_t) rand_r (&_mongoc_simple_rand_seed) & 0x7FFFu) << 15u) |
```
kevinAlbs: To fix the failing clang-format task, run:
```sh
uv run tools/format.py
```

src/libmongoc/src/mongoc/mongoc-util.c
alexey-milovidov: But then, `rand_r` called from different threads, cannot use the shared state without synchronization.

kevinAlbs: I expect it is unsafe for two threads to call `rand_r` on the same seed global. [glibc](https://github.com/bminor/glibc/blob/c744519bad81067697600bd01e90b90ae338bf08/stdlib/rand_r.c) and [musl](https://github.com/bminor/glibc/blob/c744519bad81067697600bd01e90b90ae338bf08/stdlib/rand_r.c#L25) appear to operate on the seed directly.

Suggest using a thread-local:
```c
#define _mongoc_thread_local BSON_IF_GNU_LIKE (__thread) BSON_IF_MSVC (__declspec (thread))

// Use a thread-local random seed for calls to `rand_r`:
static _mongoc_thread_local unsigned int _mongoc_simple_rand_seed = 0;
static _mongoc_thread_local bool _mongoc_simple_rand_seed_initialized = false;

static void
_mongoc_simple_rand_init (void)
{
   if (_mongoc_simple_rand_seed_initialized) {
      return;
   }
   _mongoc_simple_rand_seed_initialized = true;
   struct timeval tv;

   bson_gettimeofday (&tv);

   _mongoc_simple_rand_seed ^= (unsigned int) tv.tv_sec;
   _mongoc_simple_rand_seed ^= (unsigned int) tv.tv_usec;
}
```

kevinAlbs (changes requested):
Thank you for the contribution @azat and guidance @alexey-milovidov. Added suggestions.

src/libmongoc/src/mongoc/mongoc-util.c
azat: Thanks! Though it does not nothing to me - https://pastila.nl/?00220984/992b50f96221d5337adc8cf28cc89af2#I1mfyuBcPkXinZfWfwWjpA==

azat: I've manually applied all changes with clang-format

src/libmongoc/src/mongoc/mongoc-util.c
azat: Fixed

src/libmongoc/src/mongoc/mongoc-util.c
azat: My bad, @alexey-milovidov @kevinAlbs thanks! Fixed.

PR #1414, opened 2023-09-27
CDRIVER-4645 migrate and run load balancer tests

kevinAlbs:
# Summary

This PR partially resolves CDRIVER-4645:

- Migrate previously unused `*-loadbalanced-*` tasks to new Evergreen config.
- Run simple_http_server.py on port 18000 to avoid conflict with `haproxy` use of port 8000.
- Update load balanced tests to expect the `hello` monitoring command. https://github.com/mongodb/mongo-c-driver/pull/1256 switched from legacy hello to `hello` when in Load Balancing mode.

New tasks tested in this patch build: https://spruce.mongodb.com/version/65147abba4cf4765a33cf54d

# Background & Motivation

Testing in CI is partially motivated to test a fix specific to load balancers in CI: CDRIVER-4718.

Specifications [describe load balanced test requirements](https://github.com/mongodb/specifications/blob/14916f76fd92b2686d8e3d1f0e4c2d2ef88ca5a7/source/load-balancers/tests/README.rst#testing-requirements)

> For each server version that supports load balanced clusters, drivers MUST add two Evergreen tasks: one with a sharded cluster with both authentication and TLS enabled and one with a sharded cluster with authentication and TLS disabled.

The [Load Balancer Support specification documents](https://github.com/mongodb/specifications/blob/14916f76fd92b2686d8e3d1f0e4c2d2ef88ca5a7/source/load-balancers/load-balancers.rst#load-balancer-support) `Minimum Server Version: 5.0`

## Use of rhel8.7

The previously unused `*-loadbalanced-*` tasks used ubuntu1804. https://wiki.corp.mongodb.com/display/DBDEVPROD/Guidelines+around+Evergreen+distros suggest use of newer distros.

[...]

PR #2070, opened 2025-07-24
CDRIVER-6064 Restore cmake_policy, fix uninstall checks, and use Ninja with mingw-w64

eramongodb:
Followup to https://github.com/mongodb/mongo-c-driver/pull/2066. Identified the following error on closer inspection of the `install-uninstall-check-msvc` task output (despite task success):

```
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.15...4.0 or higher is required.  You are running version 3.11.0
-- Configuring incomplete, errors occurred!
```

This is caused by the `CMAKE` env var being [overridden](https://github.com/mongodb/mongo-c-driver/blob/e8ea70127f7bbbb4bcdb450026f059d8459b98db/.evergreen/scripts/install-uninstall-check-windows.cmd#L13) in the batch script, which ends up using an older system-provided CMake version than expected. The `install-uninstall-*` tasks for MSVC and MinGW have apparently been silently broken for years (😱) since [3165a35](https://github.com/mongodb/mongo-c-driver/commit/3165a356473a1d50d1142d7071bd46bb5bb81c2a) raised the minimum required CMake version from 3.1 to 3.15 in the 1.24.0 release on June 2023. Furthermore, [070b35ab](https://github.com/mongodb/mongo-c-driver/commit/070b35ab163230871c37dd4c8262c05c423e1d54) broke the `uninstall` batch script on Windows (but not in a way that impacts the uninstall operations themselves, as files and directories are still correctly removed), resulting in baffling error messages such as:

```
The filename, directory name, or volume label syntax is incorrect.
The parameter is incorrect.
The parameter is incorrect.
```

These issues will be fixed by a followup PR that resolves CDRIVER-6062. For now, the non-zero exit code for the uninstall command is deliberately ignored.

These issues exposed a severe lack of error-handling checks in the batch scripts, for which there is unfortunately no Bash `set -o errexit` equivalent. This PR updates all batch scripts to consistently use the `|| goto :error` pattern for all potentially-failing commands. Applying these changes revealed a whole host of silently-ignored issues which are now addressed by this PR:

- Unnecessary call to `robocopy` which copies the source directory into the binary directory.
- Obsolete `PATH` manipulation to obtain mingw-w64 GCC which is now reliably provided via Chocolatey.
- Outdated/broken path to `vsvars64.bat` to load the Visual Studio developer environment.
- Revert of removal of call to `cmake_policy()` by https://github.com/mongodb/mongo-c-driver/pull/2066 in `GenerateUninstaller.cmake` after realizing it may be evaluated in CMake script mode during the install step.
    - This in turn exposed [cmake/cmake#26678](https://gitlab.kitware.com/cmake/cmake/-/issues/26678) due to [CMP0147](https://cmake.org/cmake/help/latest/policy/CMP0147.html), which is set to `OLD` to avoid even more noisy diagnostic output.

And as drive-by improvements:

- Add `redirect_standard_error_to_output: true` to prevent confusing log output due to interleaving of stdout and stderr.
- Split `msvc` and `mingw` scripts so the giant `echo.%CC%| findstr /I "gcc">Nul && ( ... ) || ( ... )` is not treated as a single big echoed batch command.
- Updated all mingw-task control paths to use the Ninja generator (as provided by Chocolatey on all latest Windows distros).
    - The `CC` env var is also updated to correctly indicate the _compiler_ to use (`gcc`, not `mingw`), since we are only really interested in testing _compilation_ of our libraries with the mingw-w64 compiler on Windows, not the particulars of the build system of choice.

> [!NOTE]
> The `link-with-bson-mingw` and `link-with-cmake-mingw` tasks are left as effectively no-op tasks due to the lack of `pkg-config` availability on Windows distros.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -4,7 +4,7 @@ libmongoc 2.1.0 (Unreleased)
 ## Fixes
 
 * The uninstall script for Windows is fixed so it no longer "succeeds" with a non-zero `%ERRORLABEL%` value.
-* The uninstall script for non-Windows environments now calls `find "$dir" -maxdepth 1` instead of `ls --almost-all "$dir"` when detecting non-empty directories to fix MacOS compatibility.
+* The uninstall script for non-Windows environments now calls `find "$dir" -maxdepth 1` instead of `ls --almost-all "$dir"` when detecting non-empty directories to fix macOS compatibility.
 
 ## Notes
 
--- a/src/libbson/NEWS
+++ b/src/libbson/NEWS
@@ -4,7 +4,7 @@ libbson 2.1.0 (Unreleased)
 Fixes:
 
 * The uninstall script for Windows is fixed so it no longer "succeeds" with a non-zero `%ERRORLABEL%` value.
-* The uninstall script for non-Windows environments now calls `find "$dir" -maxdepth 1` instead of `ls --almost-all "$dir"` when detecting non-empty directories to fix MacOS compatibility.
+* The uninstall script for non-Windows environments now calls `find "$dir" -maxdepth 1` instead of `ls --almost-all "$dir"` when detecting non-empty directories to fix macOS compatibility.
 
 Changed:
 


.evergreen/scripts/install-uninstall-check-mingw.cmd
```diff
@@ -0,0 +1,180 @@
...
+REM    BSON_ONLY      Whether to build only the BSON library.
+
+rem Ensure Cygwin executables like sh.exe are not in PATH
+rem set PATH=C:\Windows\system32;C:\Windows
+
+echo on
+echo
+
+set SRCROOT=%CD%
+set TAR=C:\cygwin\bin\tar
```
kevinAlbs: Unused.

```suggestion
```

Also applies to `-msvc.cmd` script.

.evergreen/scripts/run-tests.sh
```diff
@@ -168,9 +168,9 @@ wait_for_server() {
   return 1
 }
 
-if [[ "${CC}" =~ gcc ]]; then
+if [[ "${OSTYPE:?}" == cygwin && "${CC}" =~ gcc ]]; then
   echo "Waiting for simple HTTP server to start..."
-  wait_for_server "simple HTTP" 8000
+  wait_for_server "simple HTTP" 18000
```
eramongodb, line 173: Overlooked by https://github.com/mongodb/mongo-c-driver/pull/1414...?

.evergreen/scripts/install-uninstall-check-mingw.cmd
```diff
@@ -0,0 +1,180 @@
...
+
+set version=1.31.0
+set major=1
```
kevinAlbs:
```suggestion
set version=2.1.0
set major=2
```

Not introduced in this PR, but I suspect the out-dated version results in some file checks always passing.

Also applies to `-msvc.cmd` script.

.evergreen/scripts/install-uninstall-check-mingw.cmd
```diff
@@ -0,0 +1,180 @@
...
+   echo %INSTALL_DIR%\lib check ok
+)
+if "%BSON_ONLY%" NEQ "1" (
+   if exist %INSTALL_DIR%\lib\pkgconfig\mongoc%major%.pc (
+     echo mongoc%major%.pc found!
+     goto :error
+   ) else (
+     echo mongoc%major%.pc check ok
+   )
+   if exist %INSTALL_DIR%\lib\cmake\mongoc-1.0\mongoc-1.0-config.cmake (
```
kevinAlbs: Not introduced in this PR: but I expect this path, and others, need to be updated to account for C driver 2.0 paths.

I do not consider it a blocker for this PR, since these paths have not changed. But I think these scripts may benefit to also check that these files exist before the uninstall step. That may help avoid checks for files that were never installed.

For consideration: I _think_ these tests can be rewritten in bash. This appears to work: https://gist.github.com/kevinAlbs/4e9ec25bb0c72ebd02ee48dbbfa38560 but I have not tested in Evergreen.

kevinAlbs (commented):
Improvements are appreciated. My comments are pre-existing issues, so I would not consider them blockers. If not addressed, I can file a ticket to track them separately.

.evergreen/scripts/compile-std.sh
```diff
@@ -187,9 +190,35 @@ else
...
 "${cmake_binary}" -S . -B "${build_dir:?}" "${configure_flags[@]}"
 "${cmake_binary}" --build "${build_dir:?}" --config Debug \
   --target mongo_c_driver_tests \
   --target mongo_c_driver_examples \
   --target public-header-warnings \
   --target "${all_target:?}"
+"${cmake_binary}" --install "${build_dir:?}" --config Debug
+
+# "lib" vs. "lib64"
+lib_dir="$(find "${install_dir:?}" -mindepth 1 -maxdepth 1 -type d -name 'lib*' -printf '%P\n')"
```
kevinAlbs:
```suggestion
lib_dir="lib"
if [ -d "${install_dir}/lib64" ]; then
  lib_dir="lib64"
fi
```

To fix observed error on macOS:
```
find: -printf: unknown primary or operator
```

This further revealed an existing issue in the uninstall generation on macOS: `ls: unrecognized option '--almost-all'`. This appears fixed by replacing in `GenerateUninstaller.cmake`:

```diff
-        list=$(ls --almost-all "$abs")
+        list=$(find "$abs" -mindepth 1)
```

.evergreen/scripts/compile-std.sh
```diff
@@ -22,7 +22,7 @@ script_dir="$(to_absolute "$(dirname "${BASH_SOURCE[0]}")")"
 declare mongoc_dir
 mongoc_dir="$(to_absolute "${script_dir}/../..")"
 
-declare install_dir="${mongoc_dir}/install-dir"
+declare mongoc_install_dir="${mongoc_dir}/install-dir"
```
kevinAlbs: Suggest renaming variables to match installed projects:
- `mongoc_install_dir` to `libmongocrypt_install_dir`
- `install_dir` to `mongoc_install_dir`

.evergreen/scripts/compile-std.sh
```diff
@@ -90,14 +90,17 @@ CXXFLAGS+=" ${flags+${flags[*]}}"
...
+install_dir="cmake-install"
+
+configure_flags_append "-DCMAKE_INSTALL_PREFIX=${install_dir:?}"
 
 # shellcheck source=.evergreen/scripts/add-build-dirs-to-paths.sh
 . "${script_dir}/add-build-dirs-to-paths.sh"
 
 export PKG_CONFIG_PATH
-PKG_CONFIG_PATH="${install_dir}/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
+PKG_CONFIG_PATH="${mongoc_install_dir:?}/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
```
kevinAlbs: Remove? I expect setting `PKG_CONFIG_PATH` is not needed.

kevinAlbs (approved):
LGTM with additional fixes for macOS and variable renames.

And thank you for fixing the existing issues with the test scripts. I am happy to see the batch scripts removed in favor of checks within `compile-std.sh`.

NEWS
```diff
@@ -1,6 +1,11 @@
 libmongoc 2.1.0 (Unreleased)
 ============================
 
+## Fixes
+
+* The uninstall script for Windows is fixed so it no longer "succeeds" with a non-zero `%ERRORLABEL%` value.
+* The uninstall script for non-Windows environments now calls `find "$dir" -maxdepth 1` instead of `ls --almost-all "$dir"` when detecting non-empty directories to fix MacOS compatibility.
```
kevinAlbs:
```suggestion
* The uninstall script for non-Windows environments now calls `find "$dir" -maxdepth 1` instead of `ls --almost-all "$dir"` when detecting non-empty directories to fix macOS compatibility.
```

To match casing on https://en.wikipedia.org/wiki/MacOS

src/libbson/NEWS
```diff
@@ -1,6 +1,11 @@
 libbson 2.1.0 (Unreleased)
 ==========================
 
+Fixes:
+
+* The uninstall script for Windows is fixed so it no longer "succeeds" with a non-zero `%ERRORLABEL%` value.
+* The uninstall script for non-Windows environments now calls `find "$dir" -maxdepth 1` instead of `ls --almost-all "$dir"` when detecting non-empty directories to fix MacOS compatibility.
```
kevinAlbs:
```suggestion
* The uninstall script for non-Windows environments now calls `find "$dir" -maxdepth 1` instead of `ls --almost-all "$dir"` when detecting non-empty directories to fix macOS compatibility.
```

To match casing on https://en.wikipedia.org/wiki/MacOS

kevinAlbs (approved):
LGTM with a NEWS casing change.

PR #2071, opened 2025-07-24
Update generated EVG config for verify-headers

eramongodb:
Followup to https://github.com/mongodb/mongo-c-driver/pull/2053 which missed config regeneration changes for the new verify-headers task to account for https://github.com/mongodb/mongo-c-driver/pull/2058.


kevinAlbs (approved):
Thank you for the fix.

PR #2075, opened 2025-07-30
CDRIVER-6062 fix Windows uninstall batch script behavior 

eramongodb:
Resolves CDRIVER-6062.

> [!NOTE]
> This PR implements [this changelog entry](https://github.com/mongodb/mongo-c-driver/blob/d4e08e5b804d39f51099f9cfe8aa292860e922e3/NEWS#L6) which was prematurely added in https://github.com/mongodb/mongo-c-driver/pull/2070. https://github.com/mongodb/mongo-c-driver/pull/2070 fixes uninstall script errors in our CI, not in the script itself. This PR fixes the script itself.

---

There is a problem with the uninstall batch script (Windows only) that is present since https://github.com/mongodb/mongo-c-driver/pull/1380 (1.25.0). This issue does not affect the _desired behavior_ of the script (all relevant files and directories _are_ being correctly deleted). However, the script completes its execution by emitting confusing error messages and a non-zero error level:

```
> uninstall.bat
Remove file ...\bin\msvcp140.dll - ok
...
Remove directory: ...\include\bson-2.1.0 - ok
The system cannot find the path specified.
The system cannot find the path specified.
The system cannot find the path specified.

> echo %ERRORLEVEL%
1
```

These errors are due to how the [`call`](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/call) command works:

> Calls one batch program from another _**without stopping the parent batch program**_.

> [!NOTE]
> These error messages may not be visible when the uninstall script is executed via the `uninstall` CMake target. These errors are exposed by executing the uninstall script directly.

---

This is the current structure of the uninstall batch script (reduced only to relevant bits):

```bat
call :init

:print
<nul set /p_=%~1
exit /b

... define rmfile and rmdir labels ...

:init

if /i "%~dp0" NEQ "%TEMP%\" (
    set tmpfile=%TEMP%\mongoc-%~nx0
    copy "%~f0" "!tmpfile!" >nul
    call "!tmpfile!" & del "!tmpfile!"
    exit /b
)

... call rmfile and rmdir labels ...
```

First issue: the `call :init` should be a `goto :init`, as otherwise, the script _continues executing_ the current script after `call` completes and falls-through into the `print` label definition.

Second issue: because `call` _continues executing_` the current script, the script cannot be deleted while it is also being executed. https://github.com/mongodb/mongo-c-driver/pull/2070 identified this problem and attempts to workaround this issue by copying the script to a temporary directory:

```bat
rem Is this script in %TEMP%? Then this script is a copy: proceed to rmfile/rmdir commands.
if /i "%~dp0" NEQ "%TEMP%\" (
    rem Otherwise, copy this script to %TEMP%...
    set tmpfile=%TEMP%\mongoc-%~nx0
    copy "%~f0" "!tmpfile!" >nul
    
    rem ... execute the copied script, delete the copied script....
    call "!tmpfile!" & del "!tmpfile!"
    
    rem ... then do nothing more (this is the original script).
    exit /b
)
```

Unfortunately, this has the same problem, but with an extra layer: the script _continues executing_ the current (now-deleted!) uninstall script after `call "!tmpfile!"`. This means the actual "call stack" looks something like this:

```
uninstall.bat
> call :init                         (in uninstall.bat)
| > call %TEMP%\mongoc-uninstall.bat (in uninstall.bat)
| | > call :init                     (in mongoc-uninstall.bat)
| | | > rmdir and rmfile commands    (uninstall.bat is deleted here!)
| | error                            (The system cannot find [mongoc-uninstall.bat])
| error                              (The system cannot find [uninstall.bat])
error:                               (The system cannot find [uninstall.bat]).
```

> [!NOTE]
> AFAICT "without stopping the parent batch program" does not imply parallelism (a la `fork`), although the behavior and output of the batch script sometimes seems to suggest it may be the case. It is very confusing.

---

Thankfully, there is [a technique](https://stackoverflow.com/questions/20329355/how-to-make-a-batch-file-delete-itself/20333575#20333575) which permits a batch script to delete itself without error:

```bat
(GOTO) 2>nul & del "%~f0"
```

Admittedly I do not understand how/why it works, only that it does what we need it to do, which is that `<commands>` in `(GOTO) 2>null & <commands>` are evaluated in the context of the _caller_ rather than in the context of the _batch script_. This means `<commands>` can include the deletion of the batch script itself.

Unfortunately, there are many limitations to the behavior of commands executed in the context of the caller, such as the inability to use ~~`set` (???) or~~ [`call :label`](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/call):

> Call has no effect at the command prompt when it is used outside of a script or batch file.

Additionally, because these commands are evaluated in the user's context, `exit` cannot be used either, as otherwise it will abruptly exit the _user's shell_ on error. Therefore, the new uninstall script expands all file and directory deletion commands inline in a massive chain of ANDs.

The end result is the following structure:

```bat
(GOTO) 2>nul & (
 <nul set /p "=Remove file: %__prefix%\bin\msvcp140.dll "
  if EXIST "%__prefix%\bin\msvcp140.dll" (
    del /Q /F "%__prefix%\bin\msvcp140.dll" && echo  - ok
  ) ELSE echo  - skipped: not present
) && (
  ...
) && (
  <nul set /p "=Remove directory: %__prefix%\include\bson-2.1.0 "
  if EXIST "%__prefix%\include\bson-2.1.0" (
    rmdir /Q "%__prefix%\include\bson-2.1.0" 2>nul && echo  - ok || echo  - skipped ^(non-empty?^)
  ) ELSE echo  - skipped: not present
) && (
  ver>nul
)
```

### DIFF
--- a/build/cmake/GenerateUninstaller.cmake
+++ b/build/cmake/GenerateUninstaller.cmake
@@ -226,7 +226,7 @@ function(add_rmdir dirname)
       set(rmdir_lines
         "  <nul set /p \"=Remove directory: ${dir} \""
         "  if EXIST \"${dir}\" ("
-        "    rmdir /Q \"${dir}\" 2>nul && echo - ok || echo - skipped (non-empty?^)"
+        "    rmdir /Q \"${dir}\" 2>nul && echo - ok || echo - skipped ^(non-empty?^)"
         "  ) ELSE echo - skipped: not present"
         ") && (")
       string(REPLACE ";" "\n" rmdir "${rmdir_lines}")


build/cmake/GenerateUninstaller.cmake
```diff
@@ -258,13 +210,37 @@ append_line("${init}")
...
 # Generate a "remove a directory" command
 function(add_rmdir dirname)
     file(TO_NATIVE_PATH "${dirname}" native)
-    append_line("${__rmdir} '${native}'")
+    if(WIN32)
+      set(dir "%__prefix%\\${native}")
+      set(rmdir_lines
+        "  <nul set /p \"=Remove directory: ${dir} \""
+        "  if EXIST \"${dir}\" ("
+        "    rmdir /Q \"${dir}\" 2>nul && echo - ok || echo - skipped (non-empty?^)"
```
kevinAlbs:
```suggestion
        "    rmdir /Q \"${dir}\" 2>nul && echo - ok || echo - skipped (non-empty?)"
```
Remove unnecessary escape? https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/echo notes:

> When inside a block terminated by parentheses (()), both opening and closing parentheses must also be escaped using the caret (`^`)

However, this `echo` is not inside a block.

build/cmake/GenerateUninstaller.cmake
```diff
@@ -298,4 +274,10 @@ foreach(dir IN LISTS dirs_to_remove)
     add_rmdir("${relpath}")
 endforeach()
 
+# Allow the batch script delete itself without error.
+if(WIN32)
+    append_line("  true")
```
kevinAlbs:
```suggestion
    append_line("  echo Uninstall complete")
```

To fix observed error:
```
'true' is not recognized as an internal or external command,
operable program or batch file.
```

build/cmake/GenerateUninstaller.cmake
```diff
@@ -258,13 +210,37 @@ append_line("${init}")
...
 endfunction()
 
 # Generate a "remove a directory" command
 function(add_rmdir dirname)
     file(TO_NATIVE_PATH "${dirname}" native)
-    append_line("${__rmdir} '${native}'")
+    if(WIN32)
+      set(dir "%__prefix%\\${native}")
+      set(rmdir_lines
+        "  <nul set /p \"=Remove directory: ${dir} \""
```
kevinAlbs, line 227: > such as the inability to use `set`

No change requested. But I expect using `set` in the caller context appears OK. I can use `set` in a command window without running a batch script. I expect the other `echo` commands could be revised to use this pattern, but see no reason to.

kevinAlbs (approved):
Thank you for the fix. LGTM with a small fix.

build/cmake/GenerateUninstaller.cmake
eramongodb: Looks like I was inadvertently depending on Cygwin (EVG) and MSYS2 (local) features which are not supported by pure CMD. Replaced with a call to `ver>nul` (see [here](https://stackoverflow.com/a/1113735) and [here](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/ver)).

build/cmake/GenerateUninstaller.cmake
eramongodb: I may have mixed up "`call` has no effect" with "`set` doesn't work" during prior testing when attempting to use labels. That being said, I think we should refrain from setting any more variables than is strictly necessary given this is being evaluated in the caller's context.

build/cmake/GenerateUninstaller.cmake
eramongodb: It is inside the following blocks marked by `<--`:

```
(GOTO) 2>nul & (
  ...
) && ( <--
  ...
  if EXIST ... ( <--
    ... || echo - skipped ^(non-empty?^)
  ) ELSE echo - skipped: not present
) && (
  ...
)
```

Without escapes, the following error is observed:

```
ELSE was unexpected at this time.
```

rcsanchez97 (approved):

PR #2080, opened 2025-08-01
Fix rpm package build 1.x

rcsanchez97:
Evergreen patch build: https://spruce.mongodb.com/version/688c0516a985180007d06f2a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/.evergreen/etc/mongo-c-driver.spec
+++ b/.evergreen/etc/mongo-c-driver.spec
@@ -10,7 +10,7 @@
 %global gh_project   mongo-c-driver
 %global libname      libmongoc
 %global libver       1.0
-%global up_version   1.30.2
+%global up_version   1.30.5
 #global up_prever    rc0
 # disabled as require a MongoDB server
 %bcond_with          tests
@@ -27,7 +27,7 @@
 Name:      mongo-c-driver
 Summary:   Client library written in C for MongoDB
 Version:   %{up_version}%{?up_prever:~%{up_prever}}
-Release:   1%{?dist}
+Release:   2%{?dist}
 # See THIRD_PARTY_NOTICES
 License:   Apache-2.0 AND ISC AND MIT AND Zlib
 URL:       https://github.com/%{gh_owner}/%{gh_project}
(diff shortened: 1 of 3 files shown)


.evergreen/scripts/build_snapshot_rpm.sh
```diff
@@ -81,7 +81,11 @@ sudo mock -r ${config} --use-bootstrap-image --isolation=simple --install rpmdev
 # This step is needed to avoid the following error on rocky+epel8:
 # Problem: conflicting requests
 #  - package utf8proc-devel-2.6.1-3.module+el8.7.0+1065+42200b2e.aarch64 from powertools is filtered out by modular filtering
-sudo mock -r ${config} --use-bootstrap-image --isolation=simple --dnf-cmd --setopt=powertools.module_hotfixes=true install utf8proc-devel
+if [[ "${config}" == "rocky+epel-8-aarch64" ]]; then
+  sudo mock -r ${config} --use-bootstrap-image --isolation=simple --dnf-cmd --setopt=powertools.module_hotfixes=true install utf8proc-devel
+else
+  sudo mock -r ${config} --use-bootstrap-image --isolation=simple --install utf8proc-devel
```
eramongodb, line 87:
```suggestion
  sudo mock -r ${config} --use-bootstrap-image --isolation=simple install utf8proc-devel
```
Should this `--install` be `install`?

.evergreen/etc/mongo-c-driver.spec
```diff
@@ -27,7 +27,7 @@
 Name:      mongo-c-driver
 Summary:   Client library written in C for MongoDB
 Version:   %{up_version}%{?up_prever:~%{up_prever}}
-Release:   1%{?dist}
+Release:   2%{?dist}
```
eramongodb, line 30: Is this indicating the version of the _spec_ itself rather than the package?

rcsanchez97: Yes, that's exactly right. It corresponds to this changelog entry (later on in the spec file):

```
* Thu Jul 24 2025 Fedora Release Engineering <releng@fedoraproject.org> - 1.30.5-2
- Rebuilt for https://fedoraproject.org/wiki/Fedora_43_Mass_Rebuild
```

.evergreen/scripts/build_snapshot_rpm.sh
rcsanchez97: I don't think this works. The man page for `mock(1)` indicates that the option is supposed to be `--install`. The reason for the bare `install` two lines up is that the `--dnf-cmd` is asking `mock` to invoke `dnf`, as opposed to the `--install` option which tells it to invoke `yum install`.

PR #2082, opened 2025-08-06
CDRIVER-6071 replace build-and-test-with-toolchain with openssl-compat

eramongodb:
Resolves CDRIVER-6071. Removes dependency on the [legacy C toolchain project](https://github.com/10gen/mongo-c-toolchain) by replacing the `build-and-test-with-toolchain` task (which began inexplicably failing since [2025-07-17](https://spruce.mongodb.com/task/mongo_c_driver_smoke_build_and_test_with_toolchain_aa371fab760eebce88a2627c3b55d7de5fe93c9f_25_07_15_19_57_42/logs?execution=1); did the artifact expire?) with a new `openssl-compat` matrix.

---

Historically, the [legacy C toolchain project](https://github.com/10gen/mongo-c-toolchain) was meant to test both SSL library compatibility and build tool minimum requirements, as currently documented by its README:

> This project exists to provide a stable set of tools and libraries at set minimum versions. 

However, its purpose has long been made obsolete or superseded by better alternatives. Support for LibreSSL was dropped in CDRIVER-5693. Boost is irrelevant (C++ Driver doesn't support it since CXX-2797, nor does it use the legacy C toolchain anyways). The minimum required CMake version was raised to 3.15 in CDRIVER-4637 (a new branch as created by https://github.com/10gen/mongo-c-toolchain/pull/2 to support this new requirement); this can easily be replaced with a [`cmake-compat` matrix](https://github.com/mongodb/mongo-cxx-driver/blob/8cd8bdc65d5ee9e0a26e92d907c18f114a81afa4/.evergreen/config_generator/components/cmake_compat.py#L18) (+ soon powered by `uvx cmake` instead of `find-cmake-latest.sh`). The legacy C toolchain also provides zstd, but AFAICT we do not actually test this minimum zstd compatibility requirement, or any other external library dependency's minimum version requirements (we can consider adding additional tasks to accomplish this if needed).

Therefore, the only value (that was) being provided by the `build-and-test-with-toolchain` task is OpenSSL library compatibility checks, which _is_ worth preserving given the wide variety of OpenSSL library versions across platforms and distros. This PR proposes preserving and refining this compatibility coverage as the new `openssl-compat` matrix.

---

First, it is important to note, per OpenSSL [docs](https://openssl-library.org/source/):

> All versions not present above are out of support and should not be used. Users of those versions are encouraged to upgrade to the newest series as soon as possible. Extended support for 1.1.1 and 1.0.2 to gain access to security fixes for those versions is available.

The versions "present above" are currently 3.0 through 3.5. This means we can/should drop compatibility coverage for any version older than 3.0 that is not either 1.0.2 and 1.1.1. Relative to the legacy C toolchain, this means dropping 0.9.8, 1.0.1, and 1.1.0.

> [!NOTE]
> The results of an EVG distro survey on 2025-08-05 report the following distros which are currently used by the C Driver that do not have provide OpenSSL v3 or newer via default system packages:
>
> * `amazon2-arm64-latest`: 1.0.2k FIPS
> * `debian-11`: 1.1.1n
> * `rhel7-latest`: 1.0.2k FIPS
> * `rhel8-latest`: 1.1.1k FIPS
> * `rhel8-power`: 1.1.1k FIPS
> * `rhel8-zseries`: 1.1.1k FIPS
> * `rhel80`: 1.1.1 FIPS
> * `rhel84`: 1.1.1g FIPS
> * `rhel94`: 1.1.1g FIPS
> * `ubuntu2004`: 1.1.1f
> * `ubuntu2004-arm64`: 1.1.1f
>
> These results support the argument for compatibility-checking pre-v3 OpenSSL versions 1.0.2 and 1.1.1 only.

Furthermore, per docs:

> The following OpenSSL version(s) are FIPS validated:
>
> * 3.1.2: FIPS 140-3
> * 3.0.9: FIPS 140-2
> * 3.0.8: FIPS 140-2
> * 3.0.0: FIPS 140-2
>
> Other OpenSSL Releases MAY use the validated FIPS provider, but MUST NOT build and use their own FIPS provider.

Given we have plenty of OpenSSL coverage with other compile-and-test matrices, I believe we do not need to extend compatibility checks for pre-v3 OpenSSL versions more than we already have. That being said, all "special" OpenSSL-related tasks have been relocated from their prior build variants into the `openssl-compat` matrix: all remaining `-openssl` tasks simply use the system-provided OpenSSL library as-is without any special steps.

That being said, even though some system-provided OpenSSL libraries are FIPS-validated, this doesn't necessarily mean the FIPS module is [enabled and used by default](https://docs.openssl.org/master/man7/fips_module/#making-all-applications-use-the-fips-module-by-default). Therefore, the `openssl-compat` matrix includes a few custom OpenSSL installations that _do_ set the appropriate `openssl.cnf` properties to use the FIPS module by default. (That being said, does mongoc even _use_ the FIPS module programmatically at all?) For now, only 3.0.9 and 3.1.2 are used for FIPS 140-2 and FIPS 140-3 coverage respectively (each being the latest patch release for its minor version series).

> [!IMPORTANT]
> This PR deliberately drops [OpenSSL FIPS Object Module 2.0](https://wiki.openssl.org/index.php/FIPS_module_2.0) integration checks. OpenSSL 3.0.9 already provides FIPS 140-2 compliance coverage. Per [Wikipedia](https://en.wikipedia.org/wiki/OpenSSL#FIPS_140_validation), "The FIPS Object Module 2.0 remained FIPS 140-2 validated in several formats until September 1, 2020, when NIST ... designated all non-compliant modules as 'Historical'. This designation includes a caution to federal agencies that they should not include the module in any new procurements." Aside from the distros listed in the survey above which provide FIPS support, I do not think we need to go out of our way to test this any longer.

The Ubuntu 24.04 distro is selected for the `openssl-compat` matrix primarily to avoid complications with registering trusted certificates with the custom OpenSSL installation on RHEL or Arch Linux (no special handling seems to be required on Ubuntu). RHEL is also complicated due to [backports of `EVP_KDF_*` functions](https://github.com/openssl/openssl/issues/11471#issuecomment-609645505) into their OpenSSL 1.1.1 package and [depending on it](https://bugzilla.redhat.com/show_bug.cgi?id=1829790#c9) (e.g. by `libk5crypto.so.3`). Since the goal of this matrix is OpenSSL version compatibility checks, platform-specific OpenSSL compatibility is deferred to other existing `-openssl` tasks.

The `openssl-compat` matrix tasks also call the `run-auth-tests.sh` script via the `run auth tests` EVG command to extend compatibility checks to runtime behavior, not just compile-time compatibility as is the case for `build-and-test-with-toolchain`. This should hopefully allow us to detect issues like CDRIVER-5998 a bit better than before.

---

Numerous fixes and improvements are made to authentication-related tasks and scripts:

- CDRIVER-5995: re-enable auth tests which were being skipped pending DEVPROD-9029, which was resolved in June 2025.
- Fix `authentication-*` and `ocsp-*` task failures due to updated CMake binary directory paths.
    - Update `test-aws-*` scripts for consistency.
- Harden Bash scripts to ensure expanded variables are defined as expected.
    - The `atlas_x509_*` Evergreen project variables required by `run-auth-tests.sh` have been moved into the EVG project repository settings so they are inherited by the `r2.1` release branch.
- Disabled several MacOS (darwinssl) OCSP tasks:
    - MacOS ARM64 only provides MongoDB Server 6.0 or newer. Filtered pre-6.0 versions from the OCSP matrix accordingly.
        - Avoid extending coverage to MacOS Intel for pre-6.0 coverage, as these distro are [very limited resources](https://wiki.corp.mongodb.com/spaces/DBDEVPROD/pages/296299060/Guidelines+for+MacOS+Distros+in+Evergreen).
    - mongoc [does not seem to implement](https://github.com/mongodb/mongo-c-driver/blob/3385ba9c2251ec184e23c12559796ce49d667c88/src/libmongoc/src/mongoc/mongoc-stream-tls-secure-transport.c#L494) support for OCSP soft failure with the Secure Transport library (darwinssl). Does [documentation](https://www.mongodb.com/docs/languages/c/c-driver/current/connect/tls/#server-certificate-verification) need to be updated per [spec](https://github.com/mongodb/specifications/blob/c13d23b91b422b348c54195fe1c49406fc457559/source/ocsp-support/ocsp-support.md#tlsdisableocspendpointcheck)? The `tlsDisableOCSPEndpointCheck` URI option is rejected by both [Secure Transport (darwinssl)](https://github.com/mongodb/mongo-c-driver/blob/3385ba9c2251ec184e23c12559796ce49d667c88/src/libmongoc/src/mongoc/mongoc-stream-tls-secure-transport.c#L697-L704) and [Secure Channel (winssl)](https://github.com/mongodb/mongo-c-driver/blob/3385ba9c2251ec184e23c12559796ce49d667c88/src/libmongoc/src/mongoc/mongoc-stream-tls-secure-channel.c#L1008-L1011), but neither of these behaviors seems to be publically documented.
    - [OCSP Test 3](https://github.com/mongodb/specifications/blob/c13d23b91b422b348c54195fe1c49406fc457559/source/ocsp-support/tests/README.md#integration-tests-permutations-to-be-tested) also consistently fails with the Secure Transport library (darwinssl) despite Test 1 consistently passing, yet they "should be identical" per spec when OCSP stabling is not supported (which I don't think is supported for Secure Transport...?). I've disabled this task for now, but I don't think it should be left disabled.


eramongodb:
> CDRIVER-5990 unskip auth tests for zSeries

Per [this patch build](https://spruce.mongodb.com/version/689a080c5a8c7100071ff12a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC), for zSeries, it seems SASL succeeds, but no-SASL fails during the "Authenticating using GSSAPI" test with:

```
ERROR:      cluster: Authentication failed: SASL Failure: (-1): generic failure
Ping failure: SASL Failure: (-1): generic failure
```

.evergreen/scripts/run-auth-tests.sh
```diff
@@ -190,46 +174,46 @@ if [[ "${ssl}" != "OFF" ]]; then
...
 
 echo "Authenticating using default auth mechanism"
 # Though the auth source is named "mongodb-cr", authentication uses the default mechanism (currently SCRAM-SHA-1).
-LD_LIBRARY_PATH="${openssl_lib_prefix}" maybe_skip "${ping}" "mongodb://${auth_mongodbcr:?}@${auth_host}/mongodb-cr?${c_timeout}"
+"${mongoc_ping:?}" "mongodb://${auth_mongodbcr:?}@${auth_host:?}/mongodb-cr?${c_timeout:?}"
 
 if [[ "${sasl}" != "OFF" ]]; then
   echo "Authenticating using GSSAPI"
-  LD_LIBRARY_PATH="${openssl_lib_prefix}" maybe_skip "${ping}" "mongodb://${auth_gssapi:?}@${auth_host}/?authMechanism=GSSAPI&${c_timeout}"
+  "${mongoc_ping:?}" "mongodb://${auth_gssapi:?}@${auth_host:?}/?authMechanism=GSSAPI&${c_timeout:?}"
```
kevinAlbs: > > CDRIVER-5990 unskip auth tests on zSeries
> 
> Per [this patch build](https://spruce.mongodb.com/version/689a110763173c0007f64e2b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC), zSeries no-SASL succeeds, but zSeries SASL fails during the "Authenticating using GSSAPI" test with:
> 
> ```
> ERROR:      cluster: Authentication failed: SASL Failure: (-1): generic failure
> Ping failure: SASL Failure: (-1): generic failure
> ```

I would rather keep the `maybe_skip` on for zSeries, and leave CDRIVER-5990 unresolved for a further look. I suspect the failure environment related (I don't think GSSAPI code has changed recently).

.evergreen/scripts/openssl-downloader.sh
```diff
@@ -0,0 +1,67 @@
...
+    url="https://github.com/openssl/openssl/releases/download/openssl-${version:?}/openssl-${version:?}.tar.gz"
+  fi
+
+  echo "${url:?}"
+}
+
+# Download the requested OpenSSL version into `openssl-<version>`.
+openssl_download() {
+  declare version
+  version="${1:?"usage: openssl_download_checksum <version>"}"
```
kevinAlbs:
```suggestion
  version="${1:?"usage: openssl_download <version>"}"
```

.evergreen/scripts/run-ocsp-test.sh
```diff
@@ -39,68 +39,78 @@ declare script_dir
...
-if [[ -d "${openssl_install_dir}" ]]; then
-  openssl_lib_prefix="${openssl_install_dir}/lib:${openssl_lib_prefix:-}"
+if [[ -d "${openssl_install_dir:?}" ]]; then
+    if [[ -d "${openssl_install_dir:?}/lib64" ]]; then
+        LD_LIBRARY_PATH="${openssl_install_dir:?}/lib64:${LD_LIBRARY_PATH:-}"
+        DYLD_LIBRARY_PATH="${mongoc_build_dir:?}/src/libmongoc"
+    else
+        LD_LIBRARY_PATH="${openssl_install_dir:?}/lib:${LD_LIBRARY_PATH:-}"
+        DYLD_LIBRARY_PATH="${mongoc_build_dir:?}/src/libbson"
+    fi
```
kevinAlbs: Fix `DYLD_LIBRARY_PATH` overwrites:

```suggestion
    if [[ -d "${openssl_install_dir:?}/lib64" ]]; then
        LD_LIBRARY_PATH="${openssl_install_dir:?}/lib64:${LD_LIBRARY_PATH:-}"
        DYLD_LIBRARY_PATH="${openssl_install_dir:?}/lib64:${DYLD_LIBRARY_PATH:-}"
    else
        LD_LIBRARY_PATH="${openssl_install_dir:?}/lib:${LD_LIBRARY_PATH:-}"
        DYLD_LIBRARY_PATH="${openssl_install_dir:?}/lib:${DYLD_LIBRARY_PATH:-}"
    fi
```

.evergreen/scripts/openssl-compat-check.sh
```diff
@@ -0,0 +1,133 @@
...
+# shellcheck source=.evergreen/scripts/find-cmake-latest.sh
+. "${script_dir}/find-cmake-latest.sh"
+declare cmake_binary
+cmake_binary="$(find_cmake_latest)"
+
+export CMAKE_BUILD_PARALLEL_LEVEL
+CMAKE_BUILD_PARALLEL_LEVEL="$(nproc)"
+
+# libmongocrypt must use the same OpenSSL library.
+echo "Installing libmongocrypt..."
```
kevinAlbs, line 41: Minor: Consider reusing `compile-libmongocrypt.sh` and passing the extra `openssl_cmake_flags` as extra arguments to ``compile_libmongocrypt` to avoid needing to update another location when updating libmongocrypt.

.evergreen/legacy_config_generator/evergreen_config_lib/tasks.py
```diff
@@ -988,16 +867,26 @@ def do_is_valid_combination(self) -> bool:
             # Secure Transport quietly ignores a must-staple certificate with no stapled response.
             prohibit(self.test == "malicious_server_test_2")
 
-        # ECDSA certs can't be loaded (in the PEM format they're stored) on Windows/macOS. Skip them.
+            # Why does this fail with Secure Transport (CSSMERR_TP_CERT_SUSPENDED)...?
+            prohibit(self.test == "TEST_3")
+
+            # Secure Transport (mongoc-stream-tls-secure-transport.c:_verify_peer) does not implement soft failure.
```
kevinAlbs:
```suggestion
            # TODO(CDRIVER-3759) Secure Transport fails soft failure test.
```

IIRC "soft-fail" behavior was only assumed after manual testing, and was assumed to be enabled with [kSecRevocationOCSPMethod](https://github.com/mongodb/mongo-c-driver/blob/3385ba9c2251ec184e23c12559796ce49d667c88/src/libmongoc/src/mongoc/mongoc-stream-tls-secure-transport.c#L525). Quoting https://mongodbcr.appspot.com/547790001/:

> I can at least validate it is contacted (and then a soft-fail occurs)

But maybe the test failures suggest otherwise. I think investigating can be done as part of CDRIVER-3759.

kevinAlbs (commented):
The script clean-ups and test coverage improvements are much appreciated.

> The `tlsDisableOCSPEndpointCheck` URI option is rejected by both [Secure Transport (darwinssl)](https://github.com/mongodb/mongo-c-driver/blob/3385ba9c2251ec184e23c12559796ce49d667c88/src/libmongoc/src/mongoc/mongoc-stream-tls-secure-transport.c#L697-L704) and [Secure Channel (winssl)](https://github.com/mongodb/mongo-c-driver/blob/3385ba9c2251ec184e23c12559796ce49d667c88/src/libmongoc/src/mongoc/mongoc-stream-tls-secure-channel.c#L1008-L1011), but neither of these behaviors seems to be publically documented.

Later sections on [Configure Transport Layer Security (TLS)](https://www.mongodb.com/docs/languages/c/c-driver/current/connect/tls/#server-certificate-verification) note:

> Setting `tlsDisableOCSPEndpointCheck` has no effect.

.evergreen/scripts/run-auth-tests.sh
eramongodb: Reverted as suggested.

.evergreen/scripts/openssl-compat-check.sh
eramongodb: Done. Extended `compile-libmongocrypt.sh` to support additional arguments which are treated as extra CMake flags.

.evergreen/scripts/run-ocsp-test.sh
```diff
@@ -39,68 +39,78 @@ declare script_dir
...
+
 # Custom OpenSSL library may be installed. Only prepend to LD_LIBRARY_PATH when
 # necessary to avoid conflicting with system binary requirements.
-declare openssl_lib_prefix="${LD_LIBRARY_PATH:-}"
-if [[ -d "${openssl_install_dir}" ]]; then
-  openssl_lib_prefix="${openssl_install_dir}/lib:${openssl_lib_prefix:-}"
+if [[ -d "${openssl_install_dir:?}" ]]; then
+    if [[ -d "${openssl_install_dir:?}/lib64" ]]; then
+        LD_LIBRARY_PATH="${openssl_install_dir:?}/lib64:${LD_LIBRARY_PATH:-}"
+        DYLD_LIBRARY_PATH="${openssl_install_dir:?}/lib64:${LD_LIBRARY_PATH:-}"
```
kevinAlbs:
```suggestion
        DYLD_LIBRARY_PATH="${openssl_install_dir:?}/lib64:${DYLD_LIBRARY_PATH:-}"
```

.evergreen/scripts/run-ocsp-test.sh
```diff
@@ -39,68 +39,78 @@ declare script_dir
...
+    else
+        LD_LIBRARY_PATH="${openssl_install_dir:?}/lib:${LD_LIBRARY_PATH:-}"
+        DYLD_LIBRARY_PATH="${openssl_install_dir:?}/lib:${LD_LIBRARY_PATH:-}"
```
kevinAlbs:
```suggestion
        DYLD_LIBRARY_PATH="${openssl_install_dir:?}/lib:${DYLD_LIBRARY_PATH:-}"
```

kevinAlbs (approved):
LGTM with a typo fix.

PR #2083, opened 2025-08-08
Suppress -Wpoison-system-directories for Apple Clang

eramongodb:
This PR replaces `find-cmake-latest.sh` with `uvx cmake` for Earthly tasks. To support `uvx cmake`, this PR also updates the Earthfile to migrate away from old/EOL/unsupported operating systems. A followup PR will migrate the rest of the EVG configuration to use `uv` to obtain preferred build tools as done in https://github.com/mongodb/mongo-cxx-driver/pull/1428.

---

First, concerning Earthly, the following operating system versions are removed due to being EOL:

- Alpine 3.16: EOL since May 2024.
- Alpine 3.17: EOL since Sep 2024.
- Alpine 3.18: EOL since Feb 2025.
- CentOS 7: EOL since June 2024.
- CentOS 8: EOL since May 2024. (Yes, it's [later than](https://wiki.centos.org/About(2f)Product.html) CentOS 7. ¯\\\_ (ツ)\_/¯)
- Ubuntu 16.04: EOL since Apr 2021 (ESM until Apr 2026).
- Ubuntu 18.04: EOL since May 2023 (ESM until Apr 2028).

> [!NOTE]
> Although we do not have our own [platform support policy document](https://docs.google.com/document/d/1SK2iCXJ1rE1PdJgEKrA9jnRSpRhYMjwNEqz6VIP1I6s/edit?tab=t.0#bookmark=id.r732ml8jz9ys), both [Evergreen](https://wiki.corp.mongodb.com/spaces/DBDEVPROD/pages/116563465/Guidelines+around+Evergreen+distros) and [Server](https://docs.google.com/spreadsheets/d/1-sZKW70HbVt2yHOWa18qwBwi-gBcn54tWmronnn89kI/edit?usp=sharing) do _not_ include "Extended Security Maintenance" mode in their range of support. Therefore, I do not think we should include them either.

Second, to avoid the need to copy mongo-c-driver scripts into the Earthly environment, to better test OS compatibility, and to take advantage of Earthly containerization, the OS' package managers are used to acquire `uv` instead of a `uv-installer.sh` script (which will be needed by EVG tasks). This leads to some [complications with using `snap`/`systemd` in containers](https://forum.snapcraft.io/t/please-remove-systemd-dependency/17580/5) (given `snap` is usually the OS-recommended alternative to obtain a missing or newer package than what is available with `apt`). However, `pipx` is available as the [second uv-recommended method](https://docs.astral.sh/uv/getting-started/installation/#pypi) to install `uv` (after `curl | sh`) on all relevant distros as a system package (when `uv` itself is not available via the system package manager), and so it is used instead when needed.

Third, Red Hat apparently [ended support for CentOS](https://www.redhat.com/en/technologies/linux-platforms/enterprise-linux/centos-migration) in 2024 in favor of [CentOS Stream](https://www.redhat.com/en/topics/linux/what-is-centos-stream) (confusingly _not_ the same thing as CentOS, but more or less its equivalent successor). Red Hat strongly recommends against continuing to use CentOS as it is "now exposed to significant risks". As part of this migration, OCI images for CentOS Stream are now provided via [quay.io](https://quay.io/repository/centos/centos) rather than [DockerHub](https://hub.docker.com/_/centos). Therefore, the CentOS Stream images are specified using fully-qualified image names to `quay.io` instead of using `$default_search_registry` (at this time, the DevProd-provided Amazon ECR instance does _not_ seem to support pull-through caching of `quay.io` images, only DockerHub).

---

This PR also includes a drive-by addition of a `-Wpoison-system-directories` suppression with Apple Clang for the `public-header-warnings` target:

```
FAILED: src/CMakeFiles/public-header-warnings.dir/c-check.c.o
/usr/bin/clang -DBSON_STATIC -DMONGOC_STATIC -D_BSD_SOURCE -D_DARWIN_C_SOURCE -D_DEFAULT_SOURCE -D_MLIB_BUILD_CONFIG=Debug -D_XOPEN_SOURCE=700 -Icmake-build/src/libmongoc/src -Isrc/libmongoc/src -Isrc/common/src -Icmake-build/src -Icmake-build/src/common/src -Isrc/libbson/src -Icmake-build/src/libbson/src -m64 -march=x86-64 -Wno-unknown-pragmas -g -std=c99 -Weverything -Werror -Wno-declaration-after-statement -Wno-disabled-macro-expansion -Wno-c++98-compat-pedantic -Wno-pre-c11-compat -Wno-pre-c2x-compat -Wno-unsafe-buffer-usage -Wno-padded -Wno-reserved-identifier -Wno-documentation-unknown-command -Wno-unknown-warning-option -fPIC -MD -MT src/CMakeFiles/public-header-warnings.dir/c-check.c.o -MF src/CMakeFiles/public-header-warnings.dir/c-check.c.o.d -o src/CMakeFiles/public-header-warnings.dir/c-check.c.o -c src/c-check.c
error: include location '/usr/local/include' is unsafe for cross-compilation [-Werror,-Wpoison-system-directories]
1 error generated.
```

This warning is strange. I do not know why the changes in this PR (or the upcoming PR to apply `uv` to EVG scripts) seem to have triggered this false-positive warning. According to [this comment](/AOMediaCodec/libavif/issues/284#issuecomment-678717800), it may be related to the implicit use of the `--sysroot` flag on recent versions of macOS...? We are definitely not attempting any cross-compilation in our macOS EVG tasks (that I am aware of), so I've opted to treat this as a false-positive and disabled it.

### DIFF
--- a/.evergreen/config_generator/components/earthly.py
+++ b/.evergreen/config_generator/components/earthly.py
@@ -27,16 +27,16 @@ _CC_PARAM_NAME = "MONGOC_EARTHLY_C_COMPILER"
 
 
 EnvKey = Literal[
-    "u16",
-    "u18",
     "u20",
     "u22",
-    "alpine3.16",
-    "alpine3.17",
-    "alpine3.18",
+    "u22",
     "alpine3.19",
+    "alpine3.20",
+    "alpine3.21",
+    "alpine3.22",
     "archlinux",
-    "centos7",
+    "centos9",
+    "centos10",
 ]
 "Identifiers for environments. These correspond to special 'env.*' targets in the Earthfile."
 CompilerName = Literal["gcc", "clang"]
@@ -65,12 +65,11 @@ def os_split(env: EnvKey) -> tuple[str, None | str]:
         # Match 'u22', 'u20', 'u71' etc.
         case ubu if mat := re.match(r"u(\d\d)", ubu):
             return "Ubuntu", f"{mat[1]}.04"
-        case "centos7":
-            return "CentOS", "7.0"
+        # Match 'centos9', 'centos10', etc.
+        case cent if mat := re.match(r"centos(\d+)", cent):
+            return "CentOS", f"{mat[1]}"
         case _:
-            raise ValueError(
-                f"Failed to split OS env key {env=} into a name+version pair (unrecognized)"
-            )
+            raise ValueError(f"Failed to split OS env key {env=} into a name+version pair (unrecognized)")
 
 
 class EarthlyVariant(NamedTuple):
@@ -152,12 +151,8 @@ class DockerLoginAmazonECR(Function):
     name = "docker-login-amazon-ecr"
     commands = [
         # Avoid inadvertently using a pre-existing and potentially conflicting Docker config.
-        expansions_update(
-            updates=[KeyValueParam(key="DOCKER_CONFIG", value="${workdir}/.docker")]
-        ),
-        ec2_assume_role(
-            role_arn="arn:aws:iam::901841024863:role/ecr-role-evergreen-ro"
-        ),
+        expansions_update(updates=[KeyValueParam(key="DOCKER_CONFIG", value="${workdir}/.docker")]),
+        ec2_assume_role(role_arn="arn:aws:iam::901841024863:role/ecr-role-evergreen-ro"),
         subprocess_exec(
             binary="bash",
             command_type=EvgCommandType.SETUP,
@@ -181,12 +176,6 @@ def task_filter(env: EarthlyVariant, conf: Configuration) -> bool:
     configuration values.
     """
     match env, conf:
-        # u16/u18/centos7 are not capable of building mongocxx
-        case e, (_sasl, _tls, cxx) if re.match(
-            r"^Ubuntu 16|^Ubuntu 18|^CentOS 7", e.display_name
-        ):
-            # Only build if C++ driver is test is disabled
-            return cxx == "none"
         # Anything else: Allow it to run:
         case _:
             return True
--- a/Earthfile
+++ b/Earthfile
@@ -135,19 +135,10 @@ test-cxx-driver:
 # PREP_CMAKE "warms up" the CMake installation cache for the current environment
 PREP_CMAKE:
     FUNCTION
-    LET scratch=/opt/mongoc-cmake
-    # Copy the minimal amount that we need, as to avoid cache invalidation
-    COPY tools/use.sh tools/platform.sh tools/paths.sh tools/base.sh tools/download.sh \
-        $scratch/tools/
-    COPY .evergreen/scripts/find-cmake-version.sh \
-        .evergreen/scripts/use-tools.sh \
-        .evergreen/scripts/find-cmake-latest.sh \
-        .evergreen/scripts/cmake.sh \
-        $scratch/.evergreen/scripts/
-    # "Install" a shim that runs our managed CMake executable:
-    RUN __alias cmake /opt/mongoc-cmake/.evergreen/scripts/cmake.sh
+    # Run all CMake commands using uvx:
+    RUN __alias cmake uvx cmake
     # Executing any CMake command will warm the cache:
-    RUN cmake --version
+    RUN cmake --version | head -n 1
 
 env-warmup:
     ARG --required env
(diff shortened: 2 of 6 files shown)


.evergreen/config_generator/components/earthly.py
```diff
@@ -27,16 +27,16 @@
...
 
 EnvKey = Literal[
-    "u16",
-    "u18",
     "u20",
     "u22",
-    "alpine3.16",
-    "alpine3.17",
-    "alpine3.18",
+    "u22",
```
kevinAlbs: Remove duplicate:
```suggestion
```

kevinAlbs (approved):
LGTM with one comment. The Earthfile maintenance and improvements are much appreciated.

Earthfile
```diff
@@ -538,6 +531,11 @@ UBUNTU_ENV:
     RUN __install curl build-essential
     ARG --required purpose
 
+    # uv is not available via apt. Avoid snapd (systemd) by using pipx instead.
+    RUN __install python3-venv pipx
+    RUN pipx install uv
+    ENV PATH="/root/.local/bin:$PATH"
```
vector-of-bool: Minor: Can this be modified to install `uv` elsewhere? e.g. set `PIPX_BIN_DIR=/opt/uv/bin` or use `--global` to install to `/usr/local/...`. I don't actually feel strongly about this, though.

.evergreen/config_generator/components/earthly.py
```diff
@@ -27,16 +27,15 @@
...
-    "alpine3.17",
-    "alpine3.18",
     "alpine3.19",
+    "alpine3.20",
+    "alpine3.21",
+    "alpine3.22",
     "archlinux",
-    "centos7",
+    "centos9",
+    "centos10",
```
vector-of-bool, line 41: Following Red Hat's (*wildly unpopular*) decision to kill CentOS, Rocky Linux and AlmaLinux were created to take up the mantle as "drop-in-replacements" for RHEL. It may be better to switch to Rocky or Alma rather than use CentOS Steram as stand-ins RHEL 9+. Rocky and Alma are also available in `docker.io`, allowing us to continue to use ECR.

(RedHat killing CentOS is also the reason why CentOS 8 EOL is earlier then 7: They kicked the project to the curb at version 8 and didn't want to support it any more).

vector-of-bool (changes requested):
Mostly LGTM, but pending comments regarding migrating away from CentOS.

.evergreen/config_generator/components/earthly.py
vector-of-bool: (This isn't a blocking request for changes, just an alternative to consider.)

Earthfile
eramongodb: Updated all `pipx` commands to install to `/opt/uv/bin` instead.

.evergreen/config_generator/components/earthly.py
eramongodb: I think we should keep CentOS Stream coverage for official Red Hat LTS compatibility coverage, but I'm fine with extending coverage to AlmaLinux as well, which cursory research suggests may have better community activity and support than Rocky Linux. Some additional context included below for reference.

Since AlmaLinux is meant to (almost) mirror CentOS Stream, a common `CENTOS_STREAM_ENV` is used to define both `env.centos*` and `env.almalinux*`.

---

Per [Red Hat](https://www.redhat.com/en/blog/furthering-evolution-centos-stream):

> CentOS Stream will now be the sole repository for public RHEL-related source code releases.

Per [AlmaLinux](https://almalinux.org/blog/future-of-almalinux/):

> [Red Hat announced](https://www.redhat.com/en/blog/furthering-evolution-centos-stream) they will no longer be providing the means for downstream clones to continue to be 1:1 binary copies of Red Hat Enterprise Linux (RHEL). [...] After much discussion, the AlmaLinux OS Foundation board today has decided to drop the aim to be 1:1 with RHEL. AlmaLinux OS will instead aim to be binary compatible with RHEL*.

> \* [Binary/ABI compatibility](https://en.wikipedia.org/wiki/Application_binary_interface) in our case means working to ensure that applications built to run on RHEL (or RHEL clones) can run without issue on AlmaLinux. Adjusting to this expectation removes our need to ensure that everything we release is an exact copy of the source code that you would get with RHEL. This includes kernel compatibility and application compatibility.

> The most remarkable potential impact of the change is that we will no longer be held to the line of “bug-for-bug compatibility” with Red Hat, and that means that we can now accept bug fixes outside of Red Hat’s release cycle. While that means some AlmaLinux OS users may encounter bugs that are not in Red Hat, we may also accept patches for bugs that have not yet been accepted upstream, or shipped downstream.

Per [Rocky Linux](https://rockylinux.org/news/keeping-open-source-open):

> Red Hat has recently expressed their perspective that they "[do not find value in a RHEL rebuild](https://www.redhat.com/en/blog/red-hats-commitment-open-source-response-gitcentosorg-changes)." While we believe this view is narrow-minded, Red Hat has taken a strong stance and limited access to the sources for RHEL to only their paying customers.

> Previously, we obtained the source code for Rocky Linux exclusively from the CentOS Git repository [as they recommended](https://developers.redhat.com/blog/2021/02/03/a-guide-for-using-centos-project-code). However, this repository no longer hosts all of the versions corresponding to RHEL. Consequently, we now have to gather the source code from multiple sources, including CentOS Stream, pristine upstream packages, and RHEL SRPMs. [...] we refuse to agree with them, which means we must obtain the SRPMs through channels that adhere to our principles and uphold our rights.

vector-of-bool (approved):
LGTM

PR #2092, opened 2025-08-18
Fix asserting arithmetic on big-endian platforms

vector-of-bool:
This fixes a latent issue on big-endian platforms that was exposed by the merge of #2074, which does an `mlib_assert_mul(int32_t, ...)`.

Broken: If the target type of a ckdint-asserting-arithmetic macro is smaller than `intmax_t`, then the return value of the function will be bogus on big-endian platforms because we neglected to adjust the temporary bits used to store the result.

Fix: If the platform is big-endian, we bit-shift the temporary result down by the number of bytes different between uintmax_t and the destination type size.

> [!note]
> This bug has never manifested in any release, as we never had any asserting-arithmetic that narrowed smaller than `intmax_t`.

### DIFF
--- a/src/common/src/mlib/ckdint.h
+++ b/src/common/src/mlib/ckdint.h
@@ -634,7 +634,7 @@ static inline uintmax_t(_mlib_assert_ckdint)(size_t dst_sz,
       // the lower place values to be in the later bytes. If the target int is
       // smaller than intmax, we must shift all the bits over to their proper
       // position. This expression is trivially constant-folded by an optimizer.
-      tmp >>= (CHAR_BIT * ((sizeof tmp) - dst_sz));
+      tmp >>= ((size_t)CHAR_BIT * ((sizeof tmp) - dst_sz));
    }
    return tmp;
 }


src/common/src/mlib/ckdint.h
```diff
@@ -629,6 +629,13 @@ static inline uintmax_t(_mlib_assert_ckdint)(size_t dst_sz,
               b_str);
       abort();
    }
+   if (!mlib_is_little_endian()) {
+      // We unconditionally set the leading bytes of `tmp`, but big-endian expects
+      // the lower place values to be in the later bytes. If the target int is
+      // smaller than intmax, we must shift all the bits over to their proper
+      // position. This expression is trivially constant-folded by an optimizer.
+      tmp >>= (CHAR_BIT * ((sizeof tmp) - dst_sz));
```
eramongodb:
```suggestion
      tmp >>= ((size_t)CHAR_BIT * ((sizeof tmp) - dst_sz));
```
To ensure signedness consistency.

PR #2097, opened 2025-08-22
Use uvx for CMake and Server Toolchain for Ninja

eramongodb:
Followup to https://github.com/mongodb/mongo-c-driver/pull/2083. Applies similar EVG configuration updates as done in https://github.com/mongodb/mongo-cxx-driver/pull/1428, but with a drive-by update to bump uv's version to 0.8.13 (latest release).

The old calls to `find_python3` and `PYTHON3_BINARY` are removed in favor of always invoking Python via `uv`. All calls to `find-cmake-latest.sh` are now replaced with `install-build-tools.sh` which obtains a latest CMake binary + a ninja binary for (optional) use by CMake. The Ninja generator is now used by default on non-Windows distros.

Batch scripts call `uvx cmake` rather than `uv tool install cmake` -> `cmake` to avoid dealing with `PATH`-related complications. These can/should hopefully be converted into Bash scripts and use Ninja+MSVC in the future...


.evergreen/scripts/patch-uv-installer.sh
kevinAlbs, line 1: I neglected to ask in prior related PRs. Is installing `uv` expected to be removed once DEVPROD-13413 is complete? Or is this intended as a long-term strategy to pin to the version `uv` used for testing?

.evergreen/config_generator/components/clang_format.py
```diff
@@ -20,7 +20,7 @@ class ClangFormat(Function):
             env={
                 "DRYRUN": "1",
             },
-            script="uv run --frozen --only-group=format tools/format.py --mode=check",
+            script='PATH="${UV_INSTALL_DIR}:$PATH" uv run --frozen --only-group=format tools/format.py --mode=check',
```
kevinAlbs, line 24: Add missing call to `InstallUV`?

.evergreen/scripts/patch-uv-installer.sh
eramongodb: > Is installing uv expected to be removed once [DEVPROD-13413](https://jira.mongodb.org/browse/DEVPROD-13413) is complete?

Yes. I would like to defer the responsibility of maintaining the `uv` binary to DevProd rather than maintaining it ourselves. If we do want to have more control over the `uv` binary version, we can/should maintain it via an Evergreen Toolchain project (DEVPROD-19855) instead of CI scripts executed per-build.

vector-of-bool (approved):
Sorry for the very long delay. This LGTM.

PR #2098, opened 2025-08-26
CDRIVER-6055 Audit array allocations

Julia-Garland:
### Motivation

"There are several points in the codebase that use `bson_malloc(sizeof(T) * N) `to allocate arrays of objects. This should not do an in-situ multiplication, since a large value of N will cause integer overflow and result in either an allocation failure or a bogus allocation size. Also, N = 0 can cause issues since malloc(0) is undefined/unspecified."

### Summary
Array allocations of the form `bson_malloc(sizeof(T) * N) ` now use a new function `bson_array_alloc()` which handles multiplication of the two terms internally.  Likewise, equivalent calls to `bson_malloc0(sizeof(T) * N)` now use `bson_array_alloc0()`.

### DIFF
--- a/src/libbson/doc/bson_array_alloc.rst
+++ b/src/libbson/doc/bson_array_alloc.rst
@@ -25,7 +25,7 @@ Description
 
 This is a portable ``malloc()`` wrapper to allocate an array of objects.
 
-If there was a failure to allocate ``num_elems * elem_size`` bytes, the process will be aborted.
+If ``num_elems * elem_size`` cannot be represented in a size_t or there was a failure to allocate ``num_elems * elem_size`` bytes, the process will be aborted.
 
 .. warning::
 
--- a/src/libbson/doc/bson_array_alloc.rst
+++ b/src/libbson/doc/bson_array_alloc.rst
@@ -8,14 +8,17 @@ Synopsis
 
 .. code-block:: c
 
+  #define BSON_ARRAY_ALLOC(Count, Type) \
+     (Type*) bson_array_alloc (Count, sizeof (Type))
+
   void *
-  bson_array_alloc (size_t type_size, size_t num_elems);
+  bson_array_alloc (size_t num_elems, size_t elem_size);
 
 Parameters
 ----------
 
-* ``type_size``: A size_t containing the size in bytes of a single object in the array. 
-* ``num_elems``: A size_t containing the number of objects to be stored in the array.
+* ``num_elems``: A size_t containing the number of objects to allocate.
+* ``elem_size``: A size_t containing the size of each object in bytes.
 
 Description
 -----------
@@ -24,7 +27,7 @@ This is a portable ``malloc()`` wrapper to allocate an array of objects.
 
 In general, this function will return an allocation at least ``sizeof(void*)`` bytes or bigger.
 
-If there was a failure to allocate ``type_size * num_elems`` bytes, the process will be aborted.
+If there was a failure to allocate ``num_elems * elem_size`` bytes, the process will be aborted.
 
 .. warning::
 
--- a/src/libbson/doc/bson_array_alloc.rst
+++ b/src/libbson/doc/bson_array_alloc.rst
@@ -25,8 +25,6 @@ Description
 
 This is a portable ``malloc()`` wrapper to allocate an array of objects.
 
-In general, this function will return an allocation at least ``sizeof(void*)`` bytes or bigger.
-
 If there was a failure to allocate ``num_elems * elem_size`` bytes, the process will be aborted.
 
 .. warning::
--- a/src/libbson/doc/bson_array_alloc.rst
+++ b/src/libbson/doc/bson_array_alloc.rst
@@ -1,7 +1,7 @@
-:man_page: bson_alloc_array0
+:man_page: bson_array_alloc
 
-bson_alloc_array0()
-===================
+bson_array_alloc()
+==================
 
 Synopsis
 --------
@@ -9,7 +9,7 @@ Synopsis
 .. code-block:: c
 
   void *
-  bson_alloc_array0 (size_t type_size, size_t num_elems);
+  bson_array_alloc (size_t type_size, size_t num_elems);
 
 Parameters
 ----------
@@ -20,7 +20,7 @@ Parameters
 Description
 -----------
 
-This is a portable ``malloc()`` wrapper to allocate an array of objects that also sets the memory to zero.
+This is a portable ``malloc()`` wrapper to allocate an array of objects.
 
 In general, this function will return an allocation at least ``sizeof(void*)`` bytes or bigger.
 
@@ -33,4 +33,4 @@ If there was a failure to allocate ``type_size * num_elems`` bytes, the process 
 Returns
 -------
 
-A pointer to a memory region which *HAS* been zeroed.
+A pointer to a memory region which *HAS NOT* been zeroed.
--- a/src/libbson/src/bson/memory.c
+++ b/src/libbson/src/bson/memory.c
@@ -257,7 +257,7 @@ void *
 bson_array_alloc(size_t num_elems /* IN */, size_t elem_size /* IN */)
 {
    void *mem = NULL;
-   size_t num_bytes;
+   size_t num_bytes = 0;
    BSON_ASSERT(!mlib_mul(&num_bytes, num_elems, elem_size));
 
    if (BSON_LIKELY(num_bytes)) {
@@ -292,7 +292,7 @@ void *
 bson_array_alloc0(size_t num_elems /* IN */, size_t elem_size /* IN */)
 {
    void *mem = NULL;
-   size_t num_bytes;
+   size_t num_bytes = 0;
    BSON_ASSERT(!mlib_mul(&num_bytes, num_elems, elem_size));
 
    if (BSON_LIKELY(num_bytes)) {
--- a/src/libbson/src/bson/memory.c
+++ b/src/libbson/src/bson/memory.c
@@ -236,31 +236,30 @@ bson_aligned_alloc0(size_t alignment /* IN */, size_t num_bytes /* IN */)
  *
  * bson_array_alloc --
  *
- *       Allocates memory for an array of objects, checking for cases of
- *       n = 0 and integer overflow in type_size * len, in which case NULL
- *       is returned.
+ *       Allocates memory for an array of objects.
+ *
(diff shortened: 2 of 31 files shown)


src/libbson/src/bson/memory.h
```diff
@@ -48,6 +48,10 @@ bson_aligned_alloc(size_t alignment, size_t num_bytes);
 BSON_EXPORT(void *)
 bson_aligned_alloc0(size_t alignment, size_t num_bytes);
 BSON_EXPORT(void *)
+bson_array_alloc(size_t type_size, size_t num_elems);
+BSON_EXPORT(void *)
+bson_array_alloc0(size_t type_size, size_t num_elems);
```
connorsmacd: This function works very similarly to `calloc`, but the argument order is reversed. Two arguments of the same type already makes this bug-prone, so I think it would be best to change the argument order for both `bson_array_alloc` and `bson_array_alloc0` to align with `calloc` since that's the ordering most C programmers would expect. That said, please let me know if this change would be internally inconsistent with other memory management functions we may have (I'm new here).

vector-of-bool: Since this is very similar to `calloc`, I recommend consolidating these into a single new function `bson_calloc` that uses the `calloc` function on the BSON vmemtable, rather than adding two new public APIs. This will also defer the `calloc` logic to the underlying allocator, which may have more smarts based on object size.

src/libmongoc/src/mongoc/mongoc-set.c
```diff
@@ -27,7 +27,7 @@ mongoc_set_new(size_t nitems, mongoc_set_item_dtor dtor, void *dtor_ctx)
    mongoc_set_t *set = (mongoc_set_t *)bson_malloc(sizeof(*set));
 
    set->items_allocated = BSON_MAX(nitems, 1);
-   set->items = (mongoc_set_item_t *)bson_malloc(sizeof(*set->items) * set->items_allocated);
+   set->items = (mongoc_set_item_t *)bson_array_alloc(sizeof(*set->items), set->items_allocated);
```
vector-of-bool: Since the pattern everywhere is almost always `(T*)bson_array_alloc(sizeof(T), N)`, this is a good time to use a function-like macro:

```c
// Plain function
void* _bson_alloc_n_impl(size_t item_size, size_t count);
// Macro that does the right thing every time:
#define bson_alloc_n(Type, Count) \
  ((Type*)_bson_alloc_n_impl(sizeof(Type), (Count))
```

This also ensures the returned pointer type is correctly used, rather than allowing the implicit-cast from `void*`.

Julia-Garland: Since most of the original function calls I changed were to bson_malloc, not bson_malloc0, would consolidating into a calloc wrapper where memory is always zeroed out be cause for any efficiency concerns?

kevinAlbs: I would rather keep a distinct non-zero alloc for cases that are performance sensitive. For example allocations in mongoc-set.c might be arbitrarily large. Maybe add a `bson_alloc_n0`?

src/libbson/doc/bson_array_alloc.rst
```diff
@@ -0,0 +1,39 @@
...
+
+* ``num_elems``: A size_t containing the number of objects to allocate.
+* ``elem_size``: A size_t containing the size of each object in bytes.
+
+Description
+-----------
+
+This is a portable ``malloc()`` wrapper to allocate an array of objects.
+
+In general, this function will return an allocation at least ``sizeof(void*)`` bytes or bigger.
```
kevinAlbs: I am not sure what is meant by:
> at least ``sizeof(void*)`` bytes

Nothing prevents a caller from requesting a smaller allocation.

Suggest removing this sentence from here and from other documentation.

src/libbson/doc/bson_array_alloc.rst
```diff
@@ -0,0 +1,39 @@
...
+
+If there was a failure to allocate ``num_elems * elem_size`` bytes, the process will be aborted.
```
kevinAlbs: Suggest noting what happens on overflow:

> If ``num_elems * elem_size`` cannot be represented in `size_t` ...

src/libbson/src/bson/memory.c
```diff
@@ -231,6 +231,79 @@ bson_aligned_alloc0(size_t alignment /* IN */, size_t num_bytes /* IN */)
...
+ *       scope of this library to handle so appropriately.
+ *
+ * Parameters:
+ *       @num_elems: The number of objects to allocate.
+ *       @elem_size: The size of each object in bytes.
+ *
+ * Returns:
+ *       A pointer if successful; otherwise abort() is called and this
+ *       function will never return. In the cases of num_elems = 0 or
+ *       integer overflow in num_elems * elem_size, NULL is returned.
```
kevinAlbs: `bson_malloc` aborts on failure to allocate (e.g. due to OOM). I expect many callers of libbson memory functions do not check for a NULL return.

Example:
```c
opts->keyaltnames = BSON_ARRAY_ALLOC(keyaltnames_count, char *);
for (uint32_t i = 0u; i < keyaltnames_count; i++) {
    opts->keyaltnames[i] = bson_strdup(keyaltnames[i]);
}
```
If `NULL` is returned, the loop still tries to write to bad addresses.


I see two options:
1. Return `NULL` on overflow (as is done now). Update callers to check if `NULL` is returned and handle appropriately (return an error, or abort).
2. Abort on overflow.

Arguably an overflow is similar to an OOM failure: you requested more memory than can be allocated. I think for consistency with other allocation functions, the array allocation functions should abort on overflow.

src/libbson/src/bson/memory.c
```diff
@@ -231,6 +231,79 @@ bson_aligned_alloc0(size_t alignment /* IN */, size_t num_bytes /* IN */)
...
+ *--------------------------------------------------------------------------
+ */
+
+void *
+bson_array_alloc(size_t num_elems /* IN */, size_t elem_size /* IN */)
+{
+   void *mem = NULL;
+   size_t num_bytes = num_elems * elem_size;
+
+   if (BSON_LIKELY(num_bytes) && BSON_LIKELY(num_elems == num_bytes / elem_size)) {
```
kevinAlbs: Consider simplifying by using the `mlib_mul` function. `mlib_mul` returns true if overflow occurred:

```c
size_t num_bytes;
BSON_ASSERT(!mlib_mul(&num_bytes, num_elems, elem_size));
```

Suggestion assumes this will be changed to assert, rather than return NULL.

src/libbson/src/bson/memory.c
```diff
@@ -231,6 +231,79 @@ bson_aligned_alloc0(size_t alignment /* IN */, size_t num_bytes /* IN */)
...
+ * Returns:
+ *       A pointer if successful; otherwise abort() is called and this
+ *       function will never return. In the cases of num_elems = 0 or
+ *       integer overflow in num_elems * elem_size, NULL is returned.
+ *
+ *--------------------------------------------------------------------------
+ */
+
+void *
+bson_array_alloc(size_t num_elems /* IN */, size_t elem_size /* IN */)
```
kevinAlbs, line 257: Suggest adding a simple isolated test for array alloc:
```c
static void
test_bson_array_alloc(void)
{
   // Can allocate an array of items:
   {
      int *arr = BSON_ARRAY_ALLOC(2, int);
      arr[0] = 1;
      arr[1] = 2;
      bson_free(arr);
   }

   // Allocating with overflow aborts:
   mlib_assert_aborts () {
      int *arr = BSON_ARRAY_ALLOC(SIZE_MAX, int);
      bson_free(arr);
   }
}
```

src/libbson/src/bson/memory.c
```diff
@@ -22,6 +22,7 @@
 #include <bson/config.h>
 
 #include <mlib/config.h>
+#include <mlib/timer.h>
```
kevinAlbs:
```suggestion
#include <mlib/ckdint.h>
```

To include direct header that defines `mlib_mul`.

src/libbson/doc/bson_array_alloc.rst
vector-of-bool: That comment appears on the other allocation function docs as well. I think it probably comes from someone long ago assuming the unspecified behavior of the underlying system allocator and publishing it as if it were part of our API. I'm not sure why we would be affording such an API guarantee on any of our allocation functions.

Also the fact that our allocating functions abort on failure is very suspicious, but I guess it's always been that way.

src/libbson/src/bson/memory.c
```diff
@@ -231,6 +232,79 @@ bson_aligned_alloc0(size_t alignment /* IN */, size_t num_bytes /* IN */)
...
+ *
+ *--------------------------------------------------------------------------
+ */
+
+void *
+bson_array_alloc(size_t num_elems /* IN */, size_t elem_size /* IN */)
+{
+   void *mem = NULL;
+   size_t num_bytes;
+   BSON_ASSERT(!mlib_mul(&num_bytes, num_elems, elem_size));
```
vector-of-bool: Because of an implementation quirk, `num_bytes` must be initialized before passing its address to `mlib_mul`. (It doesn't matter what value.) (This is the cause of the current task failures on MSVC.)

```suggestion
   size_t num_bytes = 0;
   BSON_ASSERT(!mlib_mul(&num_bytes, num_elems, elem_size));
```

src/libbson/src/bson/memory.c
```diff
@@ -231,6 +232,79 @@ bson_aligned_alloc0(size_t alignment /* IN */, size_t num_bytes /* IN */)
...
+ *       None.
+ *
+ *--------------------------------------------------------------------------
+ */
+
+void *
+bson_array_alloc0(size_t num_elems /* IN */, size_t elem_size /* IN */)
+{
+   void *mem = NULL;
+   size_t num_bytes;
```
vector-of-bool:
```suggestion
   size_t num_bytes = 0;
```

src/libbson/doc/bson_array_alloc.rst
kevinAlbs: Agreed. If we had a fresh start, I would rather allocating functions return an error. But I expect that is a significant change. Aborting on overflow seems like an improvement. I expect it avoids the possibility of an overflow resulting in a smaller-than-requested buffer.

vector-of-bool (approved):
Minor changes, but LGTM.

PR #2100, opened 2025-08-27
CDRIVER-5856 add function to check if bulkwrite is acknowledged

connorsmacd:


src/libmongoc/tests/test-mongoc-bulkwrite.c
```diff
@@ -726,6 +726,91 @@ test_bulkwrite_client_error_no_result(void *unused)
...
+   bool ok;
+
+   mongoc_bulkwrite_t *bw = mongoc_client_bulkwrite_new(client);
+   ok = mongoc_bulkwrite_append_insertone(bw, "db.coll", tmp_bson("{'_id': 123}"), NULL /* opts */, &error);
+   ASSERT_OR_PRINT(ok, error);
+
+   mongoc_bulkwritereturn_t const bwr = mongoc_bulkwrite_execute(bw, opts);
+
+   ASSERT_NO_BULKWRITEEXCEPTION(bwr);
+   ASSERT(mongoc_bulkwrite_is_acknowledged(bw) == is_acknowledged_expected);
```
connorsmacd: It doesn't look like there's a `ASSERT_CMPBOOL` or `ASSERT_EQUAL_BOOL` in `TestSuite.h`, so I did a direct comparison in the assert here. I'd be happy to implement that kind of macro unless anyone has objections.

src/libmongoc/src/mongoc/mongoc-bulkwrite.h
```diff
@@ -255,6 +255,8 @@ mongoc_bulkwrite_set_client(mongoc_bulkwrite_t *self, mongoc_client_t *client);
 // `*session` may be modified when `mongoc_bulkwrite_execute` is called.
 MONGOC_EXPORT(void)
 mongoc_bulkwrite_set_session(mongoc_bulkwrite_t *self, mongoc_client_session_t *session);
+MONGOC_EXPORT(bool)
+mongoc_bulkwrite_is_acknowledged(mongoc_bulkwrite_t const *self);
```
kevinAlbs: Also add a .rst file to document the new public function. Though CONTRIBUTING.md could use an update, it includes commands for [building docs](https://github.com/mongodb/mongo-c-driver/blob/d77a79e4b83bbcc4bc48a02cf443e4879fb36161/CONTRIBUTING.md#documentation):

```
uv run --frozen --with 'sphinx-autobuild' sphinx-autobuild -b html src/libmongoc/doc/ src/libmongoc/doc/html --re-ignore ".*.pickle" --re-ignore ".*.doctree" -j auto
```

src/libmongoc/tests/test-mongoc-bulkwrite.c
kevinAlbs: I support adding a helper macro to print both sides of the comparison on failure.

src/libmongoc/src/mongoc/mongoc-bulkwrite.h
```diff
@@ -258,6 +258,10 @@ mongoc_bulkwrite_set_session(mongoc_bulkwrite_t *self, mongoc_client_session_t *
 // `mongoc_bulkwrite_execute` executes a bulk write operation.
 MONGOC_EXPORT(mongoc_bulkwritereturn_t)
 mongoc_bulkwrite_execute(mongoc_bulkwrite_t *self, const mongoc_bulkwriteopts_t *opts);
+// `mongoc_bulkwrite_is_acknowledged` returns true if the previous call to `mongoc_bulkwrite_execute` used an
+// acknowledged write concern.
+MONGOC_EXPORT(bool)
+mongoc_bulkwrite_is_acknowledged(mongoc_bulkwrite_t const *self);
```
connorsmacd: I suppose we should also cover the case where this is called _before_ the call to `mongoc_bulkwrite_execute`. I can think of a few ways to handle this:

- Document the current behavior, which is to return `false` (simple, but potentially misleading)
- Do the same as above, but document as undefined behavior (also simple, but I generally prefer to specify behavior)
- Add `BSON_ASSERT(self->executed)` to the implementation (specifies behavior in a predictable way, but makes the error non-recoverable)
- Add a `bson_error_t` out-param to capture the error condition where `self->executed` is `false` (consistent with how other bulkwrite functions handle the case where `self->executed` is `true`, but seems like overkill)

Thoughts?

kevinAlbs: I like that idea. I would rather `BSON_ASSERT(self->executed)` or add a `bson_error_t` out-param. I slightly prefer a `bson_error_t` out-param for API consistency at the cost of a more verbose API.

connorsmacd: > I slightly prefer a `bson_error_t` out-param for API consistency at the cost of a more verbose API.

I also prefer this, but it will make this function signature a bit awkward. As far as I can tell, the API always uses `bson_error_t` out-params in conjunction with a `bool` or pointer return type, where `false` or `NULL` indicate the error is set. In this case, since we are already using the `bool` return to indicate whether or not the bulk write is acknowledged, it's not exactly clear how best to handle this.

We could use another `bool` out-param, but this makes it easy to confuse the meanings of the two `bool`s:

```c
MONGOC_EXPORT(bool)
mongoc_bulkwrite_is_acknowledged(mongoc_bulkwrite_t const *self, bool *error_is_set, bson_error_t *error);
```

We could use a tristate, and in that case I would also change the function's name slightly:

```c
typedef enum {
  MONGOC_BULKWRITE_CHECK_ACKNOWLEDGED_ERROR = -1,
  MONGOC_BULKWRITE_UNACKNOWLEDGED,
  MONGOC_BULKWRITE_ACKNOWLEDGED,
} mongoc_bulkwrite_check_acknowledged_t;

MONGOC_EXPORT(mongoc_bulkwrite_check_acknowledged_t)
mongoc_bulkwrite_check_acknowledged(mongoc_bulkwrite_t const *self, bson_error_t *error);
```

Or a struct:

```c
typedef struct {
   bool is_error;
   bool is_acknowledged;
} mongoc_bulkwrite_check_acknowledged_t;

MONGOC_EXPORT(mongoc_bulkwrite_check_acknowledged_t)
mongoc_bulkwrite_check_acknowledged(mongoc_bulkwrite_t const *self, bson_error_t *error);
```

I am just trying to pin down what the standard approach should be here, at the risk of bikeshedding.

kevinAlbs: > As far as I can tell, the API always uses `bson_error_t` out-params in conjunction with a `bool` or pointer return type, where `false` or `NULL` indicate the error is set.

Yes, I think that is a widely used pattern in the C driver.

On a brief look at declarations marked `BSON_EXPORT` and `MONGOC_EXPORT` I do not see functions with bool out-params.

I like the tristate and struct return ideas. I think it simplifies the call (no need to separately declare a short-lived bool). The tristate return might introduce a slight inconsistency with the soon-to-be-added API to get the server ID in CDRIVER-5843. Maybe a struct return could be used for both:

```c
typedef struct {
   bool is_error;
   bool is_acknowledged;
} mongoc_bulkwrite_acknowledged_t;

MONGOC_EXPORT(mongoc_bulkwrite_check_acknowledged_t)
mongoc_bulkwrite_check_acknowledged(mongoc_bulkwrite_t const *self, bson_error_t *error);

typedef struct {
   bool is_error;
   uint32_t server_id;
} mongoc_bulkwrite_serverid_t;

MONGOC_EXPORT(mongoc_bulkwrite_serverid_t)
mongoc_bulkwrite_get_serverid(mongoc_bulkwrite_t const *self, bson_error_t *error);
```

connorsmacd: Good point about CDRIVER-5843 -- I'll need to change my implementation to account for the error case. I think the struct return is the best option for both.

kevinAlbs (approved):

src/libmongoc/tests/TestSuite.h
```diff
@@ -275,6 +275,27 @@ _test_error(const char *format, ...) BSON_GNUC_PRINTF(1, 2);
 #define ASSERT_CMPDOUBLE(a, eq, b) ASSERT_CMPINT_HELPER(a, eq, b, "f", double)
 #define ASSERT_CMPVOID(a, eq, b) ASSERT_CMPINT_HELPER(a, eq, b, "p", void *)
 
+// Returns "true" or "false". The result does _not_ need to be freed.
+char const *
+_bool_to_str(bool b);
```
vector-of-bool: This can be `static inline` since it is very trivial.

vector-of-bool (approved):
LGTM

PR #2101, opened 2025-08-29
CDRIVER-5843 report last-used server ID for unacknowledged writes

connorsmacd:
# Motivation

Normally, we can read the last-used server ID for a bulk write by calling `mongoc_bulkwriteresult_serverid`. For unacknowledged writes, `mongoc_bulkwrite_execute` returns a `NULL` `mongoc_bulkwriteresult_t`, which means there is no way to get the last-used server ID for unacknowledged writes.

# Solution

The bulk write API now provides `mongoc_bulkwrite_serverid` so the last-used server ID can be accessed via the `mongoc_bulkwrite_t` instance as well. When `mongoc_bulkwrite_execute` is called, the last-used server ID is recorded in the `mongoc_bulkwrite_t` instance as well as the `mongoc_bulkwriteresult_t` instance if it is non-null.
 
> [!NOTE] 
> ## API Redundancy: `mongoc_bulkwrite_serverid` and `mongoc_bulkwriteresult_serverid`
> To prevent an API break, `mongoc_bulkwriteresult_serverid` exists alongside the new API function. ~~Since the new function can be used for both acknowledged and unacknowledged writes, we may want to consider deprecating `mongoc_bulkwriteresult_serverid`~~ (we are opting to not deprecate: https://github.com/mongodb/mongo-c-driver/pull/2101#discussion_r2323431377).

> [!IMPORTANT]
> ## No test case for unacknowledged write with a retry
> I had originally intended to include a test case for ensures the last used server ID is reported for unacknowledged writes **with a retry**. However, due to the "fire and forget" nature of unacknowledged writes, we cannot reproduce the necessary conditions with fail points, so I removed the test case for the time being. @kevinAlbs may be able to provide more insight and technical accuracy here.


src/libmongoc/src/mongoc/mongoc-bulkwrite.c
```diff
@@ -157,6 +157,10 @@ struct _mongoc_bulkwrite_t {
    // `executed` is set to true once `mongoc_bulkwrite_execute` is called.
    // `mongoc_bulkwrite_t` may not be executed more than once.
    bool executed;
+   // `serverid` is set in `mongoc_bulkwrite_execute` to identify the last used serverid. For acknowledged writes, this
+   // will be the same as `mongoc_bulkwriteresult_serverid`.
+   uint32_t serverid;
+   bool serverid_is_set;
```
connorsmacd: We currently only have a `bool` option type (`mongoc_optional_t`). We may want to consider writing some macros for creating option types for other primitives if this pattern continues to come up.

src/libmongoc/src/mongoc/mongoc-bulkwrite.c
```diff
@@ -1977,9 +1981,13 @@ mongoc_bulkwrite_execute(mongoc_bulkwrite_t *self, const mongoc_bulkwriteopts_t
...
    bson_destroy(&cmd);
-   if (ret.res && ss) {
-      // Set the returned server ID to the most recently selected server.
-      ret.res->serverid = ss->sd->id;
+   if (ss) {
+      self->serverid = ss->sd->id;
+      self->serverid_is_set = true;
+
+      if (ret.res) {
+         ret.res->serverid = self->serverid;
```
connorsmacd: Since there's no `serverid_is_set` member in `mongoc_bulkwriteresult_t`, it can actually report an invalid server ID (`0`, since `_bulkwriteresult_new` uses `bson_malloc0`) in the case where the `ss` condition above is false. This might be a defect, and perhaps another reason to consider deprecating `mongoc_bulkwriteresult_serverid`.

src/libmongoc/src/mongoc/mongoc-bulkwrite.h
```diff
@@ -269,6 +269,18 @@ typedef struct {
 MONGOC_EXPORT(mongoc_bulkwrite_check_acknowledged_t)
 mongoc_bulkwrite_check_acknowledged(mongoc_bulkwrite_t const *self, bson_error_t *error);
 
+typedef struct {
+   bool is_ok;        // true if no error
+   uint32_t serverid; // the server ID last used in `mongoc_bulkwrite_execute`
+} mongoc_bulkwrite_serverid_maybe_t;
```
connorsmacd: Some Haskell-inspired naming.

Similar my comment [here](https://github.com/mongodb/mongo-c-driver/pull/2101/files#r2322606061), we could implement a `std::expected`-esque monad for these kinds of functions.

src/libmongoc/src/mongoc/mongoc-bulkwrite.c
kevinAlbs: I am not opposed to adding macros for optional types. libmongocrypt has [optional structs](https://github.com/mongodb/libmongocrypt/blob/de39931fb3f6036dcbed04ea7dcc97c93b6ea100/src/mc-optional-private.h#L37) that could possibly be pulled in. If desired, I would rather keep them private until we have a need to expose them.

Though for this PR I am also OK with what is proposed. Consider using a struct to group the fields:
```c
struct {
    bool set;
    uint32_t value;
} serverid;
```
That may be a minor readability improvement to group the fields. And is consistent with other optional fields: [example](https://github.com/connorsmacd/mongo-c-driver/blob/1c226d51e97a2cbccf022c481153ab7ba3bc0a4e/src/libmongoc/src/mongoc/mongoc-client-side-encryption.c#L443-L462).

src/libmongoc/src/mongoc/mongoc-bulkwrite.h
kevinAlbs: If this struct is kept, consider renaming to `mongoc_bulkwrite_serverid_t` for consistency with the recently added `mongoc_bulkwrite_check_acknowledged_t`.

src/libmongoc/src/mongoc/mongoc-bulkwrite.c
kevinAlbs: > This might be a defect

I expect `0` is expected to represent an invalid server ID. Server IDs are [assigned starting at 1](https://github.com/connorsmacd/mongo-c-driver/blob/1c226d51e97a2cbccf022c481153ab7ba3bc0a4e/src/libmongoc/src/mongoc/mongoc-topology-description.c#L1390). And [mongoc_bulk_operation_get_server_id](https://mongoc.org/libmongoc/current/mongoc_bulk_operation_get_server_id.html) (part of the "collection" bulk write API) documents:

> This number is zero until the driver actually uses a server in [mongoc_bulk_operation_execute()](https://mongoc.org/libmongoc/current/mongoc_bulk_operation_execute.html)

I suggest adding a blurb to `mongoc_bulkwriteresult_serverid` to clarify:

> This return value is zero if no server was successfully selected.

---

> perhaps another reason to consider deprecating

I slightly prefer not deprecating. A possible minor caveat with deprecating: `mongoc_bulkwritereturn_t` can outlive `mongoc_bulkwrite_t`. I expect unacknowledged writes are not commonly used. There might be code like:
```c
mongoc_bulkwritereturn_t do_bulk_write ();

int main(void) {
   mongoc_bulkwritereturn_t ret =  do_bulk_write ();
   uint32_t server_id = mongoc_bulkwriteresult_serverid (ret);
   // ...
}
```
If `mongoc_bulkwriteresult_serverid` is deprecated, that code would need to call `mongoc_bulkwrite_serverid` within `do_bulk_write`. That might be a minor hurdle for users and not much benefit to us.

connorsmacd: > Server IDs are assigned starting at 1.

Good to know. I suppose that means that I could have forgone the `bool serverid_is_set` member and instead just used `0` as a sentinel value to indicate the error, but perhaps there is value in having a more explicit API.

I am fine not deprecating given the context you added. The redundancy does not have a significant maintenance burden.

src/libmongoc/src/mongoc/mongoc-bulkwrite.c
connorsmacd: For now, I don't see a pressing need to implement first-class option types as this is one of the only places in the codebase where it has come up (aside from the use of `mongoc_optional_t`). That said, I'll keep this in mind if we continue to see this pattern come up.

src/libmongoc/doc/mongoc_bulkwrite_serverid_t.rst
```diff
@@ -0,0 +1,23 @@
+:man_page: mongoc_bulkwrite_serverid_t
+
+mongoc_bulkwrite_serverid_t
+=================================
```
kevinAlbs: To align:

```suggestion
mongoc_bulkwrite_serverid_t
===========================
```

PR #2112, opened 2025-09-09
CDRIVER-4722 remove cursor support for legacy opcodes

connorsmacd:
Reference: https://jira.mongodb.org/browse/CDRIVER-4722

# Motivation

Now that our minimum wire version is 4.2, we can use exclusively use `OP_MSG` for cursors.

# Changes

- Remove all support for legacy opcodes in cursor implementation and tests.
- Simplify cursor implementation by removing dual code path logic.


src/libmongoc/CMakeLists.txt
```diff
@@ -569,8 +569,6 @@ set (MONGOC_SOURCES
    ${PROJECT_SOURCE_DIR}/src/mongoc/mongoc-cursor-cmd-deprecated.c
    ${PROJECT_SOURCE_DIR}/src/mongoc/mongoc-cursor-find.c
    ${PROJECT_SOURCE_DIR}/src/mongoc/mongoc-cursor-find-cmd.c
-   ${PROJECT_SOURCE_DIR}/src/mongoc/mongoc-cursor-find-opquery.c
```
kevinAlbs, line 572: FYI: before closing CDRIVER-4722, set `PHP Team Attention Needed: Needed`. I expect the file removal will require an update to the PHP driver build listing of source files: https://github.com/mongodb/mongo-php-driver/blob/5bf907601604618556de9698ed26a4b5a4ad54b4/config.m4#L363

src/libmongoc/src/mongoc/mongoc-cursor-find.c
```diff
@@ -44,16 +41,11 @@ _prime(mongoc_cursor_t *cursor)
    if (!server_stream) {
       return DONE;
    }
-   wire_version = server_stream->sd->max_wire_version;
    mongoc_server_stream_cleanup(server_stream);
```
kevinAlbs: Suggest removing `server_stream`. I expect it is no longer needed. The server's reported wire version was previously used to decide between OP_QUERY or OP_MSG. Now OP_MSG is always used.

src/libmongoc/src/mongoc/mongoc-cursor-find.c
```diff
@@ -26,14 +26,11 @@ typedef struct _data_find_t {
 
 extern void
 _mongoc_cursor_impl_find_cmd_init(mongoc_cursor_t *cursor, bson_t *filter);
```
kevinAlbs: Since there is now only one `find` cursor implementation, consider merging `mongoc-cursor-find-cmd.c` into `mongoc-cursor-find.c`.

PR #2113, opened 2025-09-10
CDRIVER-5779 Sync migrated unified tests and client bulkWrite QE test

mdb-ad:
https://spruce.mongodb.com/version/68e0410bf260da000771c7de

### DIFF
--- a/src/common/src/common-macros-private.h
+++ b/src/common/src/common-macros-private.h
@@ -96,4 +96,16 @@
 #define MC_DISABLE_IMPLICIT_WARNING_END
 #endif
 
+// Disable the -Wcast-qual warning
+#if defined(__GNUC__)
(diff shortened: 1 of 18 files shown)


src/libmongoc/tests/unified/entity-map.c
```diff
@@ -1057,6 +1057,11 @@ _parse_kms_provider_aws(
          if (!_append_kms_provider_value_or_getenv(&child, key, value, envvar, error)) {
             return false;
          }
+      } else if (strcmp (key, "sessionToken") == 0) {
+         const char *envvar = "MONGOC_TEST_AWS_TEMP_SESSION_TOKEN";
```
kevinAlbs, line 1066: I think this logic needs to be: if "sessionToken" is present, use `MONGOC_TEST_AWS_TEMP_*` for all three "accessKeyId", "secretAccessKey", and "sessionToken". Like this: https://github.com/kevinAlbs/mongo-c-driver/blob/e5c32351f59eaff50b4d6019225b10ddfa414d93/src/libmongoc/tests/json-test.c#L1410-L1413

PR #2114, opened 2025-09-10
CDRIVER-6091 append C++ compiler to handshake

kevinAlbs:
# Summary

Append detected C++ compiler to handshake.

Verified with this patch build: https://spruce.mongodb.com/version/68c08e74190e6700073c344b

# Background & Motivation

See CDRIVER-6091. Knowing if a C++ compiler was detected may help inform future decisions.

Example from the handshake before:

```json
{
    "driver": {
        "name": "mongoc",
        "version": "2.2.0-pre"
    },
    "os": {
        "type": "Linux",
        "name": "Ubuntu",
        "version": "22.04",
        "architecture": "x86_64"
    },
    "platform": "cfg=0x06235e88e9 posix=200809 stdc=199901 CC=GCC 11.4.0 CFLAGS=\"\" LDFLAGS=\"\""
}
```

This PR adds `CXX` to the `platform` string of the [MongoDB Handshake](https://github.com/mongodb/specifications/blob/eb7f9a253c88b8255d55308b00ebb6e91f5d20c7/source/mongodb-handshake/handshake.md):
```
                                                        vvvvvvvvvvvvvv
cfg=0x06235e88e9 posix=200809 stdc=199901 CC=GCC 11.4.0 CXX=GNU 11.4.0 CFLAGS=\"\" LDFLAGS=\"\"
                                                        ^^^^^^^^^^^^^^
```

If a C++ compiler is not found, `CXX=Unknown` is included. Alternatively: `CXX` could be omitted if not detected. But I expect it may help with analysis to easily distinguish between: "driver too old to check" and "checked and did not find".

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -226,7 +226,7 @@ if(NOT APPLE)
    endif()
 endif()
 
-# Optionally enable C++ to do some C++-specific tests
+# Check for an optional C++ compiler. Result included in handshake metadata. Enables optional C++-specific tests.
 include (CheckLanguage)
 check_language (CXX)
 if (CMAKE_CXX_COMPILER)


src/libmongoc/tests/test-mongoc-handshake.c
```diff
@@ -1342,6 +1343,19 @@ test_mongoc_handshake_race_condition(void)
    _reset_handshake();
 }
 
+static void
+test_mongoc_handshake_cpp(void)
+{
+   bson_t *handshake = _mongoc_handshake_build_doc_with_application("foo");
```
connorsmacd, line 1349:
```suggestion
   const bson_t *handshake = _mongoc_handshake_build_doc_with_application("foo");
```

Nit: Can be `const`

kevinAlbs: Leaving as-is. `_mongoc_handshake_build_doc_with_application` returns a new `bson_t*` expected to be destroyed with `bson_destroy` (which takes a non-const `bson_t *`).

src/libmongoc/CMakeLists.txt
```diff
@@ -752,6 +752,22 @@ configure_file (
...
+set(MONGOC_CXX_COMPILER_ID "Unknown")
+set(MONGOC_CXX_COMPILER_VERSION "")
+
+if(DEFINED CMAKE_CXX_COMPILER_ID)
+    set(MONGOC_CXX_COMPILER_ID "${CMAKE_CXX_COMPILER_ID}")
+endif()
+
+if(DEFINED CMAKE_CXX_COMPILER_VERSION)
+    set(MONGOC_CXX_COMPILER_VERSION "${CMAKE_CXX_COMPILER_VERSION}")
+endif()
```
eramongodb, line 764: Suggest adding a comment to the `check_language (CXX) -> enable_language (CXX)` in the root CMakeLists.txt documenting that these optional compiler ID + version handshake metadata is dependent on them (no longer just for "some C++-specific tests"):

https://github.com/mongodb/mongo-c-driver/blob/340b1f493b6aa2dd463347c3484502dd09aa1d32/CMakeLists.txt#L229-L240

kevinAlbs: Good idea. Comment updated.

PR #2119, opened 2025-09-19
CDRIVER-4489 add `mongoc_oidc_cache_t`

kevinAlbs:
Patch: https://spruce.mongodb.com/version/68cd50d489be4200072d0dd3

# Summary

Add internal OIDC cache: `mongoc_oidc_cache_t`

# Background & Motivation

`mongoc_oidc_cache_t` is an internal component tested in isolation. Subsequent PRs will use `mongoc_oidc_cache_t` to resolve CDRIVER-4689.

`mongoc_oidc_cache_t` implements behaviors described in [Credential Caching](https://github.com/mongodb/specifications/blob/4a46628a9efbc68d58134341b69be7ac1f5fada5/source/auth/auth.md#credential-caching). `mongoc_oidc_cache_t` implements the "Client Cache" (not the "Connection Cache").

`mongoc_oidc_cache_set_usleep_fn` is added to permit a custom sleep function (related to CDRIVER-4736).

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-oidc-cache.c
+++ b/src/libmongoc/src/mongoc/mongoc-oidc-cache.c
@@ -164,7 +164,7 @@ mongoc_oidc_cache_get_token(mongoc_oidc_cache_t *cache, bool *found_in_cache, bs
    // The timeout parameter (when set) is meant to be directly compared against bson_get_monotonic_time(). It is a
    // time point, not a duration.
    mongoc_oidc_callback_params_set_timeout(
-      params, mlib_microseconds_count(mlib_time_add(mlib_now(), mlib_duration(1, min)).time_since_monotonic_start));
+      params, mlib_microseconds_count(mlib_time_add(mlib_now(), mlib_duration(60, s)).time_since_monotonic_start));
 
    // Obtain write-lock:
    {


src/libmongoc/tests/test-mongoc-oidc-cache.c
```diff
@@ -0,0 +1,271 @@
...
+   // Set a callback:
+   {
+      mongoc_oidc_callback_t *cb = mongoc_oidc_callback_new(oidc_callback_fn);
+      mongoc_oidc_callback_set_user_data(cb, &ctx);
+      mongoc_oidc_cache_set_callback(cache, cb);
+      mongoc_oidc_callback_destroy(cb);
+   }
+
+   // Expect callback is called to fetch token:
+   {
```
eramongodb, line 67:
```suggestion
   }

   mlib_time_point const start = mlib_now();

   // Expect callback is called to fetch token:
   {
```
Reduce variable scope + const-correctness.

`start` isn't used until the "waits at least 100ms" test at the very end. Should there be more time-based assertions?

src/libmongoc/src/mongoc/mongoc-oidc-cache-private.h
```diff
@@ -0,0 +1,67 @@
...
+// mongoc_oidc_cache_set_usleep_fn sets a custom sleep function.
+// Not thread safe. Call before any authentication can occur.
+void
+mongoc_oidc_cache_set_usleep_fn(mongoc_oidc_cache_t *cache, mongoc_usleep_func_t usleep_fn, void *usleep_data);
+
+// mongoc_oidc_cache_get_token returns a token or NULL on error. Thread safe.
+// Sets *is_cache to indicate if the returned token came from the cache or callback.
+// Calls sleep if needed to enforce 100ms delay between calls to the callback.
+char *
+mongoc_oidc_cache_get_token(mongoc_oidc_cache_t *cache, bool *is_cache, bson_error_t *error);
```
eramongodb:
```suggestion
mongoc_oidc_cache_get_token(mongoc_oidc_cache_t *cache, bool *found_in_cache, bson_error_t *error);
```
Minor suggestion: "is_cache" -> "found_in_cache".

src/libmongoc/src/mongoc/mongoc-oidc-cache-private.h
```diff
@@ -0,0 +1,67 @@
...
+char *
+mongoc_oidc_cache_get_cached_token(mongoc_oidc_cache_t *cache);
+
+// mongoc_oidc_cache_set_cached_token overwrites the cached token. Useful for tests. Thread safe.
+void
+mongoc_oidc_cache_set_cached_token(mongoc_oidc_cache_t *cache, const char *token);
+
+// mongoc_oidc_cache_invalidate_cached_token invalidates if the cached token matches `token`. Thread safe.
+void
+mongoc_oidc_cache_invalidate_cached_token(mongoc_oidc_cache_t *cache, const char *token);
```
eramongodb:
```suggestion
mongoc_oidc_cache_invalidate_token(mongoc_oidc_cache_t *cache, const char *token);
```
Redundancy in description? If "cached" is a prerequisite, this function might return a `bool` to indicate if the to-be-invalidated token was _not_ in the cache, but I don't think this is necessary.

src/libmongoc/tests/test-mongoc-oidc-cache.c
```diff
@@ -0,0 +1,271 @@
...
+      start = mlib_now();
+      char *token = mongoc_oidc_cache_get_token(cache, &is_cache, &error);
+      ASSERT_OR_PRINT(token, error);
```
eramongodb, line 69:
```suggestion
      ASSERT_OR_PRINT(token, error);
      ASSERT_CMPSTR(token, "placeholder-token");
```
Assert not just that there is _a_ token, but that the token is the one we expect.

src/libmongoc/tests/test-mongoc-oidc-cache.c
```diff
@@ -0,0 +1,271 @@
...
+      ASSERT_CMPINT(ctx.call_count, ==, 1);
+      ASSERT(!is_cache);
+      bson_free(token);
+   }
+
+   // Expect token is cached:
+   {
+      char *token = mongoc_oidc_cache_get_cached_token(cache);
+      ASSERT(token);
```
eramongodb, line 79:
```suggestion
      ASSERT_CMPSTR(token, "placeholder-token");
```
Similar suggestion for other `token` assertions below. Consider defining a `PLACEHOLDER_TOKEN` constant macro?

src/libmongoc/tests/test-mongoc-oidc-cache.c
```diff
@@ -0,0 +1,271 @@
...
+      bson_free(token);
+   }
+
+   // Invalidating same token clears cache:
+   {
+      char *token = mongoc_oidc_cache_get_cached_token(cache);
+      mongoc_oidc_cache_invalidate_cached_token(cache, token);
+      ASSERT(token);
```
eramongodb, line 104:
```suggestion
      ASSERT_CMPSTR(token); // Copy: still valid.
```
I assume this is why the assertion comes after invalidation?

src/libmongoc/tests/test-mongoc-oidc-cache.c
```diff
@@ -0,0 +1,271 @@
...
+      bson_free(token);
+      ASSERT(!mongoc_oidc_cache_get_cached_token(cache));
+   }
+
+   // Expect subsequent call to fetch tokens waits at least 100ms.
+   {
+      char *token = mongoc_oidc_cache_get_token(cache, &is_cache, &error);
+      ASSERT_OR_PRINT(token, error);
+      mlib_duration diff = mlib_time_difference(mlib_now(), start);
+      ASSERT_CMPINT64(mlib_milliseconds_count(diff), >=, 10); // Use shorter time to avoid timing test failures.
```
eramongodb: Why so much leniency? Suggest using a clean cache object to avoid pre-existing timers on the first call to `get_cache()` (using C++ as pseudocode):

```cpp
Cache cache; // cache.ever_called == false

start = now();             // strictly before get_token()
token = cache.get_token(); // cache.last_call = now(); cache.ever_called = true;
token = cache.get_token(); // wait 100ms
stop  = now();             // strictly after get_token()

assert(stop - start >= milliseconds(100)); // >= 100ms by construction.
```

src/libmongoc/src/mongoc/mongoc-oidc-cache-private.h
```diff
@@ -0,0 +1,67 @@
...
+
+// mongoc_oidc_cache_set_callback sets the token callback.
+// Not thread safe. Call before any authentication can occur.
+void
+mongoc_oidc_cache_set_callback(mongoc_oidc_cache_t *cache, const mongoc_oidc_callback_t *cb);
+
+// mongoc_oidc_cache_get_callback gets the token callback.
+// Not thread safe. Call before any authentication can occur.
+const mongoc_oidc_callback_t *
+mongoc_oidc_cache_get_callback(mongoc_oidc_cache_t *cache);
```
eramongodb:
```suggestion
mongoc_oidc_cache_get_callback(const mongoc_oidc_cache_t *cache);
```
Const-correctness.

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
```diff
@@ -0,0 +1,205 @@
...
+      return;
+   }
+   bson_free(cache->token);
+   bson_mutex_destroy(&cache->lock);
+   mongoc_oidc_callback_destroy(cache->callback);
+   bson_free(cache);
+}
+
+char *
+mongoc_oidc_cache_get_cached_token(mongoc_oidc_cache_t *cache)
```
eramongodb:
```suggestion
mongoc_oidc_cache_get_cached_token(const mongoc_oidc_cache_t *cache)
```
Debatable whether or not thread-safe, read-only API should use `const`:

```c
bson_mutex_lock(&((mongoc_oidc_cache_t *)cache)->lock);   // C++ "mutable"
const char *token = bson_strdup(cache->token);            // const access
bson_mutex_unlock(&((mongoc_oidc_cache_t *)cache)->lock); // C++ "mutable"
```

I'm fine with rejecting this suggestion.

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
```diff
@@ -0,0 +1,205 @@
...
+{
+   BSON_ASSERT_PARAM(cache);
+
+   bson_mutex_lock(&cache->lock);
+   const char *token = cache->token;
+   bson_mutex_unlock(&cache->lock);
+   return token ? bson_strdup(token) : NULL;
```
eramongodb:
```suggestion
   bson_mutex_lock(&cache->lock);
   const char *token = bson_strdup(cache->token);
   bson_mutex_unlock(&cache->lock);
   return token;
```
Avoid reading contents of the string outside of the critical section + take advantage of `bson_strdup(NULL) == NULL`.

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
```diff
@@ -0,0 +1,205 @@
...
+
+   bson_mutex_lock(&cache->lock);
+
+   if (cache->token) {
+      bson_free(cache->token);
+      cache->token = NULL;
+   }
+
+   cache->token = token ? bson_strdup(token) : NULL;
+   bson_mutex_unlock(&cache->lock);
```
eramongodb:
```suggestion
   bson_mutex_lock(&cache->lock);
   bson_free(cache->token);
   cache->token = bson_strdup(token);
   bson_mutex_unlock(&cache->lock);
```
Reduce redundant null checks.

Note: I don't think `token == cache->token` is a concern since all `get*_token()` return a copy.

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
```diff
@@ -0,0 +1,205 @@
...
+      *is_cache = true;
+      goto unlock_and_return;
+   }
+
+   mongoc_oidc_callback_params_t *params = mongoc_oidc_callback_params_new();
+   mongoc_oidc_callback_params_set_user_data(params, mongoc_oidc_callback_get_user_data(cache->callback));
+   // From spec: "If CSOT is not applied, then the driver MUST use 1 minute as the timeout."
+   // The timeout parameter (when set) is meant to be directly compared against bson_get_monotonic_time(). It is a
+   // time point, not a duration.
+   mongoc_oidc_callback_params_set_timeout(params, bson_get_monotonic_time() + 60 * 1000 * 1000);
```
eramongodb:
```suggestion
   mongoc_oidc_callback_params_set_timeout(
      params, mlib_microseconds_count(mlib_time_add(mlib_now(), mlib_duration(1, min)).time_since_monotonic_start));
```
`bson_get_monotonic_time()` calls `mlib_now()` since https://github.com/mongodb/mongo-c-driver/commit/4285d76139e0c237a7bb53b55695020b645e06d7, so calling `mlib_now()` directly is valid (they use the same clock and representation). `_mongoc_oidc_callback_params_t::timeout` can/should be refactored to use `mlib_time_point` (not necessarily in this PR), although the public API will still need to return an `int64_t` for comparison with `bson_get_monotonic_time()`.

Perhaps internal documentation should note that `timeout` (OIDC Callback parameter) is in units of "microseconds since epoch".

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
```diff
@@ -0,0 +1,205 @@
...
+struct mongoc_oidc_cache_t {
+   // callback is owned. NULL if unset. Not guarded by lock. Set before requesting tokens.
+   mongoc_oidc_callback_t *callback;
+
+   // usleep_fn is used to sleep between calls to the callback. Not guarded by lock. Set before requesting tokens.
+   mongoc_usleep_func_t usleep_fn;
+   void *usleep_data;
+
+   // lock is used to prevent concurrent calls to callback. Guards access to token, last_called, and ever_called.
+   bson_mutex_t lock;
```
eramongodb: Given the cached token will likely be read more often than written, can this be made a `bson_shared_mutex_t` instead?

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
```diff
@@ -0,0 +1,205 @@
...
+   return oidc;
+}
+
+void
+mongoc_oidc_cache_set_callback(mongoc_oidc_cache_t *cache, const mongoc_oidc_callback_t *cb)
+{
+   BSON_ASSERT_PARAM(cache);
+   BSON_OPTIONAL_PARAM(cb);
+
+   if (cache->callback) {
```
eramongodb, line 66:
```suggestion
   BSON_OPTIONAL_PARAM(cb);

   BSON_ASSERT (!cache->ever_called);

   if (cache->callback) {
```
Precondition: "Not guarded by lock. Set before requesting tokens."

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
```diff
@@ -0,0 +1,205 @@
...
+}
+
+void
+mongoc_oidc_cache_set_usleep_fn(mongoc_oidc_cache_t *cache, mongoc_usleep_func_t usleep_fn, void *usleep_data)
+{
+   BSON_ASSERT_PARAM(cache);
+   BSON_OPTIONAL_PARAM(usleep_fn);
+   BSON_OPTIONAL_PARAM(usleep_data);
+
+   cache->usleep_fn = usleep_fn ? usleep_fn : mongoc_usleep_default_impl;
```
eramongodb, line 89:
```suggestion
   BSON_OPTIONAL_PARAM(usleep_data);

   BSON_ASSERT(!cache->ever_called);

   cache->usleep_fn = usleep_fn ? usleep_fn : mongoc_usleep_default_impl;
```

Precondition: "Not guarded by lock. Set before requesting tokens."

src/libmongoc/src/mongoc/mongoc-oidc-cache-private.h
```diff
@@ -0,0 +1,67 @@
...
+mongoc_oidc_cache_t *
+mongoc_oidc_cache_new(void);
+
+// mongoc_oidc_cache_set_callback sets the token callback.
+// Not thread safe. Call before any authentication can occur.
+void
+mongoc_oidc_cache_set_callback(mongoc_oidc_cache_t *cache, const mongoc_oidc_callback_t *cb);
+
+// mongoc_oidc_cache_get_callback gets the token callback.
+// Not thread safe. Call before any authentication can occur.
```
eramongodb: Is this precondition necessary given it is read-only + `set_callback()` cannot be called after any authentication has occurred?

kevinAlbs: Good point. Removed precondition.

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
kevinAlbs: Updated as suggested to agree with [logical const](https://isocpp.org/wiki/faq/const-correctness#logical-vs-physical-const) recommendation.

src/libmongoc/src/mongoc/mongoc-oidc-cache-private.h
kevinAlbs: Updated name. Having a token already cached is not a prerequisite. Updated comments and added a test of invalidating when no token is cached.

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
kevinAlbs: Oh great idea. I expect the majority of accesses reads. Updated.

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
kevinAlbs: Nice thinking. Updated.

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
kevinAlbs: Oof. Good spot regarding reading outside of the critical section. Updated.

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
kevinAlbs: Updated as suggested, and added internal comment on `_mongoc_oidc_callback_params_t::timeout` to clarify unit.

src/libmongoc/tests/test-mongoc-oidc-cache.c
kevinAlbs: Updated.

> Should there be more time-based assertions?

Added additional assertion before the second callback call to check that the passed time is less than 100ms.

src/libmongoc/tests/test-mongoc-oidc-cache.c
kevinAlbs: Updated here and in other successful calls to `mongoc_oidc_cache_get_token`

src/libmongoc/tests/test-mongoc-oidc-cache.c
kevinAlbs: Updated to add `PLACEHOLDER_TOKEN`.

src/libmongoc/tests/test-mongoc-oidc-cache.c
kevinAlbs: I think I intended to assert that `mongoc_oidc_cache_get_cached_token` returned a non-NULL token. Moved assert to immediately after `mongoc_oidc_cache_get_cached_token`.

src/libmongoc/tests/test-mongoc-oidc-cache.c
kevinAlbs: > Why so much leniency?

The smaller time was in response to observed test-failures on Windows, [example](https://spruce.mongodb.com/task/mongo_c_driver_mingw_windows2016_test_latest_server_ipv6_client_ipv4_noauth_nosasl_nossl_patch_223426fad31a4094c6f9d841f8175d62d2c41737_68cc623c9a562b000811e67e_25_09_18_19_49_26/logs?execution=0):

> Assert Failure: 99 >= 100

I would rather limit time-sensitive tests. Updated from 10ms to 90ms to be closer to expected sleep time and (hopefully) give enough room for time-variance.

> Suggest using a clean cache object

Separated to an isolated test `test_oidc_cache_waits_between_calls`.

src/libmongoc/src/mongoc/mongoc-oidc-cache-private.h
```diff
@@ -34,32 +34,31 @@ void
...
 char *
-mongoc_oidc_cache_get_cached_token(mongoc_oidc_cache_t *cache);
+mongoc_oidc_cache_get_cached_token(const mongoc_oidc_cache_t *cache);
 
 // mongoc_oidc_cache_set_cached_token overwrites the cached token. Useful for tests. Thread safe.
 void
 mongoc_oidc_cache_set_cached_token(mongoc_oidc_cache_t *cache, const char *token);
 
-// mongoc_oidc_cache_invalidate_cached_token invalidates if the cached token matches `token`. Thread safe.
+// mongoc_oidc_cache_invalidate_token invalidates if the cached token (if any) matches `token`. Thread safe.
```
eramongodb:
```suggestion
// mongoc_oidc_cache_invalidate_token invalidates the cached token if it matches `token`. Thread safe.
```
Wording tweak suggestion (a bit too many "if"s).

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
```diff
@@ -92,20 +96,21 @@ mongoc_oidc_cache_destroy(mongoc_oidc_cache_t *cache)
...
    BSON_ASSERT_PARAM(cache);
 
-   bson_mutex_lock(&cache->lock);
-   const char *token = cache->token;
-   bson_mutex_unlock(&cache->lock);
-   return token ? bson_strdup(token) : NULL;
+   // Cast away const to lock. This function is logically const (read-only).
+   bson_shared_mutex_lock_shared(&((mongoc_oidc_cache_t *)cache)->lock);
+   char *token = bson_strdup(cache->token);
+   bson_shared_mutex_unlock_shared(&((mongoc_oidc_cache_t *)cache)->lock);
```
eramongodb, line 114: It looks like this unfortunately triggers some Clang-specific warnings:

```
mongoc-oidc-cache.c:110:59: warning: cast from 'const struct mongoc_oidc_cache_t *' to 'struct mongoc_oidc_cache_t *' drops const qualifier [-Wcast-qual]
      |    bson_shared_mutex_lock_shared(&((mongoc_oidc_cache_t *)cache)->lock);
      |                                                           ^
```

Consider defining and using a new set of `MC_DISABLE_CAST_QUAL_(BEGIN|END)` macros.

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
```diff
@@ -0,0 +1,215 @@
...
+{
+   BSON_ASSERT_PARAM(cache);
+   BSON_OPTIONAL_PARAM(token);
+
+   bson_shared_mutex_lock(&cache->lock);
+
+   bson_free(cache->token);
+   cache->token = NULL;
+   cache->token = bson_strdup(token);
+   bson_shared_mutex_unlock(&cache->lock);
```
eramongodb:
```suggestion
   bson_shared_mutex_lock(&cache->lock);
   char *const old_token = cache->token;
   cache->token = bson_strdup(token);
   bson_shared_mutex_unlock(&cache->lock);

   bson_free(old_token);
```
Minor optimization: move deallocation out of critical zone + remove redundant assignment.

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
```diff
@@ -0,0 +1,215 @@
...
+      goto unlock_and_return;
+   }
+
+   mongoc_oidc_callback_params_t *params = mongoc_oidc_callback_params_new();
+   mongoc_oidc_callback_params_set_user_data(params, mongoc_oidc_callback_get_user_data(cache->callback));
+   // From spec: "If CSOT is not applied, then the driver MUST use 1 minute as the timeout."
+   // The timeout parameter (when set) is meant to be directly compared against bson_get_monotonic_time(). It is a
+   // time point, not a duration.
+   mongoc_oidc_callback_params_set_timeout(
+      params, mlib_microseconds_count(mlib_time_add(mlib_now(), mlib_duration(1, min)).time_since_monotonic_start));
```
eramongodb: Minor optimization: move callback param object initialization up and out of the critical section.

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
```diff
@@ -0,0 +1,215 @@
...
+         cache->usleep_fn(mlib_microseconds_count(to_sleep), cache->usleep_data);
+      }
+   }
+
+   // Call callback:
+   cred = mongoc_oidc_callback_get_fn(cache->callback)(params);
+
+   cache->last_called = mlib_now();
+   cache->ever_called = true;
+   mongoc_oidc_callback_params_destroy(params);
```
eramongodb: Minor optimization: move callback param object deallocation down and out of the critical section.

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
```diff
@@ -0,0 +1,215 @@
...
+   if (!cred) {
+      SET_ERROR("MONGODB-OIDC callback failed");
+      goto unlock_and_return;
+   }
+
+   token = bson_strdup(mongoc_oidc_credential_get_access_token(cred));
+   cache->token = bson_strdup(token); // Cache a copy.
+
+unlock_and_return:
+   bson_shared_mutex_unlock(&cache->lock);
```
eramongodb: Minor suggestion: consider scoping critical sections with `{}`:

```c
{
   // Obtain write-lock:
   bson_shared_mutex_lock(&cache->lock);

   ...

unlock_and_return:
   bson_shared_mutex_unlock(&cache->lock);
}
```

eramongodb (changes requested):

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
kevinAlbs: Updated to add and use `MC_DISABLE_CAST_QUAL_WARNING_(BEGIN|END)` macros.

src/libmongoc/src/mongoc/mongoc-oidc-cache.c
kevinAlbs: Updated here and in `mongoc_oidc_cache_invalidate_token`.

eramongodb (approved):

PR #2128, opened 2025-09-29
CDRIVER-6105 do not propagate `-fPIC`

kevinAlbs:
Patch authored by @rcsanchez97. As I understand it, this is needed to avoid propagating `-fPIC` to consumers. This is needed to fix upcoming Debian packaging.


kevinAlbs (commented):
> Note: `kms_message` sources will need to be synced with [mongodb/libmongocrypt#1079](https://github.com/mongodb/libmongocrypt/pull/1079).

Good catch. Updated to bundled kms_message to 9fff64216c06099401e2b3b2d5becb77bc17803d.

PR #2135, opened 2025-10-03
[CDRIVER-6107] Add Test Case Tag Support

vector-of-bool:
Refer: CDRIVER-6107

This changeset introduces the ability to apply label/tags to test cases, and exposes those as CTest labels.

This also includes new general mutable string and array/vector handling utilities, and refactors a lot of our test suite to use those.

## Changes:

- 6d908aa7 supports a `because` clause on all `mlib_check` assertions. This message is included in the output when the program terminates, and acts as inline documentation for why a particular assertion is present.
- d636e33d Brings `mstr`, the data-owning counterpart to `mstr_view`. This piece has been in a stash for a while, and was brought out as a better way to handle mutable strings.
- 9675ae5b Adds `vec.t.h`, a generic template header that is used to define vector types. Basically:
  - `#define` a type `T`, and then `#include <mlib/vec.t.h>`, and it will define a `T_vec` type that acts as a dynamically sized contiguous array of `T` objects.
  - Additional macros can be defined to add copy/destroy semantics to the vector objects.
- f97883e0 Adds `mstr_trim` for removing whitespace from string views.
- 47629d7a Refactors a lot of our test suite code to use `mtsr` and `vec.t.h` types. Also adds `mlib/str_vec.h` which just defines the `mstr_vec` for the very common "array of strings" type.
- b39b7957 Adds support for specifying tags/labels associated with test cases:
  - The tags are specified within the same string as the test name. (There was an attempt to add a second string parameter for tags, but this was extremely cumbersome to update all of the test cases across the codebase to pass a second empty string, especially with all the different signatures for adding test cases. A future refactor may want to consolidate our many "TestSuite_Add" functions/macros.)
  - After the test case name, place one or more spaces, and then a list of `[bracketed][tags]`.
  - To support exporting of tags for `LoadTests.cmake`, the `--tests-cmake` switch will print CMake code that defines all the test cases. I wanted to use JSON to emit the test cases, but CMake's JSON handling functionality is incredibly slow for very large JSON blobs, so directly emitting CMake code was chosen instead.
  - `LoadTests.cmake` will evaluate the emitted CMake code and use it to call `add_test` for all the test cases, as well as apply the test case labels and fixtures.
  - If a test has a tag `[uses:foo]`, then `LoadTests.cmake` will add a `FIXTURES_REQUIRED` of `mongoc/fixtures/foo`. This currently only applies to the IMDS tests, but will eventually be used for other test case fixtures.

### DIFF
--- a/src/common/src/mlib/str.h
+++ b/src/common/src/mlib/str.h
@@ -1117,9 +1117,9 @@ mstr_replace(mstr *str, mstr_view needle, mstr_view sub)
 }
 
 /**
- * @brief Like `mstr_snprintf`, but accepts a `va_list` directly.
+ * @brief Like `mstr_sprintf`, but accepts a `va_list` directly.
  */
-mlib_printf_attribute(1, 0) static inline mstr mstr_vsnprintf(const char *format, va_list args)
+mlib_printf_attribute(1, 0) static inline mstr mstr_vsprintf(const char *format, va_list args)
 {
    size_t format_strlen = strlen(format);
    size_t sz = format_strlen;
@@ -1182,7 +1182,7 @@ mlib_printf_attribute(1, 2) static inline mstr mstr_sprintf(const char *f, ...)
 {
    va_list args;
    va_start(args, f);
-   mstr ret = mstr_vsnprintf(f, args);
+   mstr ret = mstr_vsprintf(f, args);
    va_end(args);
    return ret;
 }
@@ -1193,7 +1193,7 @@ mlib_printf_attribute(1, 2) static inline mstr mstr_sprintf(const char *f, ...)
 mlib_printf_attribute(2, 0) static inline bool mstr_vsprintf_append(mstr *string, const char *format, va_list args)
 {
    mlib_check(string != NULL, because, "Output string parameter is required");
-   mstr suffix = mstr_vsnprintf(format, args);
+   mstr suffix = mstr_vsprintf(format, args);
    bool ok = mstr_append(string, suffix);
    mstr_destroy(&suffix);
    return ok;


src/libmongoc/tests/TestSuite.c
eramongodb:
```suggestion
         execle(suite->prgname, suite->prgname, "--no-fork", "--silent", "-l", test->name.data, (char *)0, envp);
```
`test->name` is now `mstr` instead of `char*`:
```
warning: passing object of class type 'mstr' (aka 'struct mstr') through variadic function [-Wclass-varargs]
  486 |          execle(suite->prgname, suite->prgname, "--no-fork", "--silent", "-l", test->name, (char *)0, envp);
      |                                                                                ^
```

I would prefer to address this warning instead of depending on "layout compatibility" through a variadic function parameter.

eramongodb:
```suggestion
         execl(suite->prgname, suite->prgname, "--no-fork", "-l", test->name.data, (char *)0);
```
`test->name` is now `mstr` instead of `char*`:
```
warning: passing object of class type 'mstr' (aka 'struct mstr') through variadic function [-Wclass-varargs]
  493 |          execl(suite->prgname, suite->prgname, "--no-fork", "-l", test->name, (char *)0);
      |                                                                   ^
```

I would prefer to address this warning instead of depending on "layout compatibility" through a variadic function parameter.

src/libmongoc/src/mongoc/mongoc-openssl-private.h
```diff
@@ -25,6 +25,8 @@
 
 #include <bson/bson.h>
 
+#if defined(MONGOC_ENABLE_SSL_OPENSSL)
```
eramongodb:
```suggestion
#include <mongoc/mongoc-config.h>

//

#if defined(MONGOC_ENABLE_SSL_OPENSSL)
```

`<mongoc/mongoc-config.h>` must be included for `defined(MONGOC_ENABLE_SSL_OPENSSL)` to behave as intended.

If this was prompted by the fix to `_test_topology_scanner`, for consistency, suggest picking one of the following:

- follow existing pattern of guarding the `#include` in `test-mongoc-topology-scanner.c`, or
- update other `#include` of this header to remove now-redundant guards.

src/common/src/mlib/str.h
```diff
@@ -565,4 +642,488 @@ mstr_contains_any_of(mstr_view str, mstr_view needle)
...
+mstr_sprintf(const char *f, ...)
+{
+   mstr ret = mstr_null();
+   va_list args;
+   va_start(args, f);
+   const bool okay = mstr_vsprintf_append(&ret, f, args);
+   va_end(args);
+   if (!okay) {
+      mstr_destroy(&ret);
+      return mstr_null();
```
eramongodb:
```suggestion
      ret = mstr_null();
```

Address the following Clang warning:

```
warning: not eliding copy on return [-Wnrvo]
1126 |    return ret;
     |           ^
```

src/common/src/mlib/vec.th
eramongodb, line 1: I prefer we avoid introducing any more "X macro headers" if we can help it.

This may be a large ask, but I believe we should be able to apply the "template" pattern used by [`MONGOC_DECL_SPECIAL_TS_POOL`](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/mongoc-ts-pool-private.h#L265) to implement the following API instead:

```c
#ifndef MLIB_VEC_H_INCLUDED
#define MLIB_VEC_H_INCLUDED

... // Implementation details.

#define mlib_vec_declare(_name, _type)     \
  typedef struct MLIB_PASTE(_name, _vec) { \
    _type *data;                           \
    ...                                    \
  };                                       \
  ...

#endif // MLIB_VEC_H_INCLUDED
```

```c
#include <mlib/vec.h> // Normal header.

...

#define TestSkipVec_DestroyElement(Skip) TestSkip_Destroy(Skip)

mlib_vec_declare(TestSkipVec, TestSkip)
```

I expect conditionally-defined API (e.g. `VecDestroyElement(Ptr)`) can be implemented using a pattern similar to `bsonPredicate`, but where the "not defined" case is ignored rather than an error: https://github.com/mongodb/mongo-c-driver/blob/b4984965877d559862e225beba09cb4e9d4a56a6/src/common/src/common-bson-dsl-private.h#L721-L724

src/libmongoc/tests/TestSuite.c
```diff
@@ -276,61 +278,53 @@ TestSuite_AddLive(TestSuite *suite, /* IN */
...
    test->func = func;
-   test->num_checks = 0;
 
+   mstr_view tag, tail;
+   tail = tags;
+   while (tail.len) {
+      mlib_check(mstr_split_around(tail, mstr_cstring("["), NULL, &tag),
+                 because,
+                 "Expected an opening bracket for the next test tag");
+      mlib_check(mstr_split_around(tag, mstr_cstring("]"), &tag, &tail), because, "Expected a closing bracket for tag");
```
eramongodb: Minor issue: this message is emitted even for cases such as `"/bson/error/basic invalid]"`:

```
TestSuite.c:315: in [_V_TestSuite_AddFull]: Check condition ⟨mstr_split_around(((mstr_view) {(((tail)).data), (((tail)).len)}), ((mstr_view) {(((mstr_cstring("["))).data), (((mstr_cstring("["))).len)}), (((void*)0)), (&tag))⟩ failed (Expected an opening bracket for the next test tag)
```

Perhaps reword the error message to be less specific.

src/libmongoc/tests/TestSuite.h
eramongodb, line 780: For consistency, consider either:

- adding support for variadic `TestSuite_AddLive*()` macros/functions to handle the live server tag append internally instead of in test name strings, or
- append the tag to all test name strings consistently to permit using `TestSuite_AddLive*()` _without_ forcing a resource-lock ("safe to run in parallel").

src/libmongoc/tests/test-mongoc-client-session.c
```diff
@@ -2727,41 +2727,42 @@ test_session_install(TestSuite *suite)
...
-   add_session_test_wc(suite, "/Session/watch", test_watch, true, test_framework_skip_if_not_replset);
-   add_session_test(suite, "/Session/aggregate", test_aggregate, true);
-   add_session_test(suite, "/Session/create", test_create, false);
-   add_session_test(suite, "/Session/database_names", test_database_names, true);
-   add_session_test(suite, "/Session/find_databases", test_find_databases, true);
-   add_session_test(suite, "/Session/find_collections", test_find_collections, true);
-   add_session_test(suite, "/Session/collection_names", test_collection_names, true);
-   add_session_test(suite, "/Session/bulk", test_bulk, false);
-   add_session_test(suite, "/Session/find_indexes", test_find_indexes, true);
+   add_session_test(suite, "/Session/cmd [lock:live-server]", test_cmd, false);
```
eramongodb: Consider appending `[lock:live-server]` within `add_session_test` and related macros instead.

src/libmongoc/tests/json-test.c
```diff
@@ -1884,9 +1885,11 @@ _install_json_test_suite_with_check(TestSuite *suite, const char *base, const ch
...
    ASSERT(realpath(joined, resolved));
 
-   if (suite->ctest_run) {
-      const char *found = strstr(suite->ctest_run, subdir);
-      if (found != suite->ctest_run && found != suite->ctest_run + 1) {
+   if (suite->ctest_run.data) {
+      // If we're running a specific test, only register if the directory we are scanning
+      // is a prefix of the requested test pathname
+      size_t where = mstr_find(suite->ctest_run, mstr_cstring(subdir));
+      if (where != 0 && where != 1) {
```
eramongodb:
```suggestion
      if (where > 1) {
```
Suggest adding a drive-by comment documenting why the `!= 1` is necessary (I'm guessing it is to support omitting the leading `/` prefix in the test name?).

build/cmake/TestFixtures.cmake
```diff
@@ -47,3 +47,11 @@ mongo_define_subprocess_fixture(
         "${_MONGOC_BUILD_SCRIPT_DIR}/bottle.py" fake_kms_provider_server:kms_provider
             --bind localhost:14987  # Port 14987 chosen arbitrarily
     )
+
+# Run our very simple HTTP server in a fixture process
+mongo_define_subprocess_fixture(
+    mongoc/fixtures/simple-http-server-18000
+    SPAWN_WAIT 1
+    COMMAND
+        $<TARGET_FILE:Python3::Interpreter> -u "${mongo-c-driver_SOURCE_DIR}/.evergreen/scripts/simple_http_server.py"
```
eramongodb, line 56: Curious: should this be `"${Python3_EXECUTABLE}"` instead of `$<TARGET_FILE:Python3::Interpreter>`?

I'm not certain [how CMake guarantees](https://cmake.org/cmake/help/latest/module/FindPython3.html) this target provides the `TARGET_FILE` property as the path to the interpreter (although it clearly _seems_ to do so given that this currently works), unlike the `${Python3_EXECUTABLE}` variable which is explicitly documented as being for that purpose:

> `Python3::Interpreter`: Python 3 interpreter. This target is defined only if the Interpreter component is found.

vs.

> `Python3_EXECUTABLE`: Path to the Python 3 interpreter.

eramongodb (changes requested):
Initial feedback. TIL `RESOURCE_LOCK`, this is fantastic. I am also very excited about the `mlib_str` and `mlib_vec` APIs.

---

Some benchmarks, averaged across 10 runs ('~' is standard deviation), testing only `/bson/*` (438 tests, all safe to execute in parallel) with `--no-fork` (`/bson/value/big_copy` is skipped):

- `test-libmongoc` (current): 0.79 secs (~0.05)
- `test-libmongoc` (PR): 0.82 secs (~0.04)
- `ctest`: 4.56 secs (~0.75)
- `ctest -j 0`: 1.57 secs (~0.12)

It is surprising to me that `ctest` _with parallelism_ is almost twice as slow as the single-threaded `test-libmongoc`. We may need to explore improving the test registration and filtering setup (in a followup PR is fine) to avoid this performance overhead as much as possible (the eager test registration skips implemented by this PR in `TestSuite_Add*()` do not seem to be sufficient).

---

Just a note that the current output of `ctest --print-labels` is:

```
All Labels:
  ipv6
  json
  lock:fake-kms
  lock:live-server
  slow
  test-libmongoc-generated
  timeout:10
  timeout:20
  timeout:30
  uses:fake_kms_provider_server
  uses:simple-http-server-18000
```

---

Some questions:

- What motivated the addition of [timeout:N] to certain tests and not others? Is the default timeout of 10 seconds too aggressive (e.g. perhaps 60 secs may be preferable)?
- What is the purpose of the `[json]` and `[slow]` tags, both of which are only used by JSON tests?

build/cmake/TestFixtures.cmake
vector-of-bool: Either form is equivalent. I just prefer making a habbit of referring to the import targets in a genex because it will catch typos immediately rather than expanding to an empty string. The `TARGET_FILE` expands to the `IMPORTED_LOCATION` value for imported targets, which should be defined or the imported target itself would not be usable for anything.

src/common/src/mlib/vec.th
vector-of-bool: While possible, defining the vec type in the same terms of many function pointers would make it significantly more complex to modify and would have performance and type-checking implications. 

Alternatively, defining everything in a gigantic macro definition would also be possible, but would be must more difficult to edit and lose editor doc comment assistance. In both cases, doing conditional inclusion of APIs (i.e. conditional copy, conditional non-trivial destroy) would be much more difficult.

eramongodb: 😔 If that is the case, I'd prefer at least that we use a non-`.h` extension so that tools do not mistaken this as a standalone-includable header (thus forcing exclusion filters to avoid errors or noisy diagnostic, e.g. as currently required by `bsoncxx/enums/*` headers in the C++ Driver [here](https://github.com/mongodb/mongo-cxx-driver/blob/8ac2a617894de9dd96d2ce6830aaaee4ae5018d5/.evergreen/scripts/abi-compliance-check-test.sh#L41), [here](https://github.com/mongodb/mongo-cxx-driver/blob/8ac2a617894de9dd96d2ce6830aaaee4ae5018d5/etc/run-clang-tidy.sh#L69), and [here](https://github.com/mongodb/mongo-cxx-driver/blob/8ac2a617894de9dd96d2ce6830aaaee4ae5018d5/src/bsoncxx/test/CMakeLists.txt#L176)).

build/cmake/TestFixtures.cmake
eramongodb: I'm in favor of using targets instead of variables when able. 👍

src/libmongoc/src/mongoc/mongoc-openssl-private.h
vector-of-bool: Going for the latter suggestions

src/libmongoc/tests/TestSuite.c
vector-of-bool: Clarified wording in assertion message to explain the cause being the whitespace introducing a tag list.

src/libmongoc/tests/json-test.c
vector-of-bool: Yes, that was the intention.

src/libmongoc/tests/TestSuite.h
vector-of-bool: I initially wanted to modify the `TestSuite_Add...` functions to support specifying tags in another argument, but it turned out to be a supermassive change for little benefit over just storing them in the `name` parameter with a whitespace delimiter.

I do want to modify our many many `TestSuite_Add...` macros/function in a future change to simplify them, because they've been a large pain in doing this PR.

src/common/src/mlib/vec.th
vector-of-bool: I've used alternative suffixes for such "special-include" files before, such as `.inl` or `.inc`. Is there any preference? (Not `.i`, as that usually means "file of preprocessor output")

src/libmongoc/tests/TestSuite.c
vector-of-bool: Good catch. Not modifying that was a mistake and only works by UB coincidence.

src/common/src/mlib/vec.th
eramongodb: I think `.inl` ("inline") is typically associated with out-of-line template definitions and/or inline functions for C++, so I slightly prefer the `.inc` extension, but it seems to be very overloaded in usage across languages and tools. I initially thought just swapping `.t.h` to `.h.t` would work, but `.t` is apparently used for Perl test scripts.

Perhaps `.th` may work. It's what we already have now ("template header"), but drops the `.` in `t.h`. I cannot find any use of this file extension which is relevant enough to be potentially conflated with our usage.

src/libmongoc/tests/TestSuite.h
eramongodb: Then I am fine with postponing this consistenty issue to a later PR that more directly addresses our test registration woes.

src/libmongoc/src/mongoc/mongoc-openssl-private.h
```diff
@@ -21,8 +21,6 @@
 
 #include <mongoc/mongoc-stream-tls-openssl-private.h>
 
-#include <mongoc/mongoc-ssl.h>
-
```
eramongodb, line 25: Is this removal necessary/intentional given the now-guarded include in test-mongoc-topology-scanner.c?

src/common/src/mlib/vec.th
eramongodb: Suggest renaming `vec.t.h` to `vec.th` (first choice) or `vec.inc` (second choice).

src/common/src/mlib/str.h
```diff
@@ -441,6 +461,63 @@ mstr_find_first_of(mstr_view hay, mstr_view const needles, mlib_upsized_integer
...
+ *
+ * @param s The string to be inspected
+ * @return mstr_view A substring view of `s` that excludes any leading whitespace
+ */
+static inline mstr_view
+mstr_trim_left(mstr_view s)
+{
+   while (s.len) {
+      char c = mstr_at(s, 0);
+      if (c == ' ' || c == '\n' || c == '\r' || c == '\t') {
```
eramongodb:
```suggestion
      if (isspace(c)) {
```
Consider using `isspace` from `<ctype.h>` instead? Also applies to `mstr_trim_right`.

src/common/src/mlib/str.h
```diff
@@ -565,4 +642,488 @@ mstr_contains_any_of(mstr_view str, mstr_view needle)
...
+   // Try to (re)allocate the region
+   char *data = (char *)realloc(str->data, alloc_size);
+   if (!data) {
+      // Failed to (re)allocate
+      return false;
+   }
+   // Note: We do not initialize any of the data in the newly allocated region.
+   // We only set the null terminator. It is up to the caller to do the rest of
+   // the init.
+   data[new_len] = 0;
```
eramongodb:
```suggestion
   data[new_len] = '\0';
```
`char` literal for `char` assignment.

src/common/src/mlib/str.h
```diff
@@ -565,4 +642,488 @@ mstr_contains_any_of(mstr_view str, mstr_view needle)
...
+/**
+ * @brief Obtain a null mstr string object.
+ *
+ * @return mstr A null string, with a null data pointer and zero size
+ */
+static inline mstr
+mstr_null(void)
+{
+   const mstr ret = {0};
+   return ret;
```
eramongodb:
```suggestion
   return (mstr){0};
```
Simplify?

src/common/src/mlib/str.h
```diff
@@ -31,8 +31,10 @@
 #include <mlib/loop.h>
 #include <mlib/test.h>
 
+#include <stdarg.h>
 #include <stdbool.h>
 #include <stddef.h>
+#include <stdint.h>
```
eramongodb:
```suggestion
#include <stdarg.h>
#include <stdbool.h>
#include <stddef.h>
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
```
IWYU for `vsnprintf`, `realloc`/`free`, and `memcpy`.

src/common/src/mlib/str.h
```diff
@@ -565,4 +642,488 @@ mstr_contains_any_of(mstr_view str, mstr_view needle)
...
+ *
+ * @param str The string object to be updated
+ * @param c The single character that will be inserted at the end
+ * @return true If the operation succeeded
+ * @return false Otherwise
+ *
+ * In case of failure, the string is not modified.
+ */
+static inline bool
+mstr_pushchar(mstr *str, char c)
```
eramongodb:
```suggestion
mstr_append_char(mstr *str, char c)
```
Consistent `mlib_append*` naming? ("push" for `mlib_vec`.)

src/common/src/mlib/str.h
```diff
@@ -565,4 +642,488 @@ mstr_contains_any_of(mstr_view str, mstr_view needle)
...
+ * @return false Otherwise
+ *
+ * @warning If the `needle` string is empty, then the program will terminate!
+ * @note If the operation fails, the content of `str` is an unspecified but valid
+ * string.
+ */
+static inline bool
+mstr_replace(mstr *str, mstr_view needle, mstr_view sub)
+{
+   mlib_check(needle.len, neq, 0, because, "Trying to replace an empty string will result in an infinite loop");
```
eramongodb: Perhaps no-op may be preferable to termination here, so users need not special-case an empty `needle`?

src/common/src/mlib/str.h
```diff
@@ -565,4 +642,488 @@ mstr_contains_any_of(mstr_view str, mstr_view needle)
...
+         // Integer overflow while advancing the offset. No good.
+         return false;
+      }
+   }
+}
+
+/**
+ * @brief Like `mstr_sprintf_append`, but accepts the va_list directly.
+ */
+MLIB_IF_GNU_LIKE(__attribute__((format(printf, 2, 0))))
```
eramongodb: This (and other occurrances of the) printf attribute in this file may need to special-case for GCC mingw compatibility by using `gnu_printf`, e.g. see "third issue" PR description of https://github.com/mongodb/libmongocrypt/pull/1054#event-19180404991 and (note: the version check is likely unnecessary at this point given other occurrances of this pattern): https://github.com/mongodb/mongo-c-driver/blob/d9c26f49e75d3de746a690db9c81ff5b4f6e21b0/src/libbson/src/bson/macros.h#L306-L312

src/common/src/mlib/str.h
```diff
@@ -565,4 +642,488 @@ mstr_contains_any_of(mstr_view str, mstr_view needle)
...
+   const size_t start_size = string->len;
+
+   // Try to guess the length of the added string
+   size_t added_len = strlen(format);
+   // Give use some wiggle room since we are inserting characters:
+   if (mlib_mul(&added_len, 2)) {
+      // Doubling the size overflowed. Not likely, but just use the original string
+      // size and grow as-needed.
+      added_len = strlen(format);
+   }
```
eramongodb:
```suggestion
   const size_t format_len = strlen(format);
   size_t added_len = format_len;
   // Give use some wiggle room since we are inserting characters:
   if (mlib_mul(&added_len, 2)) {
      // Doubling the size overflowed. Not likely, but just use the original string
      // size and grow as-needed.
      added_len = format_len;
   }
```
Avoid computing `strlen(format)` twice.

src/common/src/mlib/vec.t.h
```diff
@@ -0,0 +1,450 @@
...
+#undef fn
+#define fn(M) MLIB_PASTE_3(VecName, _, M)
+
+typedef struct VecName {
+   /**
+    * @private
+    * @brief Pointer to the first vector element, or NULL if the vector is
+    * empty.
+    *
+    * @note DO NOT MODIFY
```
eramongodb:
```suggestion
    * @note DO NOT DIRECTLY MODIFY THIS VALUE
```
Suggest being a bit more specific about what these notes mean.

src/common/src/mlib/vec.t.h
```diff
@@ -0,0 +1,450 @@
...
+ * @brief Get the maximum number of elements that can be held in the vector of
+ * a certain type.
+ */
+vec_inline_spec size_t
+fn(max_size(void)) mlib_noexcept
+{
+   // We compare against (signed) SSIZE_MAX because want to support the difference
+   // between two pointers. If we use the unsigned size, then we could have vectors
+   // with size that is too large to represent the difference between two sizes.
+   return SSIZE_MAX / sizeof(T);
```
eramongodb: `SSIZE_MAX` may require platform compatibility workarounds (e.g. see [here](https://github.com/mongodb/mongo-c-driver/blob/d9c26f49e75d3de746a690db9c81ff5b4f6e21b0/src/libbson/src/bson/compat.h#L145-L151)). Consider using `(SIZE_MAX / 2u - 1u)` instead.

src/common/src/mlib/vec.t.h
```diff
@@ -0,0 +1,450 @@
...
+ * @return T* A pointer to the newly added element, or NULL in case of allocation failure.
+ */
+// mlib_nodiscard ("Check the returned pointer for failure")
+vec_inline_spec T *
+fn(push(VecName *self)) mlib_noexcept
+{
+   size_t count = self->size;
+   if (mlib_add(&count, 1)) {
+      // Adding another element would overflow size_t. This is extremely unlikely,
+      // but precautionary.
```
eramongodb: Consider importing `BSON_(UN)LIKELY` into mlib?

src/common/src/mlib/str.h
vector-of-bool: `isspace` (and many `ctype.h` functions) tends to be a foot-gun, as it respects the global locale, and passing an invalid value is UB.

src/common/src/mlib/str.h
```diff
@@ -565,4 +642,488 @@ mstr_contains_any_of(mstr_view str, mstr_view needle)
...
+    * @brief The number of characters in the array pointed-to by `data`
+    * that precede the null terminator.
+    */
+   size_t len;
+} mstr;
+
+
+/**
+ * @brief Resize an existing or null `mstr`, without initializing any of the
+ * added content other than the null terminator. This operation is potientially
```
kevinAlbs:
```suggestion
 * added content other than the null terminator. This operation is potentially
```

src/common/src/mlib/str.h
```diff
@@ -565,4 +642,488 @@ mstr_contains_any_of(mstr_view str, mstr_view needle)
...
+mstr_resize_for_overwrite(mstr *const str, const size_t new_len)
+{
+   // We need to allocate one additional char to hold the null terminator
+   size_t alloc_size = new_len;
+   if (mlib_add(&alloc_size, 1) || alloc_size > PTRDIFF_MAX) {
+      // Allocation size is too large
+      return false;
+   }
+   // Try to (re)allocate the region
+   char *data = (char *)realloc(str->data, alloc_size);
```
kevinAlbs, line 731: Calls to `realloc` and `free` ignore functions set by `bson_mem_set_vtable`. `bson_mem_set_vtable` is used by the PHP driver.

I expect `mlib` is designed not to depend on libbson. Consider adding a function to set custom allocators from `bson_mem_set_vtable`.

src/common/src/mlib/str.h
```diff
@@ -565,4 +642,488 @@ mstr_contains_any_of(mstr_view str, mstr_view needle)
...
+{
+   const mstr ret = {0};
+   return ret;
+}
+
+/**
+ * @brief Replace the content of the given string, attempting to reuse the buffer
+ *
+ * @param inout Pointer to a valid or null `mstr` to be replaced
+ * @param s The new string contents
```
kevinAlbs, line 846:
```suggestion
 * @param s The new string contents. Must not be a view referring to data in `inout`.
```

Suggest documenting self-assignment is not supported. I expect the `mstr_resize_for_overwrite` invalidates views to `inout`.

src/common/src/mlib/str.h
```diff
@@ -565,4 +642,488 @@ mstr_contains_any_of(mstr_view str, mstr_view needle)
...
+static inline bool
+mstr_splice(mstr *str, size_t splice_pos, size_t n_delete, mstr_view insert)
+{
+   mlib_check(splice_pos <= str->len);
+   // How many chars is it possible to delete from `splice_pos`?
+   size_t n_chars_avail_to_delete = str->len - splice_pos;
+   mlib_check(n_delete <= n_chars_avail_to_delete);
+   // Compute the new string length
+   size_t new_len = str->len;
+   // This should never fail, because we should try to delete more chars than we have
```
kevinAlbs:
```suggestion
   // This should never fail, because we should never try to delete more chars than we have
```

src/common/src/mlib/str.h
```diff
@@ -565,4 +642,488 @@ mstr_contains_any_of(mstr_view str, mstr_view needle)
...
+}
+
+#define mstr_concat(A, B) mstr_concat(mstr_view_from((A)), mstr_view_from((B)))
+
+/**
+ * @brief Delete and/or insert characters into a string
+ *
+ * @param str The string object to be updated
+ * @param splice_pos The position at which to do the splice
+ * @param n_delete The number of characters to delete at `splice_pos`
```
kevinAlbs, line 932:
```suggestion
 * @param n_delete The number of characters to delete at `splice_pos`. Must not delete beyond string length.
```

Suggest documenting to alert callers of possible assert. Alternatively consider clamping the delete count:

```c
if (n_delete > n_chars_avail_to_delete) {
   // Clamp the number of chars to delete to the number that are available
   n_delete = n_chars_avail_to_delete;
}
```

That appears consistent with Javascript's `splice`, but I do not see an immediate use-case.

src/common/src/mlib/str.h
```diff
@@ -565,4 +642,488 @@ mstr_contains_any_of(mstr_view str, mstr_view needle)
...
+ * @param insert A string to be inserted at `split_pos` after chars are deleted
```
kevinAlbs, line 933:
```suggestion
 * @param insert A string to be inserted at `split_pos` after chars are deleted. Must not be a view referring to data in `str`
```

src/common/src/mlib/str.h
```diff
@@ -565,4 +642,488 @@ mstr_contains_any_of(mstr_view str, mstr_view needle)
...
+   // Insert the new data
+   memcpy(mut, insert.data, insert.len);
+   return true;
+}
+
+/**
+ * @brief Append a string to the end of some other string.
+ *
+ * @param str The string to be modified
+ * @param suffix The suffix string to be appended onto `*str`
```
kevinAlbs, line 1010:
```suggestion
 * @param suffix The suffix string to be appended onto `*str`. Must not be a view referring to data in `str`
```

Suggest documenting, since this may surprise callers:
```c
mstr s = mstr_copy_cstring("foo");
mstr_append(&s, s); // use-after-free
```

src/common/src/mlib/str.h
```diff
@@ -565,4 +642,488 @@ mstr_contains_any_of(mstr_view str, mstr_view needle)
...
+   mstr_view one = mstr_view_data(&c, 1);
+   return mstr_append(str, one);
+}
+
+/**
+ * @brief Replace every occurrence of `needle` in `str` with `sub`
+ *
+ * @param str The string object to be updated
+ * @param needle The non-empty needle string to be searched for.s
+ * @param sub The string to be inserted in place of each `needle`
```
kevinAlbs, line 1046:
```suggestion
 * @param sub The string to be inserted in place of each `needle`. Must not be a view referring to data in `str`
```

src/common/src/mlib/str.h
vector-of-bool: Consulting Python's `str.replace` and JS `replaceAll`, I've updated this function to do the same thing as those: Insert occurrences of `sub` around/between all chars on empty needles.

src/libmongoc/src/mongoc/mongoc-openssl-private.h
vector-of-bool: Accidentaly

src/common/src/mlib/vec.t.h
vector-of-bool: The correct (and more aptly named) constant macro to use would have been `PTRDIFF_MAX`. Updated.

src/common/src/mlib/str.h
```diff
@@ -441,6 +464,70 @@ mstr_find_first_of(mstr_view hay, mstr_view const needles, mlib_upsized_integer
...
+ * @param c The codepoint to be tested
+ */
+static inline bool
+mlib_is_latin_whitespace(int32_t c)
+{
+   return c == 0x20   // space
+          || c == 0xa // line feed
+          || c == 0xd // carriage return
+          || c == 0x9 // horizontal tab
+      ;
```
eramongodb:
```suggestion
   switch (c) {
   case 0x09: // horizontal tab
   case 0x0a: // line feed
   case 0x0d: // carriage return
   case 0x20: // space
      return true;

   default:
      return false;
   }
```
Switch-case instead of serial conditionals.

src/common/src/mlib/str.h
```diff
@@ -441,6 +464,70 @@ mstr_find_first_of(mstr_view hay, mstr_view const needles, mlib_upsized_integer
 #define _mstr_find_first_of_argc_3(Hay, Needle, Pos) _mstr_find_first_of_argc_4(Hay, Needle, Pos, SIZE_MAX)
 #define _mstr_find_first_of_argc_4(Hay, Needle, Pos, Len) mstr_find_first_of(Hay, Needle, mlib_upsize_integer(Pos), Len)
 
+/**
+ * @brief Test whether the given codepoint is a Basic Latin whitespace character
```
eramongodb, line 468:
```suggestion
 * @brief Test whether the given codepoint is a Basic Latin whitespace character without locale-dependent behavior
```
Include deliberate avoidance of locale-dependent behavior (i.e. `isspace`).

src/common/src/mlib/str.h
```diff
@@ -1010,7 +1019,16 @@ mstr_replace(mstr *str, mstr_view needle, mstr_view sub)
          // Integer overflow while advancing the offset. No good.
          return false;
       }
+      // Note: To support empty needles, advance one more space to avoid infinite
+      // repititions in-place.
+      // TODO: To do this properly, this should instead advance over a full UTF-8-encoded
+      // codepoint. For now, just do a single byte.
```
eramongodb: I think UTF-8 encoding is beyond the scope of these string utilities. Suggest simply documenting that this function's empty-needle behavior is "around and between every _byte_ in the string" rather than "character". Including a small representative example may also help, e.g.:
```
abc -> |a|b|c| (needle: "", sub: "|")
```

src/common/src/mlib/vec.th
```diff
@@ -0,0 +1,466 @@
...
+ *   are unspecified and reserved for future use. Elements are zero-initialized
+ *   before being passed to this macro.
+ * - Optional: Define `VecCopyElement(DstPtr, SrcPtr)` to copy data from `*SrcPtr`
+ *   to `*DstPtr`. The vector's copying function is only defined if this macro
+ *   is defined. This macro MUST evaluate to a boolean to indicate if the copy
+ *   operation succeeded. If a copy fails, then the partially copied elements
+ *   will be destroyed and the overall copy will fail.
+ *
+ *   To add a trival copying function, define `VecCopyElement` to
+ *   `VecTrivialCopyElement`.
```
kevinAlbs, line 23: `VecTrivialCopyElement` appears unused in this file (aside from Intellisense supporting macros). Is this comment out-dated?

build/cmake/LoadTests.cmake
```diff
@@ -3,57 +3,88 @@
...
+    # these as the LABELS for the test case
+    unset(labels)
+    set(labels "${MONGOC_TEST_${casename}_TAGS}")
+
+    # Find what test fixtures the test wants by inspecting labels. Each "uses:"
+    # label defines the names of the test fixtures that a particular case requires
+    list_select(labels SELECT "^uses:(.*)$" REPLACE "mongoc/fixtures/\\1" OUT fixtures)
+
+    # For any "lock:..." labels, add a resource lock with the corresponding name
+    list_select(labels SELECT "^lock:(.*)$" REPLACE "\\1" OUT lock)
```
kevinAlbs:
```suggestion
    list_select(labels SELECT "^lock:(.*)$" REPLACE "\\1" OUT locks)
```

For consistency with `labels` and `fixtures` since there may be more than one lock.

src/common/src/mlib/vec.th
vector-of-bool: The purpose of `VecTrivialCopyElement` to be used by includers, like this:

```c++
#define T int
#define VecCopyElement VecTrivialCopyElement 
#include "vec.th"
```

src/common/src/mlib/str.h
vector-of-bool: I've thought about how to approach that, and I see two alternatives. I created CDRIVER-6141 to track it separately, as adding that would be a significant change.

src/common/src/mlib/str.h
vector-of-bool: I foresee this being a footgun going forward, especially if one doesn't know the provenance of the strings in question. I've modified the code to properly handle and test self-assignment/self-insertion/self-formatting.

src/common/src/mlib/str.h
vector-of-bool: Chosen to clamp the deletion

src/common/src/mlib/str.h
```diff
@@ -565,4 +659,568 @@ mstr_contains_any_of(mstr_view str, mstr_view needle)
...
+   // the target. If not, then these are a no-op.
+   mstr_destroy(&needle_dup);
+   mstr_destroy(&sub_dup);
+   return okay;
+}
+
+/**
+ * @brief Like `mstr_snprintf`, but accepts a `va_list` directly.
+ */
+mlib_printf_attribute(1, 0) static inline mstr mstr_vsnprintf(const char *format, va_list args)
```
kevinAlbs:
```suggestion
 * @brief Like `mstr_sprintf`, but accepts a `va_list` directly.
 */
mlib_printf_attribute(1, 0) static inline mstr mstr_vsprintf(const char *format, va_list args)
```

Fix reference to `mstr_sprintf`. Suggest removing `n` from name since there is no size argument.

src/common/src/mlib/vec.th
kevinAlbs: Ah right. Thank you for the example.

kevinAlbs (approved):
LGTM!

PR #2142, opened 2025-10-07
CDRIVER-6103 AIX build failure fix

shubhamhii:
This PR fixes the signbit symbol missing issue.

Below is the issue URL.

https://jira.mongodb.org/browse/CDRIVER-6103

### DIFF
--- a/src/libbson/tests/test-json.c
+++ b/src/libbson/tests/test-json.c
@@ -1933,9 +1933,9 @@ test_bson_json_double(void)
    BSON_ASSERT(BSON_ITER_HOLDS_DOUBLE(&iter));
    ASSERT_CMPDOUBLE(bson_iter_double(&iter), ==, 0.0);
 
-/* check that "x" is -0.0. signbit not available on FreeBSD or VS 2010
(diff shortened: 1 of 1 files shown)


rcsanchez97 (commented):
Can you confirm that the Boolean logic is correct? It looks to me like it should rather be `&& !defined(_AIX)`, otherwise the logic breaks. For example the way you have implemented the check, if `__FreeBSD__` were defined  and `_AIX` were not defined(i.e., because we are compiling on FreeBSD) then the condition would be `false && (true || false) || true`. This evaluates to an overall value of `true`, and results in the call to `signbit` on a platform that doesn't have it (FreeBSD in this case).

shubhamhii:
> Can you confirm that the Boolean logic is correct? It looks to me like it should rather be `&& !defined(_AIX)`, otherwise the logic breaks. For example the way you have implemented the check, if `__FreeBSD__` were defined and `_AIX` were not defined(i.e., because we are compiling on FreeBSD) then the condition would be `false && (true || false) || true`. This evaluates to an overall value of `true`, and results in the call to `signbit` on a platform that doesn't have it (FreeBSD in this case).

Thanks for pointing it out.
Corrected the logic.

PR #2269, opened 2026-04-13
CDRIVER-6274 finalize client backpressure phase 1 rollout

connorsmacd:
CDRIVER-6274
DRIVERS-3427

# Summary

This PR implement phase 1 Client Backpressure spec changes per DRIVERS-3427:

- Remove the `adaptiveRetries` URI option and token bucket. While the [token bucket may be resurrected in the future](https://docs.google.com/document/d/1z9nU3USk7Vs4U0QN1jCtH8cIMvLjxN9aD834dLcKSlY/edit?tab=t.0#heading=h.soh5dgne92fh), this is not a guarantee, so I removed it completely.
- Add `maxAdaptiveRetries` URI option (non-negative integer, default `2`) to configure the maximum number of retries for operations that fail with a `SystemOverloadedError`.                                                                                       
- Add `enableOverloadRetargeting` URI option (boolean, default `false`) to control whether the driver deprioritizes a server that returns a `SystemOverloadedError` when selecting a server for retry.
- Update prose and unified tests.


NEWS
```diff
@@ -3,12 +3,13 @@ libmongoc 2.3.0 [UNRELEASED]
 
 ## New Features
 
-- Support retry for server overload errors.
-    - Supported on all commands.
-    - Custom application retry logic may need to be adjusted to avoid retrying too long.
-    - Upgrade is recommended to avoid impacts of server changes related to overload errors.
-        - If not upgrading, custom application retry logic may need to be adjusted to handle higher rates of overload errors.
-    - Set the URI option `adaptiveRetry=true` limit retries with a token bucket system.
+* Added support for MongoDB's Intelligent Workload Management (IWM) (<TODO: insert link to IWM landing page once it is live>) and ingress connection rate limiting features. The driver now gracefully handles write-blocking scenarios and optimizes connection establishment during high-load conditions to maintain application availability.
```
connorsmacd, line 6: Note that this is copied verbatim from DRIVERS-3427:

> Drivers should include the following release note: "Added support for MongoDB 's Intelligent Workload Management (IWM) and ingress connection rate limiting features. The driver now gracefully handles write-blocking scenarios and optimizes connection establishment during high-load conditions to maintain application availability," with a link to IWM page in Atlas, which which will be available with [DOCSP-55426](https://jira.mongodb.org/browse/DOCSP-55426) & related Driver Docs, which will be available with [DOCSP-57078](https://jira.mongodb.org/browse/DOCSP-57078).

The `<TODO: insert link to IWM landing page once it is live>` is a placeholder since the IWM landing page is not yet live. I suppose I should create a CDRIVER ticket pinned to 2.3.0 to make sure we update this before the release?

kevinAlbs: > I suppose I should create a CDRIVER ticket pinned to 2.3.0 to make sure we update this before the release?

Yes I think that is a good idea to make sure that is not forgotten.

connorsmacd: Created CDRIVER-6289.

src/libmongoc/doc/mongoc_uri_t.rst
```diff
@@ -97,7 +97,8 @@ Constant                                   Key                               Def
...
 MONGOC_URI_RETRYWRITES                     retrywrites                       true if driver built w/ TLS       If "true" and the server is a MongoDB 3.6+ replica set or sharded cluster, the driver safely retries a write that failed due to a network error or replica set failover. Only inserts, updates of single documents, or deletes of single
                                                                                                                documents are retried.
-MONGOC_URI_ADAPTIVERETRIES                 adaptiveretries                   false                             If "true", the driver will limit retry attempts during periods of prolonged server overload.
+MONGOC_URI_MAXADAPTIVERETRIES              maxadaptiveretries                2                                 The maximum number of times the driver retries an operation that fails with a `SystemOverloadedError`. Must be a non-negative integer.
+MONGOC_URI_ENABLEOVERLOADRETARGETING       enableoverloadretargeting         false                             If "true", the driver deprioritizes a server that returns a `SystemOverloadedError` when selecting a server for retry.
```
eramongodb:
```suggestion
MONGOC_URI_ENABLEOVERLOADRETARGETING       enableoverloadretargeting         false                             If "true", the driver deprioritizes servers which returned a `SystemOverloadedError` when selecting a server to retry an operation.
```
Minor wording tweak suggestion.

NEWS
```diff
@@ -3,12 +3,13 @@ libmongoc 2.3.0 [UNRELEASED]
...
+    * Supported on all commands.
+    * Custom application retry logic may need to be adjusted to avoid retrying too long.
+    * Upgrade is recommended to avoid impacts of server changes related to overload errors.
+        * If not upgrading, custom application retry logic may need to be adjusted to handle higher rates of overload errors.
+    * Add URI option `maxAdaptiveRetries` to configure the maximum number of retries for operations that fail with a `SystemOverloadedError` (default: `2`).
+    * Add URI option `enableOverloadRetargeting` to control whether retries of `SystemOverloadedError` prefer a different server (default: `false`).
```
eramongodb:
```suggestion
    * Add URI option `enableOverloadRetargeting` to control whether retries of `SystemOverloadedError` will attempt to use a different server (default: `false`).
```
Minor wording tweak suggestion.

src/libmongoc/src/mongoc/mongoc-client.c
```diff
@@ -1127,6 +1127,10 @@ _mongoc_client_new_from_topology(mongoc_topology_t *topology)
    client->client_sessions = mongoc_set_new(8, NULL, NULL);
    client->csid_rand_seed = (unsigned int)bson_get_monotonic_time();
    client->jitter_source = _mongoc_jitter_source_new(_mongoc_jitter_source_generate_default);
+   client->max_adaptive_retries =
+      mongoc_uri_get_option_as_int32(client->uri, MONGOC_URI_MAXADAPTIVERETRIES, MONGOC_DEFAULT_MAXADAPTIVERETRIES);
```
kevinAlbs: `mongoc_uri_get_option_as_int32` has the unfortunate quirk of treating 0 as unset, resulting in the default value of 2 being used.

Suggest adding an internal helper (like `_mongoc_uri_get_max_staleness_option`) so 0 is treated as 0. And consider adding a regression test:

```c
static void
test_backpressure_maxAdaptiveRetries0(void *ctx)
{
   BSON_UNUSED(ctx);

   // Create client with maxAdaptiveRetries=0
   {
      mongoc_uri_t *const uri = test_framework_get_uri();
      mongoc_uri_set_option_as_int32(uri, MONGOC_URI_MAXADAPTIVERETRIES, 0);

      client = test_framework_client_new_from_uri(uri, NULL);
      test_framework_set_ssl_opts(client);

      mongoc_uri_destroy(uri);
   }

   // Reuse prose test context to track retries:
   prose_test_3_and_4_ctx_t apm_ctx = {0};
   {
      mongoc_apm_callbacks_t *const callbacks = mongoc_apm_callbacks_new();
      mongoc_apm_set_command_started_cb(callbacks, prose_test_3_command_started_cb);
      mongoc_client_set_apm_callbacks(client, callbacks, &apm_ctx);
      mongoc_apm_callbacks_destroy(callbacks);
   }


   // Configure a failPoint to trigger with `SystemOverloadedError` and `RetryableError` labels on `find`.
   run_admin_command(BSON_STR({
      "configureFailPoint" : "failCommand",
      "mode" : "alwaysOn",
      "data" : {
         "failCommands" : ["find"],
         "errorCode" : 462, // IngressRequestRateLimitExceeded
         "errorLabels" : [ "SystemOverloadedError", "RetryableError" ]
      }
   }));

   bson_error_t error;

   // Perform a find operation with `coll` that fails.
   {
      mongoc_collection_t *const coll = mongoc_client_get_collection(client, "db", "coll");
      mongoc_cursor_t *const cursor = mongoc_collection_find_with_opts(coll, tmp_bson("{}"), NULL, NULL);

      const bson_t *doc;
      ASSERT(!mongoc_cursor_next(cursor, &doc));

      const bson_t *error_doc;
      ASSERT(mongoc_cursor_error_document(cursor, &error, &error_doc));

      // Assert that the raised error contains both the `RetryableError` and `SystemOverloadedError` error labels.
      ASSERT(mongoc_error_has_label(error_doc, MONGOC_ERROR_LABEL_RETRYABLEERROR));
      ASSERT(mongoc_error_has_label(error_doc, MONGOC_ERROR_LABEL_SYSTEMOVERLOADEDERROR));

      mongoc_cursor_destroy(cursor);
      mongoc_collection_destroy(coll);
   }

   // Assert no retries.
   ASSERT_CMPINT(apm_ctx.find_commands_started_count, ==, 0);

   disable_fail_point();

   mongoc_client_destroy(client);
}
```

kevinAlbs (commented):
Changes look good with an added check for URI value of 0.

The Evergreen task failures look unrelated to these changes. The "Incorrect shasum256" on Windows 7.0 tasks is reported on SERVER-124033.

src/libmongoc/src/mongoc/mongoc-retryable-cmd.c
```diff
@@ -68,22 +68,10 @@ _mongoc_retryable_cmd_run(const mongoc_retryable_cmd_t *cmd, bson_t *reply, bson
 
       const bool is_retryable = is_retryable_read || is_retryable_write || is_overload_retryable;
 
-      const bool is_retry_attempt = attempt > 0;
```
eramongodb, line 71: This algorithm currently maintains two separate "attempt" counters:

- the local `attempt` variable which is unconditionally incremented on every iteration and compared against `maxAdaptiveRetries` (total retry attempts regardless of reason), and
- the retry backoff generator's `attempt` data member which is only incremented when `is_overload` is true and `_mongoc_retry_backoff_generator_next()` is invoked (retry attempts due to retryable overload errors only).

According to the current spec wording:

> `attempt` is the execution attempt number (starting with 0). Note that `attempt` includes retries for errors that are not overload errors (this might include attempts under other retry policies, see Interactions with Other Retry Policies).

The spec specifies only a single `attempt` counter. I think this suggests the retry backoff generator's `attempt` counter should be removed in favor of passing the local `attempt` counter as an argument to `_mongoc_retry_backoff_generator_next()`.

connorsmacd: Great catch!

Your suggested change seems reasonable. It does however make `mongoc_retry_backoff_generator_t` stateless, so I'm not sure if "generator" makes sense in the name. Alternatively, I could add something like `_mongoc_retry_backoff_generator_skip` to increase the underlying counter without doing any calculation.

eramongodb: Adding an "increment the attempt counter" API to the generator SGTM.

> It does however make [it] stateless ...

If `_mongoc_retry_backoff_generator_next()` can be made a pure function, that also SGTM, but I expect the increment counter API will be a simpler change. IIRC the original motivation for having a generator object was to avoid having to pass `mongoc_retry_backoff_params_t` and `jitter_source` around, but perhaps that concern is somewhat reduced after https://github.com/mongodb/mongo-c-driver/pull/2240 refactored and consolidated all usage of the generator to just `mongoc_client_session_with_transaction()` and `_mongoc_retryable_cmd_run()`. 🤔

connorsmacd: I opted for the "skip" API to minimize code changes, but I'll keep this discussion in mind for any future iterations of backpressure.

src/libmongoc/src/mongoc/mongoc-retry-backoff-generator.c
```diff
@@ -67,12 +67,18 @@ _duration_double_multiply(mlib_duration duration, double factor)
    return mlib_duration((mlib_duration_rep_t)round((double)mlib_microseconds_count(duration) * factor), us);
 }
 
+void
+_increment_attempt(mongoc_retry_backoff_generator_t *generator)
```
eramongodb:
```suggestion
static void
_increment_attempt(mongoc_retry_backoff_generator_t *generator)
```
Internal helper.

eramongodb (approved):

kevinAlbs (approved):
LGTM!

PR #2273, opened 2026-04-15
CDRIVER-6291 verify retry behavior with mixed overload and non-overload errors

connorsmacd:
CDRIVER-6291
DRIVERS-3446

# Summary

Implement the prose tests added by DRIVERS-3446:

- Retryable Reads prose tests 4 and 5
- Retryable Writes prose test 6 cases 4 and 5

Additionally, there is a drive-by fix to Retryable Writes prose test 6 case 3.

## Use of global variables in Retryable Reads prose test 5 and Retryable Writes prose test 6 case 5

Both of these prose tests require some mechanism to detect how many times backoff was applied during the execution of a command. In the Python implementation, they replace the internal `RetryPolicy` type with a mock that counts the number of times the relevant function is called. While we could implement a similar customization point in the C Driver, I was concerned about having enough time for verification and code review before the April 17 release. 

For expediency, I chose to use a global counter for each test that is incremented every time the jitter source generate function is called. Since jitter is generated only when we calculate backoff, this effectively counts the number of times backoff is applied. Globals are obviously not ideal, so I am open to alternative solutions.

### DIFF
--- a/src/libmongoc/tests/test-mongoc-retryable-reads.c
+++ b/src/libmongoc/tests/test-mongoc-retryable-reads.c
@@ -827,6 +827,10 @@ test_retryable_reads_prose_4(void *unused)
    mongoc_client_destroy(client);
 }
 
+// Prose test 5 requires some means of detecting how many times backoff is applied during the execution of a command.
+// Since the C Driver lacks a formal mechanism for this, we use a global counter that is incremented every time
+// `prose_test_5_jitter_source_generate` is called. Since jitter is generated only when backoff is calculated, this
+// effectively counts the number of times backoff is applied.
 static int gProseTest5BackoffCount;
 
 static double
--- a/src/libmongoc/tests/test-mongoc-retryable-reads.c
+++ b/src/libmongoc/tests/test-mongoc-retryable-reads.c
@@ -827,17 +827,15 @@ test_retryable_reads_prose_4(void *unused)
    mongoc_client_destroy(client);
 }
 
-// Prose test 5 requires some means of detecting how many times backoff is applied during the execution of a command.
-// Since the C Driver lacks a formal mechanism for this, we use a global counter that is incremented every time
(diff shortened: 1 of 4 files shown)


src/libmongoc/tests/test-mongoc-retryable-reads.c
```diff
@@ -826,6 +827,102 @@ test_retryable_reads_prose_4(void *unused)
    mongoc_client_destroy(client);
 }
 
+static int gProseTest5BackoffCount;
```
kevinAlbs: > Globals are obviously not ideal, so I am open to alternative solutions.

I think the global is OK in this case. It is only shared by one test. Even if `test-libmongoc` were to support running tests in parallel in one process, I don't think it would cause a runtime issue.

That said, I think this could be avoided by adding a context to the jitter source:

```c
void
_mongoc_jitter_source_set_context(mongoc_jitter_source_t *source, void *context);

void *
_mongoc_jitter_source_get_context(mongoc_jitter_source_t *source);
```

Then using that to pass a local variable:
```c
int jitter_call_count = 0;
mongoc_jitter_source_t *js = _mongoc_jitter_source_new(prose_test_5_jitter_source_generate);
_mongoc_jitter_source_set_context(js, &jitter_call_count);
_mongoc_client_set_jitter_source(client, js);
```

kevinAlbs (approved):
LGTM. I do not think the global is particularly problematic, but suggested an alternative.

src/libmongoc/tests/test-mongoc-retryable-reads.c
connorsmacd: Thank you for the input.

I think this solution will work. I was originally thinking I would have to modify all the existing jitter generate functions, but I overlooked the fact that the `mongoc_jitter_source_t` object is passed as an argument (unused in all existing generate functions).

Since CI looks good and I have your approval, I'll take the time to make these changes, with the option to revert to the globals if I somehow can't get this to work before release time.

PR #2277, opened 2026-04-17
CDRIVER-6238 sync keyAltName spec tests

mdb-ad:


src/libmongoc/CMakeLists.txt
```diff
@@ -489,10 +489,10 @@ elseif (NOT ENABLE_CLIENT_SIDE_ENCRYPTION STREQUAL OFF)
       find_package (mongocrypt QUIET)
    endif ()
 
-   if (mongocrypt_FOUND AND "${mongocrypt_VERSION}" VERSION_LESS 1.15.1)
+   if (mongocrypt_FOUND AND "${mongocrypt_VERSION}" VERSION_LESS 1.18.0)
```
kevinAlbs, line 492: Also add a stub entry in NEWS to announce the new dependency requirement:

```
libmongoc 2.4.0 [UNRELEASED]
============================

## Notes

* Raise required version of libmongocrypt from 1.15.1 to 1.18.0 to support In-Use Encryption (corresponds to the CMake option: `ENABLE_CLIENT_SIDE_ENCRYPTION`).
```

kevinAlbs (approved):
LGTM with an addition to NEWS.

PR #2278, opened 2026-04-22
Sync rpm spec

rcsanchez97:
Evergreen patch build: https://spruce.corp.mongodb.com/version/69e816cd3c1c71000754a0c3/tasks

### DIFF
--- a/.evergreen/scripts/build_snapshot_rpm.sh
+++ b/.evergreen/scripts/build_snapshot_rpm.sh
@@ -128,7 +128,7 @@ sudo mock -r ${config} --use-bootstrap-image --isolation=simple --copyin "${mock
 
 sudo mock -r ${config} --use-bootstrap-image --isolation=simple --cwd "/tmp/${build_dir}" --chroot -- /bin/sh -c '(
   rpm --define "_pkgverify_level digest" -Uvh ../mock-result/*.rpm &&
-  gcc $(pkgconf --cflags bson2 mongoc2) -o example-client src/libmongoc/examples/example-client.c -lmongoc2 -lbson2
+  gcc $(pkgconf --cflags bson2 mongoc2) -o example-client src/libmongoc/examples/example-client.c $(pkgconf --libs bson2 mongoc2)
   )'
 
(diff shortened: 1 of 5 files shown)


.evergreen/scripts/build_snapshot_rpm.sh
```diff
@@ -131,10 +126,10 @@ sudo mock --resultdir="${mock_result}" --use-bootstrap-image --isolation=simple
 )
 sudo mock -r ${config} --use-bootstrap-image --isolation=simple --copyin "${mock_result}" /tmp
 
-sudo mock -r ${config} --use-bootstrap-image --isolation=simple --cwd "/tmp/${build_dir}" --chroot -- /bin/sh -c "(
-  rpm -Uvh ../mock-result/*.rpm &&
-  gcc -I/usr/include/libmongoc-1.0 -I/usr/include/libbson-1.0 -o example-client src/libmongoc/examples/example-client.c -lmongoc-1.0 -lbson-1.0
-  )"
+sudo mock -r ${config} --use-bootstrap-image --isolation=simple --cwd "/tmp/${build_dir}" --chroot -- /bin/sh -c '(
+  rpm --define "_pkgverify_level digest" -Uvh ../mock-result/*.rpm &&
+  gcc $(pkgconf --cflags bson2 mongoc2) -o example-client src/libmongoc/examples/example-client.c -lmongoc2 -lbson2
```
eramongodb: Curious: if using `pkgconf` for cflags, why not also for ldflags?
```bash
# Example when run in local environment.
$ pkgconf --libs bson2 mongoc2
-lmongoc2 -lbson2 -lrt
```

eramongodb (approved):
Minor question; otherwise, LGTM.

.evergreen/scripts/build_snapshot_rpm.sh
rcsanchez97: Good catch. Initially I had both `-I` and `-l` specified manually. I switched the includes over to `pkgconf` as a result of the new sub-directory placement. It appears that I simply overlooked the libraries. I have made this change, confirmed with a patch build, and pushed.

PR #2281, opened 2026-04-22
[CDRIVER-6174] Earthly-based Debian package builds

vector-of-bool:
Refer: CDRIVER-6174

# Summary

This changeset introduces Earthly targets and Evergreen tasks that build Debian packages of mongo-c-driver components. It is based on the process in the existing `debian_package_build.sh`, but cleaned up and incorporating some bits from the `debian/unstable` branch. The following to note:

- Existing package scripts cannot be run locally without root permissions, and are destructive to local state. They also require Debian packaging tools to be installed.
- Existing package scripts do not respect local changes, as any modifications are immediately deleted with `git clean` and `git reset` before the build runs. The Earthly build target instead adds and commits any local changes in a commit that is only used for the Earthly build
  - I spent a lot of effort trying to make it so that the commit/reset was unnecessary, but I couldn't make `git-buildpackage` respect uncommitted changes, so the `git add -A . && git commit` was used instead. Is there a way to make `git-buildpackage` respect the working copy?
- The new packaging scripts are written to mimic the behavior of the existing packaging scripts, regardless of whether the behavior of those scripts is desirable. **If we want to change the packaging behavior, now would be a good time to propose that.**
- I spent a lot of time trying to make the build operations idempotent, but couldn't manage it, as the package build necessarily modifies the source tree state and persists state between executions. As such, I just rely on the container always starting from a clean slate to create locally-reproducible builds. There is probably a way to make this work, but I couldn't justify further effort to look into it.
- I tried to updating the existing build scripts to make use of the new build scripts, but also could not justify the effort to make it work. If the new build tasks are sufficient, then the existing build scripts can be removed.
  - The simple script `debian/build_snapshot.sh` that lived in the `debian/unstable` branch has been inlined into `etc/deb/build.sh`.
- The new package build tasks run significantly faster than the existing task (26min versus <10min). Much of the time saved is in not running the `debootstrap` tool to create a chroot.
  - The tasks are split between an amd64 build and an i386 build that run concurrently. There is no ARM64 build in EVG, but one can be done locally by passing `--platform=linux/arm64`.
- The new EVG tasks do not post any build artifacts, just test the build. If we want those artifacts to be available in EVG, more changes will need to be done.
- As a drive-by change, a lot of unnecessary `.gitignore` files/entries have been deleted.

Packages can be obtained from the Earthly build by extracting the root artifact directory:

```sh
$ earthly -a +deb.packages/ deb-packages
```


docs/dev/earthly.rst
```diff
@@ -183,6 +183,31 @@ Build & Test Tasks
...
+   .. rubric:: Parameters
+   .. option:: --debian_version
+
+      Control the Debian container version used for the build.
+
+
+.. program:: +deb.test
+.. earthly-target:: +deb.test
+
+   Test the packages that are build by the `+deb.packages` target.
```
rcsanchez97:
```suggestion
   Test the packages that are built by the `+deb.packages` target.
```

etc/deb/prep-env.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+# This script prepare the Debian environment for the building of Debian packages.
```
rcsanchez97:
```suggestion
# This script prepares the Debian environment for the building of Debian packages.
```

rcsanchez97 (approved):
LGTM.

.evergreen/legacy_config_generator/evergreen_config_lib/variants.py
```diff
@@ -331,6 +331,7 @@ def days(n: int) -> int:
         [
             'debian-package-build',
```
kevinAlbs: > If the new build tasks are sufficient, then the existing build scripts can be removed.

I think yes. Remove `debian_package_build.sh` and associated Evergreen tasks as I expect they are no-longer-needed. `debian_package_build.sh` is not mentioned in `debian.rst`.

> If we want those artifacts to be available in EVG, more changes will need to be done.

Similarly, I do not expect the existing uploaded artifacts are used. Especially since the Earthly target can be easily locally run, I do not expect uploading the artifacts to be useful.

etc/deb/build.sh
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+## Script parameters
+# Directory with the mongo-c-driver repository
+: "${MCD_DIR:=$(dirname "$(dirname "$(dirname "${BASH_SOURCE[0]}")")")}"
+# The Git remote from which we will pull the Debian package control files
+: "${DEBIAN_REMOTE:="https://github.com/mongodb/mongo-c-driver"}"
+# The branh within the remote that will be used for package control files
```
kevinAlbs:
```suggestion
# The branch within the remote that will be used for package control files
```

kevinAlbs (approved):
Nicely done. The commentary, simpler local build, and speed-up are greatly appreciated. LGTM with a suggested change to remove the no-longer-needed tasks.

rcsanchez97:
The additional `debian-package-build` removal LGTM.

PR #2289, opened 2026-05-05
CDRIVER-6327 implement case and diacritic prose tests

mdb-ad:
Evergreen test: https://spruce.corp.mongodb.com/version/69fe3b156c09fa0007dd4f45

edit: included in https://github.com/mongodb/mongo-c-driver/pull/2310

### DIFF
--- a/.evergreen/scripts/compile-libmongocrypt.sh
+++ b/.evergreen/scripts/compile-libmongocrypt.sh
@@ -14,7 +14,7 @@ compile_libmongocrypt() {
   # Run `.evergreen/scripts/kms-divergence-check.sh` to ensure that there is no divergence in the copied files.
   declare -r version="1.18.0-dev"
 
-  git clone --depth=1 -q https://github.com/mongodb/libmongocrypt --branch 1.18.0 || return
+  git clone --depth=1 -q https://github.com/mongodb/libmongocrypt --branch 1.18.1 || return
 
   declare -a crypt_cmake_flags=(
     "-DMONGOCRYPT_MONGOC_DIR=${mongoc_dir}"
--- a/.evergreen/scripts/compile-libmongocrypt.sh
+++ b/.evergreen/scripts/compile-libmongocrypt.sh
@@ -14,14 +14,7 @@ compile_libmongocrypt() {
   # Run `.evergreen/scripts/kms-divergence-check.sh` to ensure that there is no divergence in the copied files.
   declare -r version="1.18.0-dev"
 
-  {
-    git clone -q https://github.com/mongodb/libmongocrypt || return
-    cd libmongocrypt || return
-    git checkout 6d6bc38254a07bd47dbd0e665cffd67adf8746a9 || return
-    cd .. || return
-  }
-  # TODO: after libmongocrypt 1.18.0 is released, replace the above block with:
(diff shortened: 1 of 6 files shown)


src/libmongoc/tests/test-mongoc-client-side-encryption.c
```diff
@@ -4868,6 +5100,154 @@ test_explicit_encryption_text_substring(void *unused)
...
+            doc(kv("$encStrContains", doc(kv("input", cstr("$encryptedText")), kv("substring", value(findPayload)))))));
+      ASSERT_OR_PRINT(ok, error);
+
+      mongoc_cursor_t *cursor;
+      const bson_t *got;
+
+      cursor = mongoc_collection_find_with_opts(eef->encryptedColl, &expr, NULL /* opts */, NULL /* read_prefs */);
+      ASSERT(mongoc_cursor_next(cursor, &got));
+      ASSERT_OR_PRINT(!mongoc_cursor_error(cursor, &error), error);
+      ASSERT_MATCH(got, "{ 'encryptedText': 'FooBarBaz' }");
```
kevinAlbs: This assert appears to be failing on some [tasks](https://spruce.corp.mongodb.com/task/mongo_c_driver_sanitizers_matrix_asan_asan_cse_sasl_cyrus_openssl_rhel8_latest_clang_test_latest_replica_auth_patch_b8925d926da23ad2c15a6f8b46385227a0b05ecc_69fa36ada3f9f600074c7785_26_05_05_18_28_01/logs?execution=0):

```
encryptedText: expected "FooBarBaz", got "foobarbaz"
```

I expect this is finding a "foobarbaz" inserted in a previous text, and is matching due to the case-insensitive.

Suggest using a different string.

kevinAlbs:
These tests use explicit-encryption both for insert and find. The tests can pass without the fix of MONGOCRYPT-903 since the bools are flipped in both insert and find payloads.

The original bug reproduces when the value is inserted with auto-encryption, and queried with explicit-encryption.

Suggest using auto encryption to insert a document in the test set-up. Then use explicit encryption to query the document in these tests.

Expect the test to fail first. Then update libmongocrypt to include MONGOCRYPT-903. Then expect the test to pass.

.evergreen/scripts/compile-libmongocrypt.sh
```diff
@@ -14,14 +14,7 @@ compile_libmongocrypt() {
...
 
-  {
-    git clone -q https://github.com/mongodb/libmongocrypt || return
-    cd libmongocrypt || return
-    git checkout 6d6bc38254a07bd47dbd0e665cffd67adf8746a9 || return
-    cd .. || return
-  }
-  # TODO: after libmongocrypt 1.18.0 is released, replace the above block with:
-  # git clone --depth=1 -q https://github.com/mongodb/libmongocrypt --branch 1.18.0 || return
+  git clone --depth=1 -q https://github.com/mongodb/libmongocrypt --branch 1.18.0 || return
```
kevinAlbs:
```suggestion
  git clone --depth=1 -q https://github.com/mongodb/libmongocrypt --branch 1.18.1 || return
```

1.18.1 has the fix of MONGOCRYPT-903.

kevinAlbs:
PR is closed since case/diacritic prose tests are included in https://github.com/mongodb/mongo-c-driver/pull/2310.

PR #2293, opened 2026-05-08
Fix variable check error in str.h

wu-zhao-min:
Correct line 1080 in str.h by changing okay = !!needle_dup.data to okay = !!sub_dup.data .
This ensures the result of the sub string duplication is checked correctly, preventing potential null pointer dereferences or the use of uninitialized memory caused by a copy-paste error.


kevinAlbs (approved):
Thank you for the fix.

`mstr_replace` is unused aside from tests. But this seems like it may prevent a future bug.

Curious: @wu-zhao-min how did you encounter this issue? Are you using mlib in your own project?

wu-zhao-min:
I discovered this issue while reading the source code. To prevent any errors from occurring during future use, I have submitted this fix.

PR #2297, opened 2026-05-11
CDRIVER-6319 Improve handling `EAGAIN` errors in OpenSSL BIO read callbacks

eramongodb:
Resolves CDRIVER-6319 and ongoing task failures on `windows-2022-latest` EVG distros. Re-enables the "run auth tests" step that were skipped in https://github.com/mongodb/mongo-c-driver/pull/2290.

Quoting Claude Code's diagnosis (version 2.1.139, Opus 1M with high effort):

> OpenSSL 4.0.0 introduced new auto-EOF detection in its legacy `BIO_read` compatibility shim (`bread_conv` in `crypto/bio/bio_meth.c`):
> 
> ```c
> // OpenSSL 4.0.0 — new logic absent in 3.6.2
> ret = bio->method->bread_old(bio, data, (int)datal);
> bio->flags &= ~BIO_FLAGS_AUTO_EOF;
> if (ret == 0) {
>     if (BIO_ctrl(bio, BIO_CTRL_EOF, 0, NULL) == 0)
>         bio->flags |= BIO_FLAGS_AUTO_EOF;   // <-- new
>     ...
> }
> ```
> 
> And `BIO_eof()` in 4.0.0 short-circuits on this flag, so the record layer sees `OSSL_RECORD_RETURN_EOF` and aborts the handshake with `SSL_R_UNEXPECTED_EOF_WHILE_READING`.
> 
> mongoc's custom BIO read (`mongoc_stream_tls_openssl_bio_read`) hit this because:
> 1. `_mongoc_stream_socket_readv` with `min_bytes=0` returns `0` (not `-1`) on `EAGAIN`.
> 2. mongoc's ctrl handler returns `0` for everything except `BIO_CTRL_FLUSH`, so `BIO_CTRL_EOF` returns `0`.
> 3. mongoc set the retry flag on the wrong BIO (`openssl->bio`, the SSL filter BIO), not on `b` (the network BIO OpenSSL is asking about).
> 
> The trio triggered `AUTO_EOF` in 4.0.0. In 3.6.2, `BIO_eof` was a plain ctrl wrapper, so a `0`-returning ctrl meant "not EOF" and the issue was masked.

Quoting OpenSSL 4.0.0 docs for [BIO_meth_new](https://docs.openssl.org/4.0/man3/BIO_meth_new/):

> Older code may call BIO_meth_get_read() and BIO_meth_set_read() instead to set an old-style read function. The parameters for the function have the same meaning as for BIO_read() and it must return values as described for BIO_read(). \[...\] Functions set by BIO_meth_set_read_ex() and BIO_meth_set_read() must call BIO_set_flags() to set the BIO_FLAGS_SHOULD_RETRY flag in relevant situations.

and [BIO_read](https://docs.openssl.org/4.0/man3/BIO_read/):

> A return value of 0 indicates that end-of-file was reached, or that a read of zero bytes was requested. A negative return value indicates an error condition.

The referenced code was introduced in [this commit](https://github.com/openssl/openssl/commit/be42447469dcccc68b7e115c7947e3c25124f3f2#diff-d34b4f116098efaead6681b47a5f3b96b7ed54fbcbebe57905335f7e0ece9c4dR134-R140), which is part of a series of (ongoing) improvements to EOF-handling by BIO read functions. Our "old-style" read callback function implementation violates the updated wording in the OpenSSL 4.0.0 function contract, which (now) explicitly documents and enforces better EOF-handling behavior, which in turn changes how _non-EOF_ behavior is handled, which thus affects the handling of `EAGAIN` errors.

---

A new regression test `/TLS/eagain` is added by this PR. To support this new test, the `ssl_test` API is extended to support wrapping the client stream. This wrapper is used to create a stream (in a manner similar to `debug_stream`) that deliberately returns `EAGAIN` errors (up to three times) before eventually succeeding.

Without setting `ret = -1`, which is required to indicate to OpenSSL 4.0.0 that the result is an error, the test failures with the originally observed error (paraphrased for brevity):

```
ERROR: mongoc: ERRORED (line: 227): TLS handshake failed: error:0A000126:SSL routines::unexpected eof while reading
ERROR: mongoc: ERRORED (line: 104): TLS handshake failed: error:0A00010B:SSL routines::wrong version number
ERROR: mongoc: msg: error:0A000139:SSL routines::record layer failure
FAIL
Assert Failure: 3 == 1 (SSL_TEST_SSL_HANDSHAKE == SSL_TEST_SUCCESS)
```

Without the `BIO_set_retry_read(b)`, which is required to indicate to OpenSSL 4.0.0 that the read operation will be retried at a later time, the test also fails (paraphrased for brevity):

```
ERROR: mongoc: ERRORED (line: 227): TLS handshake failed: Connection timed out
ERROR: mongoc: ERRORED (line: 104): TLS handshake failed: Connection timed out
ERROR: mongoc: msg: error:0A000139:SSL routines::record layer failure
FAIL
Assert Failure: 3 == 1 (SSL_TEST_SSL_HANDSHAKE == SSL_TEST_SUCCESS)
```

With both additions to `mongoc_stream_tls_openssl_bio_read()`, the test passes.

### DIFF
--- a/.evergreen/config_generator/components/openssl_compat.py
+++ b/.evergreen/config_generator/components/openssl_compat.py
@@ -73,11 +73,8 @@ def tasks():
             commands = [
                 FetchSource.call(),
                 OpenSSLSetup.call(vars=vars),
+                FunctionCall(func='run auth tests'),
             ]
-
-            # CDRIVER-6319: Atlas test servers do not yet support connections with OpenSSL 4.0.0...?
-            if version not in ['4.0.0']:
-                commands.append(FunctionCall(func='run auth tests'))
 
             yield EvgTask(
                 name=f'{TAG}-{version}-{link_type}-{distro_str}',
--- a/.evergreen/generated_configs/tasks.yml
+++ b/.evergreen/generated_configs/tasks.yml
@@ -6262,6 +6262,7 @@ tasks:
       - func: openssl-compat
         vars:
           OPENSSL_VERSION: 4.0.0
+      - func: run auth tests
   - name: openssl-compat-4.0.0-static-ubuntu2404-gcc
     run_on: ubuntu2404-large
     tags: [openssl-compat, openssl-4.0.0, openssl-static, ubuntu2404, gcc]
@@ -6271,6 +6272,7 @@ tasks:
         vars:
           OPENSSL_USE_STATIC_LIBS: "ON"
           OPENSSL_VERSION: 4.0.0
+      - func: run auth tests
   - name: openssl-compat-fips-3.0.9-shared-ubuntu2404-gcc
     run_on: ubuntu2404-large
     tags: [openssl-compat, openssl-fips-3.0.9, openssl-shared, ubuntu2404, gcc]
--- a/NEWS
+++ b/NEWS
@@ -6,6 +6,10 @@ libmongoc 2.4.0 [UNRELEASED]
 * Raise required version of libmongocrypt from 1.15.1 to 1.18.0 to support In-Use Encryption (corresponds to the CMake option: `ENABLE_CLIENT_SIDE_ENCRYPTION`).
 * In a future minor release the minimum supported MongoDB Server version will be raised from 4.2 to 4.4. This is in accordance with [MongoDB Software Lifecycle Schedules](https://www.mongodb.com/legal/support-policy/lifecycles).
 
+## Improvements
+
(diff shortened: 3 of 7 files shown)


kevinAlbs (approved):
LGTM. Thank you very much for diagnosing a fixing.

Proposal: cherry-pick this fix to `r2.3` for the upcoming 2.3.1 release. Though I do not think this is strictly a bug fix in the C driver (in response to backwards incompatible change in OpenSSL 4). CDRIVER-6308 already notes upcoming OpenSSL 4 usage. This might be a blocker for anyone upgrading their client to OpenSSL 4. I would rather release this change sooner in 2.3.1.

PR #2307, opened 2026-05-29
CDRIVER-6339 Replace Google Tag Manager with Mixpanel

eramongodb:
Resolves CDRIVER-6339 by immediately deploying the changes in https://github.com/mongodb/mongo-c-driver/pull/2306 by manually generating HTML pages for bson (2.3.0) and mongoc (2.3.0) and patching the HTML files in the gh-pages branch.

Due to changes to implicit dependencies of the Sphinx generator and Furo themes since the original 2.3.0 pages were generated, applied the changes as follows:

- Checkout the 2.3.0 tag.
- Cherry-pick https://github.com/mongodb/mongo-c-driver/commit/e5038f25d9088a9efe5befaf44e9316c0e5a927c and https://github.com/mongodb/mongo-c-driver/commit/deab4ed0c98e1b31c5148cdd1e111948455ecba3.
- Generate HTML pages and commit the diffs ("Re-generate ...").
- Cherry-pick https://github.com/mongodb/mongo-c-driver/pull/2306.
- Generate HTML pages and commit the diffs ("Replace ...").
- Revert the changes made in the "Re-generate ..." commit.

This sequence of commits isolates the resulting diffs to _just_ the addition of the Mixtape tracker script without unrelated changes caused by updated dependencies.


libbson/2.3.0/api.html
kevinAlbs: The generated files still have the Google tag. That might be harmless, but suggest removing if practical.

eramongodb: Good catch. It seems like the commit-and-revert approach did not fully work out as intended. Manually removed the stray instances in the modified files instead.

kevinAlbs (approved):

PR #2309, opened 2026-06-01
CDRIVER-6343 check embedded length in `bson_new_from_buffer`

kevinAlbs:
Add an additional length check to `bson_new_from_buffer`. See CDRIVER-6343 for details.

### DIFF
--- a/src/libbson/src/bson/bson.c
+++ b/src/libbson/src/bson/bson.c
@@ -2041,7 +2041,7 @@ bson_new_from_buffer(uint8_t **buf, size_t *buf_len, bson_realloc_func realloc_f
          return NULL;
       }
       length = mlib_read_u32le(*buf);
-      if (length > *buf_len) {
+      if (length < 5 || length > *buf_len) {
          bson_free(bson);
          return NULL;
       }
--- a/src/libbson/doc/bson_init_static.rst
+++ b/src/libbson/doc/bson_init_static.rst
@@ -25,6 +25,8 @@ The :symbol:`bson_init_static()` function shall initialize a read-only :symbol:`
 
 The resulting `bson_t` has internal references and therefore must not be copied to avoid dangling pointers in the copy.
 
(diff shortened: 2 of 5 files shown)


src/libbson/src/bson/bson.c
```diff
@@ -2041,7 +2041,7 @@ bson_new_from_buffer(uint8_t **buf, size_t *buf_len, bson_realloc_func realloc_f
          return NULL;
       }
       length = mlib_read_u32le(*buf);
-      if (length > *buf_len) {
+      if (length < 5 || length > *buf_len) {
```
eramongodb, line 2044: Note: some cases are more strict (requiring exact length), e.g.:

https://github.com/mongodb/mongo-c-driver/blob/c33e2623f4e504042e31fc61e6b53b9ef5fbe861/src/libbson/src/bson/bson.c#L2000-L2005

and:

https://github.com/mongodb/mongo-c-driver/blob/c33e2623f4e504042e31fc61e6b53b9ef5fbe861/src/libbson/src/bson/bson.c#L1904-L1908

We may want to file a followup ticket to make these length validations more consistent and/or document deliberate inconsistencies.

eramongodb (approved):

src/libbson/src/bson/bson.c
kevinAlbs: `bson_new_from_buffer` [is tested](https://github.com/mongodb/mongo-c-driver/blob/c33e2623f4e504042e31fc61e6b53b9ef5fbe861/src/libbson/tests/test-bson.c#L1059-L1082) to ensure it can accept a buffer exceeding the size of the initial BSON data:

```
// Buffer is larger than the document. Expect it to be growable without reallocating.
```

I updated the documentation to explicitly note this is a supported use of `bson_new_from_buffer` and document that `bson_new_from_data` and `bson_init_static` reject non-matching lengths.

PR #1627, opened 2024-06-04
DRIVERS-2767 QE Range Protocol v2

adriandole:
Included in this PR:
- Sync range tests from specifications
- Replace `rangePreview` with `range`
- Expose the new `trimFactor` option for range queries
- Enable range v2 in libmongocrypt

PR #2310, opened 2026-06-02
CDRIVER-6293 update QE for `prefix` and `suffix`

kevinAlbs:
# Summary

Implements changes proposed in DRIVERS-3321:

- deprecate `text` APIs for newly named `string` APIs
- replace "prefixPreview" queryType with "prefix"
- replace "suffixPreview" queryType with "suffix"
- replace "TextPreview" algorithm with "String"

## Details

Though previous documentation of the `text` APIs noted "may not be backwards compatible", I see no precedent in the C driver for removing API in a minor release based on documentation alone. Even the rename of "rangePreview" to "range" in https://github.com/mongodb/mongo-c-driver/issues/1627 did not remove the `MONGOC_ENCRYPT_ALGORITHM_RANGEPREVIEW` macro. 

This PR includes prose tests added in https://github.com/mongodb/mongo-c-driver/pull/2289.

As added test improvements:
- Add a new test fixuture (`string_explicit_encryption_fixture`) to isolate string tests from other explicit encryption tests and closely model the spec.
- Apply write concern majority to match spec.

As a drive-by fix, a missing call to `_ctx_check_error` is added to ensure an error is propagated from libmongocrypt on a failed call to `mongoc_client_encryption_encrypt_opts_set_query_type`. This was used to verify setting "prefixPreview" / "suffixPreview" triggers an error.

C++ driver tests are disabled until the C++ driver updates calls to the new API names in CXX-3503.


kevinAlbs:
On further thought, I restored the "text" APIs as deprecated aliases. Though previous documentation noted "may not be backwards compatible", I see no precedent in the C driver for removing APIs in a minor release based on documentation alone. Even the rename of "rangePreview" to "range" in https://github.com/mongodb/mongo-c-driver/issues/1627 did not remove the `MONGOC_ENCRYPT_ALGORITHM_RANGEPREVIEW` macro. 

The additional names do not seem to be much maintenance burden. I would rather err towards caution and keep the deprecated names.

With the API restored, the `abi-compliance-task` [passed](https://spruce.corp.mongodb.com/version/6a21b819fe8c4d000762aa06).

mdb-ad (approved):

PR #2312, opened 2026-06-02
[CDRIVER-5705] Add a new documentation for versioning, API, and ABI guarantees

vector-of-bool:
Refer: CDRIVER-5705

# Summary

This changeset introduces a developer+user documentation page that describes the compatibility guarantees of the libraries. For now, it lives in `docs/dev/`, but it can be migrated to anywhere as needed.

I won't detail its contents as it speaks for itself, but I did find the process interesting enough to note here.

## Adding Documentation with LLM Assistance

I took this task as an opportunity to explore Claude's technical prose and source-processing abilities in combination, and found it exceptionally helpful in accelerating the workflow significantly, potentially saving many hours of work.

**First,** I asked it to consider existing information, including existing documentation and source code contents, preexisting knowledge of semantic versioning, and knowledge about software in general, to write a basic draft of the initial document.

The initial draft appeared decent at a high level, but needed some significant work. For example: it hallucinated the existence of `BSON_EXPERIMENTAL` and `MONGOC_EXPERIMENTAL` preprocessor macros to control experimental features, it referenced deprecated CMake export target names rather than the new target names, and seemed to be unfocused on what parts of the document were intended for end users versus project contributors. There were many smaller revisions needed to make it satisfactory for me.

Overall, about three quarters of the document needed to be rewritten or revised in some way. Despite this, it was able to infer a lot of information from our existing codebase and gave me a good basis for what needed to be done to be completed.

**Second,** after hand-revising the document, I asked it to proofread the document for spelling and grammar issues, semantic issues, ambiguities, and open questions. It found several basic editorial problems, but also found potentially conflicting information or poor phrasing.

I performed this step several times until I was satisfied that there were no more notable issues. Example Claude output:

Screenshot 2026-06-02 163559

**Third,** I asked it to verify that the new document does not conflict with existing documentation in the repository. This is what it found:

Screenshot 2026-06-02 164003

I hadn't noticed it, but we do in fact use the term "micro" and "patch" interchangeably throughout the codebase when "patch" is the more common term. I updated the document to note this synonym. It was also able to verify that the suspected misspelled CMake target name `mongo::shared` was inconsistent with what was found in the `CMakeLists.txt`.

**Fourth,** I was unhappy with a structural issue in the document: We had an "ABI Stability" section but no "API Stability" section. This felt like an annoying asymmetry. Claude had also created a "What Constitutes a Breaking Change?" section that detailed the API stability guarantees, which was somewhat redundant with the versioning and ABI stability section. I asked it to add an "API Stability" section and to remove the "Breaking Change" section, consolidating redundant information between everything. It did this very well, retaining all of my existing revisions and added details without introducing any new information.

Finally, I asked it to: "Find possible redundancies between the contents of this document and other documentation in the repository." I am wary when adding new documentation of duplicating existing documentation, as they can quickly become out-of-sync or contradictory. It found the following:

Screenshot 2026-06-02 165033

I asked it to apply the suggested changes, resulting in commit 3848ce3d.

### DIFF
--- a/docs/dev/api-abi-policy.rst
+++ b/docs/dev/api-abi-policy.rst
@@ -1,0 +1,257 @@
+##################
+API and ABI Policy
+##################
+
+This page describes the stability guarantees that the C driver project makes
+about its public API and ABI across releases, how deprecation works, and what
+is explicitly *not* covered by these guarantees.
+
+.. note::
+
+   This policy applies to both **libmongoc** and **libbson**.
+
+
+.. _api-abi-policy.versioning:
+
+Versioning Scheme
+#################
+
+The C libraries use a ``<major>.<minor>.<micro>`` versioning scheme that is
+based on `Semantic Versioning <https://semver.org>`_. Prerelease and development
+versions include a suffix to indicate that the contents of the libraries are
+unstable and subject to change.
+
+- A **major** version indicates overall breaking changes. Different major
+  versions offer very few compatibility guarantees in their API and ABI.
+- A **minor** version indicates the addition of new APIs and features to the
+  libraries, or possible deprecation of APIs. Programs written against earlier
+  minor versions should continue to compile and link with a newer minor version
+  with minimal or no changes.
+- A **micro** (also known as "patch") version indicates bugfixes and tweaks and
+  will not deprecate or add any major features to libraries. Programs written
+  against earlier micro versions should compile and link against newer micro
+  with no changes unless security considerations require API or ABI changes.
+
+
+.. _api-abi-policy.public-api:
+
+What Is the Public API?
+#######################
+
+The public API consists of:
+
+* Every symbol (function, type, macro, constant) declared in a **public header**
+  — a header that is installed alongside the library and has no ``-private``
+  suffix in its name (e.g. ``mongoc/mongoc-client.h``, ``bson/bson.h``), except
+  for those noted below.
+* Every function annotated with ``MONGOC_EXPORT(...)`` or ``BSON_EXPORT(...)``.
+
+The following are **not** part of the public API and carry **no stability
+guarantee**:
+
+* Headers with a ``-private.h`` suffix (e.g. ``mongoc-client-private.h``) and
+  the contents thereof.
+* Anything under ``src/common/`` (the internal shared utility library).
+* Symbols prefixed with an underscore or otherwise documented as internal.
+* The layout of any struct whose definition lives in a ``-private.h`` header,
+  even when the opaque ``typedef`` is public.
+* The content of ``src/kms-message/``, a vendored library with its own
+  versioning.
+
+
+Header Inclusion
+****************
+
+All library header ``#include`` directives should use the ``bson/`` or
+``mongoc/`` directory prefix. Including unqualified header filenames by
+modifying the header search paths is unsupported.
+
+Header files that have the ``bson/bson-*`` or ``mongoc/mongoc-*`` (with the
+hyphen) prefix should never be included directly, despite being part of the
+public API. Only their contents are part of the public API, not the filename or
+the physical location of the entities contained within. The content of these
+headers may be moved in any version, or those headers may be removed entirely.
+Instead, users should include the umbrella headers
+``<bson/bson.h>``/``<mongoc/mongoc.h>`` or they may directly include one of the
+headers without the ``bson-`` or ``mongoc-`` prefix in its filename.
+
+
+.. _api-abi-policy.api:
+
+API Stability
+#############
+
+Within a **major** version series the project guarantees source-level backward
+compatibility: a program that compiles cleanly against version ``X.Y.Z`` using
+only the public API is expected to compile without modification against any later
+``X.Y'.Z'`` release in the same major series. Specifically:
+
+* **No removal** of public symbols — functions, types, macros, and constants
+  present in a release remain available in all subsequent releases within the
+  same major version.
+* **No rename** of a public symbol without providing a compatibility alias for
+  the original name.
+* **No incompatible signature changes** — the parameter and return types of
+  public functions do not change in ways that require callers to be updated.
+  Note that cv-qualifiers may be added or removed from parameter or return types
+  in ways that do not affect existing callers.
+* **No change in documented contract** — the observable behavior, preconditions,
+  postconditions, and error semantics of a public function do not change in a way
+  that causes a previously-correct caller to misbehave.
+* **No change in the values** of public constants or enum members in a way that
+  alters the semantics of existing programs.
+
+As a result, a program that compiles cleanly against version ``X.Y.Z`` is
+expected to compile and link without modification against any later ``X.Y'.Z'``
+release in the same major series, provided the program only uses the public API.
+
+Within a **minor** release, new public symbols may be added (see
+:ref:`api-abi-policy.additions`). Within a **micro** release, no new public
+symbols are introduced and existing behavior changes only to correct bugs.
+
+Deprecation (see :ref:`api-abi-policy.deprecation`) is the only mechanism by
+which a public symbol can eventually be removed; removal only occurs at a
+**major** release boundary. Until a symbol is actually removed it remains fully
+functional — not merely a stub.
+
+.. rubric:: Bug fixes and behavior
+
+A change that corrects a function's behavior to match its documented contract is
+not an API break, even if some callers relied on the erroneous behavior.
+Undocumented or incorrect behavior carries no stability guarantee.
+
+.. rubric:: Macros
+
+The observable behavior of a public macro is stable within a major version
+series. The specific token sequence of a macro's expansion is not guaranteed and
+may change between releases — for example, to replace a function-like macro with
+an inline function or to add internal casts. Code that depends on a specific
+expansion rather than on the macro's documented behavior is unsupported.
+
+.. rubric:: Error codes
+
+New error codes and domain values may be introduced in any **minor** release.
+Callers should handle unknown error codes gracefully rather than treating the set
+of possible values as exhaustive.
+
+
+.. _api-abi-policy.abi:
+
+ABI Stability
+#############
+
+Within a **major** version series the project guarantees:
+
+* **No removal** of exported symbols, and no change to an exported symbol's
+  link name or DLL ordinal.
+* **No signature changes** to exported functions (parameter types, return type,
+  calling convention), except for the possible addition of cv-qualifiers to
+  pointer parameters or removal of cv-qualifiers from pointer return types.
+* **No breaking changes** to the size or layout of any completely-defined struct
+  reachable in the public API, even if the contents of that struct are not
+  themselves part of the API. Where feasible, structs are kept opaque (accessed
+  only through accessor functions) to preserve this freedom.
+
+As a result, a program compiled against version ``X.Y.Z`` is expected to load
+and run correctly against any later ``X.Y'.Z'`` shared library, provided the
+program only uses the public API.
+
+.. rubric:: Incomplete types
+
+Many types (e.g. ``mongoc_client_t``) are intentionally incomplete in the public
+API and are only passed/returned by address. Callers should never rely on the
+size, alignment, or fields of such a type. Adding, removing, or rearranging the
+content of such types is not an ABI break.
+
+.. rubric:: Public aggregate types
+
+A small number of structs expose their fields in public headers as part of their
+public API (e.g. ``bson_iter_t``). Changes to these structs **are** ABI breaks
+and are therefore prohibited within a **major** version series.
+
+.. rubric:: Compilation settings
+
+Certain preprocessor macros and CMake configure-time build parameters may affect
+the ABI of the resulting build. When possible, programs should not attempt to
+link translation units together that are not compiled with the same set of
+preprocessor definitions.
+
+Users should not manually define preprocessor macros during their compilation
+that could affect the content of the exposed library APIs. Instead, users should
+rely on the CMake imported targets or pkg-config definitions to set the
+appropriate compile/link settings for their program.
+
+.. note::
+
+   The project runs an ABI compliance check in CI (the ``abi-compliance-check``
+   Evergreen task) that compares the current branch against the most recent
+   stable release. Any unexpected breakage should be caught before merge.
+
+
+.. _api-abi-policy.deprecation:
+
+Deprecation Policy
+##################
+
+Symbols that need to be removed follow a two-step cycle:
+
+1. **Mark deprecated.** Annotate the declaration with ``BSON_DEPRECATED("…")``
+   or ``BSON_DEPRECATED_FOR(replacement)`` and update the documentation to note
+   the deprecation and its replacement. This is a non-breaking change that may
+   land in any **minor** release. **mongoc** also uses the ``BSON_DEPRECATED``
+   macros (there is no ``MONGOC_DEPRECATED`` equivalent).
+
+2. **Remove.** The symbol is removed no earlier than the *next major release*
+   after it was deprecated, giving users at least one full major-version cycle
+   to migrate.
+
+A deprecated symbol must remain present and functional (not just a stub) until
+it is removed.
+
+
+.. _api-abi-policy.additions:
+
+Adding New Symbols
+##################
+
+New public symbols may be added in any **minor** release. When adding a symbol:
+
+1. Declare it in the appropriate public header.
+2. For functions, annotate it with ``MONGOC_EXPORT`` or ``BSON_EXPORT``.
+3. Add a ``.rst`` documentation page in ``src/libmongoc/doc/`` or
+   ``src/libbson/doc/``.
+4. Verify it appears in the ABI compliance report under *Added Symbols* (see
+   the ``abi-compliance-check`` Evergreen task artifact).
+
+Adding new symbols is **not** an ABI break. However, a new symbol that clashes
+with a name a downstream project already defines privately can cause linker
+confusion; prefer descriptive, namespaced names (``mongoc_`` or ``bson_``
+prefix).
+
+
+Other Notes
+###########
+
+Pre-release versions
+********************
+
+Any release with a pre-release label (e.g. ``2.0.0-alpha1``) carries **no
+stability guarantee**. Any changes that are part of a prerelease may be removed
+or modified without notice between pre-release versions.
+
+
+Experimental APIs
+*****************
+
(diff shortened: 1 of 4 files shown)


docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,257 @@
...
+- A **major** version indicates overall breaking changes. Different major
+  versions offer very few compatibility guarantees in their API and ABI.
+- A **minor** version indicates the addition of new APIs and features to the
+  libraries, or possible deprecation of APIs. Programs written against earlier
+  minor versions should continue to compile and link with a newer minor version
+  with minimal or no changes.
+- A **micro** (also known as "patch") version indicates bugfixes and tweaks and
+  will not deprecate or add any major features to libraries. Programs written
+  against earlier micro versions should compile and link against newer micro
+  with no changes unless security considerations require API or ABI changes.
```
eramongodb: > I hadn't noticed it, but we do in fact use the term "micro" and "patch" interchangeably throughout the codebase when "patch" is the more common term. I updated the document to note this synonym.

Recommend using "patch" by default (switch use of "micro" and "patch" in this paragraph + in `<major>.<minor>.<micro>` above), for consistency with SemVer spec (mentioned above) and with the C++ Driver's API Versioning policy:

> Given a version number MAJOR.MINOR.PATCH, increment the:
> 1. MAJOR version when you make incompatible API changes
> 2. MINOR version when you add functionality in a backward compatible manner
> 3. PATCH version when you make backward compatible bug fixes
> Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

as well as other wording in referenced to "patch releases" (not "micro release")

docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,257 @@
...
+
+* **No removal** of public symbols — functions, types, macros, and constants
+  present in a release remain available in all subsequent releases within the
+  same major version.
+* **No rename** of a public symbol without providing a compatibility alias for
+  the original name.
+* **No incompatible signature changes** — the parameter and return types of
+  public functions do not change in ways that require callers to be updated.
+  Note that cv-qualifiers may be added or removed from parameter or return types
+  in ways that do not affect existing callers.
```
eramongodb:
```suggestion
  Note that type-qualifiers may be added or removed from parameter or return types
  in ways that do not affect existing callers.
```
Nit: unlike C++, C uses "type qualifiers" to refer to `const`, `volatile`, and `restrict`, i.e. per [C99 spec](https://port70.net/~nsz/c/c99/n1256.html#6.7.3):

```
type-qualifier:
       const
       restrict
       volatile
```

docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,257 @@
...
+  calling convention), except for the possible addition of cv-qualifiers to
+  pointer parameters or removal of cv-qualifiers from pointer return types.
+* **No breaking changes** to the size or layout of any completely-defined struct
+  reachable in the public API, even if the contents of that struct are not
+  themselves part of the API. Where feasible, structs are kept opaque (accessed
+  only through accessor functions) to preserve this freedom.
+
+As a result, a program compiled against version ``X.Y.Z`` is expected to load
+and run correctly against any later ``X.Y'.Z'`` shared library, provided the
+program only uses the public API.
```
eramongodb: Recommend adding a bullet point concerning the stability of the _soname_ of the shared library, i.e. for reference, from the C++ Driver's [ABI Versioning policy](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/reference/api-abi-versioning/abi-versioning/#build-system-compatibility):

> - **The soname is considered part of the stable ABI.** This means, in contrast to the public API, `BSONCXX_OUTPUT_NAME` is considered part of the stable ABI, as it directly impacts the soname of the resulting bsoncxx shared library.
>
> - The "soname" is not applicable on Windows. See [Shared Libraries (MSVC Only).](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/reference/api-abi-versioning/#std-label-cpp-shared-lib-msvc)

Note the CMake option `(BSON|MONGOC)_OUTPUT_BASENAME` is a rough equivalent to the C++ Driver's `(BSONCXX|MONGOC)_OUTPUT_NAME`.

docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,257 @@
...
+Callers should handle unknown error codes gracefully rather than treating the set
+of possible values as exhaustive.
+
+
+.. _api-abi-policy.abi:
+
+ABI Stability
+#############
+
+Within a **major** version series the project guarantees:
```
eramongodb: Add a note in this section reiterating the escape hatch for security vulnerability patches as mentioned above:

> ... unless security considerations require API or ABI changes.

Note that the C++ Driver's API/ABI Versioning policy does not contain this escape hatch (which doesn't really need it either due to its use of versioned ABI namespaces), but we may want to keep it for the C Driver due to its stronger API-ABI equivalence.

docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,257 @@
...
+   the ``abi-compliance-check`` Evergreen task artifact).
+
+Adding new symbols is **not** an ABI break. However, a new symbol that clashes
+with a name a downstream project already defines privately can cause linker
+confusion; prefer descriptive, namespaced names (``mongoc_`` or ``bson_``
+prefix).
+
+
+Other Notes
+###########
```
eramongodb, line 266: Recommend adding a section here or above concerning the independence of _build system_ breaking changes from the API/ABI versioning policy to complement the existing CMake target section below, i.e. for reference, from the C++ Driver's [API Versioning policy](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/reference/api-abi-versioning/api-versioning/#build-system-compatibility):

> Although the public API depends on the configuration of the build system, the build system itself is NOT part of the public API. [...] We support the stability of the public API _per build configuration_. We do not support compatibility of the public API across different build configurations.

and from the [ABI Versioning policy](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/reference/api-abi-versioning/abi-versioning/#build-system-compatibility):

> The stable ABI policy concerning properties of the build system is mostly the same as that for the public API (see [API Build System Compatibility](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/reference/api-abi-versioning/api-versioning/#std-label-cpp-api-build-sys-compat)) with the following differences:
>
> - Instead of properties that directly impact the public API, properties that directly impact the stable ABI are considered part of the stable ABI.
> - ... (notes concerning shared library soname stability) ...

docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,257 @@
...
+
+New error codes and domain values may be introduced in any **minor** release.
+Callers should handle unknown error codes gracefully rather than treating the set
+of possible values as exhaustive.
+
+
+.. _api-abi-policy.abi:
+
+ABI Stability
+#############
```
eramongodb, line 156: This section should probably mention whether or not we want the ABI version number to remain synced with the API major version number. This has been, and still is, the case. If we want to (continue to) enforce this going forward, it should be documented by this policy. fwiw in addition to C++ Driver's ABI version number (soon-to-be set to `1`) being distinct from the API major version number (targeting `4`), the C++ Driver's [ABI Versioning policy](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/reference/api-abi-versioning/abi-versioning/#root-namespace-redeclarations) also goes out of its way to distinguish the soversion from the ABI namespace version (to support incremental deprecation/removal of ABI symbols without forcing namespace-wide invalidation):

> The integer used by an ABI namespace does _NOT_ directly correspond to an ABI version number. The ABI version number is bumped whenever _any_ binary incompatible change occurs, even when it is the removal of a _single_ symbol from the stable ABI. An ABI version number bump from `A` to `B` does NOT imply the deprecation or removal of symbols declared in the ABI namespace `vA` (if one exists).

docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,257 @@
...
+What Is the Public API?
+#######################
+
+The public API consists of:
+
+* Every symbol (function, type, macro, constant) declared in a **public header**
+  — a header that is installed alongside the library and has no ``-private``
+  suffix in its name (e.g. ``mongoc/mongoc-client.h``, ``bson/bson.h``), except
+  for those noted below.
+* Every function annotated with ``MONGOC_EXPORT(...)`` or ``BSON_EXPORT(...)``.
```
eramongodb:
```suggestion
* Every function annotated with ``MONGOC_EXPORT`` or ``BSON_EXPORT``.
```
Consistency with reference to these macros below.

docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,257 @@
...
+* Anything under ``src/common/`` (the internal shared utility library).
+* Symbols prefixed with an underscore or otherwise documented as internal.
+* The layout of any struct whose definition lives in a ``-private.h`` header,
+  even when the opaque ``typedef`` is public.
+* The content of ``src/kms-message/``, a vendored library with its own
+  versioning.
+
+
+Header Inclusion
+****************
```
eramongodb, line 63: Suggest also documenting the slightly unusual install directory structure (here or in the Versioning Scheme section above):

```
<prefix>/include
└── bson-2.3.0
    └── bson
```

Perhaps a note about `BSON_INSTALL_INCLUDEDIR` may help users who want to avoid this extra layer?

docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,257 @@
...
+Experimental APIs
+*****************
+
+APIs that are documented as **experimental** may change or be removed in any
+release, including patch releases, regardless of the usual policy. Such APIs
+will be clearly marked in the documentation.
+
+
+CMake integration API
+*********************
```
eramongodb: Suggest clearly recommending CMake as the preferable method for importing libraries as a package, i.e. from the C++ Driver's [API & ABI Versioning policy](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/reference/api-abi-versioning/#header-file-structure):

> In all cases, CMake package config files are the recommended method for importing C++ driver libraries.

Suggest also documenting our support for `pkg-config` (or `pkgconf`?) config files as fallback for non-CMake-based build systems.

docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,257 @@
...
+Versioning Scheme
+#################
+
+The C libraries use a ``<major>.<minor>.<micro>`` versioning scheme that is
+based on `Semantic Versioning <https://semver.org>`_. Prerelease and development
+versions include a suffix to indicate that the contents of the libraries are
+unstable and subject to change.
+
+- A **major** version indicates overall breaking changes. Different major
+  versions offer very few compatibility guarantees in their API and ABI.
```
eramongodb:
```suggestion
- A **major** version indicates overall breaking changes. Different major
  versions offer no compatibility guarantees in their API and ABI.
```
Suggest using simpler-but-stronger language here.

docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,257 @@
...
+- A **micro** (also known as "patch") version indicates bugfixes and tweaks and
+  will not deprecate or add any major features to libraries. Programs written
+  against earlier micro versions should compile and link against newer micro
+  with no changes unless security considerations require API or ABI changes.
+
+
+.. _api-abi-policy.public-api:
+
+What Is the Public API?
+#######################
```
eramongodb, line 40: Suggest adding bullet points for:

- documented behavior (part of public API)
- undocumented behavior (_not_ part of public API).

For reference, from the C++ Driver's [API Versioning policy](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/reference/api-abi-versioning/api-versioning/#source-compatibility):

> The _documented_ behavior of a public entity is also considered part of the public API. Behavior that is not documented is considered library-level undefined behavior and is not supported. If there is behavior that _should_ be documented but is currently not, please submit a bug report.

This is our anti-[Hyrum's Law](https://www.hyrumslaw.com/) defense (well, as much as one is able to).

docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,257 @@
...
+
+.. _api-abi-policy.abi:
+
+ABI Stability
+#############
+
+Within a **major** version series the project guarantees:
+
+* **No removal** of exported symbols, and no change to an exported symbol's
+  link name or DLL ordinal.
```
eramongodb:
```suggestion
* **No removal** of exported symbols, and no change to an exported symbol's
  link name.
```
Suggest avoiding Windows-specific terminology (+ I don't think this is a feature we use/depend-on either?).

docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,257 @@
...
+* **No breaking changes** to the size or layout of any completely-defined struct
+  reachable in the public API, even if the contents of that struct are not
+  themselves part of the API. Where feasible, structs are kept opaque (accessed
+  only through accessor functions) to preserve this freedom.
+
+As a result, a program compiled against version ``X.Y.Z`` is expected to load
+and run correctly against any later ``X.Y'.Z'`` shared library, provided the
+program only uses the public API.
+
+.. rubric:: Incomplete types
```
eramongodb, line 192:
```suggestion
program only uses the public API.

.. rubric:: Exported Symbols

...

.. rubric:: Incomplete types
```

Suggest adding a section documenting our use of export macros (which explicitly denote the set of symbols exported in the ABI) and their relationship to the public API (are all exported symbols also part of the public API?). For reference, from the C++ Driver's [ABI Versioning policy](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/reference/api-abi-versioning/abi-versioning/#binary-compatibility):

> Some entities that are part of the public API may not be part of the stable ABI (e.g. inline functions, inline variables, and template instantiations of functions and variables). Conversely, some entities that are part of the stable ABI may not be part of the public API (e.g. an exported private member function).

eramongodb (changes requested):

docs/dev/api-abi-policy.rst
vector-of-bool: There isn't really an "ABI version number" to be documented in the C library, as it doesn't have any way to do namespacing that would make it relevant. I want to avoid implying that there could be some de-sync between the ABI and the major version number, unless we wish to do ABI-breaking changes with minor versions (completely possible, but should be discussed).

docs/dev/api-abi-policy.rst
vector-of-bool: Added. I am avoiding referencing specific CMake parameters to prevent the temptation for this guide becoming too "how to".

docs/dev/api-abi-policy.rst
eramongodb: We explicitly set the SOVERSION to the API major version number, e.g. here: https://github.com/mongodb/mongo-c-driver/blob/115b2a5ad8d26114738b724457c687d9534e7a37/src/libbson/CMakeLists.txt#L240-L244

This suggestion is to explicitly document that the API major version number _is_ the soversion: that is, any ABI breaking change will constitute an API major version bump.

docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,257 @@
...
+
+What Is the Public API?
+#######################
+
+The public API consists of:
+
+* Every symbol (function, type, macro, constant) declared in a **public header**
+  — a header that is installed alongside the library and has no ``-private``
+  suffix in its name (e.g. ``mongoc/mongoc-client.h``, ``bson/bson.h``), except
+  for those noted below.
```
kevinAlbs:
```suggestion
  — a header that is installed alongside the library
```

Consider simplifying. "except for those noted" suggests some headers below are installed. But `-private` and internal headers are not installed.

I expect the references to those headers below can be similarly removed.

docs/dev/api-abi-policy.rst
kevinAlbs: Consider removing this bullet. I expect this is redundant with the "Every symbol [...] declared in a **public header**" above. I do not expect we will export a function without including it in the public header.

docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,257 @@
...
+* Every function annotated with ``MONGOC_EXPORT(...)`` or ``BSON_EXPORT(...)``.
+
+The following are **not** part of the public API and carry **no stability
+guarantee**:
+
+* Headers with a ``-private.h`` suffix (e.g. ``mongoc-client-private.h``) and
+  the contents thereof.
+* Anything under ``src/common/`` (the internal shared utility library).
+* Symbols prefixed with an underscore or otherwise documented as internal.
```
kevinAlbs:
```suggestion
* Symbols prefixed with an underscore or otherwise documented as internal or experimental.
```

Suggest echoing the "Experimental APIs" section below.

docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,308 @@
...
+of possible values as exhaustive.
+
+
+.. _api-abi-policy.abi:
+
+ABI Stability
+#############
+
+Except in the case of security patches, within a **major** version series, the
+project guarantees the following:
```
eramongodb:
```suggestion
Except in the case of security patches, the
project guarantees the following within a **major** version series:
```
Simpler sentence structure.

docs/dev/api-abi-policy.rst
```diff
@@ -0,0 +1,308 @@
...
+
+
+Other Notes
+###########
+
+Build configuration compatibility
+*********************************
+
+The build parameters are themselves subject to the same stability guarantees as
+the C API, and follow the same deprecation policy.
```
eramongodb: As suggested in https://github.com/mongodb/mongo-c-driver/pull/2312#discussion_r3350109364, I do not think we want to tie build system stability to API stability. This is difficult to reliably ensure, difficult to define the scope of stability for, and has not been reliably practiced across past v1 releases (both patch and minor). Recommend qualifying this statement to narrow the guarantee to build parameters which directly affect the API/ABI of the resulting shared libraries.

docs/dev/api-abi-policy.rst
vector-of-bool: "except for those noted" is intended to apply to the "symbols", not to "headers". Wording changed to clarify.

eramongodb (approved):

kevinAlbs (approved):

PR #2331, opened 2026-06-29
gate test executable behind `BUILD_TESTING`

kevinAlbs:
This PR applies an change equivalent to the [libmongocrypt patch](https://github.com/mongodb/libmongocrypt/blob/906010bdf5a9c7d5fc40743c49b313cdd57effa4/etc/mongo-common-test-harness.patch) to gate building the test executable in `src/common` behind `BUILD_TESTING`.


vector-of-bool (approved):
LGTM

PR #2338, opened 2026-07-06
CDRIVER-6346 Fix calls to Windows APIs when UNICODE defined

kevinAlbs:
# Summary

Fix calls to Windows APIs when UNICODE defined.

# Details

A CI compile task is added on Windows with UNICODE macros defined and the MSVC "incompatible types" warning as an error (`/we4133`) to test for regressions.

A notably irregular case is `DnsQuery_UTF8`. The out-param of `DnsQuery_UTF8` is `PDNS_RECORD`. Though `PDNS_RECORD` is defined in terms of wide-character strings when `UNICODE` is defined, `DnsQuery_UTF8` [documents](https://learn.microsoft.com/en-us/windows/win32/api/windns/nf-windns-dnsquery_utf8): "for UTF-8 encoding". So explicit casts are used.


vector-of-bool (approved):
LGTM

PR #2340, opened 2026-07-07
CDRIVER-6302 send afterClusterTime on writes in CC sessions

connorsmacd:
CDRIVER-6302
DRIVERS-3274

# Summary

This PR modifies read concern handling to send `afterClusterTime` on certain write operations in causally-consistent sessions. Changes include:

- Support `dropDatabase` in unified test runner
- Add new unified tests:
  - `causal-consistency/causal-consistency-clientBulkWrite`
  - `causal-consistency/causal-consistency-write-commands`
  - `transactions/commit`
  - `transactions/retryable-writes`
- Remove defunct legacy tests (NOTE: while the transaction legacy tests were replaced with updated equivalent unified tests, the tests for with-transaction (transaction-convenient-api) do not yet use the unified replacements due to CDRIVER-4003):
  - `transactions/commit`
  - `transactions/retryable-writes`
  - `with-transaction/callback-aborts`
  - `with-transaction/callback-commits`
- Update `_mongoc_client_session_append_read_concern` to accept a command name and send `afterClusterTime` if the command name is in the allowlist in addition to the other checks.
- Update session tests to expect read concern for some write commands.


src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -1467,17 +1498,20 @@ _mongoc_client_session_append_txn(mongoc_client_session_t *session, bson_t *cmd,
  *       are "level" and/or "afterClusterTime" - if both are empty, don't add
  *       read concern.
  *
+ *       In a causally-consistent session, afterClusterTime is also sent on
+ *       the allowlisted write commands (see
+ *       _mongoc_client_session_is_cc_write_command), identified by
```
kevinAlbs:
```suggestion
 *       _mongoc_write_command_supports_after_cluster_time), identified by
```

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -1456,6 +1456,37 @@ _mongoc_client_session_append_txn(mongoc_client_session_t *session, bson_t *cmd,
...
+      "dropDatabase",
+      "dropIndexes",
+   };
+
+   if (!command_name) {
+      return false;
+   }
+
+   for (size_t i = 0; i < sizeof(allowlist) / sizeof(allowlist[0]); i++) {
+      if (!bson_strcasecmp(command_name, allowlist[i])) {
```
kevinAlbs: Minor: but this can use `strcmp` since commands are case-sensitive (as is done [elsewhere](https://github.com/mongodb/mongo-c-driver/blob/a21e2f0f04dc0b2e99717ac2844eecd049e3daa5/src/libmongoc/src/mongoc/mongoc-client-session.c#L623) in this file):

```suggestion
      if (!strcmp(command_name, allowlist[i])) {
```

src/libmongoc/src/mongoc/mongoc-client-session.c
```diff
@@ -1499,7 +1533,10 @@ _mongoc_client_session_append_read_concern(const mongoc_client_session_t *cs,
       return;
    }
 
-   has_timestamp = (txn_state == MONGOC_INTERNAL_TRANSACTION_STARTING || is_read_command) &&
+   const bool after_cluster_time_is_supported =
+      is_read_command || _mongoc_write_command_supports_after_cluster_time(command_name);
```
kevinAlbs: The [spec notes](https://github.com/mongodb/specifications/blob/ea6be202feb255242077cbc53527559e79e5cd00/source/causal-consistency/causal-consistency.md#causally-consistent-read-and-write-commands):

> The Read and Write Concern Specification states that drivers MUST NOT add a `readConcern` field to commands that are run using a generic `runCommand` method. The same is true for causal consistency, so commands that are run using `runCommand` MUST NOT have an `afterClusterTime` field added to them.

As I understand, `mongoc_client_command_with_opts` is the generic `runCommand` helper and there are read/write variants:
- `mongoc_client_read_command_with_opts` sets `is_read_command` internally
- `mongoc_client_write_command_with_opts` sets `is_write_command` internally

Modifying the `test_cmd` helper in `test-mongoc-client-session.c` to run a "delete" command fails:

```c
static void
test_cmd(session_test_t *test)
{
   // Test both a read and write command through the generic helper:
   bson_t *cmd_read = tmp_bson("{'listCollections': 1}");
   bson_t *cmd_write = tmp_bson("{'delete': 'collection', 'deletes': [{'q': {}, 'limit': 1}]}");
   test->succeeded =
      mongoc_client_command_with_opts(test->client, "db", cmd_read, NULL, &test->opts, NULL, &test->error) &&
      mongoc_client_command_with_opts(test->client, "db", cmd_write, NULL, &test->opts, NULL, &test->error);
}
// Error: "Command should not have included readConcern"
```

So I think this check should be changed to also check `is_write_command` to exclude write commands run through the generic helper, like:

```c
const bool after_cluster_time_is_supported =
      is_read_command || (is_write_command && _mongoc_write_command_supports_after_cluster_time(command_name));
```

connorsmacd: Good catch. I implemented the fix like you suggested with a new session test case to validate it.

kevinAlbs (approved):
LGTM!

PR #2342, opened 2026-07-08
Use "default_quay_registry" for quay.io images

eramongodb:
Following [DEVPROD-32961](https://jira.mongodb.org/browse/DEVPROD-32961), Amazon ECR pull-through cache now supports Quay images. This should improve the reliability of Earthly tasks when pulling CentOS Stream images. A new `default_quay_registry` is used to specify Quay images due to Amazon ECR requiring a different format than DockerHub images. The old `default_search_registry` was renamed to `default_docker_registry` for disambiguation.


kevinAlbs (approved):
Thank you!

PR #2346, opened 2026-07-13
CDRIVER-6367 fix CMake 4.4.0 configure

kevinAlbs:
To address a newly observed error in CMake 4.4.0:
```
CMake Error: The warning category "error -DCMAKE_CXX_LINK_EXECUTABLE='echo not linking now...'" is not known.
CMake Error at cmake-build/_deps/embedded_mcd-src/src/libmongoc/CMakeLists.txt:425 (TRY_COMPILE):
  Failed to configure test project build system.
  cmake-build/_deps/embedded_mcd-src/src/libmongoc/CMakeLists.txt:449 (mongoc_get_accept_args)
```

The `-Werror` appears intended for the compiler (not `cmake`) so is moved from `CMAKE_FLAGS` to `COMPILE_DEFINITIONS`.

### DIFF
--- a/.evergreen/scripts/compile-libmongocrypt.sh
+++ b/.evergreen/scripts/compile-libmongocrypt.sh
@@ -12,7 +12,18 @@ compile_libmongocrypt() {
   # `src/kms-message`.
   #
   # Run `.evergreen/scripts/kms-divergence-check.sh` to ensure that there is no divergence in the copied files.
-  declare -r version="1.20.0"
+  declare -r version="1.21.0-dev"
+
+  {
+    git clone -q https://github.com/mongodb/libmongocrypt || return
+    cd libmongocrypt || return
+    # Check out libmongocrypt commit that pins to CMake 4.3.4 to work around MONGOCRYPT-952 until libmongocrypt updates the C driver.
+    git checkout c2f80a7153da3537fe32916eef8ff1d4cc08bc67 || return
+    cd .. || return
+  }
+  # TODO: after libmongocrypt is updated to use C driver 1.21.0, update the above to:
+  # git clone --depth=1 -q https://github.com/mongodb/libmongocrypt --branch 1.21.0 || return
+
 
   git clone --depth=1 -q https://github.com/mongodb/libmongocrypt --branch "${version:?}" || return
 
--- a/.evergreen/scripts/compile-libmongocrypt.sh
+++ b/.evergreen/scripts/compile-libmongocrypt.sh
@@ -15,7 +15,9 @@ compile_libmongocrypt() {
   declare -r version="1.21.0-dev" # TODO: set to 1.21.0 once released and change `if true` to `if false`.
   if true; then
     git clone -q https://github.com/mongodb/libmongocrypt || return
-    git -C libmongocrypt checkout --detach "63fee9ff9e2748818bfe9204038e07d455777ab3" || return
+    cd libmongocrypt || return
+    git libmongocrypt checkout --detach "63fee9ff9e2748818bfe9204038e07d455777ab3" || return
+    cd .. || return
   else
     git clone --depth=1 -q https://github.com/mongodb/libmongocrypt --branch "${version:?}" || return
   fi
--- a/.evergreen/scripts/compile-libmongocrypt.sh
+++ b/.evergreen/scripts/compile-libmongocrypt.sh
@@ -16,7 +16,7 @@ compile_libmongocrypt() {
   if true; then
     git clone -q https://github.com/mongodb/libmongocrypt || return
     cd libmongocrypt || return
-    git libmongocrypt checkout --detach "63fee9ff9e2748818bfe9204038e07d455777ab3" || return
+    git checkout --detach "63fee9ff9e2748818bfe9204038e07d455777ab3" || return
     cd .. || return
   else
     git clone --depth=1 -q https://github.com/mongodb/libmongocrypt --branch "${version:?}" || return
--- a/.evergreen/scripts/compile-libmongocrypt.sh
+++ b/.evergreen/scripts/compile-libmongocrypt.sh
@@ -12,20 +12,13 @@ compile_libmongocrypt() {
   # `src/kms-message`.
   #
   # Run `.evergreen/scripts/kms-divergence-check.sh` to ensure that there is no divergence in the copied files.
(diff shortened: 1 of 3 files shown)


.evergreen/scripts/compile-libmongocrypt.sh
```diff
@@ -12,7 +12,18 @@ compile_libmongocrypt() {
...
+
+  {
+    git clone -q https://github.com/mongodb/libmongocrypt || return
+    cd libmongocrypt || return
+    # Check out libmongocrypt commit that pins to CMake 4.3.4 to work around MONGOCRYPT-952 until libmongocrypt updates the C driver.
+    git checkout c2f80a7153da3537fe32916eef8ff1d4cc08bc67 || return
+    cd .. || return
+  }
+  # TODO: after libmongocrypt is updated to use C driver 1.21.0, update the above to:
+  # git clone --depth=1 -q https://github.com/mongodb/libmongocrypt --branch 1.21.0 || return
```
eramongodb:
```suggestion
  declare -r version="63fee9ff9e2748818bfe9204038e07d455777ab3" # TODO: set to 1.21.0 once released and change `if true` to `if false`.
  if true; then
    git clone -q https://github.com/mongodb/libmongocrypt || return
    git -C libmongocrypt checkout --detach "${version:?}" || return
  else
    git clone --depth=1 -q https://github.com/mongodb/libmongocrypt --branch 1.21.0 || return
  fi
```
Suggest preserving the commit-checkout branch for future maintainability.

Note: also update the commit hash to include https://github.com/mongodb/libmongocrypt/commit/63fee9ff9e2748818bfe9204038e07d455777ab3.

src/libmongoc/CMakeLists.txt
```diff
@@ -423,8 +423,10 @@ function (mongoc_get_accept_args ARG2 ARG3)
    ")
 
    TRY_COMPILE (RES ${CMAKE_CURRENT_BINARY_DIR}
-   ${CMAKE_CURRENT_BINARY_DIR}/accept_test${VAR}.c CMAKE_FLAGS
-   "-Werror -DCMAKE_CXX_LINK_EXECUTABLE='echo not linking now...'" OUTPUT_VARIABLE LOG2)
+   ${CMAKE_CURRENT_BINARY_DIR}/accept_test${VAR}.c
+   COMPILE_DEFINITIONS "-Werror"
+   CMAKE_FLAGS "-DCMAKE_C_LINK_EXECUTABLE=echo not linking now..."
+   OUTPUT_VARIABLE LOG2)
```
eramongodb: Suggest changing to the following:
```cmake
   TRY_COMPILE (RES
      "${CMAKE_CURRENT_BINARY_DIR}"
      SOURCES "${CMAKE_CURRENT_BINARY_DIR}/accept_test${VAR}.c"
      CMAKE_FLAGS "-DCOMPILE_DEFINITIONS=-Werror" "-DCMAKE_CXX_LINK_EXECUTABLE='echo not linking now...'"
      OUTPUT_VARIABLE LOG2
   )
```
Per CMake [docs](https://cmake.org/cmake/help/latest/command/try_compile.html):

> Compiler options can be passed in like `CMAKE_FLAGS -DCOMPILE_DEFINITIONS=-Werror`.

The `COMPILE_DEFINITIONS` arguments for preprocessor definition flags and risk being silently ignored or parsed incorrectly:

> Specify `-Ddefinition` arguments to pass to `add_definitions()` in the generated test project.

eramongodb (changes requested):

eramongodb:
Additional context: CMake 4.4.0 appears to be exposing a latent bug in how this `try_compile()` command was written, per [5abaf651](https://gitlab.kitware.com/cmake/cmake/-/commit/5abaf65154d477b06e544dc587c3b544db4016a2):

> CMake 4.4 newly complains about unrecognized warning categories, which means that uses which were silently broken now result in a hard error. In the longer term, it would be helpful to introduce a policy to repair improper use of semicolons in `CMAKE_REQUIRED_FLAGS` in order to do what the user probably meant.

The entire `CMAKE_FLAGS` argument (single string, including `echo ...`) was likely being ignored/filtered by CMake prior to the 4.4.0 release.

.evergreen/scripts/compile-libmongocrypt.sh
kevinAlbs: Suggestion was mostly applied. But `version` is kept as `1.21.0-dev` since it is later used as an argument for `BUILD_VERSION`.

kevinAlbs: I also replaced `-C` with `cd` commands since the `git` on the RHEL 7.9 distro appears not to support `-C` ([task](https://spruce.corp.mongodb.com/task/mongo_c_driver_std_matrix_std_c11_rhel7_latest_gcc_compile_patch_d9691e85a8b5f70eca91a1a94d249a5accdc785a_6a5508fb0a1f8a0007b2889e_26_07_13_15_49_19/logs?execution=0)):

```
[2026/07/13 11:51:03.171] Unknown option: -C
[2026/07/13 11:51:03.171] usage: git [--version] [--help] [-c name=value]
[2026/07/13 11:51:03.171]            [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
[2026/07/13 11:51:03.171]            [-p|--paginate|--no-pager] [--no-replace-objects] [--bare]
[2026/07/13 11:51:03.171]            [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
[2026/07/13 11:51:03.171]            <command> [<args>]
```

kevinAlbs:
Latest changes succeed on a [previously failing task](https://spruce.corp.mongodb.com/task/mongo_c_driver_std_matrix_std_c11_macos_14_arm64_clang_compile_patch_d9691e85a8b5f70eca91a1a94d249a5accdc785a_6a550ba79612f800079d25cf_26_07_13_16_00_42/logs?execution=0) and include:
- Remove `git -C` due to lack of support on RHEL 7.9 distro (DEVPROD-37143).
- Use `-Werror=author` in Earthfile following [CMake release notes](https://cmake.org/cmake/help/latest/release/4.4.html#deprecated-and-removed-features):
  > The `-W[no-][error=]dev` command-line options are deprecated. The new spelling is `-W[no-][error=]author`.

eramongodb (approved):

PR #2347, opened 2026-07-13
[r2.3] CDRIVER-6346 Fix calls to Windows APIs when UNICODE defined

kevinAlbs:
Cherry-pick of https://github.com/mongodb/mongo-c-driver/pull/2338 to r2.3.


vector-of-bool (approved):
LGTM

PR #2350, opened 2026-07-15
CDRIVER-6370 add tests and fixes for SCRAM parser

connorsmacd:
See CDRIVER-6370 for details.


src/libmongoc/src/mongoc/mongoc-scram.c
kevinAlbs, line 627: Suggest clarifying that the result of the addition does not need to be checked:

```suggestion
      ptr++; // may refer to one-past-end. But memchr with length 0 is well-defined.
```

kevinAlbs (approved):

PR #2361, opened 2026-07-31
[CDRIVER-6411] Add support for aws sts regional endpoint

bvalente:
Similar to a proposal here https://github.com/mongodb/mongo-go-driver/pull/2376

The default STS endpoint `sts.amazonaws.com` does not work in AWS Gov cloud, as stated in the documentation https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/govcloud-iam.html:
> - In the AWS GovCloud (US) Regions, there is no AWS STS global endpoint. AWS provides Regional AWS STS endpoints.
> - When using the IAM or AWS STS service in AWS GovCloud (US), you must use [AWS GovCloud (US)IAM/AWS STS endpoints](https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-endpoints.html). Use SSL (HTTPS) when you make calls to the IAM or AWS STS service in AWS GovCloud (US) Regions.

Since the driver was hardcoding the global sts endpoint, this driver wouldn't work when using IAM with IRSA (in EKS pods).

Also adding the option to enable/disable the feature with and environmental variable as per the documentation https://docs.aws.amazon.com/sdkref/latest/guide/feature-sts-regionalized-endpoints.html

For this new code to have effect, these environmental variables need to be set
```
AWS_STS_REGIONAL_ENDPOINTS=regional
AWS_REGION='your region'
```

The ideal solution would be to adapt the code to use the official aws sdk, so it can delegate all types of aws authentication. That would also add support for Pod Identity. But it's a major refactorization, like what's being done in https://github.com/mongodb/mongo-go-driver/pull/2508. That is also the pattern being followed by [Python](https://www.mongodb.com/docs/languages/python/pymongo-driver/current/security/authentication/aws-iam/#using-aws-iam-authentication-in-your-application) and [Java](https://www.mongodb.com/docs/drivers/java/sync/current/security/auth/aws-iam/#use-aws-sdk-for-java)

The Go solution was to use the Factory to register a custom authenticator, but that doesn't seem to be a possibility in this codebase. Adapting the code to be able to use a regional sts endpoint is the simplest way to support AWS Gov cloud.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-cluster-aws.c
+++ b/src/libmongoc/src/mongoc/mongoc-cluster-aws.c
@@ -460,6 +460,7 @@ _obtain_creds_from_assumerolewithwebidentity(_mongoc_aws_credentials_t *creds, b
    mongoc_stream_t *fstream = NULL;
    mcommon_string_t *token_file_contents = NULL;
    char *path_and_query = NULL;
+   char *sts_endpoint = NULL;
 
    aws_web_identity_token_file = _mongoc_getenv("AWS_WEB_IDENTITY_TOKEN_FILE");
    aws_role_arn = _mongoc_getenv("AWS_ROLE_ARN");
@@ -522,9 +523,10 @@ _obtain_creds_from_assumerolewithwebidentity(_mongoc_aws_credentials_t *creds, b
                                        aws_role_arn,
                                        MSTR_FMT(jwt_token));
 
-   // send an HTTP request to sts.amazonaws.com.
+   sts_endpoint = _mongoc_get_aws_sts_endpoint();
+
    if (!_send_http_request(true /* use_tls */,
-                           "sts.amazonaws.com",
+                           sts_endpoint,
                            443,
                            "POST",
                            path_and_query,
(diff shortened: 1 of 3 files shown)


src/libmongoc/src/mongoc/mongoc-cluster-aws.c
```diff
@@ -750,6 +755,27 @@ _obtain_creds_from_ec2(_mongoc_aws_credentials_t *creds, bson_error_t *error)
    return ret;
 }
 
+char *
+_mongoc_get_aws_sts_endpoint(void)
+{
+   char *aws_region = _mongoc_getenv("AWS_REGION");
```
kevinAlbs, line 761: I was initially concerned that adding a check for these environment variables might unexpectedly impact users if they had these environment variables set without realizing. So I was considering asking to add a `MONGOC_` prefix or similar.

However, this behavior matches the AWS SDK. [NODE-5477](https://jira.mongodb.org/browse/NODE-5477) made a similar change. So I think this is more likely to match user expectations.

kevinAlbs (commented):
@bvalente thank you for the change and explanation. LGTM 

> The ideal solution would be to adapt the code to use the official aws sdk

That request is tracked in CDRIVER-4516.

kevinAlbs (approved):

PR #2364, opened 2026-08-04
CDRIVER-6353 CDRIVER-6374 client backpressure with baseBackoffMS

connorsmacd:
CDRIVER-6353 (DRIVERS-3535)
CDRIVER-6374 (DRIVERS-3578)

# Summary

This PR implements the latest phase of changes in client backpressure.

## Handshake

The `"backpressure": True` is now `"backpressure": "2"`, where the "2" indicates the client backpressure version supported by the driver.

## Support for `baseBackoffMS`

The server may attach a `baseBackoffMS` field to the error returned on a failed attempt. If present, this value replaces the default of 100ms.

## Backoff Algorithm Changes

There are two notable differences in the backoff algorithm, which implicates `mongoc_retry_backoff_generator_t` in particular:

- The formula for exponential backoff is changed from `base_backoff * 2^(attempt - 1)` to `base_backoff * 2^attempt`, where `attempt` starts at `1`. In other words,the first attempt now uses double the base backoff rather than exactly the base backoff.
- Base backoff is passed on every invocation of `_mongoc_retry_backoff_generator_next` instead of being passed as part of initialization since it can change on each attempt. Notably, this means we cannot precompute the "max attempt" during initialization to clamp the attempt counter. To avoid integer overflow, a similar cap is computed using the same formula where the initial backoff is replaced by the smallest positive value that can be represented by `mlib_duration`, i.e., one microsecond. Any attempt value larger than this cap would result in a backoff that exceeds max backoff.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-retry-backoff-generator.c
+++ b/src/libmongoc/src/mongoc/mongoc-retry-backoff-generator.c
@@ -94,8 +94,8 @@ _mongoc_retry_backoff_generator_next(mongoc_retry_backoff_generator_t *generator
 
    const mongoc_retry_backoff_params_t *const params = &generator->params;
 
-   // `pow` may overflow to INFINITY, so clamp to `backoff_max` before converting to a duration: casting an
-   // out-of-range double to an integer type is undefined behavior.
+   // May overflow to INFINITY, so clamp to `backoff_max` before converting to a duration: casting an out-of-range
+   // double to an integer type is undefined behavior.
    const double backoff_us =
       fmin((double)mlib_microseconds_count(params->backoff_max),
            (double)mlib_microseconds_count(backoff_base) * pow(params->growth_factor, (double)generator->attempt));
--- a/src/libmongoc/src/mongoc/mongoc-retry-backoff-generator.c
+++ b/src/libmongoc/src/mongoc/mongoc-retry-backoff-generator.c
@@ -21,20 +21,27 @@
 
 #include <math.h>
 
+// MONGOC_RETRY_BACKOFF_BASE_MIN is the smallest positive base backoff representable by mlib_duration, and therefore the
+// smallest a caller can supply. It bounds the attempt counter: for any positive base backoff, backoff saturates to
+// `backoff_max` at or before the attempt cap, so incrementing the counter beyond that cannot change a result.
+#define MONGOC_RETRY_BACKOFF_BASE_MIN mlib_duration(1, us)
+#define MONGOC_DURATION_ZERO mlib_duration()
+
 struct _mongoc_retry_backoff_generator_t {
    int attempt;
-   int max_attempt;
+   int attempt_cap;
    mongoc_retry_backoff_params_t params;
    mongoc_jitter_source_t *jitter_source;
 };
 
+// _compute_attempt_cap returns the lowest attempt at which a MONGOC_RETRY_BACKOFF_BASE_MIN base backoff saturates to
+// `backoff_max`.
 static int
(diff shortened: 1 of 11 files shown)


src/libmongoc/tests/test-client-backpressure.c
```diff
@@ -715,6 +715,134 @@ test_backpressure_prose_4(void *ctx)
...
+   mongoc_client_t *client = NULL;
+   {
+      mongoc_uri_t *const uri = test_framework_get_uri();
+      mongoc_uri_set_option_as_bool(uri, MONGOC_URI_RETRYWRITES, true);
+
+      client = test_framework_client_new_from_uri(uri, NULL);
+
+      mongoc_uri_destroy(uri);
+   }
+   test_framework_set_ssl_opts(client);
```
kevinAlbs: Minor: suggest using `test_framework_new_default_client` to simplify. `retryWrites` is enabled by default. 
```suggestion
   mongoc_client_t *client = test_framework_new_default_client();
```

src/libmongoc/tests/test-mongoc-handshake.c
```diff
@@ -1381,8 +1381,14 @@ test_mongoc_handshake_includes_backpressure_flag(void)
    ASSERT(bson_has_field(request_doc, HANDSHAKE_BACKPRESSURE_FIELD));
    bson_iter_t iter;
    ASSERT(bson_iter_init_find(&iter, request_doc, HANDSHAKE_BACKPRESSURE_FIELD));
-   ASSERT(BSON_ITER_HOLDS_BOOL(&iter));
-   ASSERT(bson_iter_bool(&iter));
+   // The value MUST be the string "2", not a literal number 2.
+   ASSERT_WITH_MSG(BSON_ITER_HOLDS_UTF8(&iter),
+                   "expected \"" HANDSHAKE_BACKPRESSURE_FIELD "\" to hold a UTF-8 string, got type %d",
+                   (int)bson_iter_type(&iter));
```
kevinAlbs: Minor: to print the BSON type as a string on failure:

```suggestion
   ASSERT_WITH_MSG(BSON_ITER_HOLDS_UTF8(&iter),
                   "expected \"" HANDSHAKE_BACKPRESSURE_FIELD "\" to hold a UTF-8 string, got type %s",
                   _mongoc_bson_type_to_str(bson_iter_type(&iter)));
```

src/libmongoc/src/mongoc/mongoc-retry-backoff-generator.c
```diff
@@ -86,13 +94,13 @@ _mongoc_retry_backoff_generator_next(mongoc_retry_backoff_generator_t *generator
 
    const mongoc_retry_backoff_params_t *const params = &generator->params;
 
-   if (generator->attempt >= generator->max_attempt) {
-      return _duration_double_multiply(params->backoff_max, jitter);
-   }
-
-   const double backoff_factor = pow(params->growth_factor, (double)generator->attempt - 1);
+   // `pow` may overflow to INFINITY, so clamp to `backoff_max` before converting to a duration: casting an
```
kevinAlbs:
```suggestion
   // May overflow to INFINITY, so clamp to `backoff_max` before converting to a duration: casting an
```

Suggest removing `pow`. I expect `pow` cannot overflow since the growth factor is fixed in the driver and the attempt is capped in `_increment_attempt`. However, if a (malicious?) server returned a very large `baseBackoffMS`, this check seems important to avoid an overflow on the multiplication.

kevinAlbs (approved):
LGTM with minor comments.

PR #2377, opened 2026-08-24
Improve the runtime performance of the Python bsonjs.dumps API

maozguttman:
# Overview

Improve the runtime performance of the Python `bsonjs.dumps` API.

As requested, I have migrated my changes from the `python-bsonjs` repo (https://github.com/mongodb-labs/python-bsonjs/pull/106 pull request) to the `r1.27` branch in this repo.

Below are the results from `benchmark.py`, executed before and after my changes. `bsonjs` was compiled with GCC 14.2.0 on SLES 15 and run on Python 3.13.2.

Before my changes:
```
Timing: bsonjs.dumps(b)
10000 loops, best of 3: 0.05099046230316162
Timing: json_util.dumps(bson.decode(b))
10000 loops, best of 3: 0.38455578312277794
bsonjs is 7.54x faster than json_util

Timing: bsonjs.loads(j)
10000 loops, best of 3: 0.05889510316774249
Timing: bson.encode(json_util.loads(j))
10000 loops, best of 3: 0.41942713782191277
bsonjs is 7.12x faster than json_util
```

After my changes:
```
Timing: bsonjs.dumps(b)
10000 loops, best of 3: 0.024287620093673468
Timing: json_util.dumps(bson.decode(b))
10000 loops, best of 3: 0.39230786077678204
bsonjs is 16.15x faster than json_util

Timing: bsonjs.loads(j)
10000 loops, best of 3: 0.057051923125982285
Timing: bson.encode(json_util.loads(j))
10000 loops, best of 3: 0.41939733177423477
bsonjs is 7.35x faster than json_util
```

My internal tool processes BSON files, some of which are extremely large and are probably not representative of the BSON data that most users work with.
As a result, the performance improvements described below may be most relevant to workloads with similar characteristics.
For example, I have a BSON file with a size of 791,038,613 bytes, containing:
* 7,876,079 dictionaries
* 18,837 lists
* 31,568,552 primitives

It is important to note that these BSON files rarely contain escaped or Unicode characters.

I used Copilot to identify potential areas for runtime optimization. Some of the suggested improvements provided measurable gains in this use case.
I would appreciate it if you could review my changes and, if appropriate, provide feedback, commit them, and include them in a future release.

The overall runtime of my test case was reduced from 19.50 seconds to 4.91 seconds (a 74.8% reduction).
The table below shows the runtime reduction contributed by each change:

| Index | Change                                         | Runtime Improvement (seconds) |
| ----- | ---------------------------------------------- | ----------------------------- |
| 1     | bson_utf8_escape_for_json optimization         | 9.02                          |
| 2     | Nested string copy elimination in BSON visitor | 3.86                          |
| 3     | bson_string_append_printf optimization         | 0.47                          |
| 4     | bson_utf8_validate optimization                | 0.54                          |
| 5     | Eliminate Redundant strlen Calls               | 0.67                          |

# Testing

I ran the Python test suite and added a unit test for the "`bson_utf8_escape_for_json` optimization".
In addition, to validate the "Nested string copy elimination in BSON visitor" changes, I locally enabled the "BSON max len limited" configuration (this change was used only for testing and is not included in the commit). I tested all possible values using sample data containing BSON objects and obtained identical results both with and without my changes.
I also tested the complete set of changes on several BSON files and obtained identical results with and without the optimizations.

# Code changes

## 1. `bson_utf8_escape_for_json` optimization
Problem:
The `bson_utf8_escape_for_json` function is called for every string value and key. It always creates a new `bson_string_t` and processes escape and Unicode characters even when the string does not contain any.
This leads to unnecessary memory allocations and copying.

Fix:
Update the function to:
* Return the original string as is if it contains no escape or Unicode characters (calling function should not free it)
* Allocate and return a modified string only when needed (calling function should free it)

Files changed:
* `bsonjs/bson/bson-string.c`: `bson_string_new_n, bson_string_new`
* `bsonjs/bson/bson-string.h`: `bson_string_new_n`
* `bsonjs/bson/bson-utf8.c`
* `bsonjs/bson/bson.c`: `_bson_as_json_visit_utf8`, `_bson_as_json_visit_regex`, `_bson_as_json_visit_dbpointer`, `_bson_as_json_visit_before`, `_bson_as_json_visit_code`, `_bson_as_json_visit_symbol`, `_bson_as_json_visit_codewscope`

## 2. Nested string copy elimination in BSON visitor
Problem:
For each nested document or array:
* A separate `bson_string_t` is allocated
* The entire child JSON is built in that buffer
* The result is copied into the parent buffer
* The child buffer is then freed
This causes excessive allocations and copying.

Fix:
Use a single shared `bson_string_t` buffer across all nesting levels.

**Thread-Safety Warning:** This change eliminates an unnecessary memory copy of the content from a child node to its parent, optimizing single-threaded performance. 
However, **it is not thread-safe**. This implementation will fail if the code is run in a multi-threaded environment where child nodes are calculated concurrently on separate threads and merged into the parent once those threads complete. 
If multi-threading is required in the future, we can add a boolean configuration flag (`single/multi-threaded`) to support both modes.

Files changed:
* `bsonjs/bson/bson.c`: `_state_str_len`, `bson_json_state_t`, `_bson_as_json_visit_after`, `_bson_as_json_visit_codewscope`, `_bson_as_json_visit_document`, `_bson_as_json_visit_array`, `_bson_as_json_visit_all`

## 3. `bson_string_append_printf` optimization
Problem:
Each call to `bson_string_append_printf`:
* Allocates a temporary buffer
* Uses `vsnprintf` into it
* Appends to the main string
* Frees the temporary buffer
This results in frequent heap allocations.

Fix:
* Use a stack-allocated buffer for `vsnprintf`
* Append directly when the buffer is sufficient
* Fall back to heap allocation only when necessary

Files changed:
* `bsonjs/bson/bson-string.c`: `bson_string_append_printf`, `bson_strdupv_printf`, `bson_strdup_printf`
* `bsonjs/bson/bson-string.h`: `bson_strdupv_printf`

## 4. `bson_utf8_validate` optimization
Problem:
* An inefficient loop was used to check whether a null character exists before the end of a string.

Fix:
* Replaced the loop with `memchr` for more efficient detection.

Files changed:
* `bsonjs/bson/bson-utf8.c`: `bson_utf8_validate`

## 5. Eliminate Redundant `strlen` Calls
Problem:
* Many redundant calls to `strlen` are made even though the string length is already known.

Fix:
* Reuse previously computed string lengths instead of recalculating them with `strlen`.

Files changed:
* `bsonjs/bson/bson-iso8601.c`: `_bson_iso8601_date_format`
* `bsonjs/bson/bson-iter.c`: `bson_iter_visit_all`
* `bsonjs/bson/bson-string.c`: `bson_string_append`, `bson_string_append_c`, `bson_string_append_unichar`, `bson_string_append_printf`, `bson_strdupv_printf`, `bson_strdup_printf`
* `bsonjs/bson/bson-string.h`: `STR_AND_LEN`, `bson_string_append`, `bson_string_append_printf`
* `bsonjs/bson/bson-utf8.c`: `bson_utf8_escape_for_json`
* `bsonjs/bson/bson-utf8.h`: `bson_utf8_escape_for_json`
* `bsonjs/bson/bson.c`: `_bson_as_json_visit_utf8`, `_bson_as_json_visit_decimal128`, `_bson_as_json_visit_double`, `_bson_as_json_visit_undefined`, `_bson_as_json_visit_null`, `_bson_as_json_visit_oid`, `_bson_as_json_visit_binary`, `_bson_as_json_visit_bool`, `_bson_as_json_visit_date_time`, `_bson_as_json_visit_regex`, `_bson_as_json_visit_timestamp`, `_bson_as_json_visit_dbpointer`, `_bson_as_json_visit_minkey`, `_bson_as_json_visit_maxkey`, `_bson_as_json_visit_before`, `_bson_as_json_visit_code`, `_bson_as_json_visit_symbol`, `_bson_as_json_visit_codewscope`, `_bson_as_json_visit_document`, `_bson_as_json_visit_array`, `_bson_as_json_visit_all`, `_bson_iter_validate_before`

### DIFF
--- a/src/libbson/src/bson/bson-iso8601.c
+++ b/src/libbson/src/bson/bson-iso8601.c
@@ -287,6 +287,7 @@ _bson_iso8601_date_format (int64_t msec_since_epoch, bson_string_t *str)
    time_t t;
    int64_t msecs_part;
    char buf[64];
+   uint32_t len;
 
    msecs_part = msec_since_epoch % 1000;
    t = (time_t) (msec_since_epoch / 1000);
@@ -295,23 +296,23 @@ _bson_iso8601_date_format (int64_t msec_since_epoch, bson_string_t *str)
    {
       struct tm posix_date;
       gmtime_r (&t, &posix_date);
-      strftime (buf, sizeof buf, "%Y-%m-%dT%H:%M:%S", &posix_date);
+      len = strftime (buf, sizeof buf, "%Y-%m-%dT%H:%M:%S", &posix_date);
    }
 #elif defined(_MSC_VER)
    {
       /* Windows gmtime_s is thread-safe */
       struct tm time_buf;
       gmtime_s (&time_buf, &t);
-      strftime (buf, sizeof buf, "%Y-%m-%dT%H:%M:%S", &time_buf);
+      len = strftime (buf, sizeof buf, "%Y-%m-%dT%H:%M:%S", &time_buf);
    }
 #else
-   strftime (buf, sizeof buf, "%Y-%m-%dT%H:%M:%S", gmtime (&t));
+   len = strftime (buf, sizeof buf, "%Y-%m-%dT%H:%M:%S", gmtime (&t));
 #endif
 
    if (msecs_part) {
       bson_string_append_printf (str, "%s.%03" PRId64 "Z", buf, msecs_part);
    } else {
-      bson_string_append (str, buf);
+      bson_string_append (str, buf, len);
       bson_string_append_c (str, 'Z');
    }
 }
--- a/src/libbson/src/bson/bson-iter.c
+++ b/src/libbson/src/bson/bson-iter.c
@@ -1899,18 +1899,20 @@ bson_iter_visit_all (bson_iter_t *iter,             /* INOUT */
 {
    uint32_t bson_type = 0;
    const char *key = NULL;
+   uint32_t key_len;
    bool unsupported;
 
    BSON_ASSERT (iter);
    BSON_ASSERT (visitor);
 
    while (_bson_iter_next_internal (iter, 0, &key, &bson_type, &unsupported)) {
-      if (*key && !bson_utf8_validate (key, strlen (key), false)) {
+      key_len = bson_iter_key_len (iter);
+      if (*key && !bson_utf8_validate (key, key_len, false)) {
          iter->err_off = iter->off;
          break;
       }
 
-      if (VISIT_BEFORE (iter, key, data)) {
+      if (VISIT_BEFORE (iter, key, key_len, data)) {
          return true;
       }
 
--- a/src/libbson/src/bson/bson-string.c
+++ b/src/libbson/src/bson/bson-string.c
@@ -31,47 +31,68 @@
 #include <string.h>
 #endif
 
-// `bson_next_power_of_two_u32` returns 0 on overflow.
-static BSON_INLINE uint32_t
-bson_next_power_of_two_u32 (uint32_t v)
+/*
+ *--------------------------------------------------------------------------
+ *
+ * bson_string_new_n --
+ *
+ *       Create a new bson_string_t.
+ *
+ *       bson_string_t is a power-of-2 allocation growing string. Every
+ *       time data is appended the next power of two size is chosen for
+ *       the allocation. Pretty standard stuff.
+ *
+ *       It is UTF-8 aware through the use of bson_string_append_unichar().
+ *       The proper UTF-8 character sequence will be used.
+ *
+ * Parameters:
+ *       @str: a string to copy or NULL.
+ *       @len: number of characters to copy from str.
+ *
+ * Returns:
+ *       A newly allocated bson_string_t that should be freed with
+ *       bson_string_free().
+ *
+ * Side effects:
+ *       None.
+ *
+ *--------------------------------------------------------------------------
+ */
+
+bson_string_t *
+bson_string_new_n (const char *str, uint32_t len) /* IN */
 {
-   BSON_ASSERT (v > 0);
+   bson_string_t *ret;
+   size_t len_sz;
 
-   // https://graphics.stanford.edu/%7Eseander/bithacks.html#RoundUpPowerOf2
-   v--;
-   v |= v >> 1;
-   v |= v >> 2;
-   v |= v >> 4;
-   v |= v >> 8;
-   v |= v >> 16;
-   v++;
-
-   return v;
-}
-
-// `bson_string_ensure_space` ensures `string` has enough room for `needed` + a null terminator.
-static void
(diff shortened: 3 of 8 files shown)


maozguttman:
Hi @blink1073,

As requested, I have created a pull request with my changes to the `r1.27` branch of this repo (instead of https://github.com/mongodb-labs/python-bsonjs/pull/106 pull request).

kevinAlbs:
Hi @maozguttman thanks for the PR. 1.27 is no longer maintained. A performance improvement would be expected in the next minor release (2.6.0). Can you target this fix to the `master` branch? Once merged, I expect python-bsonjs could bump to the latest version of libbson.

Note: C driver 1.29.0 included performance improvements to `bson_utf8_escape_for_json` in CDRIVER-4740.

maozguttman:
Hi @kevinAlbs,

It looks like the code changes I made in the `bson_utf8_escape_for_json` function are conceptually similar to what was previously implemented. However, I took it one step further: if no escaping is needed, the original string is returned directly. This leaves it up to the calling function to determine whether to delete the returned string if its pointer is different than the input string.

I will try to create a pull request on the **master** branch specifically for the **Nested string copy elimination in BSON visitor** improvement. We'll see how that goes, and based on the results, I will follow up with separate pull requests for the remaining changes that had a smaller impact on runtime performance.

kevinAlbs:
> I took it one step further: if no escaping is needed, the original string is returned directly.

As a heads up: I expect that may break documented behavior of [bson_utf8_escape_for_json](https://mongoc.org/libbson/2.5.0/bson_utf8_escape_for_json.html):

> A newly allocated string that should be freed with [bson_free()](https://mongoc.org/libbson/2.5.0/bson_free.html)

But you may be able to modify the private helpers like `mcommon_json_append_escaped` without breaking behavior concerns.

maozguttman:
Hi @kevinAlbs,

It looks like the two changes providing the biggest performance boost the **bson_utf8_escape_for_json optimization** and the **Nested string copy elimination in BSON visitor** are already implemented.
I will just wait for you to bump python-bsonjs to use the latest libbson version.

For the remaining improvements:
* **bson_utf8_validate optimization**. This is a simple and, as far as I can tell, functionally equivalent change (though it still needs to be reviewed). It does not require any API prototype updates. I open a [#2378](https://github.com/mongodb/mongo-c-driver/pull/2378) PR for this.
* **Eliminate Redundant strlen Calls**. Some of these optimizations (like reusing or returning existing strlen values) do require API modifications. I will create a PR specifically for the changes that do not require any API modifications.

blink1073:
I opened https://jira.mongodb.org/browse/PYTHON-6050 to track updating `python-bsonjs` to use {{libbson}} 2.x.

kevinAlbs:
Great, thanks @maozguttman and @blink1073. Closing in favor of #2378 and PYTHON-6050.

PR #2379, opened 2026-08-25
Updates to AGENTS.md from prior Claude work

vector-of-bool:
# Summary

I was working with Claude on mongo-c-driver for a while before the `AGENTS.md` was added. This PR just consolidates a few things I found useful, as well as applies some cleanup and things discussed in the AGENTS.md for libmongocrypt.

The initial merge was performed by Opus 5 on `ultrathink`, with some iteration. I have read over the full changeset and applied manual fixes as well.

# Editorial rules applied

Two rules drove most of the diff, both from review feedback on the sibling `libmongocrypt` AGENTS.md (mongodb/libmongocrypt#1211):

1. **Don't document what the model already knows.** Generic CMake behavior is already baked-in for the model. Removed the multi-configuration generator explanations (which appeared three times), and the `-G` / `CMAKE_BUILD_TYPE` / `BUILD_SHARED_LIBS` / `CMAKE_INSTALL_PREFIX` / `CMAKE_EXPORT_COMPILE_COMMANDS` bullets. The options list is now titled "specific to this project" as a guardrail against this drifting back.
2. **Don't duplicate `CONTRIBUTING.md`.** It stays the source of truth for indentation, commit style, adding a symbol or error code, the `MONGOC_TEST_*` variables, and the Sphinx invocations. Those are now pointers, and AI-usage and portability were added to that pointer list.

The binary directory is now a `<build-dir>` placeholder rather than a prescribed name, keeping only the repo-specific fact that matters: `build/` is a tracked source directory here, so it must not be configured-into.

# Additions

- The *Private Libraries* section was expanded with more specifics about what can be found in each subcomponent.
- It specifically calls out using `ckdint.h` for arithmetic, to steer agents toward using that if there is a risk of arithmetic overflow.
- A section that directs the agent to use the bson-dsl components for decomposing and building BSON documents.
- The *Header Conventions* section documents how/why the `-private.h` headers are used and what they mean.
- The *Navigating libmongoc* section details briefly the large subsystem of the C driver, with pointers to the relevant filenames.
- A brief *Code Style* section that directs toward `CONTRIBUTING.md` AND `.clang-format`.
  - Directs to use formatting script rather than fiddling with whitespace.
  - Notes how to make sure `clangd` is running properly so it doesn't spam the agent with false diagnostics.
  - Documents header style and notes that the formatting script will rewrite/rearrange them.
  - Directs how to create new files, and tells it to duplicate the license header into new files.
- Note the `MONGODB_DEVELOPER` environment variable and how it affects configuration.
- AGENTS.md now calls out the testing skill directly, and notes that CTest is also available for test running.
- In the test skill: 
  - More emphasis that `test-libmongoc` is not built in `all`. This has continually tripped up my agents. Directs it on how to ensure it is built.
  - Direction on running tests via CTest.
  - An explanation of the test label system used in test cases.

# Removals

- Dropped general CMake knowledge from the guidance, only noting the load-bearing CMake configuration parameters. (May need more? It's sparse)
- Direction about how to run Sphinx is now a pointer to CONTRIBUTING.md

# Notes for reviewers

- No Jira ticket is attached.
- Prose wrapping is mixed: reflowed paragraphs are ones this PR touched, matching the deferral of a whole-file reformat discussed in mongodb/libmongocrypt#1211. A separate pass could normalize the file.
- This may be easier to read in the rich diff viewer.

### DIFF
--- a/.agents/skills/running-test-libmongoc/SKILL.md
+++ b/.agents/skills/running-test-libmongoc/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: running-test-libmongoc
-description: Use when running, filtering, or debugging tests in the test-libmongoc executable of the MongoDB C Driver — the custom TestSuite framework in src/libmongoc/tests/TestSuite.h with its own flags (-l, -d, -f), test-naming rules, and JSON spec tests.
+description: Use when running, filtering, or debugging tests in the test-libmongoc executable of the MongoDB C Driver — the custom TestSuite framework in src/libmongoc/tests/TestSuite.h with its own flags (-l, -d, -f), test-naming rules, tags, and JSON spec tests, whether run directly or through CTest.
 ---
 
 # Running `test-libmongoc`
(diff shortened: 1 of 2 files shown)


connorsmacd (commented):
I agree with the sentiment of not telling Claude what it already knows, but I believe many of the instructions in AGENTS.md and the testing skill are incorrect for multi-config generators:

- Configuring with `CMAKE_BUILD_TYPE`.
- The various calls to `cmake --build`, `cmake --install`, and `ctest` lack a `-C <config>` arg.
- The executable path that is referenced for `test-libmongoc` (`<build-dir>/src/libmongoc/test-libmongoc`) would be (I think) `<build-dir>/src/libmongoc/<config>/test-libmongoc` for multi-config.

Perhaps it could be worth evaluating an agent's effectiveness with a multi-config generator. It's possible that it makes the necessary changes to commands and paths without explicit instructions.

PR #2388, opened 2026-09-03
add NEWS for 1.30.9

kevinAlbs:


rcsanchez97 (changes requested):
This PR targets mater, but that doesn't seem right. Shouldn't the target branch be r1.30?

Edit: After writing this it occurred to me that I had these things reversed in my head. Approving now...

rcsanchez97 (approved):

PR #2399, opened 2026-09-04
CDRIVER-6445 Add mongoac basic package features

eramongodb:
Partially resolves CDRIVER-6445. Adds minimal support structures for API doc pages, header installation, and package config files.

API doc pages work exactly the same way as existing bson and mongoc API doc page generation, just extended to the `src/libmongoac` directory. This PR makes no updates to the EVG configuration or release scripts to support publishing mongoac API doc pages yet.

This PR adds the first mongoac public header: `version.h`, generated by CMake via `configure_file()`. Accordingly, the `mongoac::shared` and `mongoac::static` library targets are defined (mirroring `mongoc::shared` and `mongoc::static`). However, these targets are (initially) defined as `INTERFACE` targets; they will eventually need to be refactored and redefined as `IMPORTED` targets later (see: [CDRIVER-6452](https://jira.mongodb.org/browse/CDRIVER-6452)). Note the shared vs. static library requirements also mirror existing requirements for bson and mongoc (e.g. the `MONGOAC_STATIC` preprocessor macro). Only preprocessor macros are defined: functions defined by `version.rs` are deferred to an upcoming PR.

The CMake and pkg-config package config files are defined using the same approach as the [C++ Driver](https://github.com/mongodb/mongo-cxx-driver/blob/19e493b4fdf3bc1fe352cda3a35d471e92aaacfd/src/mongocxx/cmake/CMakeLists.txt). Note these config files currently do not include any system library dependencies (i.e. `libm.so` is required by certain math-related Rust crates pulled in by the Rust Driver). We may consider eventually enumerating such external library dependencies going forward.

### DIFF
--- a/src/libmongoac/cmake/CMakeLists.txt
+++ b/src/libmongoac/cmake/CMakeLists.txt
@@ -1,0 +1,81 @@
+include (CMakePackageConfigHelpers)
+
+# Generate and install modern CMake package config files.
+block ()
+  write_basic_package_version_file (
+    mongoacConfigVersion.cmake
+    VERSION "${PROJECT_VERSION}"
+    COMPATIBILITY SameMajorVersion
+  )
+
+  configure_file (
+    mongoacConfig.cmake.in
+    mongoacConfig.cmake
+    @ONLY
+  )
+
+  install (
+    FILES
+      "${CMAKE_CURRENT_BINARY_DIR}/mongoacConfigVersion.cmake"
+      "${CMAKE_CURRENT_BINARY_DIR}/mongoacConfig.cmake"
+    DESTINATION "${MONGOAC_INSTALL_CMAKEDIR}"
+  )
+endblock ()
+
+# Generate and install pkg-config package config files.
+block ()
+  function (mongoac_install_pkg_config TARGET LINK_TYPE)
+    set (output_name "$CACHE{MONGOAC_OUTPUT_BASENAME}")
+
+    if (LINK_TYPE STREQUAL "STATIC")
+      set (is_static 1)
+      set (pcfilename "${CMAKE_CURRENT_BINARY_DIR}/${output_name}-static.pc")
+    else ()
+      set (is_static 0)
+      set (pcfilename "${CMAKE_CURRENT_BINARY_DIR}/${output_name}.pc")
+    endif ()
+
+    add_custom_command (
+      OUTPUT ${pcfilename}
+      COMMAND
+        ${CMAKE_COMMAND}
+          -D "src_dir=${CMAKE_CURRENT_SOURCE_DIR}"
+          -D "bin_dir=${CMAKE_CURRENT_BINARY_DIR}"
+          -D "prefix=${CMAKE_INSTALL_PREFIX}"
+          -D "includedir=${MONGOAC_INSTALL_INCLUDEDIR}"
+          -D "libdir=${CMAKE_INSTALL_LIBDIR}"
+          -D "output_name=${output_name}"
+          -D "soname=${soname}"
+          -D "version=${PROJECT_VERSION}"
+          -D "is_static=${is_static}"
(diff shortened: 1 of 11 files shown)


src/libmongoac/doc/mongoac/version.rst
```diff
@@ -0,0 +1,24 @@
...
+
+   #define MONGOAC_VERSION            // e.g. "1.2.3-dev"
+   #define MONGOAC_VERSION_MAJOR      // e.g. 1
+   #define MONGOAC_VERSION_MINOR      // e.g. 2
+   #define MONGOAC_VERSION_PATCH      // e.g. 3
+   #define MONGOAC_VERSION_PRERELEASE // e.g. "dev" or ""
+
+   #define MONGOAC_VERSION_HEX // e.g. 0x01020300
+
+   #define MONGOAC_VERSION_CHECK(major, minor, patch)
```
kevinAlbs, line 19: In case this was unintended: consider renaming to `MONGOAC_CHECK_VERSION`.

libbson / libmongoc define `BSON_CHECK_VERSION` / `MONGOC_CHECK_VERSION`.

`MONGOAC_VERSION_CHECK` is consistent with other macros in this file, but `MONGOAC_CHECK_VERSION` might read slightly better and is consistent with libbson / libmongoc.

src/libmongoac/cmake/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+    mongoacConfig.cmake
+    @ONLY
+  )
+
+  install (
+    FILES
+      "${CMAKE_CURRENT_BINARY_DIR}/mongoacConfigVersion.cmake"
+      "${CMAKE_CURRENT_BINARY_DIR}/mongoacConfig.cmake"
+    DESTINATION "${MONGOAC_INSTALL_CMAKEDIR}"
+  )
```
kevinAlbs:
```suggestion
  if ("$CACHE{ENABLE_SHARED}" OR "$CACHE{ENABLE_STATIC}" STREQUAL "ON")
    install (
      FILES
        "${CMAKE_CURRENT_BINARY_DIR}/mongoacConfigVersion.cmake"
        "${CMAKE_CURRENT_BINARY_DIR}/mongoacConfig.cmake"
      DESTINATION "${MONGOAC_INSTALL_CMAKEDIR}"
    )
  endif ()
```

Minor: but suggest using the same condition used to guard installing `mongoac-targets.cmake`, since `mongoacConfig.cmake` references `mongoac-targets.cmake`.

kevinAlbs (approved):
LGTM with minor suggestions.

src/libmongoac/doc/mongoac/version.rst
eramongodb: The intent was to deliberately break consistency with bson/mongoc by treating macros consistently with other library entities in how they are namespace-like-prefixed:

- `mongoac_<component>_<member>` (`mongoac_version_major()`)
- `MONGOAC_<COMPONENT>_<MEMBER>` (`MONGOAC_VERSION_MAJOR`)

This extends to the `check()` function: conceptually `mongoac::version::check()`, thus `MONGOAC_VERSION_CHECK()`.

(The runtime functions, including `mongoac_version_check()`, are currently deferred due to requiring the use of export macros, which will be added in an upcoming PR.)

Feel free to push back on this if deemed insufficiently motivated.

kevinAlbs: Though I like the namespace naming convention, I am slightly more in favor of consistency. I think it is reasonable to assume apps may use both libbson with libmongoac. So this might just be slightly surprising:

```c
// Check required versions:
if (bson_check_version(2, 5, 2) && mongoc_version_check(0, 1, 1)) {
   return 0;
}
```

That said, I do not feel strongly either way. I was originally unsure if this was deliberate. But I do not object to using `MONGOC_VERSION_CHECK`.

kevinAlbs (approved):

PR #2401, opened 2026-09-04
CDRIVER-6445 Add mongoac Rust Crate

eramongodb:
Partially resolves CDRIVER-6445 by adding the `Cargo.toml` config file defining the mongoac Rust crate. CMake integration is deferred to an upcoming PR. Going forward, the lockfile must be updated using `cargo update` within the `src/libmongoac` directory. The enabled set of Rust Driver features are those required by the initial expected featureset for the mongoac; more can be enabled (or disabled) in the future as the need arises.


src/libmongoac/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+
+exclude = [
+  "cmake/**",
+  "doc/**",
+  "include/**",
+]
+
+[dependencies]
+mongodb = {
+  version = "3.8.0",
```
kevinAlbs:
```suggestion
  version = "3.9.0",
```

Suggest upgrading to latest Rust driver release.

kevinAlbs (approved):
LGTM with bumping to the latest Rust driver release.

PR #2402, opened 2026-09-04
CDRIVER-6445 Add mongoac Rust Toolchain integration with CMake

eramongodb:
Resolves CDRIVER-6445 by implementing the CMake integration required to build-and-install the mongoac libraries via the Rust Toolchain (Cargo).

Generator expressions (i.e. `$<CONFIG>`) are used whenever possible to support correct propagation of CMake build type given both single-config (e.g. GNU Makefiles, Ninja) and multi-config (i.e. Visual Studio, Ninja Multi-Config) generators. Note the mongoac library is built with `--release` for all build types that are not `Debug`. A combination of `$<INSTALL_PREFIX>` and GNUInstallDirs ensure correct support for both configure-time `CMAKE_INSTALL_PREFIX=<prefix>` and install-time `--prefix <prefix>`.

Configuration warnings are emitted on non-Linux platforms due to [CDRIVER-6452](https://jira.mongodb.org/browse/CDRIVER-6452). These will need to be replaced as part of [CDRIVER-6453](https://jira.mongodb.org/browse/CDRIVER-6453) and [CDRIVER-6454](https://jira.mongodb.org/browse/CDRIVER-6454).


src/libmongoac/CMakeLists.txt
```diff
@@ -96,41 +96,179 @@ configure_file (
...
+  "${CMAKE_CURRENT_SOURCE_DIR}/Cargo.lock"
+)
+
+set_source_files_properties (${mongoac_headers} PROPERTIES LANGUAGE C)
 
 set (mongoac_file_set_headers
   FILE_SET HEADERS
     BASE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/include" "${CMAKE_CURRENT_BINARY_DIR}/include"
+    FILES ${mongoac_headers}
+  FILE_SET HEADERS
```
kevinAlbs: `mongoac_headers` includes version.h, so this `FILE_SET` seems redundant. Consider removing this `FILE_SET` (and `include_dir`).

tools/cargo-format-all.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+#
+# format.sh
+#
+# Usage:
+#   etc/shfmt-format-all.sh
```
kevinAlbs:
```suggestion
#   tools/cargo-format-all.sh
```

src/libmongoac/CMakeLists.txt
eramongodb: Also updated/replaced the `include_dir` variable with a source+binary pair to clarify the two different include directories being referenced.

tools/cargo-format-all.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+#
+# format.sh
```
kevinAlbs:
```suggestion
# cargo-format-all.sh
```

kevinAlbs (approved):

PR #2412, opened 2026-09-10
Simplify X509_EXTENSION data retrieval

MonkeybreadSoftware:
Removed conditional compilation for OpenSSL version check and simplified the retrieval of extension data.

### DIFF
--- a/src/libmongoc/src/mongoc/mongoc-openssl.c
+++ b/src/libmongoc/src/mongoc/mongoc-openssl.c
@@ -617,12 +617,7 @@ _get_must_staple(X509 *cert)
 
    X509_EXTENSION *const ext = sk_X509_EXTENSION_value(exts, idx);
 
-   // Pre-3.0.0 `ASN1_STRING_get0_data()` expects `T*`, but post-4.0.0 `X509_EXTENSION_get_data()` returns `const T*`.
-#if OPENSSL_VERSION_NUMBER < 0x30000000L
    ASN1_STRING *const ext_data = X509_EXTENSION_get_data(ext);
-#else
-   const ASN1_STRING *const ext_data = X509_EXTENSION_get_data(ext);
-#endif
 
    /* Data is a DER encoded sequence of integers. */
    return _mongoc_tlsfeature_has_status_request(ASN1_STRING_get0_data(ext_data), ASN1_STRING_length(ext_data));


MonkeybreadSoftware:
If you wonder why:

`/Users/cs/Development/Libs/mongo/mongo-c-driver/src/libmongoc/src/mongoc/mongoc-openssl.c:628:71: error: passing 'const ASN1_STRING *const'
      (aka 'const struct asn1_string_st *const') to parameter of type 'ASN1_STRING *' (aka 'struct asn1_string_st *') discards qualifiers [-Werror,-Wincompatible-pointer-types-discards-qualifiers]
  628 |    return _mongoc_tlsfeature_has_status_request(ASN1_STRING_get0_data(ext_data), ASN1_STRING_length(ext_data));
      |                                                                       ^~~~~~~~
`

We build against OpenSSL 3.5.8 and the **const** keyword is no longer needed. So the branch seems no longer necessary.

eramongodb:
Hello, @MonkeybreadSoftware. The proposed changes break forward compatibility with OpenSSL 4.0. See [CDRIVER-6308](https://jira.mongodb.org/browse/CDRIVER-6308) (https://github.com/mongodb/mongo-c-driver/pull/2290) for details. However, those changes are not directly related to the reported issue.

> We build against OpenSSL 3.5.8...

CMake [detects](https://github.com/mongodb/mongo-c-driver/blob/8b93d5956fe5b260ce73dc9269dcdeac7cf84011/src/libmongoc/CMakeLists.txt#L268-L272) the existence of the `ASN1_STRING_get0_data` symbol and [defines](https://github.com/mongodb/mongo-c-driver/blob/8b93d5956fe5b260ce73dc9269dcdeac7cf84011/src/libmongoc/CMakeLists.txt#L275) the `MONGOC_HAVE_ASN1_STRING_GET0_DATA` [feature macro](https://github.com/mongodb/mongo-c-driver/blob/8b93d5956fe5b260ce73dc9269dcdeac7cf84011/src/libmongoc/src/mongoc/mongoc-config.h.in#L151-L158) during configuration, selecting the [const-qualified](https://docs.openssl.org/3.5/man3/ASN1_STRING_length/) `ASN1_STRING_get0_data()` when it is available:

```c
const unsigned char *ASN1_STRING_get0_data(const ASN1_STRING *x);
```

The reported compiler error suggests `MONGOC_HAVE_ASN1_STRING_GET0_DATA` is not correctly defined for OpenSSL 3.5.8 (e.g. by bypassing CMake configuration when building the mongoc library, using a mongoc library built against the wrong OpenSSL library version, or reusing a stale CMake config cache) and selecting the non-const-qualified `ASN1_STRING_data()` [as a fallback](https://github.com/mongodb/mongo-c-driver/blob/8b93d5956fe5b260ce73dc9269dcdeac7cf84011/src/libmongoc/src/mongoc/mongoc-openssl.c#L68-L70) for backward compatibility with OpenSSL 1.0.2u:

```c
unsigned char *ASN1_STRING_data(ASN1_STRING *x);
```

Per OpenSSL documentation, `ASN1_STRING_data()` "is deprecated: applications should use `ASN1_STRING_get0_data()` instead." The `MONGOC_HAVE_ASN1_STRING_GET0_DATA` feature macro is consistent with this recommendation. When building mongoc against OpenSSL 3.5.8, const-qualification is correct and consistent with the `ASN1_STRING_get0_data()` API.

Please confirm that `MONGOC_HAVE_ASN1_STRING_GET0_DATA=1` is correctly defined when configuring and building the mongoc library against OpenSSL 3.5.8, either by default in `mongoc-config.h` after CMake configuration or (when bypassing CMake) by an explicit user-defined compile definition.

MonkeybreadSoftware:
Thank you for your analysis.

From cmake configuration:
`-- Found OpenSSL: /Users/cs/Development/Libs/openssl3/linux64arm/lib/libcrypto.so (found version "3.5.8")  
-- libmongoc TLS : OpenSSL
-- Looking for ASN1_STRING_get0_data
-- Looking for ASN1_STRING_get0_data - not found`

for some reason the openssl is found, but not the symbols.

*later*

looks like zlib is not found, so the test app can't link. I've added -L to the folder and -lz for the C flags and then it works.
Strange. But thanks. This is solved.

PR #2416, opened 2026-09-15
CDRIVER-6445 Add mongoac build.rs and version.rs

eramongodb:
Resolves CDRIVER-6445 by introducing the first mongoac Rust crate `version.rs` and the first set of exported symbols.

---

The hand-written `export.h` header declares export macros [in the same manner](https://mongocxx.org/api/mongocxx-4.5.3/bsoncxx_2v1_2config_2export_8hpp.html) as the C++ Driver's export header (`MONGOAC_ABI_EXPORT`). Its implementation is slightly simplified due to its use _only_ to declare export macros for the importer; symbol export control is handled Rust-side by Cargo and do not require export macros. As usual, the `MONGOAC_STATIC` preprocessor macro must be defined when compiling against the static library.

---

This PR introduces `src/libmongoac/builds.rs`, which is automatically invoked as part of the `cargo rustc` command (before building mongoac crates) and is primarily responsible for generating the `config.rs` crate. This crate is analogous to [mongoc-config.h.in](https://github.com/mongodb/mongo-c-driver/blob/e64b43729d73a7d70b541dfe374bd4a6a29b133b/src/libmongoc/src/mongoc/mongoc-config.h.in) and [config.hpp.in](https://github.com/mongodb/mongo-cxx-driver/blob/13f379718db7b62d449d30a3e2572dc5c8cddf64/src/bsoncxx/lib/bsoncxx/v1/config/config.hpp.in): it declares constants which are dependent on the CMake build configuration such as the library version, the build type (e.g. Debug, Release, etc.) and link type (static vs. shared).

Note some of the required env vars used to forward CMake variables to Cargo were already introduced by https://github.com/mongodb/mongo-c-driver/pull/2402; this PR only adds the `MONGOAC_VERSION_FULL` env var. Only the version-related constants are used for now; build type and link type will be used to specify mongoac client metadata. Each `cargo:rerun-if-env-changed=<var>` printed to stdout instructs Cargo to rerun the build script whenever `<var>` has changed. `cargo:rerun-if-changed=<file>` is the same, but for file dependencies.

---

This PR establishes the general pattern to export a function in mongoac:

```rust
#[unsafe(no_mangle)]
pub extern "C" fn mongoac_example() -> i32 {
    42
}
```

The [no_mangle](https://doc.rust-lang.org/reference/abi.html#the-no_mangle-attribute) and [`extern`](https://doc.rust-lang.org/std/keyword.extern.html) are both required to achieve behavior analogous to `extern "C"` in C++: `no_mangle` controls the calling convention whereas `extern "C"` controls the symbol name. Though not strictly necessary (since the `mongoac` Rust crate [is not published](https://github.com/mongodb/mongo-c-driver/blob/e64b43729d73a7d70b541dfe374bd4a6a29b133b/src/libmongoac/Cargo.toml#L6)), using `pub` (together with the explicit `mongoac_*` prefix) should help explicitly denote public API functions from internal-use-only functions in Rust code.

> [!NOTE]
> The [Foreign Function Interface](https://doc.rust-lang.org/nomicon/ffi.html) page in Rustonomicon is somewhat outdated, e.g. mongoac uses Cargo edition 2024, which made the `no_mangle` attribute `unsafe`, hence `#[unsafe(no_mangle)]`.

The `version.rs` crate is unique in mongoac due to its C header being manually-written. All other crates exporting symbols are expected to have their corresponding "cargo header" generated using [cbindgen](https://github.com/mozilla/cbindgen). The `config.rs` crate is internal and not exposed in the public API, similar to [`config.hh`](https://github.com/mongodb/mongo-cxx-driver/blob/13f379718db7b62d449d30a3e2572dc5c8cddf64/src/mongocxx/lib/mongocxx/private/config/config.hh.in) in the C++ Driver.


kevinAlbs (approved):

src/libmongoac/build.rs
```diff
@@ -0,0 +1,82 @@
...
+        .expect("MONGOAC_VERSION_FULL is unset: build via the CMake configuration");
+    println!("cargo:rerun-if-env-changed=MONGOAC_VERSION_FULL");
+
+    let build_type = std::env::var("MONGOAC_CMAKE_BUILD_TYPE").unwrap_or_default();
+    println!("cargo:rerun-if-env-changed=MONGOAC_CMAKE_BUILD_TYPE");
+
+    let link_type = std::env::var("MONGOAC_LIBRARY_TYPE").unwrap_or_default();
+    println!("cargo:rerun-if-env-changed=MONGOAC_LIBRARY_TYPE");
+
+    // Assume valid SemVer string guarantees: regex not required.
```
connorsmacd, line 30: Out of curiosity, are we avoiding regex here so we don't have to pull in the regex crate, or is it just a preference?

src/libmongoac/src/version.rs
```diff
@@ -0,0 +1,51 @@
...
+pub extern "C" fn mongoac_check_version(
+    required_major: i32,
+    required_minor: i32,
+    required_patch: i32,
+) -> bool {
+    let major = mongoac_version_major();
+    let minor = mongoac_version_minor();
+    let patch = mongoac_version_patch();
+
+    major > required_major
```
connorsmacd: A more recent major version could come with breaking changes, so do we actually want to allow that here?

connorsmacd (approved):
LGTM. I just left a couple of questions to get some clarification.

src/libmongoac/build.rs
eramongodb: It is for simplicity only. The "proper" regex parsing is already [handled by CMake](https://github.com/mongodb/mongo-c-driver/blob/e64b43729d73a7d70b541dfe374bd4a6a29b133b/src/libmongoac/CMakeLists.txt#L11-L15), so it would be redundant work to regex-parse the version string again.

src/libmongoac/src/version.rs
eramongodb: I must admit this definition is merely copying what exists for [bson](https://github.com/mongodb/mongo-c-driver/blob/e64b43729d73a7d70b541dfe374bd4a6a29b133b/src/libbson/src/bson/version.h.in#L91-L95) and [mongoc](https://github.com/mongodb/mongo-c-driver/blob/e64b43729d73a7d70b541dfe374bd4a6a29b133b/src/libmongoc/src/mongoc/mongoc-version.h.in#L97-L101).

Given they are all documented as merely validating the version is "greater than or equal to" the required version, major version compatibility will require an additional explicit `mongoac_version_major() == required_major` check. Just in case, added a note to `version.rst` with examples for clarity.

PR #2417, opened 2026-09-15
CDRIVER-6445 Add mongoac_string_view_t

eramongodb:
Related to CDRIVER-6445. This PR introduces the `generate-crate-headers` tool and the first "cargo header" generated by `cbindgen` to generate `string.h` for the new `string.rs` crate.

> [!NOTE]
> Only `StringViewT` is added in this PR to minimize scope. The owning variant `StringT` is deferred to an upcoming PR.

Following this PR, the current state of a standalone mongoac install prefix looks as follows:

```
<prefix>
├── include
│   └── mongoac-0.1.0
│       └── mongoac
│           ├── export.h
│           ├── string.h
│           └── version.h
└── lib
    ├── cmake
    │   └── mongoac-0.1.0
    │       ├── mongoacConfig.cmake
    │       ├── mongoacConfigVersion.cmake
    │       └── mongoac-targets.cmake
    ├── libmongoac0.a
    ├── libmongoac0.so -> libmongoac0.so.0
    ├── libmongoac0.so.0 -> libmongoac0.so.0.1.0
    ├── libmongoac0.so.0.1.0
    └── pkgconfig
        ├── mongoac0.pc
        └── mongoac0-static.pc
```

and its current list of exported symbols is:

```
// mongoac_string_view_destroy (removed)
mongoac_version
mongoac_version_hex
mongoac_version_major
mongoac_version_minor
mongoac_check_version
mongoac_version_patch
mongoac_version_prerelease
```

---

This PR establishes the general pattern by which (most) public headers are implemented in mongoac: the `generate-crate-headers` tool (declared as a [workspace member](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields)) scans the `src/libmongoac/src` directory for public crates (not under `private/`) and not excluded by either `SKIP_CARGO_HEADERS` or `SKIP_FORWARD_HEADERS`, then generates both the normal header `mongoac/example.h` and (when appropriate) the forward header `mongoac/example-fwd.h`, collectively referred to as the "crate headers" for `example.rs`. No manual header maintenance or synchronization with crates is required: cargo headers are automatically generated by Cargo via CMake using the `mongoac_crate_headers_target` custom target.

> [!IMPORTANT]
> The `cbindgen` dependency is handled by Cargo; the user does not need to provide a `cbindgen` executable in their environment.

`generate-crate-headers` is also responsible for:

* renaming `ExampleT` structs to `mongoac_example_t` for consistency with existing C Driver conventions.
* listing required include directives required by a given crate header.
* any additional modifications to crate headers that may be added in the future (e.g. IWYU pragma comments?).

> [!NOTE]
> mongoac structs use the `T`-suffix to
>
> 1) follow the PascalCase-equivalent convention of the snake_case `_t` suffix used by the C Driver, and
> 2) easily disambiguate mongoac FFI structs from Rust Driver (the "mongodb" crate) structs going forward without forcing explicit qualification or alternative identifiers (e.g. `ClientT` is mongoac; `Client` is mongodb).
>
> `ExampleT` and `mongoac_example_t` are synonymous. The former will often be used as shorthand for the latter.

---

~On occasion (specifically for _non-opaque, non-owning, view-like structs_ such as `StringViewT`), a function stub must be declared and exported to coerce cbindgen to include the struct in the cargo header. Such function stubs (usually a single `destroy()`) are clearly documented with "Does nothing." and a short explanation in both API docs and source code.~ The `export.include` cbindgen option may be used to list structs which must be explicitly exported due to no exported function referencing the struct. These are only expected to be necessary for `StringViewT` + the upcoming `BsonViewT`, and `ArrayView<Type>T` structs.


eramongodb:
evergreen patch

> [!NOTE]
> Looks like manual patch commands do not work yet for stacked PRs even when "Manual Testing" is enabled in EVG project settings.

src/libmongoac/tools/generate-crate-headers/src/main.rs
```diff
@@ -0,0 +1,224 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::path::{Path, PathBuf};
+
+macro_rules! includes {
```
connorsmacd: I was initially quite confused when I saw this used below because I confused it with the standard [`include`](https://doc.rust-lang.org/std/macro.include.html) macro, which obviously does something much different. We may want to rename this to avoid further confusion.

Also, a brief comment with an example of what this macro expands to might be helpful since Rust macro syntax can be a bit "write-only" sometimes, e.g.:

```rust
// `includes!["stdint.h", "stdio.h"]` -> `vec!["stdint.h".into(), "stdio.h".into()]`
```

src/libmongoac/tools/generate-crate-headers/src/main.rs
```diff
@@ -0,0 +1,224 @@
...
+        rel_stem
+            .file_name()
+            .and_then(|s| s.to_str())
+            .expect("invalid UTF-8")
+    ));
+
+    let mut config = default_config();
+
+    // Only forward declarations of opaque structs.
+    config.export.item_types = vec![cbindgen::ItemType::OpaqueItems];
```
connorsmacd:
```suggestion
    let config = {
        let mut c = default_config();
        // Only forward declarations of opaque structs.
        c.export.item_types = vec![cbindgen::ItemType::OpaqueItems];
        c
    };
```

Suggest using a block expression to encapsulate the construction of `config` and limit its `mut` scope.

src/libmongoac/tools/generate-crate-headers/src/main.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    // `path/to/crate` -> `<include_dir>/path/to/crate.h`
+    let header_path = include_dir.join(&rel_stem).with_extension("h");
+    if let Some(parent) = header_path.parent() {
+        std::fs::create_dir_all(parent)
+            .unwrap_or_else(|_| panic!("failed to create include directory: {}", parent.display()));
+    }
+
+    // Default cbindgen configuration for all crates.
+    let mut config = default_config();
```
connorsmacd: Similar to the above, consider encapsulating the construction of `config`. This is arguably more applicable than my previous suggestion since the construction is more extensive.

src/libmongoac/tools/generate-crate-headers/src/main.rs
```diff
@@ -0,0 +1,224 @@
...
+    let src_dir = std::env::var("MONGOAC_SRC_DIR")
+        .expect("MONGOAC_SRC_DIR is required (the Rust source root directory)");
+    let include_dir = std::env::var("MONGOAC_INCLUDE_DIR")
+        .expect("MONGOAC_INCLUDE_DIR is required (the output directory for generated headers)");
+
+    let src_path = Path::new(&src_dir);
+    let include_path = Path::new(&include_dir);
+
+    let mut files = Vec::new();
+    find_crates(src_path, &mut files);
```
connorsmacd, line 252: I would normally expect this to be `let files = find_crates(src_path);`, but I see `find_crates` takes an out-param to facilitate recursion. I'm inclined to suggest using the current implementation of `find_crates` as a helper function to that can be used to implement my ideal version of the function, but I recognize that it might not be worth the effort in tooling code. That said, since it is just tooling code, it might be find to just eat the cost of `files.extend(find_crates(path))` to avoid the out-param entirely. Food for thought.

src/libmongoac/tools/generate-crate-headers/src/main.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    // Generate crate headers in parallel.
+    let mut handles = Vec::new();
+    for path in files {
+        let include_dir = include_path.to_path_buf();
+        let src_dir = src_path.to_path_buf();
+        handles.push(std::thread::spawn(move || {
+            generate_crate_header(&path, &src_dir, &include_dir);
+        }));
+    }
```
connorsmacd:
```suggestion
    let handles = files
        .into_iter()
        .map(|path| {
            let include_dir = include_path.to_path_buf();
            let src_dir = src_path.to_path_buf();
            std::thread::spawn(move || {
                generate_crate_header(&path, &src_dir, &include_dir);
            })
        })
        .collect::<Vec<_>>();
```

Suggest a more declarative way to initialize `handles`.

connorsmacd (approved):
LGTM. I left some comments that mostly reflect my (opinionated) preference for declarative style. I'm still trying to figure out my own Rust "style", so thank you for baring with me. I recognize this might be overkill feedback for tooling code.

src/libmongoac/doc/mongoac/mongoac_string_view_destroy.rst
eramongodb: abc

src/libmongoac/tools/generate-crate-headers/src/main.rs
eramongodb: Good catch. Replaced the ad-hoc macro with the same explicit borrow + `to_string()` pattern as in `rename_structs()`.

src/libmongoac/tools/generate-crate-headers/src/main.rs
eramongodb: Opted to leave this one as-is for now and defer its refactor to future consideration.

eramongodb (commented):
Re-requesting due to substantial changes.

Reviewing the cargo header generation script in response to feedback revealed a few more opportunities for improvement, including 1) inclusion of license header, 2) removing the noisy/redundant "Generated with cbindgen" comment in favor of the "Generated by main.rs" comment, and 3) the `export.includes` configuration option to avoid the need for do-nothing function stubs. The per-crate `configure()` function has been extended accordingly.

connorsmacd (approved):
LGTM!

PR #2424, opened 2026-09-17
CDRIVER-4530 Remove obsolete cipher suite incompatibility workaround on Windows

eramongodb:
Related to CDRIVER-4530. Although the C Driver does not yet support non-RSA certificates on Windows, the corresponding EVG tasks have since https://github.com/mongodb/mongo-c-driver/pull/1155 1) migrated from the old `windows-vs*` distros to the `windows-vsCurrent` and `windows-2022-latest` distros, 2) migrated to a uv-based Python toolchain, and 3) depends on improved DET scripts which also use uv-based dependency management. It seems some combination of these three improvements have addressed the cipher suite incompatibility issue for the Windows distros currently in use (according to discussions in [SERVER-79172](https://jira.mongodb.org/browse/SERVER-79172), the root cause is more complicated than what CDRIVER-4530 suggested, involving SChannel vs. [OpenSSL's SECLEVEL](https://docs.openssl.org/3.0/man3/SSL_CTX_set_security_level/)).

In short, this workaround, which now conflicts with latest DET scripts, is obsolete and no longer needed.


kevinAlbs (approved):
The simplification is much appreciated.

PR #2427, opened 2026-09-19
CDRIVER-6472 fix mingw-w64 cross-compilation on linux

yeads:
Cross-compiling for Windows with MinGW-w64 on Linux fails because mixed-case Windows header and library names do not match the lowercase filenames on a case-sensitive filesystem.

Use lowercase names for the Windows headers and the bcrypt, shlwapi, dnsapi, and dbghelp libraries. This also fixes the BCryptDeriveKeyPBKDF2 feature check, which previously failed at link time.

Only require utf8proc when cryptography is enabled, avoiding an unnecessary dependency in builds with TLS disabled.

Validation:
- Built Windows x64 static libraries, DLLs, and test-libmongoc with both Schannel/CNG and OpenSSL.
- Enabled SSPI, SRV, bundled zlib, bundled utf8proc, and AWS authentication.
- Each configuration passed 46 selected tests under Wine; four server-dependent authentication tests were skipped.
- Repository formatting checks and git diff --check passed.

The full test suite and native Windows runtime behavior have not been verified.


rcsanchez97:
Thanks for contributing this fix. I have reviewed and everything looks good. There was some additional logic needed around addition of the `UTF8PROC_STATIC` compile definition, which I went ahead and added and pushed. I have initiated a second CI run (to include the additional change that I made). If it comes back clean (which I expect, as the first one I ran with the initial changes was clean), then I will go ahead and merge this PR.

rcsanchez97 (approved):

